Issue #976, opened 2020-08-08
📎 Additional lint rules from IntelliJ


sebmck:
I've recently switched to Webstorm and was impressed with a lot of the rules they offer with code inspection. I haven't seen some of these in ESLint.

- ~Unnecessary `continue` as the last statement in a loop~
- Simplify statement or boolean expression
- Expression statement that is not assignment or call
- Single character regex alternations can be a charset
- Unused shadowed bindings

aravind1078:
@sebmck i wanna pick `Unnecessary continue` already started working on it, missing few cases currently will come up with a pr soon.

aravind1078:
do we want to remove the continue statement in `unnecessary continue`  rule or we just tell them that the continue is unnecessary here!

ludofischer:
I am currently trying to write `Expression statement that is not assignment or call`. Down from 12115 false positives to just 329... For the benefit of others attempting a similar task, it helps to know that the AST (Abstract Syntax Tree) format looks similar to [ESTree](https://github.com/estree/estree) and you can paste code into https://astexplorer.net/ to get an idea of how it’s parsed (I think leaving the parser set as acorn gives a good approximation of the Rome AST).

ematipico:
@ludofischer Rome has also a command called `parse` that write to screen the AST:

```
./rome parse ./test.ts
```

PR #1148, opened 2020-08-25
feat: added lint rule noUnnecessaryContinue

aravind1078:
## Summary

part of #976 
added lint rule nounnecessayContinue

## Test Plan
`./rome  test`


internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+		continue;
+		}
+		console.log(i);
+		i--;
+	}
+	"
+	"
+	while(i){
+		continue;
+		console.log(4);
```
ematipico, line 51: Come to think of it, this code is valid for this rule but it's invalid in general because `console.log(i)` will never be executed. Not sure, maybe it could be an idea for another lint rule.

internal/compiler/lint/rules/js/noUnnecessaryContinue.ts
```diff
@@ -0,0 +1,98 @@
...
+	name: "js/noUnNecessaryContinue",
+	enter(path) {
+		const {node} = path;
+		if (node.type !== "JSContinueStatement") {
+			return signals.retain;
+		} else if (
+			node.type === "JSContinueStatement" &&
+			isContinueUnNecessary(path)
+		) {
+			return path.addFixableDiagnostic(
```
ematipico, line 88: I believe we can provide an auto fix by removing the `continue;'

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
aravind1078: yeah, probably we could tell the user that with continue statement being present all the code below it in that block is kind of `dead code` or `non-reachable code`.

website/src/docs/lint/rules/js/noUnnecessaryContinue.md
```diff
@@ -0,0 +1,175 @@
+
+			---
+			title: Lint Rule js/noUnnecessaryContinue
+			layout: layouts/rule.liquid
+			showHero: false
+			description: MISSING DOCUMENTATION
```
ematipico, line 6: Description is missing, we should provide a brief explanation

website/src/docs/lint/rules/js/noUnnecessaryContinue.md
```diff
@@ -0,0 +1,175 @@
...
+			eleventyNavigation:
+				key: lint-rules/js/noUnnecessaryContinue
+				parent: lint-rules
+				title: js/noUnnecessaryContinue
+			---
+
+			# js/noUnnecessaryContinue
+
+			MISSING DOCUMENTATION
```
ematipico, line 15: Same here, we should provide a more extensive explanation of the rule

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+]
+valid: [
+	"
+	while(i){
+		if(i>5){
+		continue;
+		}
+		console.log(i);
+		i--;
+	}
```
ematipico, line 46:
```suggestion
	while (i) {
		if (i > 5) {
			continue;
		}
		console.log(i);
		i--;
	}
```

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+        console.log('sss');
+        continue;
+    }
+    else if(i>=5&&i<8){
+        console.log('s');
+    }
+    else {
+        console.log('dsd');
+    }
+	}
```
ematipico, line 29:
```suggestion
	for (let i=0; i < 10; i++) {
    	if(i>5) {
        	console.log('sss');
        	continue;
    	}
    	else if (i >= 5 && i < 8){
        	console.log('s');
    	}
	    else {
        	console.log('dsd');
    	}
	}
```

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+			if(conditionX){
+				console.log('log');
+				continue;
+			}
+			console.log('log');
+			if(conditionY){
+				console.log('log');
+			}
+		}
+	}
```
ematipico, line 66:
```suggestion
	while (condition) {
		if (conditionZ) {
			if (conditionX) {
				console.log('log');
				continue;
			}
			console.log('log');
			if (conditionY) {
				console.log('log');
			}
		}
	}
```

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+	"
+	"
+	loop: while(1){
+		forLoop: for(let i=0;i<5;i++){
+			if(someCondition){
+				continue loop;
+			}
+		}
+	}
```
ematipico, line 75:
```suggestion
	loop: while(1) {
		forLoop: for(let i = 0; i < 5; i++){
			if (someCondition) {
				continue loop;
			}
		}
	}
```

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
+filename: "file.ts"
+invalid: [
+	"
+	loop: for(let i=0;i<5;i++){
+		continue loop;
+	}
```
ematipico, line 6:
```suggestion
	loop: for (let i = 0; i < 5; i++) {
		continue loop;
	}
```

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+	"
+	"
+	while(i--)
+ 	continue;
```
ematipico, line 10:
```suggestion
	while (i--)
 		continue;
```

internal/compiler/lint/rules/js/noUnnecessaryContinue.test.rjson
```diff
@@ -0,0 +1,86 @@
...
+	"
+	"
+	while(1){
+		continue;
+	}
```
ematipico, line 15:
```suggestion
	while (1) {
		continue;
	}
```

ematipico (changes requested):
I left some suggestions around formatting. These examples will go on our website, so I think it's worth presenting them in fashion manner. Also, some documentation is missing.

I leave the final 👍  to @bitpshr

website/src/docs/lint/rules/js/noUnnecessaryContinue.md
aravind1078: @ematipico all these were auto-generated so I didn't have a look at them. can we edit these files directly or do we have to add these descriptions somewhere else so that they get populated here?

ematipico: Yes you can edit this file directly, you need to replace the placeholder "MISSING DOCUMENTATION"

sebmck:
Nice! I think this could be simplified with the usage of the `getCompletionRecords` from `js-ast-utils`:

```
import {getCompletionRecords} from "@internal/js-ast-utils";

const {node, parent} = path;

if (node.type === "ForStatement" || ...) {
    const label = parent.type === "JSLabeledStatement" ? parent.label.name : undefined;
    for (const record of getCompletionRecords(node)) {
        if (record.type === "COMPLETION" && record.node.type === "ContinueStatement") {
            const contin = record.node;
            if (contin.label === undefined || contin.label.name === label) {
                // Unnecessary continue
            }
        }
    }
}
``

We use this helper in a bunch of places to fetch statements that cause scripts to stop terminating. I think this will also handle more cases than what's in this PR.

**Comment by aravind1078** [comment]
@sebmck am not sure if `getCompletionRecords` function will help me here because it gives all the possible completion Records which also includes valid continue statements... for example
```
while (i) {
          if (i > 5) {
		continue;
	}
	console.log(i);
	i--;
}
```
in the above scenario, the function `getCompletionRecords` gives continue as one of the completion records but, the `continue` is valid here. `getcompletionRecords` can be used to determine whether continue is the last statement in one of the paths... which is what am checking in `isContinueTheLastStatement ` function I wrote but, then again we need to check whether the continue is inside the last completion path which is the only case when `continue` is unnecessary. am not sure if we can determine whether the continue is inside the last path with the completion records we get from `getCompletionRecords`. correct me if am wrong...

sebmck:
Ahh right. Looks good to me then! I think there's potential here to abstract some of this logic that will be useful for other lint rules. I believe ESLint has something called a code path. Thank you!

aravind1078:
> Ahh right. Looks good to me then! I think there's potential here to abstract some of this logic that will be useful for other lint rules. I believe ESLint has something called a code path. Thank you!

Sure i will check in eslint repo thanks!

PR #1150, opened 2020-08-26
feat: support tslint rule references

### LINKED ISSUE #1130: 💭 lint/ts/useInterfaces has incorrect documentation
I've been reading about already implemented rules and found that [`ts/useInterfaces`](https://romefrontend.dev/docs/lint/rules/ts/useinterfaces/) rather confusing. It openly states that the rule is equivalent to [`jsx-filename-extension`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-filename-extension.md) which is about file extensions, not typescript interfaces as some may assume from the rule's name. From examples we can see that this rule also reports errors for files with `.js` extension containing jsx elements which is not about typescript(the rule is under `ts` rules). Can we please clarify the rule's name or move it outside of `ts` rules to avoid further confusion?

PS: amazing work 👍

bitpshr:
## Summary

This pull request updates the `lint-rules-docs` generation script to support TSLint rule references, and updates the documentation and referenced rule for the `ts/useInterfaces` rule to prevent confusion.

Resolves #1130

## Test Plan

```sh
./rome test
```

### DIFF
--- a/scripts/generated-files/lint-rules-docs.ts
+++ b/scripts/generated-files/lint-rules-docs.ts
@@ -40,15 +40,16 @@ function highlightPre(filename: string, code: string): string {
 }
 
 // Extract the description field from the docs frontmatter
-export function extractESLintRuleInfo(
+export function extractLintRuleInfo(
 	content: string,
+	type: "eslint" | "tslint" = "eslint",
 ):
 	| undefined
 	| {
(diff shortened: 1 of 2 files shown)


sebmck:
Nice! Thank you!

PR #1152, opened 2020-08-26
fix: safely update per-token context

### LINKED ISSUE #1134: 🐛 Lint fails with `No context found` error
## Environment

```
$ rome rage --summary
Environment Variables
  Object {
    COLORTERM: "truecolor"
    LANG: "en_US.UTF-8"
    TERM: "xterm-256color"
    TERM_PROGRAM: "vscode"
  }

User Config
  unset

Rome Version
  10.0.4-beta

Node Version
  14.5.0

Platform
  darwin x64 18.7.0

Terminal Features
  Object {
    background: "unknown"
    colorDepth: 24
    columns: 162
    cursor: true
    hyperlinks: true
    isTTY: true
    unicode: true
  }

Client Flags
  Object {
    clientName: "cli"
    cwd: "/Users/user/projects/test-rome"
    realCwd: "/Users/user/projects/test-rome"
    silent: false
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 91_201_536
      pid: 7_293
      uptime: 0.955878686
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 91_201_536
        ownedBytes: 0
        ownedFileCount: 0
        pid: 7_293
        uptime: 0.955763755
      }
    ]
  }
```

## Steps To Reproduce

1. Add a file with this content:
```
const Test = () => {
  const a = 2;
  const b = 2;
  return (
    <div>
      {a && (b ? <div /> : <div />)}
      <div />
    </div>
  );
};
```
2. Run `rome check`

## Expected Results
Lint should run without failing.

## Actual Results
Lint fails with `✖ No context found` error.

bitpshr:
## Summary

This pull request updates the JavaScript parser to safely handle per-token context updates in the event that the parser context stack is empty. I'm a little suspicious that this fix may be paving over a deeper issue around incorrect context population in the first place, but the resulting AST looked correct and other upstream parser implementations do not throw in this case.

Resolves #1134

## Test Plan

Run `./rome check` on a file with the following content and verify that no internal error is thrown:

```jsx
const Test = () => {
  return (
    <div>
      {true && (false ? <div /> : <div />)}
      <div />
    </div>
  );
};
```

### DIFF
--- a/internal/js-parser/tokenizer/context.ts
+++ b/internal/js-parser/tokenizer/context.ts
@@ -65,11 +65,7 @@ tt.parenR.updateContext = tt.braceR.updateContext = function(parser) {
 		}
 	}
 
-	if (out === undefined) {
-		throw new Error("No context found");
-	}
-
-	parser.state.exprAllowed = !out.isExpr;
+	parser.state.exprAllowed = !out || !out.isExpr;
 };
 
 tt.name.updateContext = function(parser, prevType) {


sebmck:
Hmm I'm not sure either. I checked `babel-parser` and they don't have any kind of check here. It's likely only necessary because of the way we made the parser recoverable. I'll come back to this and do some investigation.

PR #1157, opened 2020-08-28
fix: apply lint rules to proper DOM elements

### LINKED ISSUE #1154: 💭 How to handle libraries that implement onClick and onKeyPress for us
Specifically, I'm using the Material UI Button component, https://material-ui.com/api/button/.

This component handles key events for us, and uses the `onClick` handler as a fallback: https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/ButtonBase/ButtonBase.js (line 212).

I don't want to have to re-implement the logic in `handleKeyDown`, and I also don't want to add suppressions to every button implementation. 

~~One possible solution in the user space is to create my own `Button` implementation, add the suppression in one place, then try and convince the other dev's to use that implementation. That sounds like a people problem.~~ This does not fix the issue because Rome still looks at the `onClick` and tells us to pair it. A **very** hacky solution would be to rename that property on my implementation (`clickerOony`), and then intercept that into `onClick`.

pseudo code verison (very icky)
```
interface MyProps {
  clickerOony: Function
}

export function MyButton(props: ButtonProps | MyProps) {
    // maybe some handling if we pass key handler and onClick instead...
    // rome-ignore lint/jsx/noPropSpreading
    return <MUIButton onClick={onClick}{...props}></MUIButton>
}
```
If it's not clear, what I'm asking is how can I avoid the above?

ematipico:
## Summary

Fixes #1154 

This PR adds a new utility called `isDomElement` that basically checks if the `JSXElement` is actually a DOM element. 

This utility has been applied to most of the `jsx-a11y` rules and I updated some test cases (not all of them).

## Test Plan

Added few valid test cases and CI.


internal/js-ast-utils/isDomElement.ts
```diff
@@ -0,0 +1,5 @@
+import {AnyNode, JSXElement} from "@internal/ast";
+
+export function isDomElement(node: AnyNode): node is JSXElement {
```
sebmck, line 3: `isJSXDOMElement` might be a better name here, since ideally all exported names should be verbose enough that their intent is clear without knowing the package name (you can imagine having `@internal/html-ast-utils` and `isDomElement` could mean checking if something isn't a web component name etc).

PR #1160, opened 2020-08-29
feat: auto completion for zsh

### LINKED ISSUE #993: 💭 Add ZSH shell completions
With rome we can enable shell completions with a simple command. It would be great if the ZSH shell was supported as it is widely used (it is the macOS' default shell).

If this can help, this article seems really helpful: https://mads-hartmann.com/2017/08/06/writing-zsh-completion-scripts.html

Thanks :)

ematipico:
## Summary

Closes #993 

## Test Plan

I checked the the result of the output but I didn't check if it actually works. 
I would like some feedback first and then I will test it manually.

### DIFF
--- a/internal/cli-flags/Parser.ts
+++ b/internal/cli-flags/Parser.ts
@@ -421,7 +421,7 @@ export default class Parser<T> {
 			}
 
 			case "zsh": {
-				path = HOME_PATH.append(".zsh-completions",`_${programName}`);
+				path = HOME_PATH.append(".zsh-completions", `_${programName}`);
 				break;
 			}
 		}
--- a/internal/cli-flags/Parser.ts
+++ b/internal/cli-flags/Parser.ts
@@ -971,17 +971,17 @@ export default class Parser<T> {
 		caseString += "\n\tesac";
 
 		return dedent`
-function _${prg} {
-    local line
+			function _${prg} {
+			    local line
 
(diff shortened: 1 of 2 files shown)


noftaly:
I just tested on MacOS 10.15.6 with zsh and the latest version of the "oh-my-zsh" framework, and I have this error when doing `npx rome --write-shell-completions zsh`:
image

(it is possible that I did not installed your branch correctly, but I see your modifications in the module)

ematipico:
@noftaly yeah it's possible that you didn't download the changes of my branch. I don't get that error, my command is correctly recognized. (I get another error which is known because I don't have a folder)

internal/cli-flags/Parser.ts
```diff
@@ -911,6 +919,71 @@ export default class Parser<T> {
...
+    local line
+
+    _arguments -C \ ${globalFlags}
+        "1: :(${commandNames.join(" ")})" \
+        "*::arg:->args"
+
+    ${caseString}
+    ${functions}
+}
+`;
```
sebmck, line 984:
```suggestion
		return dedent`
			function _${prg} {
			    local line

			    _arguments -C \ ${globalFlags}
			        "1: :(${commandNames.join(" ")})" \
			        "*::arg:->args"

			    ${caseString}
			    ${functions}
			}
		`;
```

PR #1163, opened 2020-08-30
Extensible Rome

sebmck:
## Summary

This PR adds some new config options that deviate from our philosophy of "deal with our defaults, no options".

This aims to increase adoption, and retains our philosophy by having them remain as defaults.

The new options include:

### `format.enabled`

This is `true`. But when `false`, we do not complain about any pending fixes and formatting. `--apply` has no effect. Fix suggestions are still included in individual diagnostics. Diagnostics are not flagged as fixable.

### `lint.disabledRules`

This is an array where you can put rules you wish to disable.

### `format.indentStyle`

This defaults to `"tab"` but can be set to `"space"`. This allows you to use soft tabs.

### `format.indentSize`

Specifies the amount of times to repeat an indentation character for each level.

If `indentStyle` is `"tab"` then this defaults to `1`, otherwise if it's `"space"` then it's `2`.

### `extends`

Can now take an array of paths rather than being limited to single inheritance.

## Test Plan

Added tests.


ematipico:
From my point of view, I think it would be beneficial to have the configuration around formatting inside the `.editorconfig`. It's a standard after all and the file is created when we run `rome init`.

Using `.editorconfig` would allow to format files with different options (imagine `.md` files inside `.github` folder that require space instead of tab, while we can format the rest of the `.md` files using tabs).

jamiebuilds:
I’d prefer if Rome auto-populated config options from `.editorconfig` upon `rome init` then when running Rome it complained if the final configuration disagreed with `.editorconfig`

sebmck:
@jamiebuilds Once we support TOML we can do that.

sebmck:
Going to merge this and we can iterate on other customization options later. I want to release this, alongside TOML, JSON5 and YAML support in the next minor release.

Issue #1079, opened 2020-08-15
💭 Consider changing ts noArrayShorthand rule


sebmck:
Right now we require `Array<T>` instead of `T[]` but it usually ends up with extracting the array type into one that just has an `s` on the end.

ematipico:
Is this a lint rule or a diagnostic of the compiler?

And what's the suggested change?

sebmck:
It's the lint rule `js/noShorthandArrayType` (it should actually be in the `ts` namespace). We ban:

```
T[];
```

And instead recommend:

```
Array<T>;
```

The proposed change is flipping it and recommending `T[]` instead of `Array<T>`. The reason I added this rule in the first place was that `Array<T>` is the recommended pattern in Flow, which Rome was previously using.

ljharb:
This is a good change. It looks better, is just as obvious, and is shorter.

ematipico:
I'll work on this one

PR #1167, opened 2020-08-31
feat(lint): applied new rule for array signature

ematipico:
## Summary

Closes: #1079 

__BREAKING CHANGE__: changed rule criteria from `Array<T>` to `T[]`

_The rule has been renamed from `js/noShorthandArrayType` to `ts/preferShorthandArrayType`._

Moved the rule from the `js` category to the `ts` category.

The whole code base has been moved to use the new rule. There were some strange cases that I had to fix manually:

```ts
function foo(bar: Array<() => Promise<void>>);
```

The compiler doesn't handle properly a case like this and the end result is the following:

```ts
function foo(bar: () => Promise<void>[]);

```

I found three cases in the code base and I fixed them manually
```ts
function foo(bar: (() => Promise<void>)[]);

```

Usually if I see a case like this I would tend to create a type:
```ts
type Callback = () => Promise<void>;
function foo(bar: Callback[]);
```

@sebmck , what do you think? Should we handle a case like this?

## Test Plan

New tests, type check and CI.


Zenix27 (approved):

sebmck (approved):
Awesome!

> Should we handle a case like this?

I don't think we should handle extracting these into aliases automatically but we can definitely fix the problem where it's producing malformed code. We copy the `babel-generator` heuristics for determining when something needs to be wrapping in parenthesis, but for TypeScript, Babel just YOLOs it as TypeScript has a specific node for types being parenthesized.

I'm not sure if there's somewhere we can take some comprehensive TypeScript parentheses heuristics from.

Feel free to merge once the conflicts are resolved!

PR #1188, opened 2020-09-19
fix(website): Janky scrolling

asonkeri:
## Summary
Scrolling over the headings on the website feels very choppy and the browser history gets filled with entries for each and every heading element. [History.replaceState()](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState) fixes both issues.

If populating the browser history for each scrolled over heading is the intended behavior, this can be changed to use [History.pushState()](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) instead.


ematipico:
I noticed that! This change if actually really good!!
Thank you!

Issue #1023, opened 2020-08-10
☂️ Missing Tests


jer3m01:
List of all missing tests that need to be done.

- [x] ast-utils/removeLoc.ts 
- [x] codec-js-regexp/index.ts 
- [x] codec-source-map/ArraySet.ts @nt591
- [x] codec-source-map/base64.ts
- [x] codec-source-map/MappingList.ts 
- [x] codec-source-map/SourceMapConsumer.ts
- [x] codec-source-map/SourceMapConsumerCollection.ts
- [x] codec-source-map/SourceMapGenerator.ts
- [x] js-ast-utils/assertMultipleNodes.ts @aravind1078 
- [x] js-ast-utils/assertSingleNode.ts @aravind1078 
- [x] js-ast-utils/assertSingleOrMultipleNodes.ts @aravind1078 
- [x] js-ast-utils/cleanJSXText.ts
- [x] js-ast-utils/createMemberProperty.ts
- [x] js-ast-utils/createPropertyKey.ts 
- [x] js-ast-utils/getBindingIdentifiers.ts
- [x] js-ast-utils/getImportSpecifiers.ts
- [x] js-ast-utils/getJSXAttribute.ts
- [x] js-ast-utils/getJSXElementName.ts @aravind1078
- [x] js-ast-utils/getTSQualifiedBaseFromEntityName.ts
- [x] js-ast-utils/hasJSXAttribute.ts
- [x] js-ast-utils/hasPotentialSideEffects.ts
- [x] js-ast-utils/inheritLoc.ts
- [x] js-ast-utils/isBinary.ts @nt591
- [x] js-ast-utils/isEmptyTemplateLiteral.ts 
- [x] js-ast-utils/isFunctionNode.ts @aravind1078
- [x] js-ast-utils/isInTypeAnnotation.ts 
- [x] js-ast-utils/isJSXElement.ts @aravind1078 
- [x] js-ast-utils/isNodeLike.ts @sasurau4
- [x] js-ast-utils/isValidIdentifierName.ts
- [x] js-ast-utils/removeShallowLoc.ts 
- [x] js-ast-utils/renameBindings.ts
- [x] js-ast-utils/resolveIndirection.ts
- [x] js-ast-utils/template.ts 
- [x] js-ast-utils/tryStaticEvaluation.ts @nt591
- [x] js-ast-utils/tryStaticEvaluationPath.ts @nt591
- [x] js-ast-utils/valueToNode.ts @gie3d
[...]

PR #1194, opened 2020-10-01
test: js-ast-utils/inheritLoc

jer3m01:
## Summary
Part of #1023 

Adds test for js-ast-utils/inheritLoc.ts

## Test Plan

`rome check` is successful.
`rome test` passes all tests.


sebmck:
Thank you!

PR #1195, opened 2020-10-01
test: js-ast-utils/isEmptyTemplateLiteral

jer3m01:
## Summary
Part of #1023 

Adds test for js-ast-utils/isEmptyTemplateLiteral.ts

## Test Plan

`rome check` is successful.
`rome test` passes all tests.


sebmck:
Thanks! I feel like we should actually reconsider the `expressions` and `quasi` structure of template literal nodes. It's easy to cause unpredictable behavior with them and there's likely some safer shape we can use.

PR #1198, opened 2020-10-02
fix(a11y): general a11y fixes

estevanmaito:
This PR is related to 📎 Website accessibility audit, https://github.com/romefrontend/rome/issues/839#issuecomment-672338554

Pasting fixed points here for context:

- Theme toggler should contain `role="button"` and a descriptive label
- `.token.variable` (code syntax highlighting) have a contrast ratio of 4.38 and fails WCAG AA recommended 4.5
- Info messages in code (the ones with an ℹ) have a contrast ratio of 4.33
- Code examples with "Click to Expand" areas are not keyboard accessible
- Sidebar content, top "Latest blog post" and main "Rome" logo + slogan are outside landmarks. Could be solved with some `aside`/`header`/`section` for example.
- Discord links have different `href` (triggers "Links with the same name have a similar purpose"). Footer link points to `https://discord.gg/9WxHa5d` (the same place as the top one, but that points to the pretty URL)

Let me know what you think.

### DIFF
--- a/website/src/_includes/layouts/base.liquid
+++ b/website/src/_includes/layouts/base.liquid
@@ -68,7 +68,7 @@
 			<div class="sidebar-container">
 				<div class="sidebar-stripe"></div>
 
-				<div class="sidebar">
+				<aside class="sidebar">
 					<div class="header-desktop">
 						<a class="logo" href="/">
 							{% include svg/rome-logo.svg %}
@@ -131,7 +131,7 @@
 							{% include svg/sun.svg %}
 						</button>
 					</div>
-				</div>
+				</aside>
 			</div>
 
 			<div class="main">
--- a/website/src/_includes/layouts/base.liquid
+++ b/website/src/_includes/layouts/base.liquid
@@ -150,7 +150,7 @@
 							<a href="https://github.com/romefrontend/rome/blob/main/CODE_OF_CONDUCT.md">Code of Conduct</a>
 						</li>
 						<li>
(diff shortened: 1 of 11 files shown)


estevanmaito:
Regarding the error `To fix this run ./rome run scripts/generate-all-files and commit the results`, it's impossible to run any script on Windows, as it results in multiple spaces between lines. I've already reported it here https://github.com/romefrontend/rome/issues/1065

I'll try to run it inside a Linux VM this weekend.

ematipico:
These are refreshing changes! Thank you! 

Let me know if you're not able to run the script, I can do it for you tomorrow

website/src/_includes/layouts/base.liquid
```diff
@@ -125,13 +125,13 @@
 					</div>
 
 					<div class="color-scheme-switch-container">
-						<div class="color-scheme-switch">
+						<button class="color-scheme-switch" title="Toggle theme" aria-label="Toggle theme">
```
Muhnad, line 128: I think the state of switching between 2 states should build by using input<type=radio> and label for better A11Y [prototype example](https://codepen.io/Muhnad/pen/ExyRpjq?editors=1100) in codepen.

instead of using CSS `:checked` state we can use a little bit of JS to do the same job.

sebmck:
I believe @ematipico included these changes in #1305. Thanks @estevanmaito! Much appreciated.

Issue #985, opened 2020-08-08
☂️ Markdown support


sebmck:
cc @ematipico and #883

[Edit from @ematipico ]

This is the spec that we should follow: https://github.github.com/gfm/

## Current state

The markdown support is slowly resuming and here the things to implement

- #883 will implement emphasis and strong tokenization with known limits: limit recursion of inline parsing and handle case of closing tags;
- #883 will implement parsing of code blocks;
- [missing] parsing of code spans (inline back ticks)
- [missing] parsing of links 
- [missing] parsing of tables
- [missing] parsing of HTML inside markdown
- [missing] nested list items/numeric list
- [missing] quotes
- [missing] detect square brackets inside list
- [missing] parsing strike-through 

## Linting

What sort of things would we even want to lint in Markdown that wouldn't be solved by autoformatting?

noftaly:
There is a set of Markdown Linting rules available here if you need inspiration: https://github.com/DavidAnson/markdownlint#rules--aliases.

Some relevant examples: 
- header-increment - Heading levels should only increment by one level at a time
- no-inline-html - Prohibits inline HTML
- no-empty-links - No empty links

ematipico:
> no-inline-html

This a good rule but I think we should change it. We want to support GitHub flavored markdown. They support HTML but only for certain tags: https://github.github.com/gfm/#disallowed-raw-html-extension-

ashi009:
https://github.com/DavidAnson/markdownlint is great, but the defaults for those rules are less than ideal in terms of minimizing patch size and remove noise from code reviews.

For instance, an ordered list:

```markdown
1. a
   - sub
2. b
3. c
```

Will cause multiple line changes when inserts/removes items. However, the following one won't:

```markdown
1.  a
    - sub
1.  b
1.  c
```

Also, notice the additional space between the item number and the content, which is to allow indentation alignment when there are nested lists or content.

Do we have a high-level philosophy for making those lint rule decisions? There were lots of debates over such decisions for [prettier](https://github.com/prettier/prettier/issues/8507), and they didn't end well.

ematipico:
A couple of things that the formatter could do: spaces between blocks and automatically order a numeric list:

Before
```
# Title
## Second title
1. First
20. Second
```

After 
```
# Title

## Second title

1. First
2. Second
```

ashi009:
[...]

PR #1212, opened 2020-10-25
feat(format):parsing and formatting markdown table 

Monchi:
## Summary

This PR implements parsing and formatting markdown table (part of #985). The AST structure is inspired by mdast.

### Side changes

- Rename `Strong` and `Emphasis` from "inline" to "textwrapping". "inline" should include others such as reference syntax. I'm not native english speaker and "textwrapping" maybe not a good choice. Please tell me if you have any alternative!

- Add `forceFormat` option to ignore file handler's capability on test.

## Test Plan

Input:
```markdown
|head 1|head 2|
| with space |without space|
|bottom 1|bottom 2|
```

Output:
```markdown
|   head 1   |   head 2    |
| with space |without space|
|  bottom 1  |  bottom 2   |
```


ematipico:
Hi @Monchi ,

Thank you very much for this fantastic contribution!

About the naming convention, we tries to follow the GitHub spec and here they call it "inline" 

https://github.github.com/gfm/#inlines

Monchi:
@ematipico 

Thank you very much for developing Rome continuously!

I totally agree on following the GitHub spec.

Currently `parseInline` parses only strong and emphasis. But according to the GitHub spec, "inline" also includes link reference and more.

So I renamed current `parseInline` to `parseTextWrapping` and implemented new `parseInline` to parse true "inline".

internal/markdown-parser/test-fixtures/tables/input.md
```diff
@@ -0,0 +1,3 @@
+|head 1|head 2|
+| with space |ddadada|
+|**bold**|italic|
```
ematipico, line 3: The actual italic is missing :)
```suggestion
|**bold**|_italic_|
```

internal/markdown-parser/parser/paragraph.ts
```diff
@@ -18,64 +17,23 @@ export function parseParagraph(
...
-						if (unknownToken.type === "OpenSquareBracket") {
-							return parseReference(parser);
-						}
-						return parseText(parser);
-					},
-				);
-				if (nodes) {
-					children.push(nodes);
-				}
+		if (isTable && token.type === "TablePipe") {
```
ematipico, line 20: In theory we should not encounter `TablePipe` inside a paragraph. We should encounter just simple text (the pipe inside a paragraph is treated as text)

internal/markdown-parser/parser/table.ts
```diff
@@ -0,0 +1,67 @@
+import {parseInline} from "@internal/markdown-parser/parser/inline";
+import {
+	AnyMarkdownInlineNode,
+	MarkdownTableCell,
+	MarkdownTableRow,
+} from "@internal/ast";
+
+import {MarkdownTable} from "./../../ast/markdown/tables/MarkdownTable";
```
ematipico, line 8: Wrong import, it should be imported from `@internal/ast`

internal/formatter/test-fixtures/markdown/tables/input.test.md
```diff
@@ -0,0 +1,30 @@
...
+|head 1|head 2|
+| with space |without space|
+|bottom 1|bottom 2|
+
+```
+
+### `Output`
+
+```md
+|   head 1   |   head 2    |
```
ematipico, line 26: I like the formatting but I have mixed feelings about the justifying the text. I am totally in favour of filling the table with spaces, so it looks neat. What do you think?

internal/formatter/builders/markdown/tables/MarkdownTableRow.ts
```diff
@@ -0,0 +1,9 @@
+import {MarkdownTableRow} from "@internal/ast";
+import {Builder, Token} from "@internal/formatter";
+
+export default function MarkdownTableRow(
+	builder: Builder,
+	node: MarkdownTableRow,
+): Token {
+	throw Error("Must be rendered through MarkdownTable.");
```
ematipico, line 8: Is there a reason why you decided to use `MarkdownTable` to format this node?

ematipico (changes requested):
Thank you for your contribution, lot of nice work done here! I left some feedback and question.

There's a couple of things I don't see:
- table headings https://github.github.com/gfm/#example-198
- handling of the text alignment https://github.github.com/gfm/#example-199

Do you plan to add them later with another PR?

Also, I think we should have more cases inside our tests. The only case I can see here is a "happy" case.

What about cases like this?

```
| Something
```

```
| Something | Else |
| Wow |
| Wow 
```

Also, the requires an header (`|----|`). Without the header, the table is rendered as text. I think we should consider this case.

Monchi:
@ematipico Thank you for reviewing!
I had some misunderstandings about GFM spec. I'll add some "unhappy" test cases and update parser.

BTW, I think we need too long time to stabilize our full-scratch parser.
How do you think about including [micromark](https://github.com/micromark/micromark/) as a internal? We can generate [mdast](https://github.com/syntax-tree/mdast) with micromark, then convert it to rome's AST.

ematipico:
I wasn't aware of `mdast` and we can definitely take inspiration from their categorisation and AST. Without realizing it, I think we already have some of their AST inside ours. 

About `micromark`, with Rome we want to try to do things ourselves without relying on third party packages. We can surely copy and credit third party code (we've done this already). About our markdown parser, we already implemented our logic, I'm not sure how it would play inside Rome code base. I didn't do any research. What do you think?

Monchi:
@ematipico
I think it's good to copy `micromark` **for now**.
The important point is that we won't use `mdast` directly but use Rome AST that converted from `mdast`. So we can rewrite our parser later without affecting linter, formatter, etc. We already do this in JavaScript parser with babel-parser.

IMO, we should focus on making Rome ready for use to gather more contributors.

ematipico:
Makes sense to me! Feel free to look into it if you have the time

sebmck:
Just wanted to leave a quick note around copying from other projects. We want to be very careful with licensing, and API copyright is a real area of ambiguity. Preferably we should take inspiration but never straight up copy. There's almost always room for improvement.

jamiebuilds:

Thanks again!

PR #1215, opened 2020-10-27
fix: documentation #1214

### LINKED ISSUE #1214: 🐛 Little problem with the "preferOptionalChaining" documentation
I just wanted to report that the doc page of the "preferOptionalChaining" rule, on the website, is the raw markdown: https://rome.tools/docs/lint/rules/js/preferoptionalchaining/

aravind1078:
## Summary

this PR fixes the website documentation for preferOptionalChaining lint rule

closes #1214

### DIFF
--- a/website/src/docs/lint/rules/js/preferOptionalChaining.md
+++ b/website/src/docs/lint/rules/js/preferOptionalChaining.md
@@ -1,18 +1,17 @@
+---
+title: Lint Rule js/preferOptionalChaining
+layout: layouts/rule.liquid
+showHero: false
+description: MISSING DOCUMENTATION
+eleventyNavigation:
+   key: lint-rules/js/preferOptionalChaining
+   parent: lint-rules
+   title: js/preferOptionalChaining
+---
 
-			---
-			title: Lint Rule js/preferOptionalChaining
-			layout: layouts/rule.liquid
-			showHero: false
-			description: MISSING DOCUMENTATION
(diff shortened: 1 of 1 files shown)


ematipico (approved):
Thank you!

darthmaim:
I wonder if this could have been prevented with a linting rule detecting indented frontmatter. Differentiating wrongly indented frontmatter from a codeblock (created by indentation instead of ` ``` `) starting with `---` is difficult though. Is there a linting rule prefering ` ``` ` over indentation for codeblocks?

ematipico:
We don't have a markdow linter yet but it's surely something that we can thing about.

Although, about the bad formatting (tab) of what happened here inside the `---` block, it's not something that we can lint, I think. It's not part of the spec and it's something particularly tailored for 11ty (to be more specific, the package `gray-matter`)

darthmaim:
Even though Frontmatter is not in the spec, it's not something specific to 11ty. Most static site generators use it and even github has support for rendering it.

ematipico:
Github supports it? Really? 😮  I think I didn't understand the frontmatter part.

Anyway, it's surely something that we can have once we have a working parser!

darthmaim:
Yes, this is how github renders it (see for example https://github.com/rome/tools/blob/main/website/src/docs/lint/rules/js/preferOptionalChaining.md):

image

I have added a comment to #985 about frontmatter.

PR #1219, opened 2020-11-11
Improve configuration format support

sebmck:
This is a rebase and update of a PR I started a while ago. Trying to get back into the groove of things so expect this to be the first some very rapid work.

This adds a new package called `codec-config` that contains both the JSON and TOML parser. It contains a simple API for working with possible configuration files, without needing to be concerned about the actual original language format.

I intend on removing RJSON, which would be replaced with JSON5. YAML is a superset of JSON so it makes sense for us to build it into the JSON parser with some branches.

The `codec-config` tests are currently disabled as they do not pass as they are testing TOML and YAML which are incomplete. If you want to enable them then remove the extension from the bottom of the break check in `internal/core/common/file-handlers/index.ts`, and remove the `true` argument to `createFixtureTests` in `internal/codec-config/index.test.ts`.

I also refactored the TOML parser. It doesn't make sense for us to have a full AST for configuration formats as any manipulation should happen on the derived JS values, and any additional information should be tagged in an auxiliary data structure like we do for JSON comments to retain them when formatting.


ematipico:
@sebmck I noticed that now the `package.json` file have tabs instead of spaces. Wouldn't this conflict with how `npm` formats the `package.json`? As far as I understood it uses spaces

PR #1225, opened 2020-12-09
fix: Typo on "uncommitted", updates file to return early

CodyCahoon:
## Summary

- Fixes a typo with "uncommitted"
- Returns early when git status returns an empty string

## Test Plan

`./rome test`
`./rome check --apply`

Additional testing included modifying an existing lint rule, running this script and verifying the exit code of 1, and running this script without any modifications and verifying an exit code of 0.


ematipico (approved):
Thank you!

PR #1228, opened 2020-12-10
add test for js-ast-utils/valueToNode.ts

gie3d:
## Summary

part of #1023
added tests for js-ast-utils/valueToNode.ts

## Test Plan

./rome test --all tests passed

### DIFF
--- a/internal/js-ast-utils/valueToNode.test.ts
+++ b/internal/js-ast-utils/valueToNode.test.ts
@@ -1,0 +1,84 @@
+import {test} from "rome";
+import {valueToNode} from "./valueToNode";
+import {
+	jsArrayExpression,
+	jsBooleanLiteral,
+	jsNullLiteral,
+	jsNumericLiteral,
+	jsObjectExpression,
+	jsObjectProperty,
+	jsReferenceIdentifier,
+	jsStaticPropertyKey,
+	jsStringLiteral,
+} from "@internal/ast";
+import {createPropertyKey} from "./createPropertyKey";
+
+test(
+	"verify valueToNode returns the correct supported primitive type",
+	(t) => {
+		t.looksLike(valueToNode("string"), jsStringLiteral.quick("string"));
+		t.looksLike(valueToNode(true), jsBooleanLiteral.quick(true));
+		t.looksLike(valueToNode(123), jsNumericLiteral.quick(123));
+		t.looksLike(
+			valueToNode(undefined),
+			jsReferenceIdentifier.quick("undefined"),
+		);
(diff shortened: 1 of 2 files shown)


internal/js-ast-utils/valueToNode.test.ts
```diff
@@ -0,0 +1,84 @@
...
+		);
+		t.throws(
+			() => {
+				valueToNode(() => {});
+			},
+			"Do not know how to turn this value into a literal",
+		);
+		t.throws(
+			() => {
+				valueToNode(BigInt(10));
```
sebmck, line 67: Would you like to implement this instead of turning it into a test error? We have `BigIntLiteral` that you can use. It has a `string` `value` prop that you can set with just `String(num)`.

gie3d: Thanks for the suggestion. I'll do it.

gie3d: Please correct me if I'm wrong. I initially thought that BigIntLiteral can be used like `jsBigIntLiteral.create(String(value))`. However, `string` is not acceptable then I need to parse it to JSStringLiteral first. So I come up with `jsBigIntLiteral.create(jsStringLiteral.quick(String(value)))`

internal/js-ast-utils/valueToNode.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+test(
+	"verify valueToNode returns the correct supported primitive type",
+	(t) => {
+		t.looksLike(valueToNode("string"), jsStringLiteral.quick("string"));
+		t.looksLike(valueToNode(true), jsBooleanLiteral.quick(true));
+		t.looksLike(valueToNode(123), jsNumericLiteral.quick(123));
+		t.looksLike(
+			valueToNode(BigInt(10)),
+			jsBigIntLiteral.create(jsStringLiteral.quick("10")),
```
sebmck, line 25: This is actually just using a quirk. `create` accepts an object with properties for the node. `JSBigIntLiteral` only has a single `value` property that's a `string`. Technically a `JSStringLiteral` would satisfy this type.

What you're looking for is the `quick` method that we have on nodes that only have a single property. This allows us to skip having an object entirely. (Side note: I don't like the quick name but couldn't think of anything better, open to suggestions there!)

```suggestion
			jsBigIntLiteral.quick("10"),
```

sebmck (approved):

sebmck:
Thank you @gie3d! Appreciate the contribution. I left a review with a final nit and merged my code suggestion.

PR #1229, opened 2020-12-11
test: add tests for isConditional

CodyCahoon:
## Summary

- Implement tests for `isConditional`

## Test Plan

```
./rome test
✔ 3_840 tests passed!
```

### DIFF
--- a/internal/js-ast-utils/isConditional.test.ts
+++ b/internal/js-ast-utils/isConditional.test.ts
@@ -1,77 +1,35 @@
 import {test} from "rome";
-import {parseJS} from "@internal/js-parser";
-import {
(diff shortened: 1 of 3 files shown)


sebmck:
Nice! Could you update this to use `template.expression` rather than having a `parseJS` helper? You can see an example of where we do that [here](https://github.com/rome/tools/blob/main/internal/js-ast-utils/isFunctionNode.test.ts#L9-L15).

CodyCahoon:
@sebmck Yes, updated to use template instead, thanks for pointing that out. Also fixed misspelled "substitutions" in the template file as well

ematipico (approved):

PR #1230, opened 2020-12-11
feat: no single char alts lint rule

davidnormo:
## Summary

This changes is one of those suggested in #976 

This is a new lint rule to prevent single character alternations in regex e.g. `/a|b/` and would pass character classes instead e.g. `/[ab]/`

Looks like the rule found this in the rome codebase 🤓

```
 internal/commit-parser/index.ts:58:5 
lint/js/noSingleCharRegexAlternatives ━━━━━━━━━━━━━━━━━━
  ✖ No single character alternations in regular
    expressions. Use a character class instead.

    56 │     if (
    57 │       parser.matchToken("Word") &&
  > 58 │       /("|')/.test((parser.getToken() as 
     → │ Tokens["Word"]).value)
       │ ^
    59 │     ) {
    60 │       parser.eatToken("Word");

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

✖ Found 1 problem
```

## Test Plan

TBC


sebmck:
Nice! I love it when new lint rules uncover places where we're violating them as it instantly shows their value. The rule looks good to me, what do you think about adding an autofix too that turns it into a charset?

davidnormo:
I've added the autofix advice. It took me a little while as I ran into some trouble with `addFixableDiagnostic` - it seems to accept an invalid fixed signal - in my case I was trying to replace a regex alternation with a char set but instead it should have been a regex sub expression. Is this a known issue? Perhaps something I can fix in a separate PR...

ematipico (approved):
Thank you!!

Issue #1199, opened 2020-10-04
🐛 Invalid bundling for code with optional chaining


antonk52:
## Environment

```
$ rome rage --summary
Environment Variables
  Object {
    COLORFGBG: "7;0"
    COLORTERM: "truecolor"
    LANG: "en_US.UTF-8"
    TERM: "xterm-256color"
    TERM_PROGRAM: "iTerm.app"
    TMUX: "/private/tmp/tmux-1434521695/default,73535,0"
  }

User Config
  unset

Rome Version
  10.0.4-beta-dev

Node Version
  14.9.0

Platform
  darwin x64 18.7.0

Terminal Features
  Object {
    background: "dark"
    colorDepth: 8
    columns: 173
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
  }

Client Flags
  Object {
    clientName: "cli"
    silent: false
    cwd: ___R$project$rome$$internal$path$index_ts$AbsoluteFilePath: /Users/antonk52/Documents/dev/personal/rome
    realCwd: ___R$project$rome$$internal$path$index_ts$AbsoluteFilePath: /Users/antonk52/Documents/dev/personal/rome
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 63_004_672
      pid: 85_382
      uptime: 0.47658209
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 63_004_672
        ownedBytes: 0
        ownedFileCount: 0
        pid: 85_382
        uptime: 0.476475193
      }
    ]
  }
```

## Steps To Reproduce

1. Create a `foo.ts` file with the contents
```ts
export function foo(
	arg:
		| undefined
		| {
				children: Set<string>;
			},
) {
	arg?.children.add('foo')
}
```

2. run `./rome bundle ./foo.ts ./dist`
3. Open `./dist/foo.js` to see the following

```js
(arg == null ? undefined : arg.children).add == null
	? undefined
	: (arg == null ? undefined : arg.children).add("foo");
```

This produces invalid code, which breaks at the runtime. Ie `TypeError: cannot get add of undefined`

## Expected Results

Bundle command should not introduce new ways for code to error out. To be more specific I expected to see something similar to

```js
arg == null ? undefined : arg.children.add("foo")
```

antonk52:
[...]

PR #1231, opened 2020-12-13
Optional call expressions fix

antonk52:
## Summary

This PR is intended to fix how Rome parses and formats optional call expressions. 

To be more specific it fixes issue #1199 and the issue with optional call expressions. Right now rome only marks any code as JSOptinalCallExpression that has `?.` anywhere before the calling parens, ie both `a?.b()` and `a.b?.()`. However Rome's AST does not carry any information wether there is a `?.` directly before the parens. 

I've added `optional` property to `JSOptionalCallExpression` to fix it. My motivation was
- This helps to preserve the `?.` when Rome does formatting
- Other compilers have it([babel](https://astexplorer.net/#/gist/d9636001e4480420e7d6e4eaff30f402/e2bf4f7301a716c6a1c924c053007c6468588605), [acorn](https://astexplorer.net/#/gist/d9636001e4480420e7d6e4eaff30f402/31667555bbe13cdd1d5567d0a2799027d2265480), [flow](https://astexplorer.net/#/gist/d9636001e4480420e7d6e4eaff30f402/8ec6aa01884d86b2b921d5ee4720520690c3f31b), [typescript](https://astexplorer.net/#/gist/d9636001e4480420e7d6e4eaff30f402/565784a9ccc78b17562380efda3a16f394a1d40f) with specified location of the question dot token)

## Test Plan

I've added the test case with the snapshots for the optional chaining transformer. Though I have not found other test cases for transformers and think mine can either be moved closer to the project root, since it includes other transformers. I think there are could be better ways of testing it that I've yet to discover.


internal/compiler/transforms/compile/transpile/optionalChaining.ts
```diff
@@ -7,7 +7,93 @@
...
+			});
+			current = current.object;
+		} else {
+			root = current;
+			break;
+		}
+	}
+
+	if (root === null) {
+		throw new Error(
```
ematipico, line 38: I would add a test case for this path

ematipico (approved):
Nothing major, I would just a one test case

internal/compiler/transforms/compile/transpile/optionalChaining.ts
antonk52: I've removed the null check completely since it the `root` value can never be null here because we would never exit the loop above. I've re-declared the `root` variable with definite assignment assertions and the typescript is happy now.

antonk52: Looks like rome's formatter avoided the `!` sign when printing though the information about it was in the AST, I've fixed it the last commit too 😅

sebmck:
Woo! Thank you so much! Looks great. I've avoided using optional member expressions inside of Rome itself due to this bug...

Issue #982, opened 2020-08-08
📎 Unquote valid JS identifier property keys and computed members


sebmck:
Right now if you have:

```
const foo = {
  "bar": true,
};
foo["bar"];
```

We will retain the quotes even though we can drop them. Not sure if this should be implemented in the linter as an autofix or formatter as a basic AST transform.

katungi:
@sebmck I think it should be a lint rule. Would the two options still give the same results which is no quotes??

sebmck:
@katungi Yeah it would. I don't think this makes sense as a lint rule, as it seems purely like formatting, and when we're generating any sort of code, we should still do this.

The changes would need to be made in only the file `internal/formatter/builders/js/objects/JSStaticPropertyKey.ts`. `node.value` can be a `StringLiteral` so you would need to check for that. You would also need to validate that the `name` is a valid JavaScript identifier, which can be done using `isValidIdentifierName` that you can import from `internal/js-ast-utils`.

pradel:
As https://github.com/rome/tools/pull/1235 was merged this one can probably be closed.

PR #1235, opened 2020-12-15
feat(formatter): unquote valid members

pradel:
## Summary

Hey all (first-time contributor to rome), to get a bit more familiar with the codebase I decided to try to fix issue https://github.com/rome/tools/issues/982.

A lot of changes in this pr are related to this new formatting applied to the codebase. The important files are:

For for property keys:
- `internal/formatter/builders/js/objects/JSStaticPropertyKey.ts` changes for property keys.
- `internal/formatter/test-fixtures/objects/property-key/input.test.md` test output.

For computed members:
- `internal/formatter/builders/js/expressions/JSMemberExpression.ts` changes for computed members.
- `internal/formatter/test-fixtures/objects/computed-member/input.js` test output.

From the snapshots, this is how the change looks like:

### Property keys `Input`

```js
const foo = {
  "bar": true,
  "foo-bar": true,
};
```

### Property keys `Output`

```js
const foo = {
	bar: true,
	"foo-bar": true,
};
```

### Computed members `Input`

```js
const foo = {};

foo["bar"] = true;
foo["foo-bar"] = true;
```

### Computed members `Output`

```js
const foo = {};

foo.bar = true;
foo["foo-bar"] = true;
```

## Test Plan

Tests are passing properly ✅ 

```
➜ ./rome check --apply
 Building trunk 
! This command has been hidden. Consider its usage to be experimental and do not expect support or
ℹ 1_734 files linted
✔ No known problems!
```

```
➜ ./rome test 
 Building trunk 
! This command has been hidden. Consider its usage to be experimental and do not expect support or
✔ 3_844 tests passed!
```


ematipico:
Hi @pradel , thank you for looking into this!

Regarding your issue, here's the steps you can follow to understand where to look:
- create a dummy file e.g. `test.js`
- inside the file you can write:

```js
a["b"] = "bew";
a.b = "bew"	
```

which are the two cases that you want to analyze (format from the first case to the second case)

- run the command `./rome parse test.ts`

The output should contain something like this:

```js
JSExpressionStatement {
  expression: JSAssignmentExpression {
    operator: "="
    right: JSStringLiteral {value: "bew"}
    left: JSMemberExpression {
      object: JSReferenceIdentifier {name: "a"}
      property: JSComputedMemberProperty {value: JSStringLiteral {value: "b"}}
    }
  }
}
JSExpressionStatement {
  expression: JSAssignmentExpression {
    operator: "="
    right: JSStringLiteral {value: "bew"}
    left: JSMemberExpression {
      object: JSReferenceIdentifier {name: "a"}
      property: JSStaticMemberProperty {value: JSIdentifier {name: "b"}}
    }
  }
}

```

From here, you can understand that we want to apply a different formatting inside the `JSMemberExpression` node. So you can open the file `JSMemberExpression.ts` insider the formatter package.

As we want create new nodes (and replace with the existing ones), you can use the function `jsStaticMemberProperty` that you can export from the file `ast/js/expressions/JSStaticMemberProperty.ts`.

I hope it helps.

pradel:
@ematipico thanks for the help I managed to make it work 🎉 
I updated the description of the pull request to point to the important parts of the changes. Let me know if something is missing or wrong :)

internal/formatter/test-fixtures/objects/computed-member/input.js
```diff
@@ -0,0 +1,4 @@
+const foo = {};
```
ematipico, line 1: I might make sense to add some mixed/more cases, e.g.:

```js
foo.bar["bar"]["lorem_ispsum"]["lorem-ipsum"]
```

And so on. Let's try to cover some edge case.

ematipico (changes requested):
It looks good to me! I would add some more test case inside the tests

internal/formatter/test-fixtures/objects/computed-member/input.js
pradel: Added some edge cases 👍

ematipico (approved):

PR #1236, opened 2020-12-15
chore: ignore scss, sass and less

jer3m01:
Ignore scss, sass and less as they make rome crash.

Better strategy should be used instead later.


ematipico (approved):

sebmck:
Thanks!

PR #1240, opened 2020-12-17
feat(formatter): better format for fragments

ematipico:
## Summary

I didn't like that much how rome formats the fragments (and the rest of the JSXElements).

I changed how fragments are formatted

Before:

```js
	return <>
	</>;
```

After:
```js
	return <></>;

	return <>
		Lorem ipsum
	</>;

	return <>
		<Foo />
		<Foo />
	</>;

	return <>
		<>
			<></>
		</>
	</>;
```

If approved, I can make a second PR to change also `JXSElement` formatting.

## Test Plan

Created a new test


sebmck:
Nice! Although I don't think having the `(` and `)` as explicit formatter tokens is the correct change. I'm pretty sure:

```
<>
	<></>
</>
```

Would end up being formatted as:

```
(
  <>
    (
      <></>
    )
  </>
)
```

I think you should actually just be able to add a leading `softline` (or is it `hardline`?). We already handle trailing newlines at the beginning of expressions where newlines are significant, so it will automatically wrap it in parens if necessary. ie:

```
return
foo();
```

is two statements.

ematipico:
@sebmck  I reviewed the heuristic of the formatting and looks like a good compromise! We don't have parenthesis anymore but I'm happy with the change!

PR #1242, opened 2020-12-21
fix(format): handle typeArguments in JSTaggedTemplateExpression

### LINKED ISSUE #1226: Formatter is removing type declarations before tagged templates
Formatter is removing type declarations before tagged templates

```typescript
const users = await sql<User>`SELECT * FROM ... ` 
```

becomes
```typescript
const users = await sql`SELECT * FROM ...` 
```
When the linter/formatter runs

yeonjuan:
## Summary

I added handling `typeArguments` in `tagged template`. It will fix #1226.

## Test Plan

I added a [test case](https://github.com/rome/tools/pull/1242/files#diff-08fed1028805f9130290d65f9d5bc3467ba29c25b4275342b9b1c4e4c62f72d9) ensuring it.


ematipico (approved):
LGTM! Thank you!

PR #1248, opened 2020-12-22
fix(compiler): handle optionalChaining properly

### LINKED ISSUE #1199: 🐛 Invalid bundling for code with optional chaining

yeonjuan:
## Summary
As ensuring with [the test case](https://github.com/rome/tools/pull/1248/files#diff-c9ef94c68ffb9713ca465eb168dc9c59d3961c075b201b486d6ffa39f9f3403eR48) this pr will fix #1199.
Furthermore, I found unreported bugs and those will be fixed too. 

### code
```js
// ./bug.js
a10?.c.d();
```

#### current main branch

It throws "Not implemented non js memeber expression case" error, but as I know `a10?.c.d();` is valid syntax.

```bash
$ ./rome compile ./bug.js

✖ Not implemented non js memeber expression case
```

#### PR branch

compile success

```bash
$ ./rome compile ./bug.js

a10 == void 0 ? void 0 : a10.c.d();
```

## Test Plan

I added [Test cases](https://github.com/rome/tools/compare/main...yeonjuan:issue-1199?expand=1#diff-c9ef94c68ffb9713ca465eb168dc9c59d3961c075b201b486d6ffa39f9f3403e).

### DIFF
--- a/internal/compiler/transforms/compile/transpile/optionalChaining.test.md
+++ b/internal/compiler/transforms/compile/transpile/optionalChaining.test.md
@@ -5,16 +5,34 @@
 ## `optional chaining gets compiled correct`
 
 ```
-a1 == null ? undefined : a1.b();
+a1 == void 0 ? void 0 : a1.b();
 
-(a2 == null ? undefined : a2.b) == null ? undefined : a2.b.c();
+(a2 == void 0 ? void 0 : a2.b) == void 0 ? void 0 : a2.b.c();
 
-a3.b == null ? undefined : a3.b();
+a3.b == void 0 ? void 0 : a3.b();
 
-(a4 == null ? undefined : a4.b) == null ? undefined : a4.b();
+(a4 == void 0 ? void 0 : a4.b) == void 0 ? void 0 : a4.b();
 
-a5.b == null ? undefined : a5.b();
+a5.b == void 0 ? void 0 : a5.b();
 
-(a6 ? b6 : c6) == null ? undefined : (a6 ? b6 : c6).zoo();
+(a6 ? b6 : c6) == void 0 ? void 0 : (a6 ? b6 : c6).zoo();
+
+a7 == void 0 ? void 0 : a7.b.c();
+
+a8 == void 0 ? void 0 : a8.b.c;
+
+a9.b == void 0 ? void 0 : a9.b.c.d();
+
+a10.b == void 0 ? void 0 : a10.b.c.d;
+
+(a10.b == void 0 ? void 0 : a10.b.c) == void 0 ? void 0 : a10.b.c.d;
+
+(a11.b == void 0 ? void 0 : a11.b.c.d) == void 0 ? void 0 : a11.b.c.d.e();
+
+(a12.b == void 0 ? void 0 : a12.b.c.d) == void 0 ? void 0 : a12.b.c.d.e;
+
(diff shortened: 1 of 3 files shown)


ematipico (approved):
It looks good to me! Quick question: what's the reason of changing `undefined` to `void 0`?

yeonjuan:
> Quick question: what's the reason of changing undefined to void 0?

Hi :) @ematipico 
Thanks for your review 

It's because the `undefined` is not a reserved word in the spec so it can be shadowed by the declared identifier `undefined`. (except for global scope)

* target code

```js
function foo () {
  var undefined = 1;
  obj?.a;
}
```

so, if we compile the target code using `undefined`, it affected by other user codes. 

```js
function foo () {
  var undefined = 1;
  obj == undefined ? undefined : obj.a;
}
```


On the other hand, `void` is safe.

```js
function foo () {
  var undefined = 1;
  obj == void 0 ? void 0 : obj.a;
}
```


I'm not sure whether there is a restriction disallowing the use of undefined as an identifier in Rome. so I decided to use `void`. :)

ematipico:
Thank you very much for the answer! We don't actually have a rule about the use or not of `undefined` as variable. Maybe we should think about it!

PR #1249, opened 2020-12-23
test: add test for js-ast-utils/removeShallowLoc.ts

jer3m01:
## Summary
Part of #1023 

Adds test for js-ast-utils/removeShallowLoc.ts

## Test Plan

`rome check` is successful.
`rome test` passes all tests.


ematipico (approved):

ematipico (commented):
As the function says "shallow", we should test that the children still have their `loc` object?

Issue #1244, opened 2020-12-21
📎  Typescript v4.1 & template literal types support


antonk52:
The new typescript version has been out for some time now and it introduced [Template liternal types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#template-literal-types).

This issue is for updating the typescript in node_modules and adding support for the new feature.

ematipico:
I think the version 4.1 brings to the table more features (and we don't have yet all the features we should support form version 4.0):
- variadic tuples: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types
- labeled tuples: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements
- key remappings in mapped types: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#key-remapping-in-mapped-types

ematipico:
Closing, we now support all these features! 🎉

PR #1256, opened 2020-12-23
feat(parser): support parsing key remap in mapped types

yeonjuan:
## Summary

It's for supporting `Key Remapping in Mapped Types` which introduced in TS 4.1. (#1244) 

- [ts-4.1 key-remaping-in-mapped-types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#key-remapping-in-mapped-types)

```ts
type MappedTypeWithNewKeys<T> = {
    [K in keyof T as NewKeyType]: T[K]
    //            ^^^^^^^^^^^^^
    //            This is the new syntax!
}
```

## Test Plan

I added [a test case and snapshot](https://github.com/rome/tools/compare/main...yeonjuan:ts4-1-key-remap?expand=1#diff-29177e1a4d0172414689c180568fbd0e857a1c8d8c9448dd97a4ba9c5297972d)


ematipico:
Probably we should check also the formatter part: https://github.com/rome/tools/blob/main/internal/formatter/builders/js/typescript/TSMappedType.ts

Do you prefer doing it in this PR or create a separate PR?

yeonjuan:
@ematipico
> Probably we should check also the formatter part:

Oh, Thanks :) I want to do in this PR

yeonjuan:
@ematipico Thanks :) I changed the formatter part also. Maybe should we add a test case?

ematipico:
@yeonjuan thank you for looking into this! Yes I think we should add a small test case. It could be a starting point, in case in the future we will uncover some edge case.

yeonjuan:
@ematipico Thanks for your review and proposal. :) I added test cases for it. (The CI was failed before but works now. 😅)

ematipico (approved):
This is 🔥 ! We can now start using the mapped types inside Rome! Huge thanks!

Issue #983, opened 2020-08-08
☂️ HTML support


sebmck:
## Parsing

I implemented a pretty awful HTML parser that isn't spec compliant in the slightest. We can expand upon it though and import some spec tests to ensure validity.

## Linting

We might want to make some of the JSX and Accessibility lint rules generic and allow accepting HTML.

ematipico:
Something that I wish to see here is the orchestration of different parsers: when there's a `<style>`, use the CSS parsing, same thing with `<script>`.  

But maybe this is something that can be done way in the future!

ematipico:
This resource could help: https://html.spec.whatwg.org/multipage/parsing.html

cassiebeckley:
Is anybody working on this? I'd be interested if not.

ematipico:
Not at the moment @cassiebeckley , you can pick it up if you wish! 

What's missing (as far as I know):
- `CDATA` parsing
- parsing of self closing tags

And surely there are more cases that I didn't think about.

If possible, it would be great to change the current parser and make it spec compliant. With did that with our current CSS parser and should do the same with our current HTML parser.

PR #1257, opened 2020-12-23
feat(html-parser): parse DOCTYPE and make parser recoverable

### LINKED ISSUE #772: 📎 Make HTML parser recoverable
We still have usages of `expectToken` etc that will throw hard errors.

ematipico:
## Summary

Part of #983 

Closes #772 

It adds supports for the `!DOCTYPE` token and its value.

Anything that is not `!DOCTYPE html` will throw an error. Other values other than `html` will not be supported for now because they belong to old stuff that are not part of the modern web.

Reference: https://html.spec.whatwg.org/multipage/parsing.html#the-initial-insertion-mode.

When the value is not `html`, an error is shown:

Screenshot 2020-12-23 at 16 49 16

The html parser is now recoverable and the usage of `expectToken` is removed from the parser.

## Test Plan

Created a test case.

Made sure the existing smoke tests still pass.


internal/html-parser/index.ts
```diff
@@ -336,78 +395,115 @@ function parseTag(parser: HTMLParser): HTMLElement {
 	);
 }
 
+function parserDoctype(parser: HTMLParser): HTMLDoctypeTag | undefined {
```
yeonjuan, line 398:
```suggestion
function parseDoctype(parser: HTMLParser): HTMLDoctypeTag | undefined {
```
Maybe it should be `parseDoctype` ? :)

PR #1261, opened 2020-12-25
test: add test for parser-core/ParserCore.ts

jer3m01:
## Summary
Part of #1023 

Adds test for parser-core/ParserCore.ts

## Test Plan

`rome check` is successful.
`rome test` passes all tests.


internal/parser-core/ParserCore.test.ts
```diff
@@ -0,0 +1,38 @@
+import {test} from "rome";
```
ematipico, line 1: I am not sure this test makes sense. `ParserCore` is used to create new parsers, let's see it as an abstract class, and in this test we are testing the JS parser. 

Shouldn't we test its methods like `nextToken`, `getToken`, etc.?

ematipico (changes requested):
Not sure about the nature of these tests

internal/parser-core/ParserCore.test.ts
jer3m01: Totally, my oversight, will test ParserCore itself.

jer3m01:
Looks like I am using a too old fork of rome locally, will rebase tomorrow.

ematipico (approved):
Great, much better now. If you made a rebase, we could merge it! (CI is now fixed)

PR #1262, opened 2020-12-25
test: add test for parser-core/utils.ts

jer3m01:
## Summary
Part of #1023 

Adds test for parser-core/utils.ts

## Test Plan

`rome check` is successful.
`rome test` passes all tests.


ematipico (approved):
Thank you! Looks good!

PR #1263, opened 2020-12-25
feat(lint): add autofix in noExtraBooleanCast

yeonjuan:
## Summary

I made a pr improving `noExtraBooleanCase` rule.

First, I made it fix problems automatically

```js
while (!!foo) {}

// autofixed
while (foo) {}
```

Second, I made it report some problems even if it's not in a boolean context (`if`, `for`...).

- For example, in the current version, these are lint-free
```js
let a = 1;
const foo = {};
a += !Boolean(foo); // lint-free
``` 

- In this PR, these are reported and auto fixed
```js
let a = 1;
const foo = {};
a += !Boolean(foo); // Error and it's fixed to "a += !foo"
``` 

## Test Plan

test cases

### DIFF
--- a/internal/compiler/lint/rules/js/noExtraBooleanCast.test.md
+++ b/internal/compiler/lint/rules/js/noExtraBooleanCast.test.md
@@ -8,7 +8,7 @@
 
 ```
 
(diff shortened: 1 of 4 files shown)


ematipico (approved):
Thank you! LGTM

PR #1265, opened 2020-12-27
fix(lint): handle recursive reference in noUnusedVariables

yeonjuan:
## Summary

Current "noUnusedVariables" does not handle recursive reference.

* target.js
```js
function foo() {
	foo();
}

const bar = () => {
	bar(this);
};

class Baz {
	init() {
		this.baz = new Baz();
	}
}
```

```
$ ./rome check target.js
 1 file linted
 No known problems!
```

But I believe it should report lint error because `foo`, `bar`, `Baz` is not called in outside scope.
I made a PR handle it.

```
$ ./rome check target.js
  ✖ The function variable foo is unused.

  > 1 │ function foo() {
      │          ^^^
    2 │   foo();
    3 │ }
...
  ✖ The const variable bar is unused.

    3 │ }
    4 │ 
  > 5 │ const bar = () => {
      │       ^^^
    6 │   bar(this);
    7 │ };
...
  ✖ The class variable Baz is unused.

     7 │ };
     8 │ 
   > 9 │ class Baz {
       │       ^^^
    10 │   init() {
    11 │     this.baz = new Baz();
```

## Test Plan

test cases in PR


ematipico:
We should probably resume this PR and merge it! @yeonjuan would you mind doing a rebase when you have a chance, please?

yeonjuan:
@ematipico Sure sorry for the late I'll rebase it.

ematipico (approved):

PR #1268, opened 2020-12-28
fix(formatter): handle callee parentheses

yeonjuan:
## Summary

In the below case, Rome "formatter" removes parentheses of the callee and IMO, it's incorrect results (change code behavior or syntax error)

* target.js
```js
(foo++)?.();
async () => {
  (await foo)?.();
}
(+foo)?.();
```

```
$ ./rome format target.js

foo++?.();
async () => {
    await foo?.();
};
+foo?.();
```

So, I made a PR handle it.

```
$ ./rome format target.js

(foo++)?.();
async () => {
    (await foo)?.();
};
(+foo)?.();
```

## Test Plan

test cases


ematipico (approved):
LGTM

PR #1269, opened 2020-12-28
fix(parser): parse catch param type in typescript

yeonjuan:
## Summary

Rome fails to parse catch error type.

* target.ts
```ts
try {}
catch (e: unknown) {}
```

```
$ ./rome parse target.ts

  ✖ Unclosed catch clause param

    1 │ try {}
  > 2 │ catch (e: unknown) {}
      │       ^

  ℹ We expected to find the closing character ) here
```

I made a PR to fix it

## Test Plan

test case


ematipico:
Hi @yeonjuan , thank you for catching this case! Do you think the formatter part works as it should?

yeonjuan:
@ematipico Yes I tested formatter also (in local) :) Should i add the test cases?

ematipico:
Personally, I think we should add a test, to avoid possible future regressions. (and we add a new test which is something where we lack at the moment)

yeonjuan:
@ematipico Thanks for the review :) I add a test case for formatter

ematipico (approved):

PR #1273, opened 2020-12-29
feat(formatter): handle TSAsExpression parentheses

yeonjuan:
## Summary

Current Rome adds parentheses on every `TSAsExpression`.

* target.ts
```ts
const foo = "const" as const;
foo(bar as any);
while (foo as any) {}
```

```
./rome format target.ts
```

result in.

```ts
const foo = ("const" as const);
foo((bar as any));
while ((foo as any)) {}
```

I believe it's not an expected result of formatting so made a PR handle it.


ematipico (approved):
It's amazing how we can see the value of changes directly to the Rome code base. LGTM!

PR #1276, opened 2020-12-29
fix(html-parser, formatter): handle boolean attribute

yeonjuan:
## Summary

Rome fails to parse the boolean attribute.

- target.html
```html
<input checked>
```

```
./rome parse target.html
```

I made a PR handle it.

## Test Plan

test case


ematipico (approved):
Thank you for looking into this!

PR #1280, opened 2021-01-01
feat(lint): add new a11y category

ematipico:
## Summary

**BREAKING CHANGE**

Moved `jsx-a11y/useValidLang` and `jsx-a11y/useHtmlLang` under a new category called `a11y`.

Now that we have a working html parser and we can have some rules that overlap with JSX, the lint rules should be now generic and handle different nodes.

This shows how powerful Rome is where we can have one single rule applied to different files based on the context. 

The previous rules `html/useValidLang` and `html/useHtmlLang` have been deleted and I moved their logic and tests under the new files. 

I updated also the test harness in order to handle multiple test cases. A test case now can be an array of objects, that's because we need to have a different file, which in turn creates a different AST. 

I reverted the changes I did inside the automated creation of the rules. Now they are as there were without the usage of `// rome-ignore`.

## Test Plan

Existing tests but reviewed.


ematipico:
@sebmck this is a breaking change, let me know what you think and if you're happy with the change.

scripts/generated-files/lint-rules-docs.ts
```diff
@@ -176,35 +176,70 @@ export async function main() {
...
+					} else {
+						const {invalid, valid} = singleCase;
+						if (invalid && invalid.length > 0) {
+							hasInvalid = true;
+						}
+						if (valid && valid.length > 0) {
+							hasValid = true;
+						}
 					}
 				}
```
sebmck, line 198: I think this would be a bit more DRY

```suggestion
				for (const elem of cases) {
					const cases = Array.isArray(elem) ? elem : [elem];
					for (const {invalid, valid} of cases) {
						if (invalid && invalid.length > 0) {
							hasInvalid = true;
						}
						if (valid && valid.length > 0) {
							hasValid = true;
						}
					}
				}
```

sebmck (commented):
This looks good to me! Curious what @bitpshr thinks, who did a lot of the consistency work around rule naming.

There's a conflict in `diagnostics/categories.ts` as I added an array of all valid category names so we can perform some proper validation too that will need to be updated.

Issue #984, opened 2020-08-08
☂️ CSS support


sebmck:
We already have the basis of a CSS formatter that @bitpshr has built out. Not sure how much more work needs to go into supporting additional syntax and testing.

## Linting

What sort of rules do we want? https://stylelint.io/ and https://postcss.org/ would be good references.

evilebottnawi:
@sebmck @ematipico  Need help? I have long dreamed of rewriting postcss/stylelint/cssnano (over the years, many problems have accumulated, especially related postcss - lack of opportunities of parsing). Many stylelint rules was written/rewritten by me, cssnano (minificator for CSS) lately supported by me, have worked on CSS formatting in prettier. But I would like to discuss some aspects a little, especial non standard syntax, like `less`/`scss`/`sass`/etc, improvements, fixes (small/not small roadmap).

evilebottnawi:
cc @sebmck @ematipico friendly ping

ematipico:
Hi @evilebottnawi , at the moment @bitpshr has made a stable base for the CSS parser although there are still some aspects missing (I can't remember the details but there some sub grammars that are not supported yet).

Although we could potentially start writing something around linting, which is the first step I think (together with the formatter).

About minification/purgin, there's nothing yet and having a good input on how to start is always welcome.

About non-standard syntax, I don't know if we are ready yet. SASS  and LESS are supersets of CSS so if we don't have a stable CSS parser, I don't think we can implement much around SASS and LESS. Theoretically we could start the foundations but I would wait and focus on what's more important.

ematipico:
We need to implement the following selector syntax: https://www.w3.org/TR/selectors-4/#syntax

evilebottnawi:
@ematipico Thanks for answer, I'll look at the code and get back to it in near future, maybe we should start with CSS formatter (without scss/less/etc), it should be easy and will allow finish parser, good parser = good linter

tomhodgins:
@ematipico Let me know if there's anything I can do to help, either by providing samples of valid CSS syntax, or hunting for things in specs (tokens, productions, etc). I've already got a bunch of tools I use in my own work for parsing or validating pieces of CSS syntax according to the spec :D

- tool to parse samples of CSS according to the parser entry points defined in the CSS syntax spec: https://tomhodgins.com/demo/css/

- tool to show you how the browser you are using parses a string as a CSS stylesheet: https://tomhodgins.com/demo/cssom

[...]

PR #1281, opened 2021-01-02
feat(css-parser): support selectors

yeonjuan:
## Summary
part of #984

This pr includes the implementation of parsing some CSS selectors.
It adds new CSS AST nodes, parsing, and formatting.

- Selectors
  -  CSSClassSelector : `.class`
  -  CSSIdSelector: `#id`
  - CSSTypeSelector: `type`
  - CSSPseudoClassSelector: `:class`
  - CSSPseudoElementSelector: `::element`
 - Combinators
   - descendant: ` `
   - child: `>`
   - nextSibiling: `+`
   - subsequentSibiling: `~`

## Test Plan

test case


internal/css-parser/parser/selectors.ts
```diff
@@ -0,0 +1,190 @@
...
+			patterns.push(combinator);
+		}
+
+		if (selector) {
+			patterns.push(selector);
+		}
+
+		if (!selector && !combinator) {
+			parser.unexpectedDiagnostic({
+				description: descriptions.CSS_PARSER.UNEXPECTED_TOKEN,
```
ematipico, line 166: Do you think it's possible to have a better error message here? Here we know what could be the issue, so emitting a more describing message would help to understand the cause of the problem.

internal/css-parser/test-fixtures/selectors/input.css
```diff
@@ -0,0 +1,3 @@
+div, div + #id, div ~ #id, div > #id {
```
ematipico, line 1: We should also add different cases, bad ones included. I try to suggest something:

```css
// this should be a parsing error
>, { 

}
```
```css
// this should work
span ~ + {} 
``` 
```css
+ {}
```

internal/formatter/test-fixtures/css/selectors/input.test.md
```diff
@@ -0,0 +1,94 @@
...
+div + #id {
+}
+
+div ~ #id {
+}
+
+div > #id {
+}
+
+div + #id, div ~ #id, div > #id {
```
ematipico, line 91: Now that we are starting to have something, I think we should start making some decisions about the CSS formatting.

Personally, I like having something like:


```css
div + #id, 
div ~ #id, 
div > #id {

}
```

This is my personal opinion. We can have a chat in some issue. What do you think?

internal/formatter/test-fixtures/css/selectors/input.test.md
```diff
@@ -0,0 +1,94 @@
...
```
ematipico, line 50: We should add a text case with long strings and see how the formatter behaves when it needs to go to a new line.

ematipico (commented):
Overall it looks good do me. I would add some bad case inside the parser in order to try to catch possible error and make them more descriptive. I would also try to add some edge case around the formatter

internal/formatter/test-fixtures/css/selectors/input.test.md
yeonjuan: @ematipico cool, make sense! it's actually what prettier does.

internal/css-parser/test-fixtures/selectors/input.css
yeonjuan: @ematipico Thanks :) I added a parsing error case also.

> // this should work
> span ~ + {} 
> + {}

As far as I know, these are invalid syntax right?

ematipico: Yeah my bad. The ones that I mentioned are invalid syntax.

These should be valid syntax

```css
.field + * {}
.field ~ * {}
.field * {}
```

yeonjuan: @ematipico Yes Thanks for the reply. Currently, This PR didn't contain parsing selector (`*`) implementation. I'll add that soon. :)

antonk52: It would be nice to add a test for `.long\.class\.name`, apparently class names can contain dots which should be escaped using `\` in css ie `.long\.class\.name {}`. Even though it's quite an edge case I think it would be nice to test against it.

ematipico: @yeonjuan  yes let's add the universal selector with a different PR. It makes sense! (sorry I thought it was already supported, my bad).

yeonjuan: @antonk52 
> It would be nice to add a test for .long\.class\.name, apparently class names can contain dots which should be escaped using \ in css ie .long\.class\.name {}. Even though it's quite an edge case I think it would be nice to test against it.

Sure. It makes sense. 👍  Thanks

yeonjuan:
@ematipico Hi Thanks for the reviewsI add descriptive error messages but not sure it's enough. Please give me feedback when you have time :)

ematipico (approved):
That's great! Let me know when you rebase the PR so we can merge it

yeonjuan:
@ematipico Thanks for the review! There was a minor bug in #1287 about the comment position and I added to fix the issue also - [code](https://github.com/rome/tools/pull/1281/commits/adb4b605f147fc2ffabf7dcb086424e8203b95a6)
I think It can be merged now Thanks :)

ematipico:
Cool! I noticed we still have some things to manage.

A style like this throws an error:

```css
.style {
	font-family: "Fira Code", monospace;
}
```

We don't parse the `,` after `"Fira Code"`.

yeonjuan:
@ematipico Yes you are right 👍  Could I work on that?

ematipico:
Sure! Thank you!!

PR #1287, opened 2021-01-07
fix(css-parser): handle comments

yeonjuan:
## Summary

Rome CSS parser didn't handle comment well

#### 1. parsing incorrectly
  - source
```css
a {
	color: white /* comment */;
}
```
- result
```js
CSSRoot {
  comments: Array []
  corrupt: false
  diagnostics: Array []
  filename: "project-rome/test.css"
  integrity: Object {hash: "b498e368a376f3b09805119693a54a438ce83583f4371fca48e54de4e1626cbb"}
  body: Array [
    CSSRule {
      prelude: Array [CSSIdentifier {value: "a"}]
      block: CSSBlock {
        value: Array [
          CSSDeclaration {
            name: "color"
            value: Array [
              CSSIdentifier {value: "white"}
              CSSRaw {value: " comment "} // Ops!. parsed as a CSSRaw
            ]
            important: false
          }
        ]
        startingTokenValue: "{"
      }
    }
  ]
}
```

#### 2. parsing failed

- source
```css
a {
	color: white;  /* comment */
}
```
- result
```
      ✖ Invalid declaration

      ⚠ This file has been changed since the diagnostic was produced and may be out of date
```

As far as I know, css comments can be placed almost anywhere between tokens. So I think It should be handled before calling some token methods of the parser (`eatToken`, `nextToken`, `matchToken`).

I'd like to fix this problem before #1281 merged :)

## Test Plan

test case


internal/css-parser/tokenizer/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function matchToken(parser: CSSParser, type: TokenType) {
+	skipComments(parser);
+	return parser.matchToken(type);
+}
+
+export function readToken(parser: CSSParser, type: TokenType) {
+	skipComments(parser);
+	return parser.eatToken(type);
+}
```
ematipico, line 31: I really like these abstractions!

ematipico (approved):

PR #1290, opened 2021-01-08
feat(css-parser): support universal selector

yeonjuan:
## Summary
part of #984
This PR is for support parsing universal selector (`*`).

## Test Plan

test case.


internal/css-parser/parser/selectors.ts
```diff
@@ -146,9 +159,11 @@ function tryParseSelector(parser: CSSParser) {
 	} else if (matchToken(parser, "Ident")) {
 		return parseTypeSelector(parser);
 	} else if (matchToken(parser, "Delim")) {
-		const token = readToken(parser, "Delim") as Tokens["Delim"];
```
yeonjuan, line 149: It was a bug. "Delim" token should not be consumed before parsingSelector

internal/css-parser/parser/selectors.ts
```diff
@@ -91,6 +92,18 @@ function parsePseudoSelector(
 	return undefined;
 }
 
+function parseUniversalSelector(parser: CSSParser): CSSUniversalSelector {
+	const start = parser.getPosition();
+	parser.expectToken("Delim");
```
ematipico, line 97: `expectToken` should be avoided because it throws an hard error, making the parser not recoverable. A better strategy would be to return `CSSUniversalSelector | undefined` and use `parser.matchToken()`

ematipico (changes requested):
It looks good, I would review the part where we check the token

internal/css-parser/parser/selectors.ts
yeonjuan: >  it throws an hard error, making the parser not recoverable

Oh got it. Thanks for the kind explanation :)

ematipico (approved):

PR #1292, opened 2021-01-09
feat(css-parser): support attribute selector

yeonjuan:
## Summary

part of #984

It's for support parsing attribute selectors. - [spec](https://www.w3.org/TR/selectors-4/#attribute-representation)

## Test Plan

test case


internal/css-parser/parser/selectors.ts
```diff
@@ -138,6 +142,131 @@ function tryParseCombinator(parser: CSSParser): CSSCombinator | undefined {
...
+	return value;
+}
+
+function parseAttributeSelector(
+	parser: CSSParser,
+): CSSAttributeSelector | undefined {
+	const start = parser.getPosition();
+	if (matchToken(parser, "LeftSquareBracket")) {
+		parser.nextToken();
+		readToken(parser, "Whitespace");
```
ematipico, line 207: Just noticed now but the implementation of `readToken` uses `parser.eatToken`. Should we call this function `eatToken` instead?

internal/css-parser/parser/selectors.ts
```diff
@@ -138,6 +142,131 @@ function tryParseCombinator(parser: CSSParser): CSSCombinator | undefined {
...
+			const matcher = parseAttributeMatcher(parser);
+			readToken(parser, "Whitespace");
+
+			const value = matcher && parseAttributeValue(parser);
+			readToken(parser, "Whitespace");
+
+			let modifier: AttributeModifier | undefined;
+			if (matchToken(parser, "Ident")) {
+				const ident = parser.getToken() as Tokens["Ident"];
+				if (ident.value === "i") {
```
ematipico, line 232: We should add `s` too

internal/css-parser/parser/selectors.ts
```diff
@@ -138,6 +142,131 @@ function tryParseCombinator(parser: CSSParser): CSSCombinator | undefined {
...
+
+function parseAttributeSelector(
+	parser: CSSParser,
+): CSSAttributeSelector | undefined {
+	const start = parser.getPosition();
+	if (matchToken(parser, "LeftSquareBracket")) {
+		parser.nextToken();
+		readToken(parser, "Whitespace");
+
+		if (matchToken(parser, "Ident")) {
```
ematipico, line 209: Here too, having:
```ts
if (!matchToken(parser, "Ident")) {
  parser.unexpectedDiagnostic();
  return undefined;
}
```

would help the reading of the code.

internal/diagnostics/descriptions/parsers/cssParser.ts
```diff
@@ -14,7 +14,14 @@ export const cssParser = createDiagnosticsCategory({
...
+		message: markup`Expected a left bracket <emphasis>{</emphasis>`,
 	},
 	UNEXPECTED_EMPTY_SELECTOR: {message: markup`Unexpected empty selectors`},
+	EXPECTED_CLOSING_ATTRIBUTE_SELECTOR: {
+		message: markup`Expected to close attribute selector with a right bracket <emphasis>]</emphasis>`,
+	},
+	UNKNOWN_ATTRIBUTE_MATCHER: {
+		message: markup`Unknown attribute matcher`,
+	},
+	UNKNOWN_ATTRIBUTE_MODIFIER: {message: markup`Unknown attribute modifier`},
```
ematipico, line 26: I would transform this into a function and pass the incorrect value and show it with `<emphasis>`.

Also, I would add another log telling to the user that the only correct values are `s` and `i`.

internal/diagnostics/descriptions/parsers/cssParser.ts
```diff
@@ -14,7 +14,14 @@ export const cssParser = createDiagnosticsCategory({
 	EXPECTED_IDENTIFIER: {message: markup`Expected an identifier`},
 	EXPECTED_SELECTOR: {message: markup`Expected a selector`},
 	EXPECTED_LBRACKET: {
-		message: markup`Expected a left bracket </emphasis>{</emphasis>`,
+		message: markup`Expected a left bracket <emphasis>{</emphasis>`,
```
ematipico, line 17:
```suggestion
		message: markup`Expected a left curly bracket <emphasis>{</emphasis>.`,
```

We could show the character that we actually get instead of the bracket

internal/diagnostics/descriptions/parsers/cssParser.ts
```diff
@@ -14,7 +14,14 @@ export const cssParser = createDiagnosticsCategory({
...
 	},
 	UNEXPECTED_EMPTY_SELECTOR: {message: markup`Unexpected empty selectors`},
+	EXPECTED_CLOSING_ATTRIBUTE_SELECTOR: {
+		message: markup`Expected to close attribute selector with a right bracket <emphasis>]</emphasis>`,
```
ematipico, line 21:
```suggestion
		message: markup`Expected to close attribute selector with a right square bracket <emphasis>]</emphasis>.`,
```

We could add the character we got instead of the square bracket

internal/diagnostics/descriptions/parsers/cssParser.ts
```diff
@@ -14,7 +14,14 @@ export const cssParser = createDiagnosticsCategory({
...
+	},
+	UNKNOWN_ATTRIBUTE_MATCHER: {
+		message: markup`Unknown attribute matcher`,
```
ematipico, line 24: We could suggest the available/correct attribute matchers. We have an utility called `buildSuggestionAdvice`

internal/formatter/builders/css/selectors/CSSAttributeSelector.ts
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	if (node.modifier) {
+		tokens.push(space);
+		tokens.push(node.modifier);
+	}
+
+	tokens.push("]");
+
+	return concat(tokens);
```
ematipico, line 25: I am not sure if this suggestion is correct (I don't have extensive knowledge of the formatter yet), but should we _group_ the whole attribute selector? We should add a test with long strings and make sure that the whole attribute selector is not broken on multiple lines.

internal/css-parser/test-fixtures/selectors/input.css
```diff
@@ -5,3 +5,15 @@ div, div + #id, div ~ #id, div > #id {
...
+
+a[title] {
+  color: purple;
+}
+
+a[title = "title"] {
+  color: purple;
+}
+
+a[title~="title" i] {
```
ematipico, line 17: Could we add another example with multiple attribute selectors?

```css
a[title = "title"][href="image/*"] { // let's keep the asterisk and make sure is parsed as string
  color: purple;
}
```

ematipico (changes requested):
Thank you for looking into this! Added few suggestions on making the errors a bit more descriptive and few considerations around the code

internal/formatter/builders/css/selectors/CSSAttributeSelector.ts
yeonjuan: @ematipico Thanks, I added a test case :) - [test case](https://github.com/rome/tools/pull/1292/files#diff-2436033c538e80c8f5af882dbc7b5076fa7078b3b192090777c83c748fa5a5beR130) .As far as I know, what "group" does is grouping the node's indent level :)

internal/diagnostics/descriptions/parsers/cssParser.ts
yeonjuan: @ematipico Thanks! I added `buildSuggestionAdvice` but I'm not sure but it does not work. It seems a bug in `buildSuggestionAdvice`.

```js
buildSuggestionAdvice("*", ["*="]); // it returns []
```

internal/css-parser/parser/selectors.ts
```diff
@@ -138,6 +142,131 @@ function tryParseCombinator(parser: CSSParser): CSSCombinator | undefined {
 	return undefined;
 }
 
+function parseAttributeMatcher(parser: CSSParser): AttributeMatcher | undefined {
+	let matcher: string | undefined;
```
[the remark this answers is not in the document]
yeonjuan: @ematipico Thanks :)
The matcher could include two tokens so it should be concatenated (`+=`) - [line](https://github.com/rome/tools/pull/1292/files/4da2e99128903604e6514c3a324691bff318547b#diff-0a2350dfc20a66ba1dc400464afaa6c9da34949e8238042903dc7513cf5f8d1eR163)

That's why I use `string` here :) Any good typing Idea?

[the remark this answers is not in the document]
ematipico: I do have an idea, we can use the new template literal types:

```ts
export type Matcher = "~" | "|" | "^" | "$" | "*" | "";

export type AttributeMatcher = `${Matcher}=`;
```

So now we have two types. `matcher` can start with type `Matcher` and when we do the concatenation we can cast it to `AttributeMatcher`. How does it sound? It's not the best but we can narrow down the type from `string` to `Matcher`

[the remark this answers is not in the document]
yeonjuan: @ematipico 
Thanks, nice Idea 👍
but actually, the template literal type does not make type checking strong for concatenated string - [demo](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCyBDGBjAFsKOAXjgCIA-UuAHzOqttID0GyASV0gKk9IG4AUANCRYCZKgCCMGFACWAIwCuMYJhz5CJAAZsA3urwEAvkW2CBAG2DwAtliNQAXOgeaLcgGZwAFAEYASjg9AXsNAmIyfgFjOGBLAGdUELhUuDDHSO5otPS3CIBqElJPCFLyssrS6OMhbAgAOwSIawA6SwgAcx8MzQCLASA) It's a part where ts cannot handle well

Hmm. what about an user defined type guard ?
```ts
function isAttributeMatcher(value: string): value is AttributeMatcher {
	return ATTRIBUTE_SELECTOR_MATCHERS.includes(value);
}
```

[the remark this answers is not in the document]
ematipico: Sounds like a nice idea!

ematipico (approved):

ematipico:
Hi @yeonjuan if the PR is ready, could you please rebase it? After that, we can merge it

yeonjuan:
@ematipico  Thanks for the review:) I rebased it

ematipico (approved):

PR #1296, opened 2021-01-13
feat(css-formatter): support number and percentage

yeonjuan:
## Summary

support `CSSNumber` and `CSSPercentage` formatting

## Test Plan

test cases


internal/formatter/builders/css/core/CSSPercentage.ts
```diff
@@ -1,9 +1,9 @@
 import {CSSPercentage} from "@internal/ast";
-import {Builder, Token} from "@internal/formatter";
+import {Builder, Token, concat} from "@internal/formatter";
 
 export default function CSSPercentage(
 	builder: Builder,
 	node: CSSPercentage,
 ): Token {
-	throw new Error("unimplemented");
+	return concat([String(node.value), "px"]);
```
yeonjuan, line 8: Wrong

ematipico, line 8: I think the correct implementation should be `%`
```suggestion
	return concat([String(node.value), "%"]);
```

Also, we should implement a test, so we should add something like:
```css
width: 80%;
```

ematipico (changes requested):

internal/formatter/builders/css/core/CSSPercentage.ts
yeonjuan: @ematipico Yes. right I found that while double check. Thanks 👍

ematipico (approved):

PR #1298, opened 2021-01-13
feat(compiler): foundation css prefix

jer3m01:
## Summary

This PR is a follow up to #1288, it's discussion is #1293.
Thanks to @ematipico for the initial implementation and help.

This is still in progress.

## Test Plan
Includes a test for its utils and a fixture for each of its prefixes.


internal/compiler/transforms/compile/css-handler/prefix/prefixes/transition.ts
```diff
@@ -0,0 +1,68 @@
...
+
+
+export default createPrefixVisitor({
+	name: "css-prefix",
+	enter(path, prefixConfig) {
+
+		const prefixes = new Set<string>(); // https://caniuse.com/?search=transition
+		if (matchBrowser(prefixConfig.target, "firefox 4-15")) prefixes.add("-moz-");
+
+		if (matchBrowser(prefixConfig.target, "chrome 4-25")
```
ematipico, line 21: I think the visitor should not know anything about browser versions. It's not his responsibility.

I think we should pass it an API that has more power, not just `prefixConfig`. We should pass an API that should do something like:

```js
enter(path, browser) {
	if (!browser.supports({ property: "transition" })) {
		// it could be a Set(["-webkit-", "-moz-"]);
		const prefixes = browser.retrievePrefixes({ { property: "transition" } });
	}
}
```

internal/compiler/transforms/compile/css-handler/prefix/prefixes/transition.ts
```diff
@@ -0,0 +1,68 @@
...
+} from "@internal/ast";
+import {
+	createPrefixVisitor,
+	matchBrowser,
+	nodeHasPrefix
+} from "@internal/compiler/transforms/compile/css-handler/prefix/utils";
+
+
+export default createPrefixVisitor({
+	name: "css-prefix",
```
ematipico, line 15: I think we should doing like this: in this visitor we pass the category `"transition"` and inside `createPrefixVisitor` we prepend it with `"css-prefix"` so all the visitors end up having, as a name `"css-prefix/transition"`.

Having more detailed name would help in the future

internal/compiler/transforms/compile/css-handler/prefix/prefixes/transition.ts
jer3m01: Yes totally! 

This was done before I opened and you commented #1297.
I'll put this on pause until we have a basic implementation of `internal/codec-browsers` & `internal/browser-features`.

jer3m01:
>I'll put this on pause until we have a basic implementation of `internal/codec-browsers` & `internal/browser-features`.

jer3m01:
Now that we have `browser-features` I've refactored the code to use it!
Also added a method to rename properties/values as some require "hacks" to be correct.

Would want some input from @sebmck for how to structure the target browser option.

ematipico (commented):
This is coming out real nice!

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,229 @@
...
+		if (node.value && node.value.length > 0) {
+			const propertyIndex = nodePropertyValueIndex(node, propertyName, value);
+			if (propertyIndex > -1) {
+				const property = node.value[propertyIndex] as CSSDeclaration;
+				const newDeclarations = [];
+				const prefixes = new Set(
+					targets.filter((browser) =>
+						browser.cssFeatureRequiresPrefix(browserFeaturesPropertyName)
+					).map((browser) => browser.getPrefix()),
+				);
```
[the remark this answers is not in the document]
jer3m01: I added cache on the browser class but in this method it wouldn't do anything as they're different for each property.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+export function wrapPrefixVisitor<State extends UnknownObject>(
+	visitor: PrefixVisitor<State>,
+	targets: Browser[],
+) {
+	return createVisitor<State>({
+		name: visitor.name,
+		enter: visitor.enter
+			? (path, state) => visitor.enter!(path, targets, state)
```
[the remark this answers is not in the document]
jer3m01: `TS2722: Cannot invoke an object which is possibly 'undefined'.`

internal/compiler/transforms/compile/css-handler/prefix/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import transform from "@internal/compiler/transforms/compile/css-handler/prefix/prefixes/transform";
+import display from "@internal/compiler/transforms/compile/css-handler/prefix/prefixes/display";
+import {UnknownObject} from "@internal/typescript-helpers";
+import {resolveBrowsers} from "@internal/codec-browsers";
+
+const prefixVisitors: PrefixVisitor<UnknownObject>[] = [
+	transition,
+	transform,
+	...display,
+];
```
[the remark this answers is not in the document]
jer3m01: Is there any way we could detect the export of the file? i.e. if its a single visitor add it to the generated array and if its an array of visitors add it to the array with `...`.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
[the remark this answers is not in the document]
ematipico: I think we should avoid using `!`, we are basically bypassing the compiler. Unless it's really necessary, it might be useful leaving a comment saying why.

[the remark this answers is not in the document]
jer3m01: Made it more explicit but feel like this is a typescript issue, the variable's been checked for null/undefined beforehand and should not throw that error.

internal/compiler/transforms/compile/css-handler/prefix/index.ts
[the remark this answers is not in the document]
jer3m01: Instead I just made all prefixVisitors return an array. Thanks for the idea of generating them.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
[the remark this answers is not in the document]
ematipico: It doesn't seem a bug to me. The interface `PrefixVisitor` marks `enter` and `exit` as possibly undefined https://github.com/jer3m01/rome/blob/feat/css-prefix/internal/compiler/transforms/compile/css-handler/prefix/utils.ts#L80

Where's the check against `null`/`undefined`? I might have missed it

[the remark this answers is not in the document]
jer3m01:
```ts
 enter: visitor.enter // checks if not null/undefined, latest commit linked below
	? (path, state) => visitor.enter!(path, targets, state)
	: undefined,
```
https://github.com/rome/tools/pull/1298/files#diff-5fc9fca9e5232dab4b0ddf9b1dfc6ac6790ff4338dead8f534b1ca45269b78adR108-R110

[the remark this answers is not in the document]
ematipico: That's weird indeed...

internal/compiler/transforms/compile/css-handler/prefix/prefixes/display.ts
```diff
@@ -0,0 +1,38 @@
...
+	createPrefixVisitor,
+	prefixCSSValue,
+} from "@internal/compiler/transforms/compile/css-handler/prefix/utils";
+
+// https://github.com/Fyrd/caniuse/blob/main/features-json/flexbox.json
+export default [
+	createPrefixVisitor({
+		name: "display/flex",
+		enter(path, targets) {
+			return prefixCSSValue({
```
sebmck, line 11: Could we wrap the `prefixCSSValue` calls that are in the visitors and inline it as: `if (node.type === "") {return prefixCSSValue(...);} else {return signals.retain}"`? We'll be creating a pretty large object for every single node which is potentially pretty expensive, we try to bail out as quickly as possible.

internal/compiler/transforms/compile/css-handler/prefix/prefixes/display.ts
```diff
@@ -0,0 +1,38 @@
...
+				path,
+				propertyName: "display",
+				value: "flex",
+				browserFeaturesKey: "flexbox",
+				targets,
+				rename: (propertyName) =>
+					propertyName === "-moz-flex" ? "-moz-box" : propertyName
+				,
```
sebmck, line 19: Since this is only refers to a parameter we can also put it at the top level and refer to it with a variable. That way we aren't creating new functions on every call, which can be expensive in aggregate.

internal/compiler/transforms/compile/css-handler/prefix/prefixes/display.ts
```diff
@@ -0,0 +1,38 @@
...
+			});
+		},
+	}),
+	createPrefixVisitor({
+		name: "display/inline-flex",
+		enter(path, targets) {
+			return prefixCSSValue({
```
sebmck, line 26: Ditto here.

internal/compiler/transforms/compile/css-handler/prefix/prefixes/transform.ts
```diff
@@ -0,0 +1,19 @@
...
+	createPrefixVisitor,
+	prefixCSSProperty,
+} from "@internal/compiler/transforms/compile/css-handler/prefix/utils";
+
+// https://github.com/Fyrd/caniuse/blob/main/features-json/transforms2d.json
+export default [
+	createPrefixVisitor({
+		name: "transform",
+		enter(path, targets) {
+			return prefixCSSProperty({
```
sebmck, line 11: And here

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,246 @@
...
+	}
+	return false;
+}
+
+export function nodePropertyIndex(node: CSSBlock, property: string): number {
+	return node.value
+		? node.value.findIndex((n) =>
+				n.type === "CSSDeclaration" && n.name === property
+			)
+		: -1;
```
sebmck, line 60: Not a fan of multiline ternaries, at least how we format this one:

```suggestion
	if (node.value === undefined) {
		return node.value.findIndex((n) =>
			n.type === "CSSDeclaration" && n.name === property
		);
	} else {
		return -1;
	}
```

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,246 @@
...
+			n.name === property &&
+			n.value.length === 1 &&
+			n.value[0]?.type === "CSSIdentifier" &&
+			n.value[0].value === rename(`-${prefix}-${value}`)
+		);
+	}
+	return false;
+}
+
+export function nodePropertyIndex(node: CSSBlock, property: string): number {
```
sebmck, line 55: Is there a better name for this? It wasn't immediately clear to me what this does by the function name alone. Maybe `findPropertyIndexByName`?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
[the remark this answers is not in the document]
sebmck: The visitor object is mutable, so TypeScript assumes that the property can be changed inside of functions and removes the refinement.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,246 @@
...
+) {
+	return createVisitor<State>({
+		name: visitor.name,
+		enter: visitor.enter != null
+			? (path, state) => visitor.enter!(path, targets, state)
+			: undefined,
+		exit: visitor.exit != null
+			? (path, state) => visitor.exit!(path, targets, state)
+			: undefined,
+	});
```
sebmck, line 114: If we assign the properties to variables first then TypeScript wont lose the refinement.

```suggestion
    const {enter, exit} = visitor;
	return createVisitor<State>({
		name: visitor.name,
		enter: enter === undefined
			? undefined
			: (path, state) => enter(path, targets, state),
		exit: exit === undefined
			? undefined
			: (path, state) => exit(path, targets, state),
	});
```

internal/compiler/transforms/index.ts
```diff
@@ -36,6 +36,7 @@ import scopedRomeTransform from "./compileForBundle/scopedRomeTransform";
 import asyncImportTransform from "./compileForBundle/asyncImportTransform";
 import inlineEnv from "./compileForBundle/inlineEnv";
 import {commentInjectorVisitor, variableInjectorVisitor} from "./helpers";
+import {default as cssPrefix} from "@internal/compiler/transforms/compile/css-handler/prefix";
```
sebmck, line 39: We should differentiate this from the other imports that are just plain objects to be a bit clrearer

```suggestion
import {default as createCSSPrefixVisitors} from "./compile/css-handler/prefix";
```

internal/compiler/transforms/index.ts
```diff
@@ -60,6 +61,7 @@ export const stageTransforms: TransformStageFactories = {
 		templateLiterals,
 		callSpread,
 		enums,
+		...cssPrefix(projectConfig, options),
```
sebmck, line 64:
```suggestion
		...createCSSPrefixVisitors(projectConfig, options),
```

internal/core/server/ServerRequest.ts
```diff
@@ -823,6 +823,9 @@ export default class ServerRequest {
 			inlineSourceMap: false,
 			cwd: this.client.flags.cwd,
 			resolver: mergeObjects(this.getResolverOptionsFromFlags(), resolverOpts),
+			target: this.normalizedCommandFlags.flags != null
```
sebmck, line 826: This assumes it's always called from inside `bundle`, when we also call it from other commands. If you want something command-specific then you can just leave it out.

You could then modify `Bundler.createFromServerRequest` to accept a `Partial<BundlerConfig>` that's combined with the object it gets from here, and then provide it in the `bundle` command.

internal/project/load.ts
```diff
@@ -439,13 +439,13 @@ export async function normalizeProjectConfig(
...
-			const target: ProjectConfigTarget = {
-				constraints: object.get("constraints").asImplicitMappedArray((item) =>
-					item.asString()
-				),
-			};
 			object.enforceUsedProperties("target config property");
-			config.targets.set(name, target);
+			config.targets.set(
+				name,
+				resolveBrowsers(object.asImplicitMappedArray((item) => item.asString())).map((
```
sebmck, line 445: If we have syntax errors when parsing the queries then we aren't going to have source locations which will be hard to debug. It's probably ok for now, as `resolveBrowsers` can't take an array of `ParserOptions` but we should come back to it later.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,246 @@
...
+		if (node.value && node.value.length > 0) {
+			const propertyIndex = nodePropertyIndex(node, propertyName);
+			if (propertyIndex > -1) {
+				const property = node.value[propertyIndex] as CSSDeclaration;
+				const newDeclarations = [];
+				const prefixes = new Set(
+					targets.filter((browser) =>
+						browser.cssFeatureRequiresPrefix(browserFeaturesKey)
+					).map((browser) => browser.getPrefix()),
+				);
```
sebmck, line 150: Could this be simplified as:

```suggestion
				const prefixes = new Set();

				for (const browser of targets) {
					if (browser.cssFeatureRequiresPrefix(browserFeaturesKey)) {
						prefixes.add(browser.getPrefix());
					}
				}
```

Before we were doing two loop iterations, creating functions, and had more indirection. We should optimize and consider the case where we have literally megabytes of CSS in a file, and these sorts of operations add up.

internal/compiler/transforms/compile/css-handler/prefix/index.ts
```diff
@@ -0,0 +1,31 @@
...
+export default (projectConfig: ProjectConfig, options: CompilerOptions) =>
+	prefixVisitors.map((prefixVisitor) =>
+		wrapPrefixVisitor(
+			prefixVisitor,
+			projectConfig.targets.get(options.target ?? "default")?.map((browserProps) =>
+				getBrowser(browserProps)
+			) ?? [],
+		)
+	)
+;
```
sebmck, line 31: The expressions here are pretty complex, and it's hard to tell at a glance how all the `?` interact. Might be clearer like this:

```suggestion
export default (projectConfig: ProjectConfig, options: CompilerOptions) => {
	const propsTargets = projectConfig.targets.get(options.target ?? "default") ?? [];
	const targets = propsTargets.map(props => getBrowser(props));
	return prefixVisitors.map((prefixVisitor) => wrapPrefixVisitor(prefixVisitor, targets));
};
```

internal/browser-features/Browser.ts
```diff
@@ -66,10 +67,12 @@ export abstract class Browser {
 	}
 
 	public getPrefix(): string {
-		return (
-			this.getVersionConsumer().get("p").asStringOrVoid() ??
-			this.getDefaultPrefix()
-		);
+		if (this.prefix == null) {
```
sebmck, line 70: We never actually produce `null` inside of Rome, it's always `undefined`. The only time we use `== null` is when dealing with `String.prototype.match` return values.

It's longer, but more explicit to do:

```suggestion
		if (this.prefix === undefined) {
```

internal/compiler/transforms/compile/css-handler/prefix/prefixes/transform.ts
jer3m01: Not quite sure what you mean, this one doesn't have a renamer.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,284 @@
...
+			n.value[0].value === value
+		);
+	} else {
+		return -1;
+	}
+}
+
+type PrefixCompilerPath = CompilerPath & {
+	node: CSSBlock & {
+		value: Array<AnyCSSValue | CSSRule | CSSAtRule | CSSDeclaration>;
```
ematipico, line 90: This here. I know we have it inside `CSSBlock` as optional but I wouldn't cheat the compiler like this. If there's a specific reason for this, we should create a user guard to check the value.

Also, I would move `Array<AnyCSSValue | CSSRule | CSSAtRule | CSSDeclaration>` inside `ast/css/unions.ts`, call it `CSSBlockValue` and use it here and inside `CSSBlock`

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,284 @@
...
+	visitor: PrefixVisitor<State>,
+): Visitor<State> {
+	return {
+		name: `css-handler/prefix/${visitor.name}`,
+		enter: (path: CompilerPath, state: VisitorStateEnter<State>) => {
+			if (
+				visitor.enter !== undefined &&
+				path.node.type === "CSSBlock" &&
+				path.node.value &&
+				path.node.value.length > 0
```
ematipico, line 116: This could become a user defined guard so we wouldn't cheat the compiler:

```ts
function assertCssBlockValue(path): path is PrefixCompilerPath {
	return (
		path.node.type === "CSSBlock" &&
		path.node.value &&
		path.node.value.length > 0
	)
}
```

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,284 @@
...
+}
+
+export function findPropertyIndex(node: CSSBlock, property: string): number {
+	if (node.value !== undefined) {
+		return node.value.findIndex((n) =>
+			n.type === "CSSDeclaration" && n.name === property
+		);
+	} else {
+		return -1;
+	}
```
ematipico, line 67:
```suggestion
	} 
	return -1;
```

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,284 @@
...
+		return node.value.findIndex((n) =>
+			n.type === "CSSDeclaration" &&
+			n.name === property &&
+			n.value.length === 1 &&
+			n.value[0]?.type === "CSSIdentifier" &&
+			n.value[0].value === value
+		);
+	} else {
+		return -1;
+	}
```
ematipico, line 85:
```suggestion
	} 
	return -1;
```

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,284 @@
...
+}
+
+type PrefixCompilerPath = CompilerPath & {
+	node: CSSBlock & {
```
ematipico, line 89: We have an utility for this case

```ts
RequiredProps<CSSBlock, "value">
```

It's inside `internal/typescript-helpers`

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,284 @@
...
+export function prefixCSSProperty(
+	{
+		path,
+		propertyName,
+		browserFeaturesKey,
+		rename = (propertyName) => propertyName,
+	}: PrefixCSSPropertyProps,
+) {
+	const {node} = path;
+	const propertyIndex = findPropertyIndex(node, propertyName);
```
ematipico, line 179: `findPropertyIndex` should return the index **AND** the `CSSDeclaration`, so we don't need to do a cast.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
ematipico: Yes, and instead of having `if ()`, you could call the type guard. That's what I meant

jer3m01: Could you verify I've changed what you wanted.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+export function createPrefixVisitor<State extends UnknownObject>(
+	visitor: PrefixVisitor<State>,
+): Visitor<State> {
+	return {
+		name: `css-handler/prefix/${visitor.name}`,
+		enter: (path: CompilerPath, state: VisitorStateEnter<State>) => {
+			if (visitor.enter !== undefined && isCssBlockAndHasValue(path)) {
+				return visitor.enter(path as PrefixCompilerPath, state);
```
ematipico, line 86: I think `path as PrefixCompilerPath` because the user type guard `isCssBlockAndHasValue` makes the casting for us

PR #1302, opened 2021-01-15
docs: update README logo to support dark mode

souporserious:
## Summary

Noticed the README logo wasn't showing up well when using dark mode so hope this can help fix that 🙂. I pulled the logo from the site and updated it to change depending on the user's color-scheme preference.


sebmck:
Nice! I'm not sure pulling from `prefers-color-scheme` is reliable though. eg. I use a dark mode OS but find the dark mode GitHub hard to read so I use light mode. Is there any way we could reliably detect that?

assets/SVG/logo_readme.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="1620" height="401" viewBox="0 0 1620 401" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <style>
+    .text {
+      fill: #27272A;
+    }
+    @media (prefers-color-scheme: dark) {
```
jer3m01, line 6: You could use the `data-color-mode` attribute from the html tag instead.

When using dark mode github sets it to `dark`.
image

souporserious: Oh, this is great! I'll get this updated when I get a chance.

souporserious:
@sebmck yeah I had thought about that afterward. I think @jer3m01's comment should do the trick! We can use `data-color-mode` as a better heuristic.

souporserious:
Hmm so it doesn't seem to work since the SVG is hosted on it's own there is no access to `data-color-mode` 😞 and inlining doesn't work either. Maybe we should we wait until GitHub potentially has a first-class way to handle this?

sebmck:
Yeah I think we'll have to. We could make the experience better though in dark mode by adding a white border around everything? That way there's at least some contrast in dark mode and it'd be hidden in light mode. What do you think?

souporserious:
@sebmck yeah, I think that's a great idea! Just updated it, I think it looks pretty good :)

souporserious:
Ah, actually let me rework it and flatten the path so it doesn't mess with the logo bounds since SVG aligns it to the center.

jer3m01:
Inlining the svg works, your css was incorrect.

Try with
```css
.text {
    fill: #27272A;
}
@media (prefers-color-scheme: dark) {
    html[data-color-mode="auto"] .text {
        fill: #dfdfe1;
    }
}
html[data-color-mode="dark"] .text {
    fill: #dfdfe1;
}
```

souporserious:
@jer3m01 thanks! Updated it, although, it seems that GitHub is stripping it out 🤔 I can view it locally in VS Code which is weird.

jer3m01:
Yeah that sucks, also tried using the current text color and css variables but nothing works.

Personally not a huge fan of the white outline, rather see the logo as it is.
But that's just me.

souporserious:
Ah, lame 😞. @sebmck moved it back to the outlined PNG if you want to use that, no worries if you just want to close this and leave as-is.

sebmck:
Hmm doesn't look as good as it could but I do think it's a net improvement since at least now there's some contrast. I'll merge it as-is, thank you @souporserious!

Screen Shot 2021-02-03 at 4 44 27 PM

souporserious:
No problem! Anything you were specifically looking for? Happy to do another pass to clean it up more.

sebmck:
Oh sorry I meant the best improvement would be first class for dark mode images by GitHub. I think doubling the stroke width might help even more though.

PR #1306, opened 2021-01-16
feat(css-formatter): support function

yeonjuan:
## Summary

It's a pr supporting for CSS function formating.

## Test Plan

test case.


internal/css-parser/index.ts
```diff
@@ -914,6 +914,17 @@ function parseComponentValue(
 		);
 	}
 
+	if (matchToken(parser, "Comma")) {
```
ematipico, line 917: I think the Comma should have its own node: https://www.w3.org/TR/css-syntax-3/#parse-comma-separated-list-of-component-values

internal/formatter/test-fixtures/css/function/input.css
```diff
@@ -0,0 +1,9 @@
+
```
ematipico, line 1: I think we should add also a nested function case:

```css
padding: calc(calc(8px-2px) + 8px);
```

ematipico (changes requested):
Thank you! Just a couple of suggestions

yeonjuan:
@ematipico Thanks for the review. all fixed :)

ematipico (approved):
Thank you!!

PR #1308, opened 2021-01-21
fix: add avatar img alt

cyphernull:
## Summary

adding alt attribute is more compliant to ``


cyphernull:
Is applying username string to alt attr better ? 🙂

ematipico:
Yes, much better!

ematipico (approved):

PR #1317, opened 2021-02-03
feat: add new internal/browsers-db

jer3m01:
## Summary
`internal/browsers-db` #1297

This adds a new internal package containing browser data, currently only contains data about css features.

The data is generated from caniuse.com using `scripts/generated-files/browsers-db.ts`
## Test Plan
`rome check --apply`


scripts/generated-files/browsers-db.ts
```diff
@@ -0,0 +1,546 @@
...
+			return {
+				lines: [`export default ${JSON.stringify(data)}`],
+				hash: JSON.stringify(data),
+			};
+		},
+	);
+}
+
+async function updateRegions() {
+	let regionsUsage = (<regionsFormat>{});
```
ematipico, line 493:
```suggestion
	let regionsUsage: RegionsFormat = {};
```

scripts/generated-files/browsers-db.ts
```diff
@@ -0,0 +1,546 @@
...
+
+	for (const region of REGIONS) {
+		// Can't use .forEach because of await
+		console.log(`Updating region "${region}"...`);
+
+		const rawRegionUsage = await get<{
```
ematipico, line 499: Could we move this inline type inside an interface?

scripts/generated-files/browsers-db.ts
ematipico, line 493: This could be a `Map`

scripts/generated-files/browsers-db.ts
```diff
@@ -0,0 +1,546 @@
...
+						p: v.prefix!.length === 0 ? undefined : v.prefix,
+					})),
+					cv: rawData.agents[agent].current_version,
+				};
+			}
+			return temp;
+		})(),
+		categories: rawData.cats,
+		data: (() => {
+			let temp = (<dataFormat["data"]>{});
```
ematipico, line 437: This could be a `Map`

scripts/generated-files/browsers-db.ts
```diff
@@ -0,0 +1,546 @@
...
+				[key: string]: {
+					[key: string]: number;
+				};
+			};
+		}>(regionUsageUrl.replace("<REGION>", region));
+
+		regionsUsage[region] = {
+			name: rawRegionUsage.name,
+			data: (() => {
+				let temp = (<usage>{});
```
ematipico, line 511: This could be a `Map`. What do you think about having a better name? `temp` is not really a `temp` variable because we actually put data in there and we return it at the end.

ematipico (changes requested):
I left some feedback. Also, github doesn't allow me to comment on the big file as result.

Do you think it's worth save the file as `.ts` or should we consider a `.json` format?

scripts/generated-files/browsers-db.ts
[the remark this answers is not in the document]
jer3m01: `Pick<DataFormat, "data">` would resolve as
```ts
{
    data: Map<string, Feature>
}
``` 
where as `DataFormat["data"]` is
```ts
Map<string, Feature>
```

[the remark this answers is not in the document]
ematipico: We would need to create an interface for that

sebmck (changes requested):
Nice! I added some nits just around the package name and data validation.

I think updating the database and the generation of the files should be separate. Right now in order to generate the files we also need network access. I know you didn't add the script to `generate-all-files`, so it wont be ran in CI. But it'd be good to have this added there too.

We could have a separate `scripts/update-browser-feature-database` that actually does the fetching and writing to disk. And then `scripts/generated-files/browsers-db` can do any generation that actually needs to take place.

internal/browsers-db/README.md
```diff
@@ -0,0 +1,54 @@
+# `browsers-db`
```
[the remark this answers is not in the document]
jer3m01: This package is just the data, there will be another package called `browser-features` that will convert this data into nice implementations. 
As discussed in #1297

jer3m01:
@sebmck 
> We could have a separate `scripts/update-browser-feature-database` that actually does the fetching and writing to disk. And then `scripts/generated-files/browsers-db` can do any generation that actually needs to take place.

Can't really split it as this fetches the data from `Fyrd/caniuse` and strips it down. 
There are some modifications to reduce the size of the data such as using "undefined" (removing it form the json) instead of `"<version>": "n"` and using booleans for features that require a prefix since that's the only thing we care about.

The data is never written as fetched only kept in memory then modified.
If you see a way to split it into different steps tell me.

jer3m01:
@ematipico Huge thanks for the the review, should all be done!

As for generating it as json instead of ts, last I tried `modifyGeneratedFile` didn't work with json files (can't put comments).

jer3m01:
Got it to save as `.rjson` 🚀 

@ematipico @sebmck If everything is ok for you I'll merge it.

ematipico (approved):
Great addition to the codebase! There's an issue with `rome ci`.

sebmck:
Looks good to me! There's some small improvements with the data validation that can be done but I can do them in a follow up PR. Thank you @jer3m01!

PR #1318, opened 2021-02-03
fix(formatter): handle parentheses around nested unary

yeonjuan:
## Summary

Rome formatter remove parentheses around unary in unary..

- target.ts
```ts
+(+foo);
```

```
$ ./rome format target.ts
++foo;
```

It's a pr for handle it.

```
$ ./rome format target.ts
+(+foo);
```

## Test Plan

test cases

### DIFF
--- a/internal/formatter/node/parentheses.ts
+++ b/internal/formatter/node/parentheses.ts
@@ -297,7 +297,7 @@ parens.set(
 	},
 );
 
-function needsParenUnaryExpression(
+function needsParenUnaryLikeExpression(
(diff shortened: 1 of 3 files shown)


ematipico (approved):

sebmck:
Looks good! I think we might also be missing some other conditions. eg. https://github.com/babel/babel/blob/master/packages/babel-generator/src/node/parentheses.js#L217 Might be worth looking into further. Thanks @yeonjuan!

Issue #1315, opened 2021-02-02
📎 File handler for *.txt


jer3m01:
Implement a file handler for *.txt files

sebmck:
What's the intended functionality?

jer3m01:
> What's the intended functionality?

For the moment rome crashes when it encounters a txt file, just something to ignore the file (similar way than fonts).

Don't think we can do much more than ignore the file (ie an empty file handler).

sebmck:
Oh what does the crash look like? I think we throw a pretty poor error when we resolve to a file that's missing a file handler. We can also improve that as a separate task.

jer3m01:
![image](https://user-images.githubusercontent.com/33844379/106920683-a1b18900-670b-11eb-8e87-3c482d136ad8.png)

sebmck:
Just opened #1334 which I think addresses this in the best way. When you run a command like `rome check foo.txt` or import a file we don't have a handler for you'll get the following diagnostic:

<img width="621" alt="Screen Shot 2021-02-04 at 11 51 09 PM" src="https://user-images.githubusercontent.com/853712/106996650-f7575700-6746-11eb-8d70-dce45ada7b29.png">

I also think we should probably remove a bunch of the asset extensions we have hardcoded anyway since you could also specify them in your config.

PR #1334, opened 2021-02-05
Improve max file size and missing file handler errors

### LINKED ISSUE #1328: 📎 Unknown file handler should throw a diagnostic
We throw a hard runtime exception whereas we should throw a diagnostic with the input filename. We should also enforce passing in some other information on why the file needs to be handled in the first place such as "specified via CLI", "imported here" etc.

sebmck:
## Summary

This issue closes #1328 and #1315. It changes two error conditions that we had as runtime exceptions to diagnostics.

Specifying a file that we don't have a handler for now provides a suggestion to add it as an asset extension:

Screen Shot 2021-02-04 at 11 51 09 PM

And trying to perform an operation on a file that exceeds the max file size limit now provides multiple suggestions on how to allow it and why we have that limit in the first place:

Screen Shot 2021-02-05 at 12 01 06 AM

## Test Plan

Ran `./rome ci` and manually ran commands to trigger the above screenshots.

### DIFF
--- a/internal/diagnostics/descriptions/files.ts
+++ b/internal/diagnostics/descriptions/files.ts
@@ -1,0 +1,73 @@
+import {createDiagnosticsCategory} from "./index";
+import {markup} from "@internal/markup";
+import {AbsoluteFilePath, AnyFilePath} from "@internal/path";
+import {DiagnosticAdvice} from "../types";
+
+export const files = createDiagnosticsCategory({
+	NO_FILE_HANDLER: (path: AnyFilePath) => {
+		let advice: DiagnosticAdvice = [];
+
+		if (path.hasAnyExtensions()) {
+			advice.push({
+				type: "action",
+				instruction: markup`You can treat this file extension as a binary asset by running`,
+				noun: markup`Treat this file extension as a binary asset`,
+				command: "config",
+				args: ["push", "files.assetExtensions", path.getDotlessExtensions()],
+			});
+		}
+
+		return {
+			category: "files/missingHandler",
+			message: markup`No file handler found for <emphasis>${path}</emphasis>`,
+			advice,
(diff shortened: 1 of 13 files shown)


internal/diagnostics/descriptions/files.ts
```diff
@@ -0,0 +1,73 @@
...
+					command: "config",
+					args: ["push", "files.maxSizeIgnore", relative],
+				},
+				{
+					type: "action",
+					instruction: markup`Or just increase the size limit for all files to <filesize>${String(
+						size,
+					)}</filesize>`,
+					noun: markup`Increase project max file size limit`,
+					command: "config",
```
ematipico, line 67: The command should be `config set` we we should remove `set` from the array of `args`.

That's how we registered our commands: https://github.com/rome/tools/blob/main/internal/core/server/commands.ts#L109-L110

internal/diagnostics/descriptions/files.ts
```diff
@@ -0,0 +1,73 @@
...
+					instruction: markup`You can ignore this file from linting by running`,
+					noun: markup`Ignore this file from linting`,
+					command: "config",
+					args: ["push", "lint.ignore", relative],
+				},
+				{
+					type: "action",
+					instruction: markup`Or can allow this specific file to exceed the size limit with`,
+					noun: markup`Allow only this specific file to exceed the limit`,
+					command: "config",
```
ematipico, line 58: The command should be `config push` and remove `push` from the array of `args`

internal/diagnostics/descriptions/files.ts
sebmck: Oops, you're right! I'll make a follow up PR and make the type stricter too.

PR #1336, opened 2021-02-05
feat: command auto-config

### LINKED ISSUE #1112: 📎 Add `auto-config` command
We would extract the config generation from `rome init` and move it into a `auto-config` command that can be ran with an already established Rome project. We can then supplement it with additional features such as automatic detection of invalid licenses, and other useful tools for "fixing" a project.

### LINKED ISSUE #1320: 🐛 "Expected end of file" error for file in node_modules/
Rome doesn't appear to ignore node_modules and throws an `✖ Expected end of file` error during `rome init`. This also happens during `rome check` in a project where Rome is already initialised. The file it complains about seems to be valid JSON.

```
 Welcome to Rome! Let's get you started... 

[1/2] Generating .editorconfig

 node_modules/fast-xml-parser/package.json:104:18 parse/spdxLicense ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Expected end of file

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

✖ Found 1 problem
```

## Environment

```
$ rome rage --summary

Environment Variables
  Object {
    COLORTERM: "truecolor"
    LANG: "en_NZ.UTF-8"
    TERM: "xterm-256color"
  }

User Config
  unset

Rome Version
  10.0.4-beta

Node Version
  14.10.0

Platform
  linux x64 5.8.0-41-generic

Terminal Features
  Object {
    background: "unknown"
    colorDepth: 24
    columns: 173
    cursor: true
    hyperlinks: true
    isTTY: true
    unicode: true
  }

Client Flags
  Object {
    clientName: "cli"
    cwd: "/home/joseph/dev/rome-bug-repro"
    realCwd: "/home/joseph/dev/rome-bug-repro"
    silent: false
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 29_323_264
      pid: 163_429
      uptime: 0.880797629
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 29_323_264
        ownedBytes: 0
        ownedFileCount: 0
        pid: 163_429
        uptime: 0.880722256
      }
    ]
  }
```

## Steps To Reproduce

1. Clone [this repo](https://github.com/wildlyinaccurate/rome-bug-repro) (or set up a new project with only `@aws-sdk/client-s3` installed as a dependency)
2. Run `npx rome init`

## Expected Results

Ideally Rome should completely ignore node_modules. Failing that, it should not halt the entire `init` or `check` command.

ematipico:
## Summary

Closes #1112 
Fixes #1320 

This PR makes some refactors around the `init` command and it implements a new command called `auto-config`. Also made some changes and fixes around how the licenses are handled.

- the `init` command has been split in client and server command. The `init` command now asks for the type of configuration, type of indentation and size of indentation; this change is reflected to the generated config file and the `.editorconfig`. The command now doesn't do any lint check. The `init` command is successful only if there's not configuration file present inside the directory; The installation process now support `pnpm` too by checking its lock file;
- the `auto-config` command is split in client and server command. The server command does two different things pages on the flags passed to it. The command cannot be run inside a project that doesn't have a configuration (the diagnostic suggests to run `npx rome init`). The runs lint and inspects all licenses of the projects. All the diagnostics are then passed to the client which will parse them and run the relative `config set`/`config push` commands;

In order to retrieve the diagnostics of the licenses I had to low down the severity of the diagnostics generated by the codec of the manifest, otherwise  I wouldn't been able to collect those diagnostics. @sebmck is there a better way to collect these diagnostics during the `auto-config` while we run a strict check for the rest of the commands? 

I removed the hardcoded licenses we had in the code and we use the config to shut down the errors (`invalidLices`). A nice example of a working configuration! I had to manually add them.  

I also prepared some ground around the config handler, in particular `toml`. It's not ready yet but I put there some tests. At the moment nested objects are not handled at all.

Is there something else that we could add to `auto-config`?

What's missing:
- [x] tests for `init` and tests for `auto-config`
- [x] update documentation 

## Test Plan

Run `./rome ci`


internal/core/server/commands/autoConfig.ts
```diff
@@ -0,0 +1,124 @@
...
+	descriptions,
+} from "@internal/diagnostics";
+import {UnknownObject} from "@internal/typescript-helpers";
+import {RSERValue} from "@internal/codec-binary-serial";
+
+interface Flags extends UnknownObject {
+	checkVSC: boolean;
+}
+
+export type AutoConfig = RSERValue & {
```
sebmck, line 20: Why `RSERValue &` the type of `AutoConfig` will widen.

internal/core/server/commands/autoConfig.ts
```diff
@@ -0,0 +1,124 @@
...
+		const currentProject = await server.projectManager.assertProject(cwd);
+		if (currentProject === undefined) {
+			reporter.error(
+				markup`No Rome project found at <emphasis>${cwd}</emphasis>`,
+			);
+			reporter.info(markup`Run <cmd>rome init</cmd> to boostrap your project.`);
+			return;
+		}
+
+		if (!currentProject.initialized) {
```
sebmck, line 61: This should never happen after an `assertProject` call. `initialized` marks whether we're still crawling the project on disk. When calling `assertProject` we would wait on it if it was initiated by another call.

internal/core/server/commands/config.ts
```diff
@@ -76,9 +76,9 @@ async function runCommand(
 		}
 
 		reporter.success(
-			markup`${action === "push" ? "Adding" : "Setting"} <emphasis>${keyParts}</emphasis> to <emphasis>${JSON.stringify(
+			markup`${action === "push" ? "Adding" : "Setting"} <emphasis>${JSON.stringify(
```
sebmck, line 79: Using `prettyFormat(value)` here instead would be more appropriate I think.

internal/core/server/utils/getManifest.ts
```diff
@@ -0,0 +1,24 @@
+import {AbsoluteFilePath} from "@internal/path";
+import {Manifest, normalizeManifest} from "@internal/codec-js-manifest";
+import {exists, readFileTextMeta} from "@internal/fs";
+import {json} from "@internal/codec-config";
+import {Server} from "@internal/core";
+
+export default async function getManifest(
```
sebmck, line 7: I think you can already use `server.memoryFs.getOwnedManifest(path)` to do what this method does.

internal/core/server/utils/retrieveConfigHandler.ts
```diff
@@ -0,0 +1,27 @@
+import {ConfigHandler, ConfigType} from "@internal/codec-config/types";
+import {json, rjson} from "@internal/codec-config";
+
+export default function retrieveConfigHandler(
```
sebmck, line 4: The `codec-config` package is probably a better place for this.

internal/core/server/utils/updateConfig.ts
```diff
@@ -0,0 +1,28 @@
+import {ConfigCommentMap} from "@internal/codec-config";
+import {writeFile} from "@internal/fs";
+import {ConfigHandler} from "@internal/codec-config/types";
+import {AbsoluteFilePath} from "@internal/path";
+import {ProjectConfig} from "@internal/project";
+import {DeepPartial} from "@internal/typescript-helpers";
+
+interface UpdateConfig {
+	config: Partial<ProjectConfig>;
```
sebmck, line 9: `ProjectConfig` isn't the appropriate type to use when we are dealing with the actual raw project config values as it includes a lot of normalized data. It's actually clearer to just use `RSERObject` or something instead.

sebmck (commented):
Nice! Really excited about this.

> In order to retrieve the diagnostics of the licenses I had to low down the severity of the diagnostics generated by the codec of the manifest, otherwise I wouldn't been able to collect those diagnostics. @sebmck is there a better way to collect these diagnostics during the `auto-config` while we run a strict check for the rest of the commands?

There isn't. There's probably something we can do about lowering the severity of manfiest validation errors for all third party dependencies. That way we could even relax a lot of the `loose` checks, or remove it entirely. I'll have to think about it.

internal/core/server/commands/autoConfig.ts
ematipico: Yeah that old code. Will clean it up

internal/core/server/utils/getManifest.ts
ematipico: I was looking for it everywhere!! Gosh 😅

ematipico:
@sebmck thank you for the review! I updated the PR with the changes you highlighted. Unfortunately integration tests are still turned off so there's not much I can do around it. I tested locally though.

I updated the documentation. I would be great if we could add a screenshot but I don't know how to do it.

sebmck:
🥳

PR #1341, opened 2021-02-07
Add integrations.eslint.enabled project config option

sebmck:
## Summary

This PR adds a new project config option called `integrations.eslint.enabled`:

```
integrations: {
  eslint: {
    enabled: true
  }
}
```

It does not currently do anything, I will implement that functionality in another PR.

The idea will be to use Rome as an ESLint runner. We use our parallellisation and caching, and transparently convert ESLint errors to diagnostics. An ESLint user could use Rome with some simple config and achieve instant benefits.

In the process of this I also cleaned up how we do cache keys and pass project configs to the compiler.

## Test Plan

Ran `./rome ci`

### DIFF
--- a/internal/project/types.ts
+++ b/internal/project/types.ts
@@ -14,7 +14,7 @@ import {
 	createAbsoluteFilePath,
 } from "@internal/path";
 import {Consumer} from "@internal/consume";
-import {RequiredProps} from "@internal/typescript-helpers";
+import {Dict, RequiredProps} from "@internal/typescript-helpers";
 import {SemverRangeNode} from "@internal/codec-semver";
 import {LintRuleName} from "@internal/compiler";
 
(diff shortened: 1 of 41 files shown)


internal/project/types.ts
```diff
@@ -132,13 +145,78 @@ export type ProjectConfigMetaHard = RequiredProps<
...
 
 // Final project config
-export type ProjectConfig = ProjectConfigBase & ProjectConfigObjects;
+export type ProjectConfig = ProjectConfigBase & ProjectConfigObjects & {
+	integrations: ProjectConfigIntegrations;
+};
+
+// The actual type that we allow users to specify their configuration
+// Types are deliberately wider than they need to be to more accurately represent how they will be provided. ie. `string` rather than string literals
+export type RawUserProjectConfig = {
```
sebmck, line 154: cc @ematipico Relevant for my comment on #1112

PR #1344, opened 2021-02-08
feat(css-parser, formatter): custom properties and var function

ematipico:
## Summary

Implements #984 

- `var()` function; it's a different function that probably will get its own lint rules (like, for example, only one fallback property)
- `CSSCustomProperty`

## Test Plan

New tests for formatter and parser.


internal/formatter/test-fixtures/css/function/input.test.md
```diff
@@ -38,6 +39,7 @@
 	padding: calc(calc(1px, 2px), 3px);
 	height: calc(1px, 2px, 3px);
 	background-color: rgb(0 128 255);
+	border: var(--foo, red, blue, calc(10px 1rem));
```
ematipico, line 42: @yeonjuan probably this is a bug, the `+` gets removed.

yeonjuan: @ematipico Yes, It seems a bug in parsing or formatting `CSSNumber`. I'll make a separate pr to handle it :)

internal/css-parser/index.ts
```diff
@@ -965,7 +981,30 @@ export function parseFunction(parser: CSSParser): CSSFunction {
...
+					parser.unexpectedDiagnostic({
+						description: descriptions.CSS_PARSER.INVALID_CUSTOM_PROPERTY,
+					});
+				}
+				i++;
+				params.push(parsedValue);
+			} else {
+				params.push(parsedValue);
+			}
+		}
```
yeonjuan, line 996:
```
if (parsedValue) {
	if (i === 0 && isVarFunction) {
    // ,,,
		i++;
	}
	params.push(parsedValue);
}
```
I think we can remove the `else` statement :)

yeonjuan (approved):
LGTM! I left a minor suggestion :)

Issue #1345, opened 2021-02-08
📎 Port most of the `a11y-jsx` rules under the new `a11y` category rules


ematipico:
Recently, we created a new lint category called `a11y`. This new category is meant to be generic and work for JSX and HTML at the same time.

At the moment, we ported only two rules and we should to the same for most of the `a11y-jsx` we currently have

mainangethe:
Hi @ematipico ,
I'd like to take this up. Though I'm not sure what it means when you say you "ported"

Could you give a more details surrounding what I should do to port a rule, so that I can be able to start.

ematipico:
Hi @mainangethe 

You can use this PR as example https://github.com/rome/tools/pull/1346

At the moment we have a bunch of rules that target some `a11y` requirements. These rules are tailored around JSX AST. Now that we have a working HTML AST parser, what we want to do is to make these rules a bit generic.

So we have to take those rules that are under `jsx-a11y` and move them under the category `a11y`. Not only that: we want to add HTML support, which means that one rule will be valid for JSX and HTML at the same time. 

These are the steps we should follow to move an existing JSX rule:

- create a new rule under the category `a11y`; the same of the rule should be the same of the one you plan to work on;
- copy/paste the existing rule inside the new file (and tests too);
- remove from the code base all the references to the old rule, e.g.: `"jsx-a11y/useOldRule"`, `"lint/jsx-a11y/useOldRule"`;
- extend the rule to handle HTML too;
- add new test cases for a HTML file;

zant:
Hi! I'll be working on `noHeaderScope` :)

imrishabh18:
@ematipico Are some more rules left to be ported than I can work on? I am new to this

ematipico:
> @ematipico Are some more rules left to be ported than I can work on? I am new to this

`useHeadingContent` is available! `useAnchorContent` too. They are both simple to implement.

ematipico:
Two rules missing! 🚀 
- `useValidAnchor`
- `noRedundantRoles`

harxki:
I'll be working on useValidAnchor !

UsamaHameed:
I will take on `noRedundantRoles`

ematipico:
@UsamaHameed that's a though one :) If you have any questions, please ask!

ematipico:
I want to thank you everyone that helped to move all the rules! We finished, now all the rules will work on HTML and JSX! 🎉 🎉

PR #1346, opened 2021-02-08
feat(lint): useValidAlt for HTML and JSX

ematipico:
## Summary

Implements #1345 

Ported the rule `useAltText` under the new category `a11y`.

Created new utilities that mirror the JSX ones and changed a bit the logic. 

## Test Plan

Created new tests, moved old ones inside the new ones and run `./rome ci`


internal/compiler/lint/rules/a11y/useAltText.test.rjson
```diff
@@ -0,0 +1,66 @@
...
+			"<img {...props} alt={undefined} />"
+			"<img src=\"foo\" role=\"presentation\" />"
+			"<img src=\"foo\" role=\"none\" />"
+			"<object {...props} />"
+			"<object aria-label={undefined} />"
+			"<area {...props} />"
+			"<area alt={undefined} />"
+			"<input type=\"image\" {...props} />"
+			"<input type=\"image\" {...props} alt={undefined} />"
+	]
```
ematipico, line 16: @sebmck just noticed a small issue in the rjson formmater

sebmck: Oh that's weird. I think we should get rid of RJSON ASAP so probably not worth bothering fixing.

sebmck (approved):
This is such an awesome example of how cross-language Rome could (and can!) be. Feel free to merge whenever. Might also be easier to batch up updates to multiple lint rules if you're looking at doing others.

PR #1350, opened 2021-02-09
feat: add new package internal/browser-features

jer3m01:
## Summary
`internal/browser-features` #1297

Adds a new internal package for browser implementations from `browsers-db`.

## Todo

- [x] Verify it works correctly
- [x] Add getter by id & abbreviation
- [x] Add usage example


internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,95 @@
+import {Consumer, consumeUnknown} from "@internal/consume";
+import {data, regions} from "@internal/browsers-db";
+
+export interface BrowserProps {
+	id: string,
+	version?: string,
```
ematipico, line 6: What's the case where we don't have a `version`? Why isn't it mandatory?

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,95 @@
...
+	protected getVersionConsumer(): Consumer {
+		return this.getAgentConsumer().get("vs").asImplicitArray().find(value => value.get("v").asString() === this.getVersion())!;
+	}
+
+	public getId(): string {
+		return this.id;
+	}
+
+	public getVersion(): string {
+		return this.version ?? this.getCurrentVersion();
```
ematipico, line 35: What's the difference between `this.version` and `this.getCurrentVersion`?

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,95 @@
...
+		return this.getAgentConsumer().get("vs").asImplicitArray().map(value => value.get("v").asString());
+	}
+
+	/**
+	 * Whether the css feature requires a browser prefix
+	 *
+	 * @param feature check internal/browsers-db/README.md for more info
+	 */
+	public cssFeatureRequiresPrefix(feature: string): boolean {
+		return this.getDataConsumer().get("data").get(feature).get("s").get(this.getId()).get(this.getVersion()).asBoolean(false);
```
ematipico, line 84: It might be worth considering a caching system. Nothing sophisticated. `Map<feature, boolean>`

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	/**
+	 * Get the region usage of the browser
+	 *
+	 * @param region check internal/browsers-db/README.md for more info
+	 */
+	public getRegionUsage(region: string): number|undefined {
```
ematipico, line 92: What's the use case of `getRegionUsage`?

ematipico (commented):
As for now, it seems good to me. I would be nice to see an actual usage/implementation to have a better idea of where we are going with this. I left some question and suggestion

internal/browser-features/Browser.ts
jer3m01: If a version is not specified it will use the current version of the browser. (https://github.com/rome/tools/tree/main/internal/browsers-db#data `cv`)

internal/browser-features/Browser.ts
jer3m01: The version of the current browser implementation instance or the current version of the browser. Same as https://github.com/rome/tools/pull/1350#discussion_r573695098

internal/browser-features/Browser.ts
jer3m01: Great idea, I'll add caching.

internal/browser-features/Browser.ts
jer3m01: Gets the browser usage per region (per version). Used for target browser (https://github.com/browserslist/browserslist#full-list).

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,126 @@
...
+	version?: string;
+}
+
+export type BrowserTypes = "desktop" | "mobile";
+
+export abstract class Browser {
+	private readonly id: string;
+	private readonly version: string | undefined;
+
+	private readonly cssFeatureCache = new Map<string, boolean>();
```
ematipico, line 15: This could be a `WeakMap`. I don't think we plan to use its iterator right?

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,126 @@
...
+
+	protected constructor({id, version}: BrowserProps) {
+		this.id = id;
+		if (version && !this.getVersions().includes(version)) {
+			throw new Error(`Browser "${id}" does not have a version "${version}"`);
```
ematipico, line 20: We probably want to this our diagnostic for this. You can use `createSingleDiagnosticError` and use it like:

```js
import {createSingleDiagnosticError} from "@internal/diagnostics";

throw createSingleDiagnosticError({
	description: descriptions.BROWSERS.UNKNOWN_VERSION
});
```

You will need to create a new diagnostic category

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	protected getAgentConsumer(): Consumer {
+		return this.getDataConsumer().get("agents").get(this.getId());
+	}
+
+	protected getVersionConsumer(): Consumer {
+		return this.getAgentConsumer().get("vs").asImplicitArray().find((value) =>
+			value.get("v").asString() === this.getVersion()
+		)!;
```
ematipico, line 36: Is the `!` necessary? Can't we throw an error in case it's `undefined`? (I get it's due to TS and the `find` method)

internal/browser-features/Browser.ts
```diff
@@ -0,0 +1,126 @@
...
+	 */
+	public getRawReleaseDate(): number | undefined {
+		return this.getVersionConsumer().get("r").asNumberOrVoid()
+			? this.getVersionConsumer().get("r").asNumber() * 1_000
+			: undefined;
+	}
+
+	public getReleaseDate(): Date | undefined {
+		return this.getRawReleaseDate() !== undefined
+			? new Date(this.getRawReleaseDate()!)
```
ematipico, line 89: Is the `!` still necessary after the check?

internal/browser-features/README.md
```diff
@@ -0,0 +1,75 @@
+# `browser-feature`
```
ematipico, line 1: This is an awesome REAMDE.md!! Cheers man! 🎊

internal/browser-features/index.ts
```diff
@@ -0,0 +1,168 @@
...
+	| "and_qq"
+	| "safari"
+	| "ios_saf"
+	| "samsung"
+	| "and_uc";
+
+/**
+ * Key should be the following format: `${BrowserIds}:${version: string}`
+ */
+const browserCache = new Map<string, Browser>();
```
ematipico, line 50: This could a nice occasion to use template type literals:

```ts
type Version = string;
type BrowserKey = `${BrowserIds}:${Version}`;
const browserCache = new Map<BrowserKey, Browser>();
```

IF it's correct 😅

internal/browser-features/index.ts
```diff
@@ -0,0 +1,168 @@
...
+const aliases = loadAliases();
+
+function loadAliases(): Map<string, BrowserIds> {
+	const consumer = consumeUnknown(data, "parse").get("agents");
```
ematipico, line 54: Should we call `consumer`, `agents`?

internal/browser-features/index.ts
```diff
@@ -0,0 +1,168 @@
...
+	| "safari"
+	| "ios_saf"
+	| "samsung"
+	| "and_uc";
+
+/**
+ * Key should be the following format: `${BrowserIds}:${version: string}`
+ */
+const browserCache = new Map<string, Browser>();
+const aliases = loadAliases();
```
ematipico, line 51: I'm not so sure about this... we're doing computation and memory consumption even though we're not using it; is it needed? Can't we can `loadedAliases` when we actually needed?

internal/browser-features/index.ts
```diff
@@ -0,0 +1,168 @@
...
+				break;
+			}
+			default:
+				return undefined;
+		}
+	} catch (error) {
+		if (error.toString().includes('" does not have a version "')) {
+			return undefined;
+		}
+		throw error;
```
ematipico, line 157: We could throw one of our diagnostics here

ematipico (approved):
Amazing work here!

internal/browser-features/Browser.ts
jer3m01: WeakMaps only accept objects as keys.

internal/browser-features/Browser.ts
jer3m01: The version's already been validated before, that can't be undefined.

internal/browser-features/Browser.ts
ematipico: Didn't know that 😳

internal/browser-features/Browser.ts
jer3m01: Typescript doesn't like when I remove it :/

internal/browser-features/index.ts
jer3m01: It's an unknown error, to prevent this catch block from catching everything. Don't really see a diagnostic other than "Unknown error occurred".

internal/browser-features/index.ts
jer3m01: Yeah tried that but Typescript doesn't work well with `string` type in templates. It recognizes it as literally `${string}`.

Argument of type 'string' is not assignable to parameter of type '`android:${string}` | `baidu:${strin...' 😕

internal/browser-features/Browser.ts
jer3m01: It also requires a `location`, how would I go about getting it?

ematipico: `consumer.getDiagnosticLocation()`

jer3m01: As discussed on discord,
>The error isn't really based on the data consumer and the location isn't directly related either, I'll stay with the regular error

sebmck (commented):
This looks great! The `browsers` files are pretty small. Could we move them to the same file as `Browser.ts`? What do you think?

internal/browser-features/Browser.ts
ematipico: I bet if we saved it in a variable, that would work!

internal/browser-features/index.ts
ematipico: Can't say we tried! 😁

PR #1352, opened 2021-02-11
feat(css-parser, formatter): support for the keyframe

ematipico:
## Summary

Part of #984 , it implements parsing and formatting for `@keyframe`.

Few things that we discovered from the spec are:
- `inherit`, `initial` and `unset` are INVALID names for `@keyframe`;
- `@keyframe "initial"` is a valid name, even though some IDE flag it as invalid;
- the name is case-sensitive

I also did a bit of refactor. I moved all the functions inside the `parser/` folder, so now we can divide the `parse*` functions based on their feature; the `index` file was becoming too big.

## Test Plan

Added valid and invalid test cases


sebmck:
Nice! This looks great. The only concern I have is about the AST types. Encoding the token value feels wrong to me. Why do we need to save it? Can we not easily infer it from the parents when formatting?

ematipico:
Sorry @sebmck , what do you mean precisely? Could you point me to the areas that concern you?

sebmck:
`startingTokenValue` on `CSSBlock` and `CSSKeyframe`.

ematipico:
I'll be honest with you: I don't know. It was there already when firstly implemented the CSS parser. Yes, we could easily infer the type from the parent while formatting. Unless it has something to do with some particular case of the official spec, I'd say we can nuke it.

ematipico:
@sebmck I am going to remove the `startingTokenValue` from `CSSKeyframe` but I can't do it from `CSSBlock` because it's part of the official spac: https://www.w3.org/TR/css-syntax-3/#%7b%7d-block-diagram

sebmck:
What's requiring us stick to those interfaces? Generally those sorts of interfaces are only informative and we can encode the information however we want.

ematipico:
Probably nothing. As far as I can see we are using the concept of `CSSBlock` just for instances where we have the `{` token. (for now).

We could create more informative AST nodes instead, in case we need them e.g. `CSSSquareBlock`.

PR #1353, opened 2021-02-11
Add ESLint Integration

sebmck:
This PR includes some new project config options for future integrations, some refactoring of internal types, and a new feature where Rome can be used as an ESLint runner.

### `check.dependencies: boolean` (default: `true`)

Whether we should perform dependency validation when running `rome check`.

### `lint.enabled: boolean` (default: `true`)

This allows you to disable running the Rome linter when using `rome check`. This allows you to enable ESLint integration, disable our linter, and use Rome as a glorified ESLint frontend.

### `integrations.eslint.enabled: boolean` (default: `false`)

Run ESLint when running `rome check`.

## How does the ESLint integration work?

We resolve the `eslint` module relative to your project directory, `require` it, and then use the [ESLint Node.js API](https://eslint.org/docs/developer-guide/nodejs-api) to run ESLint itself. We then convert it's error format into our diagnostic format, and that's it.

We call ESLint per-file, in the same method where we call the Rome compiler to perform a lint. As a result we automatically use our existing parallelisation and caching.

I have added all variants of ESLint config filenames to our cache breakers, so that if you change your ESLint config and run Rome, we will automatically detect and void the cache when checking validity.


sebmck:
Also important to note I've built out generic mechanisms to load untrusted code (unsandboxed). This includes resolving paths, requiring them, wrapping and catching their errors etc.

ematipico:
I have one concern: sometimes people integrate prettier inside ESlint (via `eslint-config-prettier` and `eslint-plugin-prettier`), which means that invalid format is shown as error. Hence, when we disable our lint, do we also disable our format? (to avoid possible clashes)

sebmck:
I think we should support ESLint formatters too. You're right it might be a bit weird to synchronize and could lead to wrong source locations. We can likely detect those cases and include warnings though. Thanks for the reminder, I opened #1359 to track.

ESLint integration is meant to only some incremental migration for ESLint users to Rome so I think we can be okay with it having some strong caveats. Also opened #1360 and #1361 for some other TODOs.

PR #1364, opened 2021-02-12
feat: add new package internal/codec-browsers

jer3m01:
## Summary
`internal/codec-browsers` #1297

Package to parse target browser strings *à la* `firefox >= 84` into browser implementations.

## Todo

- [x] Finish it
- [x] Make README


internal/codec-browsers/parse.ts
```diff
@@ -0,0 +1,609 @@
...
+
+						newTarget = {
+							type: "TargetBrowserRangeOperator",
+							browser,
+							version: (parser.getToken() as BrowserQueryTokens["Number"]).value,
+							operator: parser.getPreviousToken().type as
+								| "GT"
+								| "LT"
+								| "GE"
+								| "LE",
```
ematipico, line 263: Maybe we could create a new type out of them

internal/codec-browsers/parse.ts
```diff
@@ -0,0 +1,609 @@
...
+				if (parser.getToken().type !== "Percentage") {
+					throw parser.unexpected({
+						description: descriptions.BROWSERQUERY.EXPECTED_PERCENTAGE,
+						token: parser.getToken(),
+					});
+				}
+
+				const coverage = (parser.getToken() as BrowserQueryTokens["Percentage"]).value;
+
+				if (parser.lookaheadToken().type !== "In") {
```
ematipico, line 340: We should avoid the usage of `lookaheadToken` as much as possible. The reason is because this function does a tokenization without moving forward with the parsing. Which we could be expensive.

internal/codec-browsers/parse.ts
```diff
@@ -0,0 +1,609 @@
...
+				let dateStr = (parser.getToken() as BrowserQueryTokens["Number"]).value.toString();
+
+				for (let i = 0; i < 2; i++) {
+					if (parser.lookaheadToken().type !== "Hyphen") {
+						dateStr += "-01";
+						continue;
+					}
+
+					// Skip Hyphen
+					parser.nextToken();
```
ematipico, line 386: What would happen if we don't have `Hyphen`? Should we have use `expectToken` instead? It matches a token and if does, it moves forward. If it doesn't match, it throws an error.

internal/codec-browsers/parse.ts
```diff
@@ -0,0 +1,609 @@
...
+			}
+
+			return parser.finishValueToken("String", value, index.increment());
+		}
+
+		return parser.finishValueToken("Invalid", char);
+	},
+});
+
+export function parseBrowserQuery(options: ParserOptions): AnyTargetBrowser[] {
```
ematipico, line 224: This file has become really long. What do you think if we broke down various parts of  this function in small functions? e.g. `parseCover`, `parseLast`, etc.

If we use this approach, eventually we will be able to avoid the mutation `newTarget =` and return directly the object returned by these functions.

internal/codec-browsers/resolve.ts
```diff
@@ -0,0 +1,376 @@
...
+	browser: string;
+	version: number;
+	to: number;
+}
+
+interface TargetBrowserRangeOperator {
+	readonly type: "TargetBrowserRangeOperator";
+	browser: string;
+	version: number;
+	operator: "GT" | "LT" | "GE" | "LE";
```
ematipico, line 46: Yeah we probably should make a new type out of this chars. We are repeating them inside a bunch of places. We could call it `Operator`.

```ts
export type Operator = "GT" | "LT" | "GE" | "LE";
```

internal/diagnostics/descriptions/parsers/browserquery.ts
```diff
@@ -0,0 +1,46 @@
+import {createDiagnosticsCategory} from "@internal/diagnostics";
```
ematipico, line 1: We should add a full stop to all the sentences.

ematipico (commented):
Amazing work here! Truly outstanding! Especially all the test fixtures! 

I left some suggestions.

Overall I would avoid things like

```js
parser.nextToken();
parser.nextToken();
```

Because it makes the parser a little bit too loose (and we don't want that inside a non-recoverable parser). I would use methods such as `eatToken` or `expectToken`. With these methods we are more explicit and help the developer and the parser to understand what tokens should be there.

If possible, I would break down some portions of code into small functions.

ematipico (approved):
Looks good! Great work again :)

internal/browser-features/Browser.ts
```diff
@@ -122,7 +126,7 @@ export abstract class Browser {
 	public getRegionUsage(region: string): number | undefined {
 		return consumeUnknown(regions, DIAGNOSTIC_CATEGORIES.parse).get(region).get(
 			"data",
-		).get(this.getId()).get(this.getVersion()).asNumberOrVoid();
+		).get(this.getId()).get(this.getVersion().toString()).asNumberOrVoid();
```
sebmck, line 129: In #1384 I just added a new method to `Consumer` called `getPath` so in the future this could be simplified to:

```
return consumeUnknown(regions, DIAGNOSTIC_CATEGORIES.parse).getPath([region, "data", this.getId(), this.getVersion().toString()]).asNumberOrVoid();
```

internal/browser-features/index.ts
```diff
@@ -173,3 +172,47 @@ export function getBrowser(
...
+			const browser = getBrowser({name, version})!;
+			usages.push({
+				id: browser.getId(),
+				version,
+				usage: region
+					? browser.getRegionUsage(region) ?? 0
+					: browser.getGlobalUsage(),
+			});
+		});
+	});
```
sebmck, line 214: No reason not to use `for of` here instead of `forEach`.

```suggestion
	for (const name of getAllBrowserNames()) {
		for (const version of getBrowser({name})!.getVersions()) {
			const browser = getBrowser({name, version})!;
			usages.push({
				id: browser.getId(),
				version,
				usage: region
					? browser.getRegionUsage(region) ?? 0
					: browser.getGlobalUsage(),
			});
		}
	}
```

internal/browser-features/index.ts
```diff
@@ -173,3 +172,47 @@ export function getBrowser(
...
+
+export function getAllBrowserUsages(region?: string): AllBrowserUsage[] {
+	if (allBrowserUsagesCache) {
+		return allBrowserUsagesCache;
+	}
+
+	const usages: AllBrowserUsage[] = [];
+
+	getAllBrowserNames().forEach((name) => {
+		getBrowser({name})!.getVersions().forEach((version) => {
```
sebmck, line 204: I notice there's a lot of null assertions here around using the `getBrowser` method. Could we add another method called `getBrowserAssert` that always returns the type and throws if the browser doesn't exist?

We lose out on a lot of type safety and I believe the only places we currently use null assertions like this in Rome are around array methods (like `pop()`) that we're sure are being done on a populated array.

Alternatively it could be renamed to `getOptionalBrowser` and the assertion method could be called `getBrowser`. The `Assert` suffix is used throughout the codebase to refer to this kind of behaviour but both conventions are used so it's up to you.

internal/codec-browsers/README.md
```diff
@@ -0,0 +1,87 @@
+# `codec-browsers`
+
+Used to parse browser target syntax.
+Thanks to [browserslist ](https://github.com/browserslist/browserslist) for the idea and syntax.
```
sebmck, line 4:
```suggestion
Thanks to [browserslist](https://github.com/browserslist/browserslist) for the idea and syntax.
```

internal/diagnostics/descriptions/parsers/browserquery.ts
sebmck: I don't think there should be full stops. We don't currently have good conventions on writing diagnostic message text, but we only use full stops when there's more than one sentence, and that's rare as I try to break up additional context into additional `advice`.

sebmck (changes requested):
This is awesome, thank you @jer3m01! Just have some small nits around the usage of `forEach` and null assertions. I was worried that my changes in #1384 would conflict, but there doesn't appear to be much overlap. I can rebase this for you once it's ready to merge if my PR lands before this.

internal/browser-features/Browser.ts
jer3m01: Awesome! I'll change that once you merge #1384.

internal/browser-features/index.ts
jer3m01: I'll just throw an error on the base function as I can't see anywhere I would use it without checking if its undefined.

sebmck:
Awesome, thank you @jer3m01! This is a great addition 💃🏻

PR #1370, opened 2021-02-17
feat(css-parser, formatter): support calc()

ematipico:
## Summary

Part of #984 , it implement parsing and formatting of `calc()`.

I thought that this function was easy, but it turns out that it has a bit of logic and some strict syntax, which is not caught by all editors/IDEs/parser.

Source: https://www.w3.org/TR/css-values-3/#calc-syntax

At the moment I am having an error while launching the tests, I will update the snapshots tests later if I can.

- [x] update snapshot tests

## Test Plan

New tests


ematipico:
Uh, this is really relevant! The `calc-` made me think that the nodes were relevant to the `calc()` function but I was totally wrong! Thank you for sharing this!

PR #1371, opened 2021-02-19
docs(markup): add README

jer3m01:
## Summary
Adds a README for all markup tags, attributes and examples.

### DIFF
--- a/internal/markup/README.md
+++ b/internal/markup/README.md
@@ -173,10 +173,9 @@ column | number
 line | number
 
 ```ts
-markup`Error is <filelink target="test.ts" line="1" column="10">here</filelink>`
+markup`Error at <filelink target="test.ts" line="6" column="10"/>`
 ```
-
-*Doesn't seem to do anything?*
+![filelink](https://user-images.githubusercontent.com/33844379/108555619-ffda9080-72f5-11eb-8811-8b545c30a130.png)
 
 ### Inverse
 ```ts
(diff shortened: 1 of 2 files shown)


internal/markup/README.md
```diff
@@ -0,0 +1,369 @@
...
+--- | ---
+target | string
+column | number
+line | number
+
+```ts
+markup`Error is <filelink target="test.ts" line="1" column="10">here</filelink>`
+```
+
+*Doesn't seem to do anything?*
```
sebmck, line 179: Do you have a terminal emulator that supports hyperlinks https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda? `line` and `column` is used to automatically generate the inner text. Omit `here` to see it in action.

internal/markup/README.md
```diff
@@ -0,0 +1,369 @@
...
+### Success
+```ts
+markup`The <success>success</success> tag`
+```
+![success](https://user-images.githubusercontent.com/33844379/107899969-196b8900-6f40-11eb-86c0-cd845279f210.png)
+
+### Locator
+Attribute | Type
+--- | ---
+id | string
```
sebmck, line 258: This isn't fully implemented but it's a way of receiving the rendered location of something. For example, if we received a random JS value and we wanted to print a diagnostic, we could use `pretty-format` on it, pass in the key we want to reference, render the markup as `sourceText` for the `DiagnosticLocation`, and set the `line` and `column`.

internal/markup/README.md
```diff
@@ -0,0 +1,369 @@
...
+```ts
+markup`<token type="boolean">true</token> <token type="keyword">function</token> <token type="number">465</token>
+<token type="regex">/[abc]+(def).*/</token> <token type="string">hello world</token> <token type="comment">im a comment</token>
+<token type="operator">+</token> <token type="punctuation">,</token> <token type="variable">foo</token>
+<token type="attr-name">data</token> <token type="function">markup</token> <token type="attr-value">value</token>
+<token type="attr-equals">yes</token> <token type="tag">tag</token>`
+```
+![token](https://user-images.githubusercontent.com/33844379/108448934-e76d6600-7262-11eb-8305-f4d8544cb5f2.png)
+
+*Doesn't seem to do anything?*
```
sebmck, line 241: It's used for syntax highlighting. I'm not sure why you don't see it doing anything. The `syntax-markup-highlight` package generates markup. The `type` here is loosely [PrismJS](https://prismjs.com/) class names to make it easy to render to HTML (which we currently use in the docs for previews).

jer3m01:
Think this is good enough for now, and thanks for the review!

PR #1379, opened 2021-02-28
feat(css-parser, formatter): support qualified name

yeonjuan:
## Summary

Implement parsing and formatting [qualified names](https://www.w3.org/TR/css-namespaces-3/#css-qnames).

```
toto|A
|B
*|C
D
```
A qualified name can be used in type selector and attribute selector so I changed the ast of CSSTypeSelector, CSSUniversalSelector, and CSSAttributeSelector also.

## Test Plan

test cases


ematipico (commented):
There are lot of changes around existing fixtures. Is it expected?

internal/css-parser/parser/selectors.ts
```diff
@@ -16,25 +17,97 @@ import {
...
-			value: token.value,
-		},
+function matchDelim(parser: CSSParser, value: string) {
+	return (
+		matchToken(parser, "Delim") &&
+		(parser.getToken() as Tokens["Delim"]).value === value
 	);
 }
 
+function parseNamespacePrefix(parser: CSSParser) {
```
ematipico, line 33: return type is missing

internal/css-parser/parser/selectors.ts
```diff
@@ -16,25 +17,97 @@ import {
...
+	if (matchDelim(parser, "|")) {
+		nextToken(parser);
+		return "";
+	}
+
+	const lookahead = parser.lookaheadToken();
```
ematipico, line 39: Is there a way to achieve the same result without the usage of `lookaheadToken`? It's an expensive operation and its usage is discouraged. It basically runs a tokenization without moving forward.

yeonjuan: @ematipico Yes maybe:) , I'll do it this week.

yeonjuan:
@ematipico  I close it and will make another PR because it's quite late compared to the latest. Thanks for the review.

PR #1381, opened 2021-02-28
fix(markdown-parser): handle text with single newline as same paragraph

Monchi:
## Summary

Currently, text with a single newline is handled as separated paragraphs. This PR makes them handled as same paragraph with `\n` text.

## Test Plan

**input**
```markdown
first line
second line
```

**main branch**
```
MarkdownRoot {
  comments: Array []
  corrupt: false
  diagnostics: Array []
  integrity: Object {hash: "73621482ff083eca9ea88880393298f7d3f53402200780b0c16354a9beb0535a"}
  path: UIDPath<project-rome/test.md>
  body: Array [
    MarkdownParagraph {children: Array [MarkdownText {value: "first line"}]}
    MarkdownParagraph {children: Array [MarkdownText {value: "second line"}]}
  ]
}
```

**This PR**
```
MarkdownRoot {
  comments: Array []
  corrupt: false
  diagnostics: Array []
  integrity: Object {hash: "73621482ff083eca9ea88880393298f7d3f53402200780b0c16354a9beb0535a"}
  path: UIDPath<project-rome/test.md>
  body: Array [
    MarkdownParagraph {
      children: Array [
        MarkdownText {value: "first line"}
        MarkdownText {value: "\n"}
        MarkdownText {value: "second line"}
      ]
    }
  ]
}
```

### DIFF
--- a/internal/markdown-parser/index.ts
+++ b/internal/markdown-parser/index.ts
@@ -70,40 +70,6 @@ const markdownParser = createParser<MarkdownParserTypes>({
 			if (char === "#") {
 				return [state, consumeHeading(parser, index)];
 			}
-			if (char === "\n") {
-				const nextChar = parser.getInputCharOnly(index.increment());
-				if (nextChar === "#") {
-					return [state, consumeHeading(parser, index.add(1))];
-				}
-
-				if (nextChar === "`") {
-					const token = consumeCode(parser, index.add(1));
-					if (token) {
-						return [state, token];
-					}
-				}
-
-				if (isDigit(nextChar)) {
-					const token = tokenizeListItem(parser, index);
-					if (token) {
-						return [
-							{
-								isParagraph: true,
-								isListItem: true,
-							},
-							token,
-						];
-					}
-				}
-
-				return [
-					{
-						isParagraph: false,
-						isListItem: false,
-					},
-					parser.finishToken("NewLine"),
-				];
-			}
 
 			if (char === "-") {
 				const block = tokenizeBlock(parser, "-", index, char);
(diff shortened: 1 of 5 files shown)


internal/markdown-parser/parser/paragraph.ts
```diff
@@ -87,3 +92,13 @@ export function parseParagraph(
 		},
 	);
 }
+
+function isBreakingNewLine(parser: MarkdownParser) {
+	const nextToken = parser.lookaheadToken();
```
ematipico, line 97: Is there a way to achieve the same result without the usage of `.lookaheadToken`? This function is expensive and its usage is discouraged. We are basically doing a tokenization without moving forward. 

If it can help, the parser can give you the **previous token** for free. (`parser.getPreviousToken()`)

internal/markdown-parser/index.ts
```diff
@@ -70,40 +70,6 @@ const markdownParser = createParser<MarkdownParserTypes>({
 			if (char === "#") {
 				return [state, consumeHeading(parser, index)];
 			}
-			if (char === "\n") {
```
ematipico, line 73: Good catch! Totally missed this old code, thank you!

internal/markdown-parser/parser/paragraph.ts
Monchi: Ah I see, I just now read the implementation of `lookahead` :innocent:

Updates to call `nextToken` once! 42b342a

ematipico: The function is still there. We should remove it and find another way to achieve the same result without this function

Monchi: Sorry I was mistaken that `lookahead` is removed. I've completely removed it. 45b243b1f0858cdfbe701212c31db1db51a561c1

sebmck:
Going to delegate approval to @ematipico who is responsible for our Markdown parser!

ematipico (approved):
Great work! Thank you!

PR #1011, opened 2020-08-09
Feature/commit validation

ematipico:
## Summary

Closes #1009

Did some refactor and moved some functions inside the `vcs` folder, they seemed more appropriate. 
I didn't add the script inside the CI because I don't actually know how to compare branches that belong to different forks. 

One flaky part of the current implementation, it's that the script assumes the `main` branch (local one) is up to date with the upstream. If you have suggestions on how to improve the script, it would be great.

## Test Plan

CI.

PR #1384, opened 2021-03-02
Ch-ch-ch-ch-changes

sebmck:
[Turn and face the strange
Ch-ch-changes](https://www.youtube.com/watch?v=pl3vxEudif8)

This PR is a spiral of work I started when trying to build a new website. I first added a new command called `rome develop` that allows you to spin up a local web server that watches a directory for file changes, automatically bundles, and serves static files. This would allow us to self-host our own website generator.

I'm going to include the actual website in a separate PR as making this one larger than it already is would be daft.

In the future I will try to be more responsible with the how these changes are made and submitted.

### `ReporterStream` simplification

I previously implemented some magic thing where we'd keep track of the lines being written to a stream, and compensated for it when we had a progress bar to keep it at the right position.

This turned out to be not so reliable, and rather than rewrite it which would be a lot of work, I just simplified it. Progress bars will always be written to the bottom of the screen and should no longer have ghosting.

### fs

The `fs` package has been gutted, and all of it's methods moved to `AbsoluteFilePath`.

**Before**

```
await readFile(path);
```

**After**

```
await path.readFile();
```

This allows us to avoid having to import the entire `fs` package when we already have a path instance. Our `fs` methods only ever supported being passed in absolute file paths instances anyway.

### Paths

I've made our internal representation much more explicit. We now store the absolute parts of a path in specific fields. Before we encoded it in the same array where we stored relative segments, and did a lot of work to avoid messing with the beginning of the array and fetching the correct element.

We would also encode paths as strings and then reparse them when being passed over a bridge, but I've changed that to use the parsed forms instead which is faster, and in some cases more memory efficient than the original. 

### Cross-realm safety

We had a bunch of code that used `instanceof` checks in order to pretty format or encode values. However when we have multiple realms this isn't safe. ie. Is something a `Map`? What if the value came from a test?

I've switched these checks to use Node `util.types` methods that use custom host APIs to check for equivalency as it's not something we can replicate in userland.

### Bridge message format

Previously we encoded messages using an object, but I've switched it to arrays and ints instead of strings for event references. This makes the format a bit more obtuse to understand when reading it over the wire, but that's a worthwhile trade-off given the performance and size advantages.

### Standard APIs

I've reduced our dependency on Node-specific APIs such as `Buffer` and other binary methods, in favor of TypedArrays and TextEncoder/TextDecoder.

### `pretty-format` line wrapping

We will now attempt to place properties when printing an object or array on the same line until it breaks the viewport and then we place each element on individual lines. Need to come up with some heuristics though to determine the predicted complexity of an objects children and always printing them on individual lines as you can get some pretty gnarly results.


internal/cli-diagnostics/displayDiagnostic.ts
```diff
@@ -0,0 +1,288 @@
...
+	const {start, end, path} = location;
+
+	const mainAdvice = filterMainAdvice(diagnostic);
+
+	// Determine if we should skip showing the frame at the top of the diagnostic output
+	// We check if there are any frame advice entries that match us exactly, this is
+	// useful for simplifying stacktraces
+	let skipFrame = false;
+	if (start !== undefined && end !== undefined) {
+		adviceLoop: for (const item of mainAdvice) {
```
ematipico, line 85: I believe this is the first time in my life I see the label statement used inside a codebase that wasn't MDN 🤣

internal/core/server/develop/DevelopRequest.ts
```diff
@@ -0,0 +1,258 @@
...
+			}
+		}
+
+		// Fetch requested pathname
+		const pathnameStr = request.url;
+		if (typeof pathnameStr !== "string" || pathnameStr[0] !== "/") {
+			throw new Error("Malformed req.url");
+		}
+
+		this.url = createURLPath(`http://${host}${pathnameStr}`);
```
ematipico, line 50: As a note, we should be able to support possible locale `https` environments. Edge case but some users might need it.

ematipico (commented):
PR too long to go through all the changes! 😓 

I left some comment when I could. Could you remove the hardcoded line inside the ESLint integration? Line 22.

ematipico:
> In the future I will try to be more responsible with the how these changes are made and submitted.

That's a promise 😝

jer3m01:
> 
> 
> > In the future I will try to be more responsible with the how these changes are made and submitted.
> 
> That's a promise 😝

We might even get a Conventional Commits message one day 😄

ematipico:
> > 
> > 
> > > In the future I will try to be more responsible with the how these changes are made and submitted.
> > 
> > That's a promise 😝
> 
> We might even get a Conventional Commits message one day 😄 

On that note, we actually have something already! But it's not in place yet.

We have a commit parser: https://www.github.com/rome/tools/tree/main/internal%2Fcommit-parser

The parsing done by following conventional commit.

And there's a PR of mine of months ago where there's a validation for the commit: https://github.com/rome/tools/pull/1011

We should probably resume that!

PR #1385, opened 2021-03-04
fix(markdown-parser): fix markdown text ending position

Monchi:
## Summary

Currently, `MarkdownText`'s ending position is the same as starting position. This PR fixes it.

## Test Plan

See snapshot

### DIFF
--- a/internal/markdown-parser/parser/inline.ts
+++ b/internal/markdown-parser/parser/inline.ts
@@ -95,8 +95,10 @@ export function parseInline(
 			},
 		);
 	}
+	const pos = parser.getPosition();
+	parser.nextToken();
 	return parser.finishNode(
-		parser.getPosition(),
+		pos,
 		{
 			type: "MarkdownText",
 			value: token.value,
--- a/internal/markdown-parser/parser/reference.ts
+++ b/internal/markdown-parser/parser/reference.ts
@@ -55,11 +55,6 @@ export function parseReference(
 			);
 			if (inline) {
 				reference.push(inline);
-				// TODO: call nextToken in parseInline()
-				// console.log(inline.type)
-				// if (inline.type === "MarkdownText") {
-				// 	parser.nextToken();
-				// }
 			}
 		}
 	}
--- a/internal/markdown-parser/parser/reference.ts
+++ b/internal/markdown-parser/parser/reference.ts
@@ -10,9 +10,11 @@ import {parseText} from "@internal/markdown-parser/parser/text";
 
 export function parseReference(
 	parser: MarkdownParser,
-): MarkdownReferenceInline | (AnyMarkdownInlineNode[]) {
+): MarkdownReferenceInline | AnyMarkdownInlineNode[] {
+	parser.expectToken("OpenSquareBracket");
+
 	const pos = parser.getPosition();
-	let reference: MarkdownReference | (AnyMarkdownInlineNode[]) = [];
+	let reference: MarkdownReference | AnyMarkdownInlineNode[] = [];
 	let unwantedTokens = false;
 
 	while (!parser.matchToken("EOF")) {
@@ -53,9 +55,13 @@ export function parseReference(
 			);
 			if (inline) {
 				reference.push(inline);
+				// TODO: call nextToken in parseInline()
+				// console.log(inline.type)
+				// if (inline.type === "MarkdownText") {
+				// 	parser.nextToken();
+				// }
 			}
 		}
-		parser.nextToken();
 	}
 
 	return [
--- a/internal/markdown-parser/test-fixtures/lists/input.test.md
+++ b/internal/markdown-parser/test-fixtures/lists/input.test.md
@@ -18,7 +18,7 @@ MarkdownRoot {
 							children: [
 								MarkdownText {
 									value: "foo"
-									loc: SourceLocation lists/input.md 1:2-1:2
+									loc: SourceLocation lists/input.md 1:2-1:5
 								}
 							]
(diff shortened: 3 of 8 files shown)


internal/markdown-parser/parser/inline.ts
```diff
@@ -95,8 +95,13 @@ export function parseInline(
...
 	}
-	return parser.finishNode(
-		parser.getPosition(),
+	const pos = parser.getPosition();
+	return parser.finishNodeAt(
+		pos,
+		{
+			line: pos.line,
+			column: pos.column.add(token.value.length),
+		},
```
ematipico, line 104: Good catch! Although this solution seems strange.

Could you try a different approach? Adding `parser.nextToken()` before `parser.finishNode`

Monchi: Yes, this is a little hacky. Adding `parser.nextToken()` means following parsing needs some change. The affected code would be a bit large but I gonna try.

ematipico: Yeah that's my feeling too. I'll be honest, the markdown parser was my first parser here so I didn't have enough knowledge of how `ParserCore` worked.

I wouldn't mind seeing a bigger change in order to correctly align all the functions and how the nodes are created. 

As rule of thumb, we should always have

```js
parser.nextToken();
return parser.finishNode(start, {})
```

internal/markdown-parser/test-fixtures/lists/input.test.md
```diff
@@ -206,11 +206,11 @@ MarkdownRoot {
 								MarkdownReferenceInline {
 									value: ""
 									reference: Array []
-									loc: SourceLocation lists/input.md 11:2-11:4
+									loc: SourceLocation lists/input.md 11:3-11:4
 								}
```
Monchi, line 210: it's weird that `MarkdownReferenceInline` is here, but not in the scope of this PR.

internal/markdown-parser/test-fixtures/smoke/input.test.md
```diff
@@ -300,27 +300,28 @@ MarkdownRoot {
...
+							loc: SourceLocation smoke/input.md 30:16-30:43
+						}
+					]
+					loc: SourceLocation smoke/input.md 30:14-30:45
 				}
 				MarkdownText {
 					value: "__"
-					loc: SourceLocation smoke/input.md 30:45-30:45
+					loc: SourceLocation smoke/input.md 30:45-30:47
 				}
```
Monchi, line 321: This AST is also weird but not in the scope for now.

internal/markdown-parser/test-fixtures/lists/input.test.md
```diff
@@ -187,11 +187,11 @@ MarkdownRoot {
 								MarkdownReferenceInline {
 									value: ""
 									reference: []
-									loc: SourceLocation lists/input.md 11:2-11:4
+									loc: SourceLocation lists/input.md 11:3-11:4
 								}
```
Monchi, line 191: it's weird that `MarkdownReferenceInline` is here, but not in the scope of this PR.

internal/markdown-parser/test-fixtures/smoke/input.test.md
```diff
@@ -285,27 +285,28 @@ MarkdownRoot {
...
```
Monchi, line 306: This AST is also weird but not in the scope for now.

internal/markdown-parser/parser/inline.ts
Monchi: @ematipico I've finished the reimplementation. The change isn't big as I expected :)

internal/markdown-parser/test-fixtures/lists/input.test.md
ematipico: internal/markdown-parser/parser/reference.ts

`expectToken` move forward. That's why you see the change

Monchi: Yes, but it's not important because `MarkdownReferenceInline` shouldn't be here. This is checkbox of task list item.

internal/markdown-parser/parser/reference.ts
```diff
@@ -10,9 +10,11 @@ import {parseText} from "@internal/markdown-parser/parser/text";
 
 export function parseReference(
 	parser: MarkdownParser,
-): MarkdownReferenceInline | (AnyMarkdownInlineNode[]) {
+): MarkdownReferenceInline | AnyMarkdownInlineNode[] {
+	parser.expectToken("OpenSquareBracket");
```
ematipico, line 14: We can't use `expectToken` because it throws a hard error, which is not we want. Especially in a parser as loose as this one.

We should use `matchToken` and `return` in case we don't find the token we're supposed to.

internal/markdown-parser/test-fixtures/smoke/input.test.md
ematipico: Ok, no problem. We can't tackle it later with another PR.

ematipico (approved):
It looks good now and it's much better! Thank you for looking into this, it was a big oversight of mine and I apologize. The only change I asked is to not use `expectToken`. We should not use them inside recoverable parsers.

Especially inside the markdown, where we  don't have rules at all (if not few).

Monchi:
@ematipico I think this `expectToken` won't break the recoverability. 
We validate that the current token is `OpenSquareBracket` before calling `parseReference`. So if `expectToken` failed, it means there is an implementation bug but not means the given markdown is incorrect.

Issue #1324, opened 2021-02-04
📎 Propose new governance structure


sebmck:
Our governance model needs more structure. The concept of stewards doesn't make a lot of sense because with a future service to fund Rome it introduces some corporate structure. There needs to be some more delegation and assignment of roles. I have a lot of thoughts but will save it for a more concrete PR. This issue servers as a reminder and notice that I'm thinking about it.

PR #1391, opened 2021-03-08
feat(releases): Add automatic nightly releases

sebmck:
## Summary

This PR adds a GitHub action on a cron job that triggers a workflow depending on the `nightly` environment that requires approval. I'll manually approve it whenever it's triggered to remove the possibility for someone committing something dangerous.

## Test Plan

Manually testing triggering the workflow once landed.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,25 @@
+name: Nightly Release
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  release:
+    environment: nightly
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14.16.0'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Build Release
+        run: ./rome run scripts/build-nightly
+      # Publish to npm
+      - name: Publish to npm
+        run: npm publish --access public
+        working-directory: dist
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # TODO: Publish to GitHub?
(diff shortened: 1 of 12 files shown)


sebmck:
Will follow up tomorrow with some changes to automatically create a git tag and post to a dedicated Discord channel.

ematipico:
> I'll manually approve it whenever it's triggered to remove the possibility for someone committing something dangerous.

I think we should write this down inside the governance file. In case people asks "where's today's nightly?", we have a section of the project where we explain how to process works.

jer3m01:
> > I'll manually approve it whenever it's triggered to remove the possibility for someone committing something dangerous.
> 
> I think we should write this down inside the governance file. In case people asks "where's today's nightly?", we have a section of the project where we explain how to process works.

We could also allow Core Contributors to approve it (if @sebmck's ok with it).

sebmck:
@ematipico 

> I think we should write this down inside the governance file. 

Yeah, the governance document should describe the release process. I'll update it when I work on the next piece of release infra.

@jer3m01 

> We could also allow Core Contributors to approve it

With expanded access, any core contributor can be a target for a malicious attack. If you have access to a contributor account, and given that anyone can merge their own code, you can commit right before a nightly release is cut, approve it, and then it's out there. If that were to happen it's not possible for us to recover that trust.

Given that "no dependencies" is extremely uncommon, this is a power only we have and can be a compelling adoption argument for security-sensitive organizations that want more vetted tooling.

I want to reuse some of this for official releases, but only trigger on changes to `package.json` `version`. We could have a different set of reviewers for official releases and nightlies, but that feels like saying nightly releases are less safe than stable.

sebmck:
Also slightly what I alluded to in #1324 (Propose new governance structure) is having more team roles, and I can imagine there being a special one for releases.

PR #1393, opened 2021-03-08
feat: add tokenize option for parse command

Monchi:
## Summary

**This is PoC implementation before any discussions. Any ideas or discussions are welcome.**

This PR adds `rome parse --tokenize` option to tokenize file content and print the tokens. This would be useful in parser development.

## Test Plan

test.html
```html
<div>test</div>
```

Then run
```bash
$ ./rome parse test.html --tokenize
[
  TagStartOpen {
    end: ZeroIndexedNumber<1>
    start: ZeroIndexedNumber<0>
  }
  Identifier {
    value: "div"
    end: ZeroIndexedNumber<4>
    start: ZeroIndexedNumber<1>
  }
  TagEnd {
    end: ZeroIndexedNumber<5>
    start: ZeroIndexedNumber<4>
  }
  Text {
    value: "test"
    end: ZeroIndexedNumber<9>
    start: ZeroIndexedNumber<5>
  }
  TagEndOpen {
    end: ZeroIndexedNumber<11>
    start: ZeroIndexedNumber<9>
  }
  Identifier {
    value: "div"
    end: ZeroIndexedNumber<14>
    start: ZeroIndexedNumber<11>
  }
  TagEnd {
    end: ZeroIndexedNumber<15>
    start: ZeroIndexedNumber<14>
  }
  EOF {
    end: ZeroIndexedNumber<16>
    start: ZeroIndexedNumber<16>
  }
]
```


ematipico:
I really like this idea, this would really help the development of new features inside the parsers!

Although, I have mixed feelings about having this feature as command. Maybe we could add a new argument to the `parse` command? Something like:

```bash
rome parse ./test.css --only-tokens
```

Monchi:
@ematipico Thanks!

> Maybe we could add a new argument to the parse command

That's one of the choices! Actually I don't know the philosophy of Rome's command architecture and I don't have a strong opinion. I wanna ask @sebmck's thought.

sebmck:
Yeah, I think a `--tokenize` flag on the `parse` command would be preferred. I like how you've threaded it through alongside the other worker request types! Sadly there's a lot of boilerplate involved but happy you figured it out, not sure how to reduce it.

Have a few other general notes:

- Use `TokenBase[]` rather than `TokenValues<BaseTokens>[]`. We don't need a type anymore specific than that.
- `Worker#tokenize` is quite complicated and looks to be a copy+paste of the code from `parse`. We should extract out the relevant parts into helper methods to avoid the repetition. I understand that might be quite complex and intuitive given the dependencies though. If you don't feel comfortable doing that then I can do it after merging.

If you end up adding a flag to `parse`, continue to use the other worker methods you've created, rather than adding it to the `parse` worker methods/events.

Monchi:
@sebmck Thanks! I've replaced `rome tokenize` with `rome parse --tokenize`. cc: @ematipico 

> Use `TokenBase[]` rather than `TokenValues<BaseTokens>[]`. We don't need a type anymore specific than that.

Works well. I've added support for JavaScript.

> `Worker#tokenize` is quite complicated and looks to be a copy+paste of the code from parse.

Yes, I implemented it naively because it was PoC :) I've refactored that extracts the logic of generating `ExtensionParseInfo`.

sebmck:
Woah, this is awesome! Thank you @Monchi!

PR #1398, opened 2021-03-11
feat(lint): noPositiveTabindex for HTML and JSX

dctalbot:
## Summary

Ports the "no positive tab index" rule from `/jsx-a11y` to `/a11y` and enables linting html with said rule

Part of #1345

## Test Plan
- I removed references to `/jsx-a11y`
- created a test for html
- added a diagnostic handler for html
- ran `./rome ci`


ematipico:
We follow these rules: https://dequeuniversity.com/rules/axe/3.5/tabindex

What's the use case of using a positive tabindex?

ematipico (changes requested):
Documentation is the last part of this PR. The rest looks good to me! Inside `website/src/docs/` there's a file called with the same name of the rule. We should move it inside the correct folder (`a11y`) and change all the occurrences of `jsx-a11y` to `a11y` inside that file.

ematipico:
I missed it! Sorry 😔

BTW, CI is failing. You should run the script to rigenerante all files: https://github.com/rome/tools/blob/main/CONTRIBUTING.md#generated-files

ematipico (approved):
Thank you!

PR #1405, opened 2021-03-16
feat(lint): noAutofocus for HTML and JSX

harxki:
## Summary
Ported the rule noAutoFocus to the new category a11y
part of #1345 

- Migrated the /jsx-a11y/noAutoFocus to a11y/noAutoFocus
- Added HTML checks

## Test Plan
- Moved old tests from jsx-a11y/
- Added HTML tests
- run ./rome ci and passed


ematipico (approved):
Thank you!!

PR #1409, opened 2021-03-20
fix: use method instead of type for LSPProgress

elianiva:
## Summary

Rome LSP uses `type` field for LSPProgress which breaks Neovim's builtin LSP client. It should be `method` instead of `type` according to [the LSP spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#progress)

## Test Plan

I tried to run the test *before* and *after* my PR to make sure I didn't mess up. Some tests are failing after 10s (`Test worker was unresponsive for 10 seconds. We tried to collect some additional metadata but we timed out again trying to fetch it...`), so I don't think they're caused by my changes


ematipico (approved):
Thank you! Looks good to me

PR #1412, opened 2021-03-24
fix(css-parser): change feature value type CSSString to CSSIdentifier

yeonjuan:
## Summary

According to [syntax specification](https://www.w3.org/TR/mediaqueries-5/#mq-syntax), string("") cannot be a media feature value.

```css
<mf-value> = <number> | <dimension> | <ident> | <ratio>
```

Currently, the CSS parser parses `<ident>` as a CSSString Node, so I think it should be changed to CSSIdentifier.

- test.css (input)

```css
@media (min-width: 30em) and (orientation: landscape) {

}
```

```sh
$ rome format test.css
```

- result

```css
@media (min-width: 30em) and (orientation: "landscape") {

}
```

- expected

```css
@media (min-width: 30em) and (orientation: landscape) {

}
```

## Test Plan

test case


ematipico (approved):
Totally missed that! Thank you!

PR #1413, opened 2021-03-26
feat(css-parser, formatter): `@supports` support

ematipico:
## Summary

Part of #984 , it implements the `@supports` feature for parser and formatter.

The nodes are a mirror of the official spec: https://drafts.csswg.org/css-conditional-3/#at-supports

## Test Plan

New tests


jeddy3:
This is fantastic, thank you! Many of stylelint's rules currently miss declarations within `@support` (because of the coarseness of the parser used). Being able to write linting rules that target declarations (regardless of whether they belong to `@support` or rule-sets) will an improvement over stylelint.

PR #1415, opened 2021-03-29
feat(lint): noImportantInKeyframes

ematipico:
## Summary

This PR enables linting on CSS files!

I also took the chance to fix various bugs around `loc` of some needs inside the parser. They are now realistic and correct (there might be others probably). I also removed a useless AST node inside the keyframe logic.

I created a new rule called `noImportantInKeyframes`, which is taken from stylelint. I updated the script in order to add the reference to their rule (like we do for eslint). They deserve a reference as they are the first that they implemented the rule.

I reviewed a bit the logic of inside the formatter of the keyframe, they should be better now.

## Test Plan

Created new tests and updated existing ones.


ematipico:
@jeddy3 [As you can see here]( https://deploy-preview-1415--romefrontend.netlify.app/docs/lint/rules/css/noimportantinkeyframe/), the rule page points to stylelint website.

One thing that I would like to know, what's the correct name: `stylelint` (all lower case) or `Stylelint` (capital S). I would like to get this right.

ematipico:
I will do the change, because we use the tests as documentation. So I think it makes sense to use an example that actually mimic the real world usage. Thank you for pointing that out!

PR #1416, opened 2021-04-01
feat(lint): noHeaderScope for HTML and JSX

zant:
## Summary

Implements #1345

Ported `noHeaderScope` to support both HTML and JSX under `a11y`.

## Test Plan

Added test for HTML and run `./rome ci`


ematipico (approved):
Thank you! Looks good to me!

Issue #1423, opened 2021-04-08
☂️ CSS Prefix: Missing Properties


jer3m01:
## How To 
Below is a list of properties missing from our current css prefix implementation.

Use `./script compiler-create-prefix [propertyName]`. *thanks @ematipico!*
This will create:
* `internal/compiler/transforms/compile/css-handler/prefix/prefixes/[propertyName].ts`
* `internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/[propertyName]/input.css`
* `website/src/docs/css-handler/prefix/[propertyName].md`

They should all reference (using a link) at the top of their file (below imports) where the info came from.
Info like: property name, special values, ...

They should export an array of `createPrefixVisitor`s and return a `prefixCSSProperty` or `prefixCSSValue` 
(covers 99% use case, if not discuss prior). 
```ts
export default [
  createPrefixVisitor({
    name: "<PROPERTY NAME>", // <PROPERTY NAME>/<VALUE NAME>
    enter(path) {
      return prefixCSSValue({
        path,
        propertyName: "<PROPERTY NAME>",
        value: "<VALUE NAME>", // Only for prefixCSSValue
        browserFeaturesKey: "<PROPERTY DATA NAME>", // From `caniuse-db`
      });
    },
  }),
  // ...
];
```
Prefixes will automatically be added to the property/value.

If some values/properties have special names (such as `-moz-inline-flex` being `-moz-inline-box`) 
use the `rename` argument as show [here](https://github.com/rome/tools/blob/main/internal/compiler/transforms/compile/css-handler/prefix/prefixes/display.ts#L6-L8) and [here](https://github.com/rome/tools/blob/main/internal/compiler/transforms/compile/css-handler/prefix/prefixes/display.ts#L20).

Also be sure to test on https://autoprefixer.github.io/ for the same output. **Don't forget to set target as `> 0%`**.
And write tests for them in `internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/<PROPERTY NAME>/input.css`

## Examples
[...]

PR #1424, opened 2021-04-08
chore: script file to generate prefix

ematipico:
## Summary

This PR does the following:
- create a new script called `compiler-create-prefix` that creates the main boilerplate of prefix and updates the main entry point `prefix/index.ts`. We don't have documentation yet but I think we should start creating the files, what we need later is to call the compiler and automatically print the result inside the documentation. This has basically two perks: automation and using the compiler for real;
- updates the `CONTRIBUTING.md` file with the new file;
- fixes a small issue around `AbsolutePath.createDirectory` where a folder should be created only when it doesn't exist;
- fixes an existing issue inside `generated-files/css-prefix` where a property should be converted to camel case, in case the file contains a dash, e.g. `box-shadow.ts`
- creates a basic `index.md`, empty, for future documentation

Part of #1423 

## Test Plan

Run manually the scripts

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,3 +180,13 @@ The `category` is one of the category folders inside the `language` folders.
 ```
 
 The created files will be displayed in the console output.
+
+### `compiler-create-prefix`
+
+This is used to generate a new prefix boilerplate
+
+```bash
+./script compiler-create-prefix [prefixName]
+```
+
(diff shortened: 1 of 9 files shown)


jer3m01 (approved):
Great idea! And thanks for fixing my script 😅. 
Just a few variable rename.

ematipico:
Variable updated :) thank you!

PR #1426, opened 2021-04-09
feat(css-prefix): box-shadow support

ajkachnic:
## Summary

This pull request adds support of the `box-shadow` property to the auto-prefixing part of the CSS compiler.

This satisfies one part of many parts of #1423 

## Test Plan

I added a test for the property and the output was as expected. It seems like this should be working. See `internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/box-shadow/input.test.md` for the test output (I can also put it here if necessary).


jer3m01:
LGTM. Thank you!

PR #1431, opened 2021-04-13
 feat(css-prefix): box-sizing support

magsout:
## Summary

Part of #1423.

Adds `box-sizing` related prefixes.

## Test Plan

ci


internal/compiler/transforms/compile/css-handler/prefix/prefixes/box-sizing.ts
```diff
@@ -0,0 +1,19 @@
+import {
+	createPrefixVisitor,
+	prefixCSSProperty,
+} from "@internal/compiler/transforms/compile/css-handler/prefix/utils";
+
+// https://github.com/Fyrd/caniuse/blob/main/features-json/css3-boxsizing.json
+// https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L209-L211
+export default ["border-box", "content-box"].map((value) =>
+	createPrefixVisitor({
+		name: `box-sizing/${value}`,
```
jer3m01, line 10: Just put the property name, no need for the value.
```ts
name: "box-sizing",
```
(can also remove the map)

jer3m01 (changes requested):
As stated in the review, values in the name are only required when prefixing values

internal/compiler/transforms/compile/css-handler/prefix/prefixes/box-sizing.ts
magsout: done

jer3m01:
Great thank you!

PR #1435, opened 2021-04-14
chore(css-prefix): add backdrop-filter

magsout:
## Summary

Part of #1423.

Adds `backdrop-filter` related prefixes

## Test Plan

ci


jer3m01:
Could you rebase on main and regenerate the list with `./rome run scripts/generated-files/css-prefix`?

Side note: In the future write the minimum amount of require tests (enough for the feature to be tested but don't make duplicates).
This avoids tests collisions (e.g. other prefixes get added and your test uses them and fails).

magsout:
@jer3m01 

>Side note: In the future write the minimum amount of require tests (enough for the feature to be tested but don't make duplicates).

what do you mean by minimum amount of requires tests for this example? just one property is enough like?

```
.style{
  backdrop-filter: none;
}
```

and that's all?

ematipico:
@magsout yes exactly!

If by accident we put another CSS property that is also affected by our prefixer, it gets transformed. This could cause confusions in the future.

magsout:
@ematipico ok, thanks for your answer. looking at the other tests, I thought we had to add more classes. I'll update my test so.

magsout:
tests updated. is it enough?

ematipico (approved):

jer3m01:
Perfect thanks!

PR #1436, opened 2021-04-16
feat(css-prefix): user-select

magsout:
## Summary

Part of #1423.

Adds `user-select` related prefixes

## Test Plan

ci

don't know If we have to support `element` https://github.com/w3c/csswg-drafts/commit/3f1d9db96fad8d9fc787d3ed66e2d5ad8cfadd05


ematipico (approved):

jer3m01:
Strange, there shouldn't be a prefix for `-ms-`. Nothing's wrong on your side, I've opened #1441 to fix this.


> don't know If we have to support element w3c/csswg-drafts@3f1d9db

We will handle renaming and fixing css values later, probably inside `rome check`.

Thanks this PR!

PR #1438, opened 2021-04-17
fix(lsp): adhere to JSON-RPC specification

strager:
LSP is built atop the JSON-RPC specification [1]. This specification
requires messages to have a key "jsonrpc" with a value of "2.0" (as a
string). Rome's LSP server does not include a "jsonrpc" key in its
messages. This protocol violation causes some LSP clients to reject
Rome's messages.

Include the required key, fixing Rome for these clients.

[1] https://www.jsonrpc.org/specification


ematipico:
@strager thank you for the PR! CI fails as you can see, would you mind fix it so we can merge it?

ematipico (approved):

PR #1440, opened 2021-04-17
fix(lsp): fix malformed window/logMessage messages

strager:
The LSP specification requires window/logMessage notifications to
include a "type" key and a "message" key. Rome's LSP server omits the
required "type" key. This protocol violation causes problems with some
LSP clients.

Adhere to the LSP by adding the missing "type" key.


ematipico:
CI is failing due to incorrect format. We could remove the log, it's fine without comment.

ematipico (approved):

PR #1143, opened 2020-08-23
feat(lint): lint/js/preferOptionalChaining rule

antonk52:
## Summary

This PR adds a new javascript linting rule that enforces the use of optional chaining (`?.`) operator. This is currently work in progress, so far I've covered the basic cases, going to expand on it to cover more complex ones soon. Would like to get any feedback prior spending more time on this.

fixes #765

## Test Plan

I've added test cases for the rule. Going to add more complex ones as them get covered.

Issue #1437, opened 2021-04-16
🐛 Regression: `rome check` says "No files linted"


strager:
## Environment

```
$ rome rage --summary
Environment Variables
  {
    COLORTERM: "truecolor"
    LANG: "en_US.UTF-8"
    TERM: "xterm-256color"
  }

User Config
  unset

Rome Version
  10.0.7-beta.18

Node Version
  15.12.0

Platform
  linux x64 5.4.86-stragerhard2

Terminal Features
  {
    background: "unknown"
    colorDepth: 24
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<50>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
    cwd: AbsoluteFilePath</home/strager/tmp/ro
    metest>
    realCwd: AbsoluteFilePath</home/strager/tm
    p/rometest>
  }

Server Status
  {
    projects: [{id: 0}, {id: 1}]
    workers: [
      {
        astCacheSize: 0
        heapTotal: 46_706_688
        ownedBytes: 0n
        ownedFileCount: 0
        pid: 584_643
        uptime: 0.2336173
      }
    ]
    server: {
      heapTotal: 46_706_688
      pid: 584_643
      uptime: 0.233711546
    }
  }
```

## Steps To Reproduce

1. Clone Rome's source repository.
2. Check out Rome version v10.0.4-beta. (This is to avoid issues with `rome init`: #1418)
3. Create and cd into a temporary directory.
4. Run `/path/to/rome-tools/rome init`.
5. Check out Rome commit 3d09ea15e3e0baab005e4774822cdb1330320f4a (current `main`).
6. Create a file called `test.js` with contents `let x, x;`.
7. Run `/path/to/rome-tools/rome check`.

## Actual results

`rome check` succeeds with no errors, but emits a warning:
```
⚠ No files linted
✔ No known problems!
```

## Expected Results

`rome check` fails with errors:
```
 test.js:1 lint/js/useSingleVarDeclarator  FIXABLE
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Declare variables separately.

    let x, x;
    ^^^^^^^^^

  ℹ Safe fix

    1   │ - let·x,·x;
      1 │ + let·x;
      2 │ + let·x;

 test.js:1:7 lint/js/noUnusedVariables ━━━━━━━━━━━

[...]

PR #1448, opened 2021-05-02
feat(lint): Add optional call expression support to preferOptianalChaining linting rule

antonk52:
## Summary

This PR is a follow up to #1143 & #1231. Now the rule also marks the code that can be written using optional call expressions as an error. Besides adding the OCE support I've also improved the auto fix to preserve the leading and trailing comment from the original code for OCE and optional member expressions.

## Test Plan

I've added the test cases for OCE and the once that check that the comment preservation works.

I have not yet pushed the auto fix for discovered `check` problems to see the CI output in GH actions

### DIFF
--- a/internal/child-process/index.ts
+++ b/internal/child-process/index.ts
@@ -31,23 +31,19 @@ export class ChildProcess {
 
 		const {stdout, stderr} = this.process;
 
-		if (stdout != null) {
-			stdout.on(
-				"data",
-				(chunk) => {
-					this.output.push([0, chunk]);
-				},
-			);
-		}
+		stdout?.on(
+			"data",
+			(chunk) => {
+				this.output.push([0, chunk]);
+			},
+		);
 
-		if (stderr != null) {
-			stderr.on(
-				"data",
-				(chunk) => {
-					this.output.push([1, chunk]);
-				},
-			);
-		}
+		stderr?.on(
+			"data",
+			(chunk) => {
+				this.output.push([1, chunk]);
+			},
+		);
 	}
 
 	public process: childProcess.ChildProcess;
--- a/internal/cli-diagnostics/DiagnosticsPrinter.ts
+++ b/internal/cli-diagnostics/DiagnosticsPrinter.ts
@@ -576,7 +576,7 @@ export default class DiagnosticsPrinter extends Error {
 			case "github-actions": {
 				const parts = [];
 
-				if (path !== undefined && path.isAbsolute()) {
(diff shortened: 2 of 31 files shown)


ematipico:
@antonk52 it seems your rule caught some bugs :) would you mind checking them? 

Could you please change the title of your PR? https://github.com/rome/tools/blob/main/CONTRIBUTING.md#creating-pull-requests

It should start with `feat(lint): `

antonk52:
Hi, @ematipico I've initially left off the autofixes so it would be easier to review the changes.

I've regenerated files and applied autofixes. Should be good to go now.

Please note that I've added suppressions for the `website` directory since Netlify uses an older version of nodejs which does not support optional call expressions.


PS: I've rebased my branch with the current main branch and now I'm having an issue that file paths are not being detected when passed to the rome cli command, ie `./rome check ./website`

As a result I get this output:

```sh
▲ ./rome check ./website
✔ Connected to daemon

 argv:1:11 args/fileNotFound ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ No matching lint files found

    rome check ./website --no-programmatic
               ^^^^^^^^^ website
```

I've been able to reproduce it on both my branch and current main branch with nodejs versions 16.0 and 14.16.1. I can also confirm that `process.cwd()` returns the correct directory. Should I create a separate issue? Not sure if it is related to #1437

ematipico:
>  Netlify uses an older version of nodejs which does not support optional call expressions.

How does Netlify interact with the JS that we ship on our website?

ematipico:
BTW, I couldn't replicate the issue you mentioned. I run the same command with and without daemon running. No issues were thrown.

antonk52:
> How does Netlify interact with the JS that we ship on our website?

When [this script](https://github.com/rome/tools/blob/4bee16144fec6d1a749dc4b25a7ef789c136d9f9/website/src/_includes/layouts/funding.liquid#L24) is getting included the liquid template engine(?) fails parsing the file when gets to `?.` and netlify cannot build the website with an error `unexpected "."`

> BTW, I couldn't replicate the issue you mentioned. I ran the same command with and without daemon running. No issues were thrown.

The problem went away once I've restarted my computer 🙂 I've created #1458 incase other people come across it to trouble shoot further.

ematipico (approved):

PR #1450, opened 2021-05-05
Included user-select

ktfth:
## Summary

Included user-select on the css-prefix

## Test Plan

rome test internal/compiler/transforms/compile/index.test.ts --update-snapshots


jer3m01:
Oh sorry, didn't see `user-select` was already done.

The existing format was correct, the property gets prefixed and not the values so we don't use a map nor `user-select/${value}`.
Same for `browserFeaturesKey` which is retrieved from https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js and not just the name of the property.

I would recommend doing `hyphens`, if you have any questions feel free to ask on #1423 or discord.

Again sorry for this issue.

PR #1454, opened 2021-05-06
fix: invalid flag message wrong string case

### LINKED ISSUE #1395: 🐛 Invalid flag message wrong string case
## Steps To Reproduce

1. `rome test --sync-test` (without the s)
image

## Expected Results
`Unknown syncTest flag at` and `- syncTest` should be `sync-test`

cola119:
## Summary

This PR fixes #1395.

## Test Plan

This PR adds a test to make sure that the kebab case is kept as is.

### DIFF
--- a/internal/cli-flags/Parser.test.md
+++ b/internal/cli-flags/Parser.test.md
@@ -814,3 +814,43 @@
 
 
 ```
+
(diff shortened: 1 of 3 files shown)


ematipico:
Thank you!

PR #1457, opened 2021-05-06
feat(css-prefix): cursor support

ktfth:
## Summary

Included cursor on the css-prefix

## Test Plan

rome test internal/compiler/transforms/compile/index.test.ts --update-snapshots


ematipico:
@jer3m01 is it correct to have a prefix rule that doesn't create any prefix? Like in this case.

jer3m01:
For `cursor` the values are prefixed:

image

https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L538-L554

You can have a look at https://github.com/rome/tools/blob/main/internal/compiler/transforms/compile/css-handler/prefix/prefixes/display.ts which prefixes values.

internal/compiler/transforms/compile/css-handler/prefix/prefixes/cursor.ts
```diff
@@ -0,0 +1,33 @@
+import {
+	createPrefixVisitor,
+	prefixCSSProperty,
+} from "@internal/compiler/transforms/compile/css-handler/prefix/utils";
+
+// https://github.com/Fyrd/caniuse/blob/main/features-json/css3-cursors-grab.json
+// https://github.com/Fyrd/caniuse/blob/main/features-json/css3-cursors-newer.json
+// https://github.com/Fyrd/caniuse/blob/main/features-json/css3-cursors.json
+export default ["zoom-in", "zoom-out", "grab", "grabbing"].map((value) => {
```
jer3m01, line 9: `zoom-in` and `zoom-out` have a different key from `grab` and `grabbing`, a better example of how this should be done is https://github.com/rome/tools/blob/main/internal/compiler/transforms/compile/css-handler/prefix/prefixes/transform.ts.
With `prefixCSSValue` instead of `prefixCSSProperty`.

jer3m01 (changes requested):
There's a simpler way then using a condition to find the browserFeaturesKey, linked an example in the review and to prefix values `prefixCSSValue` should be used, hope this is clearer.

internal/compiler/transforms/compile/css-handler/prefix/prefixes/cursor.ts
```diff
@@ -0,0 +1,33 @@
...
+	if (value === "grab" || value === "grabbing") {
+		browserFeaturesKeySufix = "grab";
+	}
+	browserFeaturesKey = [browserFeaturesKey, browserFeaturesKeySufix].join("-");
+	browserFeaturesKey = (browserFeaturesKey as unknown) as string;
+	return [
+		createPrefixVisitor({
+			name: `cursor/${value}`,
+			enter(path) {
+				return prefixCSSProperty({
```
jer3m01, line 24: `prefixCSSValue` should be used to prefix values instead of `prefixCSSProperty`.

jer3m01 (changes requested):

ktfth:
@jer3m01 Updated following your guidance, thank you

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/cursor/input.test.md
```diff
@@ -0,0 +1,29 @@
...
+
+```
+
+```
+
+### `Input`
+
+```css
+.style {
+	cursor: none;
```
jer3m01, line 17: Could you add test cases for zoom-in, zoom-out, grab and grabbing

internal/compiler/transforms/compile/css-handler/prefix/prefixes/cursor.ts
```diff
@@ -0,0 +1,47 @@
...
+	createPrefixVisitor({
+		name: "cursor",
+		enter(path) {
+			return prefixCSSProperty({
+				path,
+				propertyName: "cursor",
+				browserFeaturesKey: "css3-cursors",
+			});
+		},
+	}),
```
jer3m01, line 46: Manually checked the data and it doesn't seem to do anything for rome, this can be removed.

jer3m01 (changes requested):
Great! Just missing tests and one can be removed.

ktfth:
@jer3m01 New modification following your review comments, check it out

jer3m01 (approved):
Awesome! One last thing, could you run `./script generated-files/css-prefix` to fix the CI.

ktfth:
> Awesome! One last thing, could you run `./script generated-files/css-prefix` to fix the CI.

Ok

jer3m01:
Thanks!

Issue #1295, opened 2021-01-13
🐛 formatter should not remove generic trailing comma in tsx


yeonjuan:
## Environment

```
$ rome rage --summary
Environment Variables
  Object {
    COLORTERM: "truecolor"
    LANG: "en_US.UTF-8"
    TERM: "xterm-256color"
    TERM_PROGRAM: "vscode"
  }

User Config
  unset

Rome Version
  10.0.4-beta-dev

Node Version
  12.16.3

Platform
  darwin x64 18.7.0

Terminal Features
  Object {
    background: "unknown"
    colorDepth: 24
    columns: 170
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
  }

Client Flags
  Object {
    clientName: "cli"
    silent: false
    cwd: ___R$project$rome$$internal$path$index_ts$AbsoluteFilePath: /Users/yeonjuan/Desktop/open-source/tools
    realCwd: ___R$project$rome$$internal$path$index_ts$AbsoluteFilePath: /Users/yeonjuan/Desktop/open-source/tools
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 27_672_576
      pid: 84_558
      uptime: 0.781767559
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 27_672_576
        ownedBytes: 0n
        ownedFileCount: 0
        pid: 84_558
        uptime: 0.78147324
      }
    ]
  }
```

## Steps To Reproduce

1. write below example in a file with `.tsx` extension 
```tsx
// test.tsx
const func = <T,>() => {
}
```

2. run formatter

```bash
$ ./rome format test.tsx
```

- output

```tsx
const func = <T>() => {}; // ts error: JSX Element `T` has no corresponding closing tag
```

## Expected Results

The result makes ts parsing error so it should not remove trailing comma.

```tsx
const func = <T,>() => {}; 
```

ematipico:
At the beginning I thought it was a typescript bug  it's not! Although it seems strange having a trailing comma

antonk52:
it's the only way for typescript to differentiate the initial type assertion syntax in tsx file with jsx ie

```ts
// *.ts
const a = <string>bar;
const b = bar as string;
```

```tsx
// *.tsx

// error
const a = <string>bar;

// fine
const a = <string,>bar;
const b = bar as string;
```

[...]

PR #1463, opened 2021-05-07
Fix(formatter): preserve generics trailing comma in arrow function of tsx

cola119:
## Summary

This PR fixes https://github.com/rome/tools/issues/1295

As you probably know, TSX arrow functions must have a trailing comma of generics when it has a single type annotation, which means `TSTypeParameterDeclaration` need to know its grandparent node but the current node structure can't retrieve ancestors as far as I know. Eventually, I tried to solve this by changing `JSFunctionHead` (the middle node). Correct me if I wrong or you have any better solution.

This patch is based on the prettier implementation (https://github.com/prettier/prettier/issues/6114).

## Test Plan

This adds some tests of both tsx and typescript.


ematipico:
I think, here in Rome, we can apply this fix at parser level. I *think* in prettier they had to do it that way because they don't control the parser. Here we can. 

The parser has access to the `path` which contains all the information about the file that is processing. I believe it would be a better fix. What do you think?

cola119:
@ematipico Thank you for your review. That makes sense so I'll try to modify the parser per se.

internal/formatter/test-fixtures/tsx/generics/input.test.md
```diff
@@ -0,0 +1,67 @@
...
+type Type1<T> = "type1";
+
+type Type2 = Two<T>;
+
+```
+
+### `Output`
+
+```tsx
+const f1 = <T,>(arg: T) => false;
```
cola119, line 44: This is the main purpose.

cola119:
@ematipico I fixed `js-parser`. Could you review it again?

internal/formatter/builders/js/typescript/TSTypeParameterDeclaration.ts
```diff
@@ -33,7 +33,13 @@ export default function TSTypeParameterDeclaration(
 		params[0].type !== "TSMappedType";
 
 	if (shouldInline) {
-		return concat(["<", builder.tokenize(params[0], node), ">"]);
+		const hasTrailingComma = "hasTrailingComma" in node && node.hasTrailingComma;
```
ematipico, line 36: Wasn't `!!node.hasTrailingComma` sufficient? Adding a string to check a boolean doesn't seem ideal

ematipico (approved):
A better solution from my point of view! Thank you very much! Just left a comment

internal/formatter/builders/js/typescript/TSTypeParameterDeclaration.ts
cola119: `node` is defined as `TSTypeParameterDeclaration | TSTypeParameterInstantiation` and `TSTypeParameterInstantiation` doesn't have `hasTrailingComma` props. 
image

So we need to narrow with in operator, I think.
https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-the-in-operator

PR #1469, opened 2021-05-07
feat(lint): useHeadingContent for HTML

JoostKersjes:
## Summary

Part of #1345.

Ported the useHeadingContent rule to the `a11y` category.

Picked this up after finding this https://github.com/rome/tools/issues/1345#issuecomment-832267503

Used #1444 as my example.

## Test Plan

```shell
$ ./rome ci
 Building trunk 
rome check

  ✔ 1_995 files linted!

rome test

  ✔ 4_009 tests passed!
```


ematipico (changes requested):
Thank you for your first contribution!

The CI fails because you should run the command `./script generate-all-files`.

This command will regenerate all the documentation files. Once run, you can commit the files that changed.

JoostKersjes:
Ran the script and pushed the changes :)

I also did this a couple of times development. Would you suggest running this script as part of the last few steps when making changes?

ematipico:
> Ran the script and pushed the changes :)
> 
> I also did this a couple of times development. Would you suggest running this script as part of the last few steps when making changes?

Yes, it should be the last step. In case of lint rules, for example, we use the test cases as part of our documentation. If you check netlify deployment, you'll be able to see the new rule updated with the HTML cases (valid and invalid).

JoostKersjes:
> 
> 
> > Ran the script and pushed the changes :)
> > I also did this a couple of times development. Would you suggest running this script as part of the last few steps when making changes?
> 
> Yes, it should be the last step. In case of lint rules, for example, we use the test cases as part of our documentation. If you check netlify deployment, you'll be able to see the new rule updated with the HTML cases (valid and invalid).

Thanks! Is replacing the "MISSING DOCUMENTATION" also part of this task, or does this go on to review next?

ematipico:
No, it's necessary for now. Eventually though, we'll need to provide some decent documentation (not just code) for all our rules

PR #1471, opened 2021-05-08
feat(css-prefix): gradient support

ktfth:
## Summary

Included gradient on the css-prefix

## Test Plan

rome test


internal/compiler/transforms/compile/css-handler/prefix/prefixes/gradient.ts
```diff
@@ -0,0 +1,99 @@
+import {
+	createPrefixVisitor,
+	prefixCSSValue,
+} from "@internal/compiler/transforms/compile/css-handler/prefix/utils";
+
+// https://github.com/Fyrd/caniuse/blob/main/features-json/css-gradients.json
+export default [
```
jer3m01, line 7: Instead of writing each one you can use `flatMap()` & `map()`:

```ts
export default [
		"background",
		"background-image",
		"border-image",
		"mask",
		"list-style",
		"list-style-image",
		"content",
		"mask-image",
	].flatMap((propertyName) => [
	    "linear-gradient",
	    "repeating-linear-gradient",
	    ...
	    ].map((value) => {
			   ...
			  prefixCSSValue({
					propertyName,
					value
				});
				...
		})
);

File: internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/gradient/input.test.md [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+
+```
+
+```
+
+### `Input`
+
+```css
+.style {
+	gradient: none;
```
jer3m01, line 17: Can you add tests cases for all of them (might be long).

jer3m01 (changes requested):
You can avoid duplicates (see review) and add missing tests.

internal/compiler/transforms/compile/css-handler/prefix/prefixes/gradient.ts
ktfth: Awesome, i can do that

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/gradient/input.test.md
ktfth: Count with me

internal/compiler/transforms/compile/css-handler/prefix/prefixes/gradient.ts
```diff
@@ -0,0 +1,35 @@
...
+	].map((value) =>
+		createPrefixVisitor({
+			name: `${propertyName}/${value}`,
+			enter(path) {
+				return prefixCSSValue({
+					path,
+					propertyName,
+					value,
+					browserFeaturesKey: "css-gradients",
+				});
```
jamiebuilds, line 31: We shouldn't quite merge this yet until `prefixCSSValue` works with complex css values such as the case with the `background` shorthand

jer3m01:
Would have to dig in deeper as the list in the source (https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L170-L194)
doesn't match with the output from autoprefixer.
Some are [defined](https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L180) but not actually prefixed?:

image

ktfth:
> Would have to dig in deeper as the list in the source (https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L170-L194)
> doesn't match with the output from autoprefixer.
> Some are [defined](https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L180) but not actually prefixed?:
> 
> image

I don't know exactly what this mean

jer3m01:
> 
> I don't know exactly what this mean

No problem, this comment is more aimed towards me for later or others members of the team. 

Essentially the code says that they should be prefixed (as you've coded) but autoprefixer doesn't actually do it. Which is really strange, if anyone else has an explanation to why they're welcomed!

jamiebuilds:

Thanks again!

PR #1482, opened 2021-05-11
docs(theme): correctly align theme toggle switch

Haroenv:
## Summary

The default padding on `button` elements isn't even on all sides, so it makes sense to remove it completely, since it looks like it was meant to be this way.

## Test Plan

look at the website and verify the new alignment.

before|after
---|---
before, light mode | after, light mode
before, dark mode | after, dark mode


EduardoLopes:
nice! this can be merged ✨

PR #1489, opened 2021-05-12
feat(lint): port noRedundantRole rule to a11y and add html support

UsamaHameed:
## Summary
This PR ports the last `jsx-a11y` rule to `a11y` and adds support for html files as well. Part of #1345 
I also removed some docs related to some other `jsx-a11y` rules that were still there.

## Test Plan
rome ci


ematipico (approved):
Top notch PR!! Thank you!!!

PR #1497, opened 2021-05-13
feat(css-prefix): width, height, size (intrinsic-width browser feature)

JustBeYou:
## Summary
CSS prefixs for width related properties. Issue #1423 

`width`, `height`, `size` and `grid` related properties are related to `instrinsic-width` feature. (https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L535). `grid` is not implemented yet, the others are.

## Test Plan
* `rome ci`
* https://autoprefixer.github.io/ - prefixes are generated correctly but in a different order


jer3m01 (approved):
For `grid` I see autoprefixer uses specific flags, would need to refactor the data to something more advanced then true/false as it doesn't cover everything.

Let's leave the TODO until I've added ways to support it.

JustBeYou:
If you need help or don't have time to deal with the refactor I could look into it too. Meanwhile, I'll look into a few more css props that could be implemented.

UsamaHameed:
Just out of curiosity, I added the prefixes for grid here: https://github.com/rome/tools/pull/1493 if what I found is correct then only `grid-template-rows` and `grid-template-columns` require prefixes. Is there something I am missing?

JustBeYou:
@UsamaHameed some `grid-*` properties support `intrinsic-width` related vales (like `min-content`, `max-content`). The prefixes are for the values, not for the properties themselves. You can take a look here https://github.com/postcss/autoprefixer/blob/main/data/prefixes.js#L535

ematipico:
Thank you! Can you please rebase and run `./script generate-all-files`?

JustBeYou:
@ematipico should be fine now!

Issue #1484, opened 2021-05-11
🐛 Lint suppression comments in JSX don't work


JoostKersjes:
## Environment

```
$ yarn rome rage --summary
yarn run v1.22.10
$ /home/joost/projects/personal/kersjes.dev/node_modules/.bin/rome rage --summary
Environment Variables
  Object {
    COLORTERM: "truecolor"
    LANG: "C.UTF-8"
    TERM: "xterm-256color"
    TERM_PROGRAM: "vscode"
  }

User Config
  unset

Rome Version
  10.0.4-beta

Node Version
  16.1.0

Platform
  linux x64 4.19.104-microsoft-standard

Terminal Features
  Object {
    background: "unknown"
    colorDepth: 24
    columns: 319
    cursor: true
    hyperlinks: true
    isTTY: true
    unicode: true
  }

Client Flags
  Object {
    clientName: "cli"
    cwd: "/home/joost/projects/personal/kersjes.dev"
    realCwd: "/home/joost/projects/personal/kersjes.dev"
    silent: false
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 61_120_512
      pid: 22_252
      uptime: 0.8071922
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 61_120_512
        ownedBytes: 0
        ownedFileCount: 0
        pid: 22_252
        uptime: 0.8070671
      }
    ]
  }


Done in 0.94s.
```

## Steps To Reproduce

1. Create a JSX linting error. For example `<Component {...pageProps} />` will create a `lint/jsx/noPropSpreading` violation.
2. Add a suppression comment in the JSX. For example `{/* rome-ignore lint/jsx/noPropSpreading: nextjs _app.tsx convention */}`.
3. Run `rome check`.

As you can see in the output below, the `lint/jsx/noPropSpreading` is still detected as a violation and the JSX comment is seen as a "formatting" fix.

```
$ yarn rome check src/pages/_app.tsx 
yarn run v1.22.10
$ /home/joost/projects/personal/kersjes.dev/node_modules/.bin/rome check src/pages/_app.tsx

[...]

PR #1501, opened 2021-05-14
fix(js-parser): handle JSXEmptyExpression position

yeonjuan:
## Summary

Rome parser does not properly handle the JSXEmptyExpresssion's position. So the comments in JSXExptyExpression attached in `trailingComments`. (But I think it should be in `innerComments`).

- test.jsx
```js
const Foo = () => {
	return <>
		{/* comment */}
	</>
}
```
 - AST (`rome parse test.jsx`)
 
```text
// ...
  JSXExpressionContainer {
                          expression: JSXEmptyExpression {trailingComments: ["1"]}
                        }
//...
```

It's the reason why #1484 occured. The Formatter prints comment at the outside of JSXContainer (`{..}`). 
```
const Foo = () => {
        return <>
                {} /* comment */
        </>;
};
```

## Test Plan

There was already test case (wrong, but will be fixed in this PR). see [internal/compiler/lint/rules/jsx/noCommentText.test.md](https://github.com/rome/tools/blob/main/internal/compiler/lint/rules/jsx/noCommentText.test.md#4-formatted)

### DIFF
--- a/internal/compiler/lint/rules/jsx/noCommentText.test.md
+++ b/internal/compiler/lint/rules/jsx/noCommentText.test.md
@@ -107,7 +107,7 @@ const a = <div>
 
 ```tsx
 const a = <div>
-	{} /* comment */
+	{/* comment */}
 </div>;
 
 ```
@@ -122,7 +122,7 @@ const a = <div>
 
 ```tsx
 const a = <div>
-	{} /** comment*/
+	{/** comment*/}
 </div>;
 
 ```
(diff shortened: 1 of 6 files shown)


ematipico:
We should probably check the formatter too. I recently changed to use the trailing commas

ematipico:
Never mind, I thought my PR was merged but it wasn't. Looks good to me!

ematipico (approved):

PR #1507, opened 2021-05-17
feat(lint): Format else conditions on own line if they had line break previously

### LINKED ISSUE #1024: 💭 Formatter feature request: comment before `else` keyword
One of the most contentious decisions in Prettier and the reason I personally no longer use it for any projects, was its forcing the removal of breaks before else blocks: https://github.com/prettier/prettier/issues/840

To illustrate two issues with this, consider the following code:

```
  // comment for if block
  if (cond1) {
    doFoo(tree, data);
    if (cond2) {
      collect(tree.left, data, depth + 1);
      collect(tree.right, data, depth + 1);
    }
  }
  // comment for else block
  else {
    doBar(data);
  }
```

As it is now, Rome will throw errors suggesting the following format

```
  // comment for if block
  if (cond1) {
    doFoo(tree, data);
    if (cond2) {
      collect(tree.left, data, depth + 1);
      collect(tree.right, data, depth + 1);
    }
  } else {
    // comment for else block
    doBar(data);
  }
```

The problems with eliminating the break before the else are that it now lines up the else block with the wrong conditional and that it eliminates the clearest place for the else block comments.

I don't see much upside in a linter being rigid to this degree, and based on the the discussion on the Prettier issue linked above, many others would agree.

---
Edit: Since the title was changed, I'll reiterate:

**I strongly prefer not removing the break before else, even in cases where there are no comments. The `else` should line up with the `if` it corresponds to, not an inner one.**

jamiebuilds:
## Summary

Fixes #1024

If `else` or `else if` conditions are printed on a line after the end of their `if` block, the newline is preserved.

```js
if (cond) {
  // ...
}
else {
  // ...
}
```

## Test Plan

Tests are included


internal/formatter/builders/js/statements/JSIfStatement.ts
```diff
@@ -53,12 +54,24 @@ export default function JSIfStatement(
...
+	else {
 		tokens.push(builder.tokenize(node.consequent, node));
 	}
 
 	if (node.alternate) {
-		tokens.push(space, "else", space, builder.tokenize(node.alternate, node));
+		let separator: Token = space;
+
+		if (compareLines(node.consequent, node.alternate) === LineComparison.After) {
+			separator = hardline;
```
jamiebuilds, line 66: I used `hardline` here, but I don't know if I fully understand when `softline` vs `hardline` are meant to be used. This seemed consistent with how its treated

internal/ast-utils/compareLines.ts
```diff
@@ -0,0 +1,24 @@
+import {AnyNode} from "@internal/ast";
+
+export enum LineComparison {
+	Unknown,
+	Same,
+	Before,
+	After,
+}
```
jamiebuilds, line 8: I used an enum for this because `isSameLine()` wasn't enough information as a boolean. I needed to specifically know that the location was known to be on a newline _after_ the end of the compared node.

ematipico (approved):

PR #1508, opened 2021-05-17
feat(cli): interactive yes/no and suggestion selection for cli commands

### LINKED ISSUE #974: 📎 Interactive yes/no when suggesting CLI commands in a TTY
Inspired by https://twitter.com/goshakkk/status/1291161477353799681

We can ask the user if they really meant the command that was suggest in the console when in a TTY.

We can perform some validation before on the suggested command flags too to make sure that accepting it wont result in more errors.

UsamaHameed:
## Summary
This pr resolves #974. 

## Test Plan

1. `rome chekc` suggests the following and the user can interactively select one
```
❯ Unknown command. Found matching commands: 
ℹ Use arrow keys and then enter to select an option
  ◉ check
  ◯ cache dir
  ◯ cache clear
```

2. `rome losg` suggests `rome logs` and the user can choose to execute `rome logs` or not
```
❯ Did you mean logs
ℹ Use arrow keys and then enter to select an option
  ◉ No (shortcut n)
  ◯ Yes (shortcut y)
```

3. throws error if it can't find a relevant match.

```
argv:1:9 flags/invalid  FATAL  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Unknown command kjsahdaslkjdlkasjdlksajldjkaslkdjalskjda

    dev-rome kjsahdaslkjdlkasjdlksajldjkaslkdjalskjda
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  ℹ To see all available commands run

    dev-rome --help

  ⚠ Rome exited as this error could not be handled and resulted in a fatal error. Please report if necessary.

```

### DIFF
--- a/internal/cli-flags/Parser.test.md
+++ b/internal/cli-flags/Parser.test.md
@@ -164,7 +164,67 @@
 
 ```
 
-## `command required with wrong command`
+## `command required with wrong command but multiple generated suggestions`
+
+### `flags`
+
+```javascript
+{}
+```
+
+### `help`
+
+```
+
+ Usage ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  test foo1 [flags]
+
+ Global Flags ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  ℹ To view global flags run
+  $ rome --help
+
+
+```
+
+### `output`
+
+```
+
+```
+
+## `command required with wrong command but only one generated suggestion`
+
+### `flags`
+
+```javascript
+{}
+```
+
+### `help`
+
+```
+
+ Usage ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  test foobar [flags]
+
+ Global Flags ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  ℹ To view global flags run
+  $ rome --help
+
+
+```
+
+### `output`
+
+```
+
+```
+
+## `command required with wrong command but provided suggestion`
 
 ### `help`
 
@@ -184,8 +244,6 @@
 
  Commands ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-  - foobar
-
   ℹ To view help for a specific command run
   $ test command_name --help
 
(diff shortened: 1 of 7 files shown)


jamiebuilds:
The double `: : check` here looks weird. Is it supposed to be removing the line entirely?

Screen Shot 2021-05-18 at 4 21 16 PM

internal/cli-reporter/Reporter.ts
```diff
@@ -154,6 +154,7 @@ export default class Reporter implements ReporterNamespace {
 					cwd: CWD_PATH,
 					home: HOME_PATH,
 				},
+				stdin: process.stdin,
```
ematipico, line 157: I don't think this is the correct approach. `Reporter` has already a `private stdin` (line 138) and it has a method called `getStdin`. We should use that one.

internal/cli-flags/Parser.ts
```diff
@@ -1141,41 +1159,53 @@ export default class Parser<T> {
...
-					args: displayArgs,
-					defaultFlags,
-					flags: rawFlags,
-				},
-				"command",
-			),
-		};
+				suggestions.forEach((suggestion) => {
+					options[suggestion] = {label: markup`${suggestion}`};
+				});
```
ematipico, line 1173: We try to stay away from this syntax. We always use `for..of` when dealing with iterators.

ematipico: BTW, you could use `.reduce`, which seems to be a better fit

internal/cli-flags/Parser.ts
```diff
@@ -1141,41 +1159,53 @@ export default class Parser<T> {
...
 
-		throw new DiagnosticsError("Unknown command", [diag]);
+				const selected = await this.reporter.select(
+					markup`Unknown command. Found matching commands: `,
```
ematipico, line 1176:
```suggestion
					markup`Unknown command <emphasis>${commandName}</emphasis>. Found matching commands: `,
```

internal/cli-flags/Parser.ts
```diff
@@ -1141,41 +1159,53 @@ export default class Parser<T> {
...
+
+				const args = Array.from(selected)[0].split(" ");
+				await this.reinit(args);
+			} else if (suggestions.length === 1) {
+				const yes = await this.reporter.radioConfirm(
+					markup`Did you mean ${suggestions[0]}`,
+				);
+				if (yes) {
+					await this.reinit(suggestions);
+				}
```
ematipico, line 1191: What would happen if the user didn't confirm?

ematipico (changes requested):
Nice PR! I left few comments around code/logic

internal/cli-flags/Parser.test.ts
```diff
@@ -394,6 +402,41 @@ test(
...
 );
 
+test(
+	"command required with wrong command but multiple suggestions",
+	async (t) => {
+		const reporter = new Reporter("ParserTest", {stdin: process.stdin});
+		reporter.select = function() {
+			const res = new Set(["foo1"]);
+			// rome-ignore lint/ts/noExplicitAny: TS generics constraint :(
+			return new Promise((resolve) => resolve(res as any));
```
ematipico, line 412:
```suggestion
			return Promise.resolve(res as any)
```

internal/cli-flags/Parser.test.md
```diff
@@ -164,7 +164,13 @@
 
 ```
 
-## `command required with wrong command`
+## `command required with wrong command and suggestion`
```
ematipico, line 167: I think the tests are a bit off

internal/cli-flags/Parser.ts
jamiebuilds: Personally I dislike `Array#reduce` I think it makes code unnecessarily confusing only to avoid local mutation which I think is perfectly fine

internal/cli-flags/Parser.ts
jamiebuilds: Should probably fall back to `throw new DiagnosticsError("Unknown command", [diag]);`

internal/cli-reporter/Reporter.ts
UsamaHameed: @ematipico that stdin field is what I am defining with `process.stdin` through the constructor when creating a `Reporter` on line 150. This reporter is returned by the `fromProcess` static method inside the Reporter class.

internal/cli-flags/Parser.ts
UsamaHameed: I avoided `reduce` for the same reason. I don't mind using it if thats what we follow. I like `for of` as well.

UsamaHameed:
> The double `: : check` here looks weird. Is it supposed to be removing the line entirely?
> 
> Screen Shot 2021-05-18 at 4 21 16 PM

The `confirm` method on `Reporter` does not clear the line but I can remove that double `:`

internal/cli-flags/Parser.ts
ematipico: Then, we should switch to `for..of` 💪

internal/cli-flags/Parser.test.md
UsamaHameed: These snapshots only capture diagnostic output, if there is none, the `Output` is empty. I don't know how much value do those tests provide us.

ematipico (approved):

UsamaHameed:
@jamiebuilds I have applied the suggestions. Please let me know if I can apply further improvements. 🙏

ematipico:
@jamiebuilds This looks good to me. Could you please do a last check? I think we can merge it.

PR #1510, opened 2021-05-18
chore: Add .rgignore file

jamiebuilds:
## Summary

```
scripts/vendor/rome.cjs
scripts/vendor/rome.cjs.map
```

These two files show up in almost every search you could make in the rome repo. Adding them to `.rgignore` makes ripgrep (`rg`) ignore them when searching through files. ripgrep is also used by vscode so it cleans up the results you see there as well.

## Test Plan

Search for something in vscode, you shouldn't find anything in these files.

Also note that you can toggle this button to turn the ignore files on/off.

Screen Shot 2021-05-17 at 4 02 23 PM

### DIFF
--- a/.rgignore
+++ b/.rgignore
@@ -1,0 +1,12 @@
+#
+# This file can be used to ignore certain files from ripgrep/vscode's search
+# results.
+#
+# Before you add anything to this file, keep in mind that it will affect
+# everyone's ability to grep for things. Consider creating your own local
+# `.rgignore` file in your `$HOME` directory.
+#
+
+# Checked-in pre-built version of Rome
+scripts/vendor/rome.cjs
+scripts/vendor/rome.cjs.map


yassere:
I think we should generally encourage contributors to manage most of their tooling-specific config individually, and that we should only add new top-level config files when they provide major convenience.

In this case, it seems easy enough to configure VS Code with `search.exclude` and to set a `RIPGREP_CONFIG_PATH` for the ripgrep cli.

But I don't have very strong opinions about this if you think this will be helpful for new contributors.

ematipico:
@yassere at the beginning we've tried this approach. I still remember I did a PR in the past for ignoring files around IDEA IDEs. This has changed: https://github.com/rome/tools/blob/main/.gitignore#L4-L5

I'd prefer having things like that inside the repo for two reasons:
- it lowers down the friction for first time contributions (we'd need to "teach" first time contributors how to ignore these files on their computer)
- we avoid to tarnish someone else's machine, what we expect to be ignored might not be the same for another project

Issue #1418, opened 2021-04-03
🐛 Rome init missing argument


jer3m01:
## Environment

```
$ rome rage --summary
Environment Variables
  {LANG: "C.UTF-8", TERM: "xterm-256color"}

User Config
  unset

Rome Version
  10.0.7-nightly.2021.2.10

Node Version
  15.11.0

Platform
  linux x64 4.4.0-19041-Microsoft

Terminal Features
  {
    background: "unknown"
    colorDepth: 8
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<472>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
    cwd: AbsoluteFilePath</mnt/d/Webstorm/test/oui>
    realCwd: AbsoluteFilePath</mnt/d/Webstorm/test/oui>
  }

Server Status
  {
    projects: [{id: 0}, {id: 1}]
    workers: [{astCacheSize: 0, heapTotal: 32_452_608, ownedBytes: 0n, ownedFileCount: 0, pid: 1_117, uptime: 0.3663973}]
    server: {heapTotal: 32_452_608, pid: 1_117, uptime: 0.3665485}
  }

```

## Steps To Reproduce

1. Create an empty project
2. `yarn rome init`
![image](https://user-images.githubusercontent.com/33844379/113492299-c68d6700-94d6-11eb-8cf4-562f1e02e052.png)

## Expected Results
Same as rome@latest

superhawk610:
The issue appears to be caused by [this line](https://github.com/rome/tools/blob/main/internal/core/server/commands/init.ts#L99), which is asserting that 1 argument is passed when in fact no arguments should be passed (there's even a check for 0 arguments a few lines earlier). It looks like this is confusing arguments and flags.

I believe there's an additional issue with the client-side `init` command where it always reports that a project already exists, but I can open a separate issue once this is resolved.

ematipico:
> The issue appears to be caused by [this line](https://github.com/rome/tools/blob/main/internal/core/server/commands/init.ts?rgh-link-date=2021-05-05T21%3A32%3A11Z#L99), which is asserting that 1 argument is passed when in fact no arguments should be passed (there's even a check for 0 arguments a few lines earlier). It looks like this is confusing arguments and flags.
> 
> ```ts
> req.expectArgumentLength(1);
> ```
> 
[...]

PR #1515, opened 2021-05-18
Rome init missing argument

jamiebuilds:
@ematipico in response to https://github.com/rome/tools/issues/1418#issuecomment-835376565

> Actually, the argument is expected to be passed. One of the most recent refactors broke the passage of arguments between commands using the Rome server.

I'm wondering what you expect this argument to be?

### DIFF
--- a/internal/core/server/commands/init.ts
+++ b/internal/core/server/commands/init.ts
@@ -73,7 +73,7 @@ export default createServerCommand<Flags>({
 		const {configType, indentSize, indentStyle, checkProject} = flags;
 
 		// Warn if provided with arguments
-		if (args.length > 0) {
+		if (args.length > 0) { // TODO: Should this assertion be changed?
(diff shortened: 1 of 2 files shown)


internal/core/server/commands/init.ts
```diff
@@ -96,7 +96,7 @@ export default createServerCommand<Flags>({
 		// - second part creates a basic configuration file and and the .editorconfig
 		// part of the command where we check if a configuration already exists
 		if (checkProject) {
-			req.expectArgumentLength(1);
+			req.expectArgumentLength(1); // TODO: Does this belong here at all?
```
ematipico, line 99: This isn't correct. Probably `expectArgumentLength` now works as expected. Although, even if we removed this, there are few issues that I need to solve. I am already working on them, I will have PR open soon.

PR #1522, opened 2021-05-20
feat(css-prefix): pseudo-selectors

JustBeYou:
## Summary

- [X] refactor `createPrefixVisitor` to allow exploring other node kinds than `CSSBlock`
- [X] prefix pseudo-classes
- [x] prefix pseudo-elements
- [x] discard unnecessary prefixes from properties if the pseudo-class was prefixed
- [x] improve naming, remove draft comments, lint the code

Related to issue #1423
Also includes refactor for `createPrefixVisitor`

## Test Plan
Snapshot test. Autoprefixer is broken on this test case, don't rely on it.


internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -243,3 +405,60 @@ function getTargets(path: CompilerPath): Browser[] {
...
+	State extends UnknownObject,
+	PathType extends CompilerPath
+> {
+	name: string;
+	enter?: (path: PathType, state: VisitorStateEnter<State>) => EnterSignal;
+	exit?: (path: PathType, state: VisitorStateExit<State>) => ExitSignal;
+}
+
+// vistitor for (CompilerPath, State) => visitor for (PathType, State)
+export function __visitorTransformer<
```
ematipico, line 441: Is there a reason why put a double underscore in  the name of this function?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
 	return signals.retain;
 }
 
+export type PrefixCSSRootCompilerPath = CompilerPath & {
+	node: CSSRoot;
+};
+
+export interface PrefixCSSRootVisitor<State extends UnknownObject> extends __TypedVisitor<
```
ematipico, line 208: I would suggest to start creating a type for this `State`. We could call it `CompilerPrefixerState`

```ts
export type CompilerPrefixerState = UnknownObject;
```

And replace all the occurrences where we have `State extends UnknownObject`

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+function isPseudoClass(node: AnyCSSPattern): node is CSSPseudoClassSelector {
+	return node.type === "CSSPseudoClassSelector";
+}
+
+function collectCSSSelectorPrefixes(
+	selector: CSSSelector,
+	namesToFeatures: Map<string, string>,
+	targets: Browser[],
+) {
+	return selector.patterns.reduce(
```
ematipico, line 232: We are trying to stay away from `.reduce`. Let's use `for..of`.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+function isCSSRule(node: AnyNode): node is CSSRule {
+	return node.type === "CSSRule";
+}
+
+function prefixCSSSelector(
+	selector: CSSSelector,
+	prefix: string,
+	namesToFeatures: Map<string, string>,
+	targets: Browser[],
+) {
```
ematipico, line 281: I think using an object is way better than passing four arguments:
- order doesn't matter
- easily extendable

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+		patterns: newPatterns,
+	});
+}
+
+function prefixCSSRule(
+	rule: CSSRule,
+	prefix: string,
+	namesToFeatures: Map<string, string>,
+	targets: Browser[],
+) {
```
ematipico, line 311: I think using an object is way better than passing four arguments:
- order doesn't matter
- easily extendable

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+
+		return pattern;
+	});
+	return cssSelector.create({
+		...selector,
+		patterns: newPatterns,
+	});
+}
+
+function prefixCSSRule(
```
ematipico, line 306: This function prefixes the `prelude` of the rule, so maybe it's worth changing its name? `prefixCSSRulePrelude`?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+	rule: CSSRule,
+	prefix: string,
+	namesToFeatures: Map<string, string>,
+	targets: Browser[],
+) {
+	// COMMENT: here we are sure that at least one selector will be changed
+	// so I think that's the most efficient way to write it
+	const newSelectors = rule.prelude.map((selector) =>
```
ematipico, line 314: `newPrelude`?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+	return cssRule.create({
+		...rule,
+		prelude: newSelectors,
+	});
+}
+
+export function prefixPseudoInCSSRoot(
+	path: PrefixCSSRootCompilerPath,
+	namesToFeatures: Map<string, string>,
+) {
```
ematipico, line 326: Could you add the return type please?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+	const newSelectors = rule.prelude.map((selector) =>
+		prefixCSSSelector(selector, prefix, namesToFeatures, targets)
+	);
+	return cssRule.create({
+		...rule,
+		prelude: newSelectors,
+	});
+}
+
+export function prefixPseudoInCSSRoot(
```
ematipico, line 323: I don't understand this function. Why do we need to replace the entire root node?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+		new Set<string>(),
+	);
+}
+
+function collectCSSRulePrefixes(
+	rule: CSSRule,
+	namesToFeatures: Map<string, string>,
+	targets: Browser[],
+) {
+	return rule.prelude.reduce(
```
ematipico, line 256: We can avoid `.reduce` here. If we're worried about possible duplicates, we can create a new `Set` at the very end of the function by giving it an array where we accumulates all the prefixes.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+
+export function prefixPseudoInCSSBlock(
+	path: PrefixCSSBlockCompilerPath,
+	namesToFeatures: Map<string, string>,
+) {
+	const targets = getTargets(path);
+
+	// COMMENT: same performance problem as above
+	// also, it's pretty duplicate. would it worth the effort to make it one function?
+	const newValue = [];
```
ematipico, line 366: We should type this guy. Inside `internal/ast/css/unions.ts` we can create a new type called `CSSBlockValue`:

```ts
export type CSSBlockValue =  Array<AnyCSSValue | CSSRule | CSSAtRule | CSSDeclaration>;
```

Use it inside `internal/ast/css/core/CSSBlock.ts` where we have the `value` property.

And use it to type it:

```ts
const newValue: CSSBlockValue = []
```

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +201,195 @@ export function prefixCSSValue(
...
+		if (prefixes.size === 0) {
+			continue;
+		}
+
+		for (const prefix of prefixes) {
+			newValue.push(prefixCSSRule(node, prefix, namesToFeatures, targets));
+		}
+	}
+
+	if (newValue.length !== path.node.value.length) {
```
ematipico, line 383: `newValue.length` will always be equals or greater than `path.node.value.length`. Perhaps we should narrow down the condition to `>`?

ematipico (changes requested):
Thank you for the PR! I left some comments and questions. I'm still not sure where we're heading though. I think this refactor makes sense but these are not utils anymore. I think we should start thinking to implement a compiler context for CSS and handle there the state. Not part of this PR of course, just putting my thoughts here.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
JustBeYou: To remember that I did not like the naming so I would change it later :joy: Also, I had the same feeling that those are not utils anymore. 
I think after making this feature work we could think to move the code away from `utils`? I could do a PR afterward

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
JustBeYou: `CSSRoot` contains multiple `CSSRules`. When prefixing we end up with newly generated rules. Is there any other way to add them to the root node than to replace it?

JustBeYou:
For the rest of the comments (where I did not reply), I will simply do the requested changes as it looks totally fine to me.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
ematipico: I think it's just sufficient to return `siglan.replace(newPrelude)` when you need to add new prefixes to a selector. That's what we do inside lint rules where we apply an autofix: we just tell to the compiler, using a signal, to "replace this specific node". The compiler does the rest.

JustBeYou: @ematipico Well, I'm not really sure it's that simple. If we replace only the prelude this would be an example result:
```
/* input */
.example:any-link {}

/* output */
.example:any-link:-moz-any-link:-webkit-any-link {}
```

And I think that the correct output should be (that's what autoprefixer does too):
```
.example:any-link {}
.example:-moz-any-link {}
.example:-webkit-any-link {}
```

Am I missing something?

ematipico: Ooooooh I see now! I completed missed that part! Yes it makes total sense now! Thank you for your answer

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +212,389 @@ export function prefixCSSValue(
...
+}
+
+function prefixCSSSelector(
+	{
+		selector,
+		prefix,
+		namesToFeatures,
+		targets,
+	}: PrefixCSSSelectorProps,
+) {
```
ematipico, line 520: Return type is missing

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/pseudo-selectors/input.css
```diff
@@ -0,0 +1,65 @@
+/* Single pseudo selectors */
```
ematipico, line 1: What do think if we moved every category of selectors in their own test file? The reason I am suggesting is because it's really difficult to keep track of what's changed and if that's correct. What I would suggest: 

```
- `pseudo-selectors/`
--- `single/`
------ `input.css
--- `multiple-classes/`
------ `input.css
--- `multiple-elements/`
------ `input.css`
```

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/pseudo-selectors/input.test.md
```diff
@@ -0,0 +1,240 @@
...
+
+.example6:any-link {
+	width: 10px;
+}
+.example6:-moz-any-link {
+	width: 10px;
+}
+.example6:-webkit-any-link {
+	width: 10px;
+}
```
ematipico, line 197: In this case we should prefix only once, ideally. 

Also, do you think it would be possible to change from:

```css
.example6:any-link {
	width: 10px;
}
.example6:-moz-any-link {
	width: 10px;
}
.example6:-webkit-any-link {
	width: 10px;
}
```
To

```css
.example6:any-link,
.example6:-moz-any-link,
.example6:-webkit-any-link {
	width: 10px;
}
```

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +212,389 @@ export function prefixCSSValue(
...
+
+/**
+ * Restrict the prefixes in the block to the prefixes
+ * present in the rule's prelude (if it's the case). 
+ */
+function purgePrefixes(prefixes: Set<string>, path: PrefixCSSBlockCompilerPath) {
+	if (isCSSRule(path.parent)) {
+		const allowedPrefixes = collectExistingPrefixesFromRule(path.parent);
+		if (allowedPrefixes.size > 0) {
+			return new Set([...prefixes].filter((x) => allowedPrefixes.has(x)));
```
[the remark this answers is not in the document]
JustBeYou: That's the proposed way of intersecting sets in MDN, but now looking closer I think it's just useless. I'll remove it.

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
```diff
@@ -228,6 +212,389 @@ export function prefixCSSValue(
...
+	prefixed: boolean;
+}
+
+/**
+ * Prefixes any rule with pseudo-selectors if it is the case.
+ * It detects if a rule already has a prefixed version or if it
+ * is prefixed itself and ignores it.
+ */
+function prefixRulesInArray(
+	array: AnyNode[],
```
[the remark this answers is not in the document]
JustBeYou: I tried to do this but TS compiler was not so cooperative. The thing is I am applying these rules to two kinds of arrays:
1. `(CSSRule | CSSAtRule)[]` for `CSSRoot`
2. `CSSBlockValue` which is a long union of types
I couldn't find a way to define a type like: `Array<thing that inherits AnyNode | CSSRule>` and work with both `CSSBlockValue` and the other. I saw something about conditional type inference, so I could extract the type of any array and extend it but it looked complicated.

Do you have any idea or should I look at the inference thing?

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/pseudo-selectors/input.test.md
JustBeYou: Not quite. If a rule is duplicated there's a big chance that the content itself is not duplicated and it's either a programming mistake or a choice. I've seen this before (it's not a good practice but some programmers still do this). I think we should focus on not having duplicated CSS rules in another feature (probably a rome lint rule), not here. What do you think?

JustBeYou:
The comments I did not respond to I will simply change.

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/pseudo-selectors/input.test.md
JustBeYou: About the style change: maybe it would be useful to do the same everywhere? I think this will simplify the code. Initially, I was thinking that throwing all the prefixed selectors in the same rule looks ugly but maybe it's not that bad. What do you think?

internal/compiler/transforms/compile/css-handler/prefix/utils.ts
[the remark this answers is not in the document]
JustBeYou: `prefixes` is a set. `[...prefixes]` is the short form of `Array.from(prefixes.values())` so I could use `.filter()`

internal/compiler/transforms/compile/test-fixtures/css-handler/prefix/pseudo-selectors/input.test.md
JustBeYou: Solved on discord.

ematipico (approved):

ematipico:
Amazing PR! Thank you very much, that was a huge contribution!

PR #1527, opened 2021-06-03
fix(compiler): correct return in getOptional

JustBeYou:
## Summary
`getOptional` was throwing an error instead of returning `undefined`.

## Test Plan
`./rome ci`

### DIFF
--- a/internal/compiler/lib/VisitorState.ts
+++ b/internal/compiler/lib/VisitorState.ts
@@ -72,10 +72,9 @@ export default class VisitorState<State extends UnknownObject>
 	public getOptional(find?: FindState<State>): undefined | State {
 		const index = this.getIndex(find);
 		if (index === -1) {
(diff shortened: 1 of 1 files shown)


ematipico (approved):

internal/compiler/lib/VisitorState.ts
```diff
@@ -72,10 +72,9 @@ export default class VisitorState<State extends UnknownObject>
 	public getOptional(find?: FindState<State>): undefined | State {
```
jamiebuilds, line 72: Looks like this isn't even used, @sebmck did you intend for this to stick around?

JustBeYou: @jamiebuilds I think I'll need it. There is no way to check if some state was already created or not. (without throwing)

PR #1530, opened 2021-06-03
fix: exit server process when daemon is stopped

### LINKED ISSUE #1528: 🐛 Stop command doesn't fully stop the rome daemon
## Environment

<details>

```
Environment Variables

User Config
  unset

Rome Version
  10.0.7-beta.18

Node Version
  14.17.0

Platform
  darwin x64 19.6.0

Terminal Features
  {
    background: "unknown"
    colorDepth: 24
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<150>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
  }
```
</details>

## Steps To Reproduce

1. `./rome start`
```
ℹ No running daemon found. Starting one...
✔ Connected to daemon
✔ Started daemon!
```
2. `./rome stop`
```
✔ Connected to daemon
✔ Stopped server.
```
3. `./rome status`
```
✔ Connected to daemon
```
The `rome-server` process is still running and listening to the socket so the client connects to it, but the Server has mostly ended so it can't respond to requests.

Looks like the `socketServer` doesn't get attached as a resource so it doesn't get released when the Server ends. Also, I think we need to add a `safeProcessExit` to the stop command.

## Expected Results

After stopping the daemon, running `./rome status` should result in `✖ Server not running.`

yassere:
## Summary

Closes #1528 

  - Adds the socketServer as a resource so that it gets cleaned up when the server ends. 
  - Adds a `safeProcessExit(0)` to the server stop command. (Is this the right approach?)
  - Prevents a started daemon from inheriting stdio from the client.

## Test Plan

Manually tested.

`./rome start`
```
ℹ No running daemon found. Starting one...
✔ Connected to daemon
✔ Started daemon!
```
`./rome stop`
```
✔ Connected to daemon
✔ Stopped server.
```
`./rome status`
```
✖ Server not running.
```

### DIFF
--- a/internal/core/server/commands/stop.ts
+++ b/internal/core/server/commands/stop.ts
@@ -9,6 +9,7 @@ import {ServerRequest} from "@internal/core";
 import {commandCategories} from "../../common/commands";
 import {createServerCommand} from "../commands";
 import {markup} from "@internal/markup";
+import {safeProcessExit} from "@internal/resources";
 
 export default createServerCommand({
 	category: commandCategories.PROCESS_MANAGEMENT,
@@ -20,5 +21,6 @@ export default createServerCommand({
 	},
 	async callback({server}: ServerRequest) {
 		await server.end();
+		await safeProcessExit(0);
 	},
 });
--- a/internal/resources/factories.ts
+++ b/internal/resources/factories.ts
@@ -142,7 +142,7 @@ export function createResourceFromServer(server: net.Server, name: string = "net
     finalize: () => {
       return new Promise((resolve, reject) => {
         server.close((err) => {
-          if (err === undefined) {
+          if (err != null) {
             reject(err);
           } else {
             resolve();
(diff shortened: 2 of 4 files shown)


internal/resources/factories.ts
```diff
@@ -142,7 +142,7 @@ export function createResourceFromServer(server: net.Server, name: string = "net
     finalize: () => {
       return new Promise((resolve, reject) => {
         server.close((err) => {
-          if (err === undefined) {
+          if (err !== undefined) {
```
sebmck, line 145: Actually shouldn't this be:

```suggestion
          if (err != null) {
```

sebmck (approved):
Looks good!

internal/core/server/commands/stop.ts
```diff
@@ -20,5 +21,6 @@ export default createServerCommand({
 	},
 	async callback({server}: ServerRequest) {
 		await server.end();
+		safeProcessExit(0);
```
jamiebuilds, line 24: Should this not be `await`'d?

PR #1536, opened 2021-06-07
feat(css-parser): `@import` 

ktfth:
## Summary

Implementation of css import

## Test Plan

./rome test


internal/css-parser/test-fixtures/import/input.test.md
```diff
@@ -0,0 +1,283 @@
...
+				CSSString {
+					value: "something.css"
+					loc: SourceLocation import/input.css 1:8-1:23
+				}
+			]
+			loc: SourceLocation import/input.css 1:0-1:23
+		}
+		CSSRule {
+			prelude: [
+				CSSSelector {
```
jamiebuilds, line 22: Why are all of these being parsed as `CSSSelector`s?

internal/css-parser/test-fixtures/import/input.css
```diff
@@ -0,0 +1,11 @@
...
+@import url("something.css");
+@import url("fineprint.css") print;
+@import url("bluish.css") projection, tv;
+@import url("narrow.css") handheld and (max-width: 400px);
+@import url("fallback-layout.css") supports(not (display: flex));
+@supports (display: flex) {
+	.ex {
+		width: 100px;
+	}
+}
```
jamiebuilds, line 11: Why the `@supports` in this file?

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,90 @@
+// https://www.w3.org/TR/css-cascade-4/#conditional-import
```
jamiebuilds, line 1: Looks like there are a bunch of missing imports in these files

jamiebuilds (changes requested):

internal/css-parser/test-fixtures/import/input.test.md
ktfth: I'm not founded the command to create tests and I have created manually

internal/css-parser/test-fixtures/import/input.css
ktfth: I'm gonna remove that

internal/css-parser/parser/at-import.ts
ktfth: Working on that

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+	onAtKeyword?: OnAtKeyword;
+	onAtDeclaration?: OnAtDeclaration;
+}
+
+export function parseAtImport(
+	{parser, onAtDeclaration, onAtKeyword}: ParseAtImport,
+): CSSAtImport {
+	const start = parser.getPosition();
+	const previousToken = parser.getToken() as Tokens["AtKeyword"];
+	const token = parser.expectToken("AtKeyword");
```
ematipico, line 29: We should not use `.expectToken` because it throws an hard error. If you need to check the existence of token, use `matchToken` and use `parser.unexpectedDiagnostic`

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+			break;
+		}
+		if (previousToken.value === "media") {
+			block = parseMediaList(parser);
+			break;
+		}
+		if (previousToken.value === "keyframes") {
+			block = parseKeyframe(parser);
+			break;
+		}
```
ematipico, line 51: Does `"keyframes"` exist inside the official spec? I didn't see it. If it's not supported, we should remove it

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		if (previousToken.value === "font-face") {
+			block = parseFontFace(parser);
+			break;
+		}
```
ematipico, line 56: `"font-face"` doesn't exist inside `@import`, we should remove it

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		if (previousToken.value === "page") {
+			block = parseAtPage(parser);
+			break;
+		}
```
ematipico, line 61: I don't think `"page"` is inside the `@import` rule. We should remove it.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (matchToken(parser, "LeftCurlyBracket")) {
+			block = parseDeclarationBlock({
+				parser,
+				parentAtKeywordToken: previousToken,
+				onAtDeclaration,
+				onAtKeyword,
+			});
+			break;
+		}
```
ematipico, line 77: There aren't any blocks inside `@import` syntax, we should remove it.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+
+interface ParseAtImport {
+	parser: CSSParser;
+	onAtKeyword?: OnAtKeyword;
+	onAtDeclaration?: OnAtDeclaration;
+}
+
+export function parseAtImport(
+	{parser, onAtDeclaration, onAtKeyword}: ParseAtImport,
+): CSSAtImport {
```
ematipico, line 26: This function should return `undefined` too.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+	const start = parser.getPosition();
+	const previousToken = parser.getToken() as Tokens["AtKeyword"];
+	const token = parser.expectToken("AtKeyword");
+	const prelude: AnyCSSValue[] = [];
+	const name = token.value;
+	let block = undefined;
+	while (true) {
```
ematipico, line 33: I don't think a loop is needed. The spec doesn't says there could be multiple values of the same token: there's only one "supports" or  one "url" or one "string":

```
@import [ <url> | <string> ]
        [ supports( [ <supports-condition> | <declaration> ] ) ]?
        <media-query-list>? ;
```

I'd suggest to read again the spec and try to understand every bit of it.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (matchToken(parser, "Semi")) {
+			break;
+		}
+		if (matchToken(parser, "EOF")) {
+			parser.unexpectedDiagnostic({
+				description: descriptions.CSS_PARSER.UNTERMINATED_AT_RULE,
+				token: parser.getToken(),
+			});
+			break;
```
ematipico, line 42: In case of errors we should exit from the function and return `undefined`. That's how we do it inside the rest of the functions.

ematipico (changes requested):
Thank you for your PR!

There are some things to review, I try to write all of them here:

- ~the PR is not ready yet, I would suggest move it to a "draft" state~ (I did myself)
- to properly start seeing what want to implement, create a file called `test.css` in the root of the project and put a simple CSS that you want to start parsing. Run the command `./rome parse test.css --tokenize` to see all the tokens that belong to the syntax inside your file;
- to start parsing the new rule, you need to understand **where** to start parsing the specific new syntax; the place is [here](https://github.com/rome/tools/blob/main/internal/css-parser/parser/rules.ts#L108-L132)

You should add a new `if` to check against the work `"import"`. Inside the new check, we can call you new function `parseAtImport` and start your new parsing.

- once you're happy with your implementation, you can run the command `./rome parse test.css` (without `--tokenize`). This command will print in command line the AST of your input file. Use this command to verify that your code is correct;

- for tests, use the command `./rome test ./internal/css-parser/index.test.ts`; add `--update-snapshots` when you need to update the snapshots of your tests. Inside `css-parser` folder there's a folder called `test-fixtures`. The way it works is that you want to create a folder with a file called `input.css` and add the cases you want to test. Run the `test` command and it will automatically create the snapshot test. No need to create it yourself (you can delete the one you created)

internal/css-parser/test-fixtures/import/input.test.md
```diff
@@ -0,0 +1,42 @@
...
+				message: RAW_MARKUP {value: "Expected a left curly bracket <emphasis>{</emphasis>."}
+			}
+			location: {
+				language: "css"
+				path: RelativePath<import/input.css>
+				end: Position 1:23
+				start: Position 1:8
+			}
+		}
+	]
```
jamiebuilds, line 38: This is all valid CSS code and shouldn't have any diagnostics

internal/ast/css/unions.ts
```diff
@@ -90,4 +92,7 @@ export type CSSAtRuleValue =
 	| CSSKeyframe
 	| CSSMediaQueryList
 	| CSSAtPage
-	| CSSFontFace;
+	| CSSFontFace
+	| CSSAtImport;
+
+export type CSSAtImportValue = string[] | CSSUrlFunction;
```
ematipico, line 98: The spec doesn't expect an array of strings. Probably the square bracket in the spec confused you. [Check this link](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#formal_syntax). If you click on the square bracket, it brings you to the explanation. 

It should be only a `string`

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,39 @@
...
+import {CSSAtImport, CSSAtImportValue} from "@internal/ast";
+import {matchToken, nextToken, readToken} from "@internal/css-parser/tokenizer";
+
+export function parseAtImport(parser: CSSParser): CSSAtImport | undefined {
+	const start = parser.getPosition();
+	let value: CSSAtImportValue;
+	while (matchToken(parser, "Whitespace")) {
+		readToken(parser, "Whitespace");
+	}
+	let token = parser.getToken() as Tokens["String"];
```
ematipico, line 12: You don't need to do a cast with `as`. It's not needed.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	if (token.type === "String") {
+		value = [token.value];
+		return parser.finishNode(
```
ematipico, line 16: Before calling `parser.finishNode`, you need to advance with the position, this means that you need to call `nextToken`:

```js
nextToken(parser);
return parser.finishNode();
```

This is needed so the parser can correctly determine the end location of the node.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,39 @@
...
+export function parseAtImport(parser: CSSParser): CSSAtImport | undefined {
+	const start = parser.getPosition();
+	let value: CSSAtImportValue;
+	while (matchToken(parser, "Whitespace")) {
+		readToken(parser, "Whitespace");
+	}
+	let token = parser.getToken() as Tokens["String"];
+
+	if (token.type === "String") {
+		value = [token.value];
```
ematipico, line 15: As pointed out, value needs to be a normal string.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,39 @@
...
+		return parser.finishNode(
+			start,
+			{
+				type: "CSSAtImport",
+				value,
+			},
+		);
+	}
+
+	return undefined;
```
ematipico, line 38: We can't return `undefined` without creating a diagnostic. If we arrive here it means that something went wrong. We need to tell the user why. Remember that these diagnostic will appear also on the VSCode plugin, which means we have to be careful about developer experience too.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,39 @@
...
+			{
+				type: "CSSAtImport",
+				value,
+			},
+		);
+	}
+
+	if (token.type === "Function") {
+		nextToken(parser);
+		token = parser.getToken() as Tokens["String"];
```
ematipico, line 27: Don't override a previous variable. Create a new one.

You can't use a cast here. It's really unsafe, why? Because you cheat the compiler and yourself to expect a token that probably would be wrong.

Take the following example:

```css
@import url(3);
```

`3` in this case is not a `"String"` token. This means that the following code should be like this:

```js
const functionArgumentToken = parser.getToken();

if (functionArgumentToken.type === "String") {
	nextToken(parser);
	return parser.finishNode();
}
parser.unexpectedDiagnostic({
	description: descriptions.CSS_PARSER.AT_IMPORT_INVALID_ARGUMENT, // we need to create this diagnostic,
	token: functionArgumentToken
});
return undefined;
```

internal/formatter/builders/css/import/CSSAtImport.ts
```diff
@@ -0,0 +1,6 @@
+import {CSSAtImport} from "@internal/ast";
+import {Builder, Token} from "@internal/formatter";
+
+export default function CSSAtImport(builder: Builder, node: CSSAtImport): Token {
+	return JSON.stringify(node.value);
```
ematipico, line 5: We have to create a test also for this. The way formatters tests are created and executed are the same of the CSS parser. You can follow the same process.

ematipico (changes requested):
The changes are better than before! Thank you! There's still some work to do though. I left some comments.

What I **strongly suggest** is to **test manually your changes before creating the tests**. Running `./rome parse test.css` with some code sample is enough to make sure that your code is correct or not.

internal/css-parser/parser/at-import.ts
ematipico: It should be `const token`

internal/formatter/builders/css/import/CSSAtImport.ts
ktfth: I have created the tests and running then they not passed with this error:

"Expected a left curly bracket {"

Can you help me understand that?

ematipico: It means there's an issue while while parsing the CSS you put inside the test fixture. You will be able to see more information if you try to parse the code using `./rome parse` command.

internal/css-parser/parser/at-import.ts
```diff
@@ -0,0 +1,51 @@
...
+	if (functionArgumentToken.type === "String") {
+		nextToken(parser);
+		value = functionArgumentToken.value;
+	}
+
+	const token = parser.getToken();
+
+	if (token.type === "Function") {
+		nextToken(parser);
+		value = token.value;
```
ematipico, line 24: This is not correct. In this case `value` should be a `CSSFunction`. Remember the type `CSSAtImport`. Its value is `export type CSSAtImportValue = string | CSSUrlFunction;`

This is the case where we need a function

```js
const functionStart = parser.getToken();
nextToken(parser);
value = parser.finishNode(functionStart, {
	type: "CSSUrlFunction",
	name: "url",
	params: [token.value]
});
```

internal/css-parser/test-fixtures/import/input.test.md
```diff
@@ -0,0 +1,35 @@
...
+		}
+		CSSAtRule {
+			name: "import"
+			prelude: []
+			block: CSSAtImport {
+				value: "url"
+				loc: SourceLocation import/input.css 2:7-2:29
+			}
+			loc: SourceLocation import/input.css 2:0-2:29
+		}
```
ematipico, line 27: @ktfth please check also this fixtures when you generate them. Did you notice that we lost information here? `"something.css"` is not there anymore.

ematipico (changes requested):

jamiebuilds:
FYI I've updated the base branch of this PR to a feature branch that other work can be merged into

ematipico:
Thank you for your contribution!

PR #1541, opened 2021-06-08
fix: match sourceMappingURL to bundler hard-coded output filename

yassere:
## Summary

Currently when bundling, the sourceMappingURL is incorrect because the bundle filename is hard-coded as 'index.js' but the sourceMappingURL is coming from the name of the entry file. This makes it harder for the debugger to find source maps.

As a temporary quick fix, is it reasonable to hard-code `index.js` here to match?

### DIFF
--- a/internal/core/server/bundler/BundleRequest.ts
+++ b/internal/core/server/bundler/BundleRequest.ts
@@ -75,7 +75,7 @@ export default class BundleRequest {
 		this.diagnostics.addAllowedUnusedSuppressionPrefix("lint");
 
 		this.sourceMap = new SourceMapGenerator({
-			path: createPath(this.resolvedEntry.getBasename()),
+			path: createPath("index.js"),
 		});
 
 		this.assets = new RelativePathMap();


jamiebuilds:
Is this the hardcoded name you mean?

https://github.com/rome/tools/blob/49ee439c3d44b05798a0b7a2833d1a7e6c9d3fd3/internal/core/server/bundler/Bundler.ts#L579

jamiebuilds:
Could `Bundler.bundle()` be refactored to calculate the paths first and pass them into BundleRequest? Or maybe go the inverse direction where BundleRequest returns them?

yassere:
There's a TODO here about using [basePath](https://github.com/rome/tools/blob/49ee439c3d44b05798a0b7a2833d1a7e6c9d3fd3/internal/core/server/bundler/BundleRequest.ts#L311-L315) for the sourceMappingURL
And a TODO here about [allowing it to be customized](https://github.com/rome/tools/blob/49ee439c3d44b05798a0b7a2833d1a7e6c9d3fd3/internal/core/server/ServerRequest.ts#L822-L823)

@sebmck Can you clarify what you had in mind there?

jamiebuilds (approved):
Happy to merge this as a temporary solution

PR #1544, opened 2021-06-09
feat(css-parser, formatter): min() and max() support

ematipico:
## Summary

Part of #984.

This PR implements the `min()` and `max()` functions. The formal syntax is [available on MDN website](https://developer.mozilla.org/en-US/docs/Web/CSS/min()#formal_syntax).

One thing to note is that these two functions, formally, can accept from one to n-`calc-sum` tokens. Which means that `min(500px)` is formally valid. Although, it doesn't make much sense to have just one argument. Maybe we could consider a lint rule here.

I also removed a constraint from the `calc()` function. I read the spec wrong and there was a diagnostic that wasn't necessary.

## Test Plan

New tests


internal/css-parser/parser/calculations.ts
```diff
@@ -234,10 +234,6 @@ function tryParseOperator(
 		const previousToken = parser.getPreviousToken();
 		const start = parser.getPosition();
 		if (!(op && (op.value === "+" || op.value === "-"))) {
-			parser.unexpectedDiagnostic({
```
ematipico, line 237: The operator is not needed here. Meaning that `calc(500px)` is formally a valid syntax. Doesn't make much sense passing a single argument, so we could consider a lint rule.

PR #1551, opened 2021-06-10
Init CodeQL analysis

jamiebuilds:
I'm not sure if this is worthwhile or not, but I was just going through repository settings and I figured I'd propose this.

For more info: https://github.com/github/codeql

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,44 @@
+name: "CodeQL"
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  schedule:
+    - cron: '35 14 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
ematipico, line 22: Should we build the project before? The entire project is in TypeScript. Does `javascript` cover TS too?

jamiebuilds: Yeah, they group javascript and typescript into one, it doesn't require building

ematipico:
It would be great to see something, for some reason the operation timed out.

jamiebuilds:
Yeah I'm not sure why its timing out. When you dig into the job it looks like it completed successfully

Issue #1556, opened 2021-06-10
📎 CSS Grid support


ematipico:
Supporting CSS grid requires a bit a work involved. There are different functions and properties. Also, it has a restricted grammar. 

Here's what I found so far:

- [x] [`repeat()`](https://csstree.github.io/docs/syntax/#Function:repeat)
- [x] [`minmax()`](https://csstree.github.io/docs/syntax/#Type:minmax())
- [x] [`fit-content()`](https://csstree.github.io/docs/syntax/#Function:fit-content)
- [ ] [`grid`](https://csstree.github.io/docs/syntax/#Property:grid)

Refers to #984

ktfth:
That task still available to work on? What is the steps to make a better contribution, what is need to see to made the work?

ematipico:
@ktfth you could try with `fit-content`. The spec is quite simple.

jamiebuilds:
Hi! Thank you for opening this issue

We recently made the decision to [migrate the Rome codebase to Rust](https://rome.tools/blog/2021/09/21/rome-will-be-rewritten-in-rust) and making some significant changes to its architecture. We've also decided not to release the previous version written in TypeScript.

We're going through and closing pull requests and issues that were specific to the TypeScript codebase, including this one. Thank you for taking the time to open it.

Thanks again!

PR #1560, opened 2021-06-11
 feat(css-parser, formatter): minmax() support

ematipico:
## Summary

Part of #1556 , it implements the `minmax()` function.

I also made a small refactor. We were using the code to skip white spaces almost everywhere. So I created a function to do that.

## Test Plan

New tests


internal/formatter/test-fixtures/css/grid/minmax/input.css
```diff
@@ -0,0 +1,3 @@
+.style {
+	grid-template-columns: minmax(    1fr     ,     auto      );
```
jeddy3, line 2: You may want to swap these parameters around in this example as I believe [`<flex>` can only be used in the 2nd parameter](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax()#formal_syntax), e.g.:

```css
grid-template-columns: minmax(    auto     ,     1fr     );
```

ematipico: Shoot! I didn't see it! Thank you! I will fix the code too

PR #1567, opened 2021-06-15
feat(lint, project): new lint configuration

### LINKED ISSUE #1529: 📎 Consider disabling all lint rules by default

ematipico:
## Summary

This PR closes #1529 

### Strict naming convention with new visitor

I created a new function called `createLintVisitor`. This function is simply a small wrapper around `createVisitor`. The wrapper is more typed and forces `name` to follow the name of the rule. 

Also, this function will be used to mark a lint rule `recommended`. In the future we could use this function to provided metadata too: description, examples, etc. useful for documentation.

### Changed how rules are stored

Previously the rules (visitors) where in stored inside a `Map` and mapped like:

```md
"js/someRule" => visitor
```

Considering the new configuration structure, we also need to rethink how we store these rules. So I changed the hierarchy of the rules:

```md
"js" => {
	"someRule" => { visitor, recommended, meta }
	"someOtherRule" => { visitor, recommended, meta }
}
"a11y" => {
	// etc
}
```

### Map of visitors separated from rules types/definitions

The reason why I created the file `categories.ts` (I will try to find a better name) was because importing some types from `index.ts` was causing some circular dependency. Having those types in a different file fixed the issue

### Disabled rules

I removed the configuration `lint.disabledRules`. It's not needed anymore as we will achieve the same result in a different way. 

## Test Plan

To create more tests


jamiebuilds:

Thanks again!

Issue #1555, opened 2021-06-10
📎 Support for path aliases 


ematipico:
There have been some discussions already in some other issue.

Some code bases take advantage of path aliases to handle how they import their modules.
```js
import Sidebar from "@components/sidebar";
```

TypeScript supports them but it doesn't transform them. I think Rome should support them too - and ultimately transform/explode them during the bundling. Path aliases will be beneficial for the linter and the bundler (and testing as direct consequence).  

We could use the command `rome auto-config` to "import" the aliases defined inside a `tsconfig.json` file. We could add a new configuration called `aliases` and have a structure like:

```json
"aliases": {
	"@components": "src/components",
	"~layouts": "src/pages/layouts"
}
```

**The name `aliases` is purely conventional and we can change it. If you have a better suggestion please leave a comment.**

The paths defined on the right-side (e.g. `"src/components"`) would start from the "root" of the project, where the configuration file is defined.

Is there anything that I missed?

JustBeYou:
`tsconfig.json` supports multiple locations for the same alias:
```json
"paths": {
     "@components": ["./components", ... other paths ...]
}
```
I think we should support the same syntax if we want it to be fully compatible.

ematipico:
@JustBeYou as you expressed your interest working on this, please consider also [this comment](https://github.com/rome/tools/issues/1137#issuecomment-677910513).

The support should be explicit and Rome should not rely on third party API/libs for features.

JustBeYou:
Sure, I know that the philosophy here is zero dependencies. I think the fix would go somewhere in the resolver, probably somewhere around `resolveLocal`.

jamiebuilds:
One question we should answer is if `tsconfig.json` paths should apply to more than just JavaScript/TypeScript.

```json
{ "paths": { "@components/*": ["./src/components/*"] } }
```

```css
@import "@components/button/styles.css"
```

etc

That might make sense at first, but consider the use case of Rome where someone isn't using TypeScript or even JavaScript.

jamiebuilds:
An alternative is to make the `paths` in `tsconfig.json` "managed" by Rome, as in automatically keep it in sync or yell if they disagree. There's already a number of ways I'd like to see tsconfigs managed by Rome anyways

I like how TypeScript uses a glob-like matching system over the regex stuff webpack does. But we also may be able to do a better job.

robpalme:
FYI this feature may overlap with Node's package imports which provides a more constrained form of remapping module specifiers with a package.

[...]

PR #1569, opened 2021-06-15
feat(resolver, project, core): support for path aliases

JustBeYou:
## Summary
Fix for #1555 . I've implemented "naive" support for aliases. It allows only one wildcard at the end of the pattern (eg. `some/path/*`). Probably it would be nice to support almost any glob-like expression but I think the effort and performance impact is not worth it. Also, in Typescript's documentation, there are no examples of a more complex expression. 

I'm waiting for a review of this draft. If it is ok, I think we could add some caching and improve where the code is placed. After that I will also implement the `auto-configuration`.

## Test Plan
I added some integration tests for the resolver, enabled them temporarily and tested the new feature.  I've also tested it on a local project.

### DIFF
--- a/internal/compiler/lib/CompilerPath.ts
+++ b/internal/compiler/lib/CompilerPath.ts
@@ -482,6 +482,7 @@ export default class CompilerPath {
 					loc: inheritLoc(old),
 				};
 			}
+			fixed = {...fixed, value};
 		}
 
 		return fixed;
--- a/internal/compiler/lint/rules/js/preferOptionalChaining.test.md
+++ b/internal/compiler/lint/rules/js/preferOptionalChaining.test.md
@@ -338,7 +338,6 @@ foo?.bar.baz;
 
 ```ts
 let foo = {};
-
 foo?.bar?.("baz")
 
 ```


JustBeYou:
Note: When resolving aliased paths we need to set `origin` to the root path of the project, otherwise it won't work as expected. I couldn't yet find a way to get the root of the project btw.

jamiebuilds:
Here's the actual documented code for how TypeScript's `"paths"` patterns work: https://github.com/microsoft/TypeScript/blob/6bb1f0792ba19b662ac0384ad1db824c5de43bd3/src/compiler/moduleNameResolver.ts#L880-L939

internal/project/load.ts
```diff
@@ -720,5 +738,8 @@ function mergePartialConfig<
 				...b.integrations.prettier,
 			},
 		},
+		// TODO: how should we correctly handle this?
+		base: b.base,
```
ematipico, line 742: I would create an object called `aliases`, and there we can have `paths` and `base`.

Doing so, would allow us to correctly doing a merge like:

```js
aliases: {
	...a.aliases,
	...b.aliases
}
```

And we are aligned with the rest of the configurations

internal/project/load.ts
```diff
@@ -720,5 +738,8 @@ function mergePartialConfig<
...
+		paths: [...a.paths, ...b.paths],
```
ematipico, line 743: To merge two arrays, you can use the function `mergeArrays`. Although don't do it here, apply the merge strategy inside the function `extendProjectConfig`

ematipico: I think the merging strategy should be more accurate.

Let's say we have something like:

```
"@components" => ["./a" "./b"],
"@components" => ["./a", "./c"]
```

It should result in:

```
"@components" => ["./a" "./b", "./c"],
```

If we just spread, we lose at least one mapping ("./b" or "./c" depending on the order of the merging)

internal/project/utils.ts
```diff
@@ -110,3 +110,43 @@ export function getParentConfigDependencies(
 
 	return deps;
 }
+
+export function parsePathAliasPattern(path: string): PathAliasPattern {
+	if (path.length === 0) {
+		throw new Error("Empty path aliases are not allowed.");
```
ematipico, line 116: Errors should be throw in a different way. In order to do that, other than passing `path`, pass also the `Consumer`.

When you have an error, you can do:

```js
throw consumer.unexpected();
```

The consumer is returned by the `.asMap` method, which was `targets`.

Alternatively, you can wrap your function like this:

```js
consumer.handleThrownDiagnostics(() => {
	parsePathAliasPattern();
})
```

I would go for the first solution though, because you will need to throw `DiagnosticError` anyway, which is something that `consumer.unexpected` already does.

As thumb rule, never throw a `Error` inside Rome unless we are inside some internals (the binary transport for example)

internal/project/utils.ts
```diff
@@ -110,3 +110,43 @@ export function getParentConfigDependencies(
...
+	}
+
+	const wildcardCount = (path.match(/\*/g) || []).length;
+	if (wildcardCount > 1) {
+		throw new Error(`Too many wildcards in ${path}`);
```
ematipico, line 121: When creating a diagnostic, as pointed out in the previous comment, let's try to advice the developer with possible solutions. In this way, we improve also the developer experience. We should always offer a way to fix bugs.

internal/project/utils.ts
```diff
@@ -110,3 +110,43 @@ export function getParentConfigDependencies(
...
+		wildcard: wildcardCount === 1,
+	};
+}
+
+export function matchAliasPattern(
+	pattern: PathAliasPattern,
+	path: string,
+): string | undefined {
+	if (!pattern.wildcard) {
+		return path === pattern.parts[0] ? path : undefined;
```
ematipico, line 135: Should we save `pattern.parts[0]` and `pattern.parts[1]` in some variable? We avoid accessing to those vars too many times.

internal/core/server/fs/Resolver.ts
```diff
@@ -517,14 +523,63 @@ export default class Resolver {
...
+			// But only if we can find a loaded project
+			if (query.origin.isAbsolute()) {
+				const project = this.server.projectManager.findLoadedProject(
+					query.origin.assertAbsolute(),
+				);
+				if (project === undefined) {
+					return resolved;
+				}
+
+				for (const aliasPath of this.getAliasedPaths(query, project.config)) {
```
ematipico, line 545: Here we could only pass `aliases` instead of the whole project configuration. Or, directly `project.config.aliases.base`.

ematipico (changes requested):

internal/project/utils.ts
JustBeYou: I did not know how to handle diagnostics and I was expecting that someone will point me in the right direction. Thank you!

internal/project/load.ts
```diff
@@ -468,6 +470,29 @@ export async function normalizeProjectConfig(
...
+				config.aliases.paths.push([
+					parsePathAliasPattern(alias, aliases),
+					targets.asMappedArray((item) =>
+						parsePathAliasPattern(item.asString(), item)
+					),
+				]);
+			}
+		}
+
+		aliases.enforceUsedProperties("aliases config property");
```
JustBeYou, line 493: I don't know if this is necessary or not but I see it is used everywhere. If it should be removed please tell me.
Also, it would be nice if you could tell me what's this function used for :)

ematipico: This function should be used yes. And it makes sure that the properties that belong to a mapped object are actually used.

For example, in your case, we have `paths` and `base` inside `aliases`. If we use `asMap` and decided to use only `base` - and ignore `paths`, this function will throw an error saying that `paths` hasn't been used.

It makes sure that all the configuration is actually parsed, used and passed to Rome's internals.

JustBeYou:
I've been thinking about how should `auto-config` handle path aliases but I couldn't really decide on something. I have the following options in my mind:
1. Check if there are custom aliases configured, if yes check if there is any `tsconfig.json`. If it does not exists, suggest creating it. If it exists and the aliases differ from Rome's configuration, suggest changing it.
2. Add a proper way of handling `tsconfig.json` that would allow auto-fixes. Something similar to what Rome is already doing for `package.json` (I see that it is already managed)

Personally, I think implementing 1 as a dirty fix or leaving `auto-config` out of the discussion, for now, would be ok. I think implementing 2 should be done in a different PR. 

@ematipico @jamiebuilds I'd be glad to know what you two think.

ematipico:
Honestly, I wouldn't put too much work around `tsconfig.json`. It shouldn't be put at the same level of `package.json`.

What I would expect from `auto-config` is the following:
1. check existence of `tsconfig.json`;
2. if not present, bail without saying nothing to the user; if present, confirm with the user that following "a", "b", etc. aliases will be imported. (if they exist) 

Other properties will be handled via official integration (like prettier and eslint).

JustBeYou:
I've added autoconfiguration. The only problem is that if I try to answer "yes" to both questions below Rome hangs. I couldn't find the cause, maybe you have a suggestion about how should I debug this.

JustBeYou:
Also, `auto-config` does not display any diagnostic if it fails. It just prints "something went wrong" or nothing at all. Which would be the right way to display them?

I was thinking about 
```typescript
if (result.type === "DIAGNOSTICS" && result.hasDiagnostics) {
  throw new DiagnosticsError("Couldn't auto-configure", result.diagnostics)
}
```
This is how the output would look like:

ematipico:
> Also, `auto-config` does not display any diagnostic if it fails. It just prints "something went wrong" or nothing at all. Which would be the right way to display them?
> 
> I was thinking about
> 
> ```ts
> if (result.type === "DIAGNOSTICS" && result.hasDiagnostics) {
>   throw new DiagnosticsError("Couldn't auto-configure", result.diagnostics)
> }
> ```
> 
> This is how the output would look like:
> 

That's how we are doing so far. You could create a new object under `INIT_COMMAND` and have all the logs, advises, etc. that you want.  

```js
throw createSingleDiagnosticsError({
	location: req.getDiagnosticLocationForClientCwd(),
	description: descriptions.INIT_COMMAND.UNCOMMITTED_CHANGES,
});
```

> I've added autoconfiguration. The only problem is that if I try to answer "yes" to both questions below Rome hangs. I couldn't find the cause, maybe you have a suggestion about how should I debug this.

About this one, I've had a look at your new changes so far. Without checking out your branch I don't really know what could be wrong. Does it occur even with just one answer? (try to comment the second one). 

I will look into it later in the afternoon.

JustBeYou:
> That's how we are doing so far. You could create a new object under INIT_COMMAND and have all the logs, advises, etc. that you want.
That's how diagnostics are created in the server command. But they should also be thrown on the local command, otherwise, we don't receive any output. At the moment, if you have uncommited files you won't receive any useful information about why the command fails. See https://github.com/rome/tools/blob/a88c94a92ffc4283eb72df963d06f6ba688d4fb7/internal/core/client/commands/autoConfig.ts#L55
That's why I was suggesting rethrowing. 

> About this one, I've had a look at your new changes so far. Without checking out your branch I don't really know what could be wrong. Does it occur even with just one answer? (try to comment the second one).
It does not occur with just one answer. Only if both are present and I try to answer "yes" to both of them.

JustBeYou:
This feature is done, but I think it should be merged after #1588 is unconfirmed or fixed and #1587 is merged.

ematipico:
> That's how diagnostics are created in the server command. But they should also be thrown on the local command, otherwise, we don't receive any output.

It's definitely a regression. I confirmed that with Sebastian.

PR #1573, opened 2021-06-17
fix(lsp): publish diagnostics from LSP server

### LINKED ISSUE #1540: 🐛 VS Code extension not reporting diagnostics
## Environment

<details>

```
Environment Variables

User Config
  unset

Rome Version
  10.0.7-beta.18

Node Version
  14.17.0

Platform
  darwin x64 19.6.0

Terminal Features
  {
    background: "unknown"
    colorDepth: 24
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<145>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
  }

Server Status
  {
    projects: [{id: 0}, {id: 1}]
    workers: [{astCacheSize: 0, heapTotal: 46_948_352, ownedBytes: 0n, ownedFileCount: 0, pid: 51_085, uptime: 0.576717229}]
    server: {heapTotal: 46_948_352, pid: 51_085, uptime: 0.576931203}
  }
```
</details>

## Steps To Reproduce

Run `./rome run scripts/run-vscode-extension.ts .`

When VS Code is done loading, the "Rome" channel in the Output tab shows that the path to the rome-dev build was discovered. It also logs when files are opened and closed. And "Format Document" still works.
But it doesn't report any diagnostics in the Problems tab.

## Expected Results

Rome should report diagnostics for code that has lint errors.

yassere:
## Summary

Closes #1540 

There are still problems with the LSP server that I’ll follow up on in separate issues, but this addresses three bugs related to publishing diagnostics.

1. The LSP server only published diagnostics for AbsoluteFilePaths but was receiving UIDPath instances.
2. The LSP server was attempting to mutate an immutable DiagnosticsProcessor.
3. And then the tricky one:

The CheckRunner was adding diagnostics to its processor keyed by UIDPath. When a file changes, it should remove old diagnostics for that path because they’re about to be recomputed. But it was attempting that with an AbsoluteFilePath, which didn’t remove the UIDPath-keyed diagnostics. So diagnostics were accumulating for the lifetime of the CheckRunner, and if you fixed a problem in an open file in VS Code, the diagnostic wouldn’t go away.

A quick fix could have been to attempt to remove old diagnostics using both an AbsoluteFilePath and its converted UIDPath, but there shouldn’t have been any UIDPath-keyed diagnostics on the processor at all. When adding diagnostics to a processor, they’re first mapped with a DiagnosticsNormalizer which can potentially turn UIDPaths into AbsoluteFilePaths, but it doesn’t have that ability built-in. Instead, its `normalizePath` method calls the `normalizePosition` function from its `markupOptions`, and it gets those `markupOptions` based on how the DiagnosticsProcessor is created.

 In the case of the CheckRunner, it uses the `createDiagnosticsProcessor` method on ServerRequest which passes in the `markupOptions` from its Reporter. The ServerRequest reporter is (usually) forked from the ServerClient reporter which in turn gets its `markupOptions` from the Logger instance on the Server. And that has a `normalizePosition` function that turns UIDPaths into AbsoluteFilePaths.

What was happening here is that the LSPServer creates server requests for its lint sessions using the `silent: true` option which causes the ServerRequest to create a new Reporter instead of forking the reporter from the ServerClient. And that new Reporter didn’t have any `markupOptions` and so the DiagnosticsNormalizer wasn’t able to work as intended.

### DIFF
--- a/internal/core/server/ServerRequest.ts
+++ b/internal/core/server/ServerRequest.ts
@@ -264,7 +264,10 @@ export default class ServerRequest {
 		);
 
 		this.reporter = query.silent
-			? new Reporter("ServerRequestSilent")
+			? new Reporter(
+					"ServerRequestSilent",
+					{markupOptions: client.reporter.markupOptions},
+				)
 			: client.reporter.fork();
 		this.resources.add(this.reporter);
 
--- a/internal/core/server/lsp/LSPServer.ts
+++ b/internal/core/server/lsp/LSPServer.ts
@@ -108,6 +108,7 @@ export default class LSPServer {
 	private createDiagnosticsProcessor(): DiagnosticsProcessor {
 		// We want to filter pendingFixes because we'll autoformat the file on save if necessary and it's just noise
 		const processor = this.request.createDiagnosticsProcessor({
+			mutable: true,
 			filter: undefined,
 		});
 		processor.addEliminationFilter({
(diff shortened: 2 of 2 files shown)


ematipico (approved):
Looks good to me! Thank you for the detailed explanation!

Issue #1548, opened 2021-06-09
📎  Replace `Zero/OneIndexed` with clearer data types


jamiebuilds:
Instead of more abstract types like `ZeroIndexed` and `OneIndexed` we should have data types like `Cursor`, `Column`, `Line`, `Position`, and `Location` which are clearer and can enforce things like _"You can't add a `Cursor` to a `Column`"_

This will require some updates to the binary transport format

jamiebuilds:
Hi! Thank you for opening this issue

We recently made the decision to [migrate the Rome codebase to Rust](https://rome.tools/blog/2021/09/21/rome-will-be-rewritten-in-rust) and making some significant changes to its architecture. We've also decided not to release the previous version written in TypeScript.

We're going through and closing pull requests and issues that were specific to the TypeScript codebase, including this one. Thank you for taking the time to open it.

Thanks again!

PR #1582, opened 2021-06-21
fix: regressions in decisions and review command

### LINKED ISSUE #1579: 🐛 Regressions in `check --review` command
## Environment

<details>

```
Environment Variables

User Config
  unset

Rome Version
  10.0.7-beta.18

Node Version
  14.17.0

Platform
  darwin x64 19.6.0

Terminal Features
  {
    background: "unknown"
    colorDepth: 24
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<147>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
  }

Server Status
  {
    projects: [{id: 0}, {id: 1}]
    workers: [{astCacheSize: 0, heapTotal: 46_845_952, ownedBytes: 0n, ownedFileCount: 0, pid: 57_884, uptime: 0.570365222}]
    server: {heapTotal: 46_845_952, pid: 57_884, uptime: 0.570518524}
  }

```
</details>

## Steps To Reproduce

1. Create a file `test.ts` containing:
```js
const foo = true;
if (foo == false) {
	console.log("Impossible!");
}
```
2. Running `rome check test.ts --review` briefly scrolls past a diagnostic then clears the terminal and displays:
```
 Reviewing diagnostics (1/1) 

❯ How do you want to resolve this?
ℹ Use arrow keys and then enter to select an option
  ◉ Add suppression comment (shortcut s)
  ◯ Do nothing (shortcut n)
  ◯ Exit (shortcut escape)
  ◯ More options... (shortcut m)
```
3. Selecting "Add suppression comment" results in:
```js
const foo = true;
// rome-ignore lint/js/noDoubleEquals: suppressed via --review
if (
	// rome-ignore lint/js/noDoubleEquals: suppressed via --review
	foo ==
	// rome-ignore lint/js/noDoubleEquals: suppressed via --review
	false
) // rome-ignore lint/js/noDoubleEquals: suppressed via --review
{
	console.log("Impossible!");
}
```

## Expected Results

  - The diagnostic should still be visible while the user is making a choice.
  - There should be an option to apply the suggested auto-fix for `lint/js/noDoubleEquals`
  - Selecting the option to add a suppression comment should only add one suppression comment.

yassere:
## Summary

Closes #1579 

- Include `verboseAdvice` options when reviewing diagnostics
- Display the diagnostic that the user is being asked to review
- Fix `less` option so that it toggles back to original options
- Prevent injecting multiple duplicate suppression comments

### DIFF
--- a/internal/numbers/indexed.ts
+++ b/internal/numbers/indexed.ts
@@ -86,7 +86,7 @@ enhanceNodeInspectClass(
 export class IndexedNumberSet<Indexed extends IndexedNumber>
 	extends MappedSet<Indexed, number> {
 	constructor(entries?: Iterable<Indexed>) {
-		super((line) => [line.valueOf(), line], entries);
+		super((num) => [num.valueOf(), num], entries);
 	}
 }
 
--- a/internal/numbers/indexed.ts
+++ b/internal/numbers/indexed.ts
@@ -1,3 +1,4 @@
+import {MappedSet} from "@internal/collections";
 import {enhanceNodeInspectClass} from "@internal/node";
 import {isObject} from "@internal/typescript-helpers";
 
@@ -81,6 +82,15 @@ enhanceNodeInspectClass(
 		return `OneIndexedNumber<${inst.valueOf()}>`;
 	},
 );
+
+export class IndexedNumberSet<Indexed extends IndexedNumber>
+	extends MappedSet<Indexed, number> {
+	constructor(entries?: Iterable<Indexed>) {
+		super((line) => [line.valueOf(), line], entries);
+	}
+}
+
+IndexedNumberSet.prototype[Symbol.toStringTag] = "IndexedNumberSet";
 
 export type IndexedNumber = OneIndexed | ZeroIndexed;
 
--- a/internal/compiler/lib/CompilerPath.ts
+++ b/internal/compiler/lib/CompilerPath.ts
@@ -446,6 +446,7 @@ export default class CompilerPath {
 			{
 				...description,
 				advice,
+				verboseAdvice,
 			},
 			{
 				...diag,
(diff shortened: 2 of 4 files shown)


ematipico (approved):

internal/numbers/indexed.ts
```diff
@@ -82,6 +83,15 @@ enhanceNodeInspectClass(
...
 );
 
+export class IndexedNumberSet<Indexed extends IndexedNumber>
+	extends MappedSet<Indexed, number> {
+	constructor(entries?: Iterable<Indexed>) {
+		super((line) => [line.valueOf(), line], entries);
+	}
+}
+
+IndexedNumberSet.prototype[Symbol.toStringTag] = "IndexedNumberSet";
```
jamiebuilds, line 93: This seems like a very one-off data structure. Do you think it would be bad just to use `MappedSet` directly above?

```js
const visitedLines = new MappedSet<OneIndexed, number>();
```

yassere: I don't think it's bad, but it would have to be constructed like this:
```js
const visitedLines = new MappedSet<OneIndexed, number>((line) => [
	line.valueOf(),
	line,
]);
```

My thought was that there might be other situations where we want to track a set of lines or positions and we wouldn't have to duplicate the serialize function in that case.

If this is just going to be a one-off, would it be simpler to make it a `Set<number>` and just add/check membership as `line.valueOf()`?

yassere: Although I realize now that I shouldn't have named the parameter `line` in the serialize function in `IndexedNumberSet`. I'd originally implemented this by using `MappedSet` directly and then I just copied over the serializer when I extracted it.

yassere: I'll merge this for now (with the renamed parameter). I think we can revisit this once the `IndexedNumber` abstraction is changed for #1548

PR #1589, opened 2021-06-24
feat(parser-core, toml): tokenize abstraction and TOML parser

ematipico:
## Summary

I cherry-picked some commits from Sebastian's PR #1561

This PR implements:

- new tokenizer abstraction; when creating a new parser, the `tokenize` method now provides a `tokenizer` object that provides a set of useful API to tokenize the input;
- implements a new TOML parser using the new `tokenizer`;
- removes RJSON from the code base and uses TOML for testing lint rules;
- some code cleanup;

## Test Plan

New tests and updated existing ones.


internal/parser-core/TokenizerCore.ts
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	public consume(str: string): boolean {
+		if (this.startsWith(str)) {
+			this.setIndex(this.index.add(str.length));
+			return true;
+		} else {
+			return false;
+		}
+	}
```
yassere, line 159: The `eat` and `consume` methods seem to be used interchangeably throughout the code. Are they both necessary?

ematipico: They are slightly different. `consume` return a `boolean` while `eat` return the string that has been passed (if it exists). This API is equals to `parser.eatToken`, it behaves in the same way.

yassere: Right, I just meant that most (all?) current uses of `tokenizer.eat` appear to be using it as a truthy check or without using the return value. It's not really a problem. Just thought it might cause a bit of confusion.

ematipico: Good point. I make the change

ematipico: @yassere done!

ematipico:
I am going to merge this and update the documentation.

PR #1597, opened 2021-06-28
fix(core): programmatic is false by default

### LINKED ISSUE #1590: 🐛 Auto-config won't show diagnostics
For example, if you have uncommitted files, running `rome auto-config` should fail and output proper diagnostics. At the moment it simply exists without reporting any error. It should be investigated if it is strictly related to `auto-config` or if the problem occurs on other local commands too. 

Related PR #1587 

## Environment

not relevant

## Steps To Reproduce

Run `rome auto-config` on an already initialized project where you have some uncommitted files.

## Expected Results

Should display diagnostics.

JustBeYou:
## Summary
Fix #1590


ematipico (approved):

ematipico:
Thank you!

Issue #1558, opened 2021-06-10
🐛 LSP Server sends notifications before InitializeResult


yassere:
## Steps To Reproduce

Send an `initialize` request to `rome lsp`

The LSP server sends `$/progress` notifications before responding to the `initialize` request.

## Expected Results

According to the spec, "the server is not allowed to send any requests or notifications to the client until it has responded with an InitializeResult" (with some exceptions that don't apply here).

PR #1600, opened 2021-06-29
fix(lsp): improve LSP message handling

### LINKED ISSUE #1578: 🐛 LSP code actions incorrectly labeled and executed
Edit: The duplicate suppression comments aren't limited to LSP: #1579 

## Environment

<details>

```
Environment Variables

User Config
  unset

Rome Version
  10.0.7-beta.18

Node Version
  14.17.0

Platform
  darwin x64 19.6.0

Terminal Features
  {
    background: "unknown"
    colorDepth: 24
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<146>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
    cwd: AbsoluteFilePath</Users/Yasser/Dev/lsp-test>
    realCwd: AbsoluteFilePath</Users/Yasser/Dev/lsp-test>
  }

Server Status
  {
    projects: [{id: 0}, {id: 1}]
    workers: [{astCacheSize: 0, heapTotal: 46_579_712, ownedBytes: 0n, ownedFileCount: 0, pid: 54_039, uptime: 0.561771282}]
    server: {heapTotal: 46_579_712, pid: 54_039, uptime: 0.562005167}
  }
```
</details>

## Steps To Reproduce

1. While using the VS Code extension, open a file containing:
```js
const foo = "test";
```
2. Clicking on `foo` and then the Code Action lightbulb shows:
```
[object Object]: lint/js/noUnusedVariables(foo)
```
3. Executing that code action results in:
```js
// rome-ignore lint/js/noUnusedVariables(foo): suppressed via --review
const // rome-ignore lint/js/noUnusedVariables(foo): suppressed via --review
foo = // rome-ignore lint/js/noUnusedVariables(foo): suppressed via --review
"test";
```

## Expected Results

The code action should be labeled: `Add suppression comment: lint/js/noUnusedVariables(foo)`
Executing it should only add a single suppression comment and it should say: `suppressed via editor`

yassere:
## Summary

This PR addresses a handful of issues related to the LSP Server. There are still some bugs and stability issues affecting LSP that I’ll continue to work though.

- Add `exit` notification handler to avoid accumulating `rome-cli` processes.
- Prevent sending messages from server until receiving `initialized` notification
- Partially fix handling of `workspace/didChangeWorkspaceFolders` notification. TODO: Bugs related to changing workspace folders. Will track separately.
- Allow tearing down project sessions if workspace changes while initial check watch is being performed.
- Fix labels on code actions and inserted suppression comments.
- Disable `$/progress` notifications for now because they have a significant performance impact. TODO: Fix this.
- Extract LSP message handlers into separate folder. TODO: Clean this up more. Maybe extract workspace commands?

Closes #1578 #1558

### DIFF
--- a/internal/core/server/lsp/LSPServer.ts
+++ b/internal/core/server/lsp/LSPServer.ts
@@ -5,22 +5,18 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {Consumer} from "@internal/consume";
 import Server from "../Server";
 import ServerClient from "../ServerClient";
 import {
 	AbsoluteFilePath,
 	AbsoluteFilePathMap,
 	AbsoluteFilePathSet,
-	createAbsoluteFilePath,
 } from "@internal/path";
 import {
 	DIAGNOSTIC_CATEGORIES,
 	Diagnostic,
 	DiagnosticsProcessor,
-	catchDiagnostics,
 	formatCategoryDescription,
-	getActionAdviceFromDiagnostic,
 } from "@internal/diagnostics";
 import {
 	PartialServerQueryRequest,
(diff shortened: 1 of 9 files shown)


internal/core/server/lsp/messages/index.ts
```diff
@@ -0,0 +1,23 @@
+import {LSPNotificationHandler, LSPRequestHandler} from "./types";
```
ematipico, line 1: I really like this refactor! It makes things easier to understand!

internal/core/server/lsp/messages/workspace.ts
```diff
@@ -0,0 +1,88 @@
...
+export const executeCommand: LSPRequestHandler = async (lsp, params) => {
+	const command = params.get("command").asString();
+	const filename = params.get("arguments").getIndex(0).asString();
+
+	const path = createAbsoluteFilePath(filename);
+	const startVersion = lsp.fileVersions.get(path);
+
+	let req: PartialServerQueryRequest | undefined;
+
+	if (command === "rome.check.apply") {
```
ematipico, line 29: Should we store these strings (`"rome.check.apply"`, `"rome.check.decisions"`, etc.) in some variables?

internal/core/server/lsp/LSPServer.ts
```diff
@@ -92,16 +78,15 @@ export default class LSPServer {
 	}
 
 	public transport: LSPTransport;
+	public request: ServerRequest;
+	public client: ServerClient;
+	public server: Server;
```
ematipico, line 83: I am not really sure we should expose these. Maybe we can have some getters or `asserRequest`, etc. Or making specific functions to expose what's needed.

internal/core/server/lsp/messages/workspace.ts
yassere: These are the commands I was also thinking of extracting, in which case there would be something like a `LSPCommandHandlers` map instead of these conditionals. Either way, I do plan to clean this up later, yeah.

internal/core/server/lsp/LSPServer.ts
yassere: I've definitely exposed some things on `LSPServer` that I want to tidy up later as part of the "TODO: Clean this up more". The `fileBuffers` and `fileVersions` certainly don't need to be public, for example.

I'm not sure yet exactly which fields I'm going to make private again, though. I don't want to add too many very specific functions to support all the things that the current and future message handlers need to do. And I think that exposing some of this on `LSPServer` for use by the message handlers is analogous to how our `Server` houses several public modules.

ematipico (approved):

PR #1601, opened 2021-06-29
feat(core, cli): config migrate and simple migrations

ematipico:
## Summary

Following a [previous conversation](https://github.com/rome/tools/discussions/1366), this PR implements a command called `rome config migrate`.

The purpose of this command is to help users to easily change their configuration in case there's been breaking changes inside a configuration file.

Config migrations are not bulletproof and they don't assure a complete mitigation from breaking changes. There might be the case where a new configuration is more articulated compared to the previous one, and it could not be possible to map the old configuration to the new one.

Although, we want to try to make sure that the developer experience is good enough to make breaking changes less painful.

___

I removed some YAML file that I copied over when we have implemnted the TOML parser and I enabled the tests around TOML parser.

I will need to enable further tests and this requires more work around how the fixtures are handled, in particular the invalid cases where we have to print the diagnostics inside the snapshot tests.

I refactored a bit the check we do for "it is a repository and there aren't uncommitted changes" so it can be used in other parts of the CLI (it was used inside the `auto-config` command but it is now used also in the `migrate` command).

## Test Plan

Created new tests.  
Once the integration tests are available again, I will add new ones.


internal/consume/Consumer.ts
```diff
@@ -169,6 +171,21 @@ export default class Consumer {
 		}
 	}
 
+	public async handleAsyncThrownDiagnostics(callback: AsyncVoidCallback) {
+		if (this.handleUnexpected === undefined) {
+			callback();
```
yassere, line 176: Should this be awaited so it's not an unhandled rejection if it throws?

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "rome-root",
 	"license": "MIT",
-	"version": "10.0.7-beta.18",
+	"version": "10.0.7",
```
jamiebuilds, line 4: Did you mean to commit this?

internal/diagnostics/descriptions/commands/initCommand.ts
```diff
@@ -20,31 +20,23 @@ const IGNORE_ADVICE: DiagnosticAdvice[] = [
...
 	EXPECTED_REPO: {
 		category: DIAGNOSTIC_CATEGORIES["commands/auto-config/expectedRepo"],
-		message: markup`Directory is not a repository. Are you sure this is where you wanted to create a project?`,
 		advice: [
 			{
 				type: "log",
-				category: "warn",
-				text: markup`This command is destructive and will format and autofix all files within.`,
+				category: "info",
+				text: markup`Are you sure this is where you wanted to create a project?`,
```
jamiebuilds, line 38: Is it still destructive?

jamiebuilds: Also, without context it's kinda confusing why it would ask you this question

internal/codec-config/test-fixtures/example-v0.3.0/input.test.md
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+```
+
+### `stringify .toml`
+
+```toml
+Table.key = value
+dog.tater.type = pug
+x.y.z.w = [object Object]
```
jamiebuilds, line 73: This is all invalid

internal/diagnostics/descriptions/commands/initCommand.ts
ematipico: I refactored the messages because I wanted the "This command is destructive..." bit inside the new command. Ultimately, the end result is always the same:

Screenshot 2021-06-30 at 11 42 02


Although there was a bug that I know fixed

internal/codec-config/test-fixtures/example-v0.3.0/input.test.md
ematipico: Fixed! Thank you

internal/core/common/commands.ts
```diff
@@ -70,3 +78,33 @@ export const commandCategories = {
...
+		if (uncommittedFiles.length > 0) {
+			throw createSingleDiagnosticsError({
+				location,
+				description: descriptions.VSC.UNCOMMITTED_CHANGES(
+					uncommittedChangesAdvice,
+				),
+			});
+		}
+	}
+}
```
yassere, line 110: Can this go somewhere else? Maybe in the `utils` directory?

internal/codec-semver/satisfies.test.ts
```diff
@@ -224,7 +224,7 @@ test(
 
 			t.true(
 				satisfiesSemver(ver, range),
-				`${range} should be satisfied by ${ver}`,
+				`${range.toString()} should be satisfied by ${ver.toString()}`,
 			);
```
jonnyowenpowell, line 228: Unnecessary calls of `toString()`.

yassere:
Did you intend to modify the vendored `rome.cjs`?

ematipico:
> Did you intend to modify the vendored `rome.cjs`?

Yes, I will upload the new version once the code review is finished

internal/core/common/commands.ts
ematipico: Not sure, I checked we have inside and I don't know if it belongs there. What do you think?

internal/core/server/ServerRequest.ts
```diff
@@ -552,6 +552,12 @@ export default class ServerRequest {
 		);
 	}
 
+	public async retrieveProjectAndConfigPaths() {
```
jonnyowenpowell, line 555: Unnecessary `async`.

internal/diagnostics/descriptions/index.ts
```diff
@@ -169,7 +171,9 @@ export const descriptions = {
 	HTML_PARSER: htmlParser,
 	MARKDOWN_PARSER: markdownParser,
 	RECOVERY_STORE: recoveryStore,
-	INIT_COMMAND: initCommand,
+	INIT_COMMAND: autoConfigCommand,
+	MIGRATE_COMMAND: migrateCommand,
 	TOML: toml,
+	VSC: vcs,
```
yassere, line 177: I think that all the spots where `VSC` is used should be `VCS`, right?

internal/diagnostics/categories.ts
```diff
@@ -409,6 +412,8 @@ export const DIAGNOSTIC_CATEGORIES: {
 	"bundler/moduleCycle": ["bundler", "moduleCycle"],
 	"bundler/topLevelAwait": ["bundler", "topLevelAwait"],
 	"childProcess/failure": ["childProcess", "failure"],
+	"vcs/uncommittedChanges": ["vcs", "uncommittedChanges"],
+	"vcs/expectedRepo": ["vcs", "expectedRepo"],
```
yassere, line 416: I think this should also be alphabetical.

scripts/core-create-migration.ts
```diff
@@ -0,0 +1,39 @@
...
+const migrationsPath = INTERNAL.append(
+	"core",
+	"server",
+	"migrate",
+	"migrations",
+);
+
+export async function main([migration]: string[]): Promise<number> {
+	if (migration === undefined) {
+		reporter.error(markup`./scripts core-create-migration [migrationName]`);
```
yassere, line 16: Should be `./script` without the `s`

internal/core/server/migrate/migrations/renameIgnorePaths/index.ts
```diff
@@ -0,0 +1,24 @@
...
+		lint.delete("ignorePaths");
+	},
+	shouldMigrate: (currentVersion, config) => {
+		if (config.has("lint")) {
+			const lint = config.get("lint");
+			return lint.has("ignorePaths");
+		}
+
+		return false;
+	},
```
yassere, line 23: How often do expect the `currentVersion` to be used here? Feels like maybe the parameters should be flipped for `shouldMigrate` if you can often just use a `(config: Consumer) => boolean`

internal/core/server/migrate/Migrator.ts
```diff
@@ -0,0 +1,64 @@
...
+		this.orderedVersions = this.orderMigrations();
+	}
+
+	private orderMigrations() {
+		const versions = Array.from(this.migrations.keys());
+		return sortSemverVersions(versions);
+	}
+
+	async run(config: Consumer) {
+		this.reporter.log(markup`Start migrating the configuration file.`);
```
yassere, line 36: If there are no migrations that need to be run, Rome still prints this "Start migrating" log but then exits silently with no further feedback, which feels odd. Maybe we can keep track of how many migrations were run so that we can report that at the end (including if that number is 0).

internal/core/common/commands.ts
yassere: Putting it in `core/server/utils` is probably fine, but it could certainly go somewhere else. I think that `core/common/commands.ts` is definitely the wrong spot though.

scripts/generated-files/migrations.ts
```diff
@@ -0,0 +1,69 @@
...
+				lines.push(`import ${toCamelCase(name)} from "./${name}";`);
+			}
+			lines.push(`import {SemverVersion} from "@internal/codec-semver";`);
+			lines.push(`import {Migration} from "../Migration";`);
+			lines.push("");
+			lines.push(
+				"export const migrations: Map<SemverVersion, Migration> = new Map();",
+			);
+			for (const {name} of defs) {
+				lines.push(`migrations.set(${name}.addedVersion, ${name});`);
```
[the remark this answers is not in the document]
ematipico: I guess I will have to use the name as key. Thank you!

ematipico:
@yassere I now changed the logic of the migrations are stored. They are not a simple `Set`. When creating a `Migrator`, the instance of the class will order them and put them in a map. `version => Array<Migration>`.

We can have various migrations per version, which makes sense. In theory, two version that run on the same version, should migrate different sections of the configuration file. If that's not the case, it means that two migrations can be merged into one.

yassere (approved):

PR #1610, opened 2021-07-01
fix(lsp): lint without unstable project-wide check watching

### LINKED ISSUE #1608: 📎 Disable project-wide check watching for LSP server
Currently, the LSP server relies on the same mechanisms as `check --watch`

This is so that it can provide project-wide diagnostics including files that aren't open in the editor, but it's also currently a source of instability and bugs.

In the interest of providing a stable LSP experience in the short term, we should modify the LSP Server to only provide diagnostics on open files for which the editor sends `textDocument` notifications.

yassere:
## Summary

This PR allows the LSP server to publish diagnostics for editor-synced `textDocument` paths without a dependence on `Checker#watch`, which has been a source of instability.

This temporarily restricts the LSP server to only operating on documents that are open in the editor. Later, project-wide checking through LSP can be restored when it can be made more stable.

This PR also prevents projects from being evicted/reloaded when a manifest is being modified unsaved in the editor. Saving a modified project manifest to disk can still cause the server to hang, which isn't addressed here.

Closes #1608

### DIFF
--- a/internal/core/server/lsp/LSPServer.ts
+++ b/internal/core/server/lsp/LSPServer.ts
@@ -16,6 +16,7 @@ import {
 	DIAGNOSTIC_CATEGORIES,
 	Diagnostic,
 	DiagnosticsProcessor,
+	catchDiagnostics,
 	formatCategoryDescription,
 } from "@internal/diagnostics";
 import {
@@ -33,9 +34,11 @@ import {LSPTransport} from "./protocol";
 import LSPProgress from "./LSPProgress";
 import {convertDiagnosticsToLSP} from "./utils";
 import {markup, readMarkup} from "@internal/markup";
-import {Event} from "@internal/events";
+import {Event, EventQueue} from "@internal/events";
 import {CommandName} from "@internal/core/common/commands";
 import {notificationHandlers, requestHandlers} from "./messages";
+import {TextDocumentSync} from "./types";
+import {promiseAllFrom} from "@internal/async";
 
 type LSPProjectSession = {
 	request: ServerRequest;
@@ -54,6 +57,10 @@ export default class LSPServer {
 		this.fileVersions = new AbsoluteFilePathMap();
 
 		this.watchProjectEvent = new Event("LSPServer.watchProject");
+		this.textDocumentSyncEvent = new EventQueue(
+			"LSPServer.textDocumentSyncEvent",
+			{toDedupeKey: ({path}) => path.join()},
+		);
 
 		request.endEvent.subscribe(async () => {
 			await this.shutdown();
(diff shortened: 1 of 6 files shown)


internal/core/server/lsp/LSPServer.ts
```diff
@@ -89,6 +96,60 @@ export default class LSPServer {
 	private projectSessions: AbsoluteFilePathMap<LSPProjectSession>;
 
 	public watchProjectEvent: Event<AbsoluteFilePath, void>;
+	public textDocumentSyncEvent: EventQueue<TextDocumentSync>;
+
+	public async init() {
```
ematipico, line 101: This function doesn't need to be `async`. There's no `await` that belongs to `public async` keyword. The only `await` we have belongs to the callback inside the `.subscribe`.

If that's done on purpose, I think we should return a custom promise that resolves when `promiseAllFrom` is called (if that's the intention).

internal/core/server/lsp/LSPServer.ts
```diff
@@ -89,6 +96,60 @@ export default class LSPServer {
...
+			return this.request.requestWorkerLint(
+				path,
+				{applySafeFixes: false, save: false},
+			);
+		});
+		this.logDiagnostics(path, diagnostics);
+		if (value === undefined) {
+			return;
+		}
+		this.diagnosticsProcessor.removePath(path);
```
ematipico, line 134: Why is the `path` removed and later we call `.getDiagnosticsForPath`?

internal/core/server/lsp/LSPServer.ts
```diff
@@ -89,6 +96,60 @@ export default class LSPServer {
...
+		const {value, diagnostics} = await catchDiagnostics(async () => {
+			return this.request.requestWorkerLint(
+				path,
+				{applySafeFixes: false, save: false},
+			);
+		});
+		this.logDiagnostics(path, diagnostics);
+		if (value === undefined) {
+			return;
+		}
```
ematipico, line 133: I think we should log the diagnostics only when `value` is `undefined`. This is because `catchDiagnostics` return a particular result:

```ts
type WrapResult<T> =
	| {
			readonly value: T;
			readonly diagnostics: undefined;
		}
	| {
			readonly value: undefined;
			readonly diagnostics: Diagnostic[];
		};
 ```

Which means that if `value` is not `undefined`, we are calling `logDiagnostics` with no diagnostics. We could save call.

File: internal/core/server/lsp/messages/general.ts [hunk 4]
```diff
@@ -12,8 +12,9 @@ export const exit: LSPNotificationHandler = async () => {
 	await safeProcessExit(0);
 };
 
-export const initialized: LSPNotificationHandler = async (lsp) => {
-	await lsp.watchPendingProjects();
+export const initialized: LSPNotificationHandler = async () => {
+	// Project-wide check watching disabled for stability
```
ematipico, line 16: Should we add a `TODO`? To remind us that this will be enabled again in the future.

internal/core/server/lsp/types.ts
```diff
@@ -381,3 +382,17 @@ export type LSPVersionedTextDocumentIdentifier = LSPTextDocumentIdentifier & {
 	 */
 	version: number | null;
 };
+
+export type TextDocumentSync =
+	| {
+			path: AbsoluteFilePath;
+			type: "DID_OPEN";
```
ematipico, line 389: Are these strings coming from the spec of LSP?

internal/core/server/lsp/LSPServer.ts
yassere: The old diagnostics for the path need to be removed from the diagnostics processor before the updated diagnostics are added.

internal/core/server/lsp/LSPServer.ts
yassere: Most of our `init` methods end up async and I forgot to change it when I ended up not needed to await anything. I can probably just move this subscription into the constructor.

internal/core/server/lsp/types.ts
yassere: The strings that come from the spec look like: `"textDocument/didOpen"`
This type is modeled after the `ServerRefreshFile` type which is used for the `refreshFileEvent` event queue.

internal/core/server/lsp/LSPServer.ts
yassere: Actually, I'll just remove the `async` so that it matches [this.](https://github.com/rome/tools/blob/95e9ba9f3e51c7914070050c8b6293779c9c4ecc/internal/core/server/fs/FileAllocator.ts#L34-L40)

internal/core/server/lsp/types.ts
ematipico: Should we consider a `enum` to store all these strings?

yassere: I don't think it's necessary. We still get type checking and code completion from tagged unions, and we already use this approach throughout the codebase.

text-document-sync

ematipico: It's not for the autocomplete. I don't like to repeat strings. In the code base we use strings because we didn't support `enum` at that time. We now support `enum` so I think we should start pushing their usage.

yassere: > I don't like to repeat strings.

Repeating strings is in the spirit of our `noShoutyConstants` lint rule. 



> We now support `enum` so I think we should start pushing their usage.

Personally, I disagree. But if that is something we want to do, we should discuss it outside the scope of this PR.

ematipico (approved):

PR #1612, opened 2021-07-01
refactor(core): remove unnecessary auto-config double call

JustBeYou:
## Summary
There were two `auto-config` queries to the server. They are unnecessary and the same job could be done with only one. I discussed this with @ematipico on Discord.


ematipico (approved):
Thank you!

Issue #1570, opened 2021-06-17
☂️   0.1 Release Tasks


jamiebuilds:
- [x] Turn off CSS support temporarily (won't be ready in time)
- [x] #1360 "Add support for fixes in ESLint integration"
- [x] #1388 "Remove hardcoded path from ESLint integration"
- [x] Add ESLint interop option (go through ESLint flags to see which ones we need to support)
- [ ] Test ESLint integration against popular ESLint plugins
- [x] Test ESLint integration against projects already using ESLint
- [x] Basic Prettier support (if you are using a `.prettierrc` it should just work)
- [x] Disable our formatter by default?
- [x] Stabilize LSP server #1608 
- [x] Improve LSP spec compliance
- [x] Fix/improve LSP code action support
- [ ] Improved Contributing Docs
- [ ] Discuss: One config format, in one place https://github.com/rome/tools/issues/1596
- [x] Disable dependency check during `rome check` command https://github.com/rome/tools/issues/1607
- [ ] Display auto-fix code actions as quick fixes in VS Code https://github.com/rome/tools/issues/1619
- [ ] Update governance structure https://github.com/rome/tools/issues/1324#issue-801438375

ematipico:
Working on the basic prettier support

yassere:
We should also stabilize the LSP Server and improve its spec compliance. I'm working on that now.

jamiebuilds:
@yassere You mind adding tasks that you need to work on to this issue?

ematipico:
Currently working on the eslint interop.

jamiebuilds:
@ematipico @yassere Could you add some details about removing dependency checking, and simplifying the LSP and `rome check` commands?

sebmck:
Opened #1607

ematipico:
The integrations are ready! Now that we decided to turn off dependency check and manifest, maybe we might able to add to the CI the integrations. 

Before it wasn't an option.

yassere:
There's more we can do to improve LSP spec compliance and code action support, but I think they're good enough now to no longer be blockers for this release.

jer3m01:
We should also add a task to update GOVERNANCE.md.

ematipico:
@sebmck @jamiebuilds the governance one should assigned to one of you.

jamiebuilds:
Hi! Thank you for opening this issue

[...]

PR #1613, opened 2021-07-02
feat(core, lsp): disable dependency graph

ematipico:
## Summary

Part of #1570 , this PR temporary disables the check of dependencies during `rome check`.

This is a temporary change and it will be turned on again under a flag in the next release, after we figure out what's the correct flow for users.

## Test Plan

Existing CI

### DIFF
--- a/internal/core/server/checker/Checker.ts
+++ b/internal/core/server/checker/Checker.ts
@@ -52,6 +52,7 @@ export type CheckerOptions = {
 	globalDecisions?: LintCompilerOptionsDecisions;
 	lintCompilerOptionsPerFile?: LinterCompilerOptionsPerFile;
 	suppressionExplanation?: string;
+	enableDependenciesCheck?: boolean;
 };
 
 type ProgressFactory = (opts: ReporterProgressOptions) => ReporterProgress;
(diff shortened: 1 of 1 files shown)


yassere:
I don't think that all this code needs to be commented out. I think it would be better if we added a flag that gets passed into `CheckerOptions` that determines whether or not the dependency analysis would run, and we could have it off by default. That would make it easier for us to test everything when we're working on adding dependency analysis back in.

ematipico:
> I don't think that all this code needs to be commented out. I think it would be better if we added a flag that gets passed into `CheckerOptions` that determines whether or not the dependency analysis would run, and we could have it off by default. That would make it easier for us to test everything when we're working on adding dependency analysis back in.

Done! Added a mandatory option and set it `false` every time we create a `Checker`. This change now affects the `LSPServer` too.

internal/core/server/checker/Checker.ts
```diff
@@ -52,6 +52,7 @@ export type CheckerOptions = {
 	globalDecisions?: LintCompilerOptionsDecisions;
 	lintCompilerOptionsPerFile?: LinterCompilerOptionsPerFile;
 	suppressionExplanation?: string;
+	enableDependenciesCheck: boolean;
```
yassere, line 55: Could this just be optional? Then it would only need to be included when you do want dependency checking enabled.

ematipico: I intentionally made it required in order to catch all the instances. Let me change it

yassere (approved):

PR #1616, opened 2021-07-04
feat(resolver, project, core): path aliases

### LINKED ISSUE #1555: 📎 Support for path aliases

JustBeYou:
## Summary

I remade the PR #1569. Fixes #1555 
At the moment `auto-config` would simply read `tsconfig` and add the missing entries. It does not sync the path aliases entirely because there is no command to pop/reset arrays in the config. Also, there is no support for Node packages remapping. 

## Testing
Manual and some unit tests, no integration tests yet.

### DIFF
--- a/internal/core/server/commands/autoConfig.ts
+++ b/internal/core/server/commands/autoConfig.ts
@@ -11,6 +11,7 @@ import {
 import {UnknownObject} from "@internal/typescript-helpers";
 import Checker from "../checker/Checker";
 import { consumeConfig } from "@internal/codec-config";
+import { aliasPatternToString } from "@internal/project/aliases";
 
 interface Flags extends UnknownObject {
 	checkVSC: boolean;
@@ -137,8 +138,17 @@ export default createServerCommand<Flags>({
 						const paths = compilerOptions.get("paths");
 						const resultPaths: [string, string[]][] = []
 						if (paths.exists()) {
-							for (const [alias, targets] of paths.asMap()) {
-								resultPaths.push([alias, targets.asMappedArray(target => target.asString())])
+							const currentPaths = currentProject.config.aliases.paths;
+							const currentPathsSet = new Set<string>();
+							for (const [alias, _] of currentPaths) {
+								currentPathsSet.add(aliasPatternToString(alias))
+							}
+
+							for (const [alias, targetsConsumer] of paths.asMap()) {
+								const targets = targetsConsumer.asMappedArray(target => target.asString());
+								if (!currentPathsSet.has(alias)) {
+									resultPaths.push([alias, targets])
+								}
 							}
 						}
 
--- a/internal/core/server/commands/autoConfig.ts
+++ b/internal/core/server/commands/autoConfig.ts
@@ -7,11 +7,13 @@ import {
 	Diagnostic,
 	createSingleDiagnosticsError,
 	descriptions,
+	DIAGNOSTIC_CATEGORIES,
 } from "@internal/diagnostics";
 import {UnknownObject} from "@internal/typescript-helpers";
 import Checker from "../checker/Checker";
-import {consumeConfig} from "@internal/codec-config";
+import {json5} from "@internal/codec-config";
 import {aliasPatternToString} from "@internal/project/aliases";
+import { consumeUnknown } from "@internal/consume";
 
 interface Flags extends UnknownObject {
 	checkVSC: boolean;
@@ -24,8 +26,8 @@ export type AutoConfig = {
 	};
 	licenses?: Diagnostic[];
 	aliases?: {
-		base: string;
-		paths: [string, string[]][];
+		base?: string;
+		paths?: [string, string[]][];
 	};
 };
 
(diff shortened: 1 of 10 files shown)


internal/diagnostics/descriptions/projectConfig.ts
```diff
@@ -14,4 +14,19 @@ export const projectConfig = createDiagnosticsCategory({
 		],
 	}),
 	RECURSIVE_CONFIG: {message: markup`Recursive config`},
+
+	TOO_MANY_WILDCARDS: (pattern: string) => ({
+		message: markup`Pattern ${pattern} contains more than one wildacard.`,
```
ematipico, line 19:
```suggestion
		message: markup`Pattern <emphasis>${pattern}</emphasis> contains more than one wildacard.`,
```

internal/diagnostics/descriptions/projectConfig.ts
```diff
@@ -14,4 +14,19 @@ export const projectConfig = createDiagnosticsCategory({
...
+		advice: [
+			{
+				type: "log",
+				category: "info",
+				text: markup`Alias patterns should have the following format \\<prefix>[*]\\<sufix>.`,
```
ematipico, line 24:
```suggestion
				text: markup`Alias patterns should have the following format <emphasis>\\<prefix>[*]\\<sufix></emphasis>.`,
```

internal/project/load.ts
```diff
@@ -562,6 +564,28 @@ export async function normalizeProjectConfig(
 		prettier.enforceUsedProperties("prettier config property");
 	}
 
+	const aliases = consumer.get("aliases");
+	if (aliases.exists()) {
+		if (aliases.has("base")) {
+			const base = aliases.get("base").asFilePath();
+			config.aliases.base = projectDirectory.resolve(base);
```
ematipico, line 571: In case `base` doesn't exist, should we fallback to the project root directory? Or, from a different point of view, is this configuration accepted:

```toml
[aliases.paths]
"@components" = [ "./components" ]
```

We don't have `base` but we still have `paths`. If accepted, we should fallback `base` to a default value. If it's not accepted, we should throw a diagnostics here saying that `base` or `paths` are mandatory.

internal/project/aliases.ts
```diff
@@ -0,0 +1,65 @@
...
+
+	if (path === prefix) {
+		return path;
+	}
+
+	return undefined;
+}
+
+export function aliasPatternToString(pattern: PathAliasPattern): string {
+	const [prefix, sufix] = pattern.parts;
```
ematipico, line 60:
```suggestion
	const [prefix, suffix] = pattern.parts;
```

internal/project/aliases.ts
```diff
@@ -0,0 +1,65 @@
...
+	return prefix;
+}
+
+export function matchAliasPattern(
+	path: string,
+	pattern: PathAliasPattern,
+): string | undefined {
+	const [prefix, sufix] = pattern.parts;
+	if (pattern.wildcard && path.startsWith(prefix) && path.endsWith(sufix)) {
+		return path.slice(prefix.length, path.length - sufix.length);
```
ematipico, line 49:
```suggestion
		return path.slice(prefix.length, path.length - suffix.length);
```

internal/project/aliases.ts
```diff
@@ -0,0 +1,65 @@
...
+		parts: path.split("*"),
+		wildcard: wildcards.length === 1,
+	};
+}
+
+export function buildPathFromAliasPattern(
+	path: string,
+	pattern: PathAliasPattern,
+): string {
+	const [prefix, sufix] = pattern.parts;
```
ematipico, line 36:
```suggestion
	const [prefix, suffix] = pattern.parts;
```

internal/project/aliases.ts
```diff
@@ -0,0 +1,65 @@
...
+	if (pattern.wildcard) {
+		return prefix + path + sufix;
```
ematipico, line 38:
```suggestion
		return prefix + path + suffix;
```

ematipico (commented):
Thank you for your PR! It's great! I left some comments around DX, messages and configuration

internal/core/server/commands/autoConfig.ts
```diff
@@ -109,6 +115,54 @@ export default createServerCommand<Flags>({
...
 				},
 			},
+			{
+				message: markup`Import path aliases from tsconfig.`,
+				async callback() {
+					const tsconfigPath = currentProject.directory.append("tsconfig.json");
+					if (await tsconfigPath.exists()) {
+						const {consumer: tsconfig} = consumeConfig({
+							path: tsconfigPath,
+							input: await tsconfigPath.readFileText(),
```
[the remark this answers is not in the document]
ematipico: The `consumeConfig` was especially created to import Rome's configuration file. Not config files in general

internal/project/load.ts
JustBeYou: https://github.com/rome/tools/pull/1616/files#diff-a0a82f5f75b7cd89d28f7d3edc674021c72a9f7ff133bc1fce4639156a855ba9R331

Here is the default value for the `base`. I think we need to call `projectDirectory.resolve()` on it too, right?

JustBeYou: I see that VCS is configured in the same way:
```
		if (vcs.has("root")) {
			config.vcs.root = projectDirectory.resolve(vcs.get("root").asFilePath());
		}
```
and
```
		vcs: {
			root: createAbsoluteFilePath("/"),
		},
```

ematipico:
@JustBeYou I'm going to put the PR on hold as we are planning to do a release soon and it would be better to not have this feature released already. (documentation is missing for example)

jamiebuilds:

Thanks again!

PR #1626, opened 2021-07-08
feat(core, project): config remove/pop commands

### LINKED ISSUE #1617: 📎 Delete/pop/reset config
`rome config` does not support any deletion related commands. I propose to have the following new commands:
* `rome config remove <key>` - deletes the configuration `key`
* `rome config pop <key> <value>` - deletes the specified `value` from an array at `key`
* `rome config reset <key>` - removes all entries from the array at `key`

I found the need for those commands when working on #1616 . When syncing `tsconfig` and `rome` paths there is no way to remove old entries. 

I will start working on this feature in the following days if you think it's worth.

JustBeYou:
## Summary
Fixes #1617 

## Test plan
Tested on the following scenarios:
* `config remove <nonexistent key>` - ignores it
* `config remove <existing key>` - removes it from config
* `config pop <nonexistent key> <value>` - ignores it
* `config pop <existing key but not array> <value>` - fails with `Expected an array`
* `config pop <exiting key and is array> <...values>` - removes them

### DIFF
--- a/internal/core/client/commands/config.ts
+++ b/internal/core/client/commands/config.ts
@@ -5,7 +5,7 @@ import {Reporter} from "@internal/cli-reporter";
 
 export const remove = createLocalCommand({
 	category: commandCategories.PROJECT_MANAGEMENT,
-	description: markup`modify a project config - remove ${"<key>"}`,
+	description: markup`modify a project config - remove "\\<key>"`,
 	usage: "<key>",
 	examples: [],
 	defineFlags() {
@@ -31,9 +31,12 @@ export const remove = createLocalCommand({
 
 export const pop = createLocalCommand({
 	category: commandCategories.PROJECT_MANAGEMENT,
-	description: markup`modify a project config - remove "<values>" from "<key>"`,
+	description: markup`modify a project config - remove "\\<values>" from "\\<key>"`,
 	usage: "<key> <...values>",
-	examples: [],
+	examples: [{
+		command: "rome config pop aliases.paths path1 path2",
+		description: markup`Remove multiple values from aliases.paths configuration.`,
+	}],
 	defineFlags() {
 		return {};
 	},
--- a/internal/core/client/commands/config.ts
+++ b/internal/core/client/commands/config.ts
@@ -1,0 +1,63 @@
+import {createLocalCommand} from "../commands";
+import {commandCategories} from "@internal/core/common/commands";
+import {markup} from "@internal/markup";
+import {Reporter} from "@internal/cli-reporter";
+
+export const remove = createLocalCommand({
+	category: commandCategories.PROJECT_MANAGEMENT,
+	description: markup`modify a project config - remove ${"<key>"}`,
+	usage: "<key>",
+	examples: [],
+	defineFlags() {
+		return {};
+	},
+	async callback(req) {
+		const {client, query} = req;
+		const {reporter} = client;
+
+		if (await confirmDeletion(reporter)) {
+			await client.query(
+				{
+					commandName: "config remove",
+					args: query.args,
+				},
+				"server",
+			);
+		}
+
+		return true;
+	},
+});
+
+export const pop = createLocalCommand({
+	category: commandCategories.PROJECT_MANAGEMENT,
+	description: markup`modify a project config - remove ${"<values>"} from ${"<key>"}`,
+	usage: "<key> <...values>",
+	examples: [],
+	defineFlags() {
+		return {};
+	},
+	async callback(req) {
+		const {client, query} = req;
+		const {reporter} = client;
+
+		if (await confirmDeletion(reporter)) {
+			await client.query(
+				{
+					commandName: "config pop",
+					args: query.args,
+				},
+				"server",
+			);
+		}
+
+		return true;
+	},
+});
+
+function confirmDeletion(reporter: Reporter): Promise<boolean> {
+	reporter.warn("Your configuration will be affected by this command.");
+	return reporter.radioConfirm(
+		"Are you sure you want to remove the specified configs?",
+	);
+}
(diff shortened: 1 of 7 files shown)


internal/core/server/commands/config.ts
```diff
@@ -63,6 +78,20 @@ async function runCommand(
...
 			]);
+		} else if (action === "remove") {
+			const lastKey = keyParts[keyParts.length - 1];
+			prevKeyConsumer.delete(lastKey);
+		} else if (action === "pop") {
+			const existing = Array.from(
+				keyConsumer.required([]).asIterable(),
+				(c) => c.asUnknown(),
+			);
+			const toRemove = Array.isArray(value) ? value : [];
```
ematipico, line 89: Shouldn't it be?

```suggestion
			const toRemove = Array.isArray(value) ? value : [value];
```

I don't understand the `[]` part.

internal/core/server/commands/config.ts
```diff
@@ -63,6 +78,20 @@ async function runCommand(
...
+			keyConsumer.setValue(
+				existing.filter((item) =>
+					typeof item !== "string" || !toRemove.includes(item)
```
ematipico, line 92: I don't understand this check `typeof item !== "string"`. What's the use case? Could you explain with an example please?

internal/core/server/commands/config.ts
```diff
@@ -63,6 +78,20 @@ async function runCommand(
 				),
 				...(Array.isArray(value) ? value : []),
 			]);
+		} else if (action === "remove") {
+			const lastKey = keyParts[keyParts.length - 1];
+			prevKeyConsumer.delete(lastKey);
+		} else if (action === "pop") {
+			const existing = Array.from(
```
ematipico, line 85: Should we rename `existing` with `existingValues`?

internal/core/server/commands/config.ts
ematipico: Should we rename `toRemove` with `valuesToRemove`?

JustBeYou: I forgot to add `[value].` Fixed.

internal/core/server/commands/config.ts
JustBeYou: I assumed that `toRemove` could contain other types than `string` but it seems it's not the case. I've changed it.

internal/core/client/commands/config.ts
```diff
@@ -0,0 +1,63 @@
...
+			);
+		}
+
+		return true;
+	},
+});
+
+export const pop = createLocalCommand({
+	category: commandCategories.PROJECT_MANAGEMENT,
+	description: markup`modify a project config - remove ${"<values>"} from ${"<key>"}`,
```
ematipico, line 34:
```suggestion
	description: markup`modify a project config - remove "<values>" from "<key>"`,
```

internal/core/server/commands/config.ts
```diff
@@ -346,3 +392,29 @@ export const migrate = createServerCommand({
 		await configPath.writeFile(stringified);
 	},
 });
+
+export const remove = createServerCommand<Flags>({
+	category: commandCategories.PROJECT_MANAGEMENT,
+	description: markup`modify a project config - remove ${"<key>"}`,
```
ematipico, line 398:
```suggestion
	description: markup`modify a project config - remove "<key>"`,
```

internal/core/server/commands/config.ts
```diff
@@ -346,3 +392,29 @@ export const migrate = createServerCommand({
...
+	defineFlags,
+	async callback(req, flags) {
+		req.expectArgumentLength(1);
+		await runCommand(req, flags, req.query.args[1], "remove");
+	},
+});
+
+export const pop = createServerCommand<Flags>({
+	category: commandCategories.PROJECT_MANAGEMENT,
+	description: markup`modify a project config - remove ${"<values>"} from ${"<key>"}`,
```
ematipico, line 411:
```suggestion
	description: markup`modify a project config - remove "<values>" from "<key>"`,
```

website/src/_includes/docs/cli-commands.md
```diff
@@ -53,13 +53,21 @@ Set the `key` to `false`.
...
 Set the `key` to the string `value`. If `value` is an absolute path then it will be made relative to the config path.
 
 #### `rome config push <key> <value>`
 
-Push the string `value` to an array at `key`. If `key` doesn't exist then it will be created.
+Push the string `value` to an array at `key`. You can pass multiple values to push separated by space.
+
+#### `rome config pop <key> <value>`
+
+Pop the string `value` from an array at `key`. If the `key` doesn't exist then it will be created. You pass multiple values to remove separated by space.
```
ematipico, line 70:
```suggestion
Pop the string `value` from an array at `key`. If the `key` doesn't exist then it will be created. You can pass multiple values to remove, separated by space. 
```

ematipico: Would it make sense to have an example?

```bash
rome config pop aliases.paths path1 path2
```

ematipico:
Now that our integration tests are enabled again, we could add some test for this new commands. We have none for the `config` command. If you want, I am going to create some tests and later on you can piggyback them and add new ones.

ematipico:
[I added few tests](https://github.com/rome/tools/blob/main/internal/core/server/commands/config.test.ts). As you can see, it's not that difficult to create a new test, we only care about the configuration file.

The only caveat is that this kind of integration test relies of snapshot tests only, which means that when you create a test, you have to make sure that the result in snapshot is what you expect.

What you get in the snapshot is simply the result of the operation.

JustBeYou:
@ematipico I've pushed the tests but I'm not able to generate snapshots. Could you take a look?

ematipico:
@JustBeYou I had to disable the config tests. They are causing some trouble, I need to spend some time to understand why. (it's probably due to the fact that the configuration is changed while a server is running).

If you fixed the lint issue, I'd be more than happy to merge the PR!

JustBeYou:
@ematipico should be ready to be merged!

jamiebuilds:

Thanks again!

Issue #1348, opened 2021-02-09
🐛 Errors handling default values with destructuring 


ajkachnic:
There appears to be a parsing error when handling default values within destructuring. It throws a parsing error on code like:
```js
const foo = ({ bar = 'foobar' }) => bar
```

This is my first time trying out Rome, so there might be something that I'm missing. I did a brief check for duplicates so there shouldn't be any, but I could be wrong

## Environment

```
$ rome rage --summary
Environment Variables
  Object {
    COLORTERM: "truecolor"
    LANG: "en_US.UTF-8"
    TERM: "xterm-kitty"
  }

User Config
  unset

Rome Version
  10.0.4-beta

Node Version
  14.15.4

Platform
  linux x64 5.4.89-1-MANJARO

Terminal Features
  Object {
    background: "unknown"
    colorDepth: 24
    columns: 89
    cursor: true
    hyperlinks: true
    isTTY: true
    unicode: true
  }

Client Flags
  Object {
    clientName: "cli"
    cwd: "/home/andrew/dev/personal/projects/main/website-redesign"
    realCwd: "/home/andrew/dev/personal/projects/main/website-redesign"
    silent: false
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 29_175_808
      pid: 234_262
      uptime: 0.574296348
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 29_175_808
        ownedBytes: 0
        ownedFileCount: 0
        pid: 234_262
        uptime: 0.574027092
      }
    ]
  }

```

## Steps To Reproduce

1. Add this code to a file in a rome project:
```js
const someFunction = ({
  message = "hello"
}) => {
  console.log(message)
}

someFunction({})
someFunction({message: 'hello world'})
```
2. Run `rome check`

## Expected Results

This should not fail to parse, as it is valid JavaScript. If you run `node` on this file, it executes fine. I assume this is just an edge case within the parser.

ematipico:
Hi @ajkachnic , what's the actual error that you get?

ajkachnic:
It outputs this error

```
 index.js:2:10 parse/js ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Unexpected token, expected ,

    1 │ const someFunction = ({
[...]

PR #1635, opened 2021-07-12
fix(js-parser): default values with destructuring

ktfth:
Trying to solve the bug related to this [issue](https://github.com/rome/tools/issues/1348)

./rome test


ematipico:
@ktfth could you please restore the template? Not the first time that happens

PR #1646, opened 2021-07-15
feat(lint): add missing TypeScript types

Conaclos:
See discussion #1640

This PR adds missing TypeScript types, such as `ArrayConstructor`. The array of types is lexicographically ordered.

# How

I have retrieved the types using the following commands:

```sh
grep -E '^type|interface' lib.es* | cut -d' ' -f2 | cut -d '<' -f1 | sort | uniq
grep -E '^declare type' lib.es* | cut -d' ' -f3 | cut -d '<' -f1 | sort | uniq | wc -l
```

in TypeScript [lib folder](https://github.com/microsoft/TypeScript/tree/main/lib).

I removed of this auto-generated list the usual JavaScript classes (Array, String, ...). Note that I also removed `Promise` of the list. I'm not sure why it was present in the rome list...

Compared to the current rome list, some types were missing:
- BufferEncoding
- BufferSource
- Console
- NodeJS
- NodeRequire
- RequestInit
- Thenable

Thus, I added these types in the list.

# Are types missing?

I deliberately omitted the types of the DOM (lib.dom*) and webworkers (lib.web*). TypeScript declares more than 1000 types for them.


ematipico:
We should definitely find a better way to define those types in the future. For now, we really appreciate this PR! It brings lot of value!

Conaclos:
@ematipico
The support of declaration files (.d.ts) is certainly the way to go.

PR #1648, opened 2021-07-16
docs(website): move diagnostics in it's own section

ematipico:
## Summary

This PR creates a new quick link called `Diagnostics`. 

Rome's diagnostics are a concept that related not only to linting, both other categories/sections. It's how we catch generic/specific errors.

As there's been lot of confusion, where a diagnostic was misunderstood as a lint error, maybe having their explanation in a separate chapter could help the users. 

## Test Plan

CI


jer3m01 (approved):
Great explanation!

PR #1652, opened 2021-07-16
feat(js-parser): enable declare TypeScript modifier on a class's field

### LINKED ISSUE #1653: 📎 Support for `declare` inside class field declarator
### Description

Enable the following TypeScript syntax:

```ts
class A {
    declare a: unknown // valid (1)
    declare b = 5 // invalid (2)
}
```

Conaclos:
Hi!

This PR adds the capability to add the `declare` TypeScript modifier on class's fields. It also updates the affected snapshots.

Closes #1653 

# Context
See discussion #1638.
The declare modifier is only allowed on class's field without initialization.
```ts
class A {
    declare a: unknown // valid (1)
    declare b = 5 // invalid (2)
}
```

# Potential issues
This PR enables the use of the declare modifier on any class's field. I'm not sure that initialized field are ruled out... I don't see any way to know if a class's field have an initialization part...

This is the first time I modify the parser, I would appreciate to get a review in order to be sure that everything is ok :)


ematipico (changes requested):
We should add new tests, as you modified some parts of the code:

**Formatter**: create a new test where we create a new `input.ts` file with the new syntax, so we make sure is properly formatted

**Parser**: we enabled syntax like `declare readonly`, `declare` or `declare abstract`. We should create three new tests for each case

The `CONTRIBUTING.md` doc should help on how to create new tests. If you have specific questions feel free to ask!

Conaclos:
@ematipico
`declare` modifier is now disallowed for private and initialized fields :)

I added tests for the formatter and the parser.

ematipico (approved):

Issue #1537, opened 2021-06-07
🐛 Regex numeric backrefs parsed incorrectly and broken by `check --apply`


yassere:
## Environment

<details>
<summary>$ rome rage --summary</summary>

```
Environment Variables
  {COLORTERM: "truecolor", LANG: "en_US.UTF-8", TERM: "xterm-256color", TERM_PROGRAM: "vscode"}

User Config
  unset

Rome Version
  10.0.7-beta.18

Node Version
  14.17.0

Platform
  darwin x64 19.6.0

Terminal Features
  {
    background: "unknown"
    colorDepth: 24
    cursor: true
    hyperlinks: true
    isTTY: true
    progress: true
    unicode: true
    columns: OneIndexedNumber<145>
  }

Client Flags
  {
    clientName: "cli"
    silent: false
    cwd: AbsoluteFilePath</Users/Yasser/Dev/rome>
    realCwd: AbsoluteFilePath</Users/Yasser/Dev/rome>
  }

Server Status
  {
    projects: [{id: 0}, {id: 1}, {id: 2}, {id: 3}, {id: 4}, {id: 5}]
    workers: [{astCacheSize: 0, heapTotal: 425_517_056, ownedBytes: 398n, ownedFileCount: 4, pid: 37_503, uptime: 115_262.351506074}]
    server: {heapTotal: 425_517_056, pid: 37_503, uptime: 115_262.351883992}
  }
```
</details>

## Steps To Reproduce

Backreferences are parsed as octal escapes even if there are capturing groups they can refer to.

1.  Create a file `test.foo.ts` containing: 
```
/f(o)\1/;
```
2. Running `./rome parse test.foo.ts` results in this expression:
```
        expression: JSRegExpSubExpression {
          body: [
            JSRegExpCharacter {value: "f"}
            JSRegExpGroupCapture {
              expression: JSRegExpSubExpression {body: [JSRegExpCharacter {value: "o"}]}
            }
            JSRegExpCharacter {value: "\x01"}
          ]
        }
```
3. Running `./rome check --apply test.foo.ts` modifies the file to:
```
/f(o)\u0001/;
```
The same thing happens even the regex has a unicode flag: `/f(o)\1/u;`

## Expected Results

  - Running `check --apply` shouldn't turn a backref into a unicode escape sequence.
  - Numeric backrefs should be parsed as `JSRegExpNumericBackReference`.

And based on a quick look at the spec, so please someone correct me if I'm wrong:
[...]

PR #1661, opened 2021-08-06
fix(compiler): fix regex parsing of back reference characters (#1537)

MichaReiser:
## Summary

This PR ensures that the regex parser parses numeric and named backreferences correctly. 

**Named Backreferences**
The regex parser always emitted an unrecoverable `UNCLOSED_NAMED_CAPTURE` diagnostic for named back
references because the `JSRegExpNumericBackReference` returned by `tokenizeEscaped` removed the `<>` from the parsed name, but the `parseCharacter` function stripped the `<` and assert that the name ends with a `>`.

This diff changes `tokenizeEscaped` to return the name with the `<>`. The `parseCharacter` function then asserts for the closing `>` and strips `<>` from the name.

**Numeric Backreferences**
The regex parser incorrectly parsed numeric backreferences in regular expressions as unicode characters. e.g. the backreference `\1` in the `/f(o)\1/;` expression gets parsed as `JSRegExpCharacter {value: '\x01'}` instead of a `JSRegExpNumericBackReference{value: 1, ...}`

This commit adds the parsing of numeric backreferences according to the [ecma script specification](https://tc39.es/ecma262/multipage/text-processing.html#prod-DecimalEscape), treating numbers starting with `1-9` as numeric back references.

## Test Plan

```
 ./rome test internal/codec-js-regexp/index.test.ts
```


internal/codec-js-regexp/index.ts
```diff
@@ -164,6 +168,20 @@ function tokenizeEscaped(
 		}
 	}
 
+	// DecimalEscape https://tc39.es/ecma262/multipage/text-processing.html#prod-DecimalEscape
+	if (/[1-9]/.test(char)) {
+		const backRefString = tokenizer.read(isDigit);
```
MichaReiser, line 173: The [specification](https://tc39.es/ecma262/multipage/text-processing.html#prod-DecimalEscape) mentions that `lookahead` should not be in `DecimalDigit`. I'm not sure what that means or requires any additional handling.

internal/codec-js-regexp/index.ts
```diff
@@ -214,21 +232,21 @@ function tokenizeEscaped(
...
 	}
 
+	// named group back reference https://github.com/tc39/proposal-regexp-named-groups#backreferences
 	if (tokenizer.eat("k")) {
 		if (parser.options.unicode) {
-			// named group back reference https://github.com/tc39/proposal-regexp-named-groups#backreferences
-			if (tokenizer.eat("<")) {
-				const value = tokenizer.read(isBackReferenceCharacter);
-				tokenizer.assert(">");
+			tokenizer.assert("<");
```
MichaReiser, line 238: Not sure if we should move this assertion to `parseCharacter` as well, according to the assertion of the `>`

internal/codec-js-regexp/index.ts
```diff
@@ -214,21 +232,21 @@ function tokenizeEscaped(
...
+			const value = `<${tokenizer.read(isBackReferenceCharacter)}${tokenizer.eat(
```
MichaReiser, line 239: `isBackReferenceCharacter` isn't spec conform. It reads until it finds a `>`. I can change this as well if we see a need for it.

internal/codec-js-regexp/index.ts
```diff
@@ -164,6 +168,20 @@ function tokenizeEscaped(
 		}
 	}
 
+	// DecimalEscape https://tc39.es/ecma262/multipage/text-processing.html#prod-DecimalEscape
+	if (parser.options.unicode && /[1-9]/.test(char)) {
```
MichaReiser, line 172: The `non-unicode` backreferences are not as straightforward. The `\` are only treated as backreferences if there's a [capture group with the given index](https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#prod-annexB-AtomEscape).

internal/codec-js-regexp/index.ts
```diff
@@ -214,27 +241,15 @@ function tokenizeEscaped(
 		);
 	}
 
+	// named group back reference https://github.com/tc39/proposal-regexp-named-groups#backreferences
 	if (tokenizer.eat("k")) {
-		if (parser.options.unicode) {
```
MichaReiser, line 218: I tested this in node and named group backreferences work fine with and without the unicode flag.

MichaReiser (commented):
I had to change the parser to keep track of the number of opening capturing parentheses (as mentioned in the spec) and only return a numeric back reference if it is valid (and otherwise return the octal code). 

This, however, breaks the `noReferenceToNonExistingGroup` lint rule.

https://github.com/rome/tools/blob/d4b8b5c5c436d4795958676ff40b8a61b3a6d7e3/internal/compiler/lint/rules/regex/noReferenceToNonExistingGroup.ts#L36-L59

The problem is that the rule matches `JSRegExpNumericBackReference` nodes, but the parser now no longer returns `JSRegExpNumericBackReference` nodes for out of bound back-references. 

I think the right fix is to replace this rule with a rule that suggests not to use octal escape sequences and instead use octal escape sequences (I understand that octal escape sequences are officially only supported in browsers). 

What are your thoughts on this?

internal/codec-js-regexp/index.ts
yassere: You're right that we should be able to parse named references in non-unicode patterns.
But if a pattern isn't unicode and it doesn't contain a [GroupName](https://tc39.es/ecma262/multipage/text-processing.html#sec-parsepattern), then `\k` should be treated as a [SourceCharacterIdentityEscape](https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#prod-annexB-SourceCharacterIdentityEscape).
 
So we shouldn't assert that `<` follows `\k` without checking the unicode flag. To be spec-compliant, we could also check for the presence of named groups to produce a diagnostic in the non-unicode case, but I think it's probably not worth it.

internal/codec-js-regexp/index.ts
yassere: I think that just means we shouldn't attempt to interpret `\12` as the decimal escape `\1` followed by a literal `2`. This looks good to me.

internal/codec-js-regexp/index.ts
```diff
@@ -575,6 +572,8 @@ function eatOperator(parser: RegExpParser, op: string): boolean {
 function parseGroupCapture(
 	parser: RegExpParser,
 ): JSRegExpGroupCapture | JSRegExpGroupNonCapture {
+	parser.setState({nCapturingParens: parser.state.nCapturingParens + 1});
```
yassere, line 575: We should only be counting capturing groups here, but this currently also increments for non-capturing groups.

internal/codec-js-regexp/index.ts
yassere: I think this is fine for now.

yassere:
Replacing the `noReferenceToNonExistingGroup` lint rule sounds good to me.

Octal escapes are still being handled incorrectly. The regex `/hello\12/` is valid and matches `”hello\n”`.
Running a `check --apply` using this branch transforms that regex into `/hello1/`.

internal/codec-js-regexp/index.ts
yassere: I don't think a second assertion is necessary. I actually think it's a bit odd that the tokenizer has to insert angle brackets into the token's value because `parseCharacter` expects them, but I think this is fine for now since we're planning on rewriting the regex parser eventually.

internal/codec-js-regexp/index.ts
MichaReiser: Oh, excellent catch!

internal/codec-js-regexp/index.ts
MichaReiser: Ahh, thanks for the links! I was wondering what the meaning of `N` is but wasn't able to find the definition in the spec. That clarifies a lot.

MichaReiser:
Thank you @yassere for the thorough review and for catching a number of bugs. I tried to address all of the comments and we should now be good to go (assuming that a breaking change like removing a lint rule is fine).

yassere (approved):
Looks good to me.

PR #1665, opened 2021-09-22
feat: CLI and formatter crates

ematipico:
TODO:
- [x] documentation
- [x] add tests
- [x] fix existing tests
- [x] lint

## Summary

This the first work we are internally doing for the formatter.

This PR creates a rough API that creates an intermediate IR from nodes. The tokens are heavily inspired from the existing API we have in Rome (TS codebase).

## Test Plan

CI and tests


crates/cli/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+							IndentStyle::Space(size.parse::<u8>().unwrap_or(2))
+						}
+						_ => IndentStyle::Tab,
+					},
+					None => IndentStyle::Tab,
+				};
+				format(input, FormatOptions::new(options));
+			}
+		}
+		Err(err) => err.exit(),
```
jamiebuilds, line 82: Should we also print `--help` here or something?

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+							if !path.exists() {
+								return Err(format!("The file \"{}\" doesn't exist.", value));
+							}
+							Ok(())
+						}),
+				),
+		)
+		.try_get_matches();
+
+	match result_matches {
```
jamiebuilds, line 61: I found a useful `Result` function the other day:

```rs
let matches = result_matches.unwrap_or_else(|error| {
	error.exit();   
});
```

It's basically `unwrap()` with custom error handling that would expect in this case `ArgMatches`. But since `Error.exit()` returns `!` it compiles without returning a new value.

I like this because I find that `match` in Rust leads to some really deeply nested code, which can be bad cause stuff never `Drop`'s, which is annoying for debugging lifetimes.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+		.version(crate_version!())
+		.subcommand(
+			App::new("format")
+				.about("Format a file")
+				.arg(
+					Arg::new("indentStyle")
+						.long("indent-style")
+						.about("The style of indentation")
+						.value_name("tab|space")
+						// .possible_values(&["tabs", "spaces"])
```
jamiebuilds, line 18: Does `.possible_values()` not work here?

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+						}),
+				)
+				.arg(
+					Arg::new("indentSize")
+						.long("indent-size")
+						.about("The size of the indent.")
+						.value_name("NUMBER")
+						.default_value("2")
+						.validator(|value| {
+							let number = value.parse::<u16>();
```
jamiebuilds, line 39: This is parsing as `u16` but you have `u8` below. I think `u8` is probably enough

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+			if let Some(matches) = matches.subcommand_matches("format") {
+				let size = matches.value_of("indentSize");
+				let style = matches.value_of("indentStyle");
+				let input = matches.value_of("input").unwrap();
+				let input = PathBuf::from(&input);
+				let options = match style {
+					Some(style) => match style {
+						"tab" => IndentStyle::Tab,
+						"space" => {
+							let size = size.unwrap_or("2");
```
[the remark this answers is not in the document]
antonk52: The default value would only apply for the indent style, not "space" in particular, wouldn’t it?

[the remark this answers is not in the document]
jamiebuilds: Yeah, I think it adds it to the `--help` output as well

crates/cli/examples/run_cli.rs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// ```bash
+///   cargo run --example run_cli -- --help
+///   cargo run --example run_cli format
+/// ```
+///
+/// To run a valid example:
+///
+/// ```bash
+///  cargo run --example run_cli format examples/input.json
```
proful, line 19: examples are good. I'm able to run the above command with no issues. May be you have few option to pretty format the outputs.

crates/cli/src/lib.rs
MichaReiser: Another alternative would be to put the whole Clap setup and argument parsing in its own function that returns a result, allowing it to use the `?` operator. The main function then deals with handling the Ok and Err case.

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+	IfBreak {
+		break_contents: Content,
+		flat_contents: Content,
+	},
+	/// A literal string, the content will be printed with quotes
+	StringLiteral(String),
+	/// A number
+	Number(u64),
+	/// A generic boolean
+	Boolean(bool),
```
MichaReiser, line 34: I'm not sure if we need `Token`s for numbers and booleans. We should be able to use `Token::string` as internal representation but I agree that it probably makes sense to either provide helpers like `Token::boolean(value: bool)` and `Token::long(u64)` or provide a `impl From<bool> for Token` implementation that does the to string conversion. However, the representation may differ between languages.

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+	},
+	// TODO Revisit, structure is a bit weird
+	IfBreak {
+		break_contents: Content,
+		flat_contents: Content,
+	},
+	/// A literal string, the content will be printed with quotes
+	StringLiteral(String),
+	/// A number
+	Number(u64),
```
MichaReiser, line 32: I expected number to be a f64 because of what I know from JS. May be good to use more specific names.

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+	pub fn new() -> Self {
+		Self { tokens: vec![] }
+	}
+
+	/// Use this utility if you know ahead of time how many tokens you will store
+	pub fn with_capacity(capacity: usize) -> Self {
+		Self {
+			tokens: Vec::with_capacity(capacity),
+		}
+	}
```
MichaReiser, line 69: This is a low-level operation. How about we implement `Extend` that allows merging an `Iterator` into `ConcatTokens` and/or implement `FromIterator` to create `ConcatTokens` from an Iterator.

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
+use crate::intersperse::Intersperse;
+
+type Content = Box<FormatTokens>;
+pub type Tokens = Vec<FormatTokens>;
+
+/// The tokens that are used to apply formatting.
+///
+/// These tokens are language agnostic.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum FormatTokens {
```
MichaReiser, line 10: The rust book use singularity for enum names, e.g. `IpAddrKind` rather than `IpAddrKinds`. 

That makes sense in my view because let's say you have a method that accepts a `FormatToken`.

```
pub fn print(token: &FormatToken) -> String { ... } 
```

The plural form might be misleading in that case, making you assume you get an array of `FormatTokens` where in fact, it's just one.

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	pub fn softline() -> FormatTokens {
+		Self::SOFT_LINE
+	}
+
+	pub fn new_line_or_space() -> FormatTokens {
+		Self::NEW_LINE_OR_SPACE
+	}
+}
```
MichaReiser, line 144: An alternative would be to make the `consts` public or make the function `const`

```
pub const fn hardline() -> FormatTokens {
  FormatTokens ::Line { mode: LineMode::Hard }
}
```

crates/formatter/src/format_json.rs
```diff
@@ -0,0 +1,159 @@
...
+	format_tokens::{ConcatTokens, FormatTokens},
+	FormatValue,
+};
+use serde_json::Value;
+
+impl FormatValue for Value {
+	fn format(&self) -> FormatTokens {
+		match self {
+			Value::String(string) => {
+				FormatTokens::from(vec!["\"".into(), string.as_str().into(), "\"".into()])
```
MichaReiser, line 11: I'm wondering if we should create a `double_quoted_str` helper on `FormatTokens`

crates/formatter/src/format_json.rs
```diff
@@ -0,0 +1,159 @@
...
+			}
+			Value::Number(number) => {
+				let number = number.as_u64().unwrap();
+				FormatTokens::from(number)
```
MichaReiser, line 15: Aren't JSON numbers `f64` since JSON is based on JS?

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+			tokens: Vec::with_capacity(capacity),
+		}
+	}
+
+	pub fn push_token<T: Into<FormatTokens>>(mut self, value: T) -> Self {
+		self.tokens.push(value.into());
+		self
+	}
+
+	pub fn format_tokens(self) -> FormatTokens {
```
MichaReiser, line 76: Nit: I expected that this function formats the tokens. How about `to_format_tokens`?

crates/cli/src/lib.rs
ematipico: That's what `.exit` does. It's part of clap's API

crates/formatter/src/format_json.rs
```diff
@@ -0,0 +1,159 @@
...
+							.push_token(",")
+							// .push_token(FormatTokens::hardline())
+							.format_tokens(),
+					);
+				}
+				ConcatTokens::with_capacity(3)
+					.push_token("{")
+					.push_token(FormatTokens::indent(FormatTokens::from(content)))
+					.push_token("}")
+					.format_tokens()
```
MichaReiser, line 38: A macro could indeed improve the ergonomics a lot. Or we may need to find some other clever way that leverages rust's type system. Maybe something we can look into next week or so.

crates/formatter/src/format_json.rs
```diff
@@ -0,0 +1,159 @@
...
+							.push_token("\"")
+							.push_token(":")
+							.push_token(FormatTokens::Space)
+							.push_token(value.format())
+							.push_token(",")
+							// .push_token(FormatTokens::hardline())
+							.format_tokens(),
+					);
+				}
+				ConcatTokens::with_capacity(3)
```
MichaReiser, line 34: Nit: This might be a bit of an over-optimization. I would expect `Vec` to allocate space for at least a few elements (around 10-16).

crates/formatter/src/format_json.rs
```diff
@@ -0,0 +1,159 @@
...
+					content.push(
+						ConcatTokens::new()
+							.push_token("\"")
+							.push_token(key.as_str())
+							.push_token("\"")
+							.push_token(":")
+							.push_token(FormatTokens::Space)
+							.push_token(value.format())
+							.push_token(",")
+							// .push_token(FormatTokens::hardline())
```
MichaReiser, line 30: We should add a `line_or_space` here so that there's a space if printed flat and a new line if printed across multiple lines.

It should also be possible to use `join` to insert the "," between elements (you currently insert a trailing comma which isn't legal in JSON).

crates/formatter/src/format_json.rs
```diff
@@ -0,0 +1,159 @@
...
+							.format_tokens(),
+					);
+				}
+				ConcatTokens::with_capacity(3)
+					.push_token("{")
+					.push_token(FormatTokens::indent(FormatTokens::from(content)))
+					.push_token("}")
```
MichaReiser, line 37: We should add `softline`s between the curly braces that are omitted when printed `flat` but are emitted if the object doesn't fit on a single line.

crates/formatter/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+mod format_tokens;
+mod intersperse;
+use crate::format_json::json_to_tokens;
+pub use format_tokens::FormatTokens;
+use std::{fs::File, io::Read, path::PathBuf, str::FromStr};
+
+/// This trait should implemented on each node/value that should have a formatted representation
+pub trait FormatValue {
+	fn format(&self) -> FormatTokens;
+}
```
MichaReiser, line 11: I like the idea because it allows static dispatch rather than dynamic dispatch to a formatter. 

How would that work if all our AST's are `Rowan` ASTs. Do we have different AST node types per language (@yassere)?

MichaReiser (approved):
Nice work! I have a few comments of which most can be applied in a later PR or are more inspirational.

crates/formatter/src/format_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+	Group(GroupToken),
+	List {
+		content: Tokens,
+	},
+	// TODO Revisit, structure is a bit weird
+	IfBreak {
+		break_contents: Content,
+		flat_contents: Content,
+	},
+	/// A literal string, the content will be printed with quotes
```
MichaReiser, line 29: We may need to rename this to `raw` or some other name. The containing string doesn't get printed in quotes but instead as is. E.g. `FormatTokens::StringLiteral("[]")` is printed as `[]`, an empty array expression. 

I wouldn't name it `StringLiteral` for that reason because it might be misleading coming from other Token/AST structures.

ematipico:
Thank you all for the feedback. As discussed with @MichaReiser,  I will merge this PR as is and address the comments in https://github.com/rome/tools/pull/1667

This is because myself and @MichaReiser are working closely on the same thing and we think that it would be best to just work on one branch.

crates/cli/src/lib.rs
ematipico: Adding this would trigger clap's validation and their error message. I still need to explore clap's API and see at what extent we're able to customize error messages.

crates/cli/src/lib.rs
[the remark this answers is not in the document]
ematipico: > Clap has a `.default_value("2")` for args

This is already done at line 37.

crates/formatter/src/format_tokens.rs
ematipico: Makes sense! Will apply the change

crates/formatter/src/format_tokens.rs
ematipico: Yes, we should make more granular. It will be useful also in other language.

crates/formatter/src/format_tokens.rs
ematipico: I like the `const` function. So we can change the name of the `consts` without changing the API.

crates/formatter/src/format_json.rs
ematipico: Probably yes. We can explore it later

crates/formatter/src/format_json.rs
ematipico: Yep! Will apply the change

crates/formatter/src/format_tokens.rs
ematipico: `clippy` requires `to_*` functions the have `&mut self` in their signature

PR #1667, opened 2021-09-23
feat: format tokens printer

MichaReiser:
## Summary

Heavily inspired by the Printer implementation of Rome "classic" but adapted to make it work efficiently with Rust.

The most significant changes are:

* Split the state into a global printer state and an arguments object, storing the state-specific to printing a single token.
* Replace throwing Exceptions with `Result`
* Split the flat/normal printing into dedicated functions to make it clear when the printer changes between the two modes.

The printer implementation lives inside of the `printer.rs`. The other changes are (99%) from #1665 on which this PR is based.

## Test Plan

I'll work on some more unit tests but wanted to share the taken approach to get feedback early. 

## Alternatives

 I thought about creating a trait that a token could implement something like this:

```rust
pub trait FormatTokenPrinter {
  pub fn print(&self, printer: &mut Printer) -> Vec<PrintTokenCall>;
  pub fn try_print_flat(&self, printer: &mut Printer) -> Result<Vec<PrintTokenCall>, LineBreakRequiredError>
}

impl Printer {
 pub fn options(&self) -> &PrinterOptions {},
 pub fn state(&self) -> &PrinterState {},
 pub fn state_mut(&mut self) -> &mut PrinterState {}

 pub fn print(mut self) -> PrintResult {...}
 pub fn print_str(&mut self, content: &str) -> () {}
 pub fn try_print_flat(&mut self, token: &Token) -> ... {}
}  
```

It could then be implemented by the different tokens. However, Rust only allows a single implementation for the `FormatTokens` enum which mean we would only be moving the logic from inside the `Printer` to the `FormatTokens`. We could create dedicated structs for some of the `FormatTokens` like `List`, `Group`, `Line` that then could implement the `FormatTokennPrinter` trait and the `FormatTokens` implementation would simply forward the call. 

What I would like about this approach is that the printing logic lives next to the token and avoids a huge `match` inside of the `Printer`. The main downside is that it requires exposing the fact that `Printer` uses its own state with `PrintTokenCall` as well as that we can't make it as ergonomic as I wish (implement the trait for each token variant).


crates/formatter/src/printer.rs
```diff
@@ -0,0 +1,523 @@
...
+			FormatToken::Indent { content } => {
+				vec![PrintTokenCall::new(content, args.with_incremented_indent())]
+			}
+
+			FormatToken::IfBreak { break_contents, .. } => {
+				vec![PrintTokenCall::new(break_contents, args)]
+			}
+
+			FormatToken::Line { .. } => {
+				self.print_str("\n");
```
ematipico, line 113: The string `"\n"` could live a `const`. Also, the new line should come from the options because of how different OS treat new lines

ematipico (approved):

crates/formatter/src/printer.rs
```diff
@@ -0,0 +1,523 @@
...
+		}
+	}
+}
+
+impl Default for PrinterOptions {
+	fn default() -> Self {
+		PrinterOptions {
+			tab_width: 2,
+			print_width: 80,
+			indent_string: String::from("  "),
```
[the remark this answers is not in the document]
MichaReiser: Oh no, all my unit tests will start failing xD

crates/formatter/src/printer.rs
MichaReiser: Moving it to a `const` shouldn't change much from a memory perspective. All constant strings are stored in the global string table of the binary. 

Giving it a name doesn't improve readability in my view. 

I'll add the `endOfLine` option so that it can be specified by the caller.

crates/formatter/src/printer.rs
```diff
@@ -0,0 +1,524 @@
...
+
+				// If the line is too long, break the group
+				if self.state.line_width > self.options.print_width {
+					return Err(LineBreakRequiredError);
+				}
+
+				// If a new line was printed, break the group
+				if current_line != self.state.generated_line {
+					return Err(LineBreakRequiredError);
+				}
```
jamiebuilds, line 165: I don't really know what would be idiomatic in Rust, but just to ask the question: Should this be an error, or should the function return a different enum like:

```rs
enum TryPrintFlatResult {
  Printed(...),
  LineBreakRequired,
}
```

MichaReiser: Using `Result` has the advantage that you can use the `?` operator in the calling code and I thought it's somewhat related to the [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) which also uses `Result`. 

I also thought about using `Option` since we aren't really interested in the error but believe that the explicit error makes it easier to understand why the conversion might fail.

PR #1668, opened 2021-09-23
feat: Add experimental parser using Tree-sitter and rowan

yassere:
## Summary

This PR adds an experimental crate that uses [Tree-sitter](https://github.com/tree-sitter/tree-sitter) to parse source code then converts the parse tree into a [rowan ](https://github.com/rust-analyzer/rowan) syntax tree. 

Tree-sitter outputs a `node-types.json` file when generating a parser. This crate uses that file to auto-generate integration code so that it can convert a Tree-sitter parse tree into a rowan syntax tree for any grammar. If we build good abstractions and infrastructure for analyzing/handling syntax trees in a language-agnostic way, this approach could significantly reduce the barrier for adding new language support within Rome.

The next step is to create an AST layer similar to what's described in the rust-analyzer [syntax guide](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/syntax.md), which will require a mix of auto-generated and hand-written code.

The sourcegen module borrows heavily from [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer).

TODO: Add more comments and tests.


crates/parser/src/languages/ts.rs
```diff
@@ -0,0 +1,693 @@
...
+		SyntaxKind::Whitespace.into()
+	}
+	fn new_root(&self, green: rowan::GreenNode) -> Self::SyntaxNode {
+		SyntaxNode::new_root(green)
+	}
+}
+impl From<u16> for SyntaxKind {
+	#[inline]
+	fn from(d: u16) -> SyntaxKind {
+		assert!(d <= (SyntaxKind::__LAST as u16));
```
MichaReiser, line 668: Since this is generated code: How about we declare the number of `SyntaxKind` nodes for the language at the top of the file. The benefit would be that match expressions don't need to handle `__LAST` to be exhaustive.

crates/parser/src/languages/tsx.rs
```diff
@@ -0,0 +1,691 @@
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(u16)]
+pub enum SyntaxKind {
```
MichaReiser, line 3: How would we share code between TS and TSX since both languages use different `SyntaxKind` types?

crates/parser/src/lib.rs
```diff
@@ -1 +1,128 @@
...
+	parser: tree_sitter::Parser,
+	language: L,
+}
+
+impl<L: ParserLanguage> Parser<L> {
+	pub fn new(language: L) -> Result<Self> {
+		let tree_sitter_language = match language.name() {
+			"tsx" => tree_sitter_typescript::language_tsx(),
+			"ts" => tree_sitter_typescript::language_typescript(),
+			name => return Err(anyhow!("Invalid syntax name: {}", name)),
```
MichaReiser, line 39: Should we generate a `Language` enum that we use in our APIs and can be converted from/to the tree sitter enum? Passing strings feels a bit fragile.

crates/parser/src/lib.rs
```diff
@@ -1 +1,128 @@
...
+	// Uses tree-sitter's cursor api to walk the parse tree and build a rowan green tree.
+	// The cursor goto methods attempt to move the cursor and return true on success.
+	fn build_green_tree(&self, src: &str, tree: tree_sitter::Tree) -> GreenNode {
+		let mut builder = GreenNodeBuilder::new();
+		let mut cursor = tree.walk();
+		let mut last_token_byte = 0;
+
+		loop {
+			let node = cursor.node();
+			let node_kind = self.language.get_kind(node.kind(), node.is_named());
```
MichaReiser, line 66: Is the `node` type a tree sitter generated type? It's a bit confusing that we call `get_kind` and pass in `kind` as the first argument. Can we rename the method to `type_name` or `kind_name` if it isn't a tree sitter generated type?

crates/parser/src/lib.rs
```diff
@@ -1 +1,128 @@
...
+			let kind = SyntaxKind(node_kind);
```
MichaReiser, line 67: Nit: Import `SyntaxKind` as `RowanSyntaxKind`. I got confused about why we need to wrap `SyntaxKind` once more.

crates/parser/src/sourcegen.rs
```diff
@@ -0,0 +1,288 @@
...
+
+			pub type SyntaxNode = rowan::SyntaxNode<#lang>;
+			pub type SyntaxToken = rowan::SyntaxToken<#lang>;
+			pub type SyntaxElement = rowan::SyntaxElement<#lang>;
+			pub type SyntaxNodeChildren = rowan::SyntaxNodeChildren<#lang>;
+			pub type SyntaxElementChildren = rowan::SyntaxElementChildren<#lang>;
+
+
+
+
```
MichaReiser, line 188: Nit: Delete additional empty lines?

MichaReiser (commented):
That's pretty cool! Impressive work @yassere 

I've had a few comments for smaller or future improvements as well as some questions. But looks all good to me!

crates/parser/Cargo.toml
```diff
@@ -4,7 +4,16 @@ version = "0.0.0"
 description = ""
 license = "MIT"
 edition = "2018"
+default-run = "parser"
 
 [dependencies]
-# tree-sitter = "0.17"
-# rowan = "0.13.2"
+anyhow = "1"
```
ematipico, line 10: Should we avoid using, for now, third party crates that handle errors? At least until we figure out the whole architecture of our the diagnostics

crates/parser/src/languages/ts.rs
```diff
@@ -0,0 +1,693 @@
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
```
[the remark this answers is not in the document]
ematipico: I second that, this file should not be modified manually (basically what we did with the old version of Rome)

crates/parser/src/languages/ts.rs
ematipico: Maybe we should use [`debug_assert`](https://doc.rust-lang.org/std/macro.debug_assert.html), unless it is intentional.

crates/parser/src/lib.rs
```diff
@@ -1 +1,128 @@
-// parser crate
+//! Experimental parser using [tree_sitter] and [rowan]
```
ematipico, line 1: We should add links to `tree_sitter` and `rowan`

crates/parser/src/sourcegen.rs
```diff
@@ -0,0 +1,288 @@
+use anyhow::Result;
```
ematipico, line 1: IMHO, I think we should move the source generation out of the parser crate. Maybe have them inside a specific folder called `scripts`. This would slim down the dependencies of the crate parser.

ematipico (commented):
This is WOW! Amazing stuff! Left some comments

crates/parser/src/languages/ts.rs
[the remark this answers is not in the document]
bnjjj: Hello, my 2 cents but maybe it could be more obvious too if you name this file like `ts_generated.rs` and keep this kind of notation when you have a generated file ?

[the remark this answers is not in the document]
bnjjj: Or having a submodule for each language and inside a `mod.rs` and  `generated.rs` and only export what you want with the right names from `mod.rs`

crates/parser/src/languages/ts.rs
```diff
@@ -0,0 +1,693 @@
...
+#[repr(u16)]
+pub enum SyntaxKind {
+	PrimaryType,
+	Declaration,
+	Expression,
+	Pattern,
```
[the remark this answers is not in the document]
yassere: It should be stable based on the order of `node-types` but we could still sort it.

crates/parser/src/languages/ts.rs
```diff
@@ -0,0 +1,693 @@
...
+	LbraceToken,
+	LbracePipeToken,
+	PipeToken,
+	PipeEqToken,
+	PipePipeToken,
+	PipePipeEqToken,
+	PipeRbraceToken,
+	RbraceToken,
+	TildeToken,
+	Whitespace,
```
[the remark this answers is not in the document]
yassere: Currently, a whitespace token will often contain mixed whitespace like `"\n\t"`. You can get the `text()` of a token, but we could also implement some helper methods to identify types of whitespace.

crates/parser/src/languages/ts.rs
yassere: This is copied from `rust-analyzer` and the run-time assertion is presumably intentional to guarantee the safety of the transmute. The `__LAST` variant could be replaced with a generated const `u16`, but it feels slightly less fool-proof because there are variants manually included in this enum that don't come 1:1 from `node-types` and that would need to be kept in sync. I doubt this enum will ever be used in a match expression without a catchall.

crates/parser/src/languages/tsx.rs
yassere: That's definitely a problem that needs to be solved and part of why this approach is experimental. Ideally, it might be possible to have smart enough codegen to create a single universal `SyntaxKind` enum that allows for overlapping types between languages but with language-prefixed variants where necessary (`JSIdentifier` vs `CSSIdentifier`). That was the original goal, but I don't know yet if it's practical.

crates/parser/src/lib.rs
yassere: This will definitely be cleaned up for aesthetics, but this is matching against a `&'static str` that's baked into the codegen for each language, so I don't think it's particularly fragile. I certainly agree that we shouldn't ever write functions that accept string parameters where strongly-typed alternatives are possible.

In the current API, this is how you create a parser for a specific language: `Parser::new(languages::TSX)`

crates/parser/src/lib.rs
yassere: Yeah, there are a number of things in this draft that should get renamed for clarity.

crates/parser/Cargo.toml
yassere: I think that would slow down our exploratory prototyping. There are a number of dependencies we'll be able to prune later, but I don't think we should be overly concerned about that until we start stabilizing crates.

crates/parser/src/sourcegen.rs
yassere: Yeah, this will get moved out of the parser crate in the future because it will also be used to generate the AST code for the syntax crate.

crates/parser/src/languages/ts.rs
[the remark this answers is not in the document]
yassere: Yeah, that was part of "TODO: Add more comments".

crates/parser/src/lib.rs
yassere: Actually, I think I misunderstood your comment at first. You're right, there should be a single enum of all languages.

crates/parser/src/languages/ts.rs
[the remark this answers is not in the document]
yassere: @bnjjj Yeah, it's a good idea to make it clear from the naming which files are generated. I'm going to put more thought into the file structure once I have a better idea of how much codegen there will be for each language.

jamiebuilds (approved):

yassere:
Thanks everyone for the feedback! I'll merge this in for now and make additional improvements when I add the AST codegen next week.

PR #1672, opened 2021-09-24
feat(formatter): Support different line endings

MichaReiser:
## Summary
Adds support for different line endings to the formatter's printer so that it emit's the users preferred line ending. 

Rome uses line feeds internally to push the complexity of having different line endings to the `Printer`, that then converts the line feeds to the user preferred ending.

## Test Plan

Added a unit test.

### DIFF
--- a/crates/formatter/src/format_token.rs
+++ b/crates/formatter/src/format_token.rs
@@ -1,4 +1,5 @@
 use crate::intersperse::Intersperse;
+use std::ops::Deref;
 
 type Content = Box<FormatToken>;
 pub type Tokens = Vec<FormatToken>;
@@ -17,8 +18,7 @@ pub enum FormatToken {
 	List(ListToken),
 	// TODO Revisit, structure is a bit weird
 	IfBreak(IfBreakToken),
-	/// A string that will be printed as is.
-	String(String),
+	String(StringToken),
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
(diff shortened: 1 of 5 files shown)


crates/formatter/src/lib.rs
```diff
@@ -36,7 +43,7 @@ impl FromStr for IndentStyle {
 }
 
 // TODO: evaluate to remove it
```
ematipico, line 45: We should remove the `TODO` now, the comment was there in the `Debug` macro

crates/formatter/src/format_token.rs
```diff
@@ -153,6 +161,48 @@ impl GroupToken {
...
+///
+/// assert_eq!(r#"["a", "b"]"#, formatted.code());
+/// ```
+/// Is not allowed to contain any new lines
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct StringToken(String);
+
+impl StringToken {
+	pub fn new(content: &str) -> Self {
+		debug_assert!(!content.contains('\r'), "The content '{}' contains a carriage return '\\r' character but string tokens must only use line feeds '\\n' as line separator. Use '\\n' instead of '\\r' and '\\r\\n' to insert a line break in strings.", content);
```
ematipico, line 181: Is this message for the developers or for the users? If it's for the users, maybe we should add a `// TODO` as reminder to have diagnostics.

ematipico (approved):

crates/formatter/src/format_token.rs
MichaReiser: It's for devs, as it's also only printed in debug builds (maybe we should change that?). 

The idea is to warn devs if they create an invalid `StringToken` that uses `\r\n` because it messes up the output (we may end up with `\r\r\n`).

PR #1673, opened 2021-09-24
fix: trailing comma in JSON formatter

ematipico:
## Summary

This PR fixes a small bug in printing a JSON where a trailing comma was added at the end of the last property.

## Test Plan

Updated current tests

### DIFF
--- a/crates/cli/examples/input.json
+++ b/crates/cli/examples/input.json
@@ -1,6 +1,8 @@
 {
 	"string": "foo",
-	"boolean": false,
-	"number": 15,
-	"object": { "something": 15 }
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):
Thanks!

> This PR fixes a small bug in printing a JSON where a leading comma was added at the end of the last property.

a ~~leading~~ **trailing** comma was added at the end of the last property.

The title is a bit confusing. What does it have to do with the test command?

ematipico:
Yeah sorry, there was a commit that was not supposed to be there and it created a wrong title.

PR #1674, opened 2021-09-24
feat(formatter): Format JSON Arrays

MichaReiser:
## Summary

Adds support for formatting JSON arrays.

## Test Plan

Added a unit test for the tokenization of JSON arrays.

Added an array to the `input.json` and ran `cargo run --example run_cli format .\crates\cli\examples\input.json`. It properly formatted the array.

Creating integration tests is something else I'm investigating but shouldn't block this PR.

### DIFF
--- a/crates/formatter/src/format_json.rs
+++ b/crates/formatter/src/format_json.rs
@@ -41,7 +41,24 @@ impl FormatValue for Value {
 				]))
 			}
 			Value::Null => FormatToken::string("null"),
-			Value::Array(_) => todo!("Implement array"),
+			Value::Array(array) => {
+				let separator = FormatToken::concat(vec![
+					FormatToken::string(","),
+					FormatToken::Line(LineToken::soft_or_space()),
+				]);
+
+				let elements = vec![
+					FormatToken::Line(LineToken::soft()),
+					FormatToken::join(separator, array.iter().map(|element| element.format())),
+				];
+
+				FormatToken::Group(GroupToken::new(vec![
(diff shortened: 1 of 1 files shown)


ematipico (approved):

crates/formatter/src/format_json.rs
```diff
@@ -41,7 +41,24 @@ impl FormatValue for Value {
...
+					FormatToken::Line(LineToken::soft()),
+					FormatToken::join(separator, array.iter().map(|element| element.format())),
+				];
+
+				FormatToken::Group(GroupToken::new(vec![
+					FormatToken::string("["),
+					FormatToken::indent(elements),
+					FormatToken::Line(LineToken::soft()),
+					FormatToken::string("]"),
+				]))
```
jamiebuilds, line 60: Just to capture from this morning, we can use:

```rs
use FormatToken::*;
```

To make this easier to write:

```rs
let separator = concat(vec![
  string(","),
  Line(LineToken::soft_or_space()),
]);

let elements = vec![
  Line(LineToken::soft()),
  join(separator, array.iter().map(|element| element.format())),
];

Group(GroupToken::new(vec![
  string("["),
  indent(elements),
  Line(LineToken::soft()),
  string("]"),
]))
```

MichaReiser: That makes sense. I'll leave it as is for now since @ematipico and I are revisiting the `FormatToken` construction as part of a separate PR in which we can apply the same approach across all files.

PR #1675, opened 2021-09-24
refactor(cli,formatter): more idiomatic rust

bnjjj:
## Summary
Some refactoring to let me dig into the project

## Test Plan
Add an e2e test for the cli format command

### DIFF
--- a/crates/cli/src/bin/cli.rs
+++ b/crates/cli/src/bin/cli.rs
@@ -1,0 +1,24 @@
+use cli::run_cli;
+
+///
+/// To run this example, run:
+///
+/// ```bash
+/// cargo run --example run_cli
+/// ```
+/// Add arguments like:
+///
+/// ```bash
+///   cargo run --example run_cli -- --help
+///   cargo run --example run_cli format
+/// ```
+///
+/// To run a valid example:
+///
+/// ```bash
+///  cargo run --example run_cli format examples/input.json
+/// ```
+///
+fn main() {
+	run_cli();
+}
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -27,7 +27,7 @@ pub fn run_cli() {
 						.default_value("2")
 						.validator(|value| {
 							value
-								.parse::<u16>()
+								.parse::<u8>()
 								.map_err(|_| "Invalid indent-size value. Try using a number")
 						}),
 				)
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -1,45 +1,34 @@
-use clap::{crate_version, App, Arg};
+use clap::{crate_version, App, AppSettings, Arg};
 use rome_formatter::{format, FormatOptions, IndentStyle};
-use std::path::PathBuf;
+use std::{path::PathBuf, str::FromStr};
 
 /// Main function to run Rome CLI
 pub fn run_cli() {
 	let matches = App::new("rome")
 		.about("The official Rome CLI")
 		.version(crate_version!())
+		.setting(AppSettings::SubcommandRequiredElseHelp)
 		.subcommand(
 			App::new("format")
 				.about("Format a file")
 				.arg(
-					Arg::new("indentStyle")
+					Arg::new("indent_style")
 						.long("indent-style")
 						.about("The style of indentation")
 						.value_name("tab|space")
 						.default_value("tab")
-						.validator(|value| {
-							if !value.is_empty() {
-								if value.eq("tab") || value.eq("space") {
-									return Ok(());
-								}
-								return Err(
-									r#"Invalid indent-style value.  Only "tabs" and "spaces" are supported. "#,
-								);
-							}
-							Ok(())
-						}),
+						.validator(|value| IndentStyle::from_str(value).map(|_| ())),
 				)
 				.arg(
-					Arg::new("indentSize")
+					Arg::new("indent_size")
 						.long("indent-size")
 						.about("The size of the indent.")
 						.value_name("NUMBER")
 						.default_value("2")
 						.validator(|value| {
-							let number = value.parse::<u8>();
-							match number {
-								Ok(_) => Ok(()),
-								Err(_) => Err("Invalid indent-size value. Try using a number"),
-							}
+							value
+								.parse::<u16>()
+								.map_err(|_| "Invalid indent-size value. Try using a number")
 						}),
 				)
 				.arg(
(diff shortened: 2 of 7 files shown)


crates/cli/src/lib.rs
```diff
@@ -55,28 +44,36 @@ pub fn run_cli() {
...
+		Some(("format", matches)) => {
+			let size = matches.value_of("indent_size");
+			let style = matches.value_of("indent_style");
+			let input = matches.value_of("input").unwrap();
+			let input = PathBuf::from(&input);
+			let options: IndentStyle = style
+				.map(|s| match s {
+					"tab" => IndentStyle::Tab,
+					"space" => {
+						let size = size.unwrap_or("2");
```
ematipico, line 63: The `unwrap_or` should not be needed because we've set a `default_value` before, right?

crates/cli/src/lib.rs
```diff
@@ -1,45 +1,34 @@
...
 						.value_name("NUMBER")
 						.default_value("2")
 						.validator(|value| {
-							let number = value.parse::<u8>();
-							match number {
-								Ok(_) => Ok(()),
-								Err(_) => Err("Invalid indent-size value. Try using a number"),
-							}
+							value
+								.parse::<u16>()
```
ematipico, line 30: Is there a reason why you changed from `u8` to `u16`? We had `u16` before but thought that `u8` was enough to store this value.

crates/formatter/src/lib.rs
```diff
@@ -53,18 +53,19 @@ impl FormatOptions {
 		Self { indent_style }
 	}
 }
-// TODO: implement me
+// TODO: implement me + handle errors
 /// Main function
 pub fn format(path: PathBuf, options: FormatOptions) {
```
ematipico, line 58: This API will change, it was created for the sake of a demo. Thank you anyway!

crates/cli/src/lib.rs
```diff
@@ -1,45 +1,34 @@
...
-								if value.eq("tab") || value.eq("space") {
-									return Ok(());
-								}
-								return Err(
-									r#"Invalid indent-style value.  Only "tabs" and "spaces" are supported. "#,
-								);
-							}
-							Ok(())
-						}),
+						.validator(|value| IndentStyle::from_str(value).map(|_| ())),
```
ematipico, line 20: What `.map(|_| ())` does in this case?

crates/cli/src/bin/cli.rs
```diff
@@ -0,0 +1,24 @@
+use cli::run_cli;
```
ematipico, line 1: Do we need to have a binary here in order to run the integration test?

crates/cli/src/lib.rs
```diff
@@ -55,28 +44,36 @@ pub fn run_cli() {
...
+		Err(err) => err.exit(),
+	};
+
+	match subcommand_matches {
+		Some(("format", matches)) => {
+			let size = matches.value_of("indent_size");
+			let style = matches.value_of("indent_style");
+			let input = matches.value_of("input").unwrap();
+			let input = PathBuf::from(&input);
+			let options: IndentStyle = style
```
ematipico, line 59: What's the benefit of using `.map` as opposed of `match {}`?

ematipico (approved):
The code will change but having a integration test set up will really helpful! Thank you!

crates/formatter/src/printer.rs
```diff
@@ -85,9 +85,12 @@ impl PrintResult {
 	}
 }
 
-/// Error returned if printing an item as a flat string fails because it either contains
-/// explicit line breaks or would otherwise exceed the specified line width.
-struct LineBreakRequiredError;
+#[derive(Debug)]
+pub enum PrinterError {
```
MichaReiser, line 89: Nice work overall. 

Could you elaborate why this is more idiomatic? I can see that this becomes useful if there are multiple error conditions but there’s only one and I can’t think of another reason why flat printing should fail than that a line break is required. That’s why I would omit the enum for now until there are more error kinds (should be easy to change since everything is internal)

crates/cli/src/lib.rs
bnjjj: Oups I think it's coming from a rebase error. I refactored the code and in the meantime you pushed on main :p

crates/cli/src/lib.rs
bnjjj: To directly return the value as the closure is waiting for a Result<()>

crates/cli/src/lib.rs
bnjjj: Get rid of one match. As we don't match on the option we only have 1 match and less indentation. Do you want me to rollback this change ?

crates/cli/src/lib.rs
bnjjj: Indeed. I just try to do not use .unwrap() as much as I can. As often clippy suggest me to use something else

crates/formatter/src/lib.rs
bnjjj: Oups my bad

crates/formatter/src/printer.rs
bnjjj: Yes indeed the main reason was just to have an easier way to add other kind of error. Sorry I didn't think too much about what other kind of error we could have. Will rollback as soon as I come back on my computer :)

crates/cli/src/bin/cli.rs
bnjjj: Yes the idea is to be able to have the ability to generate a binary with cargo or test it not only with example but with a real binary in build release mode

ematipico: FYI the real binary will leave in a different crate

bnjjj: Ok I let it there for now just to keep the e2e test

crates/cli/src/lib.rs
ematipico: Nope, it's fine! Thank you

crates/formatter/src/lib.rs
```diff
@@ -53,18 +53,19 @@ impl FormatOptions {
...
 	println!(
-		"Running formatter to: \n- file {:?} \n- with options {:?}",
+		"Running formatter to:\n- file {:?}\n- with options {:?}",
 		path, options.indent_style
 	);
 	// we assume that file exists
-	let mut file = File::open(&path).unwrap();
+	let mut file = File::open(&path).expect("cannot open the file to format");
```
MichaReiser, line 64: Ah, it's called `expect`. I knew a function to add a custom error message must exist but I wasn't able to find it after checking all `unwrap` functions. Now I know :)

MichaReiser (approved):

PR #1676, opened 2021-09-27
test: Formatter integration tests

MichaReiser:
Adds a few integration tests to the formatter. The tests discovered the following two issues that got fixed as well:

* Add a line ending at the end of JSON files (we may want to make this configurable in the future)
* String escaping.

The testing functionality is intentionally kept minimal.

* It currently doesn't use any snapshotting framework like [insta](https://insta.rs/). It seems sufficient for now to handwrite the expected code, which should only change rarely. We can research various snapshotting libraries (and maybe use our own infrastructure once build for Rome) if we see the need in the future.
* Rust's default test runner doesn't support dynamic test discovery based on external files. It's, therefore, required that all tests are declared ahead of compile time. We may want to generate the `spec_tests.rs` file in the near future, but it should be sufficient for now to manually add the test cases.

The added JSON test cases have been copied [from prettier](https://github.com/prettier/prettier/tree/main/tests/format/json) (some alternations were required to remove yet unsupported features that would require a proper JSON parser like retaining character escape sequences, number formatting (`e`/`E` notation). 

## Test Plan

`cargo test`


crates/formatter/tests/spec_tests.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[test]
+pub fn json_boolean() {
+	spec_test::spec_test("json/boolean");
+}
+
+#[test]
+pub fn json_null() {
+	spec_test::spec_test("json/null");
+}
```
MichaReiser, line 11: It should be straight forward to generate this file in the future

crates/formatter/src/format_json.rs
```diff
@@ -63,15 +65,27 @@ impl FormatValue for Value {
 	}
 }
 
+fn escape_string(string: &str) -> String {
+	string
```
MichaReiser, line 69: I benched this vs `Regex::replace_all` which always ended up to be about 10 times slower.

crates/formatter/src/format_json.rs
```diff
@@ -63,15 +65,27 @@ impl FormatValue for Value {
 	}
 }
 
+fn escape_string(string: &str) -> String {
```
ematipico, line 68: Sounds like we will need this function for a bunch of stuff. Maybe we could move it inside a `utils.rs` file

crates/formatter/tests/spec_test.rs
```diff
@@ -0,0 +1,57 @@
+use rome_formatter::{format_str, FormatOptions};
+use std::fs;
+use std::path::PathBuf;
+
+/// Tests that formatting a given input file results in the expected formatted output.
```
ematipico, line 5:
```suggestion
/// Tests that format a given input file, results in the expected formatted output.
```

crates/formatter/tests/spec_test.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// The input and expected output are stored as dedicated files in the `tests/specs` directory where
```
ematipico, line 7:
```suggestion
/// The input and the expected output are stored as dedicated files in the `tests/specs` directory where
```

crates/formatter/tests/spec_test.rs
```diff
@@ -0,0 +1,57 @@
...
+/// the input file name is `{spec_name}.json` and the output file name is `{spec_name.expected}`.
```
ematipico, line 8:
```suggestion
/// the input file name is `{spec_name}.json` and the output file name is `{spec_name.expected}.json`.
```

crates/formatter/tests/spec_test.rs
```diff
@@ -0,0 +1,57 @@
+use rome_formatter::{format_str, FormatOptions};
```
ematipico, line 1: As to-do, I think we should move this `spect_test.rs` in some crate. Probably we will need to use for other stuff.

ematipico (approved):
Looks good to me! Left few suggestions

crates/formatter/src/format_json.rs
MichaReiser: The `escaping` is likely going to be specific to JSON. I don't know if we can re-use the same logic in JavaScript

crates/formatter/tests/spec_test.rs
MichaReiser: Yes, we should do that once we see more use cases.

PR #1677, opened 2021-09-27
chore: Use rome-crate naming for all crates

MichaReiser:
Rename the cli and parser crate to match the rome-{create-name} format and avoid
conflicts with other crates.

### Test plan

```
cargo build
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,14 +78,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cli"
-version = "0.0.0"
-dependencies = [
- "clap",
- "rome-formatter",
-]
-
-[[package]]
 name = "countme"
 version = "2.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -166,22 +158,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6acbef58a60fe69ab50510a55bc8cdd4d6cf2283d27ad338f54cb52747a9cf2d"
 
 [[package]]
-name = "parser"
-version = "0.0.0"
-dependencies = [
- "anyhow",
- "indexmap",
- "proc-macro2",
- "quote",
- "rowan",
- "serde",
- "serde_json",
- "tree-sitter",
- "tree-sitter-typescript",
- "xshell",
-]
-
-[[package]]
 name = "proc-macro-error"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


ematipico:
What kind of conflicts could we encounter? `rust-analyzer` repository (which is the one that we follow) doesn't follow this convention

MichaReiser:
> 
> 
> What kind of conflicts could we encounter? `rust-analyzer` repository (which is the one that we follow) doesn't follow this convention

Hmm, my assumption was that publishing to `crates.io` requires that all package names are unique but that may then not be required (I wasn't able to find any information whatever a workspace creates a scope or not)? 

Anyhow, I still suggest following a naming strategy for our crates (e.g. the formatted is called `rome-formatter`). The question is whatever to use the `rome-` prefix or not. 

My assumption is that we also want to publish the `parse` and `formatter` packages so that people can build on top of these packages. So these package names must be unique, right? Or do we plan to only publish the `rome` crate that then calls into the different sub-crates?

ematipico:
> My assumption is that we also want to publish the parse and formatter packages so that people can build on top of these packages. So these package names must be unique, right? Or do we plan to only publish the rome crate that then calls into the different sub-crates?

We had that plan when we were writing in TypeScript. I think the statement is still valid!

MichaReiser:
> We had that plan when we were writing in TypeScript. I think the statement is still valid!

Do you mean that people can use subsets of rome? In that case, it would make sense to rename the crates to all use `rome-*`.

ematipico:
Yes. `crates.io` doesn't seem to have a scoping system which `npm` has - and we would have used that - so I think prefixing the crates is the best option at the moment.

ematipico (approved):

yassere:
The current structure is based on [rust-analyzer](https://matklad.github.io/2021/08/22/large-rust-workspaces.html) and the idea is that `/crates` is intended for internal packages that won't be published as libraries. We can certainly still choose to use a different structure, but I don't think a prefix is necessary for now.

jamiebuilds:
Rust analyzer does publish everything actually, they have a transform that renames all the crates to prefix them with [`ra_ap_`](https://crates.io/search?page=1&per_page=10&q=ra_ap_). (I have no idea what `ap` stands for there

MichaReiser:
@jamiebuilds `ap` for `auto prefix`? 

@yassere in that case it would make sense for us to rename all crates to not use the `rome-` prefix. 

However, I would still favour prefixing our crates because I don't think any of our crates are currently internal only and we can come up with a proper directory structure once we must distinguish between internal and public crates. Renaming the crates later will only introduce more changes (and more surfaces to test). However, I could also see how we could reason in the opposite direction where changing the crate naming know makes limited sense since we don't know which packages will become public and which stay internal only (in which case we could assume private by default for now and rename the `formatter` crate.

yassere:
@jamiebuilds I'm not sure why `rust-analyzer` publishes everything in practice, but I don't think that any of the packages from `/crates` are intended for external use since they all have version `"0.0.0"` in the repo.

@MichaReiser I actually think that all of our packages are currently internal only, and that we'd eventually want to put careful thought into exactly what we stabilize as public APIs. In pre-Rust Rome, most of our packages lived in `/internal` and we'd yet to put much into `/public-packages`.

I think we should assume everything is private by default and name things accordingly.

MichaReiser:
I'm fine with assuming everything is private for now. Would you @yassere  approve if I rename `rome_formatter` to `formatter` for now?

MichaReiser:
Abandoning in favor of a new PR

PR #1678, opened 2021-09-27
feat: format macro

ematipico:
## Summary

This PR implements a `format_tokens` declarative macro. The name of the macro is still up in the air and needs to be consolidated because we have APIs that have similar names.

For example, we have the API called `format_token` (singular) which is the main function exposed by the `rome_formatter` crate. 

I think we can consolidate the name of the APIs later, once this crate becomes more stable.

I created documentation and doc test, which helped to unveil some bug around some struct that were not exposed by default.

I created an alias called `cargo lint` which basically runs `cargo clippy` with all the arguments we have in the CI. This should help the local development.

## Test Plan

```
cargo test
cargo doc
cargo lint
```

### DIFF
--- a/crates/cli/tests/main.rs
+++ b/crates/cli/tests/main.rs
@@ -29,6 +29,7 @@ fn test_format_cli() {
 - file "fixtures/input.json"
 - with options Tab
 {"string": "foo", "boolean": false, "number": 15, "object": {"something": 15}}
+
 "#
 	);
 }
--- a/crates/formatter/src/format_token_macro.rs
+++ b/crates/formatter/src/format_token_macro.rs
@@ -1,0 +1,66 @@
+/// The macro `format_tokens` is a convenience macro to
+/// use when writing a list of tokens that should be at the same level
+/// without particular rule.
+///
+/// # Examples
+///
+/// Let's suppose you need to create tokens for the string `"foo": "bar"`,
+/// you would write:
+///
+/// ```rust
+/// use rome_formatter::{format_tokens, FormatToken};
+/// let token = format_tokens!("foo", ":", FormatToken::Space, "bar");
+/// ```
+///
+/// The macro can be also nested, although the macro needs to be decorated with the token you need.
+/// For example, let's try to format following string:
+///
+/// ```no_rust
+/// "foo": {
+///   "bar": "lorem"
+/// }
+/// ```
+/// You would write it like the following:
+///
+/// ```rust,no_test
+/// use rome_formatter::{format_tokens, IndentToken, LineToken, FormatToken};
+/// let token = format_tokens!(
+///   "foo",
+///   ":",
+///   IndentToken::new(
+///     format_tokens!(LineToken::hard(), "bar", ":", FormatToken::Space, "lorem")
+///   ),
+///   "}"
+/// );
+/// ```
+#[macro_export]
+macro_rules! format_tokens {
+
+	// called for things like format_tokens!("hey")
+	($token:ident) => {
+		{
+			use rome_formatter::FormatToken;
+			FormatToken::from($token)
+		}
+	};
+
+	($($token:literal),+ $(,)?) => {{
+		use rome_formatter::{FormatToken, ListToken};
+		FormatToken::from(ListToken::concat(vec![
+			$(
+					 FormatToken::from($token)
+			),+
+
+		]))
+	}};
+
+	( $( $token:expr ),+ $(,)?) => {{
+		use rome_formatter::{FormatToken, ListToken};
+		FormatToken::from(ListToken::concat(vec![
+			$(
+					 FormatToken::from($token)
+			),+
+
+		]))
+	}};
+}
--- a/crates/formatter/src/format_token_macro.rs
+++ b/crates/formatter/src/format_token_macro.rs
@@ -22,7 +22,7 @@
 /// ```
 /// You would write it like the following:
 ///
-/// ```rust, no_test
+/// ```rust
 /// use rome_formatter::{format_tokens, IndentToken, LineToken, FormatToken};
 /// let token = format_tokens!(
 ///   "foo",
@@ -32,35 +32,42 @@
 ///   ),
 ///   "}"
 /// );
+///
+/// let inner = FormatToken::concat(vec![
+///     FormatToken::from(LineToken::hard()),
+///     FormatToken::string("bar"),
+///     FormatToken::string(":"),
+///     FormatToken::Space,
+///     FormatToken::string("lorem")
+/// ]);
+/// let outer = FormatToken::concat(vec![
+///     FormatToken::string("foo"),
+///     FormatToken::string(":"),
+///     FormatToken::indent(inner),
+///     FormatToken::string("}")
+/// ]);
+///
+/// assert_eq!(token, outer);
+///
 /// ```
 #[macro_export]
 macro_rules! format_tokens {
 
 	// called for things like format_tokens!("hey")
-	($token:ident) => {
+	($token:expr) => {
 		{
-			use rome_formatter::FormatToken;
+			use $crate::FormatToken;
 			FormatToken::from($token)
 		}
 	};
 
-	($($token:literal),+ $(,)?) => {{
-		use rome_formatter::{FormatToken, ListToken};
-		FormatToken::from(ListToken::concat(vec![
-			$(
-					 FormatToken::from($token)
-			),+
-
-		]))
-	}};
-
 	( $( $token:expr ),+ $(,)?) => {{
-		use rome_formatter::{FormatToken, ListToken};
-		FormatToken::from(ListToken::concat(vec![
+		use $crate::{FormatToken, ListToken};
+		FormatToken::concat(vec![
 			$(
 					 FormatToken::from($token)
 			),+
 
-		]))
+		])
 	}};
 }
--- a/crates/formatter/src/format_token_macro.rs
+++ b/crates/formatter/src/format_token_macro.rs
@@ -64,4 +64,3 @@ macro_rules! format_tokens {
 		]))
 	}};
 }
-
--- a/crates/formatter/src/format_token_macro.rs
+++ b/crates/formatter/src/format_token_macro.rs
@@ -9,7 +9,7 @@
 ///
 /// ```rust
 /// use rome_formatter::{format_tokens, FormatToken};
-/// let token = format_tokens!("foo", ":", FormatToken::Space, "bar");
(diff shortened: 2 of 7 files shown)


crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,66 @@
...
+/// );
+/// ```
+#[macro_export]
+macro_rules! format_tokens {
+
+	// called for things like format_tokens!("hey")
+	($token:ident) => {
+		{
+			use rome_formatter::FormatToken;
+			FormatToken::from($token)
```
bnjjj, line 43: I think it's better to use full path and use `$crate` like `$crate::FormatToken::from($token)` I don't remind why and in which case it could be problematic but I remember I already had this kind of cases

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,66 @@
...
+	};
+
+	($($token:literal),+ $(,)?) => {{
+		use rome_formatter::{FormatToken, ListToken};
+		FormatToken::from(ListToken::concat(vec![
+			$(
+					 FormatToken::from($token)
+			),+
+
+		]))
```
MichaReiser, line 54: You can use `FormatToken::concat` which does some additional optimizations.

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,66 @@
...
+	( $( $token:expr ),+ $(,)?) => {{
+		use rome_formatter::{FormatToken, ListToken};
+		FormatToken::from(ListToken::concat(vec![
+			$(
+					 FormatToken::from($token)
+			),+
+
+		]))
+	}};
+}
```
MichaReiser, line 66: Can we add some doctests that show the different use cases and validate, that the macro works as we expect?

crates/formatter/src/lib.rs
```diff
@@ -1,16 +1,64 @@
...
+//!     fn format(&self) -> FormatToken {
+//!         format_tokens!(self.key.as_str(), FormatToken::Space, "=>", FormatToken::Space, self.value.as_str())
+//!     }
+//! }
+//!
+//! fn my_function() {
+//!     let key_value = KeyValue { key: String::from("lorem"), value: String::from("ipsum") };
+//!     let token = key_value.format();
+//!     let options = FormatOptions::default();
+//!     let result = format_token(&token, options);
```
MichaReiser, line 38: The name `format_token` is a bit confusing, I first thought you tried to use the macro but forgot the exclamation mark and the `s`. Or that it constructs a  single `FormatToken`. 

Or, maybe it's the name `FormatToken` that's confusing? Should we just call it `Token` since the crate add the necessary context that it's for formatting? Or  can we think of another name: `FormatIR`?

crates/formatter/src/lib.rs
```diff
@@ -1,16 +1,64 @@
...
+//!     assert_eq!(result.code(), "lorem => ipsum");
+//! }
+//!
+//! ```
+//! [IR]: https://en.wikipedia.org/wiki/Intermediate_representation
```
MichaReiser, line 43: I think you can directly embed the link inside of the comment since it should only render the label?

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,66 @@
...
+
+	( $( $token:expr ),+ $(,)?) => {{
+		use rome_formatter::{FormatToken, ListToken};
+		FormatToken::from(ListToken::concat(vec![
+			$(
+					 FormatToken::from($token)
+			),+
+
+		]))
+	}};
```
MichaReiser, line 65: Do we need the `expr` and `literal` cases or can we unify them somehow (since they're identical?) I only defined the `expr` case and it was able to handle `literals` and `expressions. 

They way I would understand the literal-declaration is that it only applies if all items are literals but we aren't doing anything different than for expressions. So just handling expressions couldbe fine?

MichaReiser (commented):
Nice! I actually like this approach better than what I prototyped (It saves us a few more characters and the implementation is roughly as simple).

crates/formatter/src/lib.rs
ematipico: I know, I just didn't want to have a long line :)

crates/formatter/src/format_token_macro.rs
ematipico: Yeah it's not needed. I will remove it

crates/formatter/src/lib.rs
ematipico: I don't mind `format_token`, although to macro `format_tokens!` can be confusing. What if we renamed `format_tokens` in `concat_tokens`?

crates/cli/tests/main.rs
```diff
@@ -29,6 +29,7 @@ fn test_format_cli() {
 - file "fixtures/input.json"
 - with options Tab
 {"string": "foo", "boolean": false, "number": 15, "object": {"something": 15}}
+
```
ematipico, line 32: This fixes the issue we have in `main`

crates/formatter/src/lib.rs
MichaReiser: I prefer `format_tokens` because it doesn't necessarily concat the tokens (which is why this approach is superior to the proposal I worked on), but sometimes only returns a single token. 

This could even be an argument why the `format_tokens` macro should be named `format_token`.... but not sure how to name the `format` function accepting a token in that case

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,73 @@
...
+/// without particular rule.
+///
+/// # Examples
+///
+/// Let's suppose you need to create tokens for the string `"foo": "bar"`,
+/// you would write:
+///
+/// ```rust
+/// use rome_formatter::{format_tokens, FormatToken};
+/// let token = format_tokens!("foo", ":", FormatToken::Space, "bar");
```
MichaReiser, line 12: Nit, can be simplified to `format_tokens!("foo:", FormatToken::Space, "bar")`

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,73 @@
...
+/// ```rust
+/// use rome_formatter::{format_tokens, IndentToken, LineToken, FormatToken};
+/// let token = format_tokens!(
+///   "foo",
+///   ":",
+///   IndentToken::new(
+///     format_tokens!(LineToken::hard(), "bar", ":", FormatToken::Space, "lorem")
+///   ),
+///   "}"
+/// );
```
MichaReiser, line 34: Nit
```suggestion
///   "foo:",
///   IndentToken::new(
///     format_tokens!(LineToken::hard(), "bar:", FormatToken::Space, "lorem")
///   ),
///   "}"
/// );
```

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,73 @@
...
+///     FormatToken::string("lorem")
+/// ]);
+/// let outer = FormatToken::concat(vec![
+///     FormatToken::string("foo"),
+///     FormatToken::string(":"),
+///     FormatToken::indent(inner),
+///     FormatToken::string("}")
+/// ]);
+///
+/// assert_eq!(token, outer);
```
MichaReiser, line 50: It might be clearer to reader what's happening here if you assert against the formatted string.
```suggestion
/// 
/// assert_eq!(r#"foo: {
  bar: lorem
}"#, format_token(&token).code());
```

It also guarantees that the created tokens match what you expect (missing opening brackets after `foo`, missing quotes).

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,73 @@
...
+///
+/// ```
+#[macro_export]
+macro_rules! format_tokens {
+
+	// called for things like format_tokens!("hey")
```
MichaReiser, line 56: Do doctests work for macros? If so, can we have one doctest here?

crates/formatter/src/format_token_macro.rs
```diff
@@ -0,0 +1,73 @@
...
+	($token:expr) => {
```
MichaReiser, line 57: This case isn't yet tested by the documentation above (All the examples only use the `vector` overload. Do macros support doctests? If so, maybe add a test in the macro definition?

crates/formatter/src/lib.rs
```diff
@@ -1,16 +1,64 @@
...
+
 mod format_json;
 mod format_token;
+mod format_token_macro;
```
MichaReiser, line 47: Nit: `format_token`**s**`_macro`

MichaReiser (commented):
Do you plan to rewrite the call-sites with your macro in a follow-up PR?

crates/formatter/src/format_token_macro.rs
ematipico:
```suggestion
/// let token = format_tokens!("foo:", FormatToken::Space, "bar");
```

crates/formatter/src/format_tokens_macro.rs
```diff
@@ -0,0 +1,61 @@
...
+///     format_tokens!(FormatToken::Space, "bar:", FormatToken::Space, "lorem")
+///   ),
+///    FormatToken::Space,
+///   "}"
+/// );
+/// let unique_token = format_tokens!("single");
+/// let options = FormatOptions::default();
+/// assert_eq!(r#"foo: { bar: lorem }"#, format_token(&token, options).code());
+/// let options = FormatOptions::default();
+/// assert_eq!(r#"single"#, format_token(&unique_token, options).code());
```
MichaReiser, line 39: Nit: Split into two examples since barely anything except the use statement is reused
```suggestion
/// use rome_formatter::{format_tokens, format_token, IndentToken, LineToken, FormatToken, FormatOptions};
/// let token = format_tokens!(
///   "foo:",
///   FormatToken::Space,
///   "{",
///   IndentToken::new(
///     format_tokens!(FormatToken::Space, "bar:", FormatToken::Space, "lorem")
///   ),
///    FormatToken::Space,
///   "}"
/// );
/// assert_eq!(r#"foo: { bar: lorem }"#, format_token(&token, options).code());
/// ```
/// Or creating a single token
/// ```
/// use rome_formatter::{format_tokens, format_token, FormatOptions};
/// let options = ;
/// assert_eq!(r#"single"#, format_token(&format_tokens!("single"), FormatOptions::default()).code());
```

MichaReiser (approved):

PR #1679, opened 2021-09-30
feat: format JS from CST

ematipico:
Add support for formatting arrays in JSON


.cargo/config.toml
```diff
@@ -1,2 +1,3 @@
 [alias]
 xtask = "run --package xtask --bin xtask --"
+lint = "clippy --workspace --all-targets --verbose -- --deny warnings"
```
jamiebuilds, line 3: nice

crates/formatter/src/ts/array.rs
```diff
@@ -0,0 +1,41 @@
+use crate::{
+	format_tokens, ts::format_syntax_token, FormatToken, FormatValue, GroupToken, ListToken,
+};
+use parser::SyntaxKind;
+use syntax::{
+	ast::{self, AstNode},
+	NodeOrToken, SyntaxNode,
+};
+
+pub fn format_node(node: SyntaxNode) -> FormatToken {
```
MichaReiser, line 10: Would it be possible to instead implement `FormatValue` for `ast::Array`. It would be nice if we could avoid the boilerplate of casting to a specific AST type and make use of rusts trait system.

I remember that you mentioned that there's often need for special handling if the `children` are empty. I think we could add a `is_empty()` method to `Children` that simply tests if the iterator returns any element (should be fairly cheap): `array.elements().iter().next().is_none()`.

crates/formatter/src/ts/array.rs
```diff
@@ -0,0 +1,41 @@
...
+					}
+
+					_ => None,
+				}
+			}
+		};
+		if let Some(token) = token {
+			tokens.push(token);
+		}
+	}
```
MichaReiser, line 33: It should be possible to rewrite this using `map` and `filter`.

```rust
let tokens = elements.map(|element| {
  match element {
	NodeOrToken::Node(node) => Some(node.format()),
	NodeOrToken::Token(token) => {
		// this is an hack until we figure out a better API to retrieve
		// token elements that are really part of an array
		match token.kind() {
			SyntaxKind::True | SyntaxKind::False | SyntaxKind::Number => {
				Some(format_syntax_token(token))
			}

			_ => None,
		}
	}
}).flatten();
```

`flatten` filters out all `None` tokens

crates/formatter/src/ts/array.rs
```diff
@@ -0,0 +1,41 @@
...
+};
+use parser::SyntaxKind;
+use syntax::{
+	ast::{self, AstNode},
+	NodeOrToken, SyntaxNode,
+};
+
+pub fn format_node(node: SyntaxNode) -> FormatToken {
+	let array = ast::Array::cast(node).unwrap();
+	let elements = array.elements();
```
MichaReiser, line 12: What happens if the array only contains an `Error` node? Does elements return the `Error` nodes too?

crates/formatter/src/ts/arrow_function.rs
```diff
@@ -0,0 +1,28 @@
...
+	let body = arrow_function.body();
+
+	if let Some(body) = body {
+		let s = body.syntax();
+		tokens.push(s.format());
+	} else {
+		let last_child = arrow_function.syntax().last_token().unwrap();
+		// the last child should be the value
+		tokens.push(format_tokens!(last_child.text()));
+	}
```
MichaReiser, line 25: What if the error function has no body because it's an error? Do we then strip the body from the formatted result?

crates/formatter/src/ts/array.rs
ematipico: Personally, I think this hack should exist at all, at least not inside the format part of the array. We should move it inside the `AstNode` API. I will keep the suggest in mind regardless! Thank you

crates/formatter/src/ts/array.rs
ematipico: The format chain keeps going. The `ERROR@` node will be formatted accordingly

crates/formatter/src/ts/arrow_function.rs
ematipico: The error node will still contain the errored tokens and the format chain will move on. I didn't test this particular case, we should first be able to replicate an error case and later see what we want to do.

crates/formatter/src/ts/array.rs
ematipico: I'm not sure this is the right way to approach the solution. If we want to apply `FormatValue` to `ast::Array`, we would ended up to this for all the AST nodes. Doing so would complicate things where we can't call `node.format` when we are iterating over the children of an expression. We would need to have a `match` inside each expression and call `cast` and `format` every time.

MichaReiser: We could provide a helper function that does the call for you and we may even be able to generate the code for it (matching on the syntax kind).

PR #1680, opened 2021-10-01
Format CST -> CST rather than to a String

MichaReiser:
## Summary

Draft for formatting a `CST` to a `CST` rather than to a `String` to not loos any of the information that we had to expensively gain by parsing the JavaScript file. The hope is that this will allow us write autofixes for lint rules without having to worry about the formatting because we can hand the CST over to the formatted that then formats the CST as desired by the user (whitespace, quotes, trailing commas, ...). 

Changing the `Printer` to print to a CST requires that we encode the `CST` in the IR of the formatted (currently called `FormatToken`). There are three new IR kinds (naming to be defined)

* `NodeToken`: Encodes that the following content belongs to this node. 
* `TokenToken`: Replaces the `StringToken`. Any content that should be present in a node
* `RawNodeToken`: The printer ignores this node. It does not reformat any of its content. Should be used for error nodes. 

The array formatter could look like this (there's probably logic that we can later extract into a shared trait to avoid repetition)

```rust
impl Tokenize for Array {
  pub fn to_format_ir(&self, context: &mut FormatContext) -> FormatIR {
    let separator = format_tokens![context.tokens.comma(), LineToken::soft_or_space()];
    let elements = format_tokens![
      LineToken::soft(),
      ListToken::join(
        separator,
        self.children().map(|child| child.to_format_ir(context))
      ),
    ];

    NodeToken::new(
      self.syntax(),
      GroupToken::new(format_tokens![
        context.tokens.left_bracket(),
        FormatToken::indent(elements),
        LineToken::soft(),
        context.tokens.right_bracket(),
      ]),
    ).into()
}
```

## Test Plan

All tests are still passing


crates/formatter/src/format_json.rs
```diff
@@ -1,182 +1,257 @@
...
 impl FormatValue for Value {
-	fn format(&self) -> FormatToken {
+	fn format(&self, context: &mut FormatContext) -> FormatToken {
 		match self {
-			Value::String(string) => {
-				FormatToken::string(format!("\"{}\"", escape_string(string)).as_str())
-			}
+			Value::String(string) => string_token(string, context),
+
 			Value::Number(number) => {
```
MichaReiser, line 19: This is super hacky and only a drop in replacement until we have a JSON -> CST parser. The code currently builds up a CST and creates the format tokens. It should use caching for nodes but it doesn't which makes it super inefficient as well.

MichaReiser:
Closing for above-mentioned reasons.

PR #1681, opened 2021-10-01
feat: add syntax crate and experimental codegen

yassere:
Edit: On second thought, maybe this shouldn't get merged with so many broken AST methods. It could get confusing even during experimentation. If we want to continue exploring this approach, it should at least detect and exclude the generated methods that won't work properly.

## Summary

This is very much an exploratory work in progress. It's messy, buggy, and incomplete. I only want to merge it so that other exploratory work can make use of it.

The AST codegen here looks more successful than it actually is. Many of the generated AST node methods won't work as expected because they rely on assumptions about the syntax trees that aren't consistent with the shape of the tree-sitter parse trees. It's possible to detect those problems in order to exclude those methods, and it's also possible to extend AST nodes with hand-written methods that would be hard to generate.

However, it seems likely that it would require forking and modifying the tree-sitter grammars in order for this approach to be truly successful.

Some breaking changes related to parsing:
- The `parser` crate is no longer dependent on rowan and so doesn't return a syntax tree directly. There's now a `parse` function exported from the `syntax` crate that can be used instead to produce a syntax tree from text.
- There's now a single `SyntaxKind` enum rather than one per language.
- The `SyntaxKind` variants are now in SCREAMING_SNAKE_CASE rather than PascalCase. I think it's harder on the eyes now, but it helps differentiate between syntax kinds and AST nodes.
- There's now a `ROOT` node for every syntax tree. If a file has some blank lines at the top, the whitespace token is attached to the root and the rest of the nodes now have the correct offset. Also, a blank file can now be parsed.


ematipico:
> On second thought, maybe this shouldn't get merged with so many broken AST methods. It could get confusing even during experimentation. If we want to continue exploring this approach, it should at least detect and exclude the generated methods that won't work properly.

What other alternatives do we have?

crates/parser/src/syntax_kind/generated.rs
```diff
@@ -0,0 +1,923 @@
+#![doc = "This file has been autogenerated. Do not modify."]
+#![doc = "To update, run: `cargo run --bin sourcegen"]
+#![allow(bad_style)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(u16)]
+pub enum SyntaxKind {
+	ABSTRACT_CLASS_DECLARATION,
```
MichaReiser, line 7: In what case are you worried that the names collide with the `Ast` node types? Are you using a `star` import to get all `SyntaxKind`s into the scope?

I personally haven't encountered any confusion until know (I'm more struggling with `syntax::SyntaxKind` vs `rowan::SyntaxKind`) and prefer the Rust naming convention.

crates/parser/src/syntax_kind/generated.rs
```diff
@@ -0,0 +1,923 @@
...
+		145u16 => LBRACE_PIPE_TOKEN,
+		80u16 => PIPE_TOKEN,
+		63u16 => PIPE_EQ_TOKEN,
+		74u16 => PIPE_PIPE_TOKEN,
+		69u16 => PIPE_PIPE_EQ_TOKEN,
+		146u16 => PIPE_RBRACE_TOKEN,
+		11u16 => RBRACE_TOKEN,
+		93u16 => TILDE_TOKEN,
+		65535u16 => SyntaxKind::ERROR,
+		_ => panic!("Syntax Kind not found: {}", kind_id),
```
MichaReiser, line 921: Nit: Can we add the language to the panic message? Just in case Rust doesn't retain the line information if tracing isn't enabled.

crates/syntax/src/ast.rs
```diff
@@ -0,0 +1,94 @@
...
+	fn syntax(&self) -> &SyntaxToken;
+
+	fn text(&self) -> &str {
+		self.syntax().text()
+	}
+}
+
+/// An iterator over `SyntaxNode` children of a particular AST type.
+#[derive(Debug, Clone)]
+pub struct AstChildren<N> {
```
MichaReiser, line 55: Can we document the generic or what does `N` stand for? Is it the type of the children? How about `TChild` or just `Child`?

Any reason `N` isn't constraint to `AstNode`? Creating a `AstChildren` seems to be of little use for any other type because the `Iterator` implementation is constraint to `AstNode`.

crates/syntax/src/node_ext.rs
```diff
@@ -0,0 +1,18 @@
...
+
+// This is an example of extending as AST node, but these are actually
+// methods that rust-analyzer is able to auto-generate
+impl ast::Array {
+	pub fn l_brack_token(&self) -> Option<SyntaxToken> {
+		support::token(&self.syntax, SyntaxKind::LBRACK_TOKEN)
+	}
+	pub fn r_brack_token(&self) -> Option<SyntaxToken> {
+		support::token(&self.syntax, SyntaxKind::RBRACK_TOKEN)
+	}
```
MichaReiser, line 17: Nit, I would go with `lbrack_token` or if you prefer, rename `SyntaxKind::LBRACK_TOKEN` to `SyntaxKind::L_BRACK_TOKEN` to be consistent. Same for `rbrack`.

crates/syntax/src/syntax_node.rs
```diff
@@ -0,0 +1,18 @@
+use crate::SyntaxKind;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum TSX {}
```
MichaReiser, line 4: Should this still be `TSX` or be more generic because we re-use `SyntaxKind`'s across languages.

MichaReiser (approved):
Nice! I would love to see this merged as soon as possible to unblock our depending work. 

I'm fine with broken AST methods if you could provide elaborate on the problematic cases and maybe give an example so that we, the users, are aware for what to look out for. 

It would be nice to see some more tests/examples/doc in future PRs to make it easier for people to understand how the APIs are intended to be used. 

I haven't reviewed all the code gen code.

crates/parser/src/lib.rs
```diff
@@ -6,125 +6,107 @@
...
+pub use crate::syntax_kind::generated::{ts_syntax_kind, tsx_syntax_kind, SyntaxKind};
 
-	fn name(&self) -> &'static str;
-	fn get_kind(&self, kind: &str, named: bool) -> u16;
-	fn whitespace_kind(&self) -> u16;
-	fn new_root(&self, green: GreenNode) -> Self::SyntaxNode;
-}
+use tree_sitter_typescript::{language_tsx, language_typescript};
+
+// TODO: Possibly add `error` method. Tree-sitter creates ERROR nodes on the parse
```
ematipico, line 23: We should write down some ideas here

crates/parser/src/lib.rs
```diff
@@ -6,125 +6,107 @@
...
-		let tree = self
-			.parser
-			.parse(src, None)
-			.ok_or_else(|| anyhow!("Failed Parsing"))?;
-
-		let green = self.build_green_tree(src, tree);
-		let tree = self.language.new_root(green);
-		Ok(tree)
-	}
+pub enum Language {
```
ematipico, line 37: Shouldn't this enum be in the `syntax` crate?

crates/syntax/src/ast.rs
```diff
@@ -0,0 +1,94 @@
...
+	where
+		Self: Sized;
+
+	fn cast(syntax: SyntaxToken) -> Option<Self>
+	where
+		Self: Sized;
+
+	fn syntax(&self) -> &SyntaxToken;
+
+	fn text(&self) -> &str {
```
ematipico, line 48: I would call this API `value`

crates/syntax/src/ast.rs
ematipico: > Can we document the generic or what does N stand for? Is it the type of the children?

Personally, I would like to avoid single letters for generics. We should give them some decent name so a developer *knows* what a generic is for.

crates/syntax/src/lib.rs
```diff
@@ -1 +1,51 @@
...
+	fn start_node(&mut self, kind: SyntaxKind) {
+		self.inner.start_node(rowan::SyntaxKind(kind.into()))
+	}
+
+	fn finish_node(&mut self) {
+		self.inner.finish_node()
+	}
+}
+
+pub fn parse(src: &str, language: Language) -> Result<SyntaxNode> {
```
ematipico, line 45: I think some doc test here would be a nice to have.

crates/syntax/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+	let args: Vec<String> = std::env::args().collect();
+
+	let lang = args.get(1).ok_or_else(|| anyhow!("Missing language"))?;
+	let path = args.get(2).ok_or_else(|| anyhow!("Missing filename"))?;
+
+	let language = match lang.as_str() {
+		"ts" => syntax::Language::Ts,
+		"tsx" => syntax::Language::Tsx,
+		lang => return Err(anyhow!("Invalid language: {}", lang)),
+	};
```
ematipico, line 13: In this case, I think we can implement the trait `FromStr` for the `enum Language`

crates/syntax/src/syntax_node.rs
ematipico: We could call it `RomeLanguage`

crates/syntax_codegen/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "syntax_codegen"
+version = "0.1.0"
```
ematipico, line 3:
```suggestion
version = "0.0.0"
```

We wouldn't want to publish it

crates/syntax_codegen/src/lib.rs
```diff
@@ -0,0 +1,354 @@
+//#![allow(unused)]
```
ematipico, line 1: to remove?

crates/parser/src/lib.rs
ematipico: I would call it `RomeLanguage`

ematipico (approved):
I feel we should merge this ASAP and create an issue with a list of all the things that we need to do in order to make the code more stable: 
- documentation
- comments
- examples
- `README.md`

yassere:
@ematipico I think that RSLint's parser could be a viable alternative. 

Ok, I'll merge this for now. But I think we should explore RSLint's parser next. If we decide to continue pursuing this approach, I'll come back to address feedback. Or if we decide to abandon this approach, I'll come back and remove this code.

PR #1683, opened 2021-10-04
chore: rename format to tokenize

ematipico:
## Summary

The team that is working on the formatter realised that using the name `.format` is actually correct. The function is not actually formatting anything, is just creating a IR for what we want to format. 

This PR does:
- rename the function `.format` to `.tokenize`
- rename the fomatter trait `FormatValue` to `TokenizeValue`

This is also in line with [what we had in Rome TS](https://github.com/rome/tools/blob/archived-js/internal/formatter/Builder.ts#L168).

I took the liberty to add a couple of cargo aliases that will help the contribution in the future.

## Test Plan

```
cargo format
cargo lint
cargo test
``

<!-- Demonstrate the code is solid. Example: The exact commands you ran and their output. -->

### CODE REVIEW

File: crates/formatter/src/lib.rs [hunk 1]
```diff
@@ -59,8 +59,8 @@ pub use printer::Printer;
 pub use printer::PrinterOptions;
 
 /// This trait should be implemented on each node/value that should have a formatted representation
-pub trait FormatValue {
-	fn format(&self) -> FormatToken;
+pub trait TokenizeValue {
```
MichaReiser, line 62: NIt: How about `Tokenizable`? `TokenizeValue` sounds more like an action, tokenize the value whereas the trait is something that can be tokenized.

MichaReiser (approved):
Looks good but would prefer if we can wait with the other renames to avoid too many merge conflicts with our not yet merged branches. 

I'm also not entirely sure if `tokenize` will be the final naming depending on whatever name we pick for `FormatToken` because I still find it weird that terminals and non-terminal nodes can be `FormatTokens` (e.g. group is a non terminal).

crates/formatter/src/lib.rs
ematipico: I like it! I will apply the change

crates/formatter/src/format_json.rs
```diff
@@ -1,9 +1,10 @@
 use crate::format_token::{GroupToken, LineToken};
-use crate::{format_token::FormatToken, FormatValue, ListToken};
+use crate::ListToken;
+use crate::{format_token::FormatToken, Tokenizable};
 use serde_json::Value;
 
-impl FormatValue for Value {
-	fn format(&self) -> FormatToken {
+impl Tokenizable for Value {
```
jamiebuilds, line 6: Based on what I've seen I think the Rust community prefers naming traits after the action rather than an adjective, i.e. `Clone` over `Cloneable`

I think this should just be `Tokenize`

PR #1685, opened 2021-10-06
temp clone rslint parser/etc into repo

jamiebuilds:
## Summary

We are doing some investigations into rslint's parser right now, we're going to temporarily check in the project to make some changes we need right away. We're _not_ planning on forking the parser.

## Test Plan

Let's see how CI does with the coverage bits


jamiebuilds:
Note: I kept the `rslint_` prefix so that there would be no confusion

MichaReiser (approved):
🎉

Haven't done a proper review in the assumption that you just pull in the files.

PR #1686, opened 2021-10-06
Parse JSON to CST

MichaReiser:
A hacky implementation that uses the rs-lint parser to parse a JSON string to a CST.

## Summary
Our current JSON parser uses serde for parsing, which works well enough but isn't lossless. For example, the parsing doesn't retain the exact number formats. We also want to get more experience working with CST's and do profiling which requires us to work with a CST. 

That's why this PR swaps the serde parser with the rslint JavaScript parser. Using a JS parser isn't ideal because it parses files that are valid JS but not JSON. I believe that's OK for now to have something to play around with CST and figure out multi-language support until we build our own JSON parser. 

## Test Plan

Tests are still passing. Parsed and formatted a large (+10 MB) JSON file

### DIFF
--- a/crates/formatter/src/format_json.rs
+++ b/crates/formatter/src/format_json.rs
@@ -1,184 +1,185 @@
 use crate::format_token::{GroupToken, LineToken};
-use crate::{format_token::FormatToken, FormatValue, ListToken};
-use serde_json::Value;
+use crate::{format_token::FormatToken, format_tokens};
+use rslint_parser::ast::{
+	ArrayExpr, GroupingExpr, Literal, LiteralProp, ObjectExpr, ObjectProp, UnaryExpr,
+};
+use rslint_parser::{parse_text, AstNode, SyntaxKind, SyntaxNode, SyntaxToken};
 
-impl FormatValue for Value {
-	fn format(&self) -> FormatToken {
-		match self {
-			Value::String(string) => {
-				FormatToken::string(format!("\"{}\"", escape_string(string)).as_str())
-			}
-			Value::Number(number) => {
-				let number = number.as_f64().unwrap();
-				FormatToken::f64(number)
-			}
-			Value::Bool(value) => FormatToken::from(value),
-			Value::Object(value) => {
-				let separator = FormatToken::concat(vec![
-					FormatToken::string(","),
-					FormatToken::Line(LineToken::soft_or_space()),
-				]);
-
-				let properties_list: Vec<FormatToken> = value
-					.iter()
-					.map(|(key, value)| {
-						FormatToken::concat(vec![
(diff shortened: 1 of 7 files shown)


crates/formatter/src/printer.rs
```diff
@@ -236,7 +236,7 @@ impl Printer {
 				let calls = self.print_token(token, args);
 
 				// If the line is too long, break the group
-				if self.state.line_width > self.options.print_width {
+				if self.state.line_width > self.options.print_width as usize {
```
MichaReiser, line 239: The indices overflowed when parsing a large file.

crates/formatter/src/format_json.rs
```diff
@@ -1,211 +1,212 @@
...
+		}
 
-	FormatToken::from(ListToken::concat(vec![
-		json.format(),
-		FormatToken::from(LineToken::hard()),
-	]))
+		SyntaxKind::OBJECT_EXPR => {
+			let object = ObjectExpr::cast(node).unwrap();
+
+			let separator = format_tokens![",", LineToken::soft_or_space(),];
```
ematipico, line 47: I think we should stick to one type of usage. I know we can use any kind of parenthesis, but for our own sanity (and the community), we should promote only one. In the rust doc I used the `format_tokens!()`. I don't mind any of those, let's just decided which one we want.

crates/formatter/src/format_json.rs
```diff
@@ -1,211 +1,212 @@
...
-				]))
-			}
-		}
+use crate::{format_token::FormatToken, format_tokens};
+use rslint_parser::ast::{
+	ArrayExpr, GroupingExpr, Literal, LiteralProp, ObjectExpr, ObjectProp, UnaryExpr,
+};
+use rslint_parser::{parse_text, AstNode, SyntaxKind, SyntaxNode, SyntaxToken};
+
+fn tokenize_token(token: SyntaxToken) -> FormatToken {
```
ematipico, line 8: Not sure if it's done on purpose or not, but I think this version of the parser is not able to identify non-valid JSON files.

For example `{ "foo": "bar", }` should fails because of the `,`.

ematipico (commented):
I'm not sure about this change. `serde_json` is able, at least, to identify non-valid JSON files while this parser is not able to do that.

crates/formatter/src/format_json.rs
MichaReiser: I'm aware that the parser won't recognise invalid JSON (as stated in the summary) and share the concerns. But I believe that having a CST based implementation helps us to experiment, figure out multi-language support etc whereas serde just doesn't fit into our outlined architecture.

PR #1687, opened 2021-10-06
feat(formatter): multiple language support

ematipico:
## Summary

This PR attempts to create a small footprint of how we can format files with different languages. The PR is not over yet:
- [x] create examples
- [x] document new crates
- [x] add more tests

I created two new crates that we will need in the future:
- `path`: a small wrapper around `std::path::PathBuf`. It will contain more information, needed to Rome in the future
- `core`: which will contain a context/app that belong to the core of the application. For now I ported the concept of [`file_handlers`](https://github.com/rome/tools/tree/archived-js/internal/core/common/file-handlers) in Rust

## Test Plan

```
cargo test
cargo lint
cargo format
```

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -9,9 +9,8 @@ pub type Handlers = HashMap<&'static str, Box<dyn ExtensionHandler>>;
 
 pub struct App {
 	handlers: Handlers,
+	unknown_handler: Box<dyn ExtensionHandler>,
 }
-
-const UNKNOWN_EXTENSION: &str = "unknown";
 
 impl Default for App {
 	fn default() -> Self {
@@ -19,8 +18,10 @@ impl Default for App {
 		map.insert("js", Box::new(JsFileHandler {}));
 		map.insert("ts", Box::new(JsFileHandler {}));
 		map.insert("json", Box::new(JsonFileHandler {}));
-		map.insert(UNKNOWN_EXTENSION, Box::new(UnknownFileHandler {}));
-		Self { handlers: map }
+		Self {
+			handlers: map,
+			unknown_handler: Box::new(UnknownFileHandler {}),
+		}
 	}
 }
 
@@ -33,7 +34,7 @@ impl App {
 		let handler = if self.handlers.contains_key(file_extension) {
 			self.handlers.get(file_extension)
 		} else {
-			self.handlers.get(UNKNOWN_EXTENSION)
+			Some(&self.unknown_handler)
 		};
 		handler.map(|handler| handler.as_ref())
 	}
--- a/crates/path/src/lib.rs
+++ b/crates/path/src/lib.rs
@@ -84,6 +84,7 @@ impl<'handler> RomePath<'handler> {
 	/// Returns the current handler associated to the file.
 	///
 	/// You need to call [deduce_handler] first in order to receive one. If not, [None] is always returned.
+	// TODO: move handler deduction inside the path crate
 	pub fn get_handler(&self) -> Option<&dyn ExtensionHandler> {
 		self.handler
 	}


crates/core/src/file_handlers/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+	fn can_have_scale(&self) -> bool {
+		false
+	}
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum FileHandlers {
+	Js(JsFileHandler),
+	Json(JsonFileHandler),
+	Base(BaseFileHandler),
```
MichaReiser, line 66: Can we call this `Default` or `UnknownFileHandler`. Base gives me the impression it's the base implementation of all handlers.

crates/core/src/file_handlers/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+	fn is_asset(&self) -> bool {
+		false
+	}
+
+	// TODO: what is this?
+	fn can_have_scale(&self) -> bool {
+		false
+	}
+}
```
MichaReiser, line 60: Can we remove it for now?

crates/core/src/file_handlers/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// Main trait to use to add a new language to Rome
+pub trait ExtensionHandler {
+	fn language(&self) -> Language;
+
+	fn mime(&self) -> Mime;
+
+	fn use_tabs(&self) -> bool {
+		true
+	}
```
MichaReiser, line 43: Does that mean it's allowed to use tabs or if you must use tabs?

crates/file/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+	type Target = PathBuf;
+
+	fn deref(&self) -> &Self::Target {
+		&self.file
+	}
+}
+
+impl<'handler> Debug for RomePath<'handler> {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		write!(f, "{:?}", self)
```
MichaReiser: Isn't this the default debug implementation?

crates/core/src/file_handlers/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+	// TODO: what is this?
+	fn can_have_scale(&self) -> bool {
+		false
+	}
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum FileHandlers {
```
MichaReiser, line 63: I think this should be a trait rather than an enum because, in my view, it's not that important in `format` if the handler is a `JS` or `JSON` handler it's only important what e.g. `capabilities` returns. 

All our other enums use the `singular` form, e.g. `FileHandler` rather than `FileHandlers`.

crates/core/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+	base_file_handler: FileHandlers,
+	json_file_handler: FileHandlers,
+}
+
+impl App {
+	pub fn new() -> Self {
+		Self {
+			js_file_handler: FileHandlers::Js(JsFileHandler {}),
+			base_file_handler: FileHandlers::Base(BaseFileHandler {}),
+			json_file_handler: FileHandlers::Json(JsonFileHandler {}),
```
MichaReiser, line 17: I personally would use a `HashMap<extension, FileHandler>` to cache the builders if needed.

crates/file/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+impl<'handler> RomePath<'handler> {
+	pub fn new(path_to_file: &str) -> Self {
```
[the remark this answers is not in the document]
ematipico: I wanted to leave the responsibility to resolve and handle paths to this crate. For example if a `Path` doesn't actually exist, this should be handled by this crate.

crates/core/src/file_handlers/mod.rs
ematipico: It means that it is allowed to use tabs

crates/core/src/file_handlers/mod.rs
ematipico: By using a trait objects would complicate things and it wasn't needed for my design. Using trait objects would cause using `Box<dyn FileHandlers>` because the compiler can't determine its size (my first implementation relied on trait objects). Basically I was trying to make things dynamic while the same thing can be resolved using a more static approach.

Probably this is achievable but it would complicate things for no reason.

crates/core/src/lib.rs
ematipico: I'd prefer to rely on a more static approach, especially for few structs that will leave for mostly the whole time the app runs - the `'static` lifetime would be a good fit.

crates/core/src/file_handlers/mod.rs
MichaReiser: Cool, how about we rename it to `may_use_tabs`, `can_use_tabs`?

crates/core/src/lib.rs
MichaReiser: Ah yeah sorry, I missed the `dyn` keyword. 

Why do you need to implement `Eq` and `Hash` for file handler if the map key is a string (or we could use an enum if we want)?

crates/core/src/file_handlers/mod.rs
xunilrj: You can still use the hashmap, for example.

```rust
pub struct JsFileHandler;
pub struct JsonFileHandler;
pub struct BaseFileHandler;

pub enum FileHandlers
{
    Js(JsFileHandler),
    Json(JsonFileHandler),
    Base(BaseFileHandler),
}

pub struct App {
    handlers: Vec<FileHandlers>,
    extensions: std::collections::HashMap<&'static str, usize>,
}

impl App {
    pub fn new() -> Self {       
        Self {
             handlers: vec![
                     FileHandlers::Base(BaseFileHandler {}),
                     FileHandlers::Js(JsFileHandler {}),
                     FileHandlers::Json(JsonFileHandler {})
             ],
             // https://docs.rs/sugars/3.0.0/sugars/macro.hmap.html
             extensions: sugars::hmap! {
                "js" => 1,
                "jsx" => 1,
                "json" => 2,
            }
        }
    }

    pub fn get_handler<S: AsRef<str>>(&self, file_extension: S) -> Option<&FileHandlers> {
        let file_extension = file_extension.as_ref();
        let i = self.extensions.get(file_extension).unwrap_or(&0);
        self.handlers.get(*i)
    }
}

```

crates/core/src/lib.rs
ematipico: Everything boils down to the fact that trait objects are not `Sized`. This means, if there's need to do some operations (like equality check), these traits have be manually implemented.

Issue #1688, opened 2021-10-06
📎 Refine Formatter API


MichaReiser:
### Description

The focus in recent weeks prototyping a formatter in Rust that works on a rowan CST. We deliberately didn't spend much time on naming nor making sure the API is consistent and super convenient. However, we believe that we have now a good enough understanding to stabilize and refine the API.

We'll use this task to outline the new Formatter API. This also involves parts that are still in very active development and aren't yet merged onto main. 

### `PrintResult`
*  Rename to `FormatResult` because it's the main result returned by `format_*` methods
* Declare outside of `printer` module
* Add a `syntax()` method that re-parses the string to a CST   

### `FormatToken`
The postfix `Token` is misleading because `Token`s commonly only appear in the leaves of a tree which isn't the case for `GroupToken`, `ListToken` that wrap other `Token`s.  That's why we propose to rename `FormatToken` to `FormatElement` which also aligns the name with `rowan` that defines a `SyntaxElement` as a `SyntaxNode` or a `SyntaxToken`. 

* Rename `FormatToken` -> `FormatElement`
* Drop the `Token` postfix from the `Element` types
* Rename `String` -> `Token` because it doesn't represent a string literal but a token that gets printed to the output as is.
* Figure out where to put the `FormatToken::*` helpers. Options are
  * Put them on `FormatContext` -> `context.soft_line()`
  * Create factory functions that can be imported using `use format_token:*` -> `soft_line()`
  * Let users use the `Line`, `Group` types directly -> `Line::soft()`

The following changes are more hygienic and related to renaming `FormatToken` to `FormatElement`
- `format_token` → `format_element` : Pretty prints a source that's already in the formatters IR
- `format_str` → `format_str` : Pretty prints a source given as string
- `format_tokens!` → `format_elements!` : Macro for conveniently build format elements.

### `ToFormatElement`
Rename the `Tokenizable` trait to `ToFormatElement` to align with the other renamings. 

Create a generic implementation on `SyntaxNode` and `SyntaxToken` that matches on `SyntaxKind` and dispatches the formatting to the `ToFormatElement` implementation of the corresponding `AstNode` or `AstToken` that allow more convenient implementations than working with `SyntaxNode`/`Token` directly.

```rust
impl ToFormatElement for SyntaxNode {
  fn to_format_element(&self, context: &FormatContext) -> Document {
    match self.kind() {
      SyntaxKind::Script => Script::cast(self.clone()).unwrap().to_format_element(context),
[...]

PR #1696, opened 2021-10-07
feat(formatter): Add `FormatContext`

MichaReiser:
## Summary

Adds the `FormatContext` that is passed to `to_format_element` to give implementations a way to access the `FormatOption`s (to, for example, determine the preferred quotation marks) and format their children. Using the `FormatContext` to format the child node gives us a central place where we can add logic that is common for all nodes, for example: 

* Handling leading and trailing comments
* Handling source maps 
* Potentially more

This PR doesn't do error handling yet which is one of the next things we have on our plate. 

Part of #1688

## Test Plan
Ran the formatter tests


ematipico:
I am still not 100% convinced by this choice. Now that I implemented part of the formatter using `rslint_parser`, I don't see enough value.

With the current implementation, we call `::cast` just once (at the root, which is always a `Script`) and then it's downhill.

PR #1703, opened 2021-10-11
avoid one alocation when calling concat_elements

xunilrj:
## Summary

Avoid allocating a std::Vec that will be consumed right after.
I think the only danger is that now we will be using stack allocation.

## Test Plan

```rust
cargo test
```

### DIFF
--- a/crates/formatter/src/format_elements.rs
+++ b/crates/formatter/src/format_elements.rs
@@ -54,11 +54,10 @@ macro_rules! format_elements {
 
 	( $( $element:expr ),+ $(,)?) => {{
 		use $crate::{FormatElement, concat_elements};
-		concat_elements(vec![
+		concat_elements([
 			$(
 					 FormatElement::from($element)
 			),+
-
 		])
 	}};
 }


MichaReiser:
Took me a sec to understand what's going on. So what the implementation does now is to allocate an array on the stack instead of a vector on the heap. I don't know what's the difference performance-wise but I'm not worried about the stack allocation: 
* The number of passed elements is always constant
* The number of passed tends to be small 
* Wrapped content is anyways allocated on the heap (recursive data type).

xunilrj:
> The number of passed elements is always constant

Ideal for being on the stack. We can have problems if we use hundreds and hundreds of items. I don't think this will happen.

> The number of passed tends to be small

Ideal for being on the stack.

> Wrapped content is anyways allocated on the heap (recursive data type).

Not a problem.

If you see the "vec!" macro kind of does this already: https://doc.rust-lang.org/src/alloc/macros.rs.html#41-51. The difference being that "box" uses placement new: https://github.com/rust-lang/rust/issues/49733#issuecomment-388858850.

So we are avoiding one allocation per concat_elements call.

PR #1710, opened 2021-10-13
Revert "refactor: replace pico-args with clap on xtask"

jamiebuilds:
Reverts rome/tools#1705, opening this for discussion

Re: https://github.com/rome/tools/pull/1705#issuecomment-941963500

> I think the point of this crate was that it compiled much faster, which is something we want for xtask commands. It matters much less that it’s a complete CLI with tons of features

Also @yassere noted that the `cargo xtask coverage` command made `coverage_query` required which it was not previously

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,9 +108,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.71"
+version = "1.0.70"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


anonrig:
Hi @jamiebuilds, original author here. Can you share the compilation time difference between the two cli dependencies? Additionally, the required coverage_query is an error on my side. Happy to fix it on a different pull request.

anonrig:
I respect with your decision of not needing a featureful CLU. I was confused since your initial assessment was `I think the point of this crate was that it compiled much faster`, which referred to time, and now you say you did not measure time.

Issue #1718, opened 2021-10-27
☂️ Rome Parser Architecture Changes


jamiebuilds:
### Description

Last week, we spent our time discussing a number of different architectural changes that we would like to make to the parser and CST/AST. We also talked to some of the maintainers about RSLint about how we would like to work together in the future.

Immediately, we've already started working on these changes in our locally checked in fork of the RSLint parser, but in the coming weeks/months we're going to collaborate to ensure that the needs of both projects are being met, and in the future Rome will be exporting its own crate to be used in RSLint with collaborators from both projects.

cc @RDambrosio016 @Stupremee

> Note: I'll create a separate issue at some point to talk about adding RSLint contributors to the rome/tools repo and what we want the complete public API of the parser crate to be. I think there might be some open questions if we should be providing the AST to RSLint or if using Ungrammar we can allow both projects to generate their own AST optimized for their use cases and only really export a tree-sink. Either way, a separate issue would be better probably.

# Store leading/trailing trivia for each token

We explored two options on how to store trivia like spaces, tabs, newlines, single and multiline comments. Let's use the following example to illustrate how RSLint works today and how we want to store trivia moving onwards. 

```rust
[
	1,
	2, // test
	3
]
```

RSLint stores the trivia as a token of the kind `Whitespace` or `Comment` and attaches it to its enclosing node. The CST for the array example would look like this (whitespace and comments are highlighted):

```yaml
script
  - array
    - l-bracket
    - whitespace: \n\t       # <<<
    - element
      - literal
        - number_token: 1 
      - comma-token
      - comment: // test     # <<<
    - whitespace: \n\t       # <<<
    - comment: // test       # <<<
    - element
      - literal
        - number_token: 2
      - comma-token
    - whitespace: " "        # <<<
    - comment: // test       # <<<
    - whitespace: \n\t       # <<<
    - element
      - literal
        - number_token: 3
    - whitespace: \n         # <<<
    - r-bracket
  - whitespace: \n           # <<<
```

There are a number of concerns that we have about this design that we'd like to resolve:

[...]

PR #1719, opened 2021-10-27
RFC: JS Grammar

MichaReiser:
## Summary

RFC for a re-structured AST Facade (see `/grammar.ungram` file). The main motivations for restructuring the grammar are (#1718 goes into more details  of why):

* Structure the AST so that every token is exposed on the AST Facade. This may require introducing new intermediate nodes
* Be explicit about the optionality of an element so that the AST Facade can return an `Err` if a mandatory element is missing. 
* Group children that depend on each other's existence in a sub-node. For example, the `else` token and the `alternative` branch of an `if` statement must both either be present or absent. Therefore, add an **optional** `ElseClause` to the `if` statement that has the **mandatory** `else` token and `alternative` branch children` (if one is present, the other must as well). 
* Ensure that children appear in fixed ordering. For example, the `UpdatExpression` (`x++` or `++x`) must be modelled as two distinct nodes to guarantee that the operator and identifier are always in the same order for each node type (`PreUpdateExpression`, `PostUpdateExpression`)
* Encode possible position for error nodes (I'll look at specifying error recovery separately).  
* Align the naming and AST structure with Rome classic (and adapt it to align with the other goals). I tried to call out the main differences to Rome in comments above the relevant nodes. 

The grammar does not yet cover typescript. This is my first time trying to design an AST. I'm certain there are some missing children, cases I didn't think of and some confusing names. 

## Scope

* `crates/rome-js-grammar/grammar.ungram` defines the JS grammar
* `crates/rome-js-grammar/rfc.md` lists open questions, proposed decisions, code gen extensions etc. 
* The other files are generated or based on @ematipico's work on migrating to ungrammar. 

## Test Plan

My imagination... There's no parser generating the defined AST today. I sketched some trees by hand but that's all. Changing the parser will bring up any cases that I missed.

### DIFF
--- a/rfcs/001-js-grammar.md
+++ b/rfcs/001-js-grammar.md
@@ -1,0 +1,107 @@
+
+## Codegen extensions
+
+* **Must**: Add support for labels so that we can influence how the child accessors are named. For example, we want to name the left and right expressions of `BinaryExpression` `left` and `right` and not `expression` and `expression. Ungrammar supports the `label: Type` syntax. All we need to do is to respect the naming when generating the fields.
+* **Must**: Support inline token-union types. For example, the binary expression operator can be any of `==, +, -, ...`. That's why it's defined as `operator: ( '==' | '+' | '-' ...)`. The source gen should only generate a single field for the operator that returns a `SyntaxToken`
+* **Must**: Generate different return types for mandatory/optional children.
+* **Must/Wish**: It's convenient to define unions in respect of other unions. For example, `JsClassMemberName = JsObjectMemberName | JsPrivateClassMemberName` where `JsObjectMemberName` is an union type as well. This is convenient to maintain the grammar but makes the Facade more awkward to use because it requires two matches: first on the outer union and then on the inner union. We can avoid this if we flatten unions inside of the source gen and automatically generate `From<InnerUnion>` implementations.
+* **Wish**: Automatically strip the `Js` extensions from field names to reduce the need for explicit labels. For example, `ExpressionStatement` must use an explicit label for its inner expression so that it isn't named `js_expression`.
+* **Wish**: Generation of a union type over all nodes/tokens of a language, `JsElement = ..., JsNode = ..., JsToken = ...`
+
+## Proposed Changes
+
+### JS Prefixing
+* Prefix nodes  with `Js` to avoid conflicts with other languages
+* Prefix tokens with `Js` if they are not representing a specific character sequence. For example, prefix `js_string` because different languages use different quotes for strings but don't prefix `;` or `true` because these are "pure" character sequences. `true` might be somewhat controversial because different languages use different casing `TRUE` vs `true` vs `True`. I don't think the casing is relevant. All these represent a `true` token and the corresponding parser shouldn't generate `TRUE` for a language expecting `true` and each language-impl can provide helpers to build common tokens.
+
+
+### Separated sequences
+
+It's common in languages to have a sequence of elements that all are separated by a specific separator. Examples in JavaScript where elements are comma separated are:
+
+* Array elements
+* Object properties
+* Import/Export specifiers
+* ...
+
+There are two ways how we can model this. I'll use a simplified version of JS arrays that ignores array holes.
+
+**Wrap elements in a node**
+
+One option is to create a new node representing an `ArrayElement` that contains the value and an optional trailing comma.
+
+```
+JsArray = '[' elements: JsArrayElement* ']
+
+JsArrayElement =
+  value: JsArrayElementValue
+  trailing_comma: ','?
+
+JsArrayElementValue = JsExpression | JsSpread
+```
+
+Creating a new node has a few downsides:
+
+* Undesired nesting
+* Naming becomes tricky if the element values are over a union type. How to avoid the name clash of the union type and the `Element` type?
+* Repetitive grammar: JS defines multiple element-lists that are separated by commas.
+* Mutating the elements requires extra care to ensure that the appropriate number of commas are in (the right) place
+
+I saw that Roslyn' uses a [`SeparatedList`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.separatedsyntaxlist-1). It's just a nice view over a list of children and I think we can adopt a similar approach in the AST Facade by returning `AstSeparatedChildren` for such element lists.
+`AstSeparatedChildren` is a view over a range of children that alternate between element and separator. It may provide the following methods:
+
+* `nodes() -> Iterable<Result<T>>`: Returns the nodes inside of the list. The return type is a result in case a node is missing between two separators
+* `separators() -> Iterable<SyntaxToken>`: returns all the separators. There should be no need for an `Option` or `Result` because the parser never starts a new element if a separator is missing
+* `elements() -> Iterable<(Result<T>, Option<SyntaxToken>)>`: Returns a list of elements with their separator, for example for use inside of the formatter. The separator is optional because the last element may or may not have a trailing separator.
+
+We could add additional methods for mutating a separated list, querying a specific `node`/`separator`/`element` etc.
+
+I believe this gives us the same advantages as having distinct elements without the repetition:
+
+```
+JsArray = '[' elements: (JsArrayElement  (',' JsArrayElement )* ','?) ']
+
+JsArrayElement = JsExpression | JsSpread
+```
+
+We should even be able to automatically generate the proper return type on the field by testing whenever we see the `(T (',' T)* ','?)` pattern inside of the grammar, something that [rust-analyzer does too](https://github.com/rust-analyzer/rust-analyzer/blob/dfa355b4313a7ea3eb5a262ee9f1da71f50884d5/crates/syntax/src/tests/sourcegen_ast.rs#L676) already does.
+
+## Open Questions
+
+### Member vs Property naming
+
+Should we call object and class members `Members` or `Properties`?
+
+* `MemberExpression` has an `object` and `property` field today which feels inconsistent. Shouldn't property be called member or should it be a `PropertyExpression`
+* A property can be a value property, method, constructor, or a getter/setter. How to we prevent that we don't end up with property property?
+* The term member is commonly used in other languages
+
+The main downside I see is that property is a commonly used term in the JS ecosystem. Changing property to member could be an entry barrier for people familiar with the term property.
+
+### AST conformance
+
+How strict should our AST API be?
+
+* Should directives be allowed into `BlockStatements` or should we have separate `FunctionBody` and `BlockStatement` nodes where only `FunctionBody` allows directives?
+* Should a `script` allow module statements inside of the body?
+* Should a `module` have an explicit field for `import` statements at the top?
+* Should assignment expression allow left-hand sides that aren't valid assignment targets?
+
+There's probably no correct answer to this question but we can favour one over the other.
+
+A strict AST has the benefit that the mutation API prevents users from constructing invalid trees but it comes at the cost that our parser must fall back to `Unknown*`  nodes more often and a more complex AST.
+
+I would recommend being more lenient so that the parser can flag the use of `import/export` inside a script but we can still provide auto-completion on imports/exports before the user converts the script to a module. But it's probably something we must decide case by case but would love to hear your thoughts.
+
+### Node granularity
+
+Having fine granular nodes has the advantage that the API allows querying for a very specific node when, for example, using `node.descendants::<PreIncrementExpression>()`.
+
+Having more coarse-grained nodes on the other hand has the advantage that it's easier to implement common behaviour that, for example, applies to all Binary expressions.
+
+I guess, this isn't a specific question, but we should think about if we favour fine granular nodes, being even as extreme as having a different node for each `BinaryExpression` operator.
+
+
+## Ungrammar extension proposals
+
+Add support for `///` comments that we can use to document nodes and tokens.
--- a/rfcs/001-js-grammar.ungram
+++ b/rfcs/001-js-grammar.ungram
@@ -1,0 +1,799 @@
+// Rules
+// Create a node when:
+// - contains multiple children (not just a token)
+// - it is part of an union type over nodes
+// otherwise, create a token
+
+// Create a wrapper node when the optionality of two or more children depend on each other.
+// - For example, the 'else' token and the alternate branch must either be both present or absent.
+//      -> Create an ElseClause wrapping the token and the alternate branch.
+// Create a wrapper node for lists of children that use a common separator, for example
+//   `ArrayElement = value: Expression, trailing_comma: ','`
+//   This to guarantee that all tokens can be accessed and that you have a single type of children nodes
+
+JsScript =
+    interpreter: 'js_shebang'? // rslint: shebang
+    directives: 'js_directive'* // Or a directive node?
+    stmts: JsAnyStatement* // Called items
+
+JsModule =
+    interpreter: 'js_shebang'?
+    directives: 'js_directive'*
+    stmts: JsAnyStatement*
+
+// --------------------- Unknowns ---------------------
+// Hack to make ungrammar happy. Let's hope they never ban recursive types xD
+// Add special handling in the source generator to skip the `SyntaxNode`, `SyntaxToken`, and `SyntaxElement` types, so that they resolve to the global `SyntaxNode`, `SyntaxToken`, and `SyntaxElement` definitions.
+SyntaxNode = SyntaxNode
+SyntaxToken = SyntaxToken
+SyntaxElement = SyntaxNode | SyntaxToken
+
+JsUnknownStatement = SyntaxElement*
+JsUnknownExpression = SyntaxElement*
+JsUnknownPattern = SyntaxElement*
+JsUnknownMember = SyntaxElement*
+JsUnknownBinding = SyntaxElement*
+JsUnknownAssignmentTarget = SyntaxElement*
+
+// --------------------- Statements ---------------------
+
+                                        // RSLint
+JsAnyStatement = JsBlockStatement       // BlockStmt
+    | JsBreakStatement                  // BreakStmt
+    | JsClassDeclaration                // ClassDecl
+    | JsDebuggerStatement               // DebuggerStmt
+    | JsEmptyStatement                  // EmptyStmt
+    | JsExpressionStatement             // ExprStmt
+    | JsForStatement                    // ForStmt
+    | JsForInStatement                  // ForInStmt
+    | JsForOfStatement                  // ForOfStmt
+    | JsFunctionDeclaration             // FnDecl
+    | JsIfStatement                     // IfStmt
+    | JsLabeledStatement                // LabelledStmt
+    | JsReturnStatement                 // ReturnStmt
+    | JsSwitchStatement                 // SwitchStmt
+    | JsTryStatement                    // TryStmt
+    | JsVariableDeclarationStatement    // VarDecl
+    | JsWhileStatement                  // WhileStmt
+    | JsWithStatement                   // WithStmt
+    | JsExport                          // ExportDecl
+    | JsExportDefault                   // ExportDefaultDecl
+    | JsExportFrom                      // ExportNamed
+    | JsExportAllFrom                   // ExportWildcard
+    | JsImport                          // ??
+    | JsImportModule                    // ??
+    | JsUnknownStatement
+
+JsBlockStatement =
+    '{' stmts: JsAnyStatement* '}'
+
+JsBreakStatement =
+    'break'
+    (label: 'js_identifier')?
+    ';'?
+
+JsContinueStatement =
+    'continue'
+    (label: 'js_identifier')?
+    ';'?
+
+JsDebuggerStatement =
+    'debugger' ';'?
+
+// RSLint uses Condition for: (test)
+JsDoWhileStatement =
+    'do' body: JsAnyStatement
+    'while' '(' test: JsAnyExpression ')' ';'?
+
+JsEmptyStatement =
+    ';'
+
+JsExpressionStatement =
+    expression: JsAnyExpression ';'?
+
+// Difference to RSLint: no ForInit, ForTest and ForUpdate: RSLint needed the extra kinds so that it could keep the test/update apart
+JsForStatement =
+    'for' '('
+    init: JsForInit? init_semicolon: ';'
+    test: JsAnyExpression? test_semicolon: ';'
+    update: JsAnyExpression?
+    ')'
+    body: JsAnyStatement
+
+JsForInit = JsAnyExpression | JsVariableDeclaration
+
+JsForInStatement =
+    'for' '(' left: JsForLeft 'in' right: JsAnyExpression ')'
+    body: JsAnyStatement
+
+JsForOfStatement =
+    'for' 'await'?
+    '('
+    left: JsForLeft
+    'of'
+    right: JsAnyExpression
+    ')'
+    body: JsAnyStatement
+
+JsForLeft = JsVariableDeclaration |  JsAnyAssignmentTarget
+
+JsFunctionDeclaration =
+    'function' '*'?
+    id: JsIdentifierBinding
+    parameter_list: JsParameterList
+    body: JsFunctionBody
+
+// Doesn't exist in Rome, rome uses the same type for block/function bodies.
+// Reason for splitting: Directives can only exist in function bodies (anything introducing new scope)
+JsFunctionBody =
+    '{'
+    directives: 'directive'*
+    body: JsAnyStatement*
+    '}'
+
+JsIfStatement =
+    'if' '(' test: JsAnyExpression ')'
+    consequence: JsAnyStatement
+    else_clause: JsElseClause?
+
+JsElseClause =
+    'else'
+    alternate: JsAnyStatement
+
+JsLabeledStatement =
+    label: 'js_identifier' ':'
+    body: JsAnyStatement
+
+JsReturnStatement =
+    'return' argument: JsAnyExpression? ';'?
+
+JsSwitchStatement =
+    'switch' '(' discriminant: JsAnyExpression ')'
+    cases: JsSwitchCase*
+
+// Introduce a case clause / default because the optionality of the tokens depend on which variant it is
+// An alternative would be to have preDefaultCases, defaultCases, postDefaultCases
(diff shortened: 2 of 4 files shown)


crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+JsArrayRestBinding =
+    '...'
+    binding: JsBinding
+
+
+// let {a, b, ...rest } =
+//     ^^^^^^^^^^^^^^^^
+JsObjectBinding =
+    '{'
+        properties: (AnyJsPropertyBinding (',' AnyJsPropertyBinding)* ','?)
```
ematipico, line 684: I think that "orphan" semicolons are forbidden inside an object binding. While they are OK inside an array.

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
+
+## Codegen extensions
+
+* **Must**: Add support for labels so that we can influence how the child accessors are named. For example, we want to name the left and right expressions of `BinaryExpression` `left` and `right` and not `expression` and `expression. Ungrammar supports the `label: Type` syntax. All we need to do is to respect the naming when generating the fields.
```
ematipico, line 4:
```suggestion
* **Must**: add support for labels so that we can influence how the child accessors are named. For example, we want to name the left and right expressions of `BinaryExpression` `left` and `right` and not `expression` and `expression. Ungrammar supports the `label: Type` syntax. All we need to do is to respect the naming when generating the fields.
```

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+* **Must**: Support inline token-union types. For example, the binary expression operator can be any of `==, +, -, ...`. That's why it's defined as `operator: ( '==' | '+' | '-' ...)`. The source gen should only generate a single field for the operator that returns a `SyntaxToken`
```
ematipico, line 5:
```suggestion
* **Must**: support inline token-union types. For example, the binary expression operator can be any of `==, +, -, ...`. That's why it's defined as `operator: ( '==' | '+' | '-' ...)`. The source gen should only generate a single field for the operator that returns a `SyntaxToken`
```

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+* **Must**: Generate different return types for mandatory/optional children.
```
ematipico, line 6:
```suggestion
* **Must**: generate different return types for mandatory/optional children.
```

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+* **Must/Wish**: It's convenient to define unions in respect of other unions. For example, `JsClassMemberName = JsObjectMemberName | JsPrivateClassMemberName` where `JsObjectMemberName` is an union type as well. This is convenient to maintain the grammar but makes the Facade more awkward to use because it requires two matches: first on the outer union and then on the inner union. We can avoid this if we flatten unions inside of the source gen and automatically generate `From<InnerUnion>` implementations.
```
ematipico, line 7:
```suggestion
* **Must/Wish**: it's convenient to define unions in respect of other unions. For example, `JsClassMemberName = JsObjectMemberName | JsPrivateClassMemberName` where `JsObjectMemberName` is an union type as well. This is convenient to maintain the grammar but makes the Facade more awkward to use because it requires two matches: first on the outer union and then on the inner union. We can avoid this if we flatten unions inside of the source gen and automatically generate `From<InnerUnion>` implementations.
```

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+* **Wish**: Automatically strip the `Js` extensions from field names to reduce the need for explicit labels. For example, `ExpressionStatement` must use an explicit label for its inner expression so that it isn't named `js_expression`.
```
ematipico, line 8:
```suggestion
* **Wish**: automatically strip the `Js` extensions from field names to reduce the need for explicit labels. For example, `ExpressionStatement` must use an explicit label for its inner expression so that it isn't named `js_expression`.
```

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+* **Wish**: Generation of a union type over all nodes/tokens of a language, `JsElement = ..., JsNode = ..., JsToken = ...`
```
ematipico, line 9:
```suggestion
* **Wish**: generation of a union type over all nodes/tokens of a language, `JsElement = ..., JsNode = ..., JsToken = ...`
```

ematipico (commented):
Left some comments.

ematipico:
Re: the `+` operator, I think we should evaluate creating a PR inside `ungrammar` repository. At least, propose it and see it our feature makes sense.

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+JsUnknownExpression = SyntaxElement*
+JsUnknownPattern = SyntaxElement*
+JsUnknownMember = SyntaxElement*
+JsUnknownBinding = SyntaxElement*
+JsUnknownAssignmentTarget = SyntaxElement*
+
+// --------------------- Statements ---------------------
+
+                                        // RSLint
+JsStatement = JsBlockStatement          // BlockStmt
```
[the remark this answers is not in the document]
MichaReiser: Makes sense. I'll abandon this PR once we agreed on the important points and we can then change your grammar one by one, making sure we use consistent formatting.

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+
+JsLabeledStatement =
+    label: 'js_identifier' ':'
+    body: JsStatement
+
+JsReturnStatement =
+    'return' argument: JsExpression? ';'?
+
+JsSwitchStatement =
+    'switch' '(' discriminant: JsExpression ')'
```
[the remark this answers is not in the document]
MichaReiser: haha, copied from roslyn :)

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+    cases: JsSwitchCase*
```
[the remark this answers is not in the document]
MichaReiser: That's correct but I'm not sure if we can encode this into the grammar (without making it super awkward). The main issue is that we can't change the ordering. So having something like `cases: JsSwitchCase* default: JsDefaultCase` doesn't work because the `default` case must not necessarily be the last case. Do you have a suggestion on how to encode this?

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+
+// Introduce a case clause / default because the optionality of the tokens depend on which variant it is
+JsSwitchCase = JsCaseClause | JsDefaultClause
+JsCaseClause = 'case' test: JsExpression ':'
+JsDefaultClause = 'default' ':'
```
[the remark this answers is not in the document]
MichaReiser: Lol yes :)

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+JsVariableDeclarationStatement =
+    declaration: JsVariableDeclaration ';'?
+
+// TODO unable to type: at least one child. Add support for + (at least one) to ungrammar or split into declarator: JsVariableDeclarator, other_declarators: JsVariableDeclarator*
+JsVariableDeclaration =
+    kind_token: ('var' | 'let' | 'const')
+    declarators: JsVariableDeclarator*
+
+JsVariableDeclarator =
+    id: JsBinding '=' init: JsExpression?
```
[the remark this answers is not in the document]
MichaReiser: Right, we should change `declarators` to `(JsVariableDeclarator (',' JsVariableDeclarator)*)` (maybe with trailing)?

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+    | JsStringLiteral                   // Literal
+    | JsTemplateLiteral
+    | JsUnknownExpression
+
+JsArrayExpression =
+    '['
+    // Not 100% correct because array holes must have a trailing comma. But enforcing
+    // this would make it too awkward to use
+    elements: (JsArrayElement (',' JsArrayElement)* ','?)
+    ']'
```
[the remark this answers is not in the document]
MichaReiser: Isn't needed. An array with a `;` at the end is an expression statement containing an array expression.

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+JsCallExpression =
+    callee: JsExpression
+    argument_list: JsArgumentList
+
+JsConditionalExpression =
+    test: JsExpression
+    '?'
+    consequent: JsExpression
+    ':'
+    alternate: JsExpression
```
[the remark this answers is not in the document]
MichaReiser: Same here, expressions are wrapped in an ExpressionStatement if they end with a `;`

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+// Parser guarantees that this list is never empty, always contains at least on element
+// Rome uses two different lists: expressions and quasis
+JsTemplateLiteral =
+    left_tick: '`'
+    elements: JsTemplateElement*
+    right_tick: '`'
+
+JsTemplateElement = JsStringTemplateElement | JsExpression
+
+JsStringTemplateElement = value: 'js_string_template_element'
```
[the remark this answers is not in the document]
MichaReiser: I created `JsStringTemplateElement` because unions should always be over nodes and not tokens. 

I created a `js_string_template_element` because it isn't a quoted string as it is the case for `js_string`

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+JsImportClause = JsImportDefaultBinding | JsNamespaceImportClause | JsNamedImportClause
+
+// Rome wraps the local name in JSImportSpecifierLocal
+JsImportDefaultBinding =
+    local_name: JsBindingIdentifier
+    // The comma is required if followed by a {} but otherwise forbidden. Introducing another node just to
+    // enforce this in namspace import clause / named import clause feels overkill
+    trailing_comma: ','?
+
+// Flatten the 'as' + name or use 'JsImportBinding'?
```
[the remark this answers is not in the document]
MichaReiser: For what reason?

crates/rome-js-grammar/src/grammar.ungram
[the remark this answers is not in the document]
ematipico: Right, I didn't consider the ordering. Unfortunately no, I don't have a better suggestion (only suggestions that would feel strange). My only conclusion would be to leave the grammar like this, and make a manual implementation of `default() -> Result<DefaultClause, MissingElement>`, where the function checks if there's at least one and not more than one child of type `DefaultClause`

crates/rome-js-grammar/src/grammar.ungram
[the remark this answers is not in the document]
ematipico: Yeah I think it makes sense!

crates/rome-js-grammar/src/grammar.ungram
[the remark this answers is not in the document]
ematipico: I believe they "belong" to the same piece of grammar, meaning that if the future something changes, there's just one place. 

Other than that, I don't have any other strong opinions.

crates/rome-js-grammar/src/grammar.ungram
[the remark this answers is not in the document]
MichaReiser: Having a manual implementation makes sense but we should probably define a new error for that use case, because the default isn't really missing.

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+**Recommendation**
+I would suggest going with Option A for now. We can still go for option B if we see us re-implementing this pattern over and over again.
+
+### Member vs Property naming
+
+Should we call object and class members `Members` or `Properties`?
+
+* `MemberExpression` has an `object` and `property` field today which feels inconsistent. Shouldn't property be called member or should it be a `PropertyExpression`
```
yassere, line 115: The terminology seems in line with the [spec](https://tc39.es/ecma262/#sec-property-accessors), if I'm understanding it correctly. I think this is probably fine as-is.

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+
+A strict AST has the benefit that the mutation API prevents users from constructing invalid trees but it comes at the cost that our parser must fall back to `Unknown*`  nodes more often and a more complex AST.
+
+I would recommend being more lenient so that the parser can flag the use of `import/export` inside a script but we can still provide auto-completion on imports/exports before the user converts the script to a module. But it's probably something we must decide case by case but would love to hear your thoughts.
+
+### Node granularity
+
+Having fine granular nodes has the advantage that the API allows querying for a very specific node when, for example, using `node.descendants::<PreIncrementExpression>()`.
+
+Having more coarse-grained nodes on the other hand has the advantage that it's easier to implement common behaviour that, for example, applies to all Binary expressions.
```
yassere, line 140: Could we get the best of both worlds by having additional specific AST nodes that allow you to refine a coarse node? The specific nodes would need additional checks on the `can_cast` method (that would probably be hand-written in many cases). And we'd implement `From<SpecificNode> for CoarseNode` to make them easier to use.

Alternately, we could add granularity through utility functions: `node.descendants().filter_map(UnaryExpression::is_pre_increment)`

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+    body: JsFunctionBody
+
+
+// --------------------- MODULES ---------------------
+
+JsExport =
+    'export' declaration: JsExportDeclaration
+    ';'?
+
+JsExportDeclaration = JsFunctionDeclaration | JsClassDeclaration | JsVariableDeclarationStatement
```
jamiebuilds, line 556: Should we make it a rule to name all of these unions `Any*`?

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+JsArgumentList =
+    '('
+    arguments: (JsArgument (',' JsArgument)* ','?)
+    ')'
+
+JsArgument = JsExpression | JsSpread
+
+
+JsIdentifier =
+    name: 'js_identifier'
```
jamiebuilds, line 799: Should this be named like `JsStatic*Name` or just get folded into `JsStaticMemberName`?

crates/rome-js-grammar/src/grammar.ungram
```diff
@@ -0,0 +1,802 @@
...
+
+JsBinding = JsObjectBinding | JsArrayBinding | JsBindingIdentifier | JsUnknownBinding
+
+JsDefaultValueClause =
+    '='
+    value: JsExpression
+
+// let x = OR function(test) {}
+//     ^               ^^^^
+JsBindingIdentifier = name: 'js_identifier'
```
jamiebuilds, line 658: Should this be named `JsIdentifierBinding`?

jamiebuilds:
I think we should also take a moment to consider how some of these nodes relate to the TypeScript and JSX grammars. For example there are a number of different types of identifiers that we have already:

- `JsBindingIdentifier`
- `JsIdentifierAssignmentTarget`
- `JsIdentifier`
- `JsReferenceIdentifierExpression`

TypeScript has parallels for almost all of these (except the `JsIdentifierAssignmentTarget`):

```ts
type TsBindingIdentifier<
  TsBindingIdentifier extends TsIdentifierReferenceExpression = TsIdentifierReferenceExpression
> = {
  TsStaticMemberName: TsIdentifierReferenceExpression.TsStaticMemberName
}
```

JSX also has a few:

```tsx
<>
  {/* member expression */}
  <jsxIdentifier.jsxIdentifier jsxIdentifier/>
  <JsxReferenceIdentifier.jsxIdentifier/>
  {/* namespaced name */}
  <JsxReferenceIdentifier:jsxIdentifier jsxIdentifier:jsxIdentifier/>
  {/* tsx */}
  <JsxReferenceIdentifier<TsReferenceIdentifier> />
</>
```

Initially I was leaning towards saying that we should reuse grammar names when possible (i.e. No separate `JsBindingIdentifier` and `TsBindingIdentifier`) thinking that would be better for letting people operate on these nodes abstractly. However, that also makes it worse for people that are trying to work with a specific grammar and now need their own logic to figure out where in the grammar they are. It would also imply that we should try aligning in more places like using `JsMemberExpression` instead of `JsxMemberExpression`, but these have important differences in grammar (`<foo[bar]/>` is not valid JSX syntax for example).

For the people who do want to work with identifiers in the abstract, we could have separate tools for doing so:

```rs
// just as an example, each identifier-like node could implement a shared trait
trait Identifier { fn name() -> String }
impl Identifier for JsBindingIdentifier { .. }
impl Identifier for JsIdentifierAssignmentTarget { .. }
// or belong to some enum
enum AnyIdentifier { .. }
// etc...
if node.is_identifier() { .. } // check if a node is some identifier-like node
any_identifier.name(); // get the name of some identifier-like node
```

We should come up with general categories for these in that case:

- Reference Identifiers: `JsIdentifierReferenceExpression`, `TsIdentifierReferenceExpression`, `JsxIdentifierReferenceExpression`
- Binding Identifiers: `JsBindingIdentifier`, `TsBindingIdentifier`

Maybe even further into the grammar:

- Static Member Access: `JsMemberExpression`, `TsQualifiedName`, `JsxMemberExpression`, `JsxNamespacedName`
- Computed Member Access: `JsMemberExpression`, `TsIndexedAccessType`

If we do want to go further into the grammar, we might want to consider splitting some nodes like `JsMemberExpression` up into things like `JsStaticMemberExpression` and `JsComputedMemberExpression` (This is what Shift AST does).

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+## Open Questions
+
+### Variable Declaration
+
+`JsVariableDeclaration` can either contain one or many `declarators`. The grammar could define this as:
+
+```
+JsVariableDeclaration =
+    kind_token: ('var' | 'let' | 'const')
+    declarators: JsVariableDeclarator+
```
jamiebuilds, line 78: Can't this be:

```
JsVariableDeclaration =
    kind_token: ('var' | 'let' | 'const')
    declarators: (JsVariableDeclarator (',' JsVariableDeclarator)*)
```

crates/rome-js-grammar/src/rfc.md
jamiebuilds: Am I misunderstanding something here, it looks like we're using `member: JsMember` not `property`

crates/rome-js-grammar/src/rfc.md
```diff
@@ -0,0 +1,148 @@
...
+
+Having more coarse-grained nodes on the other hand has the advantage that it's easier to implement common behaviour that, for example, applies to all Binary expressions.
+
+I guess, this isn't a specific question but we should think about if we favour fine granular nodes, being even as extreme as having a different node for each `BinaryExpression` operator.
+
+
+
+## Ungrammar extension proposals
+
+Add support for `///` comments that we can use to document nodes and tokens.
```
jamiebuilds, line 148: JSDoc and fam use `/** ... */` fairly universally. Did you want to parse inside these comments?

crates/rome-js-grammar/src/rfc.md
yassere: Sorry, I meant that I think it's fine the way Rome's original AST has it. But I don't have a strong opinion here.

MichaReiser:
> 
> 
> I think we should also take a moment to consider how some of these nodes relate to the TypeScript and JSX grammars. For example there are a number of different types of identifiers that we have already:
> 
>     * `JsBindingIdentifier`
> 
>     * `JsIdentifierAssignmentTarget`
> 
>     * `JsIdentifier`
> 
>     * `JsReferenceIdentifierExpression`
> 
> 
> TypeScript has parallels for almost all of these (except the `JsIdentifierAssignmentTarget`):
> 
> ```ts
> type TsBindingIdentifier<
>   TsBindingIdentifier extends TsIdentifierReferenceExpression = TsIdentifierReferenceExpression
> > = {
>   TsStaticMemberName: TsIdentifierReferenceExpression.TsStaticMemberName
> }
> ```
> 
> JSX also has a few:
> 
> ```tsx
> <>
>   {/* member expression */}
>   <jsxIdentifier.jsxIdentifier jsxIdentifier/>
>   <JsxReferenceIdentifier.jsxIdentifier/>
>   {/* namespaced name */}
>   <JsxReferenceIdentifier:jsxIdentifier jsxIdentifier:jsxIdentifier/>
>   {/* tsx */}
>   <JsxReferenceIdentifier<TsReferenceIdentifier> />
> </>
> ```
> 
> Initially I was leaning towards saying that we should reuse grammar names when possible (i.e. No separate `JsBindingIdentifier` and `TsBindingIdentifier`) thinking that would be better for letting people operate on these nodes abstractly. However, that also makes it worse for people that are trying to work with a specific grammar and now need their own logic to figure out where in the grammar they are. It would also imply that we should try aligning in more places like using `JsMemberExpression` instead of `JsxMemberExpression`, but these have important differences in grammar (`<foo[bar]/>` is not valid JSX syntax for example).

I haven't spent much time thinking about that yet (maybe I should have). My first question would be if we feel comfortable moving forward with a JS only grammar for now and figure out some of JS's dialects after (or while implementing the JS grammar). My motivation for delaying the decision is that implementing the JS grammar will give us a better understanding of what worked well and what doesn't. 

I'm very much undecided right now on which approach to take and would need to look closer into how the grammars differ but I'm probably more leaning toward different nodes if they differ significantly. My main thoughts from a grammar perspective are:

* Whatever a node's child is required should never depend on the presence (or absence) of a sibling. This is mainly a concern if certain tokens/nodes are required in TypeScript but not in JavaScript. 
* We should define a new node whenever the supported syntax for a child differs (JSX member expression example) 


> For the people who do want to work with identifiers in the abstract, we could have separate tools for doing so:
> 
> ```rust
> // just as an example, each identifier-like node could implement a shared trait
> trait Identifier { fn name() -> String }
> impl Identifier for JsBindingIdentifier { .. }
> impl Identifier for JsIdentifierAssignmentTarget { .. }
> // or belong to some enum
> enum AnyIdentifier { .. }
> // etc...
> if node.is_identifier() { .. } // check if a node is some identifier-like node
> any_identifier.name(); // get the name of some identifier-like node
> ```

I'm leaning towards an enum here because we already have the infrastructure in place to define it inside of our grammar and testing if something is an instance of it ``AnyJsIdentifier::cast(node.syntax())`


> 
> For the people who do want to work with identifiers in the abstract, we could have separate tools for doing so:
> 
> ```rust
> // just as an example, each identifier-like node could implement a shared trait
> trait Identifier { fn name() -> String }
> impl Identifier for JsBindingIdentifier { .. }
> impl Identifier for JsIdentifierAssignmentTarget { .. }
> // or belong to some enum
> enum AnyIdentifier { .. }
> // etc...
> if node.is_identifier() { .. } // check if a node is some identifier-like node
> any_identifier.name(); // get the name of some identifier-like node
> ```
> 
> We should come up with general categories for these in that case:
> 
>     * Reference Identifiers: `JsIdentifierReferenceExpression`, `TsIdentifierReferenceExpression`, `JsxIdentifierReferenceExpression`
> 
>     * Binding Identifiers: `JsBindingIdentifier`, `TsBindingIdentifier`
> 
> 
> Maybe even further into the grammar:
> 
>     * Static Member Access: `JsStaticMemberExpression`, `TsQualifiedName`, `JsxMemberExpression`, `JsxNamespacedName`
> 
>     * Computed Member Access: `JsComputedMemberExpression`, `TsIndexedAccessType`
> 

I like this categorisation and having these "abstract" types will simplify writing analysis a lot. Roslyn has the concept of `IOperation` that abstract a common operation across languages (for example, creating an array). 

> If we do want to go further into the grammar, we might want to consider splitting some nodes like `JsMemberExpression` up into things like `JsStaticMemberExpression` and `JsComputedMemberExpression` (This is what Shift AST does).

I actually liked Rome's approach of computed/static a bit better because it gets away with fewer node types but it certainly makes sense to split it if it helps to categorise the node types.

crates/rome-js-grammar/src/grammar.ungram
ematipico: > I mainly ask myself the question if it's important for authors to know if they're working with a Union or a concrete node type. 

This is really a good point, especially if we plan to flatten the unions. If we decided to stick with certain naming convention, this would help developers and contributors.

crates/rome-js-grammar/src/rfc.md
MichaReiser: @sebmck what's your opinion on calling properties "member" in the new AST? 

We don't think aligning with the spec is that important because the names may change in the future and the name property seems more historical.

sebmck: Sure, I don't think it matters what we call them as long as we follow some rules to remain consistent. We should definitely be using spec terminology where we can but deviating is entirely fine if we think it makes it more clear.

crates/rome-js-grammar/src/grammar.ungram
jamiebuilds: I kinda worry about using the more generic sounding `JsIdentifier` because the other types have been split out:

```
JsBindingIdentifier
JsIdentifierAssignmentTarget
JsIdentifier
JsReferenceIdentifierExpression
```

A user might match on `JsIdentifier` thinking it will return all kinds of identifiers, especially since that is the case in Babel/etc.

crates/rome-js-grammar/src/grammar.ungram
MichaReiser: babel doesn't seem to care about a trailing comma.

MichaReiser:
Moved the files to the RFC folder. Updated the JS grammar with the proposed changes.

Issue #1725, opened 2021-10-29
☂️ AST Façade Improvements


MichaReiser:
### Description

> RSLint's and Rust-analyzer's child accessors defined on the AST nodes return `Option<TNode>`. For example, the `IfStmt` child accessors are defined as follow (implementations omitted for brevity):
> 
> ```rust
> impl IfStmt {
> 	pub fn if_token(&self) -> Option<SyntaxToken>;
> 	pub fn condition(&self) -> Option<Condition>;
> 	pub fn cons(&self) -> Option<Stmt>;
> 	pub fn else_token(&self) -> Option<SyntaxToken>;
> 	pub fn alt(&self) -> Option<Stmt>;
> }
> ```
> 
> We identified that such an API has two major shortcomings for our use case. Let's explore these using the following example:
> 
> ```rust
> if (test)) else {}
> ```
> 
> RSLint creates the following CST for this script:
> 
> ```
> SCRIPT@0..19
>   IF_STMT@0..18
>     IF_KW@0..2 "if"
>     WHITESPACE@2..3 " "
>     CONDITION@3..9
>       L_PAREN@3..4 "("
>       NAME_REF@4..8
>         IDENT@4..8 "test"
>       R_PAREN@8..9 ")"
>     **ERROR@9..10
>       R_PAREN@9..10 ")"**
>     WHITESPACE@10..11 " "
>     ELSE_KW@11..15 "else"
>     WHITESPACE@15..16 " "
>     BLOCK_STMT@16..18
>       L_CURLY@16..17 "{"
>       R_CURLY@17..18 "}"
>   WHITESPACE@18..19 "\n"
> ```
> 
> 1. The condition of the `IfStmt` has one right parenthesis too many. RSLint excellently recovers from this error by inserting an `Error` in the place of the `if`'s consequence. The problem with the API is that accessing `if_stmt.cons()` returns `None` because `Error` isn't a `Stmt`. This has the advantage that a linter will not bother and try to dive into the body but a formatter needs the `Error` node to format its content (even if that just means printing it as it was in the original source).
> 2. It's not clear from the API what are the mandatory and optional children. Correctly handling the case where a mandatory child is missing is important where, on the other hand, it's safe to skip missing optional children. That's why the API should guide users to handle missing mandatory children. 
> 
> Because of this, we plan to make the following two changes to the facade:
> 
[...]

PR #1736, opened 2021-11-01
Change the visibility of green nodes to internal

MichaReiser:
## Summary

Changes the visibility of the `Green*` and `cursor::Syntax*` types to `pub(crate)` from previously `pub`. 

The motivation behind changing the visibility is that the `Green*` types are an internal optimisation of our CST to reduce memory usage by reusing identical nodes and tokens. The cursor types are also just internal representations that help navigate through a tree and is used by the public `Syntax*` types.

This PR has to remove the `replace_with` methods for now as they used to operate on `GreenNode`s. We should revisit the exact signature and behaviour when working on the mutation API. I may even suggest that the functionality should be super limited because mutating with `SyntaxNode`s directly voids all guarantees that the mutations will form a valid CST tree. Ideally, mutations should be limited to `AstNode`s (that e.g. use a `TreeWriter` to mutate the tree). 

RSLint exposes the `GreenNode` on `Script` with a note that that's useful for multithreading. I believe it should be sufficient for our use cases to just clone the root of the `SyntaxTree` and move it to the new thread rather than passing `GreenNode`s around.

Part of #1725 

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_rowan/src/tree_builder.rs
+++ b/crates/rome_rowan/src/tree_builder.rs
@@ -3,7 +3,6 @@ use crate::{
 	green::{GreenElement, NodeCache},
 	Language, NodeOrToken, SyntaxNode,
 };
-use std::marker::PhantomData;
 
 /// A checkpoint for maybe wrapping a node. See `GreenNodeBuilder::checkpoint` for details.
 #[derive(Clone, Copy, Debug)]
@@ -15,7 +14,6 @@ pub struct TreeBuilder<'cache, L: Language> {
 	cache: CowMut<'cache, NodeCache>,
 	parents: Vec<(L::Kind, usize)>,
 	children: Vec<(u64, GreenElement)>,
-	ph: PhantomData<L>,
 }
 
 impl<L: Language> Default for TreeBuilder<'_, L> {
@@ -24,7 +22,6 @@ impl<L: Language> Default for TreeBuilder<'_, L> {
 			cache: CowMut::default(),
 			parents: Vec::default(),
 			children: Vec::default(),
-			ph: PhantomData,
 		}
 	}
 }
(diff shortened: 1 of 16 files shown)


crates/rome_rowan/src/tree_builder.rs
```diff
@@ -1,47 +1,61 @@
...
-#[derive(Default, Debug)]
-pub struct GreenNodeBuilder<'cache> {
+/// A builder for a syntax tree.
+#[derive(Debug)]
+pub struct TreeBuilder<'cache, L: Language> {
 	cache: CowMut<'cache, NodeCache>,
-	parents: Vec<(SyntaxKind, usize)>,
+	parents: Vec<(L::Kind, usize)>,
 	children: Vec<(u64, GreenElement)>,
+	ph: PhantomData<L>,
```
ematipico, line 18: What's the use case of having `PhantomData` here?

ematipico (approved):

crates/rome_rowan/src/tree_builder.rs
MichaReiser: Nice catch! It was needed in one of my first versions because I didn't change the `SyntaxKind` to `L::Kind`. The problem then was that the generic argument `L` was never used but Rust requires that all generic params are used in at least one field. That's when you could use `PhantomData`, it's a field that, to my understanding, has no runtime representation but allows the compiler to catch the generic parameter.

yassere:
I'm not 100% sure there's no valid use case for exposed green nodes, but I think this change is fine for now. If there's ever a need to expose green nodes in the future, we can discuss the justification at that point.

> I believe it should be sufficient for our use cases to just clone the root of the SyntaxTree and move it to the new thread rather than passing GreenNodes around.

Syntax nodes in this version of rowan are not thread-safe, and I think before changing that, we'd want to measure the impact of atomic reference counting during every traversal. One alternative could be to make a thread-safe `SyntaxRoot` type that can be converted back to a `SyntaxNode` after sending it to a new thread.

MichaReiser:
Good point. I was about to create a SyntaxTree node that holds a reference to the green node. Calling root on the tree creates the SyntaxNode. We could make this type Sync and Send as you suggested

Issue #1720, opened 2021-10-27
☂️ Attach trivia to tokens


xunilrj:
### Description

part of: #1718

# Pending Decisions

- storing number of spaces/tabs (swift) or a string containing all whitespace (Roslyn)?
- Token { kind: Ident, text: "foo", trailing_trivia: Trivia::spaces(1), span: Span(4) } (one option)
modelling trivia as cached references (similar to token/nodes) or as enum (Space, Tab, NewLine, SingleLineComment, - - MultiLineComment) and store them by value?
- unifying GreenNode , GreenToken and GreenTrivia into a single GreenNode(Roslyn) or keep as is (Rust analyzser, Swift)?

# Tasks

- [x] Modify lexer to generate two tokens for "\n ". "\n" will be appended to the trailing trivia of the last token, and " " will be prepended to the leading trivia of the next token. See #1716
- [x] Fix lexer, because we actually want the "newline" token to always be in the leading, not in the trailing. Including the ```EOF``` token.
- [x] Changes to Rowan. See #1783 and #1798
- [x] Changes to the parser. See #1801

# Research

## Lexer merges newline and trailing spaces

```rust
let text = "let a = 1; // nice variable \n /*hey*/ let b = 2; // another nice variable";
let (mut tokens, mut errors) = tokenize(text, 0);
let mut i = 0;
for token in &tokens {
    println!("{:?} {:?}", &text[i..(i + token.len)], token);
    i += token.len;
}
```

generates

```
"let" Token { kind: IDENT, len: 3 }
" " Token { kind: WHITESPACE, len: 1 }
"a" Token { kind: IDENT, len: 1 }
" " Token { kind: WHITESPACE, len: 1 }
"=" Token { kind: EQ, len: 1 }
" " Token { kind: WHITESPACE, len: 1 }
"1" Token { kind: NUMBER, len: 1 }
";" Token { kind: SEMICOLON, len: 1 }
" " Token { kind: WHITESPACE, len: 1 }
"// nice variable " Token { kind: COMMENT, len: 17 }
"\n " Token { kind: WHITESPACE, len: 2 }
"/*hey*/" Token { kind: COMMENT, len: 7 }
" " Token { kind: WHITESPACE, len: 1 }
"let" Token { kind: IDENT, len: 3 }
" " Token { kind: WHITESPACE, len: 1 }
"b" Token { kind: IDENT, len: 1 }
" " Token { kind: WHITESPACE, len: 1 }
[...]

PR #1738, opened 2021-11-01
Attaching trivia to GreenToken

xunilrj:
part of: #1720

# Tasks

- [x] ```GreenToken``` will have an enum with the trivial case and a fallback using Vec.
- [ ] ```SyntaxTrivia``` will wrap ```GreenTokenTrivia```
- [x] Change ```PartialEq``` and ```Hash``` implementations for the *Token structs to consider the trivia.
- [x] Change ```Display``` and ```Debug``` implementations to print trivia.
- [x] Improve ```Display``` and ```Debug```
- [x] Change the ```NodeCache``` to respect trivia (not fixed by changing ```Hash``` implementation)
- [ ] Change ```LossyTreeSink```
- [x] Change ```LosslessTreeSink```
- [x] Fix GreenToken slice to contains leading, token and trailing trivia. In one slice. Otherwise the ```Syntax``` layer end up with all offsets wrong.
- [x] Fix doctests
- [x] Improve ```test_trivia_attached_to_tokens```

## Debug

```
  VAR_DECL@0..11
    " " IDENT@0..5 "let" " "         <- Range is bigger than "let" because we need the trivia now
    DECLARATOR@5..10
      SINGLE_PATTERN@5..7
        NAME@5..7
           IDENT@5..7 "a" " "
       EQ@7..9 "=" " "
      LITERAL@9..10
         NUMBER@9..10 "1"
     SEMICOLON@10..11 ";"
```

# Decisions

- Now the token "slice" contains all its trivia.
- To trim this trivia we have to ask leading().text_len() and trailing().text_len() and computer where the token text is.
- Formatter is not working because we must manually insert trailing space between tokens. Maybe we need a couple of helper methods for this.

# How to test

```rust
> cargo test -p rslint_parser
> cargo xtask coverage
...
...
``` 

Coverage is the same as before.

### DIFF
--- a/crates/rome_rowan/src/green/token.rs
+++ b/crates/rome_rowan/src/green/token.rs
@@ -152,6 +152,7 @@ impl GreenTokenData {
 
 	/// Returns the length of the text covered by this token.
 	#[inline]
+	#[allow(dead_code)]
 	pub fn text_len(&self) -> TextSize {
 		TextSize::of(self.text())
 	}
@@ -181,6 +182,13 @@ impl GreenTokenData {
 
 	pub(crate) fn cache_hash(&self) -> u64 {
 		Self::cache_hash_of(self.kind(), self.text_with_trivia())
+	}
+
+	pub fn leading_trailing_total_len(&self) -> (TextSize, TextSize, TextSize) {
+		let leading_len = self.data.header.leading_trivia.text_len();
+		let trailing_len = self.data.header.trailing_trivia.text_len();
+		let total_len = self.data.slice().len() as u32;
+		(leading_len, trailing_len, total_len.into())
 	}
 }
 
--- a/crates/rome_rowan/src/green/token.rs
+++ b/crates/rome_rowan/src/green/token.rs
@@ -127,6 +127,16 @@ impl GreenTokenData {
 	pub fn text_len(&self) -> TextSize {
 		TextSize::of(self.text())
 	}
+
+	#[inline]
+	pub fn leading(&self) -> &[GreenTokenTrivia] {
+		self.data.header.leading_trivia.as_slice()
+	}
+
+	#[inline]
+	pub fn trailing(&self) -> &[GreenTokenTrivia] {
+		self.data.header.trailing_trivia.as_slice()
+	}
 }
 
 impl GreenToken {
--- a/crates/rome_rowan/src/green/token.rs
+++ b/crates/rome_rowan/src/green/token.rs
@@ -1,11 +1,12 @@
 use std::{
 	borrow::Borrow,
-	fmt,
+	fmt::{self, Pointer},
 	mem::{self, ManuallyDrop},
 	ops, ptr,
 };
 
 use countme::Count;
+use text_size::TextLen;
 
 use crate::{
 	arc::{Arc, HeaderSlice, ThinArc},
@@ -13,17 +14,88 @@ use crate::{
 	TextSize,
 };
 
-#[derive(PartialEq, Eq, Hash, Debug)]
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub enum Trivia {
+	Whitespace(usize),
+	Comment(usize),
+}
+
+impl Trivia {
+	pub fn as_thin(self, text: &str) -> ThinTrivia {
+		let ptr = ThinArc::from_header_and_iter(self, text.bytes());
+		ThinTrivia(ptr)
+	}
+
+	fn text_len(&self) -> TextSize {
+		match self {
+			Trivia::Whitespace(n) => (*n as u32).into(),
+			Trivia::Comment(n) => (*n as u32).into(),
+		}
+	}
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct ThinTrivia(ThinArc<Trivia, u8>);
+
+impl ThinTrivia {
+	#[inline]
+	pub fn trivia(&self) -> &Trivia {
+		&self.0.header
+	}
+
+	#[inline]
+	pub fn text(&self) -> &str {
+		unsafe { std::str::from_utf8_unchecked(self.0.slice()) }
+	}
+
+	#[inline]
+	fn text_len(&self) -> TextSize {
+		self.0.header.text_len()
+	}
+}
+
+impl fmt::Debug for ThinTrivia {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		write!(f, "{:?}", self.0.slice())
+	}
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
 pub enum GreenTokenTrivia {
-	Whitespace,
+	None,
+	One(ThinTrivia),
+	Many(Vec<ThinTrivia>),
+}
+
+impl GreenTokenTrivia {
+	fn text_len(&self) -> TextSize {
+		match self {
+			GreenTokenTrivia::None => 0.into(),
+			GreenTokenTrivia::One(x) => x.text_len(),
+			GreenTokenTrivia::Many(v) => v.iter().fold(0.into(), |len, x| len + x.text_len()),
+		}
+	}
+}
+
+impl fmt::Debug for GreenTokenTrivia {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		match self {
+			Self::None => Ok(()),
+			Self::One(trivia) => write!(f, "{:?}", trivia.text()),
+			Self::Many(trivia) => {
+				let items: Vec<_> = trivia.iter().map(|x| x.text()).collect();
+				f.debug_list().entries(items).finish()
+			}
+		}
+	}
 }
 
 #[derive(PartialEq, Eq, Hash)]
 struct GreenTokenHead {
 	kind: SyntaxKind,
 	_c: Count<GreenToken>,
-	leading_trivia: Vec<GreenTokenTrivia>,
-	trailing_trivia: Vec<GreenTokenTrivia>,
+	leading_trivia: GreenTokenTrivia,
+	trailing_trivia: GreenTokenTrivia,
 }
 
 type Repr = HeaderSlice<GreenTokenHead, [u8]>;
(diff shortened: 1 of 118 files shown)


crates/rome_rowan/src/green/builder.rs
```diff
@@ -32,16 +32,27 @@ impl GreenNodeBuilder<'_> {
...
+	// Adds new token to the current branch.
 	#[inline]
-	pub fn token(&mut self, kind: SyntaxKind, text: &str) {
-		let (hash, token) = self.cache.token(kind, text);
+	pub fn token(
+		&mut self,
+		kind: SyntaxKind,
+		text: &str,
+		leading: Vec<super::token::GreenTokenTrivia>,
+		trailing: Vec<super::token::GreenTokenTrivia>,
```
MichaReiser, line 42: We probably don't want to use `Vec`s to store the leading and trailing trivia, instead, we should store it inside of the `HeaderSlice` (as the slice)`. This avoids another heap allocation. 

```
type Repr = HeaderSlice<GreenTokenHead, [u8]>;
```

I'm not sure how we want to do that tough, because the token currently stores its string as content. Maybe something like: 
* Store leading and trailing offset in header
* Change the content to `enum { Char(c), Trivia(trivia) }`

Accessing the str, leading or trailing trivia then simply is slicing the data and mapping it. 

Another alternative could be to not store a `HeaderSlice` but have multiple slices instead?

crates/rome_rowan/src/green/token.rs
```diff
@@ -13,10 +13,17 @@ use crate::{
...
+pub enum GreenTokenTrivia {
+	Whitespace,
+}
+
 #[derive(PartialEq, Eq, Hash)]
 struct GreenTokenHead {
 	kind: SyntaxKind,
 	_c: Count<GreenToken>,
+	leading_trivia: Vec<GreenTokenTrivia>,
+	trailing_trivia: Vec<GreenTokenTrivia>,
```
MichaReiser, line 26: We probably don't want to use `Vec`s to store the leading and trailing trivia, instead, we should store it inside of the `HeaderSlice` (as the slice)`. This avoids another heap allocation. 

```
type Repr = HeaderSlice<GreenTokenHead, [u8]>;
```

I'm not sure how we want to do that tough, because the token currently stores its string as content. Maybe something like: 
* Store leading and trailing offset in header
* Change the content to `enum { Char(c), Trivia(trivia) }`

Accessing the str, leading or trailing trivia then simply is slicing the data and mapping it. 

Another alternative could be to not store a `HeaderSlice` but have multiple slices instead?

crates/rome_rowan/src/green/builder.rs
```diff
@@ -32,16 +32,27 @@ impl GreenNodeBuilder<'_> {
...
+	) {
```
MichaReiser, line 43: Could be nice to have an additional overload to build a token without trivia... I sometimes use the builder inside tests.

crates/rome_rowan/src/green/token.rs
```diff
@@ -13,10 +13,17 @@ use crate::{
 	TextSize,
 };
 
+#[derive(PartialEq, Eq, Hash, Debug)]
+pub enum GreenTokenTrivia {
+	Whitespace,
```
MichaReiser, line 18: Do we need to store the count as well?

xunilrj: Yes, and another one for comments.

crates/rome_rowan/src/green/token.rs
xunilrj: > We can decide to follow up on the representation in a follow-up PR but we should avoid needing 3 heap allocations for each token.

Agreed!

crates/rome_rowan/src/green/builder.rs
ematipico: Would it make more sense to leave this fucntion as is and create a new one from scratch? e.g. `token_with_trivia`?

Do we need to mandatory pass an empty vector in case, for example, we don't have trailing trivia? Would it make more sense to have it as `Option<Vec<super::token::GreenTokenTrivia>>`?

crates/rslint_lexer/src/lib.rs
```diff
@@ -124,6 +124,17 @@ impl<'src> Lexer<'src> {
 		tok
 	}
 
+	fn should_break_whitespace_now(&self, byte: u8) -> bool {
+		if !is_linebreak(byte as char) {
+			false
+		} else {
+			!matches!(
+				(byte as char, self.bytes.get(self.cur + 1).copied()),
+				('\r', Some(next)) if (next as char) == '\n'
```
ematipico, line 133: I find this comparison difficult to read. What are we comparing here?

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -162,78 +131,63 @@ impl<'a> LosslessTreeSink<'a> {
...
+		// println!("LosslessTreeSink::do_token {:?} {:?}", kind, len);
 		let range = TextRange::at(self.text_pos, len);
 		let text = &self.text[range];
 		self.text_pos += len;
 		self.token_pos += 1;
-		self.inner.token(kind, text);
+
+		let trailing = self.get_trivia(true);
+		let leading = self.get_trivia(false);
+		let leading = std::mem::replace(&mut self.next_token_leading_trivia, leading);
```
ematipico, line 150: What exactly are we doing here?

crates/rome_rowan/src/green/token.rs
MichaReiser: It could make sense to split this PR into three smaller ones to make it easier to review 

* changes to the token type, maybe hashing as well and the tree builder. Demonstrate functionaliy with unit test 
* Lexer changes: not sure I understand what’s changing there
* Tree sink changes

crates/rome_rowan/src/green/builder.rs
MichaReiser: I personally try to avoid ‘Option<Vec>>’ . An empty vec has the same size and has no heap allocation.

the question is: would ‘Option<Vec>>’ have a semantically different meaning than an empty vec?

crates/rslint_lexer/src/lib.rs
xunilrj: I agree... but this is clippy suggestion; originally I used a match. I will try to improve it.

xunilrj: I agree... but this is clippy suggestion; originally I used a match. I will try to improve it.

crates/rslint_parser/src/lossless_tree_sink.rs
xunilrj: at line 148, "self.next_token_leading_trivia" has the leading trivia for the current token. I will start collecting the trailing trivia, and the leading for the next token.
The replace swaps the "current leading trivia" with the "new leading trivia".

Maybe it will be more clear if I process the current leading, and update the vec after this.

crates/rome_rowan/src/green/token.rs
xunilrj: Lexer changes were from the previous PR. I forgot to rebase before creating the PR. Done.

yassere (commented):
Sorry I didn't get to this sooner. I left some comments, but I don't think the memory-related concerns need to be addressed before this PR merges.

crates/rome_rowan/src/green/token.rs
yassere: >- Store leading and trailing offset in header
>- Change the content to enum { Char(c), Trivia(trivia) }

If the `trivia` here is a pointer to a single trivia token, then I don't think that would be viable. An enum is the size of its largest variant (plus the discriminant), so we probably can't afford to replace the `u8`s in `HeaderSlice` with enums.

If you mean that we could store two offsets that divide a single slice of `u8`s into leading/token/trailing, that seems possible.  But I think that would mean discarding any "TriviaKind" information (i.e. distinguishing between whitespace and comments) and having to determine that again later whenever we want to separate a single mixed `leadingTrivia` into individual trivia tokens.

crates/rome_rowan/src/green/token.rs
```diff
@@ -13,10 +13,88 @@ use crate::{
...
+		write!(f, "{:?}", self.0.slice())
+	}
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub enum GreenTokenTrivia {
+	None,
+	One(ThinTrivia),
+	Many(Vec<ThinTrivia>),
+}
```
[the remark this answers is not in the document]
MichaReiser: I agree with @yassere that we probably haven't found the right balance yet. I also spent some more time thinking about a possible representation for trivia and start to like Roslyn's approach of heal allocating trivia and caching them more and more:

* `TriviaCollection`: `ThinArc<TriviaCollectionHead, GreenTrivia>`. Heap allocated collection storing the leading or trailing trivia of a token. We cache the collections to reduce heap-allocations and memory usage. 
* `GreenTrivia`: `ThinArc<TriviaKind, u8>` (very similar to a token): Stores a single trivia with its string

I do like this approach because it finds a good balance between total memory consumption and the number of heap allocations. My assumption is that the leading/trailing trivia can be shared across many nodes. Therefore, caching the collections significantly reduces the number of heap allocations and total memory consumption, because the collection can be shared across a single file, even projects. 

Having two pointers increases the token size only by 2 words (16) bytes which is less than the cost of a vector. 

I'm not sure if it's worth wrapping the pointer in an option (but it probably doesn't matter because we don't pay any size overhead) because every token at least has a leading trivia and storing a pointer to a shared empty collection is little overhead. 

I also agree with @yassere that we don't need to make these changes in this PR but we should follow up on the data structure.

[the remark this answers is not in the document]
yassere: > every token at least has a leading trivia and storing a pointer to a shared empty collection is little overhead.

Storing a thin pointer to a shared empty collection means that you have to dereference the pointer to know that the collection is empty. You mentioned avoiding `Option<Vec>>`, but I think that's different. An owned vector can tell it's empty without any indirection because it stores its length independently from its slice pointer.

This might not actually matter in practice, but it could be worth taking advantage of null-pointer optimization to be able to identify the lack of leading/trailing trivia without following a pointer and without increasing mem size.

It's true that most tokens have leading trivia, but not always. The file `console.log("Hello World!");` contains 7 tokens all without any trivia.

xunilrj:
I created three other PR's that are more easy to understand. See #1720.

PR #1742, opened 2021-11-02
feat: unions of tokens in codegen

### LINKED ISSUE #1741: Implode alternatives of tokens
> Support inline token-union types. For example, the binary expression operator can be any of `==, +, -, ...`. That's why it's defined as `operator: ( '==' | '+' | '-' ...)`. The source gen should only generate a single field for the operator that returns a `SyntaxToken`

ematipico:
## Summary

This PR closes #1741 

During the code generation, we check if a `Rule::Alt` contains a vector of only `Rule::Token`. If so, we then collect all its tokens and store them.

When the moment of generating the code, check if `token_kinds` exist and we generate a different code.

## Test Plan

```
cargo lint
cargo format
cargo test
```


crates/rslint_parser/src/ast/generated/nodes.rs
```diff
@@ -499,65 +499,61 @@ pub struct UnaryExpr {
...
-	pub fn decrement_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T ! [--]) }
-	pub fn delete_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T![delete]) }
-	pub fn void_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T![void]) }
-	pub fn typeof_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T![typeof]) }
-	pub fn plus_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T ! [+]) }
-	pub fn minus_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T ! [-]) }
-	pub fn bitwise_not_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T ! [~]) }
-	pub fn excl_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T![!]) }
-	pub fn await_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T![await]) }
+	pub fn operation(&self) -> Option<SyntaxToken> {
```
MichaReiser, line 502: Can we call this `operator` to align with rome classic? Same for other fields.

crates/rslint_syntax/src/generated_new.rs
```diff
@@ -0,0 +1,561 @@
+//! Generated file, do not edit by hand, see `xtask/src/codegen`
+
```
MichaReiser, line 2: What's the meaning of the `new` postfix?

xtask/src/codegen/ast.rs
```diff
@@ -149,3 +158,43 @@ fn handle_rule(
...
+	}
+
+	let tokens: Vec<_> = rule
+		.iter()
+		.map(|rule| {
+			match rule {
+				Rule::Token(token) => clean_token_name(grammar, token),
+				// it should not go here, we checked if they all have token rules
+				_ => {
+					panic!("Something wrong happened. It seems there's some error in your grammar.")
```
MichaReiser, line 188: This isn't an issue with the grammar but more that we shouldn't have reached this line in the first place because `all_tokens` should then have been false. 

One way to avoid the double matching would be to rewrite this as a for each loop. 

```
for rule in rule.iter() {
  match rule {
    Rule::Token(token) => ... ,
    _ => return false 
    
  }
}
```

xtask/src/codegen/ast.rs
```diff
@@ -149,3 +158,43 @@ fn handle_rule(
...
+				}
+			}
+		})
+		.collect();
+
+	let field = Field::Token {
+		name: label.to_string(),
+		tokens: Some(tokens),
```
MichaReiser, line 196: I find it confusing that a token has some more tokens. We could change this to token_kinds if we still want to explicitly match on the token kind. We can otherwise replace this with a single flag: `union`

xtask/src/codegen/kinds_src.rs
```diff
@@ -386,16 +389,44 @@ impl Field {
...
-				let token: proc_macro2::TokenStream = token.parse().unwrap();
+			Field::Token { name, .. } => {
+				let token: proc_macro2::TokenStream = name.parse().unwrap();
 				Some(quote! { T![#token] })
 			}
 			_ => None,
 		}
 	}
+
+	pub fn extract_tokens(&self) -> Option<proc_macro2::TokenStream> {
```
MichaReiser, line 400: The name here isn't very clear. I had to read the source to understand what it's doing. 

Anyway, I think I'm in favour of just accepting any non-trivial token for now and later accepting any token. The parser should ensure that it doesn't generate crap ;)

crates/rslint_syntax/src/generated_new.rs
ematipico: Oh shoot, it shouldn't be here! Good catch 😅

crates/rslint_parser/src/ast/generated/nodes.rs
```diff
@@ -499,65 +499,61 @@ pub struct UnaryExpr {
...
+				T![delete],
+				T![void],
+				T![typeof],
+				T ! [+],
+				T ! [-],
+				T ! [~],
+				T![!],
+				T![await],
+			],
+		)
```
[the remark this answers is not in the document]
ematipico: Although we wouldn't want to get a token that it isn't supposed to be there, I think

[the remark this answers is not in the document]
MichaReiser: I think the point is that a token that isn't supposed to be there shouldn't be there in the first place. But I guess we need it for now until we have fixed slots to avoid returning the wrong token. But we may want to revisit if we need to have an explicit token check because the problem is that we otherwise skip a token that might be in it's place (which is an error too) .

xtask/src/codegen/ast.rs
```diff
@@ -90,6 +92,18 @@ fn handle_alternatives(grammar: &Grammar, rule: &Rule) -> Option<Vec<String>> {
 	}
 }
 
+fn clean_token_name(grammar: &Grammar, token: &Token) -> String {
+	let mut name = grammar[*token].name.clone();
+
+	// These tokens, when parsed to proc_macro2::TokenStream, generates a stream of bytes
+	// that can't be recognized by [quote].
+	// Hence, they need to be decorated with single quotes.
```
MichaReiser, line 100: Thanks!

xtask/src/codegen/kinds_src.rs
```diff
@@ -362,7 +362,10 @@ pub struct AstNodeSrc {
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum Field {
-	Token(String),
+	Token {
+		name: String,
+		token_kinds: Option<Vec<String>>,
```
MichaReiser, line 367: Using `Vec<String>` should be sufficient. You can test for `token_kinds.is_empty()` to see if it has any.

MichaReiser (approved):
Thanks for improving our code gen :) Nice to see how straightforward it is to handle new cases. 

I may re-visit the decision to test for an explicit token when working on fixed children because it may then no longer be required. I hope you're ok with this.

Issue #1737, opened 2021-11-01
Weekly goals - week 44


MichaReiser:
## Highlevel Goals

* Finish the work around the rowan data structure changes needed for attaching trivia to tokens and using fixed slots for children.
* Decision on lint-transformation API

## Work

- [ ] Publish a design proposal for CST based Lint transformation, agree on a proposal and communicate the decision
- [ ] Attach leading/trailing trivia to tokens #1720 
- [ ] Extend the AST codegen to support the features required by the new JS grammar #1725 
- [ ] Store children in fixed slots #1721 
- [ ] Explore new GitHub features for issue management
- [ ] Explore deployment options and features for crates
- [ ] Align crate licences 
- [ ] Use `rome_` naming convention for all our crates.  @mrkldshv

MichaReiser:
@mrkldshv are you still interested in working on the crate renaming? We've now agreed to go for `rome_` for now. Please don't feel pressured to do work on it before the end of this week.

mrkldshv:
@MichaReiser yeah, I'd like to work on renaming.

PR #1744, opened 2021-11-03
chore: rename rome crates

mrkldshv:
## Summary

Apply `rome_` prefix naming convention for all rome crates as agreed in #1737.

Renamed crates: 
* cli
* core
* formatter
* path

## Test Plan

```
cargo lint
cargo test
```


mrkldshv:
I wasn't really sure if we should rename `tests_macros` package. Would you suggest keeping it as it is?

ematipico (approved):
Looks good to me! Thank you!

ematipico:
> I wasn't really sure if we should rename `tests_macros` package. Would you suggest keeping it as it is?

I think we can skip it for now. Probably we could consider to move it somewhere else, later.

xunilrj (approved):

ematipico:
Thank you @mrkldshv !

mrkldshv:
@ematipico Happy to help! Would be glad to pick up something else.

PR #1745, opened 2021-11-03
feat: flat union of unions

### LINKED ISSUE #1743: Flatten unions of unions during the code generation
> It's convenient to define unions in respect of other unions. For example, `JsClassMemberName = JsObjectMemberName | JsPrivateClassMemberName` where `JsObjectMemberName` is an union type as well. This is convenient to maintain the grammar but makes the Facade more awkward to use because it requires two matches: first on the outer union and then on the inner union. We can avoid this if we flatten unions inside of the source gen and automatically generate `From<InnerUnion>` implementations.

ematipico:
## Summary

Closes #1743

We now generate a slight different code in case we encounter an enum that points to an enum.

This solution supports infinite number of enums and at any level. An example - **please check comments inside the snippet**:

```rust

impl From<SpreadElement> for ExprOrSpread {
	fn from(node: SpreadElement) -> ExprOrSpread { ExprOrSpread::SpreadElement(node) }
}
impl AstNode for ExprOrSpread {
	fn can_cast(kind: SyntaxKind) -> bool {
		match kind {
			SPREAD_ELEMENT => true,
			// here we fall back to use `ca_cast` and can contain other enums checks
			k if Expr::can_cast(k) => true,
			_ => false,
		}
	}
	fn cast(syntax: SyntaxNode) -> Option<Self> {
		let res = match syntax.kind() {
			SPREAD_ELEMENT => ExprOrSpread::SpreadElement(SpreadElement { syntax }),
			_ => {
				// this is where we fallback and put all the "enums of enums".
				// This means that we could have other enums without problems!
				if let Some(expr) = Expr::cast(syntax.clone()) {
					return Some(ExprOrSpread::Expr(expr));
				}
				return None;
			}
		};
		Some(res)
	}
	fn syntax(&self) -> &SyntaxNode {
		match self {
			ExprOrSpread::SpreadElement(it) => &it.syntax,
			ExprOrSpread::Expr(it) => it.syntax(),
		}
	}
}
```

## Test Plan

```
cargo test
cargo lint
cargo format
```

`cargo xtask coverage` on main

```
│   17604   │ 16765  │  838   │   1    │  95.23   │
```

`cargo xtask coverage` on this PR

```
```


MichaReiser:
I know it's not ready yet. But can you add `cargo xtask coverage` to your test plan. I believe we broke a few tests when we merged ungrammar before.. whoops

ematipico:
> I know it's not ready yet. But can you add `cargo xtask coverage` to your test plan. I believe we broke a few tests when we merged ungrammar before.. whoops

It might be better to have a separate PR we fix the coverage AND we add it to the CI. It seems it's not running! (I didn't know we had that)

MichaReiser:
It's not about increasing coverage. Just making sure we don't break anything new :)

ematipico:
@MichaReiser The `CI / Test262 coverage (pull_request) ` runs `cargo xtask coverage`. So I don't get what you're asking me to do?

MichaReiser:
Run the `cargo xtask coverage` once with and once without your changes and then verify if the numbers are still the same (or at least, not more panic / failing).

crates/formatter/src/ts/expr_or_block.rs
```diff
@@ -5,8 +5,35 @@ impl ToFormatElement for ExprOrBlock {
 	fn to_format_element(&self, formatter: &Formatter) -> Option<FormatElement> {
 		match self {
 			ExprOrBlock::BlockStmt(block) => block.to_format_element(formatter),
-			ExprOrBlock::Expr(expr) => expr.to_format_element(formatter),
 			ExprOrBlock::Literal(literal) => literal.to_format_element(formatter),
+			ExprOrBlock::ArrowExpr(node) => node.to_format_element(formatter),
+			ExprOrBlock::Template(_) => todo!(),
```
xunilrj, line 10: hum... I am wondering here if these variants that were flattened and don't have a suffix can be a source of confusing in the future.

ematipico: You hit the spot. Yes, it will be a source of confusion because what we have in the grammar will be different from what we generate.

I might have some other idea but I need to test it first.

ematipico:
Moving the PR to draft.

I believe I found a better to way to handle union of unions without remove them from the final code gen. I will update the PR once I achieved the solution.

ematipico:
PR updated with thew new code generation. Description has been updated with a brief explanation of what I've done.

MichaReiser:
Can you add an example on how this API would be used e.g. when matching on an union containing another union. Do users have to first match on the outer and then on the inner union? 

Do we have many nested unions, especially ones that contain more than two sub unions? I'm trying to understand what the runtime cost of the trying out each sub-union is.

MichaReiser:
> Can you add an example on how this API would be used e.g. when matching on an union containing another union. Do users have to first match on the outer and then on the inner union? 
> 
> Do we have many nested unions, especially ones that contain more than two sub unions? I'm trying to understand what the runtime cost of the trying out each sub-union is.

OK, the performance shouldn't be a problem. We can (in the future or in a follow up PR) unroll the match of subunions in 'cast' and 'can_cast'. 

Something that might be worthwhile adding is a check that no two sub-unions are overlapping.

ematipico:
> Something that might be worthwhile adding is a check that no two sub-unions are overlapping.

I was thinking to have it as a task after this PR is merged and make it and "first good issue". It's an optimization and it's small enough.

ematipico:
> Can you add an example on how this API would be used e.g. when matching on an union containing another union. Do users have to first match on the outer and then on the inner union?

The API is still the same. I basically took the manual approach used by `rslint_parser` and made more generic, so it can be auto generated without restrictions.

> Do we have many nested unions, especially ones that contain more than two sub unions? I'm trying to understand what the runtime cost of the trying out each sub-union is.

Like one or two, so it's not that bad. I think those are more exceptions to the rule and I am pretty sure that won't be many in the future, even we will extend a language or create a new one.

MichaReiser:
> > Something that might be worthwhile adding is a check that no two sub-unions are overlapping.
> 
> 
> 
> I was thinking to have it as a task after this PR is merged and make it and "first good issue". It's an optimization and it's small enough. 

That's a great idea. We could do the same for unrolling the cast and can cast methods

crates/rome_formatter/src/ts/expr_or_spread.rs
```diff
@@ -6,13 +6,6 @@ impl ToFormatElement for ExprOrSpread {
 		match self {
 			ExprOrSpread::SpreadElement(spread) => spread.to_format_element(formatter),
 			ExprOrSpread::Expr(expr) => expr.to_format_element(formatter),
-			ExprOrSpread::Literal(literal) => literal.to_format_element(formatter),
-			ExprOrSpread::ObjectExpr(object_expression) => {
-				object_expression.to_format_element(formatter)
-			}
-			ExprOrSpread::ArrayExpr(array_expression) => {
-				array_expression.to_format_element(formatter)
-			}
```
MichaReiser, line 15: 🥳

xtask/js.ungram
```diff
@@ -325,9 +326,9 @@ ExprOrSpread =
 	Expr
 	| SpreadElement
 	// @ematipico these have been added because the current enums are not flatten
-	| Literal
-	| ObjectExpr
-	| ArrayExpr
+	// | Literal
+	// | ObjectExpr
+	// | ArrayExpr
```
MichaReiser, line 331: Delete?

xtask/src/codegen/generate_nodes.rs
```diff
@@ -100,6 +100,7 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
 		.enums
 		.iter()
 		.map(|en| {
+			// here we collect all the variants, regardless
```
MichaReiser, line 103: regardless of what?

xtask/src/codegen/generate_nodes.rs
```diff
@@ -112,8 +113,20 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
...
-				.variants
+			// here we make the partition
+			// inside an enum, we can have variants that point to a "flat" type or to another enum
+			// we want to divide these variants as we will generate a different code based on these requirements
+			let (variant_of_variants, simple_variants): (Vec<_>, Vec<_>) =
+				en.variants.iter().partition(|current_enum| {
+					let this_variant_has_variants = ast
+						.enums
+						.iter()
+						.any(|v| v.name.eq(*current_enum) && !v.variants.is_empty());
```
MichaReiser, line 124: Nit: Unlikely an issue since this is a codegen but you're re-iterating the variants collection for every variant to do a name lookup. Would it make sense to build a map of name -> variant once and use that instead of using `any`?

MichaReiser (approved):

xtask/src/codegen/generate_nodes.rs
ematipico: Thank you! I knew I could improve it but I couldn't see how :D

MichaReiser:
@ematipico would you mind creating the two follow up issues discussed in this PR?

ematipico:
> @ematipico would you mind creating the two follow up issues discussed in this PR?

Created 

#1757 
#1756

PR #1746, opened 2021-11-03
AST Separated List

MichaReiser:
## Summary

Implements an AST view for lists that contain nodes separated by tokens, for example an array expression where the elements are separated by commas. 

The PR introduces the new `AstSeparatedList`, changes the code gen to recognise the `T (',' T)* ','?` and `T (',' T)*` patterns, and uses the new separated list in the js grammar. 

The PR doesn't  use the new list yet for:
* arrays because this would require defining a new `JsArrayHole` type as well. 
* Import clauses, because not all of them must be separated by a ','

We can address these issues when changing our Ast Facade.

Part of #1725 

## Test Plan

Passing / failing / panics stay the same

```
│   17608   │ 16767  │  840   │   1    │  95.22   │
```

## Review

It's probably easiest to review the commits for now until the `AstList` PR is merged.

### DIFF
--- a/crates/rslint_parser/src/ast.rs
+++ b/crates/rslint_parser/src/ast.rs
@@ -603,8 +603,7 @@ mod tests {
 		// list([,3])
 		let list = build_list(vec![(None, Some(",")), (Some(3), None)]);
 
-		// This should panic because having two successive separators is invalid.
-		// Grammars should instead model a "hole" node if this is a valid language construct.
+		// This should panic because the first element is a separator instead of a node.
 		let _ = list.elements().collect::<Vec<_>>();
 	}
 
--- a/crates/rslint_parser/src/ast.rs
+++ b/crates/rslint_parser/src/ast.rs
@@ -185,10 +185,171 @@ impl<N: AstNode> IntoIterator for AstNodeList<N> {
 	}
 }
 
+#[derive(Debug, Clone)]
+pub struct AstSeparatedElement<N> {
+	node: N,
+	trailing_separator: Option<SyntaxToken>,
+}
+
+#[derive(Debug, Clone)]
+pub struct AstSeparatedList<N> {
+	list: SyntaxList,
+	ph: PhantomData<N>,
+}
+
+impl<N: AstNode> AstSeparatedList<N> {
+	fn new(list: SyntaxList) -> Self {
+		Self {
+			list,
+			ph: PhantomData,
+		}
+	}
+
+	/// Returns an iterator over all nodes with their trailing separator
+	pub fn elements(&self) -> AstSeparatedListElementsIterator<N> {
+		AstSeparatedListElementsIterator {
+			next: self.list.iter(),
+			ph: PhantomData,
+		}
+	}
+
+	/// Returns an iterator over all separator tokens
+	pub fn separators(&self) -> impl Iterator<Item = SyntaxToken> {
+		self.elements()
+			.flat_map(|element| element.trailing_separator)
+	}
+
+	/// Returns an iterator over all nodes
+	pub fn iter(&self) -> AstSeparatedListNodesIterator<N> {
+		AstSeparatedListNodesIterator {
+			inner: self.elements(),
+		}
+	}
+
+	#[inline]
+	pub fn is_empty(&self) -> bool {
+		self.len() == 0
+	}
+
+	pub fn len(&self) -> usize {
+		// TODO 1724 replace with (self.list.len() + 1) << 2 once trivia are attached to tokens
+		self.iter().count()
+	}
+
+	pub fn trailing_separator(&self) -> Option<SyntaxToken> {
+		// TODO 1724: Replace with simple match once trivia is no longer stored in the body
+		// match self.list.last() {
+		// 	NodeOrToken::Token(token) => Some(token),
+		// 	_ => None
+		// }
+
+		self.elements().last()?.trailing_separator
+	}
+}
+
+#[derive(Debug, Clone)]
+pub struct AstSeparatedListElementsIterator<N> {
+	next: SyntaxElementChildren,
+	ph: PhantomData<N>,
+}
+
+impl<N> AstSeparatedListElementsIterator<N> {
+	// TODO 1724: Replace with call to next once trivia are no longer stored in tokens.
+	fn next_non_trivia_or_error(&mut self) -> Option<SyntaxElement> {
+		self.next.find_map(|element| match &element {
+			NodeOrToken::Node(node) => {
+				if node.kind() == SyntaxKind::ERROR {
+					None
+				} else {
+					Some(element)
+				}
+			}
+			NodeOrToken::Token(token) => {
+				if token.kind().is_trivia() {
+					None
+				} else {
+					Some(element)
+				}
+			}
+		})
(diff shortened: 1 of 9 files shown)


crates/rslint_parser/src/ast.rs
```diff
@@ -185,9 +185,163 @@ impl<N: AstNode> IntoIterator for AstNodeList<N> {
...
+}
+
+impl<N: AstNode> Iterator for AstSeparatedListElementsIterator<N> {
+	type Item = AstSeparatedElement<N>;
+
+	fn next(&mut self) -> Option<Self::Item> {
+		let node_or_token = self.next_non_trivia_or_error()?;
+
+		let element = match node_or_token {
+			// The node for this element is missing if there's right a separator token
```
ematipico, line 287: "if there's right a separator token" doesn't sound right

crates/rslint_parser/src/ast.rs
```diff
@@ -185,9 +185,163 @@ impl<N: AstNode> IntoIterator for AstNodeList<N> {
...
+			NodeOrToken::Token(token) => panic!(
```
ematipico, line 288: Is using `panic!` the right choice? If so, should we leave a comment? 
I would expect a diagnostic of something else to handle gracefully the exception.

crates/rslint_parser/src/ast.rs
```diff
@@ -232,3 +393,149 @@ mod support {
...
+		);
+		assert!(list.trailing_separator().is_some());
+	}
+
+	#[test]
+	#[should_panic(
+		expected = "Missing element in separated list, found COMMA@2..3 \",\" token instead"
+	)]
+	fn separated_with_two_successive_separators() {
+		// list([1,,])
```
ematipico, line 522: Should we test `[,1]` and `[]` too? First, a list that starts with a separators, second an empty list

xtask/js.ungram
```diff
@@ -6,15 +6,16 @@
...
-//   label:A     -- suggested name for field of AST node
+//   //          				-- comment
+//   Name =      				-- non-terminal definition
+//   'ident'     				-- token (terminal)
+//   A B         				-- sequence
+//   A | B       				-- alternation
+//   A*          				-- zero or more repetition
+//   (A (',' A)* ','?)	-- repetition of node A separated by ',' (allowing a trailing comma)
+//   A?          				-- zero or one repetition
+//   (A)         				-- same as A
```
ematipico, line 17: Comment aligned
```suggestion
//   A*          				-- zero or more repetition
//   (A (',' A)* ','?)			-- repetition of node A separated by ',' (allowing a trailing comma)
//   A?          				-- zero or one repetition
//   (A)         				-- same as A
```

xtask/src/codegen/ast.rs
```diff
@@ -156,12 +157,81 @@ fn handle_rule(
...
 		}
-		Rule::Seq(rules) | Rule::Alt(rules) => {
+		Rule::Alt(rules) => {
+			for rule in rules {
+				handle_rule(fields, grammar, rule, label, false, false);
+			}
+		}
+
+		Rule::Seq(rules) => {
+			if lower_comma_list(fields, grammar, label, rules.as_slice()) {
```
ematipico, line 167: `handle_comma_list`?

crates/rslint_parser/src/ast.rs
MichaReiser: This is one of the consequences of the new AST facade. Our new AST facade gives new guarantees that will require panics/asserts to satisfy Rust's type checker. For example, we define that the 3rd element of an `if` statement will always be the condition. That means, we'll need to *cast* to a condition and *unwrap* the node. 

The parser should never generate a parse tree that validates this assumption. It's a parser bug if it does. 

That's why working on the syntax level is highly discouraged and our mutation API should ideally only work on the AST level, so that invalidating the tree never is possible (and, therefore, casting is safe).

yassere: > That's why working on the syntax level is highly discouraged and our mutation API should ideally only work on the AST level, so that invalidating the tree never is possible (and, therefore, casting is safe).

This is something we may need to put more thought into. We probably want something like Swift's make/with/builder APIs that work at the AST level, and that should be the preferred way of working with syntax nodes on a small scale.

But I think it's very useful to have some methods like those in Roslyn's [SyntaxNodeExtensions](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxnodeextensions?view=roslyn-dotnet-3.11.0) (ReplaceNodes especially) and I'm not sure where exactly we fit tree validation into that.

I think we should try to avoid panics even if transforms misbehave, so if we need to panic on certain invalid states, we'd want to prevent transforms from being able to create those states. So that would mean going a step further than just discouraging certain APIs and we should consider which ones we'd have to rule out entirely if we can't otherwise avoid these sorts of panics.

MichaReiser: > This is something we may need to put more thought into. We probably want something like Swift's make/with/builder APIs that work at the AST level, and that should be the preferred way of working with syntax nodes on a small scale

Certainly, but that's something I consider out of scope for the parser work. Also if it is indeed limited to the small scale. 


> But I think it's very useful to have some methods like those in Roslyn's SyntaxNodeExtensions (ReplaceNodes especially) and I'm not sure where exactly we fit tree validation into that.

These may be useful but I still think that most refactorings/lint autofixes should have no need for these extensions. The main use case that I see for working with `SyntaxNode` directly is if you want to implement something generically, regardless of the node type. 

> I think we should try to avoid panics even if transforms misbehave, so if we need to panic on certain invalid states, we'd want to prevent transforms from being able to create those states. So that would mean going a step further than just discouraging certain APIs and we should consider which ones we'd have to rule out entirely if we can't otherwise avoid these sorts of panics.

I'm not sure if we can make it impossible but agree, we should make it as hard as possible for transforms to generate invalid source code. 

Casting here just reveals an error in the tree after a transform (or parser bug) that would result in a syntax error when emitting the file. Adding checks to avoid emitting files with syntax errors will also prevent that the cast here can fail.

crates/rome_formatter/src/ts/mod.rs
```diff
@@ -28,6 +28,7 @@ mod test {
 	fn arrow_function() {
 		let src = "let v = (value  , second_value) =>    true";
 		let tree = parse_text(src, 0);
+		println!("{:#?}", tree.syntax());
```
ematipico, line 31: To delete

crates/rslint_parser/src/ast.rs
```diff
@@ -232,3 +400,175 @@ mod support {
...
+	#[test]
+	#[should_panic(
+		expected = "Missing element in separated list, found COMMA@0..1 \",\" token instead"
+	)]
+	fn separated_with_leading_separator() {
+		// list([,3])
+		let list = build_list(vec![(None, Some(",")), (Some(3), None)]);
+
+		// This should panic because having two successive separators is invalid.
+		// Grammars should instead model a "hole" node if this is a valid language construct.
```
ematipico, line 559: I believe the comment is not correct for this particular test case

ematipico (approved):

Issue #1732, opened 2021-10-29
Change the codegen to respect a child's optionality and generate the appropriate code, fix consuming code


ematipico:

MichaReiser:
Implemented as part of #1749

PR #1749, opened 2021-11-04
Feature/syntax result

### LINKED ISSUE #1731: Introduce `SyntaxResult`

### LINKED ISSUE #1740: Change Formatter to return `Result<FormatElement, FormatError>` instead of an `Option`. Prevents accidentally returning when skipping an optional element. `enum FormatError { MissingNode, UnknownNode }` . Implement `From<SyntaxError> for FormatError` so that the try operator works nicely.

ematipico:
## Summary

Closes #1731
Closes #1740 

This PR applies the following changes:
- it extends the logic of the code generation by tracking nodes that are marked as optional. The nodes/tokens that are marked as optional will emit code as `Optional<Node>`/`Optional<SyntaxToken>`. Nodes and tokens that are not marked as optional, will now generated code as `SyntaxResult<Node>`/ `SyntaxResult<SyntaxToken>`;
- created a new type called `SyntaxResult`;
- updated the formatter APIs to handle `Result` instead of `Option`;
- modified the manual implementation of some nodes; there, I left some `TODO` where we know that we will need to remove/change some logic. Some of that logic was kept there to maintain compatibility. Once we will start migrating to a better syntax, we will change (or remove) that logic too;
- updated doc tests where it was needed;

## Test Plan

```
cargo test
cargo format
cargo lint
```


crates/formatter/src/formatter.rs
```diff
@@ -28,16 +31,17 @@ impl Formatter {
 	pub fn format_root(self, root: &SyntaxNode) -> FormatResult {
 		let element = self
 			.format_syntax_node(root)
-			.unwrap_or_else(|| self.format_raw(root));
+			// TODO: we have an error here. Diagnostic?
```
jamiebuilds, line 34: Should the formatter ever be responsible for reporting syntax errors? I assume the parser would still emit diagnostics for every syntax error even if we recover and run other tools like the formatter.

Maybe this would be the difference between logging levels though, if you passed a `--log-level=verbose` it would also emit these kinds of events from other tools so we can debug why the formatter didn't format what we expected.

ematipico: Yes it makes sense. Although executing an unwrap without a comment or way to handle it doesn't seem right, hence the `TODO` :)

MichaReiser: I think we should either return an error or return the text as is

crates/rome_formatter/src/cst.rs
```diff
@@ -10,7 +10,7 @@ use rslint_parser::ast::{
 use rslint_parser::{AstNode, AstToken, SyntaxKind, SyntaxNode, SyntaxToken};
 
 impl ToFormatElement for SyntaxNode {
-	fn to_format_element(&self, formatter: &Formatter) -> Option<FormatElement> {
+	fn to_format_element(&self, formatter: &Formatter) -> Result<FormatElement, FormatError> {
```
MichaReiser, line 13: Can we define a `FormatResult` type that is an alias of `Result<T, FormatError>`? 

Yes, the name is clashing with `FormatResult` but I think we should rename `FormatResult` because the `Result` is misleading, reader might think it's a `Result` type. How about `format_root` returns a `FormatResult<Formatted>`?

crates/rome_formatter/src/formatter.rs
```diff
@@ -28,16 +31,17 @@ impl Formatter {
 	pub fn format_root(self, root: &SyntaxNode) -> FormatResult {
 		let element = self
 			.format_syntax_node(root)
-			.unwrap_or_else(|| self.format_raw(root));
+			// TODO: we have an error here. Diagnostic?
+			.unwrap_or_else(|_| self.format_raw(root));
```
MichaReiser, line 35: We can bail out if we change `format` to return an error too (see previous comment).

crates/rome_formatter/src/lib.rs
```diff
@@ -76,7 +78,23 @@ use std::str::FromStr;
...
-	fn to_format_element(&self, formatter: &Formatter) -> Option<FormatElement>;
+	fn to_format_element(&self, formatter: &Formatter) -> Result<FormatElement, FormatError>;
+}
+
+#[derive(Debug, PartialEq)]
+/// Series of errors encountered during formatting
+pub enum FormatError {
+	/// Node is missing and it should be required for a correct formatting
+	MissingNode(SyntaxKind),
+	UnknownNode,
```
MichaReiser, line 89: I'm not sure yet if we need the `UnknownNode` case. We can probably as well just call `format_verbatim` for unknown nodes? We should probably experiment and see how well the formatter plays with unknown nodes and introduce it if the results aren't satisfying.

crates/rome_formatter/src/ts/class/prop.rs
```diff
@@ -28,14 +29,15 @@ impl ToFormatElement for ClassProp {
...
+				Ok(format_elements![
 					static_token,
 					key,
 					equal_and_value,
 					token(";")
 				])
 			}
-			None => None,
+			// TODO: review, we can have a better error
+			None => Err(FormatError::UnknownNode),
```
MichaReiser, line 40: Isn't this a `MissingNode` error? Either the `equal` or `value` were missing. You can get the right kind by extending the match on line 17 to handle the `Some(_), None` and `None, Some(_)` cases to return a `Missing` error. This will be nicer once we improved the AST Facade.

crates/rome_formatter/src/ts/declarators/var_decl.rs
```diff
@@ -1,18 +1,21 @@
...
+		if self.is_const() {
+			tokens.push(formatter.format_token(&self.const_token()?)?);
+		} else if self.is_var() {
+			tokens.push(formatter.format_token(&self.var_token()?)?);
+		} else if self.is_let() {
+			// TODO: #1725 remove this custom code once #1745 is merged
+			tokens.push(formatter.format_token(&self.let_token().unwrap())?);
 		} else {
-			return None;
+			return Err(FormatError::MissingNode(self.syntax().kind()));
```
MichaReiser, line 18: The var decl kind isn't the missing element. We miss a `const`, `var` or `let` token. 

This use case reveals a problem when we pass in the expected kind and we'll have the same problem with all other places where we expect a union type (e.g. a `Expr`, there's no single expected kind). 

I think it's best if we remove the `SyntaxKind` parameter for missing in that case because it's just misleading.

crates/rome_formatter/src/lib.rs
```diff
@@ -76,7 +78,23 @@ use std::str::FromStr;
...
 pub trait ToFormatElement {
-	fn to_format_element(&self, formatter: &Formatter) -> Option<FormatElement>;
+	fn to_format_element(&self, formatter: &Formatter) -> Result<FormatElement, FormatError>;
+}
+
+#[derive(Debug, PartialEq)]
+/// Series of errors encountered during formatting
+pub enum FormatError {
+	/// Node is missing and it should be required for a correct formatting
+	MissingNode(SyntaxKind),
```
MichaReiser, line 88:
```suggestion
	MissingElement(SyntaxKind),
```

It can be either a token or node that is missing. 

There's also a problem with passing in the missing kind in case the missing child is a union, e.g. an expression. There isn't a single missing kind in that case. 

We should, therefore, either remove the parameter or re-define the error. An alternative that I see is to change it to `MissingRequiredChild { parent: SyntaxNode }`. We may be able to add the "slot" of the missing child in the future so that it is clear which child was missing.

crates/rome_formatter/src/ts/patterns/assign_pattern.rs
```diff
@@ -1,14 +1,16 @@
...
+		let key = if let Ok(key) = self.key() {
 			formatter.format_node(key)?
 		} else {
 			empty_element()
 		};
-		let assign = if let Some(eq_token) = self.eq_token() {
+		let assign = if let Ok(eq_token) = self.eq_token() {
 			formatter.format_token(&eq_token)?
 		} else if let Some(colon_token) = self.colon_token() {
 			formatter.format_token(&colon_token)?
```
MichaReiser, line 16: There are probably more cases but we shouldn't format a missing token. We should just bail out if a mandatory token is missing. This is true in all formatting code (I only just realised this now, there are probably more such examples in other files).

Or is this not a mandatory token in which case the grammar is incorrect? It might then be worth adding a TODO (pointing to #1725) or explain in the PR summary that you intentionally didn't change the grammar.

```suggestion
		let assign = formatter.format_token(&eq_token?)?
```

crates/rome_formatter/src/ts/statements/try_statement.rs
```diff
@@ -18,9 +22,9 @@ impl ToFormatElement for TryStmt {
 			empty_element()
 		};
 		if handler.is_empty() && finalizer.is_empty() {
-			None
+			Err(FormatError::MissingNode(self.syntax().kind()))
```
MichaReiser, line 25: The missing node is the catch handler or the finalizer and not the try statement. We should probably just remove the "expected" syntax kind parameter of `MissingNode`

crates/rslint_parser/src/ast.rs
```diff
@@ -102,6 +109,31 @@ mod support {
...
+			.children_with_tokens()
+			.filter_map(|it| it.into_token())
+			.find(|it| it.kind() == kind)
+	}
+
+	pub(super) fn as_mandatory_node<N: AstNode>(parent: &SyntaxNode) -> SyntaxResult<N> {
+		parent
+			.children()
+			.find_map(N::cast)
+			.ok_or_else(|| SyntaxError::MissingElement(parent.kind()))
```
MichaReiser, line 123: Shouldn't it be the missing element's kind rather than what the parent was?

crates/rslint_parser/src/ast/expr_ext.rs
```diff
@@ -28,10 +28,12 @@ impl CondExpr {
...
 
 impl LiteralProp {
-	pub fn key(&self) -> Option<PropName> {
+	pub fn key(&self) -> SyntaxResult<PropName> {
 		if PropName::can_cast(
 			support::children::<PropName>(self.syntax())
-				.next()?
+				.next()
+				// TODO: review here as it's not correct
+				.unwrap()
```
MichaReiser, line 36: Can't we keep `next()?` which would give the same behaviour as before?

xtask/js.ungram
```diff
@@ -537,7 +537,7 @@ FnDecl =
 	return_type:TsType?
 	body:BlockStmt
 
-VarDecl = 'var' 'const' manual__:'let' declared:Declarator* ';'?
+VarDecl = ('var' | 'const' | manual__:'let') declared:Declarator* ';'?
```
MichaReiser, line 540: Do we need a label here? Or what name picks the code generator for the `kind` operator?

xtask/src/codegen/ast.rs
```diff
@@ -187,6 +188,8 @@ fn handle_tokens_in_unions(
 	let field = Field::Token {
 		name: label.to_string(),
 		token_kinds,
+		// TODO: review this, it's not correct
+		optional: false,
```
MichaReiser, line 192: This should be fine because the token shouldn't be optional (you return `false` in that case on line 184). Or should the optional be passed in on line 124

MichaReiser (changes requested):
Nice work!

Your code made it clear to me that there's an issue with the kind parameter of `MissingElement`. There's no obvious choice for the missing kind if the child is an union, e.g. the binary expression operator (different tokens) or if it's an expression. 

I think we should either drop the param or change the error to `MissingRequiredChild { parent: SyntaxNode }`

crates/rome_formatter/src/cst.rs
ematipico: Sure, thank you. Actually at first I created a type and it was clashing... As I didn't have a better naming in mind, I left as is and continued the refactor.

crates/rome_formatter/src/lib.rs
ematipico: Makes sense. After I played with the refactor, I get  the sense that we won't have any errors other than `MissingNode`.

crates/rome_formatter/src/ts/class/prop.rs
ematipico: Unfortunately it's impossible to get the `.kind()` of `equal` and `value` unless we actually decide to hardcode it inside the error `MissingNode`.

Also, what if `.value()` returned a `UnknownExpression`? Wouldn't it count as `UnknownNode`? That's the part where I am struggling at the moment, understanding and handling missing nodes VS unknown nodes inside the formatter.

crates/rslint_parser/src/ast/expr_ext.rs
ematipico: We can't because `.next` returns a `Option` while the function return type is `SyntaxResult` (which is a `Result<T>` under the hood). One possible to solution would be to transform `Option<T>` to `Result<T, E>` using [`ok_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or)?

xtask/js.ungram
ematipico: We do, we still have some manual implementation called `let_token`.

xtask/src/codegen/ast.rs
ematipico: I **think** it should be passed on line 124.

I was worried about cases like `op: ('+' | '-')?`. I think it should be safe passing it.

crates/rome_formatter/src/ts/class/prop.rs
MichaReiser: > Also, what if .value() returned a UnknownExpression? Wouldn't it count as UnknownNode? That's the part where I am struggling at the moment, understanding and handling missing nodes VS unknown nodes inside the formatted.

Maybe, depends on how we want to do the formatting for `Unkown` nodes:
* If we want to format the nodes as is, then implementing `ToFormatElement` for `UnknownExpression` is enough. 
* If we want to bail out of formatting, then having an `UnknownNode` error is the way to go and the `Unknown*` implementation would just bail out with that error.

ematipico:
> `MissingRequiredChild`

So I tried to implement `MissingRequiredChild { parent: SyntaxNode }` and this would required to `.clone()` the `AstNode` every time we fire an error; this is because the generic called `.syntax()` would return to us a reference - `&SyntaxNode`. Not sure if it's worth it...

Adding a `{ parent: &SyntaxNode}` instead, would involve introducing lifetimes, inside errors. Not sure we need that.

MichaReiser:
> > `MissingRequiredChild`
> 
> So I tried to implement `MissingRequiredChild { parent: SyntaxNode }` and this would required to `.clone()` the `AstNode` every time we fire an error; this is because the generic called `.syntax()` would return to us a reference - `&SyntaxNode`. Not sure if it's worth it...
> 
> Adding a `{ parent: &SyntaxNode}` instead, would involve introducing lifetimes, inside errors. Not sure we need that.

Cloning is cheap but yeah, we don't have a use case for the param yet. Should we go with dropping the kind param for now and then add whatever we need for our tracking purposes when we need it?

crates/rome_formatter/src/lib.rs
```diff
@@ -71,12 +72,40 @@ use rome_core::file_handlers::Language;
...
+pub type FormatResult<F> = Result<F, FormatError>;
+
+#[derive(Debug, PartialEq)]
+/// Series of errors encountered during formatting
+pub enum FormatError {
+	/// Node is missing and it should be required for a correct formatting
+	MissingRequiredChild,
+
+	/// Cases where the parser can't figure out the type of node.
+	UnknownNode,
```
MichaReiser, line 94: The comment here doesn't match how you are using it in the code. I would recommend deleting it for now and use a `MissingRequiredChild` in the place where you used `UnknownNode` until we figured out what to do about `Unknown*` nodes.

crates/rome_formatter/src/lib.rs
```diff
@@ -71,12 +72,40 @@ use rome_core::file_handlers::Language;
...
+
+	/// In case our formatter doesn't know how to format a certain language
+	UnsupportedLanguage,
+
+	/// When the ability to format the current file has been turned off on purpose
+	CapabilityDisabled,
```
MichaReiser, line 100: Is it worth differentiating between `UnsupportedLanguage` and `CapabilityDisabled`? Isn't the `capability` disabled for the case where our formatter doesn't support formatting (`UnsupportedLanguage`?).

crates/rome_formatter/src/lib.rs
```diff
@@ -71,12 +72,40 @@ use rome_core::file_handlers::Language;
...
+}
+
+impl From<SyntaxError> for FormatError {
+	fn from(syntax_error: SyntaxError) -> Self {
+		match syntax_error {
+			SyntaxError::MissingElement(_node) => FormatError::MissingRequiredChild,
```
MichaReiser, line 106: Should we use the same names/structure for the errors?

MichaReiser (approved):

crates/rome_formatter/src/lib.rs
ematipico: Yes it is worth it. Supporting a language and being able to format it should be two separate things. For example, in RomeJS we had support for parsing CSS but the format was turned off publicly. In the meantime, we could programmatically turn it on and test it, incrementally, until we were happy with the final result.

MichaReiser: That's fair. Is it something Formater consumers should be aware of tough? For them, it's an unsupported language. 

Please go forward with what you think best.

ematipico: Not necessarily, we might have linting enabled while formatter still disabled (for the same language). If things changes, we are still in time for unifying everythind and having the broad "unsupported" error.

MichaReiser:
closes #1732

Issue #1733, opened 2021-10-29
Strip `Js` from codegen and enforce usage of labels in codegen


ematipico:
> Change the codegen to strip the `Js` extensions from field names to reduce the need for explicit labels. For example, `ExpressionStatement` must use an explicit label for its inner expression so that it isn't named `js_expression`.

MichaReiser:
What a title 😄

ematipico:
Yeah I used the new github features that automatically creates tasks out of a checkbox list.

PR #1750, opened 2021-11-05
feat: strip prefix from methods in code generation

ematipico:
## Summary

 Part of https://github.com/rome/tools/issues/1725

Closes https://github.com/rome/tools/issues/1733

This PR improves the codegen by forcing to use labels when writing ungrammar.

## Test Plan

```
cargo test
cargo check
cargo lint
```

### DIFF
--- a/crates/rslint_parser/src/ast/generated/nodes.rs
+++ b/crates/rslint_parser/src/ast/generated/nodes.rs
@@ -1410,7 +1410,7 @@ impl TsTypeAliasDecl {
 	pub fn eq_token(&self) -> SyntaxResult<SyntaxToken> {
 		support::as_mandatory_token(&self.syntax, T ! [=])
 	}
-	pub fn ts_type(&self) -> SyntaxResult<TsType> { support::as_mandatory_node(&self.syntax) }
+	pub fn ty(&self) -> SyntaxResult<TsType> { support::as_mandatory_node(&self.syntax) }
 }
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct TsNamespaceDecl {
@@ -1424,9 +1424,7 @@ impl TsNamespaceDecl {
 	pub fn dot_token(&self) -> Option<SyntaxToken> {
 		support::as_optional_token(&self.syntax, T ! [.])
 	}
-	pub fn ts_namespace_body(&self) -> SyntaxResult<TsNamespaceBody> {
-		support::as_mandatory_node(&self.syntax)
-	}
+	pub fn body(&self) -> SyntaxResult<TsNamespaceBody> { support::as_mandatory_node(&self.syntax) }
 }
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct TsModuleDecl {
--- a/xtask/src/codegen/kinds_src.rs
+++ b/xtask/src/codegen/kinds_src.rs
@@ -375,6 +375,8 @@ pub enum Field {
 		name: String,
 		token_kinds: Vec<String>,
 		optional: bool,
+		/// if the name was generated from a label
+		name_from_label: bool,
 	},
 	Node {
 		name: String,
@@ -503,7 +505,15 @@ impl Field {
 				};
 				format_ident!("{}_token", name)
 			}
-			Field::Node { name, .. } => {
+			Field::Node {
+				name,
+				name_from_label,
+				ty,
+				..
+			} => {
+				if !name_from_label {
+					panic!("The node {} doesn't have a label", ty);
+				}
 				if name == "type" {
 					format_ident!("ty")
 				} else {
--- a/xtask/src/codegen/kinds_src.rs
+++ b/xtask/src/codegen/kinds_src.rs
@@ -3,6 +3,8 @@
 
 use quote::{format_ident, quote};
 
+const LANGUAGE_PREFIXES: [&str; 4] = ["js_", "ts_", "jsx_", "tsx_"];
+
 pub struct KindsSrc<'a> {
 	pub punct: &'a [(&'a str, &'a str)],
 	pub keywords: &'a [&'a str],
@@ -514,10 +516,14 @@ impl Field {
 				if !name_from_label {
 					panic!("The node {} doesn't have a label", ty);
 				}
-				if name == "type" {
+				let mut final_name = name.clone();
+				for prefix in LANGUAGE_PREFIXES {
+					final_name = final_name.replace(prefix, "");
+				}
+				if final_name == "type" {
 					format_ident!("ty")
 				} else {
-					format_ident!("{}", name)
+					format_ident!("{}", final_name)
 				}
 			}
 		}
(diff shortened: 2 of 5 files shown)


crates/rslint_parser/src/ast/generated/nodes.rs
```diff
@@ -265,7 +265,7 @@ impl TsTypeAliasDecl {
 	pub fn type_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T![type]) }
 	pub fn type_params(&self) -> Option<TsTypeParams> { support::child(&self.syntax) }
 	pub fn eq_token(&self) -> Option<SyntaxToken> { support::token(&self.syntax, T ! [=]) }
-	pub fn ts_type(&self) -> Option<TsType> { support::child(&self.syntax) }
+	pub fn ty(&self) -> Option<TsType> { support::child(&self.syntax) }
```
xunilrj, line 268: I hate that we can't have "type" as identifiers in Rust. There is one option: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=528a1f56f1209a1b4e5492ecec89c4ac

```rust
pub struct A {
    r#type: String
}
```

It is not great, but maybe it is something we want?

xunilrj (approved):

crates/rslint_parser/src/ast/generated/nodes.rs
ematipico: Does it work for methods? Because `ty` is a method in this case.

MichaReiser: It does work but it also requires to change the call-sites to `n.r#type()` [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b0bd55e189eaadda36be9ebe16b33738)

We may be able to change the `type` name in the future. For example, is this a type reference or a type binding?

xtask/src/codegen/kinds_src.rs
```diff
@@ -492,11 +498,23 @@ impl Field {
...
+				..
+			} => {
+				if !name_from_label {
+					panic!("The node {} doesn't have a label", ty);
+				}
+				let mut final_name = name.clone();
+				for prefix in LANGUAGE_PREFIXES {
+					final_name = final_name.replace(prefix, "");
+				}
+				if final_name == "type" {
```
MichaReiser, line 514: Why do we still need this if we use explicit labels anyway? Or is this for the node types?

xtask/src/codegen/kinds_src.rs
```diff
@@ -492,11 +498,23 @@ impl Field {
...
+				name_from_label,
+				ty,
+				..
+			} => {
+				if !name_from_label {
+					panic!("The node {} doesn't have a label", ty);
+				}
+				let mut final_name = name.clone();
+				for prefix in LANGUAGE_PREFIXES {
+					final_name = final_name.replace(prefix, "");
```
MichaReiser, line 512: You don't want to replace all occurrences but only if it's a prefix. You can use [`strip_prefix`](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix)

MichaReiser (commented):
There might have been a misunderstanding in #1733 The title says that labels should be enforced but the content of the task sets the goal to remove the need for explicit labels: 

> Change the codegen to strip the Js extensions from field names to reduce the need for **explicit** labels. For example, ExpressionStatement must use an explicit label for its inner expression so that it isn't named js_expression.

I personally would find it valuable if I don't need to prefix all children because it makes the grammar super verbose.

ematipico:
> There might have been a misunderstanding in #1733 The title says that labels should be enforced but the content of the task sets the goal to remove the need for explicit labels:
> 
> > Change the codegen to strip the Js extensions from field names to reduce the need for **explicit** labels. For example, ExpressionStatement must use an explicit label for its inner expression so that it isn't named js_expression.
> 
> I personally would find it valuable if I don't need to prefix all children because it makes the grammar super verbose.

I can close the PR if you want. Sorry for the misunderstanding.

MichaReiser:
> > There might have been a misunderstanding in #1733 The title says that labels should be enforced but the content of the task sets the goal to remove the need for explicit labels:
> > > Change the codegen to strip the Js extensions from field names to reduce the need for **explicit** labels. For example, ExpressionStatement must use an explicit label for its inner expression so that it isn't named js_expression.
> > 
> > 
> > I personally would find it valuable if I don't need to prefix all children because it makes the grammar super verbose.
> 
> I can close the PR if you want. Sorry for the misunderstanding.

I do find it valuable that we automatically strip the `Js` prefix because it would then not be required to have explicit labels and I think your PR already accounts for that, doesn't it?

crates/rslint_parser/src/ast/generated/nodes.rs
ematipico: Personally, I wouldn't call it `type` at all because it could create confusion depending on what we are evaluating.

For example:

```
function foo<T, S, V>(opts: Options): Foo {}
```

If we had a simple `r#type()` function, what's the type referred to? `Foo` or `<T, S, V>`?  I'd prefer to have more descriptive names.

ematipico:
> > > There might have been a misunderstanding in #1733 The title says that labels should be enforced but the content of the task sets the goal to remove the need for explicit labels:
> > > > Change the codegen to strip the Js extensions from field names to reduce the need for **explicit** labels. For example, ExpressionStatement must use an explicit label for its inner expression so that it isn't named js_expression.
> > > 
> > > 
> > > I personally would find it valuable if I don't need to prefix all children because it makes the grammar super verbose.
> > 
> > 
> > I can close the PR if you want. Sorry for the misunderstanding.
> 
> I do find it valuable that we automatically strip the `Js` prefix because it would then not be required to have explicit labels and I think your PR already accounts for that, doesn't it?

I removed the logic that forces to use a label and kept the stripping part. Thank you for the suggestions!

xtask/src/codegen/kinds_src.rs
```diff
@@ -501,10 +503,18 @@ impl Field {
...
 			Field::Node { name, .. } => {
-				if name == "type" {
+				let mut final_name = name.clone();
+				for prefix in LANGUAGE_PREFIXES {
+					final_name = String::from(
+						final_name
+							.strip_prefix(prefix)
+							.unwrap_or_else(|| final_name.as_str()),
+					);
+				}
```
MichaReiser, line 513: Nit: We may now end up stripping multiple prefixes (probably super rare). Might be safer to find the first prefix and then strip it [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=539c75135043e586d7266255367e4558)

```rust
 let prefixes = HashSet::from(["js", "ts", "tsx"]);
  let input = "js_expression";
  
  
  let (prefix, tail) = input.split_once('_').unwrap_or(("", input));
  
  let name = if prefixes.contains(prefix) {
      tail
  } else {
      input
  };
```

MichaReiser (approved):
And our code gen becomes better and better every day :)

xtask/src/codegen/kinds_src.rs
ematipico: Ultra rare but let's do it! The safer, the better

PR #1752, opened 2021-11-05
feat: unknown nodes to track incorrect code

### LINKED ISSUE #1751: Change the codegen to skip the `SyntaxNode`, `SyntaxToken`, and `SyntaxElement` nodes and instead, resolve them to the types from the parser crate.

ematipico:
## Summary

This PR closes #1751 

New nodes have been introduced to track specific error inside the code that is parsed.

The code inside the formatter and `rslint_parser` has been updated but no implementation has been done. This will be part of another PR.

## Test Plan

```
cargo lint
cargo test
cargo format
```


xtask/src/codegen/generate_nodes.rs
```diff
@@ -5,11 +5,30 @@ use crate::{
...
+		.collect();
+
+	let filtered_nodes: Vec<_> = ast
+		.nodes
+		.iter()
+		.filter(|e| !DENIED_NODES.contains(&e.name.as_str()))
+		.collect();
+
 	let (node_defs, node_boilerplate_impls): (Vec<_>, Vec<_>) = ast
 		.nodes
```
MichaReiser, line 25: Should this use `filtered_nodes` instead?

xtask/js.ungram
```diff
@@ -20,6 +20,23 @@
...
+// - SyntaxNode, SyntaxToken and SyntaxElement will be stripped from the codegen
+// - Unknown nodes are special nodes used to keep track of broken code; they are
+//   not part of the grammar but they will appear inside the green tree
+//
+
+
+
+SyntaxNode = SyntaxNode
+SyntaxToken = SyntaxToken
+SyntaxElement = SyntaxNode | SyntaxToken
```
MichaReiser, line 32: I saw that having all three definitions complicates the codegen and we only ever use `SyntaxElement` in the grammar. I think it's sufficient if we change the `SyntaxElement` definition and add a comment explaining what it is
```suggestion
SyntaxElement = SyntaxElement
```

xtask/src/codegen/generate_nodes.rs
```diff
@@ -44,10 +63,20 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
...
+					// it will return tokens a and nodes regardless because there's an error
+					// inside the code
+					if *is_unknown_node {
+						quote! {
+							pub fn #method_name(&self) -> SyntaxElementChildren {
+								support::element(&self.syntax)
+							}
+						}
+					} else if *optional {
 						quote! {
```
MichaReiser, line 80: I find the comment here a bit misleading because we test `is_unknown_node` which is true if the return type is `SyntaxElement` and not if the node is called `Unknown`. Should we change the boolean to `is_syntax_element_type` instead? 

It also decouples the concepts of `Unknown` and `SyntaxElement`. Any node could contain a child that returns a `SyntaxElement`.

MichaReiser (commented):
Looks great! There are some compile error handles in the formatter because the `Unknown` cases aren't handled. I think you can add a `todo()` for now.

Issue #1724, opened 2021-10-29
☂️ Use fixed slots for children


MichaReiser:
### Description

> RSLint and rust-analyzer only store nodes or tokens if they've been present in the source file. Let's take the following example of a `for` statement that only has a condition but misses the `init` and `update` nodes. 
> 
> ```jsx
> for (;i < 10;) {
> }
> ```
> 
> RSLint generates the following CST (ignoring trivia):
> 
> ```yaml
> script:
>   - for-stmt:
>     - for-keyword
>     - l-paren    #// <-- No ForInit after the l-paren
>     - semicolon  
>     - for-test:
>       - binary-expression:
>         - name-ref:
>           - ident(test)
>         - l-angle 
>         - literal
>           - number(10)
>     - semicolon 
>     - r-paren
>     - block
> ```
> 
> The downside of not storing missing (even optional) nodes is that accessing a particular field requires iterating over the children of a node to find its position. For example, the `for-test` is the 4th element if the for-loop has no `for-init` but is the 5th element otherwise. 
> 
> The CST can guarantee fixed positions of its children if missing optional or mandatory children are marked as missing inside its children collection. There are multiple ways to do so, some options are:
> 
> - Storing `None` for missing children
> - Using a fake missing node/token
> - Creating the missing nodes and tokens and flag them as "missing"
> - Use an additional bitmask to encode which nodes are present or missing
> 
> ```yaml
> script:
>   - for-stmt:
>     - for-keyword
>     - l-paren
>     - None # <<<
>     - semicolon
>     - for-test:
>       - binary-expression:
>         - name-ref:
>           - ident(test)
>         - l-angle 
>         - literal:
>           - number(10)
>     - semicolon 
>     - None # <<<
>     - r-paren
>     - block
> ```
> 
> Storing missing elements increases memory consumption because the data structure must encode the information whatever an element is present or missing. We explored different solutions, including bitflags, to mark present and missing elements to keep the overhead to a minimum.

Part of #1718 

See #1721 for discussion around the implementation

- [x] Support empty slots in rowan.  #1761
- [x] Change the parser to inform the `TreeSink` about missing children #1763
- [ ] Change the parser to emit the `missing` children. Probably requires going through the parsing of every node. #1762 
- [x] Create the `SyntaxKind::List` that represents a sequence of children (for example, the statements in a block statement OR the array elements in an array) #1728
[...]

PR #1753, opened 2021-11-08
Pre and PostUpdateExpression

MichaReiser:
## Summary

Splits out the `PreUpdateExpression` and `PostUpdateExpression` from the `UnaryExpr`.
Separating the two is necessary for implementing fixed slots for children (#1724) because it's otherwise unclear whatever the operand or operator comes first.

It would have been sufficient to only introduce the `PostUpdateExpression` and model the `PreUpdateExpression` as an `UnaryExpr`. However, I do think that it's more important to keep the symmetry of the Update expressions and the previously existing `is_update()` helpers on the `UnaryExpr` suggest that it's important for passes to know if they're operating on an `Update` expression (which is exactly what splitting the nodes gives you).

## Test Plan

Coverage doesn't change 

```

│   17608   │ 16767  │  840   │   1    │  95.22   │
```


xunilrj (approved):

ematipico (changes requested):
We should have a test for `PRE_UPDATE_EXPRESSION` TOO

MichaReiser:
I added a dedicated test for pre update expressions

PR #1755, opened 2021-11-09
feat(codegen): Respect label for tokens & support optional token unions

MichaReiser:
## Summary

Changes the syntax codegen to respect the label for single tokens, for example, `optional: '?.'` and to create a single field for an optional token union, for example, `operator: ('a' | 'b')?`

Part of #1725 

## Test Plan

Ran the codegen on the grammar proposed in #1719 and the codegen correctly generated the union types & and no longer failed on the `?.` token.


ematipico:
Yeah that's the missing part I was trying to fix. Thank you for looking into this!

ematipico (approved):

PR #1760, opened 2021-11-09
feat: move away from generic errors

### LINKED ISSUE #1759: Revisit the parse rules and replace the generic `Error` handler
Revisit the parser logic and replace the `SyntaxKind` `"ERROR"` with the new nodes that are prefixed with `Unknown*`

ematipico:
## Summary

This PR closes #1759 

As explained in #1725 , we want to be able to detect errors inside the unions, meaning that when the parser sees an error, instead of using a generic token called `ERROR`, we use now more specific `Unknown*` tokens that become part of the grammar.

What's been done:

- added new `Unknown*` nodes to the grammar (ungrammar file);
- added to the code gen a new utility file, it will be useful when we will remove `ERROR` from the code base, probably;
- removed the usage of `ERROR` from the parser and used `JS_UNKNOWN*` tokens instead;

## Test Plan

```
cargo test
cargo lint
cargo format
```


crates/rslint_parser/src/ast.rs
```diff
@@ -156,7 +156,7 @@ impl<N: AstNode> Iterator for AstNodeListIterator<N> {
 		self.inner.find_map(|e| {
 			let syntax = e.into_node()?;
 
-			if syntax.kind() == SyntaxKind::ERROR {
+			if syntax.kind() == SyntaxKind::ERROR || is_unknown_kind(syntax.kind()) {
```
MichaReiser, line 159: Your changes should actually allow removing the whole workaround because let's say this is an `AstSeparatedList<Expr>`  (sequence expression). The parser should only generate `UnknownExpr` that are castable to `Expr`. So, `syntax.to::<Expr>()` should succeed.

xtask/src/codegen/generate_utils.rs
```diff
@@ -0,0 +1,38 @@
+use super::kinds_src::{AstSrc, KindsSrc};
+use crate::codegen::Result;
+use quote::{format_ident, quote};
+
+pub fn generate_utils(_ast: &AstSrc, grammar: KindsSrc) -> Result<String> {
+	let matches: Vec<_> = grammar
+		.nodes
+		.iter()
+		.filter_map(|node| {
+			if !node.contains("UNKNOWN") || node.contains("TS_UNKNOWN") {
```
ematipico, line 10: I don't like the `|| node.contains("TS_UNKNOWN")` but it's an unfortunate coincidence

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -108,7 +108,7 @@ impl<'a> TreeSink for LosslessTreeSink<'a> {
 					it.kind,
 					self.text
 						.get(start..end)
-						.unwrap_or_else(|| self.text.get(start - 1..end).unwrap()),
+						.unwrap_or_else(|| self.text.get(start - 1..end).unwrap_or("")),
```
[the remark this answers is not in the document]
ematipico: Beats me. Once I started to replace `ERROR` with `JS_UNKNOWN_*` this error popped here. I tried to debug it but my knowledge of internals of the parser is still too low to understand that part.

ematipico:
> It could also make sense to tackle this on a per node basis or node type (e.g. all statements, all expressions). You may also want to start with the simpler nodes first (statements?).

I would yeah. As I highlighted on discord, we don't have enough error test cases to make sure that our changes are actually correct or not.

MichaReiser:
I spend some more time thinking about how to handle TypeScript syntax inside of JavaScript and came to the conclusion that we should apply the same strategy as for module syntax inside of a script (or forbidden syntax inside of a module), for example, import and export statements. 

I believe it's best if we parse import/export statements even inside of scripts but then add diagnostics that these are not permitted. This has a few advantages

* Autocompletion and all analysis continue to work correctly
* We can give the user two options: Convert the import/export to an alternative syntax or convert the file to a Module
* It's still clear that the module contains invalid syntax because of the added diagnostics and the compiler can decide if lint rules, transformers, etc. should run or not. 

The alternative of instead inserting an `UnknownStatement` comes at the cost that we loos a lot of information and may make it harder for us to provide actionable fixes. 

I believe we shouldn't tread typescript nodes in JS any different from import/export inside scripts. Meaning, we should parse it assuming it's typescript but add diagnostics , explaining that the syntax isn't valid in JS. That can mean that we promote a JS node to a typescript node if we model TS/JS separately or that we insert the child in the expected slot. 

Regarding the PR. I think it would make sense to split it into smaller pieces, e.g. a group of statements/expressions, or around a topic (object expressions). Splitting the PR eases the code review, gives an opportunity to discuss, document, and test the error recovery.

ematipico:
I am going to close this PR and start over. 

I will create a new branch and all my PRs regarding this work will be merged into that branch. This should help. Thank you for the feedback!

PR #1761, opened 2021-11-10
Add support for empty slots to nodes

MichaReiser:
## Summary

Our goal is to use fixed offsets to get a specific child instead of iterating and trying to match the child by its type. 

This PR extends the rowan `*Node` types to support empty slots, a slot for a child that wasn't present in the tree either because it's an optional child or because of a syntax error. 

The proposed implementation renames the `GreenChild` enum to `Slot` and adds an `Empty` variant. This isn't the most space-efficient representation if nodes have many optional children / sparse nodes (requires 16 bytes for every empty slot). But I believe it's sufficient for now and we can research more efficient sparse array data structures in the future if this turns out to be a problem measuring with real-world programs. 

Part of #1724 

## Test Plan

Added new tests for siblings and children. Ran the coverage and there's no change in the test results.

### DIFF
--- a/crates/rome_rowan/src/cursor.rs
+++ b/crates/rome_rowan/src/cursor.rs
@@ -89,13 +89,14 @@ use std::{
 	iter,
 	mem::{self, ManuallyDrop},
 	ops::Range,
-	ptr, slice,
+	ptr,
 };
 
 use countme::Count;
 
+use crate::green::{Child, Children};
 use crate::{
-	green::{GreenChild, GreenElementRef, GreenNodeData, GreenTokenData, SyntaxKind},
+	green::{GreenElementRef, GreenNodeData, GreenTokenData, SyntaxKind},
 	sll,
 	utility_types::Delta,
 	Direction, GreenNode, GreenToken, NodeOrToken, SyntaxText, TextRange, TextSize, TokenAtOffset,
@@ -118,7 +119,7 @@ struct NodeData {
 
 	rc: Cell<u32>,
 	parent: Cell<Option<ptr::NonNull<NodeData>>>,
-	index: Cell<u32>,
+	slot: Cell<u32>,
 	green: Green,
 
 	/// Invariant: never changes after NodeData is created.
@@ -141,7 +142,7 @@ unsafe impl sll::Elem for NodeData {
 		&self.next
 	}
 	fn key(&self) -> &Cell<u32> {
-		&self.index
+		&self.slot
 	}
 }
 
@@ -227,7 +228,7 @@ impl NodeData {
 	#[inline]
 	fn new(
 		parent: Option<SyntaxNode>,
-		index: u32,
+		slot: u32,
 		offset: TextSize,
 		green: Green,
 		mutable: bool,
@@ -237,7 +238,7 @@ impl NodeData {
 			_c: Count::new(),
 			rc: Cell::new(1),
 			parent: Cell::new(parent.as_ref().map(|it| it.ptr)),
-			index: Cell::new(index),
+			slot: Cell::new(slot),
 			green,
 
 			mutable,
@@ -255,7 +256,7 @@ impl NodeData {
 				) {
 					sll::AddToSllResult::AlreadyInSll(node) => {
 						if cfg!(debug_assertions) {
-							assert_eq!((*node).index(), (*res_ptr).index());
+							assert_eq!((*node).slot(), (*res_ptr).slot());
 							match ((*node).green(), (*res_ptr).green()) {
 								(NodeOrToken::Node(lhs), NodeOrToken::Node(rhs)) => {
 									assert!(ptr::eq(lhs, rhs))
@@ -332,20 +333,28 @@ impl NodeData {
 			Green::Token { ptr } => GreenElementRef::Token(unsafe { &*ptr.as_ref() }),
 		}
 	}
+
+	/// Returns an iterator over the siblings of this node. The iterator is positioned at the current node.
 	#[inline]
-	fn green_siblings(&self) -> slice::Iter<GreenChild> {
-		match &self.parent().map(|it| &it.green) {
-			Some(Green::Node { ptr }) => unsafe { &*ptr.get().as_ptr() }.children().raw,
-			Some(Green::Token { .. }) => {
-				debug_assert!(false);
-				[].iter()
+	fn green_siblings(&self) -> Option<Siblings> {
+		match &self.parent()?.green {
+			Green::Node { ptr } => {
+				let parent = unsafe { &*ptr.get().as_ptr() };
+
+				Some(Siblings::new(parent, self.slot()))
 			}
-			None => [].iter(),
+			Green::Token { .. } => {
+				debug_assert!(
+					false,
+					"A token should never be a parent of a token or node."
+				);
+				None
+			}
 		}
 	}
 	#[inline]
-	fn index(&self) -> u32 {
-		self.index.get()
+	fn slot(&self) -> u32 {
+		self.slot.get()
 	}
 
 	#[inline]
@@ -365,9 +374,8 @@ impl NodeData {
 		while let Some(parent) = node.parent() {
 			let green = parent.green().into_node().unwrap();
 			res += green
-				.children()
-				.raw
-				.nth(node.index() as usize)
+				.slots()
+				.nth(node.slot() as usize)
 				.unwrap()
 				.rel_offset();
 			node = parent;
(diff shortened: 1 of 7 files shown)


crates/rome_rowan/src/cursor.rs
```diff
@@ -332,20 +333,28 @@ impl NodeData {
...
 	}
+
+	/// Returns an iterator over the siblings of this node. The iterator is positioned at the current node.
 	#[inline]
-	fn green_siblings(&self) -> slice::Iter<GreenChild> {
+	fn green_siblings(&self) -> Option<Siblings> {
 		match &self.parent().map(|it| &it.green) {
-			Some(Green::Node { ptr }) => unsafe { &*ptr.get().as_ptr() }.children().raw,
+			Some(Green::Node { ptr }) => {
+				let parent = unsafe { &*ptr.get().as_ptr() };
```
ematipico, line 342: I know this is existing code, but I wonder if we could start having `// SAFETY: ` comments around `unsafe` code that explain why it's "safe" to use `unsafe` code.

crates/rome_rowan/src/cursor.rs
```diff
@@ -1329,4 +1317,44 @@ impl Iterator for PreorderWithTokens {
...
+		Self {
+			parent,
+			start_slot: 0,
+		}
+	}
+
+	pub fn start_at_slot(&mut self, start: u32) {
+		assert!(
+			(start as usize) < self.parent.slots().len(),
+			"Start slot out of bounds"
```
ematipico, line 1338: Should we provide a better message? We could provide the `len()` as additional information inside the message.

crates/rome_rowan/src/cursor.rs
```diff
@@ -1329,4 +1317,44 @@ impl Iterator for PreorderWithTokens {
...
+		);
+		self.start_slot = start;
+	}
+
+	pub fn forwards(&self) -> Children<'a> {
```
ematipico, line 1343: We should document this function and what it does.

crates/rome_rowan/src/cursor.rs
```diff
@@ -1329,4 +1317,44 @@ impl Iterator for PreorderWithTokens {
...
+		let mut slots = self.parent.slots().enumerate();
+
+		slots.nth(self.start_slot as usize);
+
+		Children::new(slots)
+	}
+
+	pub fn backwards(&self) -> impl Iterator<Item = Child<'a>> {
```
ematipico, line 1351: We should document this function and what it does.

crates/rome_rowan/src/cursor.rs
```diff
@@ -332,20 +333,28 @@ impl NodeData {
...
-			Some(Green::Node { ptr }) => unsafe { &*ptr.get().as_ptr() }.children().raw,
+			Some(Green::Node { ptr }) => {
+				let parent = unsafe { &*ptr.get().as_ptr() };
+
+				let mut siblings = Siblings::new(parent);
+				siblings.start_at_slot(self.slot());
+				Some(siblings)
+			}
 			Some(Green::Token { .. }) => {
 				debug_assert!(false);
```
ematipico, line 349: This debug assert doesn't make much sense, I think. Can we delete it?

crates/rome_rowan/src/cursor.rs
```diff
@@ -1329,4 +1317,44 @@ impl Iterator for PreorderWithTokens {
...
+
+		slots.nth(self.start_slot as usize);
+
+		Children::new(slots)
+	}
+
+	pub fn backwards(&self) -> impl Iterator<Item = Child<'a>> {
+		let mut slots = self.parent.slots().enumerate();
+
+		slots.nth_back(slots.len() - 1 - self.start_slot as usize);
```
ematipico, line 1354: What's the point of using `nth_back` if we don't use the returned element?

crates/rome_rowan/src/cursor.rs
```diff
@@ -1329,4 +1317,44 @@ impl Iterator for PreorderWithTokens {
...
+			(start as usize) < self.parent.slots().len(),
+			"Start slot out of bounds"
+		);
+		self.start_slot = start;
+	}
+
+	pub fn forwards(&self) -> Children<'a> {
+		let mut slots = self.parent.slots().enumerate();
+
+		slots.nth(self.start_slot as usize);
```
ematipico, line 1346: What's the point of using `nth` if we don't use the returned element?

crates/rome_rowan/src/green/node.rs
```diff
@@ -143,15 +160,19 @@ impl GreenNodeData {
 	/// Children of this node.
 	#[inline]
 	pub fn children(&self) -> Children<'_> {
-		Children {
+		Children::new(self.slots().enumerate())
+	}
+
+	pub fn slots(&self) -> Slots<'_> {
```
ematipico, line 166: Should we `#[inline]` the function? Also, let's add a small doc comment

crates/rome_rowan/src/green/node.rs
```diff
@@ -160,43 +181,55 @@ impl GreenNodeData {
...
-		Some((idx, child.rel_offset(), child.as_ref()))
+		Some((idx, slot.rel_offset(), slot))
 	}
 
+	/// Replaces the child in a slot with a new element or removes it
 	#[must_use]
-	pub fn replace_child(&self, index: usize, new_child: GreenElement) -> GreenNode {
-		let mut replacement = Some(new_child);
-		let children = self.children().enumerate().map(|(i, child)| {
+	pub fn replace_slot(&self, index: usize, replacement: Option<GreenElement>) -> GreenNode {
```
ematipico, line 193: Out of curiosity: does that mean that we can replace a slot with a `None`? If so, what's the use case?

crates/rome_rowan/src/green/node.rs
```diff
@@ -373,4 +416,147 @@ impl<'a> DoubleEndedIterator for Children<'a> {
...
+	pub fn rel_offset(&self) -> TextSize {
+		self.rel_offset
+	}
+	pub fn element(&self) -> GreenElementRef<'a> {
+		self.element
+	}
+}
+
+impl<'a> TryFrom<(usize, &'a Slot)> for Child<'a> {
+	type Error = ();
```
ematipico, line 441: Why an empty error?

crates/rome_rowan/src/green/node_cache.rs
```diff
@@ -49,47 +50,60 @@ fn token_hash(token: &GreenTokenData) -> u64 {
...
+	})
 }
 
 impl NodeCache {
+	/// Hash used for nodes that haven't been cached because it has too many slots or
+	/// one of its children wasn't cached.
+	const UNCACHED_NODE_HASH: u64 = 0;
+
+	/// Hash for an empty slot
+	const EMPTY_SLOT_HASH: u64 = 1;
```
ematipico, line 77: I presume this value won't ever conflict with the hash generated by `node_hash` and `token_hash`?

crates/rome_rowan/src/green/node_cache.rs
```diff
@@ -104,24 +118,27 @@ impl NodeCache {
...
-				let lhs = node.0.children();
-				let rhs = children_ref.iter().map(|(_, it)| it.as_deref());
+			node.0.kind() == kind && node.0.slots().len() == slots_ref.len() && {
+				let lhs = node.0.slots();
+				let lhs = lhs.map(|slot| element_id(slot.as_ref()));
 
-				let lhs = lhs.map(element_id);
-				let rhs = rhs.map(element_id);
+				let rhs = slots_ref.iter().map(|(_, element)| {
+					// Converts &Option<NodeOrToken<GreenNode, GreenToken> to Option<NodeOrToken<&GreenNode, &GreenToken>>
```
ematipico, line 126:
```suggestion
					// Converts &Option<NodeOrToken<GreenNode, GreenToken>> to Option<NodeOrToken<&GreenNode, &GreenToken>>
```

crates/rome_rowan/src/tree_builder.rs
```diff
@@ -46,7 +46,12 @@ impl<L: Language> TreeBuilder<'_, L> {
 	#[inline]
 	pub fn token(&mut self, kind: L::Kind, text: &str) {
 		let (hash, token) = self.cache.token(L::kind_to_raw(kind), text);
-		self.children.push((hash, token.into()));
+		self.children.push((hash, Some(token.into())));
+	}
+
+	#[inline]
+	pub fn empty(&mut self) {
```
ematipico, line 53: `empty` let me think to a function that tries to remove  all the children. What do you think of `empty_slot`?
Or if think `empty` if fine, maybe some documentation docs might help

ematipico (commented):
I still need time to wrap my head around this PR. I left some questions and comments, I will come back to review it

crates/rome_rowan/src/cursor.rs
```diff
@@ -332,20 +333,28 @@ impl NodeData {
...
+			}
 			Some(Green::Token { .. }) => {
 				debug_assert!(false);
-				[].iter()
+				None
 			}
-			None => [].iter(),
+			None => None,
 		}
 	}
```
yassere, line 354: Since this is returning an option now, adding a `?` to the end of `&self.parent().map(|it| &it.green)` would make this a bit cleaner. You could remove the `Some` wrapping from every match arm and the `None => None` arm entirely.

crates/rome_rowan/src/cursor.rs
MichaReiser: That makes sense but I would need to dig deeper here to understand it myself. It probably makes sense to revisit the tree structure (and all its unsafeness) after we decided to keep or throw away the inline-mutation. It might then no longer be needed to have any unsafe code.

crates/rome_rowan/src/cursor.rs
MichaReiser: It does make sense because that branch should never be executed (the parent should never be a token). The code throws in debug mode but recovers in production.

crates/rome_rowan/src/cursor.rs
MichaReiser: We traverse the iterator from the back until we reach our start element (in which we're not interested in, we are only interested in siblings, but not the element from which we started). Calling `next_back` (or `next()` on the `rev` iterator created further down) will then return the element right in front of the starting element.

crates/rome_rowan/src/cursor.rs
MichaReiser: Same as for `nth_back`. We make the next "cursor" of the iterator point to the start node so that calling `next()` then returns the element right after the starting element (the next sibling)

crates/rome_rowan/src/green/node.rs
MichaReiser: Good point. We should probably keep the `replace_child` name since we're not replacing the slot but its content (and None means the slot is empty).

crates/rome_rowan/src/green/node.rs
MichaReiser: Because it's sufficient for the use case here where all we need to know is if the transformation was successful or not.

crates/rome_rowan/src/green/node_cache.rs
MichaReiser: In theory, it could but should be unlikely. That's why the cache performs a shallow equal comparison if two nodes match by the hash.

crates/rome_rowan/src/cursor.rs
ematipico: Ah lol
Then... comment? I mean, it was unexpected O.o

crates/rome_rowan/src/cursor.rs
ematipico: Got it! 
I now understand why we use this "trick" to position to the correct slot, but the next developer that will arrive here might be confused like me. Do think it's worth leaving the exact answer you gave to me inside the code as comment?

crates/rome_rowan/src/cursor.rs
ematipico: Got it! 
I now understand why we use this "trick" to position to the correct slot, but the next developer that will arrive here might be confused like me. Do think it's worth leaving the exact answer you gave to me inside the code as comment?

yassere (commented):
Just a couple minor style suggestions.

MichaReiser:
Updated PR with suggestions from the reviews (thanks everyone)

PR #1762, opened 2021-11-10
Use fixed offsets for Script, Module, ...

MichaReiser:
## Summary

Changes the parser to generate the missing events and changes the codegen to use fixed offsets for the following nodes:

* Script
* Module

Part of #1724 

Depends on #1738 and #1760


MichaReiser:
Abandon, there's too much that happend since...

PR #1763, opened 2021-11-10
Parser: Create empty slots for expected tokens

MichaReiser:
## Summary

Changes the parser to generate a `missing` event if it expected a specific token that isn't present in the source code. Tracking the missing children of a node is required to implement fixed offsets for children. 

It's probably best to only review the second commit as the first commit is part of #1761 

Part of #1724 

## Test Plan

Verified that coverage isn't changing.

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -119,6 +119,14 @@ impl<L: Language> SyntaxNode<L> {
 		SyntaxNode::from(cursor::SyntaxNode::new_root(green))
 	}
 
+	/// Returns the element stored in the slot with the given index. Returns [None] if the slot is empty.
+	///
+	/// ## Panics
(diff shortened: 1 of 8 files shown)


ematipico (approved):
Well, I believe we're doing something amazing here and we will see the benefits really soon!

yassere:
This looks like it handles missing mandatory fields. Will missing optional fields have to be handled case-by-case throughout the parser?

MichaReiser:
> 
> 
> This looks like it handles missing mandatory fields. Will missing optional fields have to be handled case-by-case throughout the parser?

Yes, we'll need to handle every missing field in the parser (which is something I started in #1767). The PR already adds a neat helper to eat an optional token and otherwise emit the missing event. I hope to come up with something similar for nodes to make this not too painful.

PR #1764, opened 2021-11-10
feat(codegen): optimize cast codegen for enums

### LINKED ISSUE #1756: 📎 Optimize `cast` codegen for enums with only one variant
### Description

We recently changed some logic around how we enums that fallback in other enums.

This change of logic can be optimized even more while checking if an enum has one variant OR more than one variant. 

In case of multiple variants, [we emit this code](https://github.com/rome/tools/blob/main/crates/rslint_parser/src/ast/generated/nodes.rs#L5429-L5435), which is correct.

What we can optimize is that, in case of **one single variant**, like [the linked example](https://github.com/rome/tools/blob/main/crates/rslint_parser/src/ast/generated/nodes.rs#L5404-L5407), the `.clone()` call can be removed because the value gets moved just once, making `.clone` not needed.

Once thing optimization is correctly implemented, we can also remove the [clippy rule](https://github.com/rome/tools/blob/main/crates/rslint_parser/src/ast/generated/nodes.rs#L5).

To generate the new nodes, `cargo xtask syntax` must be run.

colinaaa:
## Summary

Follow the instruction of #1756, try to optimize the `.clone()`

fix #1756

### DIFF
--- a/crates/rslint_parser/src/ast/generated/nodes.rs
+++ b/crates/rslint_parser/src/ast/generated/nodes.rs
@@ -2,7 +2,6 @@
 
 #![allow(clippy::enum_variant_names)]
 #![allow(clippy::match_like_matches_macro)]
-#![allow(clippy::redundant_clone)]
 use crate::{
 	ast::*,
 	SyntaxKind::{self, *},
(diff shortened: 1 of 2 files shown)


ematipico (approved):
Looks good! Thank you for implementing this!

PR #1767, opened 2021-11-10
refactor (rslint_parser): Rename Script & Module, Stmt, and Expr

MichaReiser:
## Summary

I'm taking baby steps here :) 

This PR is part of adopting the AST Facade to our liking (#1725) and performs the following changes:

* Renames `Script` and `Module` to `JsScript` and `JsModule` and updates the children to match the grammar from the RFC (#1719)
* Renames `Stmt` to `JsAnyStatement`
* Renames `Expr` to `JsAnyExpression`
* Fix the code gen to correctly output the token kind if the token uses a label (`interpreter: 'js_shebang'`)

Part of #1725 

## Test Plan

Ran `xtask coverage` and the same number of tests are passing.


xtask/js.ungram
```diff
@@ -47,14 +47,22 @@ JsUnknownBinding = SyntaxElement*
 JsUnknownAssignmentTarget = SyntaxElement*
 
 Ident = 'ident'
-Script = 'shebang'? items:Stmt*
-Module = 'shebang'? items:ModuleItem*
+
+// TODO #1725 add directives (JsDirective (';' JsDirective)*)
+JsScript =
+	interpreter: 'js_shebang'?
+	statements: JsAnyStatement*
```
MichaReiser, line 54: I'm starting to write rules across multiple lines (and insert spaces between label and the type) because I find single-line and without spaces rather difficult to parse: 

```
ExportNamed = 'export' 'type'? 'from'?  '{' specifiers:(Specifier (',' Specifier)* ','?) * '}'
```

For example, where's the `specifier` part starting/ending?

xtask/src/codegen/generate_nodes.rs
```diff
@@ -31,38 +32,39 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
...
-								support::find_token(&self.syntax, #tokens)
-							}
-						}
-					} else if is_optional {
-						quote! {
-							pub fn #method_name(&self) -> Option<SyntaxToken> {
-								support::as_optional_token(&self.syntax, #token_kind)
+					match kind {
+						TokenKind::Many(kinds) => {
+							let tokens = token_kinds_to_code(kinds);
```
MichaReiser, line 46: I introduced the new enum `TokenKind` to differentiate between a many and single token. I had to add the single kind to correctly generate the kind if the token has a label. 

For example, for `shebang: 'js_shebang'`

Before

```rust
pub fn shebang(&self) -> Option<SyntaxToken> {
  support::as_optional_token(&self.syntax, T[shebang]); // Shebang token doesn't exist
}
```

After

```rust
pub fn shebang(&self) -> Option<SyntaxToken> {
  support::as_optional_token(&self.syntax, T[js_shebang]); 
}
```

xtask/src/codegen/generate_nodes.rs
```diff
@@ -381,3 +383,17 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
...
+}
+
+fn token_kinds_to_code(kinds: &Vec<String>) -> proc_macro2::TokenStream {
+	let streamed_tokens: Vec<proc_macro2::TokenStream> =
+		kinds.iter().map(token_kind_to_code).collect();
+
+	quote! {
+			&[#(#streamed_tokens),*]
+	}
+}
```
MichaReiser, line 399: I moved these functions and pass the kinds directly to avoid a `Option` return type that we are then `unwrap`ing in the generator.

xtask/js.ungram
ematipico: Yeah it makes total sense!

ematipico (approved):
Looks good to me!

PR #1770, opened 2021-11-11
refactor(rslint_parser): Try Statement Facade

MichaReiser:
## Summary

Changes the structure and naming of the Try and Throw statements

* Try Statement
  * Split into `TryStatement` and `TryFinallyStatement` so that we can correctly model the optionality of the catch/finally clauses.
  * Rename `test` child to `body`
  * Rename `handler` to `catch_clause`
  * Rename `finalizer` to `finally_clause` to align with `catch_clause`
  * Extract the `JsCatchDeclaration` from the `JsCatchClause` to group the optional `(e)`
  * Extend the integration tests
* Throw Statement
  * `JsThrowStatement`
  * Rename the `exception` child to `argument`

Based on #1769 and part of #1725

## Test Plan

Verified that coverage doesn't change. Added new integration tests for try statement.

PR #1771, opened 2021-11-11
refactor(rslint_parser): IfStatement Ast Node

MichaReiser:
## Summary

Adopts the Switch and If statement facade as described in #1725 

* If statement
  * Rename from `IfStmt` to `JsIfStatement`
  * Inline the condition
  * Group the `else` token and the alternative branch in an else clause
* Switch statement
  *  Rename from `SwitchStm` to `JsSwitchStatement`
  * Inline the condition and rename the child to `discriminant` (it's not a condition)
  * Rename the case union type to `JsAnySwitchClause`
  * Renamed `cons` to `consequence`

Based on #1770  and part of #1725

## Test Plan

Verified that this PR doesn't change the coverage.


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -619,11 +619,16 @@ pub(crate) fn block_items(
 /// An expression wrapped in parentheses such as `()`
```
ematipico, line 619: We have now two functions with the same comment docs. Probably we should change one of them

crates/rome_formatter/src/ts/statements/if_stmt.rs
```diff
@@ -1,34 +1,40 @@
...
+use rslint_parser::ast::{JsElseClause, JsIfStatement};
 
-impl ToFormatElement for IfStmt {
+impl ToFormatElement for JsIfStatement {
 	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-		let mut result = format_elements![
+		let formatted_else_clause = if let Some(else_clause) = self.else_clause() {
+			format_elements![space_token(), formatter.format_node(else_clause)?]
+		} else {
+			empty_element()
```
ematipico, line 12: Can't stress enough, but I love how the `empty_element()` is really coming in handy!

ematipico (approved):

PR #1773, opened 2021-11-11
feat: add replace_nodes method for SyntaxNode

yassere:
## Summary

This draft PR is a piece of the CST transform work related to #1739 

It adds a method for replacing multiple syntax nodes from a single immutable syntax tree, building a new syntax tree all at once. This allows for identifying and collecting references to multiple nodes of interest then replacing them all simultaneously.

Attempting to immutably replace multiple target nodes one-by-one can be a challenge because each replacement creates a new tree which the remaining targets nodes are no longer a part of. Roslyn has a [TrackNodes](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxnodeextensions.tracknodes?view=roslyn-dotnet-4.0.1) method which is another solution to this problem that we may also want to consider.

This approach attempts to make multiple replacements efficiently, maintaining structural sharing where possible and avoiding traversal into subtrees that don’t contain nodes targeted for replacement.

It first sorts the replacements to make use of binary search and to restrict the slice of replacements checked by subtrees during recursion. This optimization hasn’t been validated yet with benchmarks. If we do decide to use the `replace_nodes` approach as part of our CST transform infrastructure, we should test performance and evaluate assumptions (e.g. sorting a vector of replacements may not pay off in every situation).

TODO: 
 - Add tests and recheck logic in `SyntaxNodeReplacer::matches`
 - Put more thought into naming things and clean up the code/data structures/API

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -114,6 +114,20 @@ impl<L: Language> SyntaxNode<L> {
 		SyntaxNode::from(cursor::SyntaxNode::new_root(green))
 	}
 
+	pub fn replace_nodes(&self, replacements: &[(SyntaxNode<L>, SyntaxNode<L>)]) -> SyntaxNode<L> {
+		let replacer = cursor::SyntaxNodeReplacer::from_pairs(replacements);
+		SyntaxNode::new_root(self.raw.replace_nodes(&replacer))
+	}
+
+	pub fn replace_nodes_with(
+		&self,
+		nodes: &[SyntaxNode<L>],
+		compute: impl Fn(&SyntaxNode<L>) -> SyntaxNode<L>,
+	) -> SyntaxNode<L> {
+		let replacer = cursor::SyntaxNodeReplacer::from_fn(nodes, compute);
+		SyntaxNode::new_root(self.raw.replace_nodes(&replacer))
+	}
+
 	pub fn kind(&self) -> L::Kind {
 		L::kind_from_raw(self.raw.kind())
 	}
--- a/crates/rome_rowan/src/cursor.rs
+++ b/crates/rome_rowan/src/cursor.rs
@@ -882,6 +882,152 @@ impl SyntaxNode {
 		};
 		self.data().attach_child(index, data)
 	}
+
+	pub(crate) fn replace_nodes(&self, replacer: &SyntaxNodeReplacer) -> GreenNode {
+		let range = 0..replacer.replacements.len();
+		self.replace_nodes_with_range(replacer, range)
+	}
+
+	pub(crate) fn replace_nodes_with_range(
+		&self,
+		replacer: &SyntaxNodeReplacer,
+		range: Range<usize>,
+	) -> GreenNode {
+		match replacer.matches(self, range) {
+			ReplacerMatch::Exact(syntax) => syntax.green().into_owned(),
+			ReplacerMatch::None => self.green().into_owned(),
+			ReplacerMatch::Overlap(new_range) => {
+				let children = self.children_with_tokens().map(|c| match c {
+					NodeOrToken::Token(t) => NodeOrToken::Token(t.green().to_owned()),
+					NodeOrToken::Node(n) => {
+						NodeOrToken::Node(n.replace_nodes_with_range(replacer, new_range.clone()))
+					}
+				});
+				return GreenNode::new(self.kind(), children);
+			}
+		}
+	}
+}
+
+struct SyntaxNodeReplacement {
+	old: SyntaxNode,
+	new: SyntaxNode,
+}
+
+pub(crate) struct SyntaxNodeReplacer {
+	replacements: Vec<SyntaxNodeReplacement>,
+}
+
+impl SyntaxNodeReplacer {
+	pub fn from_pairs<N>(pairs: &[(N, N)]) -> Self
+	where
+		N: ToOwned,
+		N::Owned: Into<SyntaxNode>,
+	{
+		let mut replacements: Vec<SyntaxNodeReplacement> = pairs
+			.iter()
+			.map(|(old, new)| SyntaxNodeReplacement {
+				old: old.to_owned().into(),
+				new: new.to_owned().into(),
+			})
+			.collect();
+
+		replacements.sort_unstable_by_key(|replace| {
+			(
+				replace.old.text_range().start(),
+				replace.old.text_range().end(),
+			)
+		});
+
+		Self { replacements }
+	}
+
+	pub fn from_fn<N, C>(nodes: &[N], compute: C) -> Self
+	where
+		N: ToOwned,
+		N::Owned: Into<SyntaxNode>,
+		C: Fn(&N) -> N,
+	{
+		let mut replacements: Vec<SyntaxNodeReplacement> = nodes
+			.iter()
+			.map(|node| SyntaxNodeReplacement {
+				old: node.to_owned().into(),
+				new: compute(node).to_owned().into(),
+			})
+			.collect();
+
+		replacements.sort_unstable_by_key(|replace| {
+			(
+				replace.old.text_range().start(),
+				replace.old.text_range().end(),
+			)
+		});
+
+		Self { replacements }
+	}
+}
+
+impl SyntaxNodeReplacer {
+	fn matches(&self, node: &SyntaxNode, search_range: Range<usize>) -> ReplacerMatch {
+		let mut overlap_range = search_range.clone();
+		let mut has_overlap = false;
+
+		// Replacements are ordered by:
+		//   1. the start offset of the target node
+		//   2. the end offset of the target node
+		//
+		// Use binary search to find the exclusive upper bound of replacements that
+		// start before the end of the current node. If all replacements start
+		// before this node ends, the offset will be the length of this slice.
+		// If no replacements do, the offset will be zero.
+		let current_node_end = node.text_range().end();
+		let replacements_slice = &self.replacements[search_range.clone()];
+		let partition_offset =
+			replacements_slice.partition_point(|r| r.old.text_range().start() < current_node_end);
+
+		// If no replacements could overlap with the current node, return.
+		if partition_offset == 0 {
+			return ReplacerMatch::None;
+		}
+
+		// Exclude non-overlapping replacements from current slice and possible overlap range
+		let replacements_slice = &replacements_slice[..partition_offset];
+		overlap_range.end = overlap_range.start + partition_offset;
+
+		// Starting from the last replacement that could overlap with the current node,
+		// iterate backwards until either finding a match or finding a replacement that
+		// ends before the current node starts.
+		for (index, replacement) in replacements_slice.iter().enumerate().rev() {
+			// If the current node exactly matches the target node of a replacement,
+			// then return the new syntax node.
+			if node == &replacement.old {
+				return ReplacerMatch::Exact(replacement.new.to_owned());
+			}
+
+			// If the target node ends before the current node starts, exclude it and all earlier
+			// replacements from the possible overlap range.
+			if replacement.old.text_range().end() < node.text_range().start() {
+				overlap_range.start += index;
+				break;
+			}
+
+			// If the loop never reaches here, then there was either an exact match or
+			// no possible overlap.
+			has_overlap = true;
+		}
+
+		if has_overlap {
+			ReplacerMatch::Overlap(overlap_range)
+		} else {
+			ReplacerMatch::None
+		}
+	}
+}
+
+enum ReplacerMatch {
+	Exact(SyntaxNode),
+	None,
+	Overlap(Range<usize>),
 }
 
 impl SyntaxToken {
@@ -931,6 +1077,11 @@ impl SyntaxToken {
 				""
 			}
 		}
+	}
+
+	#[inline]
+	fn green(&self) -> &GreenTokenData {
+		self.data().green().into_token().unwrap()
 	}
 
 	#[inline]
@@ -1193,12 +1344,14 @@ impl Iterator for SyntaxNodeChildren {
 #[derive(Clone, Debug)]
 pub(crate) struct SyntaxElementChildren {
 	next: Option<SyntaxElement>,
+	length: usize,
 }
 
 impl SyntaxElementChildren {
 	fn new(parent: SyntaxNode) -> SyntaxElementChildren {
 		SyntaxElementChildren {
 			next: parent.first_child_or_token(),
+			length: parent.green_ref().children().len(),
 		}
 	}
 }
@@ -1210,6 +1363,12 @@ impl Iterator for SyntaxElementChildren {
 			self.next = next.next_sibling_or_token();
 			next
 		})
+	}
+}
+
+impl ExactSizeIterator for SyntaxElementChildren {
+	fn len(&self) -> usize {
+		self.length
 	}
 }
 


crates/rome_rowan/src/cursor.rs
```diff
@@ -1193,12 +1344,14 @@ impl Iterator for SyntaxNodeChildren {
...
 pub(crate) struct SyntaxElementChildren {
 	next: Option<SyntaxElement>,
+	length: usize,
 }
 
 impl SyntaxElementChildren {
 	fn new(parent: SyntaxNode) -> SyntaxElementChildren {
 		SyntaxElementChildren {
 			next: parent.first_child_or_token(),
+			length: parent.green_ref().children().len(),
```
MichaReiser, line 1354: This is now an `O(n)` operation because we need to traverse all slots to determine the length of the children. What do we need the length for?

yassere: `Children` is an `ExactSizeIterator` and I thought we could get the length of the iterator cheaply since the slice should already know its [length](https://github.com/rome/tools/blob/93bbf0128714bb399e6e3c3fa60ac55ae4186872/crates/rome_rowan/src/arc.rs#L246-L271).

I wanted the length in order to implement `ExactSizeIterator` for `SyntaxElementChildren` to avoid allocating a vector before passing the mapped children to `GreenNode::new`.

But I'll take a closer look at this. If this is `O(n)` then I'll definitely change it.

yassere: Actually, the `ExactSizeIterator` implementation I added is flawed (because it's only accurate before the iterator starts being consumed) so thanks for drawing attention to this.

crates/rome_rowan/src/api.rs
```diff
@@ -114,6 +114,20 @@ impl<L: Language> SyntaxNode<L> {
 		SyntaxNode::from(cursor::SyntaxNode::new_root(green))
 	}
 
+	pub fn replace_nodes(&self, replacements: &[(SyntaxNode<L>, SyntaxNode<L>)]) -> SyntaxNode<L> {
```
MichaReiser, line 117: Nit: Should we use a struct for replacements to make it clear what new/old are?

crates/rome_rowan/src/api.rs
```diff
@@ -114,6 +114,20 @@ impl<L: Language> SyntaxNode<L> {
...
+		let replacer = cursor::SyntaxNodeReplacer::from_pairs(replacements);
+		SyntaxNode::new_root(self.raw.replace_nodes(&replacer))
```
MichaReiser, line 119: Nit: Should we test if the green node did change and otherwise return a cloned self?

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+		N: ToOwned,
+		N::Owned: Into<SyntaxNode>,
+		C: Fn(&N) -> N,
+	{
+		let mut replacements: Vec<SyntaxNodeReplacement> = nodes
+			.iter()
+			.map(|node| SyntaxNodeReplacement {
+				old: node.to_owned().into(),
+				new: compute(node).to_owned().into(),
+			})
```
MichaReiser, line 956: What's the use case of this if we're calling the compute function immediately even before starting the traversal? I expected that we would only call the `compute` fn if the source node is present in the tree. 

Nit: You can change `from_pairs` to accept an `IntoIterator` instead and then call `from_pairs` here (after mapping the replacements) to avoid duplicating the sort etc.

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+				replace.old.text_range().start(),
+				replace.old.text_range().end(),
+			)
+		});
+
+		Self { replacements }
+	}
+}
+
+impl SyntaxNodeReplacer {
```
MichaReiser, line 970: What's the reason for two `impl SyntaxNodeReplacer`? 

Nit: Should the replacer be moved to its own file. cursor.rs starts to grow significantly

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+					NodeOrToken::Token(t) => NodeOrToken::Token(t.green().to_owned()),
+					NodeOrToken::Node(n) => {
+						NodeOrToken::Node(n.replace_nodes_with_range(replacer, new_range.clone()))
+					}
+				});
+				return GreenNode::new(self.kind(), children);
+			}
+		}
+	}
+}
```
MichaReiser, line 910: Nit: Would it be possible to move this logic into the `Replacer` so that it controls the whole "replacement" process or is there some visibility issue?

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+		//   2. the end offset of the target node
+		//
+		// Use binary search to find the exclusive upper bound of replacements that
+		// start before the end of the current node. If all replacements start
+		// before this node ends, the offset will be the length of this slice.
+		// If no replacements do, the offset will be zero.
+		let current_node_end = node.text_range().end();
+		let replacements_slice = &self.replacements[search_range.clone()];
+		let partition_offset =
+			replacements_slice.partition_point(|r| r.old.text_range().start() < current_node_end);
```
MichaReiser, line 986: There's also a [`binary_search`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) operation, at least one slices. Not sure if it could be useful here.

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+		let replacements_slice = &replacements_slice[..partition_offset];
+		overlap_range.end = overlap_range.start + partition_offset;
+
+		// Starting from the last replacement that could overlap with the current node,
+		// iterate backwards until either finding a match or finding a replacement that
+		// ends before the current node starts.
+		for (index, replacement) in replacements_slice.iter().enumerate().rev() {
+			// If the current node exactly matches the target node of a replacement,
+			// then return the new syntax node.
+			if node == &replacement.old {
```
MichaReiser, line 1003: I thirst thought that `SyntaxNode` implements referential equality but it actually compares if the nodes point to the same `GreenNode`. 

I'm wondering if this can lead to unexpected results. For example using this arbitrary example:

```
let x = "a";

function test() {
  let x = "b";
}
```

Now let's say my pass wants to rename the 'x' identifier inside test because it's shadowing the outer 'x' variable. 

So, I hold a reference to the `x` identifier node in the `test` function and now call the replacer with `(test_x, JsIdentifier::make("new_name"))`. 

My expectation would be that only the inner `x` changes but not the outer. 

However, rowan is clever and only allocates a single `JsIdentifier` green node for all `x` identifiers in the program... so the syntax nodes of the inner and outer `x` identifier point to the same green node and are, therefore equal.

So, I believe the output of my replacement then is

```
let new_name = "a";

function test() {
  let new_name = "b";
}
```
Which is not what I would have expected. 

I'm not sure if `SyntaxNode`s should even implement equality based on their green nodes. It kind of makes sense but can result in very unexpected behaviour, as seen here. 

I guess, the equality behaviour that I would expect is referential equality but that only works if we cache `SyntaxNode`s.

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+			ReplacerMatch::None
+		}
+	}
+}
+
+enum ReplacerMatch {
+	Exact(SyntaxNode),
+	None,
+	Overlap(Range<usize>),
 }
```
MichaReiser, line 1031: I like this enum, it makes the code fairly easy to understand.

crates/rome_rowan/src/cursor.rs
MichaReiser: You probably want to iterate over `slots` if you call `GreenNode::new` or you'll drop all empty slots.

MichaReiser (commented):
Neat implementation! I'm interested to see what you have in mind for validation :)

I believe there's an issue with how rowan implements `SyntaxNode` equality that can lead to unexpected outcome for pass authors.

crates/rome_rowan/src/cursor.rs
yassere: The use case is that an analyzer might not call `replace_nodes_with` directly, but could instead provide the target nodes and a compute closure to the analysis/autofix infrastructure which only calls it if necessary. Adding additional laziness here is a good suggestion though.

Yeah, I plan to clean up these function signatures and restructure some things.

crates/rome_rowan/src/cursor.rs
yassere: The `partition_point` method is a binary search [under the hood](https://github.com/rust-lang/rust/blob/e4c23daeb461ac02413eb36c8cefcc5530638a05/library/core/src/slice/mod.rs#L3571-L3576).

yassere:
@MichaReiser Thanks for the feedback and the good suggestions! I'll address them next week when I clean up this code. The equality implementation for `SyntaxNode` should incorporate absolute offset, so I don't think there's a problem with confusing two different syntax nodes from the same tree. But I'll take a closer look at it.

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+	replacements: Vec<SyntaxNodeReplacement>,
+}
+
+impl SyntaxNodeReplacer {
+	pub fn from_pairs<N>(pairs: &[(N, N)]) -> Self
+	where
+		N: ToOwned,
+		N::Owned: Into<SyntaxNode>,
+	{
+		let mut replacements: Vec<SyntaxNodeReplacement> = pairs
```
ematipico, line 927: Isn't `Vec<_>` enough?

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+			.iter()
+			.map(|(old, new)| SyntaxNodeReplacement {
+				old: old.to_owned().into(),
+				new: new.to_owned().into(),
+			})
+			.collect();
+
+		replacements.sort_unstable_by_key(|replace| {
```
ematipico, line 935: Why `sort_unstable_by_key` and not `sort_by_key`?

crates/rome_rowan/src/cursor.rs
ematipico: What's the meaning of `"current node exactly matches the target node"` in this case? How's the comparison done in this case? If the comparison is done on their length inside the the tree, we might need to expand the comment and specify it. 

If not, then I think the comparison could lead to issues.

crates/rome_rowan/src/cursor.rs
```diff
@@ -882,6 +882,152 @@ impl SyntaxNode {
...
+
+		if has_overlap {
+			ReplacerMatch::Overlap(overlap_range)
+		} else {
+			ReplacerMatch::None
+		}
+	}
+}
+
+enum ReplacerMatch {
```
ematipico, line 1027: Would it make sense to have a `Overflow` match? I mean, if the new node is going to have a range that is greater than the old one, it could give us the possibility to bail earlier from the `matches` logic and avoid a binary search.

crates/rome_rowan/src/cursor.rs
MichaReiser: Including the length may be sufficient for most practical use cases but can still be dangerous if authors accidentally start mixing nodes from different files. We may want to include a key from the `SyntaxRoot` or implement "referential" equality.

crates/rome_rowan/src/cursor.rs
yassere: The [docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by_key) recommend unstable sorting when applicable because it's faster and uses less memory. I don't think we need a stable sort here. Also, one of the constraints we might need to add is that the target nodes won't be allowed to overlap, in which case there wouldn't be any "equal" elements for this sort to reorder.

crates/rome_rowan/src/api.rs
```diff
@@ -114,6 +114,20 @@ impl<L: Language> SyntaxNode<L> {
 		SyntaxNode::from(cursor::SyntaxNode::new_root(green))
 	}
 
+	pub fn replace_nodes(&self, replacements: &[(SyntaxNode<L>, SyntaxNode<L>)]) -> SyntaxNode<L> {
+		let replacer = cursor::SyntaxNodeReplacer::from_pairs(replacements);
```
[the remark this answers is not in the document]
yassere: If the old and new node in a pair are the same syntax node, then the replacement will be a no-op because the underlying green node is the same. We can add a pre-check to exclude no-op replacements from the vector to reduce traversal. But we might also need to add more checks like preventing target nodes from overlapping, and we couldn't just exclude those replacements, so it's possible this whole operation needs to return a `Result`.

crates/rome_rowan/src/cursor.rs
yassere: No reason. Just haven't cleaned things up yet.

Sure, this could be in its own file.

crates/rome_rowan/src/cursor.rs
yassere: Currently, this doesn't efficiently handle the possibility that the range of the matchee node is fully contained within the range of a replacement target, and you're right that we should check for that. That should only be possible if a replacement targets the ancestor of the syntax node on which `replace_nodes` is called, so we can detect that before we start traversing and either exclude it or return an error.

I don't think an `Overflow` variant on the `ReplacerMatch` enum is necessary because we should handle this before we get to that point.

crates/rome_rowan/src/api.rs
yassere: Yeah, the data structures and function signatures in this draft are definitely not in their final form.

crates/rome_rowan/src/cursor.rs
yassere: > What's the meaning of "current node exactly matches the target node" in this case?

It incorporates absolute offset, so Micha's example does work correctly as long as the nodes all come from the same syntax tree. But it is worth adding additional checks and possibly returning an error if a replacement targets a node in a different tree than the one that's being traversed.

> We may want to include a key from the `SyntaxRoot`

What exactly do you mean by this? A new `SyntaxRoot` type that contains a unique key?

> implement "referential" equality.

I'm not sure why syntax nodes don't currently have referential equality, but I think it's so that they can be converted into a `SyntaxNodePtr` and later reconstructed into a node that's still "equal" to the original node. That may not be relevant to us in this case, though, so pointer-based equality might be a solution here.

crates/rome_rowan/src/cursor.rs
yassere: Oh right. Empty slots exist now! And `Slots` already implements `ExactSizeIterator` so that's perfect.

crates/rome_rowan/src/cursor.rs
ematipico: > But it is worth adding additional checks and possibly returning an error if a replacement targets a node in a different tree than the one that's being traversed.

Is it actually possible? I mean, theoretically it is possible but are we going to permit to do so via API? 
What I am trying to say, I guess that these APIs won't be called directly from the consumers, and the compiler will make sure that two nodes belong to the same  tree.

crates/rome_rowan/src/cursor.rs
ematipico: Great explanation! Thank you!

crates/rome_rowan/src/cursor.rs
yassere: > Is it actually possible to pass two nodes from different trees? I mean, theoretically it is possible but are we going to permit to do so via API?

The `old` and `new` nodes can be from different trees and often will be because the `new` node might come from a Builder API. In principle, an analyzer could build a new node and mistakenly target that as the 'old' node in a replacement, in which case the `old` node will never match anything and the replacement won't do anything.

And yes, we probably want the compiler to check for that and early error rather than traversing the entire tree. As you said, I don't expect these APIs will be called directly by analyzers, but I'm not necessarily ruling that out yet if we can find a good solution for validation.

ematipico:
@yassere what's the status of this PR? Do we intend to proceed forward?

yassere:
@ematipico I still plan to proceed with a `replace_nodes` method, but there are a number of changes necessary to make it compatible with the current version of `rome_rowan`. I'll work on those changes in tandem with the analyzer infrastructure, so I'll close this PR for now and then link to it later in a future PR to make sure I address all the feedback here.

PR #1774, opened 2021-11-12
refactor(rslint_parser): Update while loops to match new AST Facade

MichaReiser:
## Summary

Changes the js grammar to match our new AST facade as defined in #1725 (and proposed in #1719)

* `Continue` Statement
  * Rename from `ContinueStmt` to `JsContinueStatement`
  * Inline the label identifier (It's not a normal "variable" reference but a label reference) 
* `Break` Statement
  * Rename from `BreakStmt` to `JsBreakStatement`
  * Inline the label identifier (It's not a normal "variable" reference but a label reference) 
* `While` Statement
  * Rename from `WhileStmt` to `JsWhileStatement`
  * Rename `cons` to `body`
  * Inline the condition and rename `condition` to `test`
* `DoWhileStatement`
  * Rename to `JsDoWhileStatement`
  * Inline condition
  * rename `cons` to `body`
  * rename `condition` to `test`

Part of #1725 

## Test Plan

Verified the output in the updated `rast` files to correctly reflect the parsing changes and verified that the coverage isn't changing.


crates/rome_formatter/src/ts/statements/break_statement.rs
```diff
@@ -1,21 +1,21 @@
...
 };
-use rslint_parser::ast::BreakStmt;
+use rslint_parser::ast::JsBreakStatement;
 
-impl ToFormatElement for BreakStmt {
+impl ToFormatElement for JsBreakStatement {
 	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-		let break_element = formatter.format_token(&self.break_token()?)?;
-		let ident = if let Some(ident_token) = self.ident_token() {
-			group_elements(format_elements![
```
MichaReiser, line 11: The group here didn't had any effect. Groups only change the output if they contain any `soft` indent/space/line breaks.

ematipico (approved):

PR #1775, opened 2021-11-12
Update Variable Declaration to match new AST Facade

MichaReiser:
## Summary

Changes the js grammar of the variable declaration to match our new AST facade as defined in #1725 (and proposed in #1719)

* Split `VarDecl` into `JsVariableDeclarationStatement` and `JsVariableDeclaration` to correctly account for variable declarations inside for loops (that aren't statements).
* Change the parser to emit a `let` token instead of an ident if let is used inside a variable declaration (let is a valid identifier which is why the lexer returns an `ident` token)
* Rename `Declarator` to `JsVariableDeclarator`
* Split out the `init` into a `JsEqualValueClause` that holds the `=` token and the initializer expression.

This PR also includes a fix for the codegen to correctly type `('let' | 'const' | 'var')` as a required child rather than optional.

## Test Plan

Verified the rest changes manually and checked that the coverage isn't changing.


crates/rome_formatter/src/ts/declarators/variable_declaration_statement.rs
```diff
@@ -0,0 +1,47 @@
...
+		let mut declarators = Vec::with_capacity(self.declarators().len());
+
+		for declarator in self.declarators().iter() {
+			declarators.push(formatter.format_node(declarator)?);
+		}
+
+		Ok(format_elements![
+			formatter.format_token(&self.kind_token()?)?,
+			space_token(),
+			join_elements(token(","), declarators),
```
ematipico, line 29: No spaces and/or line breaks?

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -118,118 +116,6 @@ pub fn get_precedence(tok: SyntaxKind) -> Option<u8> {
 	})
 }
 
-/// Check the bound names of a variable declaration and issue errors according to `13.3.1.1`
```
ematipico, line 126: What this deleted code was for?

xtask/src/codegen/generate_nodes.rs
```diff
@@ -92,7 +101,7 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
 					} else if *optional {
 						quote! {
 							pub fn #method_name(&self) -> Option<#ty> {
-								support::as_optional_node(&self.syntax)
+								support::node(&self.syntax)
```
ematipico, line 104: What was the reason behind the change of names? 
`as_optional_node` => `node`
`as_mandatory_node` => `required_node`

ematipico:
Do we have tests for cases like:
```js
let foo = "lorem", bar = "ipsum", third = "value", fourth = 6;
```

MichaReiser:
Addressed review comments and added a new test.

ematipico (approved):

ematipico:
@MichaReiser please let some feedback around the questions I left. Nothing major but I would like have them answered.

MichaReiser (commented):
Sorry, @ematipico . It seems I forgot to finish the review...

xtask/src/codegen/generate_nodes.rs
ematipico: Oh it's OK, don't worry. One suggestion, I would prefix `optional_` to `node` and `token` => `optional_node`. This will make a good counter part to `required_`

MichaReiser: That's true. Let me revisit the naming when I work on using fixed offset

PR #1776, opened 2021-11-12
Update literals to match new AST Facade

MichaReiser:
## Summary

Refines the AST Facade for literals to match the grammar proposed in #1719. This change is part of #1725.

* Introduce new `JsAnyLiteral`
* Split `Literal` into `JsStringLiteral`, `JsBooleanLiteral`, `JsNullLiteral`, `JsRegexLiteral`, and `JsBigIntLiteral`. This allows to implement custom methods on the corresponding literal nodes.
* Renames the `number` and kinds to `JS_NUMBER_LITERAL_TOKEN` `JS_STRING_LITERAL_TOKEN` to avoid conflicts with the TS `number` keyword (and keep symmetry).
* Removed some unused keywords and special handling inside of the code gen.

## Test Plan

Manually verified the `*.rast` changes and checked that the coverage isn't changing.


ematipico (changes requested):
I think we're missing `JsUndefinedLiteral`

MichaReiser:
@ematipico I thought so too. But `undefined` is modelled as an identifier, the same way as babel/shift does (following the shift parse tree). 

```
{
  "type": "Module",
  "directives": [],
  "items": [
    {
      "type": "ExpressionStatement",
      "expression": {
        "type": "IdentifierExpression",
        "name": "undefined"
      }
    }
  ]
}
```

It's just a global. `undefined` isn't even a keyword :O

ematipico:
Oh yeah it's true!! It's possible to shadow `undedined`. Totally passed my mind.

ematipico (approved):

Issue #1747, opened 2021-11-03
📎 Test coverage diff on PRs


ematipico:
### Description

[Follow up from discord](https://discord.com/channels/678763474494423051/841927329265090590/905512345881694250).

We need to implement a way to show a diff of the coverage of the test262 on each PR. This can be done via bot or (better) via workflow directly. Via workflow would avoid us in some permission issue, where implementing a bot would require admin rights to enable them.

Ideally, the algorithm should do a diff between the test262 coverage on `main` against the one run the PR.

Stupremee:
If you need inspiration, [boa](https://github.com/boa-dev/boa/blob/main/.github/workflows/test262.yml) has the same integration in their CI. They additionally publish a web version to GitHub Pages, but that isn't required I guess.

jonahsnider:
Might be a good use case for [Danger](https://danger.systems/js/). React uses it to automate their bundle size diffs ([their config](https://github.com/facebook/react/blob/main/dangerfile.js), [example PR comment](https://github.com/facebook/react/pull/22674#issuecomment-956694730))

Boshen:
This can be closed since we now have the bot and `xtask compare`?

PR #1779, opened 2021-11-12
ci: first draft of comapre results

ematipico:
## Summary

**This is a first PR out of two for implementing #1747**

The reason why I will need to split it in two is because we will run the command `cargo xtask compare` in the PR and in the `main` branch, *but at the moment the `main` branch doesn't have this command*. Once this PR is merged, I will move into the second part.

What's done:
- created a new command called `cargo xtask compare` which compares two reports and emits the result;
- moved the workflow of test262 into a separate workflow file;
- created some logic to show the results in markdown/CLI, most of the code was taken from what [`boa` is doing](https://github.com/boa-dev/boa/blob/main/boa_tester/src/results.rs)
- optimized the JSON serialization/de-serialization  of the information, by removing some (we're not interested in the panic messages, for example)

What I plan to do next, is to print the compare using github bot to the PR

## Test Plan

Current CI


ematipico:
Going to merge so I can hit the second part of the PR. Feel free to leave some feedback, I will address it later

.github/workflows/test262.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Rename the emitted file
+        run: |
+          mv base_results.json new_results.json
+      - name: Save test results
+        uses: actions/upload-artifact@v2
+        with:
+          name: new_results
+          path: new_results.json
+
+  compare:
```
MichaReiser, line 46: It could be useful to document what the goal of each job is. 

Nit: How about naming this job `coverage-comparison` to make it clear that this job is related to coverage?

.github/workflows/test262.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.56.1
+          profile: minimal
+      - name: Cache
+        uses: Swatinem/rust-cache@v1
+      - name: Run Test262 suite
+        continue-on-error: true
+        run: cargo xtask coverage --json
```
MichaReiser, line 36: Nit: It's a bit surprising where the `base_results.json` file is coming from. How about the `cargo xtask coverage --json` prints the results to stdout and you then pipe it to a file? That would also eliminate the need for renaming the file later

```sh
cargo xtask coverage --json > new_coverage_results.json
```

xtask/src/compare/mod.rs
```diff
@@ -0,0 +1,40 @@
+use crate::{project_root, BASE_RESULT_FILE, NEW_RESULT_FILE};
```
MichaReiser, line 1: Nit: Should we rename this to `compare_json`? or is it specific to coverage results? Then how about naming it `coverage_compare`?

xtask/src/compare/results.rs
```diff
@@ -0,0 +1,245 @@
...
+
+pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
+	let base_results: TestResults =
+		serde_json::from_reader(File::open(base).expect("Can't read the file of the base results"))
+			.expect("Can't parse the JSON file of the base results");
+	let new_results: TestResults =
+		serde_json::from_reader(File::open(new).expect("Can't read the file of the new results"))
+			.expect("Can't parse the JSON file of the new results");
+
+	let base_total = base_results.summary.tests_ran as isize;
```
MichaReiser, line 14: Nit: Why `isize` rather than `usize`?

xtask/src/compare/results.rs
```diff
@@ -0,0 +1,245 @@
...
+				} else {
+					""
+				},
+			),
+		);
+
+		if !report_diff.fixed.is_empty() {
+			println!();
+			println!(
+				"<details><summary><b>Fixed tests ({}):</b></summary>",
```
MichaReiser, line 103: I love it!

xtask/src/coverage/files.rs
```diff
@@ -165,3 +173,103 @@ pub enum FailReason {
...
+pub struct TestResults {
+	#[serde(rename = "s")]
+	pub summary: Summary,
+	#[serde(rename = "p")]
+	pub details: Vec<TestResult>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Summary {
+	#[serde(rename = "a")]
```
MichaReiser, line 194: Is it worth trimming the property names from a performance perspective? It makes the JSON rather difficult to read as a human.

MichaReiser (commented):
That's neat! And I love it that you implemented it as an `xtask` so that I can now use it locally as well to identify the tests my changes broke.

xtask/src/compare/results.rs
```diff
@@ -0,0 +1,245 @@
...
+fn compare_diffs<'a>(
+	base_results: &'a TestResults,
+	new_results: &'a TestResults,
+) -> ReportDiff<'a> {
+	let mut report_diff = ReportDiff::new();
+	for base_result in &base_results.details {
+		let test_to_analyze = new_results
+			.details
+			.iter()
+			.find(|new_test| new_test.path.as_os_str().eq(base_result.path.as_os_str()));
```
MichaReiser, line 216: It would make sense to build a hash map from `path -> new_test` to avoid the `O(test^2)` complexity.

.github/workflows/test262.yml
ematipico: It makes sense! Will do

xtask/src/compare/results.rs
ematipico: Because the `total_*` variables can be negative. So, we apply the cast when we extract the variables or we do a cast when we run the diff e.g. `(new_total - base_total) as isize`.

xtask/src/compare/mod.rs
ematipico: I think the folders are named after the CLI command that they run - `xtask compare` in this case. That's why I called the folder `compare/`. I can rename the `run` function to `coverage_compare` but I would leave the name of the folder as is.

xtask/src/coverage/files.rs
ematipico: We don't care about readability in this case - I mean, we care but we are showing the result into a nice markdown table. The JSON exported is REALLY BIG so if there's a change to save some byte, I think we should take it. 

If we do care about readability inside the JSON, then I would add a new CLI option. At the moment the JSON is saved using serde's [`to_string`](https://docs.rs/serde_json/1.0.70/serde_json/fn.to_string.html) which emits a **minified** JSON.

PR #1782, opened 2021-11-12
ci: comment compare

ematipico:
## Summary

This PR implements workflow actions to tell github-bot to run the coverage on `main` branch and execute the comparison between the coverage results of this PR and the coverage results on `main`. 

This the an example of comment in case there's an increase of coverage:

| Total | 17608 | 17608 | 0 |
| Passed | 16767 | 16787 | ✅ ⏫ **+20** |
| Failed | 840 | 820 | ✅ ⏬ **-20** |
| Panics | 1 | 16787 | 0 |
| Coverage | 95.22% | 95.52% | **+0.30%** |

And this is an example in case the coverage decreases:

| Total | 17608 | 17608 | 0 |
| Passed | 16767 | 16687 | ❌ ⏬ **-80** |
| Failed | 840 | 900 | ❌ ⏫ **+60** |
| Panics | 1 | 1 | 0 |
| Coverage | 95.22% | 95.52% | **+0.30%** |

*At the moment the command is failing because I changed the logic of the command when using `--json` as suggested in a previous PR*.

## Test Plan

I run the commands manually against this PR and the `main` branch and made sure it works. Also, pasted the results emitted locally.

### DIFF
--- a/.github/workflows/test262.yml
+++ b/.github/workflows/test262.yml
@@ -50,7 +50,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+#        os: [windows-latest, macos-latest, ubuntu-latest]
+        os: [ubuntu-latest]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
@@ -73,7 +74,7 @@ jobs:
         with:
           name: new_results
       - name: Compare results on ${{ matrix.os }}
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request_target'
         id: comparison
         shell: bash
         run: |
@@ -84,12 +85,12 @@ jobs:
           echo "::set-output name=comment::$comment"
 
       - name: Get the PR number
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request_target'
         id: pr-number
         uses: kkak10/pr-number-action@v1.3
 
       - name: Find Previous Comment
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request_target'
         uses: peter-evans/find-comment@v1.3.0
         id: previous-comment
         with:
@@ -97,7 +98,7 @@ jobs:
           body-includes: Test262 conformance changes
 
       - name: Update comment
-        if: github.event_name == 'pull_request' && steps.previous-comment.outputs.comment-id
+        if: github.event_name == 'pull_request_target' && steps.previous-comment.outputs.comment-id
         uses: peter-evans/create-or-update-comment@v1.4.5
         continue-on-error: true
         with:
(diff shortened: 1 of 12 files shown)


MichaReiser:
Should we add some summary text to the coverage results (with some emoji?) so that it's immediately clear if everything is good or if it requires attention?

ematipico:
> Should we add some summary text to the coverage results (with some emoji?) so that it's immediately clear if everything is good or if it requires attention?

Feel free to suggest what you want. At the moment if some number changes, they are highlighted in bold.

xunilrj (approved):

.github/workflows/test262.yml
```diff
@@ -67,10 +64,55 @@ jobs:
         uses: Swatinem/rust-cache@v1
       - name: Run Test262 suite on main branch
         continue-on-error: true
-        run: cargo xtask coverage --json
+        run:  cargo xtask coverage --json > base_results.json
```
MichaReiser, line 67: Thanks @ematipico for taking the effort to change this even after it was merged!

crates/rslint_parser/src/parser.rs
```diff
@@ -206,12 +207,17 @@ impl<'t> Parser<'t> {
 		let m = self.start();
 		self.error(error);
 		self.bump_any();
-		m.complete(self, SyntaxKind::ERROR);
+		m.complete(self, unknown_syntax_kind);
```
MichaReiser, line 210: This looks unrelated...

xtask/src/compare/results.rs
```diff
@@ -35,9 +35,37 @@ pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
 
 	if markdown {
 		/// Generates a proper diff format, with some bold text if things change.
-		fn diff_format(diff: isize) -> String {
+		fn diff_format(diff: isize, i_am_passed_results: bool, show_increase: bool) -> String {
+			let good = "✅ ";
+			let bad = "❌ ";
+			let up = "⏫ ";
+			let down = "⏬ ";
```
MichaReiser, line 42: Love it!

MichaReiser (approved):
Looks good. Just make sure to strip out the error changes.

.github/workflows/test262.yml
MichaReiser, line 67: I noticed that it now takes quite a while until the results show up on the PR because the coverage command must run twice. Do you think there's a way to publish the result for "main" and store it (e.g. using the commit number in the name) whenever a commit gets merged so that we may just retrieve it (and only run `xtask coverage` in the case it's missing for whatever reason)

ematipico: Yes, this is something I discussed with the others during the standup. 

Boa uses Github pages to store the results.

 Cloudflare doesn't have a storage service yet. Github actions artifacts are not readable across workflows. Maybe, for now, we should use Github pages

PR #1783, opened 2021-11-12
trivia attached to GreenTokens

xunilrj:
part of: #1720
Improved changes of #1738

# Tasks

- [x] ```GreenToken``` own a list of its trivia lengths
- [x] ```SyntaxTrivia``` wraps ```GreenTokenTrivia```
- [x] ```PartialEq``` and ```Hash```
- [x] ```Display``` and ```Debug```
- [x] ```NodeCache``` must take into consideration the token trivia (not fixed by changing ```Hash``` implementation)

# Decisions

- Now the GreenToken "slice" contains all its trivia.
- To trim this trivia we have to ask leading().text_len() and trailing().text_len() and compute where the token text actually is.
- Given that ```GreenToken``` already has its trivia, we ignore the trivia when hashing and caching. Advantage: faster; Disadvantage: dangerous. At least in theory, two equal strings can come from two different parsers and end up with two different trivia.

- GreenToken now is 32 bytes bigger (16 bytes for each trivia). Hard to improve here because ```Box<>``` wants to be aligned.

```rust
pub enum GreenTokenTrivia {
	None,
	One(Trivia),
	Many(Box<Vec<Trivia>>),
}

print-type-size type: `GreenTokenTrivia`: 16 bytes, alignment: 8 bytes
print-type-size     discriminant: 8 bytes
print-type-size     variant `Whitespace`: 8 bytes
print-type-size         field `.0`: 8 bytes
print-type-size     variant `Comment`: 8 bytes
print-type-size         field `.0`: 8 bytes
print-type-size     variant `Many`: 8 bytes
print-type-size         field `.0`: 8 bytes
```

- Box<Vec<Trivia>> is frown upon. See https://github.com/rust-lang/rust-clippy/blob/b7f3f7f6082679da2da9a0b3faf1b5adef3afd3b/clippy_lints/src/types/box_collection.rs. But I still think it may be a good idea here. We will need to measure.

# How to test

```rust
> cargo test -p rome_rowan
> cargo xtask coverage
...
...
``` 

Coverage is the same as before.

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -32,6 +32,28 @@ impl Language for RawLanguage {
 	}
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub enum Trivia {
+	Whitespace(usize),
+	Comments(usize),
+}
+
+impl Trivia {
+	#[inline]
+	pub fn text_len(&self) -> TextSize {
+		match self {
+			Trivia::Whitespace(n) => (*n as u32).into(),
+			Trivia::Comments(n) => (*n as u32).into(),
+		}
+	}
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct SyntaxTrivia<L: Language> {
+	raw: cursor::SyntaxTrivia,
+	_p: PhantomData<L>,
+}
+
 #[derive(Clone, PartialEq, Eq, Hash)]
 pub struct SyntaxNode<L: Language> {
 	raw: cursor::SyntaxNode,
@@ -114,6 +136,12 @@ impl<L: Language> From<SyntaxToken<L>> for SyntaxElement<L> {
 	}
 }
 
+impl<L: Language> SyntaxTrivia<L> {
+	pub fn text(&self) -> &str {
+		self.raw.text()
+	}
+}
+
 impl<L: Language> SyntaxNode<L> {
 	pub(crate) fn new_root(green: GreenNode) -> SyntaxNode<L> {
 		SyntaxNode::from(cursor::SyntaxNode::new_root(green))
@@ -133,6 +161,47 @@ impl<L: Language> SyntaxNode<L> {
 
 	pub fn text_range(&self) -> TextRange {
 		self.raw.text_range()
+	}
+
+	pub fn text_trimmed_range(&self) -> TextRange {
+		let leading_len = self
+			.raw
+			.first_token()
+			.map(|x| x.leading().text_len())
+			.unwrap_or_else(|| 0.into());
+
+		let trailing_len = self
+			.raw
+			.last_token()
+			.map(|x| x.trailing().text_len())
+			.unwrap_or_else(|| 0.into());
+
+		let range = self.text_range();
+		TextRange::new(range.start() + leading_len, range.end() - trailing_len)
+	}
+
+	pub fn leading_range(&self) -> TextRange {
+		let leading_len = self
+			.raw
+			.first_token()
+			.map(|x| x.leading().text_len())
+			.unwrap_or_else(|| 0.into());
+
+		let range = self.text_range();
+		let start = range.start();
+		TextRange::new(start, start + leading_len)
+	}
+
+	pub fn trailing_range(&self) -> TextRange {
+		let trailing_len = self
+			.raw
+			.last_token()
+			.map(|x| x.trailing().text_len())
+			.unwrap_or_else(|| 0.into());
+
+		let range = self.text_range();
+		let end = range.start();
+		TextRange::new(end - trailing_len, end)
 	}
 
 	pub fn text(&self) -> SyntaxText {
@@ -304,12 +373,28 @@ impl<L: Language> SyntaxToken<L> {
 		self.raw.text_range()
 	}
 
+	pub fn text_trimmed_range(&self) -> TextRange {
+		self.raw.text_trimmed_range()
+	}
+
+	fn leading_range(&self) -> TextRange {
+		self.raw.leading_range()
+	}
+
+	fn trailing_range(&self) -> TextRange {
+		self.raw.trailing_range()
+	}
+
 	pub fn index(&self) -> usize {
 		self.raw.index()
 	}
 
 	pub fn text(&self) -> &str {
 		self.raw.text()
+	}
+
+	pub fn text_trimmed(&self) -> &str {
+		self.raw.text_trimmed()
 	}
 
 	pub fn parent(&self) -> Option<SyntaxNode<L>> {
@@ -348,6 +433,22 @@ impl<L: Language> SyntaxToken<L> {
 	pub fn detach(&self) {
 		self.raw.detach()
 	}
+
+	#[inline]
+	pub fn leading(&self) -> SyntaxTrivia<L> {
+		SyntaxTrivia {
+			raw: self.raw.leading(),
+			_p: PhantomData,
+		}
+	}
+
+	#[inline]
+	pub fn trailing(&self) -> SyntaxTrivia<L> {
+		SyntaxTrivia {
+			raw: self.raw.trailing(),
+			_p: PhantomData,
+		}
+	}
 }
 
 impl<L: Language> SyntaxElement<L> {
@@ -355,6 +456,27 @@ impl<L: Language> SyntaxElement<L> {
 		match self {
 			NodeOrToken::Node(it) => it.text_range(),
 			NodeOrToken::Token(it) => it.text_range(),
+		}
+	}
+
+	pub fn text_trimmed_range(&self) -> TextRange {
+		match self {
+			NodeOrToken::Node(it) => it.text_trimmed_range(),
+			NodeOrToken::Token(it) => it.text_trimmed_range(),
+		}
+	}
+
+	pub fn leading_range(&self) -> TextRange {
+		match self {
+			NodeOrToken::Node(it) => it.leading_range(),
+			NodeOrToken::Token(it) => it.leading_range(),
+		}
+	}
+
+	pub fn trailing_range(&self) -> TextRange {
+		match self {
+			NodeOrToken::Node(it) => it.trailing_range(),
+			NodeOrToken::Token(it) => it.trailing_range(),
 		}
 	}
 
@@ -386,12 +508,14 @@ impl<L: Language> SyntaxElement<L> {
 			NodeOrToken::Token(it) => it.next_sibling_or_token(),
 		}
 	}
+
 	pub fn prev_sibling_or_token(&self) -> Option<SyntaxElement<L>> {
 		match self {
 			NodeOrToken::Node(it) => it.prev_sibling_or_token(),
 			NodeOrToken::Token(it) => it.prev_sibling_or_token(),
 		}
 	}
+
 	pub fn detach(&self) {
 		match self {
 			NodeOrToken::Node(it) => it.detach(),
@@ -592,7 +716,9 @@ impl<L: Language> IntoIterator for SyntaxList<L> {
 
 #[cfg(test)]
 mod tests {
-	use crate::api::RawLanguage;
+	use text_size::TextRange;
+
+	use crate::api::{RawLanguage, Trivia};
 	use crate::{Direction, Language, SyntaxKind, SyntaxList, TreeBuilder};
 
 	#[test]
@@ -804,6 +930,76 @@ mod tests {
 		assert_eq!(
 			first_semicolon.prev_sibling_or_token(),
 			first_semicolon.siblings_with_tokens(Direction::Prev).next()
+		);
+	}
+
+	#[test]
+	pub fn syntax_text_and_len() {
+		let mut builder: crate::TreeBuilder<crate::api::RawLanguage> = crate::TreeBuilder::new();
+		builder.start_node(crate::SyntaxKind(0));
+		builder.token_with_trivia(
+			crate::SyntaxKind(0),
+			"\n\t let \t\t",
+			&[Trivia::Whitespace(3)],
+			&[Trivia::Whitespace(3)],
+		);
+		builder.finish_node();
+
+		let node = builder.finish();
+		assert_eq!("\n\t let \t\t", node.text());
+
+		let token = node.first_token().unwrap();
+		assert_eq!("\n\t let \t\t", token.text());
+		assert_eq!("let", token.text_trimmed());
+
+		assert_eq!("\n\t ", token.leading().text());
+		assert_eq!(" \t\t", token.trailing().text());
+	}
+
+	#[test]
+	pub fn syntax_range() {
+		let mut builder: crate::TreeBuilder<crate::api::RawLanguage> = crate::TreeBuilder::new();
+		builder.start_node(crate::SyntaxKind(0));
+		builder.token_with_trivia(
+			crate::SyntaxKind(0),
+			"\n\t let \t\t",
+			&[Trivia::Whitespace(3)],
+			&[Trivia::Whitespace(3)],
+		);
+		builder.token_with_trivia(
+			crate::SyntaxKind(0),
+			"a ",
+			&[Trivia::Whitespace(0)],
+			&[Trivia::Whitespace(1)],
+		);
+		builder.token_with_trivia(
+			crate::SyntaxKind(1),
+			"\n=\n",
+			&[Trivia::Whitespace(1)],
+			&[Trivia::Whitespace(1)],
+		);
+		builder.token(crate::SyntaxKind(0), "1");
+		builder.token(crate::SyntaxKind(0), ";");
+		builder.finish_node();
+
+		let node = builder.finish();
+
+		let eq_token = node
+			.descendants_with_tokens()
+			.find(|x| x.kind().0 == 1)
+			.unwrap();
+		assert_eq!(TextRange::new(11.into(), 14.into()), eq_token.text_range());
+		assert_eq!(
+			TextRange::new(12.into(), 13.into()),
+			eq_token.text_trimmed_range()
+		);
+		assert_eq!(
+			TextRange::new(11.into(), 12.into()),
+			eq_token.leading_range()
+		);
+		assert_eq!(
+			TextRange::new(13.into(), 14.into()),
+			eq_token.trailing_range()
 		);
 	}
 }
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -1,8 +1,9 @@
 use std::{fmt, iter, marker::PhantomData, ops::Range};
 
 use crate::{
-	cursor, Direction, GreenNode, NodeOrToken, SyntaxKind, SyntaxText, TextRange, TextSize,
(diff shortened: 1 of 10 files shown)


crates/rome_rowan/src/api.rs
```diff
@@ -135,10 +168,32 @@ impl<L: Language> SyntaxNode<L> {
...
+
 	pub fn text(&self) -> SyntaxText {
 		self.raw.text()
 	}
 
+	pub fn text_trimmed(&self) -> SyntaxText {
+		self.raw.text_trimmed()
+	}
+
+	pub fn leading(&self) -> Option<SyntaxTrivia<L>> {
```
MichaReiser, line 183: My understanding is that only tokens have trivia. Does this method return the first trivia in the tree (and trailing the last)? If so, I think we should rename it to `first_leading_trivia`, and `last_trailing_trivia`.

crates/rome_rowan/src/cursor.rs
```diff
@@ -148,6 +148,23 @@ unsafe impl sll::Elem for NodeData {
...
+impl Clone for SyntaxTrivia {
+	#[inline]
+	fn clone(&self) -> Self {
+		Self {
+			offset: self.offset,
+			token: self.token.clone(),
+			is_leading: self.is_leading,
+		}
+	}
+}
```
MichaReiser, line 166: Would it be possible to derive the `Clone` implementation? Or what's preventing it?

crates/rome_rowan/src/cursor.rs
```diff
@@ -543,6 +560,36 @@ impl NodeData {
...
+	}
+
+	pub(crate) fn text_len(&self) -> TextSize {
+		let g = self.token.green();
+		if self.is_leading {
+			g.leading().text_len()
+		} else {
+			g.trailing().text_len()
+		}
+	}
```
MichaReiser, line 590: Nit: I personally struggled to understand what `g` is referring to. How about calling it `token`, `green_token` or `green`?

crates/rome_rowan/src/cursor.rs
```diff
@@ -631,6 +678,29 @@ impl SyntaxNode {
...
+			.unwrap_or_else(|| 0.into());
+
+		let trailing_len = self
+			.last_token()
+			.map(|x| x.trailing().text_len())
+			.unwrap_or_else(|| 0.into());
+
+		let range = self.text_range();
+		TextRange::new(range.start() + leading_len, range.end() - trailing_len)
+	}
```
MichaReiser, line 694: Is it intentional, that `trim` only "trims" the first leading/trailing trivia, but not any other trivia?

crates/rome_rowan/src/api.rs
```diff
@@ -32,6 +33,28 @@ impl Language for RawLanguage {
...
+
+impl Trivia {
+	#[inline]
+	pub fn text_len(&self) -> TextSize {
+		match self {
+			Trivia::Whitespace(n) => (*n as u32).into(),
+			Trivia::Comments(n) => (*n as u32).into(),
+		}
+	}
+}
```
MichaReiser, line 50: Where is this used today? 

I also find the name somewhat confusing, how's `Trivia` different from `SyntaxTrivia`. My understanding is that `SyntaxTrivia` is a collection of `Trivia`s. Swift uses the term `SyntraxTriviaPiece` for the items in the collection and I kind of like that terminology. Wdyt?

MichaReiser (changes requested):
That looks neat! I especially love how you paid attention to avoiding allocating any `String` to get the text and trimmed text of nodes/trivia and tokens.

I think it would be good to document all added methods on the API as well as the newly introduced concepts. It took me a while to understand that `*Trivia` is a collection of `TriviaPiece`s . I think we should document
* what a trivia is
* what are the specific parsing rules
* what's a trimmed/normal text
* What does a trivia consist of (trivia pieces). We should probably have a follow-up PR exposing the pieces. 

Please add the additional documentation and I do think that this PR is then good to go.

However, I do believe that there's some work we have to follow up with. This doesn't have to be part of this PR.

* Expose trivia pieces and discuss the pieces we want to support. I'm still convinced that it would be valuable to have a new line trivia piece. Yes, we could add helpers to `SyntaxTriviaPiece` to further split it into new-line/white space but isn't that exactly what a trivia piece is?
* I believe it would be nice if we can have different implementations for each syntax trivia piece, so that the new line piece could expose helpers like `is_cariage_return`, `is_windows_line_ending`, etc. 
* I'm personally still not convinced that the way you store trivia in this PR is ideal in regards to total memory consumption. I brought it up in your initial PR but, unfortunately, didn't receive any feedback on it. Each trivia is 16 bytes. That means, we pay an overhead of 32 bytes even if a token doesn't have any trivia at all. Furthermore, most trivia is repeated frequently and storing it multiple times significantly increases memory consumption. The storage layout also differs significantly from how rowan stores nodes/tokens, which, in my view, introduces complexity by introducing a new concept and the layout has limited extensibility (separating new line and whitespace pieces may prevent us from avoiding the heap allocation for a very common use case for leading trivia).  That's why I believe it would be better to heap allocate trivia and only store the leading/trailing pointers in the token. This brings down the token overhead to 16 bytes (8 bytes each), reducing the overhead by 50% for tokens without trivia. Furthermore, it allows us to cache trivia across tokens which should significantly bring down the total amount of heap allocations and the total memory consumption per file and across files. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=273d8d140efae75bb503779f1aa5bece) for an example on how this could be implemented.

crates/rome_rowan/src/api.rs
```diff
@@ -114,6 +137,16 @@ impl<L: Language> From<SyntaxToken<L>> for SyntaxElement<L> {
 	}
 }
 
+impl<L: Language> SyntaxTrivia<L> {
+	pub fn text(&self) -> &str {
+		self.raw.text()
+	}
```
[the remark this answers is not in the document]
xunilrj: > Something that might be nice (not required for this PR) is to have different trivia implementations per kind so that we could add additional methods, for example, is_cariage_return for new lines.

It does seem like a good idea. Not sure is going to be super ergonomic though, given that each implementation needs to be behind an enum.

> Should we expose a method to get the trivia kind?

Yes.

But I agree that these can be in future PRs.

crates/rome_rowan/src/api.rs
xunilrj: I like the "Piece" suffix. The trivia enum is shared between the Syntax* and the Green* tree, so I prefer to keep it without any prefix.

We can have
- TriviePiece
- SyntaxTrivia
- GreenTrivia

I think it is consistent with other Syntax* and Gree* structs.

crates/rome_rowan/src/api.rs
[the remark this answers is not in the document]
MichaReiser: Yes, this would probably need to be behind an enum. The enum could implement the logic that is useful for all pieces .

```rust
pub enum SyntaxTriviaPiece {
  Whitespace(SyntaxWhitespaceTriviaPiece),
  NewLine(SyntaxNewLineTriviaPiece),
  ...
}

impl SyntaxTriviaPiece{
  pub fn text(&self) -> ... 
}
```

crates/rome_rowan/src/api.rs
MichaReiser: That works if the API doesn't need to hold any additional data. How do you plan to implement a `text` method on a trivia piece? I think that's something that we'll need sooner or later in the API but probably isn't something we need on the Green layer (because it requires keeping a reference to the `Token` that stores the text).

crates/rome_rowan/src/api.rs
```diff
@@ -135,10 +168,32 @@ impl<L: Language> SyntaxNode<L> {
 		self.raw.text_range()
 	}
 
+	pub fn text_trimmed_range(&self) -> TextRange {
```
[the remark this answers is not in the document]
xunilrj: hum... I think makes more sense to move some of the unit tests to doc tests... will do it!

xunilrj:
> I'm personally still not convinced that the way you store trivia in this PR is ideal in regards to total memory consumption... 

Let me try to come up with a way to test/measure, so we can understand which solution would be the best.

MichaReiser:
> Let me try to come up with a way to test/measure, so we can understand which solution would be the best.

Feel free to do this after merging this PR and move the data structure decision into a discussion.

crates/rome_rowan/src/api.rs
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_rowan/src/api.rs
```diff
@@ -135,10 +168,32 @@ impl<L: Language> SyntaxNode<L> {
...
+	}
+
+	pub fn leading(&self) -> Option<SyntaxTrivia<L>> {
+		self.raw.leading().map(|raw| SyntaxTrivia {
+			raw,
+			_p: PhantomData,
+		})
+	}
+
+	pub fn trailing(&self) -> Option<SyntaxTrivia<L>> {
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Done.

crates/rome_rowan/src/cursor.rs
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Will do this in another PR.

crates/rome_rowan/src/api.rs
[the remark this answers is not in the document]
xunilrj: Will do this in another PR.

crates/rome_rowan/src/cursor.rs
```diff
@@ -1097,6 +1197,37 @@ impl SyntaxElement {
...
 // region: impls
 
+impl PartialEq for SyntaxTrivia {
+	#[inline]
+	fn eq(&self, other: &Self) -> bool {
+		self.token == other.token && self.is_leading == other.is_leading
+	}
+}
+
+impl Eq for SyntaxTrivia {}
```
MichaReiser, line 1207: Isn't it possible to use `#[derive(Eq)]` if you manually implement `PartialEq`?

crates/rome_rowan/src/green/token.rs
```diff
@@ -95,30 +152,95 @@ impl GreenTokenData {
...
+	pub fn text_leading(&self) -> &str {
+		let leading_len = self.leading().text_len();
+
+		let end: usize = leading_len.into();
+		let text = unsafe { std::str::from_utf8_unchecked(self.data.slice()) };
+		&text[0..end]
+	}
+
+	#[inline]
+	pub fn text_trailing(&self) -> &str {
```
MichaReiser, line 189: Nit: `text_leading_trivia` and `text_trailing_trivia` to match the documentation of `text_trimmed` (you mentioned that it excludes all `trivia`, not `leading and `trailing`)

crates/rome_rowan/src/green/token.rs
```diff
@@ -95,30 +152,95 @@ impl GreenTokenData {
...
 	#[inline]
 	pub fn text_len(&self) -> TextSize {
 		TextSize::of(self.text())
 	}
+
+	#[inline]
+	#[allow(dead_code)]
+	pub fn text_trimmed_len(&self) -> TextSize {
+		TextSize::of(self.text_trimmed())
+	}
```
MichaReiser, line 207: Nit: Delete?

crates/rome_rowan/src/green/token.rs
```diff
@@ -95,30 +152,95 @@ impl GreenTokenData {
...
+	#[inline]
+	pub fn trailing(&self) -> &GreenTokenTrivia {
+		&self.data.header.trailing
+	}
 }
 
 impl GreenToken {
-	/// Creates new Token.
 	#[inline]
+	#[allow(dead_code)]
```
MichaReiser, line 222: Delete?

crates/rome_rowan/src/api.rs
```diff
@@ -358,6 +655,27 @@ impl<L: Language> SyntaxElement<L> {
...
+
+	pub fn leading(&self) -> Option<SyntaxTrivia<L>> {
+		match self {
+			NodeOrToken::Node(it) => it.first_leading_trivia(),
+			NodeOrToken::Token(it) => Some(it.leading_trivia()),
+		}
+	}
+
+	pub fn trailing(&self) -> Option<SyntaxTrivia<L>> {
+		match self {
```
MichaReiser, line 673: Nit: `leading` and `trailing` trivia to align it with the methods on `SyntaxNode`?

crates/rome_rowan/src/api.rs
```diff
@@ -32,6 +33,34 @@ impl Language for RawLanguage {
...
+		}
+	}
+}
+
+pub enum SyntaxTriviaPiece {
+	Whitespace {},
+	NewLine {},
+	Comments {},
+}
+
```
MichaReiser, line 57: `SyntaxTriviaPiece` doesn't seem to be used at the moment and there's the `Trivia`. 

We should only have a single representation for a `Trivia` piece in the API (including the builder). Meaning, that the `TreeBuilder` should accept `SyntaxTriviaPiece` if that's what the `SyntaxToken` API returns.

crates/rome_rowan/src/api.rs
```diff
@@ -32,6 +33,34 @@ impl Language for RawLanguage {
...
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct SyntaxTrivia<L: Language> {
```
MichaReiser, line 59: Would you mind adding some documentation explaining what `SyntaxTrivia` is, what the rules are for `leading` and `trailing` trivia and what the trivia pieces are?

MichaReiser (approved):
Looks good to me but would be nice if someone else also has a look at it, maybe @yassere ? 

Thanks for addressing all the feedback and excellent documentation. This is a very neat PR!

crates/rome_rowan/src/cursor.rs
xunilrj: I think I can derive everything, actually.
The manual impls came from control+c, control+v from other Syntax*.

crates/rome_rowan/src/green/token.rs
xunilrj: Done.

crates/rome_rowan/src/green/token.rs
xunilrj: The attribute or the method? The method is still not being used, I created it because we also have a text_len().
I will remove it, just to clean things.

xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Removing this. I will create another PR for the TriviaPiece.

crates/rome_rowan/src/api.rs
xunilrj: Mind if I create this in another PR?

crates/rome_rowan/src/green/token.rs
```diff
@@ -8,14 +8,69 @@ use std::{
...
+					GreenTokenTrivia::Many(Box::new(vec![Trivia::Comments(len), second]))
+				}
+				(GreenTokenTrivia::Many(mut v), second) => {
+					v.push(second);
+					GreenTokenTrivia::Many(v)
+				}
+			}
+		}
+		trivia
+	}
```
yassere, line 66: I think we should avoid pushing onto a mutable vector (which can cause reallocations) when we already know how many elements there will be. This seems like a good candidate for slice patterns. Maybe something like:
```rs
fn from(trivias: &[Trivia]) -> Self {
    match trivias {
        [] => GreenTokenTrivia::None,
        [Trivia::Whitespace(len)] => GreenTokenTrivia::Whitespace(*len),
        [Trivia::Comments(len)] => GreenTokenTrivia::Comments(*len),
        many => GreenTokenTrivia::Many(Box::new(Vec::from(many))),
    }
}
```

crates/rome_rowan/src/api.rs
```diff
@@ -32,6 +33,34 @@ impl Language for RawLanguage {
 	}
 }
 
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub enum Trivia {
+	Whitespace(usize),
+	Comments(usize),
+}
```
yassere, line 40: Should we derive a `Copy` implementation here too?

crates/rome_rowan/src/api.rs
```diff
@@ -131,12 +170,186 @@ impl<L: Language> SyntaxNode<L> {
...
+	///     );
+	/// });
+	/// assert_eq!("\n\t let \t\ta; \t\t", node.text());
+	/// ```
+	pub fn text(&self) -> SyntaxText {
+		self.raw.text()
+	}
+
+	/// Returns the text of all descendants tokens combined,
+	/// excluding the fist token leading trivia, and the last token trailing trivia.
```
yassere, line 200: fist -> first

crates/rome_rowan/src/api.rs
```diff
@@ -131,12 +170,186 @@ impl<L: Language> SyntaxNode<L> {
...
+	/// assert_eq!(14u32, range.end().into());
+	/// ```
 	pub fn text_range(&self) -> TextRange {
 		self.raw.text_range()
 	}
 
-	pub fn text(&self) -> SyntaxText {
-		self.raw.text()
+	/// Returns the range corresponding for the text of all descendants tokens combined,
+	/// excluding the fist token leading  trivia, and the last token trailing trivia.
```
yassere, line 256: fist -> first

yassere (approved):
Left a couple suggestions, but this looks good to me. We can discuss more about optimizing memory/performance in the future when we start benchmarking.

crates/rome_rowan/src/api.rs
xunilrj: hum... We can.

crates/rome_rowan/src/green/token.rs
xunilrj: Done. I decided to also move the vector. I think it will make more sense in the parser PR.

crates/rome_rowan/src/api.rs
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Done.

crates/rome_rowan/src/green/token.rs
xunilrj: This is actually used in tests.

PR #1784, opened 2021-11-15
Unify `JsModule` and `JsScript` into `JsRoot` and parse directives

MichaReiser:
## Summary

This PR unifies the `JsModule` and `JsScript` into the `JsRoot` and adds the source type to the `Parse` result (if it's a module or a script).

The main reason for doing so is that we decided to use the exact same grammar for both scripts and modules (we're not preventing import statements inside of scripts). Mainly, to not have to parse module syntax as unknown statements inside of scripts (flagging them as errors should be sufficient).

Unifying modules and scripts further simplify the API usage because there's now a unified root type.

This PR further refines the parsing of directives and adds directives to our grammar by introducing a new `FunctionBody` node for blocks that are direct children of any function declaration (and, therefore, allow directives).

Part of #1725 

## Test Plan

Verified that the coverage isn't decreasing. Actually, the change fixes an panic when parsing `built-ins\Number\isFinite\finite-numbers.js`


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -454,37 +468,84 @@ pub fn block_stmt(
...
+		p.bump_any();
+
+		// eat semicolon if present, correct termination guaranteed by is_directive
+		p.eat(SyntaxKind::SEMICOLON);
+
+		directive.complete(p, JS_DIRECTIVE);
+
+		let directive_text = p.token_src(&directive_token);
+
+		if &directive_text[1..directive_text.len() - 1] == "use strict" {
```
xunilrj, line 523: This is just to skip the quotes?

Because just testing the whole thing, is probably going to be faster and easier to understand.

I did a quick test using https://github.com/bheisler/iai

```rust

fn m1() {
    let _ = black_box("\"dasdasdasdasdadas\"") == black_box("\"use strict\"");
}

fn m2() {
    let s = "\"dasdasdasdasdadas\"";
    let _ = black_box(&s[1..s.len()-1]) == black_box("use strict");
}

```

```
m1
  Instructions:                  12
  L1 Accesses:                   18
  L2 Accesses:                    0
  RAM Accesses:                   1
  Estimated Cycles:              53

m2
  Instructions:                  12
  L1 Accesses:                   17
  L2 Accesses:                    1
  RAM Accesses:                   1
  Estimated Cycles:              57
```

crates/rslint_parser/test_data/inline/ok/directives.js
```diff
@@ -0,0 +1,35 @@
...
+
+	"use strict"; // not a directive
+});
+
+let b = () => {
+	"use strict";
+
+	let a = 10;
+
+	"use strict";
```
xunilrj, line 30: This one is a directive or are we just missing the comments?

xunilrj (approved):

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: I agree that it improves readability. The issue is that we need to test for `"use strict"` and `'use strict'`, but I guess the compiler might be able to optimize this as well.

crates/rslint_parser/test_data/inline/ok/directives.js
MichaReiser: It's missing the comment. let me add it

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -431,14 +430,29 @@ pub fn empty_stmt(p: &mut Parser) -> CompletedMarker {
 	m.complete(p, JS_EMPTY_STATEMENT)
 }
 
+pub(crate) fn function_body(
```
ematipico, line 433: Small comment to explain what this function is for?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -454,37 +468,84 @@ pub fn block_stmt(
...
+		let directive_text = p.token_src(&directive_token);
+
+		if directive_text == "\"use strict\"" || directive_text == "'use strict'" {
+			if old_state == None {
+				old_state = Some(p.state.clone());
+			}
+
+			let mut new_state = p.state.clone();
+			new_state.strict(p, directive_token.range);
+			p.state = new_state;
```
ematipico, line 530: What are we doing here?

ematipico (commented):
One question, I guess I missed that information: when are we going to add the space after the labels inside ungrammar? `body:JsBody` => `body: JsBody`

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: The parser tracks whatever it is inside of a "strict" code block or not. This logic sets the required flags inside of the parser and we return the old state so that the caller can restore it once it's done parsing.

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Got it! I wonder if adding a `parse_` prefix to these kind of function might help. This was something we had inside RomeJS.

PR #1786, opened 2021-11-15
Function Expression/Declaration AST Facade

MichaReiser:
## Summary

Updates the facade for function declarations and function expression to match the grammar defined in #1719. 

* Function declaration
  * Rename from `FnDecl` to `JsFunctionDeclaration`
  * Rename `ParameterList` to `JsParameterList`
  * Introduce new `JsIdentifierBinding` type for `id`
  * Parses `...rest` as `JS_REST_PARAMETER` instead of `JS_REST_PATTERN`.
  * Split out `TsReturnType` for the `:` and the return type
* Function expression
  * Rename from `FnExpr` to `JsFunctionExpression`
  * Otherwise the same as for declaration
* Arrow function expression
  * Rename from `ArrowExpr` to `JsArrowFunctionExpression`
  * Rename `ExprOrBlock` to `JsAnyArrowFunctionBody`
  * Otherwise the same as for declaration.

This PR cleans up some of the function parsing (extracts helpers rather than using boolean params). This improves error reporting for the `func_decl` use case if the name is missing. 

This PR also removes the support for decorators. The parser had some support for decorators but they haven't been part of the grammar. We should re-add them when we decide to properly support them (and add tests)

Part of #1725 

## Test Plan

Verified that the coverage isn't changing.


crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -81,85 +47,6 @@ fn class_prop_name(p: &mut Parser) -> Option<(CompletedMarker, Option<usize>)> {
...
-			)
-			.primary(guard.cur_tok().range, "");
-
-		guard.error(err);
-	}
-	args_body(guard);
-	m.complete(guard, FN_DECL)
-}
-
 #[allow(clippy::unnecessary_unwrap)]
```
MichaReiser, line 50: I merely moved these functions to `function.rs`. I extracted smaller functions from `args_body`, splitted `func_decl` into `function_declaration` and `function_expression`, and renamed `fn_body` to `function_body_or_declaration`.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -915,18 +914,7 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
...
-				p.bump_remap(T![async]);
-				let mut complete = function_decl(
-					&mut *p.with_state(ParserState {
-						in_async: true,
-						..p.state.clone()
-					}),
-					m,
-					true,
-				);
-				complete.change_kind(p, FN_EXPR);
```
MichaReiser, line 928: We repetitively parsed the `async` keyword. I decided to move the consumption of the `async` keyword into `function_expression`/`function_declaration`. This fixes multiple issues where the `async` identifier wasn't correctly re-mapped to the `async` keyword.

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -0,0 +1,136 @@
...
+
+pub(super) fn function_body(p: &mut Parser) -> Option<CompletedMarker> {
+	block_impl(p, JS_FUNCTION_BODY, None)
+}
+
+// TODO 1725 This is probably not ideal (same with the `declare` keyword). We should
+// use a different AST type for function declarations. For example, a function declaration should
+// never have a body but that would be allowed with this approach. Same for interfaces, interface
+// methods should never have a body
+/// Either parses a typescript declaration body or the function body
```
MichaReiser, line 91: I only copied this method which is why I didn't address the `declare` problem.

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -0,0 +1,136 @@
...
+
+fn function(p: &mut Parser, kind: SyntaxKind) -> CompletedMarker {
+	let m = p.start();
+
+	if kind == JS_FUNCTION_DECLARATION {
+		// TS function declaration
+		p.eat(T![declare]);
+	}
+
+	let in_async = p.at(T![ident]) && p.cur_src() == "async";
```
xunilrj, line 40: Do you think makes sense to have eat_if and bump_remap_if methods?

xunilrj (approved):

crates/rslint_parser/src/syntax/function.rs
MichaReiser: I personally like the explicit control flow of the `if`, I find this clearer than too inline condition. However, something like that may come in handy when I work on signalling the "missing" slots but I don't know exactly how the API would look like. 

That's why I prefer to defer to delay adding any such methods until I have a better understanding of the requirements.

crates/rslint_parser/src/syntax.rs
```diff
@@ -9,6 +9,7 @@
 
 pub mod decl;
 pub mod expr;
+mod function;
```
ematipico, line 12: Why isn't this `pub` like the others?

crates/rslint_parser/src/syntax/expr.rs
ematipico: Are there new tests for this new logic? I get we fixed some bugs, but are they reflected inside some test?

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	m.complete(guard, kind)
+}
+
+pub(super) fn function_body(p: &mut Parser) -> Option<CompletedMarker> {
+	block_impl(p, JS_FUNCTION_BODY, None)
+}
+
+// TODO 1725 This is probably not ideal (same with the `declare` keyword). We should
```
ematipico, line 87: If you add a hash `#` before the issue number (`#1725`), it integrates with VSCode/Idea and it will show the issue inside the editor

crates/rslint_parser/src/syntax.rs
MichaReiser: The parsing logic should be internal and not public. That also helps to detect dead code (there's plenty). Ideally, all the other modules should be private too

crates/rslint_parser/src/syntax/function.rs
MichaReiser: Uh neat!

PR #1787, opened 2021-11-16
Restructure Object AST Tree strcuture

MichaReiser:
## Summary

This PR restructures the AST structure of object expressions and related nodes (members, etc.) and is part of #1725.

* Renames `ObjectExpr` to `JsObjectExpression`
* Renames `props` to `members`
* Renames the `key` of all members to `name`
* Renames `IdentProp` to `JsShorthandPropertyObjectMember` that directly stores a `ident` token
* Renames `LiteralProp` to `JsPropertyMember` and changes the `key` from `Name` to `JsStaticObjectMemberName`
* Introduces `JsGetterObjectMember` (unclear if it should be shared with classes in the future)
* Introduces `JsSetterObjectMember` (unclear if it should be shared with classes in the future)
* Introduces `JsMethodObjectMember` (unclear if it should be shared with classes in the future)
* Renames `SpreadProp` to `JsSpread`

The PR fixes some issues when it comes to `setter/getter` parsing.

The main concern around sharing `Js*Member`s with classes is that objects are more restrictive when it comes to naming because classes support private properties which object members don't. I think the better approach is to introduce a `JsAnyGetter = JsGetterObjectMember | JsGetterClassMember` union that implements helpers to retrieve the shared properties.

## Test Plan

Verified that coverage doesn't drop. Ran tests manually. 

The PR fixes the following coverage tests (all containing getters with literal names):

```
Test 'built-ins\Array\prototype\reverse\length-exceeding-integer-limit-with-object.js' incorrectly threw an error
Test 'built-ins\Array\prototype\reverse\length-exceeding-integer-limit-with-proxy.js' incorrectly threw an error
Test 'built-ins\Array\prototype\unshift\length-near-integer-limit.js' incorrectly threw an error
Test 'built-ins\Function\prototype\apply\get-index-abrupt.js' incorrectly threw an error
```


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1157,150 +1158,6 @@ pub fn spread_element(p: &mut Parser) -> CompletedMarker {
...
-			if p.at(T!['}']) {
-				break;
-			}
-		}
-		object_property(p);
-	}
-	props_list.complete(p, LIST);
-
-	p.expect(T!['}']);
-	m.complete(p, OBJECT_EXPR)
```
MichaReiser, line 1184: I moved all the `object` related parsing into its own `syntax/object` file. I did cleanup `object_property` and copied `method` from `syntax/decl` because the structure differs for the time being (and may as well in the future).

xtask/js.ungram
```diff
@@ -480,6 +477,109 @@ JsPostUpdateExpression =
     operator:('++' | '--')
 
 
+///////////////
+// OBJECTS
+///////////////
+
+// { a: ... }
+//   ^
```
ematipico, line 485: I like these comments, really helpful!

ematipico (approved):

crates/rome_formatter/src/ts/object_members/object_member_name.rs
```diff
@@ -0,0 +1,22 @@
+use rslint_parser::ast::{JsAnyObjectMemberName, JsStaticObjectMemberName};
+
+use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
+
+impl ToFormatElement for JsAnyObjectMemberName {
```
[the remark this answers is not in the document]
MichaReiser: I moved the static object member implementation to its own file

MichaReiser:
> 
> ### Test262 comparison coverage results on ubuntu-latest
> Total 	17608 	17608 	0
> Passed 	16767 	16771 	✅ ⏫ **+4**
> Failed 	840 	836 	✅ ⏬ **-4**
> Panics 	1 	1 	0
> Coverage 	95.22% 	95.25% 	**+0.02%**
> Fixed tests (4):
> 
> ```
> xtask/src/coverage/test262/test/built-ins/Array/prototype/unshift/length-near-integer-limit.js
> xtask/src/coverage/test262/test/built-ins/Array/prototype/reverse/length-exceeding-integer-limit-with-object.js
> xtask/src/coverage/test262/test/built-ins/Array/prototype/reverse/length-exceeding-integer-limit-with-proxy.js
> xtask/src/coverage/test262/test/built-ins/Function/prototype/apply/get-index-abrupt.js
> ```

That's amazing @ematipico. Thanks for working on the bot integration.

PR #1788, opened 2021-11-16
feat: pass error kind via parameter

ematipico:
## Summary

First PR that implements #1759 

This change is very simple, every time we need to mark error inside the tree, we explicitly pass the `SyntaxKind::ERROR` as parameter instead of having it hardcoded. 

This change will make things much easier and smoother when applying specific errors during the parsing phase.

## Test Plan

Existing CI

### DIFF
--- a/crates/rslint_parser/src/parser.rs
+++ b/crates/rslint_parser/src/parser.rs
@@ -185,6 +185,8 @@ impl<'t> Parser<'t> {
 		error: impl Into<ParserError>,
 		recovery: TokenSet,
 		include_braces: bool,
+		// The kind of unknown token the parser wanted to try to guess at that position
+		unknown_syntax_kind: SyntaxKind,
 	) -> Option<()> {
 		if self.state.no_recovery {
 			return None;
@@ -206,12 +208,17 @@ impl<'t> Parser<'t> {
 		let m = self.start();
 		self.error(error);
 		self.bump_any();
-		m.complete(self, SyntaxKind::ERROR);
+		m.complete(self, unknown_syntax_kind);
 		Some(())
 	}
 
 	/// Recover from an error but don't add an error to the events
-	pub fn err_recover_no_err(&mut self, recovery: TokenSet, include_braces: bool) {
+	pub fn err_recover_no_err(
+		&mut self,
+		recovery: TokenSet,
+		include_braces: bool,
+		unknown_syntax_kind: SyntaxKind,
+	) {
 		match self.cur() {
 			T!['{'] | T!['}'] if include_braces => {
 				return;
@@ -225,7 +232,7 @@ impl<'t> Parser<'t> {
 
 		let m = self.start();
 		self.bump_any();
-		m.complete(self, SyntaxKind::ERROR);
+		m.complete(self, unknown_syntax_kind);
 	}
 
 	/// Starts a new node in the syntax tree. All nodes and tokens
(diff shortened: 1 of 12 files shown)


crates/rslint_parser/src/parser.rs
```diff
@@ -185,6 +185,8 @@ impl<'t> Parser<'t> {
 		error: impl Into<ParserError>,
 		recovery: TokenSet,
 		include_braces: bool,
+		// The kind of unknown token the parser wanted to try to guess at that position
+		unknown_syntax_kind: SyntaxKind,
```
MichaReiser, line 189: I find this comment a bit hard to understand. One reason is that `Unknown` kinds can never be tokens but always must be nodes (they group a set of tokens/nodes). I'm also not sure if `guess` is the right word here. My understanding of the method is:

* It tries to recover from the error by seeing if any token in the recovery set OR `{` and `}` are present
* If that's the case, emit a parse error but don't add any node 
* Otherwise, wrap the next token in an error node

So it's mainly a way to deal with unexpected tokens. I believe the caller should precisely know what kind of `Unknown` node we expect (which is why you pass it). 

Maybe something like this:

```
The kind of the unknown node the parser inserts if it isn't able to recover because the current token is neither in the recovery set nor any of `{` or `}`.
```

We may want to rename that method if we fully understand its use case because I had no idea what it's doing by just reading its name. 

Nit: Should this comment be part of the method documentation?

Future: The method starts to have many parameters. We may want to think of an alternative API, especially if we need to add more parameters.

crates/rslint_parser/src/parser.rs
```diff
@@ -206,12 +208,17 @@ impl<'t> Parser<'t> {
...
 		Some(())
 	}
 
 	/// Recover from an error but don't add an error to the events
-	pub fn err_recover_no_err(&mut self, recovery: TokenSet, include_braces: bool) {
+	pub fn err_recover_no_err(
+		&mut self,
+		recovery: TokenSet,
+		include_braces: bool,
+		unknown_syntax_kind: SyntaxKind,
```
MichaReiser, line 220: How about we add the same comment for `unknown_syntax_kind` as you added to `err_recover`?

MichaReiser (approved):

crates/rslint_parser/src/parser.rs
ematipico: > So it's mainly a way to deal with unexpected tokens. I believe the caller should precisely know what kind of Unknown node we expect (which is why you pass it).

What do you think if changed the API in this way:
1. call the parameter `expected_node`;
2. in case the parser expects a `JS_STATEMENT`, we then pass `SyntaxKind::JS_STATEMENT`;
3. the API will be called like `p.err_recover(err, T!['{'], true, JS_STATEMENT)`
3. `err_recover` will map `JS_STATEMENT` to `JS_UNKNOWN_STATEMENT`  using a `match` into a variable called `unresolved_node` or `unknown_node`

Would that make sense?

> We may want to rename that method if we fully understand its use case because I had no idea what it's doing by just reading its name.

Yes, I plan to rename the methods in order to better fit what we are trying to do. But I didn't want to put too much stuff and I wanted to lay ground for the future PRs.

> Future: The method starts to have many parameters. We may want to think of an alternative API, especially if we need to add more parameters.

This is something I already had in mind. Thank you!

MichaReiser: That would be neat but there are no kinds of unions like expression and statements.

PR #1791, opened 2021-11-17
Refactor Class AST Tree Structure

MichaReiser:
## Summary

This PR refactors the AST structures for class declarations and class expressions as defined in #1725. 

* Class declaration & Expression
  * Rename from `ClassDecl`/`ClassExpr` to `JsClassDeclaration`/`JsClassExpression`
  * Renamed `name` to `id`
  * Introduced a `JsExtendsClause` that wraps the `extends` keyword and the parent class
  * Renamed `body` to `members` and removed the `ClassBody` node
* Constructor
  * Rename from `Constructor` to `JsConstructorClassMember` 
  * Rename `name` to `id`
  * Parse "constructor" and 'constructor' methods as `JsConstructorClassMember`s
* Method
  * Renamed `Method` to `JsMethodClassMember`
* Properties
  * Renamed `ClassProp` to `JsPropertyClassMember`
* Getter
  * Renamed `Getter` to `JsGetterClassMember`
  * Removed the parameter list (getters can't have parameters)
* Setter
  * Renamed `Setter` to `JsSetterClassMember`
  * Replaced the parameter list with a single `parameter` 
* Private properties
  * Deleted the `PrivateProp` and instead introduce a `JsPrivatePropertyClassMemberName` type since all members can be private (except constructors)
* Replaced `JsEmptyStatement` with `JsEmptyMember` (these aren't proper statements)
* TS Accessibility: Rename to access modifier to match typescripts class documentation
* Rename `JsComputedObjectMemberName` and `JsStaticObjectMemberName` to `JsComputedMemberName` and `JsLiteralMember` because they can be shared between classes and objects
* Rename `TsReturnType` to `TsTypeAnnotation` so that it can also be used for properties, parameters, etc.  

This PR also adds a set of new tests for classes to verify the member parsing. I did so because I had to refactor much of the parsing logic to make sense of it (and reduce some repetitive code and fix some parsing issues related to semicolons). 

I further extracted the class parsing from the `decl.rs` into its own `class.rs`. You can see my refactorings if you look at the individual commits.

## Test Plan

Added new tests, verified that coverage isn't changing.

## Disclaimer

I didn't verify if the typescript parsing is working correctly or if the typescript grammar is correct. We need to tackle typescript support separately.


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -0,0 +1,854 @@
+use crate::state::StateGuard;
```
MichaReiser, line 1: Extracted from `decl`. You can review the individual commits if you want to see what I've refactored (quiet a bit)

crates/rome_formatter/src/ts/class/class_declaration.rs
```diff
@@ -0,0 +1,39 @@
+use crate::{
```
MichaReiser, line 1: Hmm, sorry. Don't know why git didn't recognise my renames of the formatter files, maybe because I rebased the commits?

crates/rome_formatter/tests/specs/js/class.js
```diff
@@ -24,5 +24,5 @@ class Foo extends Boar {
 
 	new_prop = 5
 
-	second_prop = [5,3,4]
+	second_prop = [5,3,4];;
```
yassere, line 27: Is this intentionally testing the extra semicolon?

MichaReiser: Yes, should I add a comment?

yassere: Sure, a comment would be helpful. Or maybe naming it something like `double_semicolon`.

crates/rslint_parser/test_data/inline/ok/getter_class_member.rast
```diff
@@ -0,0 +1,181 @@
...
+        IDENT@6..13 "Getters"
+      WHITESPACE@13..14 " "
+      L_CURLY@14..15 "{"
+      WHITESPACE@15..17 "\n\t"
+      LIST@17..140
+        JS_GETTER_CLASS_MEMBER@17..29
+          GET_KW@17..20 "get"
+          WHITESPACE@20..21 " "
+          JS_STATIC_MEMBER_NAME@21..24
+            IDENT@21..24 "foo"
```
yassere, line 17: The naming of the `JS_STATIC_MEMBER_NAME` node feels a bit confusing since it doesn't have anything to do with `static` members. Does the "static" in this case just mean "literal"?

MichaReiser: No, Literals are represented as `JsStringLiteral` or `JsNumberLiteral`. It's an identifier. Static as it's the opposite of computed. 

How would you feel about `JsIdentifierName` or do you have another idea?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -0,0 +1,856 @@
...
+	}
+
+	// test static_method
+	// class foo {
+	//  static foo(bar) {}
+	//  static *foo() {}
+	//  static async foo() {}
+	//  static async *foo() {}
+	// }
+	let has_access_modifier = matches!(p.cur_src(), "public" | "private" | "protected");
```
ematipico, line 216: there's `"readonly"` too

crates/rslint_parser/test_data/inline/ok/getter_class_member.rast
yassere: Yeah, I meant literal as in `LiteralPropertyName` from the [spec](https://tc39.es/ecma262/#prod-PropertyName) which includes identifiers, string literals, and numeric literals.

Would we want a `LiteralMemberName` based on the member/property naming we've chosen?
Apparently `class Foo { 2 = "two" }` is valid?

MichaReiser: Hehe yep, that's something I learned today too :D 

`LiteralMemberName` would only add another level of indirection (and I would then prefer `static` as the opposite of `computed). Because its definition must be an union and unions should either be over tokens or nodes. In this case, it must be over nodes because literals are nodes. 

```
LiteralMemberName = JsStringLiteral | JsNumberLiteral | JsIdentifierMemberName?
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -65,6 +63,24 @@ pub fn semi(p: &mut Parser, err_range: Range<usize>) {
 	}
 }
 
+/// Eats a semicolon or verifies that the current position allows for an automatic semicolon insertion
+/// Returns false if neither a semicolon was present nor does the current position allow for an automatic
+/// semicolon insertion
+pub(crate) fn optional_semi(p: &mut Parser) -> bool {
```
ematipico, line 69: The name of the function doesn't reflect the description and what it does. Maybe `has_semi_or_allows_semi`?

ematipico (approved):
Nothing major popped into my eye! I am glad that we decoupled the parsing logic of classes into a separate file!

xtask/js.ungram
```diff
@@ -579,6 +505,136 @@ PrivatePropAccess = lhs:JsAnyExpression '.' rhs:PrivateName
...
+	TsConstructorParam
+	| Pattern
+
+// class Test { a; b = 1; static c; }
+//              ^^ ^^^^^^ ^^^^^^^^^
+JsPropertyClassMember =
+	'declare'?
+	access_modifier: TsAccessibility?
+	'abstract'?
+	'static'?
```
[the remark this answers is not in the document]
MichaReiser: That's correct but encoding this correctly could be difficult

crates/rslint_parser/src/syntax/class.rs
MichaReiser: I think that's handled differently and is TS specific (which I'm neither extending nor removing).

ematipico: Thank you!

crates/rslint_parser/test_data/inline/ok/getter_class_member.rast
MichaReiser: I changed the `JsStaticMemberName` to 

```
JsStaticMemberName = name: ('string' | 'number' | 'ident') 
```

and updated the parsing. We now have `JsAnyClassMemberName = JsStaticMemberName | JsComputedMemberName | JsPrivateMemberName`. 

I hope that helps.

xtask/js.ungram
```diff
@@ -579,6 +505,136 @@ PrivatePropAccess = lhs:JsAnyExpression '.' rhs:PrivateName
...
+JsConstructorParameterList =
+	'('
+	parameters: (JsAnyConstructorParameter (',' JsAnyConstructorParameter)* ','?)
+	')'
+
+JsAnyConstructorParameter =
+	TsConstructorParam
+	| Pattern
+
+// class Test { a; b = 1; static c; }
```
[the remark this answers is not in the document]
MichaReiser: I don't think we need to list all possible cases (declare, access modifier, optional, !).

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: You're right. I updated the comment. That should fix the issue.

PR #1794, opened 2021-11-17
feat: unknown statements

ematipico:
## Summary

This PR does the following:

- created a new struct called `ParseRecover` where information the information needed for recovery are created passed to the parser;
- unified the functions `err_recover` and `err_recover_no_err` in one single function;
- the recovery logic is now encapsulated inside `ParseRecover` and the recovery logic is inside the `recover` function;
- created new test cases to track errors that were not tracked before;
- removed `ERROR` and added `JS_UNKNOWN_STATEMENT` in some particular case, specifically the parser has a function called [`stmt()`](https://github.com/rome/tools/blob/main/crates/rslint_parser/src/syntax/stmt.rs#L87) that is called in those functions that tries to parse particular statements: `if/else`, `do/while`, `while`, `for`, etc., so every time this function fails, we emit an error;

## Test Plan

```
cargo lint
cargo format
cargo test
```


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1023,8 +1032,13 @@ pub fn switch_stmt(p: &mut Parser) -> CompletedMarker {
...
 				temp.error(err);
 			} else {
-				first_default = Some(range);
+				// if the range is 0, it means that we found a "case"
+				if !range.is_empty() {
+					first_default = Some(range);
+				}
 			}
+		} else {
+			break;
```
ematipico, line 1041: Here we have an error, meaning that we haven't found `case` or `default`, so we need to `break` from the `while` loop otherwise we fall into a loop.

yassere:
What’s the advantage of combining these parameters into a `RecoveryBag` struct? It feels like it’s adding more boilerplate at the error recovery call sites in the parser.

crates/rslint_parser/src/recovery_bag.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct RecoveryBag {
+	/// The [Diagnostic] to emit
+	error: Option<ParserError>,
+	/// It tells the parser to recover if the position is inside a set [tokens](TokenSet)
+	recovery: TokenSet,
+	/// It tells the parser to recover if the current token is a curly brace
+	include_braces: bool,
+	/// The kind of the unknown node the parser inserts if it isn't able to recover because
+	/// the current token is neither in the recovery set nor any of `{` or `}`.
+	mysterious_node: SyntaxKind,
```
MichaReiser, line 15: That sounds mysterious...

How about unknown_node_kind?

crates/rslint_parser/src/recovery_bag.rs
```diff
@@ -0,0 +1,57 @@
+use crate::{Parser, ParserError, TokenSet};
+use rslint_errors::Diagnostic;
+use rslint_lexer::{SyntaxKind, T};
+
+/// This struct contains the information needed to the parser to recover from a certain error
+pub struct RecoveryBag {
```
MichaReiser, line 6: Nit: `RecoveryBag` isn't telling readers much about what it is. Is it a `RecoveryConfiguration`?

crates/rslint_parser/src/recovery_bag.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl RecoveryBag {
+	pub fn new(recovery: TokenSet, include_braces: bool, mysterious_node: SyntaxKind) -> Self {
+		Self {
+			error: None,
+			recovery,
+			include_braces,
+			mysterious_node,
+		}
+	}
```
MichaReiser, line 26: Nit: I find boolean flags hard to understand. Should we remove the `include_braces` and initialize it with the most common default and then expose a `with_braces_included` and `without_braces_included` functions to toggle the flag?

crates/rslint_parser/src/recovery_bag.rs
```diff
@@ -0,0 +1,57 @@
...
+			error: Some(error.into()),
+			recovery,
+			include_braces,
+			mysterious_node,
+		}
+	}
+
+	pub fn has_braces(&self, parser: &Parser) -> bool {
+		matches!(parser.cur(), T!['{'] | T!['}'] if self.include_braces)
+	}
```
MichaReiser, line 44: Nit: To align with the `is_at_token_set`: Rename to `is_at_braces`. 

We may not need both these operations and just exposing a `is_at_recovery_token` could be sufficient.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1023,8 +1032,13 @@ pub fn switch_stmt(p: &mut Parser) -> CompletedMarker {
 
 				temp.error(err);
 			} else {
-				first_default = Some(range);
+				// if the range is 0, it means that we found a "case"
+				if !range.is_empty() {
+					first_default = Some(range);
+				}
```
MichaReiser, line 1038: Nit: can the `if` / `else` be unified into a single `else if`?

```
} else if range.is_empty() {
  first_default = Some(range);
}
```

It may also improve readability if we rename `range` to `default_range` inside of the `if let Some(range) = switch_clause(...)` to make it clear what kind of range this is.

crates/rslint_parser/src/recovery_bag.rs
ematipico: Configuration doesn't attract me... What about a simple `Recovery`? Or `RecoveryKit`? `AidKit`? Trying to use some fantasy lol

crates/rslint_parser/src/recovery_bag.rs
ematipico: Yes, that should improve the usability of the API

ematipico:
> What’s the advantage of combining these parameters into a `RecoveryBag` struct? It feels like it’s adding more boilerplate at the error recovery call sites in the parser.

The parameters of the functions we use to recover from errors were becoming too many. Micha suggested to revisit it with a `struct` and I agree to. Also, it allows us to unify some logic in one single place

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -108,7 +108,7 @@ impl<'a> TreeSink for LosslessTreeSink<'a> {
 					it.kind,
 					self.text
 						.get(start..end)
-						.unwrap_or_else(|| self.text.get(start - 1..end).unwrap()),
+						.unwrap_or_else(|| self.text.get(start - 1..end).unwrap_or("")),
```
ematipico, line 111: This part of the code is still obscure to me, but when I swap `ERROR` with `JS_UNKNOWN_*`, it triggers the `unwrap` error. The `unwrap_or` fixes the panic.

crates/rslint_parser/src/recovery_bag.rs
xunilrj: ```ParseRecoverer```?

crates/rslint_parser/test_data/inline/err/switch_stmt_no_cases.rast
```diff
@@ -0,0 +1,59 @@
+JS_ROOT@0..34
+  LIST@0..0
```
xunilrj, line 2: This node with this range makes sense?

ematipico: Yeah [I asked the same thing on discord](https://discord.com/channels/678763474494423051/841927329265090590/910520914574209024).

It is intended, that's the list that contains the directives, e.g. `"use strict"`

crates/rslint_parser/test_data/inline/err/switch_stmt_no_cases.rast
```diff
@@ -0,0 +1,59 @@
...
+                IDENT@20..23 "bar"
+            WHITESPACE@23..24 " "
+            JS_EQUAL_VALUE_CLAUSE@24..30
+              EQ@24..25 "="
+              WHITESPACE@25..26 " "
+              JS_BOOLEAN_LITERAL@26..30
+                TRUE_KW@26..30 "true"
+      SEMICOLON@30..31 ";"
+    WHITESPACE@31..32 "\n"
+    JS_UNKNOWN_STATEMENT@32..33
```
xunilrj, line 32: hum... I would expect the ```*_UNKNOWN_*``` node in this case to surround the whole statement; but looking at its range, it seems its pointing only to new line after the semicolon.

ematipico: Me too! That's what I am trying to understand. Although, also TS parser highlights the same:
Screenshot 2021-11-17 at 15 52 07

crates/rslint_parser/src/lossless_tree_sink.rs
MichaReiser: @xunilrj do you know where this might be coming from?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -976,7 +989,10 @@ fn switch_clause(p: &mut Parser) -> Option<Range<usize>> {
 					"Expected the start to a case or default clause here",
 				);
 
-			p.err_recover(err, STMT_RECOVERY_SET, true, ERROR);
+			p.recover_on_unexpected_node(
```
MichaReiser, line 992: I think the best recovery here could be to insert a case clause where the unknown statement is the body. The recovery set should contain semicolon, }, new line and eof

This should also remove the need for the added break in the outer loop

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -66,7 +67,11 @@ pub fn semi(p: &mut Parser, err_range: Range<usize>) {
 }
 
 /// A generic statement such as a block, if, while, with, etc
-pub fn stmt(p: &mut Parser, recovery_set: impl Into<Option<TokenSet>>) -> Option<CompletedMarker> {
+pub fn stmt(
+	p: &mut Parser,
+	recovery_set: impl Into<Option<TokenSet>>,
+	unknown_node_on_error: SyntaxKind,
```
MichaReiser, line 73: I don't understand tha unknown parameter here. Shouldn't a non parsable statement always become an unknown statement. Because only unknown statements are allowed in places of a statement

crates/rslint_parser/src/parser.rs
```diff
@@ -181,72 +182,30 @@ impl<'t> Parser<'t> {
...
 			return;
 		}
 
 		let m = self.start();
+		if let Some(error) = error {
+			self.error(error);
+		}
 		self.bump_any();
-		m.complete(self, unknown_node);
+		m.complete(self, recovery_bag.get_unknown_node_kind());
```
MichaReiser, line 208: I believe we can reduce the number of branches for the error handling

```

if let Some(error) = error {
  self.error(error);
}

if !recovery_bag.parsing_is_recoverable(self) {
 	let m = self.start();  
	self.bump_any();
	m.complete(self, recovery_bag.get_unknown_node_kind());
}

File: crates/rslint_parser/src/recovery_bag.rs (hunk 3 above)
Comment (MichaReiser) [reply crates/rslint_parser/src/recovery_bag.rs:9]: Like that, or `ParseRecovery`

File: crates/rslint_parser/src/recovery_bag.rs [hunk 13]
```diff
@@ -0,0 +1,78 @@
...
+	pub fn without_braces_included(mut self) -> Self {
+		self.include_braces = false;
+		self
+	}
+
+	/// Enable check of curly braces as recovery tokens
+	pub fn with_braces_included(mut self) -> Self {
+		self.include_braces = true;
+		self
+	}
```
MichaReiser, line 54: Nit: It's a bit confusing that `with_error` is a factory function whereas these `with` functions are instance methods. 

How about `enable_braces_recover()` and `disable_braces_recovery()` (is the disable used anywhere?)

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -113,16 +118,14 @@ pub fn stmt(p: &mut Parser, recovery_set: impl Into<Option<TokenSet>>) -> Option
...
-			p.err_recover(
+			p.recover_on_unexpected_node(RecoveryBag::with_error(
+				dbg!(recovery_set.into().unwrap_or(STMT_RECOVERY_SET)),
+				unknown_node_on_error,
 				err,
-				recovery_set.into().unwrap_or(STMT_RECOVERY_SET),
-				false,
-				ERROR,
-			);
+			));
```
MichaReiser, line 128: For example, these handlers insert error nodes in places where otherwise a statement would have been. Shouldn't it then always insert an `UnknownStatement`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -216,7 +219,7 @@ fn expr_stmt(p: &mut Parser) -> Option<CompletedMarker> {
 
 		let m = expr.undo_completion(p);
 		p.bump_any();
-		stmt(p, None);
+		stmt(p, None, JS_UNKNOWN_EXPRESSION);
 		return Some(m.complete(p, JS_LABELED_STATEMENT));
```
MichaReiser, line 223: The body of a labeled statement must be a stmt. Why is it that we pass expression here?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -560,6 +563,7 @@ pub fn condition(p: &mut Parser) -> CompletedMarker {
 /// An expression wrapped in parentheses such as `()`
 pub fn parenthesized_expression(p: &mut Parser) {
 	p.state.allow_object_expr = p.expect(T!['(']);
+	// p.recover_on_unexpected_node(recovery_bag)
```
MichaReiser, line 566: Delete?

crates/rslint_parser/test_data/inline/err/do_while_stmt_err.rast
```diff
@@ -30,7 +30,7 @@ JS_ROOT@0..42
                 JS_BOOLEAN_LITERAL@37..41
                   TRUE_KW@37..41 "true"
                 WHITESPACE@41..42 "\n"
-                ERROR@42..42
+                JS_UNKNOWN_EXPRESSION@42..42
```
MichaReiser, line 33: An expression in the place of the `While` statement's body isn't valid. This should be a `UnknownStatement`

crates/rslint_parser/test_data/inline/err/while_stmt_err.rast
```diff
@@ -39,7 +39,7 @@ JS_ROOT@0..52
         TRUE_KW@44..48 "true"
       R_PAREN@48..49 ")"
       WHITESPACE@49..50 " "
-      ERROR@50..51
+      JS_UNKNOWN_EXPRESSION@50..51
```
MichaReiser, line 42: The body of a while statement must be a statement and not an expression

MichaReiser (changes requested):
I like the `RecoveryBag` and overall great improvements! 

There are some cases where we now insert invalid children (e.g. expressions instead of statements). 

Would you mind listing in the PR summary for parser rules your changing the error recovery?

crates/rslint_parser/src/lossless_tree_sink.rs
xunilrj: No. But I will remove this code in my next PR, today. So I would not care too much.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -588,13 +592,17 @@ pub fn if_stmt(p: &mut Parser) -> CompletedMarker {
 
 	// body
 	// allows us to recover from `if (true) else {}`
-	stmt(p, STMT_RECOVERY_SET.union(token_set![T![else]]));
+	stmt(
```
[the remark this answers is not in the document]
ematipico: Actually, this is needed. As the comment says, it allows the parser to recover in case the the `if` is missing the needed curly brackets. We are not parsing the consequent yet. 

Here, we are telling to the parser that it should recover if it sees a `TokenSet` that includes the `else` too. But, if we have, for example:

```js
if (true)) else {}
-------^
```

The additional bracket will be marked as `JS_UNKNOWN_STATEMENT`

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -961,10 +972,12 @@ fn switch_clause(p: &mut Parser) -> Option<Range<usize>> {
 			p.expect(T![:]);
 			let cons_list = p.start();
 			while !p.at_ts(token_set![T![default], T![case], T!['}'], EOF]) {
-				stmt(p, None);
+				stmt(p, None, JS_UNKNOWN_STATEMENT);
 			}
 			cons_list.complete(p, LIST);
 			m.complete(p, JS_CASE_CLAUSE);
+			// We return an empty range, to tell the parser that we haven't found a default clause, but there's not error
+			return Some(0..0);
```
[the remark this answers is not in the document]
ematipico: I thought that too. I guess I am still too concerned about changing too much the parser 😓 
I will apply the change :)

crates/rslint_parser/src/parser.rs
ematipico: So easy! Thank you!

crates/rslint_parser/src/recovery_bag.rs
ematipico: Thank you all for the suggestions!

crates/rslint_parser/src/syntax/stmt.rs
ematipico: > Is this inserting the UnknownStatement as a child of a switch? 

The error recovery here handles cases like this:

```js
switch (foo) {
	var bar = true;
}
```

I didn't understand the rest of the comment though, sorry.

crates/rslint_parser/test_data/inline/err/switch_stmt_no_cases.rast
```diff
@@ -0,0 +1,59 @@
...
+      WHITESPACE@6..7 " "
+      L_PAREN@7..8 "("
+      JS_REFERENCE_IDENTIFIER_EXPRESSION@8..11
+        IDENT@8..11 "foo"
+      R_PAREN@11..12 ")"
+      WHITESPACE@12..13 " "
+      L_CURLY@13..14 "{"
+      WHITESPACE@14..16 "\n\t"
+      LIST@16..16
+    JS_VARIABLE_DECLARATION_STATEMENT@16..31
```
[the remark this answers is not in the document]
ematipico: Let's tackle this in a different PR. Now that we have a test case, we can refine it later

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: Yes, what's the output it generates in this case? 

One thing we could do to recover from this case is to generate the following tree

```
SWITCH_STATEMENT
  SWITCH_CASE_CLAUSE
    MISSING (case keyword)
    MISSING (test node)
    MISSING (:)
    VARIABLE_DECLARATION_STATEMENT
```

So we would parse the statement and insert a non existing switch case clause. This is fine, because we add an error to the case and passes will see that it's broken when they visit the test node / case keyword...

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Probably the fact that we are inside function called `expr_stmt` led me to this mistake. Will change

crates/rslint_parser/src/syntax/stmt.rs
ematipico: It is true, although, as we didn't have enough test cases, I wanted to make sure that the parent passed the unknown node, so I knew who passed it. 

I am going to refactor the change now

crates/rslint_parser/src/parse_recoverer.rs
```diff
@@ -0,0 +1,72 @@
...
+	/// It tells the parser to recover if the position is inside a set of [tokens](TokenSet)
+	recovery: TokenSet,
+	/// It tells the parser to recover if the current token is a curly brace
+	include_braces: bool,
+	/// The kind of the unknown node the parser inserts if it isn't able to recover because
+	/// the current token is neither in the recovery set nor any of `{` or `}`.
+	unknown_node_kind: SyntaxKind,
+}
+
+impl ParseRecoverer {
```
MichaReiser, line 21: nit: When this is called `ParseRecoverer`, should it then have a `recover(p: &mut Parser)` method, so that it handles all recovery logic?

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Here's the output:

```
    JS_SWITCH_STATEMENT@0..16
      SWITCH_KW@0..6 "switch"
      WHITESPACE@6..7 " "
      L_PAREN@7..8 "("
      JS_REFERENCE_IDENTIFIER_EXPRESSION@8..11
        IDENT@8..11 "foo"
      R_PAREN@11..12 ")"
      WHITESPACE@12..13 " "
      L_CURLY@13..14 "{"
      WHITESPACE@14..16 "\n\t"
      LIST@16..16
    JS_VARIABLE_DECLARATION_STATEMENT@16..31
      JS_VARIABLE_DECLARATION@16..30
        VAR_KW@16..19 "var"
        WHITESPACE@19..20 " "
        LIST@20..30
          JS_VARIABLE_DECLARATOR@20..30
            SINGLE_PATTERN@20..23
              NAME@20..23
                IDENT@20..23 "bar"
            WHITESPACE@23..24 " "
            JS_EQUAL_VALUE_CLAUSE@24..30
              EQ@24..25 "="
              WHITESPACE@25..26 " "
              JS_BOOLEAN_LITERAL@26..30
                TRUE_KW@26..30 "true"
      SEMICOLON@30..31 ";"
    WHITESPACE@31..32 "\n"
    JS_UNKNOWN_STATEMENT@32..33
      R_CURLY@32..33 "}"
```

> So we would parse the statement and insert a non existing switch case clause

We can't assume we have a `case` clause. It could be a `default` clause too. I believe this is a particular case that requires a particular strategy to recover from the error. I think we should do it in a different PR. It's good that we a test case now.

MichaReiser: Does that mean that it is inserting the `JS_UNKNOWN_STATEMENT` if it finds any unexpected token? 

Does that mean that we can end up in a situation where the parser inserts two statements into the parent, first the `switch` statement and then the `UnknownStatement` for any unexpected content?

ematipico: > Does that mean that it is inserting the JS_UNKNOWN_STATEMENT if it finds any unexpected token?

Yes

> Does that mean that we can end up in a situation where the parser inserts two statements into the parent, first the switch statement and then the UnknownStatement for any unexpected content?

Yes. And this is valid not for just a switch statement, but all similar situations:

```js
function foo() } } } } { }
```

Will emit:

```
    JS_FUNCTION_DECLARATION@0..14
      FUNCTION_KW@0..8 "function"
      WHITESPACE@8..9 " "
      JS_IDENTIFIER_BINDING@9..12
        IDENT@9..12 "foo"
      JS_PARAMETER_LIST@12..14
        L_PAREN@12..13 "("
        LIST@13..13
        R_PAREN@13..14 ")"
    JS_UNKNOWN_STATEMENT@14..15
      R_PAREN@14..15 "}"
    JS_UNKNOWN_STATEMENT@15..16
      R_CURLY@15..16 "}"
    JS_UNKNOWN_STATEMENT@16..17
      R_PAREN@16..17 "}"
    JS_UNKNOWN_STATEMENT@17..18
      R_CURLY@17..18 "}"
    JS_BLOCK_STATEMENT@18..25
      L_CURLY@18..19 "{"
      LIST@19..25
        JS_BLOCK_STATEMENT@19..25
```

ematipico:
I reverted the changes I did around the parsing of `switch`. It fixed a infinite recursion inside the parser but it lowered the coverage, meaning that we'd need to spend more time around it.

ematipico:
> There are some cases where we now insert invalid children (e.g. expressions instead of statements).

Errors, now fixed.

> Would you mind listing in the PR summary for parser rules your changing the error recovery?

Done

crates/rslint_parser/src/parse_recoverer.rs
```diff
@@ -0,0 +1,94 @@
...
+			}
+		}
+		if !self.parsing_is_recoverable(p) {
+			let m = p.start();
+			p.bump_any();
+			m.complete(p, self.get_unknown_node_kind());
+		}
+	}
+
+	/// Checks if the parsing phase is recoverable by checking curly braces and toke set
```
yassere, line 72: toke -> token

crates/rslint_parser/src/parse_recoverer.rs
```diff
@@ -0,0 +1,94 @@
...
+	fn parsing_is_recoverable(&self, parser: &Parser) -> bool {
+		self.is_at_token_set(parser) || self.is_at_braces(parser)
+	}
+
+	/// It returns the diagnostic
+	fn get_error(&self) -> Option<Diagnostic> {
+		self.error.to_owned()
+	}
+
+	/// It return the unknown node kind that will be used to complete the parsing
```
yassere, line 82: return -> returns

crates/rslint_parser/src/parser.rs
```diff
@@ -179,75 +179,24 @@ impl<'t> Parser<'t> {
...
+	// 	let error = recovery_bag.get_error();
+	// 	if let Some(error) = error {
+	// 		self.error(error);
+	// 	}
+	// 	if !recovery_bag.parsing_is_recoverable(self) {
+	// 		let m = self.start();
+	// 		self.bump_any();
+	// 		m.complete(self, recovery_bag.get_unknown_node_kind());
+	// 	}
+	// }
```
yassere, line 199: Is this just an outdated version that will be removed?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -588,6 +588,7 @@ pub fn if_stmt(p: &mut Parser) -> CompletedMarker {
 
 	// body
 	// allows us to recover from `if (true) else {}`
+	// FIXME: in this case
```
MichaReiser, line 591: TODO?

MichaReiser (approved):
Looks good. Please take a look at the comments from @yassere before merging.

PR #1799, opened 2021-11-18
Refactor MemberExpression AST structure

MichaReiser:
## Summary

Refactors the AST tree structure for MemberExpressions as part of #1725 

* Renames `BracketExpr` to `JsComputedMemberExpression`
* Renames `DotExpr` to `JsStaticMemberExpression`
* Introduces the new `JsReferenceIdentifierMember` and `JsReferencePrivateMember` which are references to member names (analog to `JsReferenceIdentifierExpression` that references an identifier and `JsBindingIdentifier` that defines an identifier)
* Merge `PrivatePropAccess` into `JsStaticMemberExpression` (enabled by introducing `JsReferenceMember`
* Introduce `SuperExpression` so that `super.test()` works 
* Add new check that verifies that calling `super()` only works inside of constructor (I leave checking if we're inside of a subclass to another PR).
* Delete `SuperCall` as this can now be modelled using `CallExpr`
* Deleted some no longer used nodes/kinds

## Test Plan

Verified the coverage and that all tests are still passing. Added new integration tests.

The coverage is increasing because of the added check if `super` is called from inside a constructor.


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -553,54 +550,72 @@ pub fn subscripts(p: &mut Parser, mut lhs: CompletedMarker, no_call: bool) -> Co
...
 				.primary(priv_range.range(p), "");
 
 			p.error(err);
 			return m.complete(p, ERROR);
 		}
-		if let Some(range) = range {
-			let err = p
-				.err_builder("optional chaining cannot contain private identifiers")
-				.primary(range, "");
+	}
```
MichaReiser, line 580: This isn't true. `err?.#test` works perfectly fine.

xtask/js.ungram
```diff
@@ -848,6 +841,28 @@ JsComputedMemberName =
...
+// A reference to a private member
+// a.#b
+//   ^^
+// let { #b: c } = a
+//       ^^
+JsReferencePrivateMember =
+	'#'
+	name: 'ident'
+
+
```
MichaReiser, line 865: I'm not entirely sure about the naming here. Should these be postfixed with `JsAnyReferenceMemberName` since these are names and not members? Might sound a bit weird in the case of `JsReferenceIdentifierMemberName`

MichaReiser: @jamiebuilds wdyt?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -457,30 +456,38 @@ pub fn member_or_new_expr(p: &mut Parser, new_expr: bool) -> Option<CompletedMar
...
-				m.complete(p, BRACKET_EXPR)
+			T![.] => static_member_expression(p, super_completed, T![.]),
+			T!['['] => computed_member_expression(p, super_completed, false),
+			T![?.] => {
+				super_completed.change_kind(p, JS_UNKNOWN_EXPRESSION);
+				p.error(
+					p.err_builder(
+						"Super doesn't support optional chaining as super can never be null",
+					)
+					.primary(super_completed.range(p), ""),
```
ematipico, line 471: Should we suggest  to remove the `?.`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -457,30 +456,38 @@ pub fn member_or_new_expr(p: &mut Parser, new_expr: bool) -> Option<CompletedMar
...
-			}
-			T!['['] => {
-				p.bump_any();
-				expr(p);
-				p.expect(T![']']);
-				m.complete(p, BRACKET_EXPR)
+			T![.] => static_member_expression(p, super_completed, T![.]),
+			T!['['] => computed_member_expression(p, super_completed, false),
+			T![?.] => {
+				super_completed.change_kind(p, JS_UNKNOWN_EXPRESSION);
```
ematipico, line 466: I'm not sure having an unknown node is the correct approach.

Actually, I think that this approach is different from how the current parser treats the errors. Is this what we want?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1158,15 +1185,22 @@ pub fn spread_element(p: &mut Parser) -> CompletedMarker {
...
-	}
+	let lhs = if p.at(T![super]) && p.nth_at(1, T!['(']) {
+		let mut super_marker = super_expression(p);
+
+		if !p.state.in_constructor {
+			p.error(
+				p.err_builder("`super` is only valid inside of a class constructor of a subclass.")
+					.primary(super_marker.range(p), ""),
+			);
+			super_marker.change_kind(p, JS_UNKNOWN_EXPRESSION);
```
ematipico, line 1196: Is this the approach we should start implementing? Changing type at every error? Or, to make the question from a different angle, what's the logic of *when* using unknown nodes?

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: I'm not entirely decided on this yet to be honest. I see the advantage of leaving it as is for example in the case where we find a import/export statement inside of a script. We may then still be able to provide auto completion. 

However, it means that plugin passes won't know that somethings off except if they test the if the node has a diagnostic. 

That's why I think we should change the node type because not changing the node type would mean that we return a broken AST.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: I don't think it's at every error. Missing children is fine. Changing the type is only necessary if the node isn't correct in this context (I think xD)

xtask/js.ungram
jamiebuilds: I don't know that I would call this a "reference" since it's not actually looking anything up in scope. Accessing a member uses completely different logic that could be affected by getters/proxies/etc, so relating it to scope references seems incorrect.

I like the idea of using `JsStaticMember` (or `JsLiteralMember`) 

```js
JsStaticMember(Name?) -- or -- JsLiteralMember(Name?)
JsPrivateMember(Name?)
```

- We could use the `*Name` suffix for all miscellaneous "identifier-like" nodes
- We could use "literal" to mean "any value that can be determined from the source text alone" which would make "static" less overloaded since it's another language feature

MichaReiser: That's an excellent point and something I haven't considered. My main crux is that the potential of name clashes.

* Classes and Object have `Js*ClassMember` and `Js*ObjectMember`, and it might be a nice addition in the future to have a `JsAnyMember` that is a union over all possible members. Using `JsStaticMember` and `JsPrivateMember` would then not be part of that union which is confusing from a naming perspective
* Classes and Object permit identifiers, string literals, and number literals as member names which are why this PR introduces a `JsLiteralMemberName`. The problem is, string literals and number literals are not permitted in `MemberExpression's, which is why they need two different types (with different names, calling both `*LiteralMemberName` might be confusing if they are not interchangeable 

That's why I opted for the `Reference` prefix to distinguish the use cases clearly, but, as mentioned, I'm only somewhat happy with it. 

An alternative that I considered is to prefix the member names used for classes and object expressions with `Binding` because they introduce new bindings. That's probably more true for classes that have their own scope, and each member defines a new binding in that scope. Objects may define members, but they don't introduce a new binding to the scope (they bind the members on the object, but I don't think that's the meaning of Binding we're using). 

Overall, my problem is that these names are similar but have different semantics nonetheless.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: That's a good idea. I'll add it on Monday

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: I'm coming to the conclusion that this is indeed what we should do. Let's talk about it on Tuesday

xtask/js.ungram
MichaReiser: I talked about the naming with @jamiebuilds . We don't have a better name yet but we also came up with some more renames for other structures. So I think it's best to move on with this PR and fix the name when renaming all other fields/ nodes.

PR #1800, opened 2021-11-18
Rename `*Literal` nodes to `*LiteralExpression`

MichaReiser:
## Summary

#1799  introduced a new `JsLiteralMemberName` that is defined as `JsLiteralMemberName = value: ('ident' | 'js_string_literal_token' | 'js_number_literal_token')`. That means, quering for all `StringLiterals` using `root.descendants().filter(JsStringLiteral::cast)` will not return the `string_literal_token` inside of the `JsLiteralMemberName`. 

This is expected, because the `JsLiteralMemberName` contains the token and not the `JsStringLiteral` node but it may be unexpected for pass authors. 

I believe that not using a `JsStringLiteral` inside of member names or import statements (module name) is desired because authors may otherwise be inclined to, for example, replace all `JsStringLiteral` with `JsTemplateLiteralExpression` but that will create an invalid syntax tree in the case of import statements. 

That's why this PR renames `Js*Literal`s to `Js*LiteralExpression` to make it clear that these are the literal expressions. It then renames `JS_*_LITERAL_TOKEN` to `JS_*_LITERAL` as we no longer need the token postfix to avoid the name collision. 

The PR does not change the parsing for `PropName` since this will go away soon anyway.

## Test Plan

Verified the changes rast outputs to make sure the parser changes were correctly applied.


MichaReiser:
The CI job will run when the base branch is merged.

ematipico:
That's the thing. The CI is not running here and there's now way for reviewers to see if these new changes are healthy or not. What if the coverage goes down?

ematipico (approved):

Issue #1726, opened 2021-10-29
☂️ Formatter API improvements 


MichaReiser:
Part of #1718 

Umbrella task to update the `Formatter` after we made the improvements to our parser. 

- [x] #1740
- [x] Rename `format_raw` to `format_verbatim` #1976 
- [x] **Decision**: Add helpers for formatting nodes
    - A) Implement extension methods on `SyntaxResult` , `Option`, and `AstNode`/ `SyntaxToken` to format the element (to reduce the number of try operators). You can find some ideas [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3d25881c7e0eb2be701d0fc418d99242) but please spend some time to think about alternatives (for example, should we add helpers to the `FormatResult` type?).
    - **B) Provide helpers to format optional tokens, for example, `formatter.format_optional_token(token) -> FormatElement` and overload `format_token` to accept an `Option`: `format_token<T: Into<Option<Token>>`? Or implement the `format` methods as extension methods?**
        
        ```rust
        formatter.format_token(stmt.semicolon_token()).unwrap_or_else(|| token(";"));
        formatter.format_optional_token(element.comma_token());
        ```
        
        Implemented as part of #1916: `format_token` is generic over optional tokens, can take either a `SyntaxToken` or `Option<SyntaxToken>` and return a `FormatElement` or `Option<FormatElement>` respectively
        
- [x] Change the formatter to visit every token, including`,` and `;`.
- [x] #1916 
- [x] Re-evaluate CST-> CST printing? This should now be possible with explicit rules to which token trivia gets attached.

MichaReiser:
We probably also want to explore to codegen the union formatter implementations

ematipico:
> We probably also want to explore to codegen the union formatter implementations

It makes sense. The only drawback is that if we create new nodes out of it, we must implement also the `ToFormatElement` for those nodes. 

I think we could also automate that part, at a certain extent. RomeJS used to create a builder for each node. If we decided to go in the same direction, we might be able to auto generate a simple scaffolding for new nodes.

xunilrj:
When finishing the trivia changes I had to decide what to do when formatting statements with problems. See https://github.com/rome/tools/pull/1801/files#r752272632

Basically we are trimming "whitespaces" around the statement; but keeping everything inside as it is.
Would be interesting to revisit this later and decide what we want to do.

Boshen:
>  Decision: Add helpers for formatting nodes
[...]

PR #1801, opened 2021-11-18
LosslessTreeSinker attaching trivia to tokens

xunilrj:
part of: #1720
Improved changes of #1738

# Tasks

- [x] LosslessTreeSink
- [x] LossyTreeSink

I also needed to make some formatter changes.

- [x] Trim leading and trailing trivia from ```SynyaxNode``` with error. See ```if_stmt_err.js```


crates/rome_formatter/src/ts/statements/mod.rs
```diff
@@ -29,26 +29,9 @@ pub fn format_statements(
...
-						} else {
-							FormatElement::List(list)
-						}
-					}
-					_ => verbatim,
-				}
-			})
+			formatter
+				.format_node(stmt.clone())
+				.unwrap_or_else(|_| formatter.format_raw(stmt.syntax()).trim_start().trim_end())
```
xunilrj, line 34: When we insert a problematic statement into the formatter, we are trimming all "whitespace" from the start and from the end. Is that what we want to do?

MichaReiser: I don't know. Can we add a todo and mention it in the the formatter tracking task?

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +743,80 @@ impl FormatElement {
...
+	/// the [FormatElement].
+	/// Including characters of the [FormatElement::Token] variant.
+	pub fn trim_start(&self) -> FormatElement {
+		match self {
+			FormatElement::Empty => FormatElement::Empty,
+			FormatElement::Space => FormatElement::Empty,
+			FormatElement::Line(_) => FormatElement::Empty,
+			FormatElement::Indent(_) => FormatElement::Empty,
+			FormatElement::Group(_) => todo!(),
+			FormatElement::ConditionalGroupContent(_) => todo!(),
```
MichaReiser, line 757: This should call `trim_start` on the content

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +743,80 @@ impl FormatElement {
...
+	/// the [FormatElement].
+	/// Including characters of the [FormatElement::Token] variant.
+	pub fn trim_end(&self) -> FormatElement {
+		match self {
+			FormatElement::Empty => FormatElement::Empty,
+			FormatElement::Space => FormatElement::Empty,
+			FormatElement::Line(_) => FormatElement::Empty,
+			FormatElement::Indent(_) => FormatElement::Empty,
+			FormatElement::Group(_) => todo!(),
+			FormatElement::ConditionalGroupContent(_) => todo!(),
```
MichaReiser, line 793: Same, traverse into content

crates/rslint_parser/src/tests.rs
```diff
@@ -172,3 +172,104 @@ fn assert_errors_are_absent(errors: &[ParserError], path: &Path, syntax: &Syntax
 		syntax
 	);
 }
+
+#[test]
+pub fn test_trivia_attached_to_tokens() {
+	use crate::util::SyntaxNodeExt;
+
+	let text = "/**/let a = 1; // nice variable \n /*hey*/ let \t b = 2; // another nice variable";
+	let m = parse_module(text, 0);
```
MichaReiser, line 181: Would you mind adding a test for

```
// only a comment
```

I'm wondering where we attach the comment to.

crates/rslint_parser/test_data/inline/err/async_arrow_expr_await_parameter.rast
```diff
@@ -1,35 +1,28 @@
-JS_ROOT@0..26
+JS_ROOT@0..25
```
MichaReiser, line 1: What's the reason that the root contains a character less? What happens with the trailing whitespace?

xunilrj: Fixed.

crates/rslint_parser/src/tests.rs
xunilrj: Done. See "just_trivia_must_be_appended_to_eof" below.

crates/rome_formatter/src/ts/statements/mod.rs
xunilrj: https://github.com/rome/tools/issues/1726#issuecomment-973051032

crates/rome_formatter/src/format_element.rs
```diff
@@ -154,6 +154,18 @@ pub fn token(text: &str) -> FormatElement {
 	}
 }
 
+/// Creates a token that gets written as is to the output. See fucntion [token] for more details
+/// and examples.
+/// This function moves the string to be more efficient.
+#[inline]
+pub fn token_from_string(text: String) -> FormatElement {
```
ematipico, line 161: Wasn't changing the existing function to `token(text: Into<&str>)` enough?

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +743,80 @@ impl FormatElement {
 	pub fn is_empty(&self) -> bool {
 		self == &FormatElement::Empty
 	}
+
+	/// Remove all empty, space, line breaks, idents from the start of
```
ematipico, line 747: I think that "empty" are not removed. On line 752, they are the same. Probably we should remove the "empty" part from the doc?

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +743,80 @@ impl FormatElement {
...
+			FormatElement::Line(_) => FormatElement::Empty,
+			FormatElement::Indent(_) => FormatElement::Empty,
+			FormatElement::Group(_) => todo!(),
+			FormatElement::ConditionalGroupContent(_) => todo!(),
+			FormatElement::List(list) => {
+				let mut content: Vec<_> = list
+					.iter()
+					.skip_while(|e| match e {
+						FormatElement::Empty => true,
+						FormatElement::Space => true,
```
ematipico, line 763:
```suggestion
						FormatElement::Empty |
						FormatElement::Space => true,
```

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +743,80 @@ impl FormatElement {
...
+							let s = t.trim_start();
+							s.is_empty()
+						}
+						_ => false,
+					})
+					.map(Clone::clone)
+					.collect();
+				if let Some(FormatElement::Token(s)) = content.get_mut(0) {
+					s.0 = s.trim_start().to_string()
+				}
```
ematipico, line 776: What exactly is happening here?

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +743,80 @@ impl FormatElement {
...
+				FormatElement::List(List::new(content))
+			}
+			FormatElement::Token(s) => token(s.trim_start()),
+		}
+	}
+
+	/// Remove all empty, space, line breaks, idents from the end of
```
ematipico, line 783: "empty" can be removed from the doc

crates/rome_formatter/src/ts/statements/mod.rs
ematipico: Prettier at the moment removes all "whitespace" if there are more than one - meaning at least 2 new lines.  As we want to mimic their behaviour, I think we should Prettier as starting point in case we have some decision to take.

crates/rome_rowan/src/api.rs
```diff
@@ -98,7 +98,7 @@ impl<L: Language> SyntaxTriviaPiece<L> {
 	/// ```
 	pub fn text(&self) -> &str {
 		let txt = self.raw.text();
-		let start = self.offset;
+		let start = self.offset - self.raw.offset();
```
ematipico, line 101: What's this change about?

crates/rome_rowan/src/api.rs
```diff
@@ -311,7 +311,7 @@ impl<L: Language> Iterator for SyntaxTriviaPiecesIterator<L> {
 }
 
 impl<L: Language> SyntaxTrivia<L> {
-	/// Returns the text of all descendants tokens combined, including all trivia.
+	/// Returns all pieces of this trivia.
```
ematipico, line 314: Should we link "pieces" to the correct `struct`? If it's read without context and without reference, it might cause some confusion

crates/rome_rowan/src/api.rs
```diff
@@ -622,6 +622,11 @@ impl<L: Language> SyntaxNode<L> {
 		self.raw.descendants().map(SyntaxNode::from)
 	}
 
+	pub fn descendants_tokens(&self) -> impl Iterator<Item = SyntaxToken<L>> {
+		self.descendants_with_tokens()
```
ematipico, line 626: Why don't we use the `self.raw` here? Like in the other APIs

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -18,6 +19,9 @@ pub struct LosslessTreeSink<'a> {
 	state: State,
 	errors: Vec<ParserError>,
 	inner: SyntaxTreeBuilder,
+	needs_eof: bool,
```
ematipico, line 22: Can we document the new property please?

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -18,6 +19,9 @@ pub struct LosslessTreeSink<'a> {
...
+	/// Trivia start Offset and its pieces.
```
ematipico, line 23: Can we link "pieces" please? And "Offset" too I think

crates/rslint_parser/test_data/inline/err/binary_expressions_err.rast
```diff
@@ -1,6 +1,6 @@
 JS_ROOT@0..35
   LIST@0..0
-  LIST@0..34
+  LIST@0..35
```
ematipico, line 3: Is it normal that some trees are wider?

crates/rome_formatter/src/format_element.rs
xunilrj: Done.

crates/rome_formatter/src/format_element.rs
xunilrj: Done.

xunilrj:
> Can we delete these now?

I think we can. I will create another PR to clean everything we don't need anymore.

crates/rslint_parser/test_data/inline/err/binary_expressions_err.rast
xunilrj: I think yes, it is.
Indentation is not helping here... but this is happening because the trailing whitespace now is part of the node as trailing trivia, for example.

For example, a smaller example:

Before
```
              JS_REFERENCE_IDENTIFIER_EXPRESSION@4..7
                IDENT@4..7 "foo"
              WHITESPACE@7..8 " "
``` 

After

```
              JS_REFERENCE_IDENTIFIER_EXPRESSION@4..8
                IDENT@4..8 "foo "
```

crates/rome_formatter/src/ts/statements/mod.rs
xunilrj: hum.. I think we can keep what we have now, and revisit if it does not makes sense.

crates/rome_formatter/src/format_element.rs
xunilrj: No because I actually want to move the String, so I can create a ```Token``` without any extra allocation.

Places where you only have a &str, there is nothing we can do. You will need to call ```.to_string()``` and allocate.

But places where you own a String, and don't need it, you can move it to the Token. If this string is "casted" to ```&str```, I will need to ```to_string()``` it, what I am trying to avoid here.

I could do the inverse actually ```Into<String>```.

Complete example:

```rust

#[derive(Debug)]
pub struct Token(String);

pub fn token_into_str<'a, T: Into<&'a str>>(s: T) -> Token {
    Token(s.into().to_string())
}

pub fn token_into_string<T: Into<String>>(s: T) -> Token {
  Token(s.into())
}

pub fn token_as_ref<'a, T: AsRef<str>>(str: T) -> Token {
    Token(str.as_ref().to_string())
}

pub fn token_with_move(s: String) -> Token {
    Token(s)
}

fn main() {
    let allocations = allocation_counter::count(|| {
        let s = "a".to_string();
        println!("{:?}", token_into_str(s.as_str()));
    });
    println!("token_into_str: {}", allocations);

    let allocations = allocation_counter::count(|| {
        let s = "a".to_string();
        println!("{:?}", token_into_string(s.as_str()));
    });
    println!("token_into_string(&str): {}", allocations);

    let allocations = allocation_counter::count(|| {
        let s = "a".to_string();
        println!("{:?}", token_into_string(s));
    });
    println!("token_into_string(String): {}", allocations);

    let allocations = allocation_counter::count(|| {
        let s = "a".to_string();
        println!("{:?}", token_as_ref(&s));
    });
    println!("token_as_ref(&String): {}", allocations);

    let allocations = allocation_counter::count(|| {
        let s = "a".to_string();
        println!("{:?}", token_as_ref(s.as_str()));
    });
    println!("token_as_ref(&str): {}", allocations);

    let allocations = allocation_counter::count(|| {
        let s = "a".to_string();
        println!("{:?}", token_with_move(s));
    });
    println!("token_with_move: {}", allocations);
}
```

```
Token("a")
token_into_str: 3
Token("a")
token_into_string(&str): 2
Token("a")
token_into_string(String): 1
Token("a")
token_as_ref(&String): 2
Token("a")
token_as_ref(&str): 2
Token("a")
token_with_move: 1
```

https://replit.com/@xunilrj/MutedDimTest#src/main.rs

xunilrj: Done.

crates/rome_formatter/src/format_element.rs
xunilrj: Done.

crates/rome_formatter/src/format_element.rs
xunilrj: Because tokens have leading trivia now, at the formatter we can have:

```
   Indent(Token("\t\tlet"))
```

When we call ```.trim_start()" for example, these three lines are trimming the token to Token("let").

crates/rome_formatter/src/format_element.rs
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Each ```SyntaxTriviaPiece``` has its offset in the document. But to slice the token text, I need the offset from within the token text (including trivia).

For example

```rust
function foo() {\n/*hi*/let a = 0;}
```

The token ```let``` has:
leading trivia: "\n/*hi*/"
trailing trivia: " "

When you ask for the SyntaxTriviaPiece of the comments, it has a global offset of 17..23. But to get the text of the trivia piece, we need to slice the token text, that is "\n/*hi*/let"; The comments offset of 17..23, actually is 1..7.

crates/rome_rowan/src/api.rs
xunilrj: Done.

crates/rome_rowan/src/api.rs
xunilrj: Because using ```self.raw``` would return a ```cursor::SyntaxToken``` and I would need to create the correct ```api::SyntaxToken<L>``` from them. This method is doing this for me.

crates/rslint_parser/src/lossless_tree_sink.rs
xunilrj: Done.

crates/rslint_parser/src/lossless_tree_sink.rs
xunilrj: Piece link: Done.
Offset, no idea where to link this to.

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +732,80 @@ impl FormatElement {
...
+							s.is_empty()
+						}
+						_ => false,
+					})
+					.map(Clone::clone)
+					.collect();
+				content.reverse();
+				if let Some(FormatElement::Token(s)) = content.last_mut() {
+					s.0 = s.trim_end().to_string()
+				}
```
yassere, line 803: Rather than allocating a vector and then reversing it in place, could we find the `position` of the first non-whitespace element in reverse iteration so we can `take` up to that position in forward iteration and collect that into a new vector?

crates/rslint_parser/test_data/inline/err/binary_expressions_err.rast
```diff
@@ -9,40 +9,31 @@ JS_ROOT@0..35
           L_PAREN@3..4 "("
           LIST@4..9
             JS_BINARY_EXPRESSION@4..9
-              JS_REFERENCE_IDENTIFIER_EXPRESSION@4..7
-                IDENT@4..7 "foo"
-              WHITESPACE@7..8 " "
+              JS_REFERENCE_IDENTIFIER_EXPRESSION@4..8
+                IDENT@4..8 "foo "
```
MichaReiser, line 13: This format is a bit misleading. It reads as if the identifier name is `foo ` (with trailing space). Would it make sense to format the trivia more explicitly, for example

```
IDENT@4..8 text: (trivia) (text="foo") (trivia whitespace: " ")
```

Printing the trivia pieces one by one has the benefit that changes to the trivia parsing are tested as well.

crates/rslint_parser/test_data/inline/err/conditional_expr_err.rast
```diff
@@ -1,43 +1,31 @@
...
+          IDENT@20..24 "bar "
+        JS_CONDITIONAL_EXPRESSION@24..40
+          JS_REFERENCE_IDENTIFIER_EXPRESSION@24..28
+            IDENT@24..28 "baz "
+          QUESTION@28..30 "? "
+          JS_REFERENCE_IDENTIFIER_EXPRESSION@30..34
+            IDENT@30..34 "foo "
+          COLON@34..36 ": "
+          JS_REFERENCE_IDENTIFIER_EXPRESSION@36..40
+            IDENT@36..40 "bar\n"
```
MichaReiser, line 28: Shouldn't we see the `EOF` token as part of the test files?

MichaReiser (changes requested):
I really love this change, excellent work!

There are two last things that I think we should improve before moving forward.

## Debug Format:

I find it a bit misleading. It looks as if an identifier was parsed as `foo ` or keywords/punctuations as `\n!`. 

Would it make sense to format the trivia more explicitly, for example

```
IDENT@4..8 text: (trivia) (text="foo") (trivia whitespace: " ")
```

Swift actually prints tokens more like nodes since they now can contain rather bulky content like comments

```
IDENT@4..8 
  (trivia newline "\n")
  (trivia whitespace "    ")
  (text: "foo")
  (trivia whitespace: " ")
```

Printing the trivia pieces one by one has the benefit that changes to the trivia parsing are tested as well. In that spirit. Would you mind adding an integration test for some trivia (single line comment, mutliline comment etc.)

## EOF

I believe we shoulnd't append the EOF token conditionally, only when needed. Authors would see, oh, there's an EOF token I can test for and then bum, it isn't there in some cases. 

We should also make it clear where EOF tokens can appear in our grammar. I think it's sufficient to add it to the `JsRoot` rule for now. We can figure out the `parse_text` use case later (probably best to introduce a "fake" node for parsed text.

crates/rome_formatter/src/format_element.rs
xunilrj: Done. But this ```trim_start()``` and ```trim_end()``` are a little bit naive. If we decide to stick with them, I would try to improve them: code-wise, performance-wise and make them work with more cases.

crates/rome_formatter/src/format_element.rs
ematipico: Great explanation! Thank you! :)

crates/rome_rowan/src/api.rs
ematipico: Thank you! Makes sense now

crates/rslint_parser/src/lossless_tree_sink.rs
ematipico: probably offset is the `TextRage`?

crates/rome_formatter/src/format_element.rs
```diff
@@ -731,6 +732,87 @@ impl FormatElement {
...
+						FormatElement::Space => true,
+						FormatElement::Line(_) => true,
+						FormatElement::Indent(_) => true,
+						FormatElement::Token(t) => {
+							let s = t.trim_end();
+							s.is_empty()
+						}
+						_ => false,
+					})
+					.next();
```
yassere, line 799: For getting rid of the lint error, I think you could replace these three method calls with a [position](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position)

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -161,87 +134,110 @@ impl<'a> LosslessTreeSink<'a> {
...
-						return trivias.next().unwrap().0 + 1;
-					}
-				}
-				COMMENT => {
-					return idx + 1;
-				}
-				_ => {}
-			},
-			_ => return 0,
+	fn get_trivia(&mut self, break_on_newline: bool) -> (TextRange, Vec<TriviaPiece>) {
```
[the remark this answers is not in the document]
xunilrj: For all cases but the first and the last trivia of the file ```break_on_newline = false``` is for the trailing. That is why I prefer this name, because it is what the method is doing.

crates/rslint_parser/test_data/inline/err/binary_expressions_err.rast
xunilrj: What about now?

```
JS_VARIABLE_DECLARATION@0..13
        LET_KW@0..8 "let" [Comments("/**/")] [Whitespace(" ")]
```

crates/rome_formatter/src/format_element.rs
xunilrj: Done.

crates/rome_formatter/src/format_element.rs
xunilrj: Makes sense. Done.

crates/rslint_parser/test_data/inline/err/binary_expressions_err.rast
MichaReiser: Works for me. It might be helpful to be a bit more descriptive, especially because the leading/text are now out of order

```
        LET_KW@0..8 "let" (leading: COMMENTS("/**/")) (trailing: WHITESPACE(" "))
```

I would be OK not printing the trivia if it's empty.

Anyhow, that's more my personal taste now.

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -161,87 +134,110 @@ impl<'a> LosslessTreeSink<'a> {
...
+	/// Finishes the tree and return the root node with possible parser errors.
+	///
+	/// If tree is finished with pending trivia, but no tokens were generated, for example,
+	/// a completely commented file, a [SyntaxKind::EOF] will be generated and all pending trivia
+	/// will be appended to its leading trivia.
 	pub fn finish(mut self) -> (SyntaxNode, Vec<ParserError>) {
+		if self.needs_eof && !self.next_token_leading_trivia.1.is_empty() {
+			self.do_token(SyntaxKind::EOF, 0.into());
+		}
+
```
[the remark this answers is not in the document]
xunilrj: I am forcing a EOF at the tree sink for now. This way is easier and we never loose the last trivia of the file.

crates/rslint_parser/test_data/inline/err/conditional_expr_err.rast
xunilrj: ...and now you can see.

MichaReiser (approved):
Thanks! 

Can we remove the `hidden` from `EOF`. 

https://github.com/rome/tools/blob/8adc8b5dcd1777e45204a1401f9d595619dee6eb/crates/rslint_syntax/src/generated.rs#L11-L12

As this is now a public visible token.

PR #1803, opened 2021-11-19
Implement Drop for Marker

MichaReiser:
## Summary

This PR adds a drop bomb to the Parser's `Marker` implementation to ensure we always call either `abandon` or `complete` on every marker. 

I was stunned to see that the parser is able to produce a valid tree even with nodes in the tree that were never completed, and, therefore, don't have a `SyntaxKind` set. 

The reason why this works is that the parser sets the kind to `TOMBSTONE` when starting a new node and these get skipped when building the final tree. The problem is, skipped means the children of the `TOMBSTONE`d node get attached to the parent. 

This "bubbling up" of children is problematic, because let's say we have a node that calls into `expr`. The expectation would be that `expr` either produces a `UnknownExpression` or doesn't add any nodes to the parse tree. However, this isn't the case. `expr` may start parsing and then decide that it doesn't know how to continue and it then bails out with the result that all parsed tokens and nodes get attached to the parent. 

Bailing out is fine, but we should be very considerate about what we want to do in those cases, either completing the node with an `UNKNOWN` kind or propagating the error to the parent (by, for example, returning a `Result` type). 

## Test Plan

Verified coverage.

### DIFF
--- a/crates/rslint_parser/src/parser.rs
+++ b/crates/rslint_parser/src/parser.rs
@@ -3,11 +3,11 @@
 //! the parser yields events like `Start node`, `Error`, etc.
 //! These events are then applied to a `TreeSink`.
 
+use drop_bomb::DropBomb;
 use rslint_errors::Diagnostic;
 use std::borrow::BorrowMut;
 use std::cell::Cell;
 use std::ops::Range;
-use std::thread::panicking;
 
 use crate::*;
 
@@ -522,7 +522,7 @@ impl<'t> Parser<'t> {
 }
 
 /// A structure signifying the start of parsing of a syntax tree node
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Debug)]
 pub struct Marker {
 	/// The index in the events list
 	pub pos: u32,
@@ -530,7 +530,7 @@ pub struct Marker {
 	pub start: usize,
 	pub old_start: u32,
 	pub(crate) child_idx: Option<usize>,
-	finished: bool,
+	bomb: DropBomb,
 }
 
 impl Marker {
@@ -540,7 +540,7 @@ impl Marker {
 			start,
 			old_start: pos,
 			child_idx: None,
-			finished: false,
+			bomb: DropBomb::new("Marker must either be `completed` or `abandoned` to avoid that children are implicitly attached to a markers parent."),
 		}
 	}
 
@@ -555,7 +555,7 @@ impl Marker {
 	/// and mark the create a `CompletedMarker` for possible future
 	/// operation like `.precede()` to deal with forward_parent.
 	pub fn complete(mut self, p: &mut Parser, kind: SyntaxKind) -> CompletedMarker {
-		self.finished = true;
+		self.bomb.defuse();
 		let idx = self.pos as usize;
 		match p.events[idx] {
 			Event::Start {
@@ -576,7 +576,7 @@ impl Marker {
 	/// Abandons the syntax tree node. All its children
 	/// are attached to its parent instead.
 	pub fn abandon(mut self, p: &mut Parser) {
-		self.finished = true;
+		self.bomb.defuse();
 		let idx = self.pos as usize;
 		if idx == p.events.len() - 1 {
 			match p.events.pop() {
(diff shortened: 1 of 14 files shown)


crates/rslint_parser/src/parser.rs
```diff
@@ -522,14 +523,15 @@ impl<'t> Parser<'t> {
 }
 
 /// A structure signifying the start of parsing of a syntax tree node
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, PartialEq, Eq, Hash)]
```
MichaReiser, line 526: Cloning markers is super dangerous. Finishing both markers will certainly result in a panic.

crates/rslint_parser/src/syntax/pat.rs
```diff
@@ -237,6 +237,7 @@ fn object_binding_prop(p: &mut Parser, parameters: bool) -> Option<CompletedMark
 	let name = if let Some(n) = name {
 		n
 	} else {
+		m.abandon(p);
```
MichaReiser, line 240: I don't think we really should be calling abandon here and instead insert unknown nodes in all cases. Because abandoning simply attaches the parsed children to its parent... Probably containing elements that it didn't expect at all. However, we should do this in a dedicated PR.

yassere:
Is this inspired by the `DropBomb` on markers in the [rust-analyzer parser](https://github.com/rust-analyzer/rust-analyzer/blob/4c20d6879f7b23c532dee908031caab178bb79fa/crates/parser/src/parser.rs#L255-L294)?

MichaReiser:
> Is this inspired by the `DropBomb` on markers in the [rust-analyzer parser](https://github.com/rust-analyzer/rust-analyzer/blob/4c20d6879f7b23c532dee908031caab178bb79fa/crates/parser/src/parser.rs#L255-L294)?

I only knew that Rust uses drop bombs for the diagnostics builder and I copied the implementation from there. But yes, the idea is the same. I can certainly extract a `DropBomb` struct. 

The overall idea is to make error handling explicit. I need to have a closer look what the behavior is when you implicitly drop a  marker (if it is automatically abandoned). However, we should handle these explicitly everywhere because it often means that we parsed something that now gets attached to a parent that probably doesn't want some random child tokens.

yassere:
> I need to have a closer look what the behavior is when you implicitly drop a marker (if it is automatically abandoned).

What do you mean by implicitly dropping a marker? This drop implementation will run whenever any marker goes out of scope, right?

MichaReiser:
> > I need to have a closer look what the behavior is when you implicitly drop a marker (if it is automatically abandoned).
> 
> What do you mean by implicitly dropping a marker? This drop implementation will run whenever any marker goes out of scope, right?

That's correct. I mean if someone's not explicitly completing a marker.

yassere:
> I mean if someone's not explicitly completing a marker.

I still don't understand. Whenever a `Marker` is created, it will eventually go out of scope and this drop implementation will panic if the marker wasn't `finished` in between, right? Sorry if I'm completely misunderstanding what you're saying.

MichaReiser:
> > I mean if someone's not explicitly completing a marker.
> 
> I still don't understand. Whenever a `Marker` is created, it will eventually go out of scope and this drop implementation will panic if the marker wasn't `finished` in between, right? Sorry if I'm completely misunderstanding what you're saying.

That's correct. I'm mainly wondering why it was OK to not explicitly terminate markers by calling `complete` or `abandon`.  Because the `events` structure will contain `start` events without a matching `finish` event.

The underlining question is: Why did this work before

MichaReiser:
OK the trick is that `start` inserts a node of type `TOMBSTONE` and these get skipped. 

https://github.com/rome/tools/blob/85d1b8f48f90ca479ed5d999c85e36546e3c4b1b/crates/rslint_parser/src/parser.rs#L235-L239

I don't think this is greater because it then makes it difficult to reason how the final tree structure will look like

yassere:
Ah, thanks for clarifying. I didn't realize that you were talking about the existing behavior prior to this PR. Yeah, I agree that requiring markers to be explicitly abandoned is better.

ematipico (changes requested):
We can have some test that shows that this new feature actually works as expected?

ematipico (approved):

yassere (approved):

PR #1804, opened 2021-11-19
feat: unknown bindings and unknown patterns

ematipico:
## Summary

This PR adds more test cases and adds the correct unknown nodes inside existing test cases:

- created new test cases to cover existing errors
- `eval`, `argments` and `yield` will emit `JS_UNKNOWN_BINDING`
- when parsing a pattern, we decide the unknown node based on the state of the parser
- an incorrect usage of `let` will emit a `JS_UNKNOWN_BINDING`
- an incorrect object pattern can generate a `JS_UNKNOWN_PATTERN`
	```js
	let { 5 } = foo;
	```
- an incorrect parameters list will generate a `JS_UNKNOWN_PATTERN`
	```js
	function foo(true) {}
	```
- incorrect functions will generate `JS_UNKNOWN_BINDING`
	```js
	function {}
	```
	```
      JS_PARAMETER_LIST@23..38
        LIST@23..37
          JS_UNKNOWN_BINDING@23..24
            L_CURLY@23..24 "{"
          JS_UNKNOWN_BINDING@24..25
            R_CURLY@24..25 "}"
          WHITESPACE@25..26 "\n"
          ERROR@26..34
            FUNCTION_KW@26..34 "function"
          WHITESPACE@34..35 " "
          JS_UNKNOWN_BINDING@35..36
            STAR@35..36 "*"
          ERROR@36..37
            L_PAREN@36..37 "("
        R_PAREN@37..38 ")"
	```
- updated the test suite with all the correct tests, extracted directly from the source code (comments)

Please refers to the new tests in the PR for more details. Make sure to search for `JS_UNKNOWN_BINDING` and `JS_UNKNOWN_PATTERN`.

## Test Plan

```
cargo lint
cargo test
cargo format
```


yassere:
Were those new tests added manually into the `/inline` directory? I think those tests are intended to be [extracted](https://github.com/rome/tools/blob/4672e1c98aaaf52f90d8dc6b41a863acddce451f/xtask/src/codegen/parser_tests.rs#L47-L78).

ematipico:
> Were those new tests added manually into the `/inline` directory? I think those tests are intended to be [extracted](https://github.com/rome/tools/blob/4672e1c98aaaf52f90d8dc6b41a863acddce451f/xtask/src/codegen/parser_tests.rs#L47-L78).

What do you mean by "extracted"? 

Myself and Micha have been adding new tests in the `/inline` folder every time we change things in the parser, and it worked as intended - they are like snapshot tests.

yassere:
The tests themselves would work as intended, but I think the idea is that you add the test cases "inline" to comments right next to the parser code they're testing and then the codegen extracts those comments into the `inline` directory.

So for example, these [ok tests](https://github.com/rome/tools/blob/4672e1c98aaaf52f90d8dc6b41a863acddce451f/crates/rslint_parser/test_data/inline/ok/assign_expr.js) come from [here](https://github.com/rome/tools/blob/4672e1c98aaaf52f90d8dc6b41a863acddce451f/crates/rslint_parser/src/syntax/expr.rs#L196-L201) and these [err tests](https://github.com/rome/tools/blob/4672e1c98aaaf52f90d8dc6b41a863acddce451f/crates/rslint_parser/test_data/inline/err/binary_expressions_err.js) come from [here](https://github.com/rome/tools/blob/4672e1c98aaaf52f90d8dc6b41a863acddce451f/crates/rslint_parser/src/syntax/expr.rs#L303-L306).

ematipico:
Damn! I wished there was some documentation somewhere! If there was, I didn't find it anywhere!

 Thank you for letting me know, I will change that right away!

ematipico:
@MichaReiser I restored your tests by adding the new cases as commented code.

MichaReiser:
> @MichaReiser I restored your tests by adding the new cases as commented code. 

Thanks ema. How can I generate the inline tests? Is it a macro?

ematipico:
> > @MichaReiser I restored your tests by adding the new cases as commented code. 
> 
> Thanks ema. How can I generate the inline tests? Is it a macro?

It's an xtask: `cargo xtask codegen`.

This will generate the JS files out of the comments.

Then we run the command to update the .rast files.

I guess we should start documenting all these findings

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -106,6 +106,10 @@ fn object_member(p: &mut Parser) -> Option<CompletedMarker> {
 			// test object_expr_method
 			// let b = {
 			//  foo() {},
+				// foo() {},
+				// "bar"(a, b, c) {},
+				// ["foo" + "bar"](a) {},
+				// 5(...rest) {}
```
yassere, line 112: The indentation is off here and the `foo() {}` is duplicated.

crates/rslint_parser/src/syntax/pat.rs
```diff
@@ -175,9 +181,10 @@ pub fn array_binding_pattern(
 			m.complete(p, REST_PATTERN);
 			break;
 		} else if binding_element(p, parameters, assignment).is_none() {
+			// TODO: find a away to land
```
yassere, line 184: What does this TODO mean?

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -48,7 +48,7 @@ pub(super) fn object_expr(p: &mut Parser) -> CompletedMarker {
 /// An individual object property such as `"a": b` or `5: 6 + 6`.
 fn object_member(p: &mut Parser) -> Option<CompletedMarker> {
 	match p.cur() {
-		// test object_expr_getter_setter
+		// test object_expr_getter_getter
```
yassere, line 51: Is the duplicate `getter` here intentional?

yassere (commented):
This looks good to me, but I haven't personally worked much with the internals of this part of the parser.

crates/rslint_parser/src/syntax/pat.rs
```diff
@@ -70,14 +70,16 @@ pub fn pattern(p: &mut Parser, parameters: bool, assignment: bool) -> Option<Com
 			m.complete(p, SINGLE_PATTERN)
 		}
 		_ => {
+			let mut unknown_node_kind = JS_UNKNOWN_BINDING;
 			let err = p
 				.err_builder("Expected an identifier or pattern, but found none")
 				.primary(p.cur_tok().range, "");
 			let mut ts = token_set![T![ident], T![yield], T![await], T!['['],];
 			if p.state.allow_object_expr {
+				unknown_node_kind = JS_UNKNOWN_PATTERN;
```
MichaReiser, line 79: It's probably sufficient to go with `JS_UNKNOWN_PATTERN` for now because the current grammar doesn't distinguish between assignment target and bindings. 
```suggestion
```

I also don't think that testing for `p.state.allow_object_expr` is sufficient to decide if it is one or the other because the following is valid too

```
({ a } = b);
```

MichaReiser (approved):
Looks good and thanks for recovering the tests. 

The one thing I would change is to not distinguish between binding and pattern and only use `JS_UNKNOWN_PATTERN` because the current grammar doesn't make the distinction as well (that's what I've been working on when extracting assignment targets, and then later introduce bindings). 

We may need to give this a second pass once we have refined our approach to error recovery.

crates/rslint_parser/src/syntax/pat.rs
```diff
@@ -100,19 +100,21 @@ pub fn opt_binding_identifier(p: &mut Parser) -> Option<CompletedMarker> {
...
+		kind_to_change = JS_UNKNOWN_BINDING;
 		p.error(err);
 	}
 
 	if p.at(T![await]) && p.state.in_async {
 		let err = p
 			.err_builder("Illegal use of `await` as an identifier in an async context")
 			.primary(p.cur_tok().range, "");
-
+		kind_to_change = JS_UNKNOWN_BINDING;
```
MichaReiser, line 117: There were some more unknown bindings ;)

ematipico: Sorry @MichaReiser , I make this change later in another PR

PR #1806, opened 2021-11-22
Print empty slots in debug output

MichaReiser:
## Summary

We plan to use fixed offsets to access any child of a node. That requires, that the parser inserts "empty slots" if a child is missing (either because it's optional or because of a syntax error).

Our current means of testing the parser are the snapshot tests that use the alternative debug format to assert the tree structure. However, this format currently doesn't contain any output for empty slots.

This PR adds the empty slots to the alternate debug output of SyntaxNodes so that we can verify that the parser inserts missing slots where needed.

This PR introduces a new `eat_optional` for cases where one wants to eat an optional token if present or insert a missing slot otherwise.

Part of #1724 

## Test Plan

Verified rast output changes.


crates/rslint_parser/src/parser.rs
```diff
@@ -256,9 +267,7 @@ impl<'t> Parser<'t> {
 	/// Advances the parser by one token
 	pub fn bump_any(&mut self) {
 		let kind = self.nth(0);
-		if kind == SyntaxKind::EOF {
-			return;
-		}
```
MichaReiser, line 261: This is a bug in the caller. It should never happen that we try to eat the EOF token.

crates/rslint_parser/src/parser.rs
```diff
@@ -198,7 +209,7 @@ impl<'t> Parser<'t> {
 			_ => (),
 		}
 
-		if self.at_ts(recovery) {
+		if self.at_ts(recovery) || self.cur() == EOF {
```
MichaReiser, line 212: We don't want to create an empty error node (`bump_any` skipped the `EOF` token). So test if we're at EOF and if so, don't create an error node.

crates/rslint_parser/test_data/inline/err/bracket_expr_err.rast
```diff
@@ -1,12 +1,14 @@
 JS_ROOT@0..19
+  (empty-slot)
```
MichaReiser, line 2: This is the shebang token

crates/rslint_parser/test_data/inline/err/do_while_stmt_err.rast
```diff
@@ -1,39 +1,50 @@
...
+              (empty-slot)
+              (empty-slot)
+          (empty-slot)
+          (empty-slot)
+          (empty-slot)
+          (empty-slot)
+      (empty-slot)
+      (empty-slot)
+      (empty-slot)
+      (empty-slot)
```
MichaReiser, line 46: This looks a bit wild :D The errors are gone because of the EOF fix. 

The empty slots are for missing `(`, `)`, `while`... tokens

crates/rslint_parser/test_data/inline/err/bracket_expr_err.rast
xunilrj: Probably not possible, but there is a way to print the expected node?

```
JS_ROOT@0..19
  JS_SHEBANG@(empty-slot)
```

MichaReiser: Unfortunately not. The information which node means what is only available in the `AstNode`. 

Should we print the indices too: E.g. 

```
  slot-0: (empty)
  slot-1: LIST@
```

xunilrj (approved):

crates/rslint_parser/test_data/inline/err/bracket_expr_err.rast
xunilrj: hum... that seems interesting. Specially for the case below with multiple empties...

MichaReiser: Wdyt? I kind of like it and dislike it at the same time :D

ematipico: I like the numbered slots. It helps the reading, considering that sometimes the formatting is hard to read on github

crates/rome_rowan/src/cursor.rs
```diff
@@ -1524,6 +1538,127 @@ impl Iterator for PreorderWithTokens {
 	}
 }
 
+#[derive(Debug, Clone)]
+pub(crate) enum SyntaxSlot {
```
ematipico, line 1542: Documentation?

crates/rome_rowan/src/cursor.rs
```diff
@@ -1524,6 +1538,127 @@ impl Iterator for PreorderWithTokens {
...
+impl From<SyntaxElement> for SyntaxSlot {
+	fn from(element: SyntaxElement) -> Self {
+		match element {
+			SyntaxElement::Node(node) => SyntaxSlot::Node(node),
+			SyntaxElement::Token(token) => SyntaxSlot::Token(token),
+		}
+	}
+}
+
+pub(crate) struct SyntaxSlots<'a> {
```
ematipico, line 1557: Documentation

crates/rome_rowan/src/cursor.rs
```diff
@@ -1524,6 +1538,127 @@ impl Iterator for PreorderWithTokens {
...
+	parent: &'a SyntaxNode,
+	raw: Enumerate<Slots<'a>>,
+}
+
+impl<'a> SyntaxSlots<'a> {
+	fn map_item(&self, item: (usize, &Slot)) -> SyntaxSlot {
```
ematipico, line 1563: Documentation

crates/rome_rowan/src/cursor.rs
```diff
@@ -1524,6 +1538,127 @@ impl Iterator for PreorderWithTokens {
...
+}
+
+impl<'a> ExactSizeIterator for SyntaxSlots<'a> {
+	fn len(&self) -> usize {
+		self.raw.len()
+	}
+}
+impl<'a> FusedIterator for SyntaxSlots<'a> {}
+
+pub(crate) struct SlotsPreorder {
```
ematipico, line 1609: Documentation?

crates/rome_rowan/src/cursor.rs
```diff
@@ -1524,6 +1538,127 @@ impl Iterator for PreorderWithTokens {
...
+						None => WalkEvent::Leave(SyntaxSlot::Node(parent)),
+					}
+				}
+			})
+		});
+		next
+	}
+}
+
 #[derive(Debug, Clone)]
```
ematipico, line 1662: Would it be possible to add some unit test of the new features, please?

ematipico:
@MichaReiser I left some comments, could you please address them in your next PR?

crates/rome_rowan/src/cursor.rs
MichaReiser: I'm not sure how much-added value we get by adding additional unit tests, considering that it is tested extensively by all our integration tests.

crates/rome_rowan/src/cursor.rs
MichaReiser: This is an internal function. I don't think we need documentation for every single function in our code base

MichaReiser:
> @MichaReiser I left some comments, could you please address them in your next PR?

I added some brief documentation (considering that these are all crate internal types) in #1799

ematipico:
> > @MichaReiser I left some comments, could you please address them in your next PR?
> 
> I added some brief documentation (considering that these are all crate internal types) in #1799

Even though they are internal, in the past we struggled with internal code, where it was really difficult to understand some parts of it. As we're doing a rewrite, I promised myself that we should document any code that we write, so that any developer can have some decent context in case they need to extend/fix code that didn't write.

MichaReiser:
> > > @MichaReiser I left some comments, could you please address them in your next PR?
> > 
> > 
> > I added some brief documentation (considering that these are all crate internal types) in #1799
> 
> Even though they are internal, in the past we struggled with internal code, where it was really difficult to understand some parts of it. As we're doing a rewrite, I promised myself that we should document any code that we write, so that any developer can have some decent context in case they need to extend/fix code that didn't write.

That makes sense.

crates/rome_rowan/src/cursor.rs
ematipico: Unfortunately, I don't agree. In the past with RomeJS we struggled to understand various parts of the code base because of missing comments/documentation. I would like to avoid that now.

MichaReiser: We should probably have that discussion as a team because it seems we have different opinions here.

PR #1814, opened 2021-11-23
feat: unknown expression nodes

ematipico:
## Summary

- inside `expr.rs`, use the `JS_UNKNOWN_EXPRESSION` node
- removed the usage of `JS_UNKNOWN_BINDING` in favour of `JS_UNKNOWN_PATTERN` as highlighted in #1802 
- generated tests and updated existing ones
- removed `ERROR` from the rest of the recoveries, expect for typescript

## Test Plan

```
cargo lint
cargo test
cargo format
```


crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -209,7 +209,7 @@ pub(super) fn parameters_list(
 						T![...],
 						T![')'],
 					],
-					ERROR,
+					JS_UNKNOWN_EXPRESSION,
```
ematipico, line 212: I couldn't find a the piece of code that triggers this recovery, so I don't know if this unknown node is actually correct.

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -301,7 +304,7 @@ fn method_object_member_body(p: &mut Parser) -> Result<(), ()> {
 			.err_builder("expected a method definition, but found none")
 			.primary(p.cur_tok().range, "");
 
-		ParseRecoverer::with_error(BASE_METHOD_RECOVERY_SET, ERROR, err).recover(p);
+		ParseRecoverer::with_error(BASE_METHOD_RECOVERY_SET, JS_UNKNOWN_MEMBER, err).recover(p);
```
ematipico, line 307: I believe that this part, the `else` is never reached in the logic of the compiler because the check on the current character at line 296 is always done by the parent that calls this function. Because of this, I couldn't find the code that triggers this recovery error.

crates/rslint_parser/test_data/inline/err/block_stmt_in_class.rast
```diff
@@ -9,7 +9,7 @@
       2: L_CURLY@7..8 "{" [] []
       3: LIST@8..9
         0: JS_UNKNOWN_MEMBER@8..9
-          0: ERROR@8..9
+          0: JS_UNKNOWN_MEMBER@8..9
```
yassere, line 12: Is it intentional here that there's a `JS_UNKNOWN_MEMBER` nested inside a `JS_UNKNOWN_MEMBER`?

ematipico: For now, it is intentional. This a drawback of the current recovery logic, where it's really difficult to understand when/if a `CompletedMarker` is optional/required. 

Also, the change that we have here is the same change that we have in the [`class_decl_errr.rast`](https://github.com/rome/tools/pull/1814/files/bcb74e236ef4221ab77b847351517b633f1374b6#diff-b88160cbef70efa81d30797b0648898358053d41a2efccf464b767584e1d0161L39)

crates/rslint_parser/src/syntax/object.rs
MichaReiser: `method_object_member` doesn't perform this check but it's fine. I reworked that parsing already so that it does what we want :)

crates/rslint_parser/test_data/inline/err/block_stmt_in_class.rast
MichaReiser: I think it's fine if we revisit this with the new recovery logic in place.

MichaReiser (approved):
Looks good. Let's address the smaller remaining issues when we reworked our parser to use the new API from #1815

PR #1805, opened 2021-11-19
Separate Assignment target from Pattern

MichaReiser:
## Summary

Introduces new assignment target nodes that are used in places where a value is assigned to a target. Previously, RSLint reused patterns for assignment targets. 

The tricky part about assignment targets is that the parser only knows it is supposed to be an assignment target (and not any expression) after it reached the first assignment character. That's why the implementation has methods to convert from `expression` to `assignment target`. 

* Introduces new assignment target nodes
* Implements parsing for assignment target nodes
* Improves error recovery compared to the existing pattern parsing (now able to detect more invalid syntax, group invalid tokens together, and provide better error messages)

The implementation already adds "missing" slots whenever the primitives return `ParsedSyntax`. 

I also had to update the `for`, `for in` and `for of` grammar to account for that `for` allows any expression in the init whereas `for in` and `for of` only allows assignment targets. The grammar isn't final yet and will be revisited in #1725. We should also use this as an opportunity to clean up the parsing a bit.

## Test Plan

Added plenty of new tests and verified the rast output

PR #1789, opened 2021-11-16
feat: mark errors with "guessed" unknown nodes instead of errors

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/1759

This PR groups a number of other PRs that were already merged and reviewed. We removed the usage of `ERROR` in places the parser can't recover and marks nodes as `Unknown`. 

## Test Plan

Current CI

PR #1815, opened 2021-11-23
Parse and Error Recovery API

MichaReiser:
## Summary

The preliminary goal of this PR is to refine the `parse_*` and `recover_` APIs so that they guide developers in building correct syntax trees and avoid infinite loops. The secondary goal is to add means to improve error recovery and simplify adding `missing` slots if a child is missing.

### Correctness

The preliminary goal of this PR is to refine the API of the `parse_*` and `recover` methods to guide parser authors to create correct syntax trees. One shortcoming of the current API is that it's unclear for the caller when they must handle an error and when not because many methods return an `Option<CompletedMarker>` but it has a different meaning for different rule implementations

* a) Try to parse a node of type X
* b) Parse a node of type X, add an error if it's missing
* c) Parse a nod of type X, add an error and perform some error recovery

The problem is that handling a missing node is required for a) if the calling rule expects this to be a required parent but doesn't have to do anything in case of b) or c). There are other situations where there are three different variants of the same parse rule only to support the three different cases a), b), and c). 

Our API (and the compiler) should guide developers to do the necessary recovery when needed and provide means to propagate errors in case they can't handle the error on their own. It should further not be required to implement the same rule multiple times to support the different cases. That's why this PR introduces a new `ParsedSyntax` that must be handled and redefines the contract of parse rules. The contract of a parsing rule is:

* Returns `Present(completed)` if it was able to parse a (partial) node. Partial means, even if it means that it only parsed the `for (` head of a for statement and all other children are missing
* Returns `Absent` if the node wasn't present. The rule doesn't add any error in that case and the rule isn't allowed to progress the parser position. 

The PR further addresses the `recover` APIs (and unifies them) and enforce handling whatever the recovery is successful or not to avoid infinite loops. It does so by introducing a new `RecoveryResult` and a recovery function is only successful if:

* The parser isn't at the EOF
* The recovery consumed at least one token (it did some recovery)

### Missing slots

#1724 requires that the parser adds a "missing slot" for every optional or required child that isn't present in the source text. This requires that the `parse_*` methods expose the information if they parsed a node or not. 

This PR adds helper methods to the before introduced `ParsedSyntax` to accomplish that. 

* `make_required`: Adds an error and a missing slot if the parse method failed to parse the expected node, doesn't do anything otherwise
* `make_optional`: Adds a missing slot if the parse method didn't return a node 

It further introduces two helpers `precede_required`, and `precede_optional` that are useful if a parsing rule can only be parsed if another parse rule succeeded:

```
let lhs = expr(p)?;
let binary = lhs.precede_required(p); // inserts binary as a parent of lhs but only if lhs is present in the source
```
The benefit of this is that this avoids creating a marker that then must be explicitly `abandoned` in case the `lhs` isn't present in the source code.

### Recovery

A rule often doesn't know enough about its surroundings to decide for the best error recovery strategy. Rules may then be forced to only "eat" the next token and wrap it in an "unknown" node (if that is even allowed in that context). The problem is, there's no guarantee that the next token is valid in this context which has the result, that the parser will insert many diagnostics. 

Ideally, the parser groups as many invalid tokens as possible into a single `Unknown` node and only adds a single diagnostic. For example, an array expression can use a more aggressive recovery if it failed to parse the next element and can eat all tokens up to the next `,`, `]`, `}`, `;` into an `Unknown*` node. However, that only works if e.g. parsing an expression doesn't perform any error recovery as well. 

This is why this PR proposes to move error recovery to the call sites, that have the required context to perform good error recovery. 

### Conditional Syntax

There are different syntaxes that are only valid in a certain context:

* `with`: Loose mode only
* typescript: Typescript files only
* import/export: top of a module
* experimental syntax

The difficulty is that the parser must wrap such conditional syntax inside of an `Unknown`* node but unknown nodes don't exist for every node type. For example, the whole function declaration must be wrapped in an `UnknownStatement` if any parameter has a typescript type annotation. However, this can't be done in the `parse_parameter` rule because there's no `UnknownParameter` node type. That's why the rule must propagate the error to the caller until it reaches the `FunctionDeclaration` implementation that then can handle the case.

This PR introduces a `ConditionalParsedSyntax` that must be handled to address this need. It should only be returned by parse rules that may return conditional syntax (and can't convert the node to an `Unknown` node). 

## Usage
The PR rewrote some parsing rules to show how the API is intended to be used. I also rewrote the assignment target parsing to use the new API [in this commit](https://github.com/rome/tools/pull/1805/commits
/b9f616e634e39a6bf5f0942d7ca8cd5193f402bc) (part of #1805)

## Proposal

Rename the parsing rules from `assignment_expression` to `parse_assignment_expression`, etc..

## Examples

<details>
  <summary>Parsing a list with error recovery</summary>

https://github.com/rome/tools/blob/f57b7e347fb132e92b7ffba6406a1d30d5e4b770/crates/rslint_parser/src/syntax/object.rs#L23-L50
</details>

<details>
  <summary>Rule with conditional syntax</summary>

https://github.com/rome/tools/blob/7212349a4396e0086ee82b3cc08d533b531bb0c0/crates/rslint_parser/src/syntax/stmt.rs#L611-L632
</details>

<details>
  <summary>Type Script parse method</summary>
  
Nothing special, return a regular `ParsedSyntax`

https://github.com/rome/tools/blob/7212349a4396e0086ee82b3cc08d533b531bb0c0/crates/rslint_parser/src/syntax/function.rs#L163-L169

</details>

<details>
 <summary>JS with may contain TS</summary>

  https://github.com/rome/tools/blob/7212349a4396e0086ee82b3cc08d533b531bb0c0/crates/rslint_parser/src/syntax/function.rs#L44-L120

</details>

## Test Plan

`cargo test` and `cargo xtask coverage`


crates/rslint_parser/src/parser.rs
```diff
@@ -796,3 +806,49 @@ mod tests {
...
+		}
+	}
+
+	if empty {
+		list.abandon(p);
+		p.missing();
+		None
+	} else {
+		Some(list.complete(p, LIST))
+	}
```
MichaReiser, line 853: Inspirational and to show how to parse a list. Not part of the error recovery API design.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -86,10 +87,10 @@ pub(super) fn is_semi(p: &Parser, offset: usize) -> bool {
 /// A generic statement such as a block, if, while, with, etc
 pub fn stmt(p: &mut Parser, recovery_set: impl Into<Option<TokenSet>>) -> Option<CompletedMarker> {
 	let res = match p.cur() {
-		T![;] => empty_stmt(p),
-		T!['{'] => block_stmt(p, recovery_set).unwrap(), // It is only ever None if there is no `{`,
-		T![if] => if_stmt(p),
-		T![with] => with_stmt(p),
+		T![;] => empty_stmt(p).unwrap(), // It is only ever Err if there's no ;
```
MichaReiser, line 90: `unwrap` won't be needed once `stmt` returns `ParseResult`

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -427,15 +418,15 @@ pub(super) fn block_impl(
 		None
 	};
 
-	statements(p, false, true, recovery_set);
+	statements(p, false, true, None);
```
MichaReiser, line 421: We won't need this recovery set anymore, callers should perform the error recovery in place

crates/rslint_parser/test_data/inline/err/paren_or_arrow_expr_invalid_params.rast
```diff
@@ -67,6 +68,13 @@ error[SyntaxError]: Expected an expression, but found none
 1 │ (5 + 5) => {}
   │       ^ Expected an expression here
 
+--
+error[SyntaxError]: expected a function body but found ')' instead
+  ┌─ paren_or_arrow_expr_invalid_params.js:1:7
+  │
+1 │ (5 + 5) => {}
+  │       ^
+
```
MichaReiser, line 77: This file is actually a good example. We don't want to progress character by character when doing error recovery. Instead, the `ParameterList` should skip all tokens until it fins a safe token (`,`, `)` or maybe the start of another pattern).

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// Extension trait for `ParseResult<CompletedMarker>`
+pub trait ParsedSyntax {
+	/// Makes the node returned by the parse rule a required child and returns the completed marker as an [Option].
+	///
+	/// It returns `Some(completed)` if the parse rule successfully parsed a node (this is `Ok(completed)`)
+	///
+	/// It adds a missing marker, adds a diagnostic that it expected a node of [ExpectedNodeError], and
+	/// returns `None` if the parse rule failed to parse the node (this is `Err`).
+	fn make_required(self, p: &mut Parser) -> Option<CompletedMarker>;
```
MichaReiser, line 152: I'm not very fond of this name yet (and `make_optional`). I first went with `into_required` but the problem is that it also has side effects that users must be aware of. 

I then tried to use `p.required(parse_expression(p))` but the problem is that Rust doesn't permit that because `p` gets borrowed twice as `&mut` (even though this is probably safe). 

Let me know if you have any ideas to improve the naming.

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,363 @@
...
+/// or not. This may be because
+/// * Syntax is only available in strict or sloppy mode: For example, with statements
+/// * Syntax support depends on the file type: Typescript, JSX, Import / Export statements
+/// * Syntax depends on a newer language version: experimental features, private field existence test
+///
+/// See [CompletedMarker.requires_syntax_feature] that creates an `Err(Unsupported)` if the feature
+/// isn't available in the current parsing context.
+pub type ConditionalSyntaxParseResult = Result<CompletedMarker, ConditionalSyntaxError>;
+
+pub trait ConditionalParsedSyntax {
```
MichaReiser, line 340: I'm also not super happy with this name. Any suggestions?

crates/rslint_parser/src/parser.rs
```diff
@@ -349,12 +358,21 @@ impl<'t> Parser<'t> {
...
 			};
 
-			self.missing();
 			self.error(err);
 			false
 		}
 	}
 
+	/// Tries to eat a specific token kind, adds a missing marker and an error to the events stack if it's not there.
+	pub fn expect_required(&mut self, kind: SyntaxKind) -> bool {
```
MichaReiser, line 367: I introduced this new `expect_required` because we want to add an error message when the `,` token is missing when parsing array elements but we don't want to add a missing slot.

crates/rslint_parser/src/parser.rs
```diff
@@ -743,6 +762,31 @@ impl CompletedMarker {
...
+
+	pub fn to_parse_result(self) -> ParseResult {
+		Ok(self)
+	}
+
+	/// Converts this completed marker to a conditional parse result and adds
+	/// an error if the syntax is not supported in the current parsing context
+	pub fn to_conditional_parse_result<F, E>(
+		self,
+		p: &mut Parser,
```
ematipico, line 774: It's really weird to see a `self` that is a `Parser` to have a method that accepts a `Parser` as parameter :D 

But that's how ownership works!

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+/// first expected token isn't present, for example if the `for` keyword isn't present when calling
+/// `parse_for_statement`. There are other rules that can recover even if the
+/// first expected token isn't present. For example, the `parse_assignment_target_with_optional_default`
+/// can still parse `[ = "b"] = test` even tough the assignment target before the `=` is missing but
+/// it can recover once it sees the `=` token and create a `JsAssignmentTargetWithDefault` that has a
+/// missing `assignment_target`.
+///
+/// The parse rule must rewind the parser if it parsed an incomplete node but can't correctly determine its type
+/// to ensure that the caller and do the proper error recovery.
+pub type ParseResult = Result<CompletedMarker, ExpectedError>;
```
ematipico, line 39: My mind tells me that `UnexpectedError` works better for these kind of errors, even though the message says `expected {}`. Although `Expected*` work too. I'm sure we can find a good naming convention

crates/rslint_parser/src/parser.rs
MichaReiser: The `self` is a `CompletedMarker` here. We may want to move some types into their own files to make this clear.

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: That makes sense and are open for other suggestions. I wouldn't name it `Unexpected` because the error conveys the name of the expected node. Whatever part does the error recovery contributes the "unexpected tokens" which can be a single or multiple tokens.

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+
+	/// It creates a new marker that precedes the node returned by the parse rule.
+	///
+	/// It returns [CompletedMarker.precede] if this [Result] holds a parsed node.
+	/// It returns a new marker by calling [p.start()] if this [Results] doesn't contain a parsed node
+	/// and adds a missing marker. No diagnostic will be added if the node isn't present (since it's optional)
+	///
+	/// See [CompletedMarker.precede]
+	#[must_use]
+	fn precede_optional(self, p: &mut Parser) -> Marker;
```
ematipico, line 142: Here too, we leave the implementation of the method to the developer. Without context and help from the types, it will be really difficult to understand how the logic should be implemented. 

Do you think it's possible to change the return type?

crates/rslint_parser/src/parser.rs
ematipico: Damn, I just noticed the name of the file and noticed only now that it was inside `CompletedMarker`, thank you! 

Yeah I think having it inside a different file might help!

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+	fn into_unsupported(self) -> ConditionalSyntaxParseResult {
+		match self {
+			Ok(c) => Err(ConditionalSyntaxError::Unsupported(c)),
+			Err(expected) => Err(ConditionalSyntaxError::Expected(expected)),
+		}
+	}
+}
+
+#[derive(Debug)]
+pub enum ConditionalSyntaxError {
```
ematipico, line 265: I'm not super fond of the `ConditionalSyntax` prefix. It doesn't tell us what's that for. What about `LanguageFeature`?  Which is exactly what we are trying to support/handle. 

It's not just related to super languages such as TS or SASS. But also the new features of the language... JS, CSS, etc. get new features overtime that we can unlock/handle using a `LanguageFeature` naming convention

crates/rslint_parser/src/parser/parse_result.rs
ematipico: `LanguageFeatureParseResult`?

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+/// * Syntax depends on a newer language version: experimental features, private field existence test
+///
+/// See [CompletedMarker.requires_syntax_feature] that creates an `Err(Unsupported)` if the feature
+/// isn't available in the current parsing context.
+pub type ConditionalSyntaxParseResult = Result<CompletedMarker, ConditionalSyntaxError>;
+
+pub trait ConditionalParsedSyntax {
+	/// Converts this to a [ParsedSyntax] by converting any contained `Unsupported` node to the passed
+	/// in `unknown_kind`.
+	fn or_unsupported_to_unknown(self, p: &mut Parser, unknown_kind: SyntaxKind) -> ParseResult;
```
ematipico, line 305: Should it be `or_unknown_to_unsupported`? 🤔

ematipico (commented):
I like the proposed changes overall. I have been having hard time working with a `Option<CompletedMarker>` because it doesn't tell us when something is a real error or it is expected.

These changes will make things easier us when we want to work on the recover strategies.

I left some questions around the `trait` that are proposed. I believe we should tighten the types to  avoid to misuse the implementation of the traits.

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: I don't think users will ever implement this trait. The only reason why this trait exists is so that we can add extension methods to `Result<CompletedMarker, ParseError>`.

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+				Ok(recovered) => {
+					let diagnostic = error.into_diagnostic(p, recovered.range(p));
+					p.error(diagnostic);
+					Ok(recovered)
+				}
+
+				Err(_) => {
+					let diagnostic = error.into_diagnostic(p, p.cur_tok().range);
+					p.error(diagnostic);
+					Err(RecoveryError::Eof)
```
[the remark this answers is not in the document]
MichaReiser: What do you mean by redirecting? The errors have two different spans.

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: It converts a `unsupported` to an `unknown` node which matches the order in the name.

crates/rslint_parser/src/parser/parse_result.rs
ematipico: Sorry for the misunderstanding, when I said "user" I meant us developers 🤣 

I mean, the developer will need to implement this trait if they need to create a new parser. I just try to see myself or some other developer that didn't work on this feature and could have hard time.

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: I don't know of another type I could use here. What you want is a `Marker`. These methods aren't implemented by users. The trait only exists so that we can add extension methods to `Result`. 

The alternative is to not use `Result` and instead use our own `enum` but I noticed that I had to re-implement many of the `Result` methods (`ok`, `map`, `map_err`, ...) which is why I decided to use `Result` in the end.

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: This should be part of the parser infrastructure that is shared across all parsers. I don't see a reason why it needs to differ between languages. 

Can you elaborate with what part your struggling the most so that I can address it in the documentation?

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+
+/// Parse result for a parsing rule where the parsed syntax depends if a language feature is available
+/// or not. This may be because
+/// * Syntax is only available in strict or sloppy mode: For example, with statements
+/// * Syntax support depends on the file type: Typescript, JSX, Import / Export statements
+/// * Syntax depends on a newer language version: experimental features, private field existence test
+///
+/// See [CompletedMarker.requires_syntax_feature] that creates an `Err(Unsupported)` if the feature
+/// isn't available in the current parsing context.
+pub type ConditionalSyntaxParseResult = Result<CompletedMarker, ConditionalSyntaxError>;
```
MichaReiser, line 300: The benefit of separating `ConditionalSyntaxParseResult` and `ParseResult` is that most methods will return `ParseResult` which is easier to handle (`ConditionalSyntaxParseResult` requires to handle on the additional case). 

That's why I separated the two results.

crates/rslint_parser/src/parser/parse_result.rs
ematipico: Simply this: 

```rust
	fn make_required(self, p: &mut Parser) -> Option<CompletedMarker>; 										
	fn make_optional(self, p: &mut Parser) -> Option<CompletedMarker>;
```

> This should be part of the parser infrastructure that is shared across all parsers. I don't see a reason why it needs to differ between languages.

This is exactly my concern, having a trait that is shared across parsers where we have two methods that have the exact same signature (parameters and return type) and they differ only in naming and documentation. How can we make sure that `make_required` is used properly or not? Right now, we can't enforce it, and that's what worries me most.

ematipico: As alternative, we could pass to `make_required` a `Diagnostic

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -139,67 +149,72 @@ fn object_member(p: &mut Parser) -> Option<CompletedMarker> {
...
-					Some(m.complete(p, JS_PROPERTY_OBJECT_MEMBER))
+					Ok(m.complete(p, JS_PROPERTY_OBJECT_MEMBER))
 				} else {
-					Some(m.complete(p, JS_UNKNOWN_MEMBER))
+					// It turns out that this isn't a valid member after all. Make sure to throw
+					// away everything that has been parsed so far so that the caller can
+					// do it's error recovery
+					dbg!(p.cur_src());
+					m.abandon(p);
+					p.rewind(checkpoint);
```
MichaReiser, line 161: This may look a bit weird and has some performance implications as well. 

The current rules are that a parse rule isn't allowed to progress if it isn't able to parse a (partial) node. The problem here is, that we started parsing and only then figured out that this doesn't seem to be a valid member after all. 

An alternative to rewinding could be to add an `Incomplete(Marker)` Error to `ConditionalSyntax` so that we don't need to reset the parser and instead can continue from there. 

I don't think this will safe many cycles since recovery is pretty cheap (while loop testing on the next token and eating otherwise). It's also unclear what this would mean for errors. Let's say the parsing added errors up to this point (because the member was missing). Would we then end up having two errors: One for member missing and one that the whole thing should be a member but wasn't any?

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: That's a great suggestion. It would also make it cheaper to construct the `Err` nodes and only create the error when it is added to the parse context. 

The only downside is that it we need to repeat the "errors" for the same nodes in multiple places but I guess we can address this by creating some helpers

ematipico: Yeah I know! 😕 

Personally, it's a downside I'm willing to pay if it helps improving the usage and makes things more explicit

crates/rslint_parser/src/parser/parse_result.rs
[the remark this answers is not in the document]
ematipico: I mean  at line 231 we have `Err(_)`, and we are ignoring the error on purpose. Is it correct? What if the error is not of type `RecoveryError::Eof`? Why do we hard code `RecoveryError::Eof`?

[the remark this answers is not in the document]
MichaReiser: whoops. That was from before I introduced a second recovery error. Nice catch

crates/rslint_parser/src/parser/parse_result.rs
ematipico: Then, we can pass to `precede_required` a `Diagnostic`, which will be in line with `make_required`. What do you think?

MichaReiser: Yes, I'm currently making this change tough I don't plan to pass a diagnostic because we only want to build the diagnostic in case of an error (reduces the number of allocations). Same for `or_recover` tough I'm not sure yet if the error builder should be part of `ParseRecovery` or not.

crates/rslint_parser/src/parser/parse_recovery.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{CompletedMarker, Parser, TokenSet};
```
ematipico, line 1: This file in https://github.com/rome/tools/pull/1789 was named `parse_recoverer`. Should we decide which naming convention to use?

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: I hope you like it :)

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: Renamed it to `AbsentError`. We can also be more specific if we wish, e.g. naming it `AbsentNodeError`.

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,325 @@
...
+	/// The passed message is displayed in between: `expected {message} but instead found...`
+	pub fn new(message: &'static str) -> Self {
+		Self {
+			message,
+			primary: None,
+		}
+	}
+
+	/// Overrides the default primary message with a custom one
+	pub fn with_primary(mut self, primary: &'static str) -> Self {
```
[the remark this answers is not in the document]
MichaReiser: Deleted it for now since it isn't used but we can add a `primary` method to the `ExpectedNodeDiagnosticBuilder` once needed.

crates/rslint_parser/src/parser/parse_recovery.rs
MichaReiser: `recoverer` doesn't seem to be a real english word according to the oxford dictionary but `recovery` is 

https://www.oxfordlearnersdictionaries.com/spellcheck/english/?q=recoverer

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: > We could use the prefix wrap_, which might be the reverse of unwrap:

`unwrap` "unwraps" the Ok value of a result. Therefore, `wrap` would "wrap" a `CompletedMarker` as a `ParseResult` (what `Ok(completed)` does. That's why I believe that `wrap` might be misleading.

crates/rslint_parser/src/parser/parse_recovery.rs
ematipico: Great, I will change it in my PR!

MichaReiser: We actually may need both for now since they have slightly different behaviour. We should probably sunset one after we finished the migration from the old recovery to the new recovery.

crates/rslint_parser/src/parser/parse_result.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// Parse result for a parsing rule where the parsed syntax depends if a language feature is available
+/// or not. This may be because
+/// * Syntax is only available in strict or sloppy mode: For example, with statements
+/// * Syntax support depends on the file type: Typescript, JSX, Import / Export statements
+/// * Syntax depends on a newer language version: experimental features, private field existence test
+///
+/// See [CompletedMarker.requires_syntax_feature] that creates an `Err(Unsupported)` if the feature
+/// isn't available in the current parsing context.
+pub type ConditionalSyntaxParseResult = Result<CompletedMarker, ConditionalSyntaxError>;
```
MichaReiser, line 267: I'm a bit inclined to actually change this to a struct that also stores the feature it belongs to (or condition or however we want to call it) so that we can then implement `make_required` and `make_optional` that then only apply for that specific feature. Or, to change the `Error` options to `Absent(Supported|Unsupported)`,  `Unsupported(completed)` so that it's possible to distinguish "absent" from "absent" and the feature isn't available. The conversion wouldn't change, meaning that absent would still convert to `AbsentError` since this doesn't hurt. 

A possible use case for this is TypeScript syntax where we, for example, parse the return type of a function which is optional in the case of `TypeScript`, but should error in JS.  That's why it would be nice if we could write something like:

```
completed.requires_feature(p, TypeScript, JsErrors::typescript_syntax)
   .make_required(p, TsErrors::expected_something)
````

I need to explore this in more detail tomorrow.

File: crates/rslint_parser/src/syntax/js_parse_errors.rs [hunk 19]
```diff
@@ -0,0 +1,66 @@
+use crate::parser::{ParseErrors, ToDiagnostic};
+use crate::Parser;
+use rslint_errors::Diagnostic;
+use std::ops::Range;
+
+/// Provides factory function to create common diagnostics for the JavaScript syntax
+pub(crate) struct JsParseErrors;
+
+impl JsParseErrors {
```
yassere, line 9: Same question as with `ParseErrors`. Is there an advantage to using singletons rather than just organizing these functions using modules?

yassere (commented):
I don’t have much hands-on experience working with these parse methods so I don’t have much feedback, but this does seem like a good idea to me.

Do you think it’s necessary for ergonomics that the parse methods return an actual `Result` rather than a custom enum where things like `Absent` and `Unsupported` are simply variants?  It feels a bit odd that a parse method that doesn’t find an optional node would consider that an error state and return an `Err(AbsentError)` (if I’m understanding it correctly, so please correct me if I'm wrong). But if bubbling up errors using `?` is going to be common here, then that makes sense.

MichaReiser:
> I don’t have much hands-on experience working with these parse methods so I don’t have much feedback, but this does seem like a good idea to me.
> 
> 
> 
> Do you think it’s necessary for ergonomics that the parse methods return an actual `Result` rather than a custom enum where things like `Absent` and `Unsupported` are simply variants?  It feels a bit odd that a parse method that doesn’t find an optional node would consider that an error state and return an `Err(AbsentError)` (if I’m understanding it correctly, so please correct me if I'm wrong). But if bubbling up errors using `?` is going to be common here, then that makes sense.

I agree on the ergonomics and my first version actually used custom enums but I had to reimplement many of the generic Result methods like 'ok', 'map' etc. 

The try operator can be useful in some rare cases but I don't think it's super important here.

Let me revisit the result/custom enum decision tomorrow.

I also came to the conclusion that we most certainly want a custom enum for 'ConditionalSyntax'

crates/rslint_parser/src/lib.rs
```diff
@@ -245,3 +245,18 @@ impl From<FileKind> for Syntax {
...
+
+/// Syntax feature that tests if the current scope runs in sloppy / loose mode.
+#[doc(alias = "LooseMode")]
+pub struct SloppyMode;
+
+impl SyntaxFeature for SloppyMode {
+	fn is_available(&self, p: &mut Parser) -> bool {
+		p.state.strict.is_none()
+	}
+}
```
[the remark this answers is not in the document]
MichaReiser: I like having a trait so that we can share this infrastructure with other languages without the need for a shared enum. But we can have a JS specific enum

crates/rslint_parser/src/parser/parse_error.rs
```diff
@@ -0,0 +1,89 @@
...
+	/// Creates a diagnostic saying that the node [name] was expected at range
+	pub fn expected_node(name: &str, range: Range<usize>) -> ExpectedNodeDiagnosticBuilder {
+		ExpectedNodeDiagnosticBuilder::with_single_node(name, range)
+	}
+
+	/// Creates a diagnostic saying that any of the nodes in [names] was expected at range
+	pub fn expected_any(names: &[&str], range: Range<usize>) -> ExpectedNodeDiagnosticBuilder {
+		ExpectedNodeDiagnosticBuilder::with_any(names, range)
+	}
+}
```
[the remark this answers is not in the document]
MichaReiser: Mainly discoverability and fewer named imports. What do you prefer?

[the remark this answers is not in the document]
yassere: > Mainly discoverability and fewer named imports.

I think you can get these same advantages by importing the module and using the module name to qualify function calls: `parse_error::expected_node(name, range)` vs `ParseErrors::expected_node(name, range)`

You can use modules (inline or otherwise) in order to group functions as granularly as you want, so I'd probably prefer that approach.

crates/rslint_parser/src/parser/parse_recovery.rs
MichaReiser: I'll probably rename the "existing" one to `SingleTokenParseRecovery` as that's what it does.

crates/rslint_parser/src/parser/parse_result.rs
MichaReiser: I updated the PR and extended the documentation. I think `ConditionalParsedSyntax` is a good name because it is more generic than just supporting specific language features. It means, that this syntax can be valid or invalid depending on the parsing context. This can be because of a syntax feature or some other condition.

crates/rslint_parser/src/lib.rs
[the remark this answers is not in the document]
MichaReiser: Done

crates/rslint_parser/src/parser/parse_error.rs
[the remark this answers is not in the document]
MichaReiser: Makes sense and gives the parsing rule author the option to choose explicit imports vs using the module prefix.

Done

MichaReiser:
I reworked the PR following @yassere suggestions

* Replaced `ParseResult` with `ParsedSyntax` which is our own enum. Eliminates the need for an `AbsentError` and using `Present(completed)` and `Absent` in parse rules certainly gives more information than `Ok(completed)` and `Err(AbsentError)`. 
* Replaced `ConditionalSyntaxParseResult` with our own enum `ConditionalParsedSyntax`, mainly for the same reasons as above 
* Introduced a new `InvalidParsedSyntax` for syntax that is invalid in the current parsing context. 

Thank you all for your feedback. The API feels now much better and more consistent than with what I initially started.

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -0,0 +1,385 @@
...
+		match self {
+			Present(syntax) => Some(syntax),
+			Absent => {
+				p.missing();
+				let diagnostic = error_builder(p, p.cur_tok().range);
+				p.error(diagnostic);
+				None
+			}
+		}
+	}
```
ematipico, line 82: I just love it

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -0,0 +1,385 @@
...
+			None => Absent,
+		}
+	}
+}
+
+/// A parsed syntax that is only valid in some parsing contexts but not in others.
+/// One use case for this is for syntax that is only valid if the parsing context supports
+/// a certain language feature, for example:
+///
+/// * Syntax that is only supported in strict or sloppy mode: for example, with statements
```
ematipico, line 297:
```suggestion
/// * Syntax that is only supported in strict or sloppy mode: for example, `with` statements
```

crates/rslint_parser/test_data/inline/err/function_decl_err.js
```diff
@@ -5,3 +5,4 @@ async function() {}
 async function *() {}
 function *foo() {}
 yield foo;
+function test(): number {}
```
ematipico, line 8: It seems that this `function_decl_err.js`file is not inside any comments of the codegen. (`// test_err function_decl_err`), hence not created by the command `cargo xtask codegen`. It seems that I have missed it. We should do it later.

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -234,6 +234,7 @@ pub fn check_lhs(p: &mut Parser, expr: JsAnyExpression, marker: &CompletedMarker
 }
 
 /// Check if the var declaration in a for statement has multiple declarators, which is invalid
+#[allow(deprecated)]
```
ematipico, line 237: Should we add a TODO here? I presume we will need to remove this at some point

ematipico (approved):
Overall, this new changes are refreshing and will make the usage of the parser better once we get used to it.

Now that we have an infrastructure that allows us to understand when we have errors, optional children and unsupported language features, I think all these features should live inside their own crate. Mainly because this will be used by other parsers.

Imagine creating a JSON parser where we can use `ConditionalParsedSyntax` to support comments inside a JSON file.

I would suggest the following changes once the PR is merged:

- create a new crate called `parser_core`
- move `ConditionalParsedSyntax`, `ParsedSyntax`, `RecoveryError`, `ParseRecovery`, `parse_error.rs` and `ExpectedNodeDiagnosticBuilder` inside `parser_core`
- move all markers (and probably events too) inside `parser_core`
- - document `parser_core` and create a small example (in a `README.md` file or inside a doc comment) of how to use the new parsing infrastructure 
- create a trait called `ParserCore` which will contain all the existing functions for parsing a file (`eat`, `expect`, `at`, etc.)
- `rslint_parser` will become a `js_parser` and will implement `ParserCore`

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -198,6 +201,7 @@ fn class_members(p: &mut Parser) -> CompletedMarker {
 	members.complete(p, LIST)
 }
 
+#[allow(deprecated)]
```
xunilrj, line 204: is this really deprecated? ```class``` is calling this, right?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -226,10 +230,10 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
 	if declare && !has_access_modifier {
 		// declare() and declare: foo
 		if is_method_class_member(p, offset) {
-			literal_member_name(p); // bump declare as identifier
+			literal_member_name(p).ok().unwrap(); // bump declare as identifier
```
xunilrj, line 233: Should we leave this ```unwrap()``` here? If yes, why the ```.ok()```?

crates/rslint_parser/src/parser/parse_recovery.rs
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// TODO: Add a `recover_until` which recovers until the parser reached a token inside of the recovery set
+	// or the passed in `parse_*` rule was able to successfully parse an element.
+
+	/// Tries to recover by parsing all tokens into an `Unknown*` node until the parser finds any token
+	/// specified in the recovery set, the EOF, or a line break (depending on configuration).
+	/// Returns `Ok(unknown_node)` if recovery was successful, and `Err(RecoveryError::Eof)` if the parser
+	/// is at the end of the file (before starting recovery).
+	pub fn recover(&self, p: &mut Parser) -> RecoveryResult {
```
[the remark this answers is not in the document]
MichaReiser: I don't think so but I might be wrong. My hope is that we can delete the whole `state.no_recovery` thing but there might be a use case where we still need it and we should then add the check here as well.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: `ParsedSyntax` doesn't have an `unwrap` method. We need the `ok` to first convert it into an `Option`. 

We can remove the `unwrap` once the enclosing function returns a `ParsedSyntax` to. I just didn't want to rewrite all parse rules in this commit :D

MichaReiser:
> Overall, this new changes are refreshing and will make the usage of the parser better once we get used to it.
> 
> Now that we have an infrastructure that allows us to understand when we have errors, optional children and unsupported language features, I think all these features should live inside their own crate. Mainly because this will be used by other parsers.
> 
> Imagine creating a JSON parser where we can use `ConditionalParsedSyntax` to support comments inside a JSON file.
> 
> I would suggest the following changes once the PR is merged:
> 
>     * create a new crate called `parser_core`
> 
>     * move `ConditionalParsedSyntax`, `ParsedSyntax`, `RecoveryError`, `ParseRecovery`, `parse_error.rs` and `ExpectedNodeDiagnosticBuilder` inside `parser_core`
> 
>     * move all markers (and probably events too) inside `parser_core`
> 
>     * * document `parser_core` and create a small example (in a `README.md` file or inside a doc comment) of how to use the new parsing infrastructure
> 
>     * create a trait called `ParserCore` which will contain all the existing functions for parsing a file (`eat`, `expect`, `at`, etc.)
> 
>     * `rslint_parser` will become a `js_parser` and will implement `ParserCore`

Moving out the core logic makes sense but I don't think I'll tackle this right now because i don't know all the constraints yet. But what I started doing is to move `Js` specific features into the `syntax` folder.

crates/rslint_parser/src/syntax/util.rs
MichaReiser: I inlined all `allow` next to the problematic statement.

Yes, this should go away when we apply the changes to the AST that @jamiebuilds suggests

MichaReiser: and the main for now is that we don't add any new uses... I'm less concerned about the existing ones.

PR #1816, opened 2021-11-24
xtask to download and parse js libs

xunilrj:
The current version is as simple as possible. Download and parse every file at ```xtask\src\libs\libs.txt```. If the file is already downloaded, just use the cache, which is at ```target``` folder.

Currently we can't parse ```jquery``` and ```vue```. The only two examples I pre-selected.

# How to run

```
cargo xtask  coverage-libs
   Compiling xtask v0.0.0 (C:\github\rome\tools\xtask)
    Finished dev [unoptimized + debuginfo] target(s) in 3.17s
     Running `target\debug\xtask.exe coverage-libs`
jquery.min.js - using target\jquery.min.js
thread 'main' panicked at 'Marker must either be `completed` or `abandoned` to avoid that children are implicitly attached to a markers parent.', C:\Users\xunil\.cargo\registry\src\github.com-1ecc6299db9ec823\drop_bomb-0.1.5\src\lib.rs:113:13
vue.global.prod.js - using target\vue.global.prod.js
thread 'main' panicked at 'Marker must either be `completed` or `abandoned` to avoid that children are implicitly attached to a markers parent.', C:\Users\xunil\.cargo\registry\src\github.com-1ecc6299db9ec823\drop_bomb-0.1.5\src\lib.rs:113:13
end
```

### DIFF
--- a/xtask/src/libs/mod.rs
+++ b/xtask/src/libs/mod.rs
@@ -1,0 +1,39 @@
+use std::{path::PathBuf, str::FromStr};
+
+pub fn run() {
+	let out = PathBuf::from_str("target").unwrap();
+	let libs = include_str!("libs.txt").lines();
+	for lib in libs {
+		let url = url::Url::from_str(lib).unwrap();
+		let filename = url.path_segments().unwrap().last().unwrap();
+
+		let mut file = out.clone();
+		file.push(filename);
+
+		let code = match std::fs::read_to_string(&file) {
+			Ok(code) => {
+				println!("{} - using {}", filename, file.display());
+				code
+			}
+			Err(_) => {
+				println!("{} - Downloading {} to {}", filename, lib, file.display());
+				match ureq::get(lib).call() {
+					Ok(response) => {
+						let mut reader = response.into_reader();
+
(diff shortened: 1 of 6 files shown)


xtask/src/libs/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+					Ok(response) => {
+						let mut reader = response.into_reader();
+
+						let _ = std::fs::remove_file(&file);
+						let mut writer = std::fs::File::create(&file).unwrap();
+						let _ = std::io::copy(&mut reader, &mut writer);
+
+						std::fs::read_to_string(&file).unwrap()
+					}
+					Err(_) => todo!(),
```
MichaReiser, line 30: Can we add a more meaningful message here? Like: Download of filename failed and include some status code or whatnot.

MichaReiser (approved):
Looks good and it reveals some more parser bugs, nice! :)

xtask/src/libs/mod.rs
xunilrj: Done.

ematipico:
Do you think it's worth considering adding this task to CI?

PR #1818, opened 2021-11-24
feat: correct unknown nodes on partial recovery

ematipico:
## Summary

Small PR to address the last part of the recovery usage.

## Test Plan

```
cargo lint
cargo test
cargo format
```

### DIFF
--- a/crates/rslint_parser/src/parser.rs
+++ b/crates/rslint_parser/src/parser.rs
@@ -10,9 +10,6 @@ use std::borrow::BorrowMut;
 use std::cell::Cell;
 use std::ops::Range;
 
-use crate::parse_recovery::ParseRecovery;
-use crate::syntax::decl::BASE_METHOD_RECOVERY_SET;
-use crate::syntax::expr::EXPR_RECOVERY_SET;
 use crate::*;
 
 /// An extremely fast, error tolerant, completely lossless JavaScript parser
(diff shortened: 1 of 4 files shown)


crates/rslint_parser/src/parser.rs
```diff
@@ -467,14 +468,11 @@ impl<'t> Parser<'t> {
 		};
 
 		if self.at(T![;]) {
-			let m = self.start();
 			let err = self
 				.err_builder("expected an expression, but found `;` instead")
 				.primary(self.cur_tok().range, "");
+			ParseRecovery::with_error(token_set![], unknown_syntax_kind, err).recover(self);
```
MichaReiser, line 474: This feels overkill to me. 

My understanding of the recovery logic is:
* create a new node and stuff the `;` in there
* Add an error

The trick of an empty token set is required to force the recovery to always eat the `;`.  Overall, the old logic was easier to understand for me than what we have now, except there is some hidden gem that I don't see.

ematipico: The old logic is more understandable but you wouldn't know that there we are recovering.  

That's is basically doing what `ParseRecovery` is for; I would prefer to use `ParseRecovery` because it will allow use to do better refactors.

I can keep to old code but I think a recovery should be done by `ParseRecovery` and not manually, on the loose, using the parser. Doing so would risk the make the code harder to maintain.

MichaReiser: Maybe not worth a long discussion since this code may go away anyway. But I think I can tell that this code is doing a recovery from its enclosing method. Also, the error message is giving it away, isn't it ;)

The main argument I see is that it's easier to find the code that is doing error recovery.

Please write some documentation of what's going on here, especially explaining the empty token set trick.

ematipico: I will restore the old code, as it seems that this logic is causing some misunderstanding.

As explained in the description, **the parser is not recovering here**.

MichaReiser (approved):

PR #1821, opened 2021-11-25
"Short" guide on how to author parsing rules

MichaReiser:
## Summary

Start writing a guide on how to implement parsing rules. This isn't intended to be the final version. We want to revisit the guide after we finished our changes to the parser.

### DIFF
--- a/crates/rslint_parser/docs/authoring_parse_rules.md
+++ b/crates/rslint_parser/docs/authoring_parse_rules.md
@@ -1,0 +1,227 @@
+# Authoring Parse Rules
+
+This is a short, or not so short, guide on implementing parse rules using the Rome parser infrastructure.
+
+## Naming
+The convention is to prefix your parse rule with `parse_` and then use the name defined in the grammar file.
+
+For example, `parse_for_statement` or `parse_expression`.
+
+## Signature
+Most parse rules take a mutual reference to the parser as their only parameter and return a `ParsedSyntax`.
+
+```rust
+fn parse_rule_name(&mut: Parser) -> ParsedSyntax {}
+```
+
+You're free to add additional parameters to your function if needed. There are rare cases where you want to consider returning `ConditionalParsedSyntax` as explained in [conditional syntax](#conditional-syntax)
+
+
+## Parsing a single node
+
+Let's assume you want to parse the JS `if` statement:
+
+```
+JsIfStatement =
+ if
+ (
+ test: JsAnyExpression
+ )
+ consequent: JsBlockStatement
+ else_clause: JsElseClause?
+```
+
+### Presence Test
+
+Now, the parsing function must first test if the parser is positioned at an `if` statement and return `Absent` if that's not the case.
+
+```rust
+if !p.at(T![if]) {
+ return Absent;
+}
+```
+
+Why return `Absent`? The function must return `Absent` if the rule can't predict by the next token(s) if they form the expected node or not. Doing so allows the calling rule to decide if this is an error and perform an error recovery if necessary.  The second reason is to ensure that the rule doesn't return a node where all children are missing.
+
+Your rule implementation may want to consider more than just the first child to determine if it can parse at least some of the expected children.
+For example, the if statement rule could test if the parser is located at an `else` clause and then create an `if` statement where all children are missing except the `else` clause:
+
+```rust
+if !p.at(T![if]) && !p.at(T![else]){
+  return Absent
+}
+```
+
+Your implementation can also call into another parsing rule if the first child is a node and not a token.
+
+```rust
+let assignment_target = parse_assignment_target(p);
+
+if assignment_target.is_absent() {
+  return Absent;
+}
+
+let my_node = assignment_target.precede_or_missing();
+```
+
+But be careful with calling other rules. Your rule mustn't progress the parser if it returns `Absent`.
+
+
+### Parse children
+The parse rules will guide you in how to write your implementation and the parser infrastructure provides the following convenience APIs:
+
+* Optional token `'ident'?`: Use `p.eat_optional(token)`. It eats the next token if it matches the passed-in token. Adds a missing marker if the token isn't present in the source code.
+* Required token `'ident'`: Use`p.expect_required(token)`. It eats the next token if it matches the passed-in token. It adds an `Expected 'x' but found 'y' instead` error and a missing marker if the token isn't present in the source code.
+* Optional node `body: JsBlockStatement?`: Use`parse_block_statement(p).or_missing(p)`. It parses the block if it is present in the source code and adds a missing marker if it isn't.
+* Required node `body: JsBlockStatement`: Use `parse_block_statement(p).or_missing_with_error(p, error_builder)`: It parses the block statement if it is present in the source code and adds a missing marker and an error if not.
+
+Using the above-described rules result in the following implementation for the `if` statement rule.
+
+```rust
+fn parse_if_statement(p: &mut Parser) -> ParsedSyntax {
+ if !p.at(T![if]) {
+  return Absent;
+ }
+
+ let m = p.start();
+
+ p.expect_required(T![if]);
+ p.expect_required(T!['(']);
+ parse_any_expression(p).or_missing_with_error(p, js_parse_errors::expeced_if_statement);
+ p.expect_required(T![')']);
+ parse_block_statement(p).or_missing_with_error(p, js_parse_errors::expected_block-statement);
+ parse_else_clause(p).or_missing();
+
+ Present(m.complete(p, JS_IF_STATEMENT));
+}
+```
+
+Hold on, what are these *missing* markers? Rome's AST facade uses fixed offsets to retrieve a particular child from a node. For example, the 3rd child of the if statement is the condition. However, the condition would become the second element if the opening parentheses `(` isn't present in the source text. That's where missing elements come into play. Missing elements (added by calling `p.missing()`) represent placeholders for syntax that isn't present in the source text to guarantee that the children always appear in the same order.
+
+## Parsing Lists & Error Recovery
+
+Parsing lists is different from parsing single elements with a fixed set of children because it requires looping until the parser reaches a terminal token (or the end of the file).
+
+You may remember that `parse_*` methods shouldn't progress parsing if they return `Absent`. Not progressing the parser is problematic inside `while` loops because it inevitably results in an infinite loop.
+
+That's why you must do error recovery when parsing lists. Luckily, the parser comes with infrastructure to make error recovery a piece of cake.
+The general structure for parsing a list is (yes, that's something the parser infrastructure should provide for you):
+
+```rust
+let list = p.start();
+let mut first = true;
+
+while !p.at(T![']']) {
+ if first {
+  first = false;
+ } else {
+  p.expect(T![,]);
+
+  if p.at(T!['}']) {
+   break;
+  }
+ }
+
+ let recovered_element = parse_array_element(p).or_recover(
+  p,
+  ParseRecovery::new(JS_UNKNOWN_EXPRESSION, token_set![T![,], T![']'], T![;]])
+   .enable_recovery_on_line_break(),
+  js_parse_error::expected_array_element,
+ );
+
+ if recovered_element.is_err() {
+  break;
+ }
+}
+
+list.complete(p, LIST);
+```
+
+Let's run through this step by step:
+
+```rust
+let recovered_element = parse_array_element(p).or_recover(
+ p,
+ ParseRecovery::new(JS_UNKNOWN_EXPRESSION, token_set![T![,], T![']'], T![;]])
+  .enable_recovery_on_line_break(),
+ js_parse_error::expected_array_element,
+);
+```
+
+The `or_recover` performs an error recovery if the `parse_array_element` method returns `Absent`; there's no array element in the source text. The recovery eats all tokens until it finds one of the tokens specified in the `token_set`, a line break (if you called `enable_recovery_on_line_break`) or the end of the file. The recovery doesn't throw the tokens away but instead wraps them inside a `UNKNOWN_JS_EXPRESSION` node (first parameter). There exist multiple `UNKNOWN_*` nodes. You must consult the grammar to understand which `UNKNOWN*` node is supported in your case.
+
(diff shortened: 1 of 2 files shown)


crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+
+This is a short, or not so short, guide on implementing parse rules using the Rome parser infrastructure.
+
+## Naming
+The convention is to prefix your parse rule with `parse_` and then use the name defined in the grammar file.
+
+For example, `parse_for_statement` or `parse_expression`.
+
+## Signature
+Most parse rules take a mutual reference to the parser as their only parameter and return a `ParsedSyntax`.
```
xunilrj, line 11: mutual reference?

xunilrj (approved):

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+The `StrictMode.excluding_syntax` creates a conditional syntax that is invalid in strict mode and adds an error if the parser is currently in strict mode.
+
+You can convert the `ConditionalParsedSyntax` to a regular `ParsedSyntax` by calling `or_invalid_to_unknown`, which wraps the whole parsed `with` statement in an `UNKNOWN` node if the parser is in strict mode and otherwise returns the unchanged `with` statement.
+
+What if there's no `UNKNOWN` node matching the node of your parse rule? You must then return the `ConditionalParsedSyntax` without making the `or_invalid_to_unknown` recovery. It's then up to the caller to recover the potentially invalid syntax.
+
+
+## Summary
+
+* Pare rules are named `parse_rule_name`
```
mrkldshv, line 222: Small typo: parse rules :)

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
+# Authoring Parse Rules
+
+This is a short, or not so short, guide on implementing parse rules using the Rome parser infrastructure.
```
ematipico, line 3:
```suggestion
This is a short, or not so short, guide to implement parse rules using the Rome parser infrastructure.
```

crates/rslint_parser/docs/authoring_parse_rules.md
ematipico:
```suggestion
Most parse rules take a `&mut` reference to the parser as their only parameter and return a `ParsedSyntax`.
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+ else_clause: JsElseClause?
+```
+
+### Presence Test
+
+Now, the parsing function must first test if the parser is positioned at an `if` statement and return `Absent` if that's not the case.
+
+```rust
+if !p.at(T![if]) {
+ return Absent;
```
ematipico, line 40:
```suggestion
 return ParsedSyntax::Absent;
```

ematipico: Or alternatively, add: 

```rust 
use crate::ParsedSyntax::Absent;
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+}
+```
+
+Why return `Absent`? The function must return `Absent` if the rule can't predict by the next token(s) if they form the expected node or not. Doing so allows the calling rule to decide if this is an error and perform an error recovery if necessary.  The second reason is to ensure that the rule doesn't return a node where all children are missing.
```
ematipico, line 44:
```suggestion
Why return `ParsedSyntax::Absent`? The function must return `ParsedSyntax::Absent` if the rule can't predict by the next token(s) if they form the expected node or not. Doing so allows the calling rule to decide if this is an error and perform an error recovery if necessary.  The second reason is to ensure that the rule doesn't return a node where all children are missing.
```

ematipico: I would add an additional paragraph:

```md
In our example the parser can't find the `if` token, then we tell the caller that we don't have any `JsIfStatement`s.
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+let assignment_target = parse_assignment_target(p);
+
+if assignment_target.is_absent() {
+  return Absent;
+}
+
+let my_node = assignment_target.precede_or_missing();
+```
+
+But be careful with calling other rules. Your rule mustn't progress the parser if it returns `Absent`.
```
ematipico, line 67: Maybe we can expand of what `progress the parser` mean? I know what it means but new contributors might find it confusing.

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+
+```rust
+let assignment_target = parse_assignment_target(p);
+
+if assignment_target.is_absent() {
+  return Absent;
+}
+
+let my_node = assignment_target.precede_or_missing();
+```
```
ematipico, line 65: The previous paragraph doesn't respect the example. The paragraph talks about "being a node and not a token" but the example does an existence check (`is_absent()`).

I would add more description and I would add a small comment (maybe a reference to the correct paragraph) that explains what `precede_or_missing` is doing.

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+  return Absent;
+ }
+
+ let m = p.start();
+
+ p.expect_required(T![if]);
+ p.expect_required(T!['(']);
+ parse_any_expression(p).or_missing_with_error(p, js_parse_errors::expeced_if_statement);
+ p.expect_required(T![')']);
+ parse_block_statement(p).or_missing_with_error(p, js_parse_errors::expected_block-statement);
```
ematipico, line 92:
```suggestion
 parse_block_statement(p).or_missing_with_error(p, js_parse_errors::expected_block_statement);
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+ parse_else_clause(p).or_missing();
```
ematipico, line 93: Maybe we can add an inline comment there saying that we use `or_missing` because the `else {}` clause is optional in the JS grammar.

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+
+Hold on, what are these *missing* markers? Rome's AST facade uses fixed offsets to retrieve a particular child from a node. For example, the 3rd child of the if statement is the condition. However, the condition would become the second element if the opening parentheses `(` isn't present in the source text. That's where missing elements come into play. Missing elements (added by calling `p.missing()`) represent placeholders for syntax that isn't present in the source text to guarantee that the children always appear in the same order.
+
+## Parsing Lists & Error Recovery
+
+Parsing lists is different from parsing single elements with a fixed set of children because it requires looping until the parser reaches a terminal token (or the end of the file).
+
+You may remember that `parse_*` methods shouldn't progress parsing if they return `Absent`. Not progressing the parser is problematic inside `while` loops because it inevitably results in an infinite loop.
+
+That's why you must do error recovery when parsing lists. Luckily, the parser comes with infrastructure to make error recovery a piece of cake.
```
ematipico, line 107:
```suggestion
That's why you must do error recovery when parsing lists. Luckily, the parser comes with the infrastructure to make error recovery a piece of cake.
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+The general structure for parsing a list is (yes, that's something the parser infrastructure should provide for you):
+
+```rust
```
ematipico, line 110: Let's give some context about the example and say that we are doing. I guess the objective of the code is parsing a JS array?

Maybe a snippet of what we want to parse might be useful.

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+
+ Present(m.complete(p, JS_IF_STATEMENT));
+}
+```
+
+Hold on, what are these *missing* markers? Rome's AST facade uses fixed offsets to retrieve a particular child from a node. For example, the 3rd child of the if statement is the condition. However, the condition would become the second element if the opening parentheses `(` isn't present in the source text. That's where missing elements come into play. Missing elements (added by calling `p.missing()`) represent placeholders for syntax that isn't present in the source text to guarantee that the children always appear in the same order.
+
+## Parsing Lists & Error Recovery
+
+Parsing lists is different from parsing single elements with a fixed set of children because it requires looping until the parser reaches a terminal token (or the end of the file).
```
ematipico, line 103: I think we should precisely explain what "error recovery" means in this context. It is a concept that not all the parsers have and I believe how it works in our parser.

For example, the error recovery we had in RomeJS is different from what we have here.

Perhaps having a snippet might help:

```js
function foo() }}} {}
```

And we can explain that `}}}` are "unknown nodes", the parser will move forward until it will recover itself when it finds the first `{`. 

Also, I can see that you explain the concept after the example but without snippet it might be difficult. Maybe a "visual" example could help to better understand the strategy of the recovery.

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+```rust
+let recovered_element = parse_array_element(p).or_recover(
+ p,
+ ParseRecovery::new(JS_UNKNOWN_EXPRESSION, token_set![T![,], T![']'], T![;]])
+  .enable_recovery_on_line_break(),
+ js_parse_error::expected_array_element,
+);
+```
+
+The `or_recover` performs an error recovery if the `parse_array_element` method returns `Absent`; there's no array element in the source text. The recovery eats all tokens until it finds one of the tokens specified in the `token_set`, a line break (if you called `enable_recovery_on_line_break`) or the end of the file. The recovery doesn't throw the tokens away but instead wraps them inside a `UNKNOWN_JS_EXPRESSION` node (first parameter). There exist multiple `UNKNOWN_*` nodes. You must consult the grammar to understand which `UNKNOWN*` node is supported in your case.
```
ematipico, line 151:
```suggestion
The `or_recover` performs an error recovery if the `parse_array_element` method returns `Absent`; there's no array element in the source text. The recovery eats all tokens until it finds one of the tokens specified in the `token_set`, a line break (if you called `enable_recovery_on_line_break`) or the end of the file. The recovery doesn't throw the tokens away but instead wraps them inside a `JS_UNKNOWN_EXPRESSION` node (first parameter). There are multiple `UNKNOWN_*` nodes. You have to consult the grammar to understand which `UNKNOWN*` node is supported in your case.
```

ematipico: When talking about `token_set`, I think it might be helpful to refer to example above and explains why we passed those tokens. (perhaps a reference to a possible snippet might help too)

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+There are two reasons why the recovery can fail
+
+The parser reached the end of the file.
+The next token is one of the tokens specified in the recovery set, meaning: There is nothing to recover from.
+
+## Conditional Syntax
+
+The conditional syntax allows you to express that some syntax may not be valid in all source files. Some use cases are
```
ematipico, line 171:
```suggestion
The conditional syntax allows you to express that some syntax may not be valid in all source files. Some use cases are:
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+
+* Syntax that is only supported in strict or sloppy mode: for example, `with` statements
```
ematipico, line 173: Let's expand on `with` statement. Meaning that `with` statement is not valid when a JavaScript file uses `"use strict"` or is a module.

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+* Syntax that is only supported in certain file types: Typescript, JSX, Import / Export statements
```
ematipico, line 174:
```suggestion
* Syntax that is only supported in certain file types: Typescript, JSX, modules
```

crates/rslint_parser/docs/authoring_parse_rules.md
```diff
@@ -0,0 +1,227 @@
...
+* Syntax that is only available in specific language versions: experimental features
```
ematipico, line 175:
```suggestion
* Syntax that is only available in specific language versions: experimental features, different versions of the language e.g. (ECMA versions for JavaScript)
```

ematipico (approved):
Nice guide! 

I possible, I would expand a bit around the usage of `.or_missing_with_error(p, js_parse_errors::expeced_if_statement);` and explain how to create errors.

One important bit, I think, is to explain the usage of the `.precede_*` methods, with some example maybe.

We can add these bits later!

crates/rslint_parser/docs/authoring_parse_rules.md
ematipico: I think the recovery doesn't eat all the tokens at the moment. It eats only the next token. It will only when we will implement `recover_until`.

MichaReiser: It does... 

https://github.com/rome/tools/blob/22cd68b3fdacd2feb6933bceb2ba3746b8a93e2d/crates/rslint_parser/src/parser/parse_recovery.rs#L79-L81

Recover until would just be a neat helper so that we don't need to duplicate the tokens from the `parse_element` into the token set.

ematipico: Forgot about that change :) never mind then!

PR #1822, opened 2021-11-25
chore: rename parse methods and use new parser API

ematipico:
## Summary

This PR changes the signature of some methods from `Option<CompletedMarker>` to `ParsedSyntax`

It mainly tackles the parsing of parameters and statements, which are pretty much self contained.

- migrated function `formal_param_pat`. Although the error handling is still not correct (check the snapshot test, the range of the error is out of bounds) and I will leave it as is because I know @MichaReiser is working on patterns;
- created tests for the new error handling for setters, `set do() {}` inside classes and objects;
- renamed function `parameter_list`, no issues;
- migrated function `parameters_list`, no issues;
- migrated `arrow_body`, no issues;
- migrated `getter_object_member`, no issues;
- migrated `empty_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `block_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `if_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `with_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `while_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `for_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `do_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `switch_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `try_statement`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `return_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `break_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `continue_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `throw_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- migrated `debugger_stmt`, no issues; the method still uses `.ok().unwrap()` as temporary workaround;
- fixed error handling for `parse_switch_statement` and changed its logic. It doesn't use the range anymore and it's more simple to understand. Added a new test for error recovery in case `default` and `case` are missing;
- updated code inside the formatter because it was syntactically wrong, updated it with correct code;
- renamed all the methods  from `_stmt` to `_statement`

## Test Plan

```
cargo lint
cargo format
cargo test
```

Waiting for coverage result


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -555,7 +555,7 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
 
 					member_marker.complete(p, JS_GETTER_CLASS_MEMBER)
 				} else {
-					formal_param_pat(p);
+					parse_formal_param_pat(p).or_missing(p);
```
ematipico, line 558: @MichaReiser this is a case where the compiler warned where the `ParsedSyntax` needed to be handled. Is it the correct approach?

ematipico: I think that probably in this case the `.ok()` was the better approach 🤔

MichaReiser: You want to use `or_missing_with_error` here because it's an error if a setter has no parameter. 

`ok` should always never be used because it doesn't handle the error at all.

MichaReiser:
Would you mind explaining in more detail what the goal of this PR is. Which parse functions have now been completely migrated to the new parsing logic? 

I would recommend to only rename fully migrated functions to `parse_` so that we can track what uses the new logic and what isn't.

It's hard to do a proper review otherwise (and keep track of what has been done and what not).

ematipico:
> Would you mind explaining in more detail what the goal of this PR is. Which parse functions have now been completely migrated to the new parsing logic?
> 
> I would recommend to only rename fully migrated functions to `parse_` so that we can track what uses the new logic and what isn't.
> 
> It's hard to do a proper review otherwise (and keep track of what has been done and what not).

I expanded the description of the PR.

Please check the error handling part and please let me know if you're OK to leave it as is for the time being. It will be properly fixed once the `pattern()` function will be migrated.

MichaReiser:
> Please check the error handling part and please let me know if you're OK to leave it as is for the time being. It will be properly fixed once the `pattern()` function will be migrated.

The pattern part is fine. I'll probably rewrite it from scratch (by merging it with the assignment target implementation)

ematipico:
> > Please check the error handling part and please let me know if you're OK to leave it as is for the time being. It will be properly fixed once the `pattern()` function will be migrated.
> 
> The pattern part is fine. I'll probably rewrite it from scratch (by merging it with the assignment target implementation)

I don't think it's fine (maybe the error is somewhere else), please check this part of the snapshot:

```
--
error[SyntaxError]: expected a parameter but instead found '{'
  ┌─ setter_class_member.js:2:13
  │
2 │   set foo() {}
  │             ^ Expected a parameter here


```

As you can see the range is not correct, it seems that the parser has been advanced too much.

MichaReiser:
> > > Please check the error handling part and please let me know if you're OK to leave it as is for the time being. It will be properly fixed once the `pattern()` function will be migrated.
> > 
> > 
> > The pattern part is fine. I'll probably rewrite it from scratch (by merging it with the assignment target implementation)
> 
> I don't think it's fine (maybe the error is somewhere else), please check this part of the snapshot:
> 
> ```
> --
> error[SyntaxError]: expected a parameter but instead found '{'
>   ┌─ setter_class_member.js:2:13
>   │
> 2 │   set foo() {}
>   │             ^ Expected a parameter here
> ```
> 
> As you can see the range is not correct, it seems that the parser has been advanced too much.

I mean fine in... maybe broken but pattern parsing will change completely... that's why it's fine :)

ematipico:
Perfect, thank you!

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -555,7 +560,9 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
 
 					member_marker.complete(p, JS_GETTER_CLASS_MEMBER)
 				} else {
-					formal_param_pat(p);
+					// TODO: review error handling once the pattern functions is refactored
+					parse_formal_param_pat(p)
+						.or_missing_with_error(p, js_parse_error::expected_parameter);
```
MichaReiser, line 565: This is correct on your side. If the pattern code is inserting nodes even if it isn't able to parse a pattern than the bug is in `parse_formal_param_pat` and not how you call it! 

I'll have a look when working on patterns.

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -99,20 +99,19 @@ pub(super) fn formal_param_pat(p: &mut Parser) -> Option<CompletedMarker> {
 
 		kind = ASSIGN_PATTERN;
 	}
-
-	Some(m.complete(p, kind))
+	ParsedSyntax::Present(m.complete(p, kind))
 }
 
-pub(super) fn parameter_list(p: &mut Parser) -> CompletedMarker {
+pub(super) fn parse_parameter_list(p: &mut Parser) -> CompletedMarker {
```
MichaReiser, line 105: You mentioned in the summary that you migrated `parameters_list` but it still returns `ParsedSyntax`. OK with me, but can you remove it from the summary?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -781,11 +788,11 @@ fn constructor_class_member_body(p: &mut Parser, member_marker: Marker) -> Compl
...
 fn constructor_parameter_list(p: &mut Parser) -> CompletedMarker {
 	let m = p.start();
-	parameters_list(p, constructor_parameter);
+	parse_parameters_list(p, constructor_parameter);
 	m.complete(p, JS_CONSTRUCTOR_PARAMETER_LIST)
 }
 
-fn constructor_parameter(p: &mut Parser) -> Option<CompletedMarker> {
+fn constructor_parameter(p: &mut Parser) -> ParsedSyntax {
 	let m = p.start();
```
MichaReiser, line 796: Do I understand this correctly that all modifiers are parsed into this `m`. Can we rename it to `modifiers_marker`? 

Makes the code a bit more clearer.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -826,12 +833,12 @@ fn constructor_parameter(p: &mut Parser) -> Option<CompletedMarker> {
 
 	if !has_accessibility && !has_readonly {
 		m.abandon(p);
-		formal_param_pat(p)
+		parse_formal_param_pat(p)
 	} else {
-		if let Some(ref mut pat) = formal_param_pat(p) {
+		if let Some(ref mut pat) = parse_formal_param_pat(p).ok() {
```
MichaReiser, line 838: You can use 

```
if let Present(mut pat) = parse_formal_param_pat(p) {
    ...
}
```

No need to convert it to a `Option` first.

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -37,7 +37,7 @@ pub(super) fn formal_param_pat(p: &mut Parser) -> Option<CompletedMarker> {
 		pattern
 	} else {
 		m.abandon(p);
-		return None;
+		return ParsedSyntax::Absent;
```
MichaReiser, line 40: This won't be correct in case there was any typescript modifier. But we can leave it for now because it's patterns.

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -189,7 +188,7 @@ pub(super) fn parameters_list(
 		} else {
 			// test_err formal_params_no_binding_element
 			// function foo(true) {}
-			if let Some(res) = parse_param(p) {
+			if let Some(res) = parse_param(p).ok() {
```
MichaReiser, line 191: `if let Present(res) = parse_param(p) {`

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -819,7 +821,7 @@ pub fn paren_or_arrow_expr(p: &mut Parser, can_be_arrow: bool) -> CompletedMarke
 				}
 			}
 			p.expect_no_recover(T![=>])?;
-			arrow_body(p).ok()
+			parse_arrow_body(p).ok()
```
MichaReiser, line 824: `ok` is mainly there for compatibility but is rarely what you want. 

We're fairly confident that this is an arrow function at this point and an arrow function always must have a body. That's why you need to use `or_missing_with_err` here to add an error if the arrow function body is missing. 

Maybe add a test as well?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -933,7 +935,7 @@ pub fn expr(p: &mut Parser) -> Option<CompletedMarker> {
 
 /// A primary expression such as a literal, an object, an array, or `this`.
 pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
-	if let Some(m) = literal_expression(p) {
+	if let Some(m) = parse_literal_expression(p).ok() {
```
MichaReiser, line 938: `if let Present(m) = parse_literal_expression(p) {`

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -214,7 +221,8 @@ fn setter_object_member(p: &mut Parser) -> ParsedSyntax {
 	object_member_name(p).or_missing_with_error(p, js_parse_error::expected_object_member_name);
 
 	p.state.allow_object_expr = p.expect_required(T!['(']);
-	formal_param_pat(p);
+	// TODO: review error handling once the pattern functions is refactored
```
MichaReiser, line 224: This is correct. I don't think we need the TODO here. We just need to fix the `parse_formal_param_pat` implementation.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -387,14 +398,18 @@ pub fn continue_stmt(p: &mut Parser) -> CompletedMarker {
...
+pub fn parse_return_stmt(p: &mut Parser) -> ParsedSyntax {
 	// test_err return_stmt_err
 	// return;
 	// return foo;
+	if !p.at(T![return]) {
+		return Absent;
+	}
 	let m = p.start();
 	let start = p.cur_tok().range.start;
 	p.expect_required(T![return]);
```
MichaReiser, line 410: Nit: `bump` or `bump_any`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -387,14 +398,18 @@ pub fn continue_stmt(p: &mut Parser) -> CompletedMarker {
...
 	if !p.has_linebreak_before_n(0) && p.at_ts(STARTS_EXPR) {
+		// TODO: review this part and make sure it plays well with the new recovery logic
 		p.expr_with_semi_recovery(false);
```
MichaReiser, line 413: This is actually incorrect to my understanding. It adds an error if the parser is at a `;`. But `return;` is valid JS. 

This should be simply

```rust
!p.has_linebreak_before_n(0) {
    parse_expression(p).or_missing();
} else {
    p.missing();
}
```

But don't think this can easily be done when `expr` isn't returning `ParsedSyntax`

We certainly want the `p.missing()` in the `else` branch to make sure we add a missing marker if the `return` argument is missing.

Sitenote: We should probably have helpers `is_at_expression` instead of `p.at_ts(STARTS_EXPR)`.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -315,11 +321,13 @@ pub fn throw_stmt(p: &mut Parser) -> CompletedMarker {
...
+		return Absent;
+	}
 	let m = p.start();
 	let start = p.cur_tok().range;
-	p.expect_required(T![break]);
-
+	p.bump_any(); // break keyword
 	let end = if !p.has_linebreak_before_n(0) && p.at(T![ident]) {
 		let label_token = p.cur_tok();
 		p.bump_any();
```
MichaReiser, line 333: Err... not sure why GitHub doesn't allow me to comment on line 338. 

We need to add a `missing` marker if the label is missing to ensure we always have the same number of children.

Add `p.missing()` on line 338

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -340,7 +348,7 @@ pub fn break_stmt(p: &mut Parser) -> CompletedMarker {
 		p.error(err);
 	}
```
MichaReiser, line 349: You need to convert the `break` statement to a `JS_UNKNOWN_STATEMENT` if used outside of a `switch`, `loop`, or `labeled statement`.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -376,7 +387,7 @@ pub fn continue_stmt(p: &mut Parser) -> CompletedMarker {
...
 
-	m.complete(p, JS_CONTINUE_STATEMENT)
+	Present(m.complete(p, JS_CONTINUE_STATEMENT))
```
MichaReiser, line 390: Same as for break: Add missing marker for label. Return `Unknown_STMT` if used outside of a loop.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -407,13 +422,13 @@ pub fn return_stmt(p: &mut Parser) -> CompletedMarker {
 
 		p.error(err);
 	}
-	complete
+	Present(complete)
```
MichaReiser, line 425: You need to return a `UnknownStatement` if used outside of a function.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -630,7 +645,7 @@ pub fn if_stmt(p: &mut Parser) -> ParsedSyntax {
 }
```
MichaReiser, line 645: You must add a missing marker if the else clause is missing

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -630,7 +645,7 @@ pub fn if_stmt(p: &mut Parser) -> ParsedSyntax {
...
 
 /// A with statement such as `with (foo) something()`
-pub fn with_stmt(p: &mut Parser) -> ParsedSyntax {
+pub fn parse_with_stmt(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 648: Can we rename all the functions to match the name in the grammar

* `parse_if_stmt` -> `parse_if_statement`
* `parse_with_stmt` -> `parse_with_statement`
* ...

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1039,7 +1064,7 @@ fn switch_clause(p: &mut Parser) -> Option<Range<usize>> {
 //  case bar:
 //  default:
 // }
-pub fn switch_stmt(p: &mut Parser) -> CompletedMarker {
+pub fn parse_switch_stmt(p: &mut Parser) -> CompletedMarker {
```
MichaReiser, line 1067: * Should this return `ParsedSyntax`?
* We shouldn't allow for two `default` cases. It may be easier to pass `first_default` to `switch_clause` (parse_switch_clause) so that it then can create a `SwitchCase` where we stuff the whole `default` case into an `UnknownStatement`.
* We should do error recovery in case no case was found.

MichaReiser (changes requested):
Can we give this PR a more descriptive title? 
I know, it's hard because it changes many different methods. But even something like `use new parse API for simple statements`. 

* `parse_parameter_list` You only changed the name but it still returns a `CompletedMarker`. 

We should add tests for every migrated statement testing the valid and error cases. This also allows us to verify if we're adding the right amount of missing markers.

crates/rslint_parser/src/syntax/decl.rs
ematipico: It's still part of the work: rename methods and and use new API. I will update the description.

crates/rslint_parser/src/syntax/decl.rs
ematipico: I will add a TODO

crates/rslint_parser/src/syntax/decl.rs
ematipico: API changed

crates/rslint_parser/src/syntax/stmt.rs
ematipico: I'd prefer to leave it as it is, and review it in another PR. I think that `expr_with_semi_recovery` should go away, but not in this PR

crates/rslint_parser/src/syntax/stmt.rs
ematipico: > We should do error recovery in case no case was found.

You mean if no `case` or `default` was found?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1039,43 +1062,67 @@ fn switch_clause(p: &mut Parser) -> Option<Range<usize>> {
...
+						.err_builder(
+							"Multiple default clauses inside of a switch statement are not allowed",
+						)
+							.secondary(
+								default_found.range(&temp),
+								"the first default clause is defined here",
+							)
+							.primary(marker.range(&temp), "a second clause here is not allowed");
+
+						temp.error(err);
```
ematipico, line 1103: Here the error has to be added manually. We can't do `clause.or_missing_with_error` because `clause` is actually `Present`.

MichaReiser: Just changing it to a default statement isn't valid here. YOu then get the structure 

```
switch
 - unknown statement
```

I would pass the `first_default` to `parse_switch_clause` like this `parse_switch_clause(&mut *temp, first_default)`. 

The `parse_switch_clause can then test if the `default` clause is already present. If that's the case, then start a new case clause with:

* 'case' token missing
* test node missing
* consequent: The whole `default clause`.

ematipico: > Just changing it to a default statement isn't valid here

Could you tell us why?

I don't understand the final result.  Why do we need to mark a `case_clause` while it actually is a `default_clause`? And why do we need to mark the first two tokens as missing? I may understand the first one (`default` VS `case`), but why also the `:`? It's actually present, I can't understand why we need to mark it as missing.

Wouldn't be better like this?

```
switch
	- default_clause: 
		...
	- case_clause:
		- unknown // it's not missing; it exists but it's wrong
		- colon
		- unknown_statement
			- list
		- 
```

MichaReiser: There are only specific places where unknown nodes are allowed to appear and this is specified in our grammar. Unknown nodes are only allowed if you find a grammar rule of the format `left => JsUnknown*`. 

There's no such rule for `JsAnySwitchClause` that allows any `unknown` node. That's why having an unknown node as a direct child of `JsSwitchStatement` is invalid. 

We can argue whatever we should keep it as a `JsDefaultClause`. I believe that we should have a single rule what to do with nodes that aren't valid in this context and this is to wrap them as `Unknown` to make it clear that this syntax isn't valid. We need to do the same if we find a `with` statement in strict mode, `import/export` statements inside scripts or not at the top of the file.

I do like your suggestion to not put it into the statement but instead interpret the `default` as the `discriminant` of the `case` clause. I wasn't aware that JS allows expressions for case clause discriminates: 

```
- switch
  - default_clause:
    - ...
   - case_clause:
     - missing "catch" token
     - discriminant: 
       - JsUnknownExpression
         - default token
       - colon: ':'
       - body (if any)
```

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -99,20 +100,19 @@ pub(super) fn formal_param_pat(p: &mut Parser) -> Option<CompletedMarker> {
...
-	Some(m.complete(p, kind))
+	ParsedSyntax::Present(m.complete(p, kind))
 }
 
-pub(super) fn parameter_list(p: &mut Parser) -> CompletedMarker {
+pub(super) fn parse_parameter_list(p: &mut Parser) -> ParsedSyntax {
 	let m = p.start();
-	parameters_list(p, formal_param_pat);
-	m.complete(p, JS_PARAMETER_LIST)
+	parse_parameters_list(p, parse_formal_param_pat);
```
MichaReiser, line 108: Should this return `Absent` if the current token isn't `(`?

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -189,7 +189,7 @@ pub(super) fn parameters_list(
 		} else {
 			// test_err formal_params_no_binding_element
 			// function foo(true) {}
-			if let Some(res) = parse_param(p) {
+			if let Present(res) = parse_param(p) {
```
MichaReiser, line 192: Can we use our new error recovery here? 

```
let result = parse_param(p).or_recover(p, ParseRecovery::new(
    JS_UNKNOWN_PATTERN,
    token_set![
		T![ident],
		T![await],
		T![yield],
		T![,],
		T!['['],
		T![...],
		T![')'],
    ]).enable_recovery_on_line_break());

if let Ok(result) {
    if result.kind() == ASSIGN_PATTERN && p.state.in_binding_list_for_signature ... 
} else {
    break;
}
```

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -846,7 +850,7 @@ pub fn paren_or_arrow_expr(p: &mut Parser, can_be_arrow: bool) -> CompletedMarke
 			if params_marker.is_none() {
 				// Rewind the parser so we can reparse as formal parameters
 				p.rewind(checkpoint);
-				parameter_list(p);
+				parse_parameter_list(p).or_missing(p);
```
MichaReiser, line 853: Isn't the parameter list mandatory?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -975,7 +979,7 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
 					let m = p.start();
 					p.bump_remap(T![async]);
 					if p.at(T!['(']) {
-						parameter_list(p);
+						parse_parameter_list(p).or_missing(p);
```
MichaReiser, line 982: Isn't the parameter list mandatory? At the other hand, it can never be missing if we're at a `(` token. So `or_missing` should be fine (or even `ok()`)

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -99,7 +99,7 @@ fn function(p: &mut Parser, kind: SyntaxKind) -> ConditionalParsedSyntax {
 		type_parameters.or_missing(guard);
 	}
 
-	parameter_list(guard);
+	parse_parameter_list(guard).or_missing(guard);
```
MichaReiser, line 102: The parameter list is mandatory

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -318,7 +325,7 @@ fn method_object_member_body(p: &mut Parser) {
 	p.state.in_function = true;
 
 	ts_parameter_types(p);
-	parameter_list(p);
+	parse_parameter_list(p).or_missing(p);
```
MichaReiser, line 328: The parameter list is required.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -973,27 +1011,50 @@ pub fn for_stmt(p: &mut Parser) -> CompletedMarker {
...
+) -> ParsedSyntax {
 	let m = p.start();
 	match p.cur() {
 		T![default] => {
-			p.bump_any();
+			// in case we have two `default` expression, we mark the second one
+			// as `JS_CASE_CLAUSE` where the the "default" keyword is an Unknown node
+			let syntax_kind = if first_default.is_some() {
+				// we bump the "default" keyword but we mark it as unknown
+				p.bump_remap(JS_UNKNOWN_PATTERN);
```
MichaReiser, line 1029: This isn't valid. A token can never have the type `JS_UNKNOWN_PATTERN`. 

I liked your suggestion to create a `JS_CASE_CLAUSE` in that case where we interpret the `default` keyword as the `discriminant`. 

```
let case_kind = if first_default.is_some() {
    p.missing(); // Missing 'case' clause
    let discriminant = p.start();
    p.bump_any(); // interpret `default` as the discriminant
    discriminant.complete();
    JS_CASE_CLAUSE
} else {
    p.bump_any();
    JS_DEFAULT_CLAUSE
}
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1039,43 +1087,49 @@ fn switch_clause(p: &mut Parser) -> Option<Range<usize>> {
 //  case bar:
 //  default:
 // }
-pub fn switch_stmt(p: &mut Parser) -> CompletedMarker {
+pub fn parse_switch_statement(p: &mut Parser) -> ParsedSyntax {
 	// test_err switch_stmt_err
 	// switch foo {}
 	// switch {}
+	// switch { var i = 0 }
```
MichaReiser, line 1094: Can we add a test case for multiple `default` clauses?

crates/rslint_parser/test_data/inline/err/paren_or_arrow_expr_invalid_params.rast
```diff
@@ -147,6 +163,21 @@ error[SyntaxError]: Expected a statement or declaration, but found none
 2 │ (a, ,b) => {}
   │         ^^ Expected a statement or declaration here
 
+--
+error[SyntaxError]: Expected an expression, but found none
+  ┌─ paren_or_arrow_expr_invalid_params.js:4:1
+  │
+4 │
```
MichaReiser, line 170: This looks a bit weird. Do you know why it isn't pointing to any valid source? Is it because it is at the EOF?

crates/rslint_parser/test_data/inline/err/setter_class_member.rast
```diff
@@ -0,0 +1,51 @@
...
+  3: EOF@32..33 "" [Whitespace("\n")] []
+--
+error[SyntaxError]: Expected an identifier or pattern, but found none
+  ┌─ setter_class_member.js:2:11
+  │
+2 │   set foo() {}
+  │           ^
+
+--
+error[SyntaxError]: expected a parameter but instead found '{'
```
MichaReiser, line 35: This recovery looks weird but I guess it's becaues pattern contains some code to do error recovery which then moves the parser position till after the `)` (and adds two errors)

crates/rslint_parser/test_data/inline/err/switch_stmt_err.rast
```diff
@@ -54,6 +76,49 @@ error[SyntaxError]: expected `')'` but instead found `{`
...
+
+--
+error[SyntaxError]: expected a default, or a case but instead found 'var'
+  ┌─ switch_stmt_err.js:3:10
+  │
+3 │ switch { var i = 0 }
+  │          ^^^ Expected a default, or a case here
+
+--
+error[SyntaxError]: expected `'}'` but instead found `var`
```
MichaReiser, line 108: Do you know why we generate two errors pointing to the `var`? 

Is it because the var eats over the switch? 

Would it result in better error recovery if we wrap any statements inside of a `CASE_CLAUSE`?

Let's say we could do this (maybe not today, but in the future)

```
match p.cur() {
    T![default] => ... 
    T![case] => ...
    _ => {


        // We found some content that isn't a case nor default clause. Let's assume the author forgot the default/case or deleted it
        // Ideally, loop and eat all statements but eating a statement one by one works to

        let m = p.start();
        p.missing(); // case
        p.missing(); // discriminant
        p.missing(); // colon

        let recovered = statement(p).or_recover(p, ParseRecovery::new(JS_UNKNOWN_STATEMENT, token_set![T![default], T![case], T!['}']]));

        if recovered.is_err() {
            // Can only happen at the EOF or when we reached T![}] but this method should never be called for those positions
            // Maybe use `unreachable` instead?
            m.abandon();
            return None;
        }
    }
}
```

MichaReiser:
Can we give this PR a more descriptive title? 
I know, it's hard because it changes many different methods. But even something like `use new parse API for simple statements`. 

* `parse_parameter_list` You only changed the name but it still returns a `CompletedMarker`. 

We should add tests for every migrated statement testing the valid and error cases. This also allows us to verify if we're adding the right amount of missing markers.

MichaReiser (commented):
There are some more places where I believe we should use `or_missing_with_err` and the switch structure isn't correct yet (having the AST facade as part of the tests will help verify the correctness, especially #1829 )

crates/rslint_parser/test_data/inline/err/paren_or_arrow_expr_invalid_params.rast
ematipico: I strongly believe that the issue is the function `pattern`, which moves the parser forward too much

crates/rslint_parser/test_data/inline/err/setter_class_member.rast
ematipico: Exactly

crates/rslint_parser/test_data/inline/err/switch_stmt_err.rast
ematipico: Your suggestion was really useful! I managed to fix a couple of issues!

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -99,20 +100,23 @@ pub(super) fn formal_param_pat(p: &mut Parser) -> Option<CompletedMarker> {
...
 		kind = ASSIGN_PATTERN;
 	}
-
-	Some(m.complete(p, kind))
+	ParsedSyntax::Present(m.complete(p, kind))
 }
 
-pub(super) fn parameter_list(p: &mut Parser) -> CompletedMarker {
+/// parse the whole list of parameters, brackets included
+pub(super) fn parse_parameter_list(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 107: Not related to your changes but took me an awfully long time to spot the difference between `parse_parameter_list` and `parse_parameters_list`. 

Not sure what other names we could use for the generic one: Maybe `parse_parameter_list_impl`, `parse_parameter_list_common`. Or we rename `parse_parameter_list` to `parse_function_parameter_list` but that may be confusing because it then no longer matches the name in the grammar file.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -977,15 +981,18 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
...
+					let parsed_parameters = parse_parameter_list(p);
+					if parsed_parameters.is_absent() {
 						let m = p.start();
 						// test_err async_arrow_expr_await_parameter
 						// let a = async await => {}
 						p.bump_remap(T![ident]);
 						m.complete(p, JS_IDENTIFIER_BINDING);
+					} else {
+						parsed_parameters
+							.or_missing_with_error(p, js_parse_error::expected_parameters);
```
MichaReiser, line 993: You won't need the `or_missing_with_error` because the `if parsed_parameters.is_absent` already covers the case where it is missing.
```suggestion
					let parsed_parameters = parse_parameter_list(p);
					if parsed_parameters.is_absent() {
						let m = p.start();
						// test_err async_arrow_expr_await_parameter
						// let a = async await => {}
						p.bump_remap(T![ident]);
						m.complete(p, JS_IDENTIFIER_BINDING);
					} 
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -973,27 +1011,53 @@ pub fn for_stmt(p: &mut Parser) -> CompletedMarker {
...
 		T![default] => {
-			p.bump_any();
+			// in case we have two `default` expression, we mark the second one
+			// as `JS_CASE_CLAUSE` where the the "default" keyword is an Unknown node
+			let syntax_kind = if first_default.is_some() {
+				// missing "case" keyword
+				p.missing();
+				let discriminant = p.start();
+				p.bump_any(); // interpret `default` as the test of the case
+				discriminant.complete(p, IDENT);
```
MichaReiser, line 1032: This must be a `JsUnknownExpression`. The `discriminant` must be a `expression`.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -982,7 +986,8 @@ pub fn ts_non_array_type(p: &mut Parser) -> Option<CompletedMarker> {
 			}
 		}
 		JS_NUMBER_LITERAL | JS_STRING_LITERAL | TRUE_KW | FALSE_KW | JS_REGEX_LITERAL => Some(
-			literal_expression(p)
+			parse_literal_expression(p)
+				.ok()
 				.unwrap()
 				.precede(p)
```
MichaReiser, line 992: Nit: You can use `parse_literal_expression(p).precede_or_missing()`
```suggestion
			parse_literal_expression(p)
				.precede_or_missing(p)
```

crates/rslint_parser/test_data/inline/err/switch_stmt_err.rast
```diff
@@ -18,7 +18,77 @@
...
+          0: (empty)
+          1: (empty)
+          2: (empty)
+          3: JS_UNKNOWN_STATEMENT@54..65
+            0: VAR_KW@54..58 "var" [] [Whitespace(" ")]
+            1: IDENT@58..60 "i" [] [Whitespace(" ")]
+            2: EQ@60..62 "=" [] [Whitespace(" ")]
+            3: JS_NUMBER_LITERAL@62..63 "0" [] []
+            4: SEMICOLON@63..65 ";" [] [Whitespace(" ")]
+        1: JS_CASE_CLAUSE@65..80
```
MichaReiser, line 53: Nice!

MichaReiser (approved):
Looks all good. There are mostly a few nits. The only thing that needs to be fixed before landing is the type of the `discriminant` in the case of multiple default clauses.

crates/rslint_parser/src/syntax/decl.rs
ematipico: I would personally remove one of the functions. It's too confusing

ematipico:
Thank you @MichaReiser for your patience!

PR #1823, opened 2021-11-25
ci: split workflows for PRs and `main` branch

ematipico:
## Summary

This PR puts some order to our workflows:

- `pull_request.yml` is the workflow where we run on pull requests. This is where we might have some bot that does some fancy stuff, like the coverage;
- `main.yml` is the workflow that we trigger when something is pushed on `main` branch. It does usual checks like in the PR. It also run the coverage and prints it in the command line (different behaviour from the PR workflow). This will also set the foundation for other jobs that we want to run only in certain occasions, specifically those jobs that take long time;
- `test262.yml` stays the same that runs on pull requests only. I plan to merge it into `pull_request.yml` but only when I figure out a wait to save artifacts between workflows

## Test Plan

Current CI

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,10 @@
-name: CI
+
+# This workflow run when something is pushed on main and it does:
+  # - normal checks like in the normal PRs
+  # - runs the coverage and prints in the command line
+name: CI on main
 on:
   push:
-    branches:
-      - main
-  pull_request:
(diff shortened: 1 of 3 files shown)


MichaReiser:
Looks good. Would you mind changing the title to something explaining what "better workflows" mean? For example, split github workflows into separate files.

MichaReiser (approved):

PR #1826, opened 2021-11-26
Add helper for parsing identifiers

MichaReiser:
## Summary

Parsing identifiers is a super common task and there's currently no centralized helper that takes care of validating the identifier.

This PR adds a `parse_identifier` method that parses any node wrapping an identifier and makes sure it performs the necessary validations.

Fixes a couple more parsing errors :)

## Test Plan

Verified that the now passing tests are related to identifier parsing


crates/rslint_parser/src/syntax/pat.rs
```diff
@@ -68,7 +71,17 @@ pub fn pattern(p: &mut Parser, parameters: bool, assignment: bool) -> Option<Com
...
 				}
 			}
-			binding_identifier(p);
+
+			if let Present(mut identifier) = parse_identifier_binding(p) {
+				let mapped_ident_kind = match identifier.kind() {
+					JS_IDENTIFIER_BINDING => NAME,
+					JS_UNKNOWN_BINDING => JS_UNKNOWN_PATTERN,
+					k => k,
+				};
```
MichaReiser, line 80: This is a bit silly but is only needed temporarily until patterns are migrated to bindings.

crates/rslint_parser/test_data/inline/err/binding_identifier_invalid.rast
```diff
@@ -84,13 +84,6 @@ error[SyntaxError]: Illegal use of `yield` as an identifier in generator functio
 3 │    let yield = 5;
   │        ^^^^^
 
---
-error[SyntaxError]: Illegal use of `yield` as an identifier in strict mode
-  ┌─ binding_identifier_invalid.js:3:8
-  │
-3 │    let yield = 5;
-  │        ^^^^^
-
```
MichaReiser, line 93: The error is still there, we just don't add it twice anymore (see line 81-84)

crates/rslint_parser/src/syntax/pat.rs
ematipico: If it's temporary, can we add a `TODO` please?

ematipico (approved):

PR #1827, opened 2021-11-29
Assert AST Test in parser integration test

MichaReiser:
## Summary

We test our parser by comparing the expected syntax tree with the actual tree produced by the parser. This works well but doesn't give us any guarantee that what the parser produces matches the AST structure. 

This PR removes the derived `Debug` implementation for `AstNode` and replaces it with a generated implementation that prints the children of a node:

```rust
impl std::fmt::Debug for JsConstructorClassMember {
	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
		f.debug_struct("JsConstructorClassMember")
			.field(
				"access_modifier",
				&support::DebugOptionalNode(self.access_modifier()),
			)
			.field("name", &support::DebugSyntaxResult(self.name()))
			.field(
				"parameter_list",
				&support::DebugSyntaxResult(self.parameter_list()),
			)
			.field("body", &support::DebugSyntaxResult(self.body()))
			.finish()
	}
}
```

and adds the AST output to the parser's `ok` integration tests. 

This helps us 

* The format is easier to read than the syntax node output because it labels children
* It allows us to spot structural mismatches: New children will always be "missing" until the necessary parser changes are made. Mismatching field types also show up as `missing` (see included fix for class members named "static")
* Simplifies debugging for pass-authors that may be less familiar with our internal syntax level representation. 

Should we print the syntax and the ast tree: I think yes because:
* The syntax tree can be useful when debugging
* Changes that are only reflected in one may point out structural mismatches 

The PR doesn't add the AST assertions for failing tree tests yet because some tests are panicking because of `AstSeparatedList` children that have missing separators/elements. We'll ned to rework the `AstSeparatedList` to allow for empty elements and missing separators to be able to recover from `fn(a b)` (missing separator) or `fn(,a)` because we don't want to insert non existing tokens into the parse tree (fixing missing separators) nor empty elements (fixing missing elements). I plan to follow up on this in a new PR.

I also tried to wrap the printing with `std::panic::catch_unwind` but `Cell`s are not `UnwindSafe`. 

## Test Plan

I verified some of the AST tree outputs by hand.

### DIFF
--- a/crates/rome_formatter/src/ts/parameter_list.rs
+++ b/crates/rome_formatter/src/ts/parameter_list.rs
@@ -2,7 +2,7 @@ use crate::{
 	format_elements, group_elements, join_elements, soft_indent, soft_line_break_or_space, token,
 	FormatElement, FormatResult, Formatter, ToFormatElement,
 };
-use rslint_parser::ast::JsParameterList;
+use rslint_parser::ast::{JsAnyParameter, JsParameterList};
 
 impl ToFormatElement for JsParameterList {
 	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
@@ -16,5 +16,14 @@ impl ToFormatElement for JsParameterList {
 			),),
 			formatter.format_token(&self.r_paren_token()?)?
 		]))
+	}
+}
+
+impl ToFormatElement for JsAnyParameter {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		match self {
+			JsAnyParameter::Pattern(pattern) => pattern.to_format_element(formatter),
+			JsAnyParameter::JsRestParameter(_) => todo!("rest parameter"),
+		}
 	}
 }
--- a/crates/rslint_parser/src/ast.rs
+++ b/crates/rslint_parser/src/ast.rs
@@ -384,9 +384,10 @@ mod support {
 		AstNode, AstNodeList, AstSeparatedList, SyntaxElementChildren, SyntaxKind, SyntaxNode,
 		SyntaxToken,
 	};
-	use crate::ast::AstChildren;
+	use crate::ast::{AstChildren, JsAnyParameter};
 	use crate::SyntaxList;
 	use crate::{SyntaxError, SyntaxResult};
+	use std::fmt::{Debug, Formatter};
 
 	pub(super) fn node<N: AstNode>(parent: &SyntaxNode) -> Option<N> {
 		parent.children().find_map(N::cast)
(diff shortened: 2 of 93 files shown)


MichaReiser:
@xunilrj I guess there's no way to bypass the `UnwindSafe` for `Cell`?

crates/rome_formatter/src/ts/parameter_list.rs
```diff
@@ -18,3 +18,12 @@ impl ToFormatElement for JsParameterList {
 		]))
 	}
 }
+
+impl ToFormatElement for JsAnyParameter {
```
MichaReiser, line 22: Another bug this testing revealed: `JsParameters` was restricted to only allow for `Pattern` but not `JsRestParameter`.

crates/rslint_parser/src/ast.rs
```diff
@@ -435,6 +460,32 @@ mod support {
 		find_token(parent, possible_kinds)
 			.ok_or_else(|| SyntaxError::MissingRequiredChild(parent.clone()))
 	}
+
+	/// New-type wrapper to flatten the debug output of syntax result fields when printing [AstNode]s.
+	/// Omits the [Ok] if the node is present and prints `missing (required)` if the child is missing
+	pub(super) struct DebugSyntaxResult<N>(pub(super) SyntaxResult<N>);
```
MichaReiser, line 466: This is to flatten `Ok(JsUnaryExpression)` to just `JsUnaryExpression` or `missing (required)` in case it is missing. Not strictly necessary but I found having a less nested tree helps with readability

crates/rslint_parser/test_data/inline/ok/for_stmt.rast
```diff
@@ -1,3 +1,159 @@
...
+                                    expression: JsNumberLiteralExpression {
+                                        value_token: JS_NUMBER_LITERAL@13..14 "5" [] [],
+                                    },
+                                },
+                            },
+                            trailing_separator: None,
+                        },
+                    ],
+                },
+                semicolon_token: missing (required),
```
MichaReiser, line 33: This is fixed in #1805 The parser parses the `;` into the `For` stmt but the grammar incorrectly defines that they're part of the `ForStmtInit` or `ForStmtUpdate`

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -127,7 +127,7 @@ pub(super) fn parameters_list(
 			p.eat(T![,]);
 			break;
 		} else {
-			p.expect_required(T![,]);
+			p.expect(T![,]);
```
ematipico, line 130: What's this change about?

ematipico (approved):
I like this improvement!

crates/rslint_parser/src/syntax/decl.rs
MichaReiser: Let me undo this. The separated list in this version doesn't expect any missing children which `expect_required` adds. However, the new implementation actually requires "missing" slots for missing separators/nodes.

PR #1829, opened 2021-11-29
Allow missing tokens/nodes in AstSeparatedList

MichaReiser:
## Summary

I initially proposed that the `AstSeparatedList` should panic whenever a separator or node is missing. However, that means that the only way for the parser to recover from a missing separator (e.g. `a b`) or from a missing node (`,a`) is to create an empty token (COMMA with "text": "") or an empty element.

Having these empty tokens and nodes breaks with the convention used in the rest of the AST facade where missing required elements return a `SyntaxError::MissingRequiredChild`.

This PR changes the `AstSeparatedListElement` to store the node and separator tokens as `SyntaxResult<T>`s and removes the `panics` in the implementation.

This allows us now to also add the `Ast` snapshot to the failing dir tests.

## Test Plan

Updated the `AstSeparatedList` tests. Verified that the error tests now print AST trees.

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -622,6 +622,7 @@ impl<L: Language> SyntaxNode<L> {
 		}
 	}
 
+	/// Returns an iterator over all the slots of this syntax node.
 	pub fn slots(&self) -> SyntaxSlots<L> {
 		SyntaxSlots {
 			raw: Some(self.raw.slots()),
@@ -1112,10 +1113,16 @@ impl<L: Language> From<SyntaxElement<L>> for cursor::SyntaxElement {
 	}
 }
 
+/// Each node has a slot for each of its children regardless if the child is present or not.
+/// A child that isn't present either because it's optional or because of a syntax error
+/// is stored in an [SyntaxSlot::Empty] to preserve the index of each child.
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum SyntaxSlot<L: Language> {
+	/// Slot that stores a node child
 	Node(SyntaxNode<L>),
+	/// Slot that stores a token child
 	Token(SyntaxToken<L>),
+	/// Slot that marks that the child in this position isn't present in the source code.
 	Empty,
 }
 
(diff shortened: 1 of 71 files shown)


crates/rslint_parser/src/syntax/pat.rs
```diff
@@ -75,13 +75,18 @@ pub fn pattern(p: &mut Parser, parameters: bool, assignment: bool) -> Option<Com
...
-			#[allow(deprecated)]
-			SingleTokenParseRecovery::with_error(ts, JS_UNKNOWN_PATTERN, err).recover(p);
-			return None;
+			if parameters {
+				ts = ts.union(token_set![T![,], T![')']])
+			}
+			p.error(err);
+			return match ParseRecovery::new(JS_UNKNOWN_PATTERN, ts).recover(p) {
+				Ok(recovered) => Some(recovered),
+				Err(_) => None,
```
MichaReiser, line 88: This is a temporary workaround. Ideally, pat shouldn't perform any error recovery here. We'll change this when reworking patterns.

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -137,9 +137,9 @@ pub fn check_for_stmt_lhs(p: &mut Parser, expr: JsAnyExpression, marker: &Comple
 			}
 		}
 		JsAnyExpression::JsArrayExpression(expr) => {
-			let elem_count = expr.elements().len();
+			let elem_count = expr.elements().iter().count();
 
-			for (idx, elem) in expr.elements().iter().enumerate() {
+			for (idx, elem) in expr.elements().iter().flatten().enumerate() {
```
MichaReiser, line 142: The whole code in here will go away when we rewrite the parsing of for statement.

crates/rome_rowan/src/api.rs
```diff
@@ -1096,6 +1112,74 @@ impl<L: Language> From<SyntaxElement<L>> for cursor::SyntaxElement {
 	}
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum SyntaxSlot<L: Language> {
```
ematipico, line 1116: Should we document all these new APIs?

ematipico (approved):

PR #1839, opened 2021-12-01
Rename Pattern to Bindings and share parsing logic with Assignment Target

MichaReiser:
Sorry, it seems this escalated quite a bit. I wasn't aware, that patterns are used in so many different places in our codebase

## Summary

#1805 separated assignment targets out from patterns. This PR now renames the "remaining" patterns to bindings. Bindings are different from assignment targets because they bind a new name in the current scope and assign a value to it rather than assigning a value to an existing name.  

This PR further restructures Patterns as specified in #1719. 

* Renames `Pattern` to `JsAnyBinding`
* Replaces `AssignPattern` with `JsBindingWithDefault`
* Deletes `PatternOrExpr`
* Replace `SinglePattern` with `JsIdentifierBinding`
* Replace `RestPattern` with `JsArrayRestBinding` and `JsObjectRestBinding`. Remove it from the `Pattern` union (only permitted inside of arrays/objects)
* Rename `ObjectPattern` to `JsObjectBinding`
* Replace `KeyValuePattern` with `JsShorthandPropertyBinding` and `JsPropertyBinding`
* Rename `ArrayPattern` to `JsArrayBinding`
* Remove `ExprPattern`
* Unify `SpreadElement` and `JsSpread`
* Deleted `InitializedProp` (not valid JS nor TS syntax???)

This PR extracts the common logic for parsing `Patterns` from `assignment_target` as traits into `pattern.rs` that I then used to parse bindings.

Part of #1725 

## Test Plan

Verified coverage, added new test cases.

All libs are now parsing without panicking. Performance is roughly the same


crates/rome_formatter/src/ts/bindings/array_binding.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl ToFormatElement for JsArrayBinding {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		let l_bracket = formatter.format_token(&self.l_brack_token()?)?;
+		let elements = formatter.format_separated(self.elements())?;
+		let r_bracket = formatter.format_token(&self.r_brack_token()?)?;
+
+		Ok(format_elements![group_elements(format_elements![
+			l_bracket,
+			join_elements(space_token(), elements),
```
ematipico, line 15: Should we use the same code we use for array elements?

```rust
			soft_indent(join_elements(soft_line_break_or_space(), elements)),
			if_group_breaks(token(",")),
```

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -47,6 +47,16 @@ impl ParsedSyntax {
 		}
 	}
 
+	pub fn map<F>(self, mapper: F) -> ParsedSyntax
```
ematipico, line 50: Can we document this new API, please? Also, I wouldn't call it only "map" because it's not a real map, it doesn't transform the input to "any" output.  But if this is the idiomatic pattern, I would describe it in the documentation

crates/rslint_parser/src/syntax/assignment_target.rs
```diff
@@ -105,92 +106,27 @@ pub(crate) fn parse_simple_assignment_target(
...
-
-	elements.complete(p, LIST);
-	p.expect(T![']']);
-
-	Present(m.complete(p, JS_ARRAY_ASSIGNMENT_TARGET))
 }
 
+struct ArrayAssignmentTarget;
+
 // test array_assignment_target_rest
```
ematipico, line 130: Are the comments still valid? I mean, should we move them next to the function that does the job?

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -0,0 +1,273 @@
+use crate::parser::{expected_any, ToDiagnostic};
+#[allow(deprecated)]
+use crate::syntax::class::parse_equal_value_clause;
```
ematipico, line 3: `TODO` to remove it?

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -145,7 +78,7 @@ pub(super) fn parse_parameters_list(
 				p.error(err);
 				let m = p.start();
 				p.bump_any();
-				m.complete(p, JS_UNKNOWN_PATTERN);
+				m.complete(p, JS_UNKNOWN_BINDING);
```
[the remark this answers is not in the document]
ematipico: `TODO`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -539,15 +541,28 @@ pub(super) fn any_reference_member(p: &mut Parser) -> Option<CompletedMarker> {
...
+		parse_reference_identifier_member(p).ok()
 	}
 }
 
-fn reference_identifier_member(p: &mut Parser) -> Option<CompletedMarker> {
-	identifier_name(p).map(|mut ident| {
-		ident.change_kind(p, JS_REFERENCE_IDENTIFIER_MEMBER);
-		ident
-	})
+pub(super) fn is_at_reference_identifier_member(p: &Parser) -> bool {
```
ematipico, line 548: Small doc?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -539,15 +541,28 @@ pub(super) fn any_reference_member(p: &mut Parser) -> Option<CompletedMarker> {
...
+	p.at(T![ident]) || p.cur().is_keyword()
+}
+
+pub(super) fn parse_reference_identifier_member(p: &mut Parser) -> ParsedSyntax {
```
ematipico, line 552: Small doc?

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -156,14 +156,6 @@ fn parse_object_member(p: &mut Parser) -> ParsedSyntax {
 				parse_method_object_member_body(p);
 				Present(m.complete(p, JS_METHOD_OBJECT_MEMBER))
 			} else if let Some(mut member_name) = member_name {
-				// test object_expr_assign_prop
-				// let b = { foo = 4, foo = bar }
```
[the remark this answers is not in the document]
ematipico: No, it's not a valid syntax

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -0,0 +1,306 @@
+///! Provides traits for parsing pattern like nodes
```
ematipico, line 1: Should we expand on what kind of patterns these traits are meant to parse?

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -0,0 +1,306 @@
...
+
+/// Trait for parsing an array like pattern of the form `[a, b = "c", { }]`
+pub(crate) trait ParseArrayPattern<P: ParseWithDefaultPattern> {
+	/// The kind of an unknown pattern. Used in case the pattern contains elements that aren't valid patterns
+	fn unknown_pattern_kind() -> SyntaxKind;
+	/// The kind of the array like pattern (array assignment or array binding)
+	fn array_pattern_kind() -> SyntaxKind;
+	/// The kind of the rest pattern
+	fn rest_pattern_kind() -> SyntaxKind;
+	/// Creates a diagnostic saying that the parser expected an element at the passed in position
```
ematipico, line 48:
```suggestion
	/// Creates a diagnostic saying that the parser expected an element at an arbitrary position, passed as input
```

ematipico (commented):
It's a lot of code and it will take time for me to review. For now I leave some feedback and question:

- we should add new tests to the formatter for array binding. It seems we don't have them, other than `simple.js`;
- why did you use traits in this case? what's the endgame? The documentation doesn't explain that part and I can see you implemented the traits just once, which means that it could have been a `struct` or an `enum`;

MichaReiser:
> It's a lot of code and it will take time for me to review. For now I leave some feedback and question:

Thanks for taking the time! 

>     * we should add new tests to the formatter for array binding. It seems we don't have them, other than `simple.js`;

I'm not sure if that code is final. I just moved the files from the `pattern` to the `bindings` folder and renamed them. I don't know why git doesn't recognize the moves. I would leave it for now because this PR isn't touching the formatter code and we need to revisit all implementations anyway.

>     * why did you use traits in this case? what's the endgame? The documentation doesn't explain that part and I can see you implemented the traits just once, which means that it could have been a `struct` or an `enum`;

The traits are implemented twice: Once for `Bindings` and once for `AssignmentTarget`. For example, `ObjectAssignmentTarget` and `ArrayAssignmentTarget` both implement the `ParseObjectPattern` trait.

crates/rome_formatter/src/ts/bindings/array_binding.rs
MichaReiser: Probably, but I just moved this logic from `ArrayPattern`. Will leave it, for now, to not change even more behaviour.

crates/rslint_parser/src/parser/parsed_syntax.rs
MichaReiser: Sure! And you're right. Renamed it to `and_then` to match the method on [`Result`](https://doc.rust-lang.org/std/result/enum.Result.html#method.and_then)

crates/rslint_parser/src/syntax/assignment_target.rs
MichaReiser: The implementation is now shared between bindings and assignment targets. I still want to test both because they have slightly different behaviour. 

I moved it to `rest_pattern_kind` so that they're closer where they are used.

crates/rslint_parser/src/syntax/binding.rs
MichaReiser: Uh, not sure how that deprecated made it in there. Probably because I removed the `SingleTokenParseRecovery` use and it now moved to the `parse_equal_value_clause`.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: What would you write in this documentation (same for `parse_reference_identifier_member`. I can only think of

> Tests if the parser is positioned at a reference identifier member

or 

> Parses a reference identifier member

which doesn't provide more information than the function names. Especially since `parse_*` is now a common pattern and `is_at_*` will be more common too.

crates/rslint_parser/src/syntax/pattern.rs
MichaReiser: I can mention `array`, `object`, and `with_default` patterns but I guess that's already mentioned by the traits and will only become outdated if we change any of them.

crates/rslint_parser/src/syntax/expr.rs
ematipico: I would describe/simplify the "reference identifier member" part. Also, I don't know what `is_keyword` does, which makes me think what is method doing.

MichaReiser: Hmm... the `reference_identifier_member` is a name of an AST node. Yes, we should rename that node but we haven't found a better name yet. But that's why I can't simplify the method name.

`is_keyword` tests if that token is a `keyword`. For example, `for` is a valid member even tough it is a keyword.

MichaReiser: This is the corresponding grammar: 

https://github.com/rome/tools/blob/914f0c2dd9087db42f5c24a3215f28a60d1aa8b1/xtask/js.ungram#L955-L961

And we have a name now... `JsName`, keep it simple :) Put we should to this in a separate PR

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -0,0 +1,285 @@
...
+	// test rest_property_binding
+	// let { ...abcd } = a;
+	// let { b: { ...a } } = c;
+	//
+	// test_err rest_property_binding_err
+	// let { ... } = a;
+	// let { ...c = "default" } = a;
+	// let { ...{a} } = b;
+	// let { ...rest, other_assignment } = a;
+	// let { ...rest, } = a;
```
ematipico, line 265: Let's add a test for 
```js
let {  ...await } = a;
```
It should blow for incorrect use of the keyword `await`

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -0,0 +1,285 @@
...
+		if p.at(T![...]) {
+			let m = p.start();
+			p.bump(T![...]);
+
+			let inner = parse_binding(p);
+
+			if let Present(mut inner) = inner {
+				if inner.kind() != JS_IDENTIFIER_BINDING {
+					inner.change_kind(p, JS_UNKNOWN_BINDING);
+					p.error(p.err_builder("Expected identifier binding").primary(inner.range(p), "Object rest patterns must bind to an identifier, other patterns are not allowed."))
```
ematipico, line 276: I wonder if we could suggest a way to fix the error based on what we have? Like, if we know we have brackets, to remove them; if it's a number to replace it with a correct name, etc.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -67,31 +73,34 @@ fn class(p: &mut Parser, kind: ClassKind) -> CompletedMarker {
...
+				if guard.typescript() && DISALLOWED_TYPE_NAMES.contains(&text) {
+					let err = guard
+						.err_builder(&format!(
+							"`{}` cannot be used as a class name because it is already reserved as a type",
+							text
+						))
+						.primary(id.range(&*guard), "");
+
+					guard.error(err);
+				}
```
ematipico, line 92: Do we really need typescript here? I mean, can we remove it for now?

crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -195,8 +129,8 @@ pub(super) fn parse_parameters_list(
 			// function (a++, c) {}
 			let recovered_result = parse_param(p).or_recover(
 				p,
-				ParseRecovery::new(
-					JS_UNKNOWN_PATTERN,
+				&ParseRecovery::new(
```
ematipico, line 132: I see that you changed it! Good :) 
I had to change too

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -0,0 +1,306 @@
...
+
+		let m = p.start();
+		let rest_end = p.cur_tok().range.end;
+		p.bump(T![...]);
+
+		let with_default = self.pattern_with_default();
+
+		with_default
+			.parse_pattern(p)
+			.or_missing_with_error(p, |p, _| P::expected_pattern_error(p, rest_end..rest_end));
```
ematipico, line 140: Is `rest_end...rest_end` right? 🤔

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -0,0 +1,306 @@
...
+					Err(RecoveryError::Eof) => break,
+					Err(RecoveryError::AlreadyRecovered) => {
+						guard.error(Self::expected_property_pattern_error(
+							guard,
+							guard.cur_tok().range,
+						));
+						break;
+					}
+					_ => {}
+				}
```
ematipico, line 212: I don't understand this part. If `recover_result` is `Err`, it will never reach this `match` because it will hit the `break;`. Also, why marking the error *again*? I think `or_recover` already does that.

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -0,0 +1,306 @@
...
+		p.bump(T![=]);
+
+		if let Ok(recovered) = recovery.recover(p) {
+			recovered.undo_completion(p).abandon(p); // append recovered content to parent
+		}
+		p.error(
+			p.err_builder("rest element cannot have a default")
+				.primary(
+					default_start..p.cur_tok().range.start,
+					"Remove the default value here",
```
ematipico, line 285: The default should also contain the `=`. The snapshot shows that the diagnostic has only the range of the default value.

I mean, if we remove only the default value, we still have invalid syntax because also the `=` should be gone.

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -0,0 +1,306 @@
...
+	} else if p.at(T![,]) && p.nth_at(1, end_token) {
+		p.error(
+			p.err_builder("rest element may not have a trailing comma")
+				.primary(p.cur_tok().range, "Remove the trailing comma here")
+				.secondary(rest.range(p), "Rest element"),
+		);
+	} else {
+		p.error(
+			p.err_builder("rest element must be the last element")
+				.primary(rest.range(p), "Move the rest element to the end"),
```
ematipico, line 301: I would add `to the end of the pattern, right before the closing curly bracket`

crates/rslint_parser/src/syntax/binding.rs
MichaReiser: That would be neat and should be possible. But let's wait fo the autofix API to do this.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: That's copied from further below. I didn't change anything.

crates/rslint_parser/src/syntax/decl.rs
MichaReiser: hehe yeah, was annoying that it moved :)

crates/rslint_parser/src/syntax/pattern.rs
MichaReiser: Yes, we want to point the error directly at the end of the rest token. 

```
[ ...    ]
    ^ // point right after the last dot and not..

[ ...     ]
         ^ // to the next token

File: crates/rslint_parser/src/syntax/pattern.rs (hunk 16 above)
Comment (MichaReiser) [reply crates/rslint_parser/src/syntax/pattern.rs:212]: Nice catch! That makes no sense at all

File: crates/rslint_parser/src/syntax/pattern.rs (hunk 17 above)
Comment (MichaReiser) [reply crates/rslint_parser/src/syntax/pattern.rs:285]: Looks correct to me

```
error[SyntaxError]: rest element cannot have a default
  ┌─ rest_property_binding_err.js:2:12
  │
2 │ let { ...c = "default" } = a;
  │       ---- ^^^^^^^^^^^^ Remove the default value here
  │       │     
  │       Rest element
```

File: crates/rslint_parser/src/syntax/binding.rs (hunk 11 above)
Comment (MichaReiser) [reply crates/rslint_parser/src/syntax/binding.rs:265]: This is actually valid, at least outside of async functions.

Comment (ematipico) [reply crates/rslint_parser/src/syntax/binding.rs:265]: I just tried it with Babel, shift and espree. They all emit an error on incorrect use of `await`

File: crates/rslint_parser/src/syntax/binding.rs [hunk 19]
```diff
@@ -0,0 +1,285 @@
...
+// test_err binding_identifier_invalid
+// async () => { let await = 5; }
+// function *foo() {
+//    let yield = 5;
+// }
+// let eval = 5;
+// let let = 5;
+// const let = 5;
+// let a, a;
+pub(crate) fn parse_identifier_binding(p: &mut Parser) -> ConditionalParsedSyntax {
```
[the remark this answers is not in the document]
MichaReiser: Yes, that makes sense. I also need to revisit the `ConditionalParsedSyntax`. The way it currently works (where `Absent` values can be invalid too`) is super awkward to use.

crates/rslint_parser/src/syntax/pattern.rs
ematipico: Sorry, I must have had some illusion 😆

crates/rslint_parser/src/syntax/binding.rs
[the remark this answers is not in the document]
MichaReiser: I'm also still figuring this out. So far I thought it's a good idea to make the `or_invalid_to_unknown` conversion right inside of the parse function but it turned out that is dangerous. 

For example, the class or function id, a single arrow function parameter or an object rest pattern don't support `JsAnyBinding` and, therefore, returning a `JsUnknownBinding` is invalid in these cases. So it seems, all we can do in these cases is to bubble up the error until we come to a place where we can handle it. 

This is a bit annoying. An alternative would be to just add a diagnostic and keep the binding as is. This would certainly be easier but we then no longer have a strict rule when to use unknown:

* any syntax error (additional syntax) -> Content is inside of an `Unknown` node
* missing syntax -> Empty slot, handled in the AST facade

We need to get some more experience how to handle and we may come back to this change.

MichaReiser:
Update on `await`. It's a valid identifier in scripts (outside of async) functions but not in modules. 

I added one more check to disallow `let` as an identifier in `strict mode` (see coverage improvements)

crates/rslint_parser/src/syntax/binding.rs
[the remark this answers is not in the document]
ematipico: I bumped into this too while refactoring `class.rs`, where some function uses `or_invalid_to_unknown` under to hood but it is "hidden" and we don't know what's happening

ematipico (approved):

PR #1840, opened 2021-12-01
refactor: parse statements

ematipico:
## Summary

This PR moves all the functions inside `stmt.rs` to use `ParsedSyntax`.

- `stmt()` is now called `parse_statement` and it always returns `Option<CompletedMarker>`. The recovery done inside the function. Added a small chapter in the documentation
- `class_declaration` is now `parse_class_declaration` and returns `ParsedSyntax`. Although it's just a workaround, where I added a `TODO`
- `parse_function_expression` doesn't handle any error and it is intentional
- created a new `unwrap` function for `ParsedSyntax`, utility that should be used when a `ParsedSyntax` can't be `Absent`. If so, it panics
- added more tests to cover some changes around do while statements

## Test Plan

```
cargo test
cargo lint
cargo format
```

Made sure that the coverage is still the same


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -94,68 +94,77 @@ pub(super) fn is_semi(p: &Parser, offset: usize) -> bool {
...
-pub fn stmt(p: &mut Parser, recovery_set: impl Into<Option<TokenSet>>) -> Option<CompletedMarker> {
+///
+/// Error handling and recovering happens inside this function, so the
+/// caller has to pass a recovery set.
+///
+/// If not passed, [STMT_RECOVERY_SET] will be used as recovery set
+pub fn parse_statement(
+	p: &mut Parser,
+	recovery_set: impl Into<Option<TokenSet>>,
+) -> Option<CompletedMarker> {
```
MichaReiser, line 105: What's the reason for returning `Option<CompletedMarker>`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -894,7 +894,10 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
 			// let a = async function() {};
 			// let b = async function foo() {};
 			if p.nth_at(1, T![function]) {
-				function_expression(p)
+				parse_function_expression(p)
+					// it's fine to not handle the error the because the check on tokens is done beforehand
+					.or_missing(p)
```
MichaReiser, line 899: It should then be sufficient to call `ok().unwrap()` directly because it can never be `missing`. Should we add an `.expect()` function to `ParsedSyntax` if this is a common-pattern?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -946,7 +949,9 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
 			// test function_expr
 			// let a = function() {}
 			// let b = function foo() {}
-			function_expression(p)
+
+			// it's fine to not handle the error because the check on the tokens is done beforehand
+			parse_function_expression(p).or_missing(p).unwrap()
```
MichaReiser, line 954: Same here, let's introduce `ParsedSyntax::expect` or `unwrap`

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -979,7 +984,9 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
 		// (foo)
 		T!['('] => paren_or_arrow_expr(p, p.state.potential_arrow_start),
 		T!['['] => array_expr(p),
-		T!['{'] if p.state.allow_object_expr => parse_object_expression(p).or_missing(p).unwrap(),
+		T!['{'] if p.state.allow_object_expr => parse_object_expression(p)
+			.or_missing_with_error(p, js_parse_error::expected_object_expression)
```
MichaReiser, line 988: `or_missing_with_error` and `unwrap` is probably not what you want because we would add an error and then panic because of the `unwrap.` Use `.expect` / `.unwrap` if we're certain that `parse_*` returns a result or `or_missing_with_error`. We should need both.

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -5,11 +5,12 @@ use crate::syntax::js_parse_error;
 use crate::syntax::pat::parse_identifier_binding;
 use crate::syntax::stmt::{is_semi, parse_block_impl};
 use crate::syntax::typescript::{ts_type_or_type_predicate_ann, ts_type_params};
+use crate::syntax::util::is_at_async_function;
```
MichaReiser, line 8: Nit: Should this be in `function.rs`?

crates/rslint_parser/src/syntax/js_parse_error.rs
```diff
@@ -92,3 +92,15 @@ pub(crate) fn expected_property_assignment_target(p: &Parser, range: Range<usize
...
+
+pub(crate) fn expected_object_expression(p: &Parser, range: Range<usize>) -> Diagnostic {
+	expected_node("object expression", range).to_diagnostic(p)
+}
+
+pub(crate) fn expected_variable_declaration(p: &Parser, range: Range<usize>) -> Diagnostic {
+	expected_node("variable declaration", range).to_diagnostic(p)
+}
+
+pub(crate) fn expected_statement(p: &Parser, range: Range<usize>) -> Diagnostic {
```
MichaReiser, line 104: Should this be named `expected_statement_or_declaration`? What declarations do we have that aren't statements?

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -400,15 +404,20 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
 		}
 
 		if p.at(T![class]) {
-			class_declaration(&mut *p.with_state(ParserState {
+			// TODO: handle possible error later
```
MichaReiser, line 407: I think the `ok().unwrap()` is fine here. `parse_class_declaration` should always parse something after the `class` keyword.

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -400,15 +404,20 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
+			.unwrap();
 			return m.complete(p, EXPORT_DEFAULT_DECL);
 		}
 
-		if p.cur_src() == "async" && p.nth_at(1, T![function]) && !p.has_linebreak_before_n(1) {
-			function_declaration(p);
+		if is_at_async_function(p, true) {
+			parse_function_declaration(p)
+				// it's fine to not handle the error the because the check on tokens is done beforehand
+				.or_missing(p);
```
MichaReiser, line 420: Use `unwrap` or `expect` in that case. You don't want to add a `missing` marker here.

Or you can rewrite this so that you don't need the `unwrap`/`expect`. This works because `parse_function` is guaranteed to not progress if it isn't positioned at a function.

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -422,25 +431,26 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
-				&& p.nth_at(1, T![function])
-				&& !p.has_linebreak_before_n(1)))
+		is_at_async_function(p, true))
 	{
-		function_declaration(p);
+		parse_function_declaration(p)
+			// it's fine to not handle the error the because the check on tokens is done beforehand
+			.or_missing(p);
 	} else if !only_ty && p.at(T![const]) && p.nth_src(1) == "enum" {
 		ts_enum(p).err_if_not_ts(p, "enums can only be used in TypeScript files");
```
MichaReiser, line 446: `expect`/`unwrap` should be fine here. Or you can call `parse_variable_declaration_statement` and only go into this branch if it returns `Present`

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -94,68 +94,77 @@ pub(super) fn is_semi(p: &Parser, offset: usize) -> bool {
 }
 
 /// A generic statement such as a block, if, while, with, etc
-#[allow(deprecated)]
-pub fn stmt(p: &mut Parser, recovery_set: impl Into<Option<TokenSet>>) -> Option<CompletedMarker> {
+///
+/// Error handling and recovering happens inside this function, so the
```
MichaReiser, line 98: Does it have to? Or is this just for now?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -722,21 +731,24 @@ pub fn parse_while_statement(p: &mut Parser) -> ParsedSyntax {
...
 	// test_err var_decl_err
 	// var a =;
 	// const a = 5 let b = 5;
 	let m = p.start();
 	let start = p.cur_tok().range.start;
 
-	variable_declaration(p, false);
+	parse_variable_declaration(p, false).or_missing(p);
 
 	semi(p, start..p.cur_tok().range.start);
```
MichaReiser, line 743: We need to check if we're indeed at a variable declaration statement and otherwise return absent. You can do this like this:

```suggestion
if let Present(declaration) = parse_variable_declaration(p, false) {
  let m = declaration.precede(p);
  semi(p, start..p.cur_tok().range.start);
  Present(m.complete(p, JS_VARIABLE_DECLARATION_STATEMENT))
} else {
  Absent
}
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -722,21 +731,24 @@ pub fn parse_while_statement(p: &mut Parser) -> ParsedSyntax {
...
-	m.complete(p, JS_VARIABLE_DECLARATION_STATEMENT)
+	Present(m.complete(p, JS_VARIABLE_DECLARATION_STATEMENT))
 }
 
-fn variable_declaration(p: &mut Parser, no_semi: bool) -> CompletedMarker {
+/// Parses a list of JS_VARIABLE_DECLARATION
+///
+/// Error handling is delegated to the subsequent functions, so it's safe to call `or_missing`
+fn parse_variable_declaration(p: &mut Parser, no_semi: bool) -> ParsedSyntax {
 	let m = p.start();
```
MichaReiser, line 752: What does that mean and why? Is this a different pattern?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -722,21 +731,24 @@ pub fn parse_while_statement(p: &mut Parser) -> ParsedSyntax {
...
 	let mut is_const = None;
```
MichaReiser, line 753: The function should return `Absent` if there's no `let`/`var`/`const` identifier.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -774,7 +786,7 @@ fn variable_declaration(p: &mut Parser, no_semi: bool) -> CompletedMarker {
 
 	declared_list.complete(p, LIST);
 	p.state.name_map.clear();
-	m.complete(p, JS_VARIABLE_DECLARATION)
+	Present(m.complete(p, JS_VARIABLE_DECLARATION))
 }
```
MichaReiser, line 791: Will you change `variable_declarator` in its own PR?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -879,18 +891,17 @@ pub fn parse_do_statement(p: &mut Parser) -> ParsedSyntax {
...
-	Present(m.complete(p, JS_DO_WHILE_STATEMENT))
+	let mut guard = p.with_state(ParserState {
+		continue_allowed: true,
+		break_allowed: true,
+		..p.state.clone()
+	});
+	parse_statement(&mut *guard, None);
+	guard.expect_required(T![while]);
+	parenthesized_expression(&mut *guard);
+	let end_range = guard.cur_tok().range.end;
```
MichaReiser, line 902: `break` and `continue` aren't allowed inside the `while` token nor the `parenthesized expression. That's why I think the existing code was correct.

Maybe add some test cases for using `break`/`continue` inside a while loop?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -168,6 +175,8 @@ pub(crate) fn ts_declare(p: &mut Parser) -> Option<CompletedMarker> {
 			p.bump_remap(T![declare]);
 			// unwrap the marker so its children go to `m`
 			variable_declaration_statement(p)
+				.or_missing_with_error(p, js_parse_error::expected_variable_declaration)
```
MichaReiser, line 178: Either use unwrap or `or_missing_with_error`. It makes no sense to use them together (adding an error to then panic)

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -176,6 +185,8 @@ pub(crate) fn ts_declare(p: &mut Parser) -> Option<CompletedMarker> {
 			let m = p.start();
 			p.bump_remap(T![declare]);
 			variable_declaration_statement(p)
+				.or_missing_with_error(p, js_parse_error::expected_variable_declaration)
```
MichaReiser, line 188: Same

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -77,3 +77,14 @@ pub fn check_for_stmt_declaration(p: &mut Parser, marker: &CompletedMarker) {
 		p.error(err);
 	}
 }
+
+#[inline]
+/// Checks if the parser is inside a "async function"
+pub(super) fn is_at_async_function(p: &mut Parser, check_line_break: bool) -> bool {
```
MichaReiser, line 83: Nit: It's hard to guess what the `bool` param is for without looking at the function. 

Should we introduce an enum to make it more explict?

crates/rslint_parser/test_data/inline/err/block_stmt_in_class.rast
```diff
@@ -72,5 +72,12 @@ error[SyntaxError]: Expected a statement or declaration, but found none
 1 │ class S{{}}
   │           ^ Expected a statement or declaration here
 
+--
+error[SyntaxError]: expected a statement, or a declaration but instead found ''
```
MichaReiser, line 76: Do you know where this second error is now coming from?

crates/rslint_parser/test_data/inline/err/function_broken.rast
```diff
@@ -118,32 +103,11 @@ error[SyntaxError]: expected a function body but instead found ')'
...
-1 │ function foo())})}{{{  {}
-  │                 ^ Expected a statement or declaration here
-
---
-error[SyntaxError]: Expected a statement or declaration, but found none
-  ┌─ function_broken.js:1:18
-  │
-1 │ function foo())})}{{{  {}
-  │                  ^ Expected a statement or declaration here
+  │               ^^^^ Expected a statement, or a declaration here
```
MichaReiser, line 110: Much better!

crates/rslint_parser/test_data/inline/err/object_expr_error_prop_name.rast
```diff
@@ -217,31 +145,12 @@ error[SyntaxError]: Expected a statement or declaration, but found none
   │            ^ Expected a statement or declaration here
 
 --
-error[SyntaxError]: Expected a semicolon or an implicit semicolon after a statement, but found none
```
MichaReiser, line 220: Haha lol, nice that our parser recovered from this :D

MichaReiser:
Nice work!

There are some smaller issues with using `or_missing*` with `unwrap` and I have some conceptual questions about error recovery. Otherwise looks great.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: Ideally, this function will return `ParsedSyntax` in the future and all the `except`/`unwrap` calls will be gone.

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -400,15 +404,20 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
 		}
 
-		if p.cur_src() == "async" && p.nth_at(1, T![function]) && !p.has_linebreak_before_n(1) {
-			function_declaration(p);
+		if is_at_async_function(p, true) {
+			parse_function_declaration(p)
+				// it's fine to not handle the error the because the check on tokens is done beforehand
+				.or_missing(p);
 			return m.complete(p, EXPORT_DEFAULT_DECL);
 		}
```
MichaReiser, line 422:
```suggestion
		
		if let Present(function) = parse_function_declaration(p) {
			return m.complete(p, EXPORT_DEFAULT_DECL);
		}
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -94,68 +94,77 @@ pub(super) fn is_semi(p: &Parser, offset: usize) -> bool {
...
+				js_parse_error::expected_statement,
+			);
 
-	Some(res)
+			if recovered.is_ok() {
+				recovered.ok()
+			} else {
+				None
+			}
+		}
```
MichaReiser, line 165: You can simply use `recovered.ok()`

```suggestion
			recovered.ok(
		}
```

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -422,25 +431,26 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
-		// async function ...
-		(p.cur_src() == "async"
-				&& p.nth_at(1, T![function])
-				&& !p.has_linebreak_before_n(1)))
+		is_at_async_function(p, true))
 	{
-		function_declaration(p);
+		parse_function_declaration(p)
+			// it's fine to not handle the error the because the check on tokens is done beforehand
+			.or_missing(p);
```
MichaReiser, line 444: Use `unwrap`/`expect`

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -422,25 +431,26 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
 	} else if !only_ty && p.at(T![const]) && p.nth_src(1) == "enum" {
 		ts_enum(p).err_if_not_ts(p, "enums can only be used in TypeScript files");
 	} else if !only_ty
 		&& (p.at(T![var])
 			|| p.at(T![const])
 			|| (p.cur_src() == "let" && FOLLOWS_LET.contains(p.nth(1))))
 	{
-		variable_declaration_statement(p);
+		variable_declaration_statement(p)
+			.or_missing_with_error(p, js_parse_error::expected_variable_declaration);
```
MichaReiser, line 453: `expect`/`unwrap` should be fine here

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -901,7 +912,9 @@ fn for_head(p: &mut Parser) -> SyntaxKind {
 			include_in: false,
 			..p.state.clone()
 		});
-		let decl = variable_declaration(&mut *guard, true);
+		let decl = parse_variable_declaration(&mut *guard, true)
+			.or_missing(&mut *guard)
+			.unwrap();
```
MichaReiser, line 917: You don't want to use `or_missing` with `unwrap`. What you want here is `expect` or re-use the pattern I proposed before

MichaReiser (changes requested):

crates/rslint_parser/src/syntax/stmt.rs
ematipico: It's a workaround for what we have in `stmt.rs:224`.

Once `expr_or_assignment` will return `ParsedSyntax`, we can migrate to use `Present` instead. I will add a `TODO`

crates/rslint_parser/src/syntax/stmt.rs
ematipico: For now it has to, yes. `expr_with_semi_recovery` is the culprit. I'd prefer to leave that function as it is now and tackle it later. Too many changes for now.

crates/rslint_parser/src/syntax/function.rs
ematipico: I assume that `function.rs` will contains functions to **parse**, while util can contain various functions

crates/rslint_parser/src/syntax/js_parse_error.rs
ematipico: Declaration is a term that belonged to the old syntax. I will change it to just "statement"

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Yes

crates/rslint_parser/test_data/inline/err/block_stmt_in_class.rast
ematipico: It's coming from the recovery that we have inside `parse_statement`

crates/rslint_parser/src/syntax/util.rs
ematipico: Not a fan of making an enum just to replace a boolean, but I can do it

crates/rslint_parser/test_data/inline/err/object_expr_error_prop_name.rast
ematipico: It's a monster parser, ultra resilient!

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -90,6 +90,20 @@ impl ParsedSyntax {
 		}
 	}
 
+	/// It returns a [CompletedMarker] from the current syntax
+	///
+	/// # Panics
+	///
+	/// The current syntax is [ParsedSyntax::Absent]
```
MichaReiser, line 97:
```suggestion
	/// Panics if the syntax is [ParsedSyntax::Absent].
```

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -90,6 +90,20 @@ impl ParsedSyntax {
...
+	pub fn unwrap(self) -> CompletedMarker {
+		match self {
+			Absent => {
+				panic!("Can't call unwrap on a missing marker");
```
MichaReiser, line 101:
```suggestion
				panic!("Called `unwrap` on an `Absent` syntax");
```

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: Remove the `missing`

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: Thanks for adding the todos

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -230,38 +237,39 @@ fn expr_statement(p: &mut Parser) -> Option<CompletedMarker> {
 				"Tried to get the ident of a name node, but there was no ident. This is erroneous",
 			);
 
-		let text_range = TextRange::new((range.start as u32).into(), (range.end as u32).into());
-		let text = p.source(text_range);
-		if let Some(range) = p.state.labels.get(text) {
-			let err = p
-				.err_builder("Duplicate statement labels are not allowed")
-				.secondary(
```
MichaReiser, line 238: Oh god, I'll hate merging xD

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -761,6 +771,8 @@ fn variable_declaration(p: &mut Parser, no_semi: bool) -> CompletedMarker {
 				.primary(p.cur_tok().range, "");
 
 			p.error(err);
+			m.abandon(p);
```
MichaReiser, line 774: We shouldn't add an error if it is Absent. Adding errors is left to the caller.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: Still relevant. We shouldn't use `guard` to parse `while` and the expression

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -150,11 +150,15 @@ pub(crate) fn ts_declare(p: &mut Parser) -> Option<CompletedMarker> {
...
+		T![function] => parse_function_declaration(p).unwrap(),
 		T![class] => {
 			let m = p.start();
 			p.bump_remap(T![declare]);
-			class_declaration(p).undo_completion(p).abandon(p);
+			parse_class_declaration(p)
+				.ok()
+				.unwrap()
+				.undo_completion(p)
+				.abandon(p);
```
MichaReiser, line 161: Nit
```suggestion
			parse_class_declaration(p)
				.unwrap()
				.undo_completion(p)
				.abandon(p);
```

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -218,7 +224,11 @@ pub(crate) fn ts_decl(p: &mut Parser) -> Option<CompletedMarker> {
 			p.error(err);
 			return None;
 		}
-		class_declaration(p).undo_completion(p).abandon(p);
+		parse_class_declaration(p)
+			.ok()
+			.unwrap()
```
MichaReiser, line 229: Nit
```suggestion
		parse_class_declaration(p)
			.unwrap()
```

MichaReiser (approved):
Overall looks good. There are some smaller issues that need fixing before merging.

crates/rslint_parser/src/syntax/function.rs
MichaReiser: The main reason why I would move it is that `is_at_async_function` may need to be updated when `parse_function` changes. That's why I always placed `is_at_*` methods right next to the corresponding parse method.

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -400,15 +404,20 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
 		if p.at(T![class]) {
-			class_declaration(&mut *p.with_state(ParserState {
+			// TODO: handle possible error later
+			parse_class_declaration(&mut *p.with_state(ParserState {
 				in_default: true,
 				..p.state.clone()
-			}));
+			}))
+			.ok()
+			.unwrap();
```
MichaReiser: Nit
```suggestion
			.unwrap();
```

crates/rslint_parser/src/syntax/program.rs
MichaReiser: Use `unwrap` here. Missing should never be possible (and union rules should never end up adding `missing` slots to the parent)

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Good catch, I forgot to remove the error

crates/rslint_parser/src/syntax/stmt.rs
ematipico: I can't recall why I changed that. There was a compiler error. I restored the changes to what they were.

Issue #1837, opened 2021-11-30
📎 Parser Optimization


xunilrj:
### Description

Doing a more low level analysis of our parser we discovered that when parsing real libs, like https://github.com/mrdoob/three.js, we were spending 70% of the time just allocating/cloning and moving memory.

To discover that we run

```
> valgrind --tool=callgrind --dump-instr=yes --simulate-cache=yes --collect-jumps=yes target/release/xtask coverage-libs --filter=three
```
And used kcachegrind

![image](https://user-images.githubusercontent.com/83425/144065374-b86ef72f-45f6-407d-920a-e2afb77a96e0.png)

The main issue is that to "look ahead" or to do tentative parsing we are cloning the parser. That was cloning the whole Tokenized model and all the parsing events.

A more long term solution would be to come up with a more optimized model for forking the parser; but for a short term solution, and to show that this is actually our problem, I did two optimizations.

For some reason, on Windows we are much slower than on WSL (allocator?)

Before on WSL
```
> cargo run -p xtask --release -- coverage-libs --filter=three
    Finished release [optimized] target(s) in 1.73s
     Running `target/release/xtask coverage-libs --filter=three`
[three.min.js] - using [target/three.min.js]
Tokenizer
        Time
                took 28.460508ms
Parser
        Time
                took 1.365925465s
TreeSink
        Time
                took 37.974648ms
Total Time: 1.436760875s
end

> cargo run -p xtask --release -- coverage-libs | grep "Total Time:"
    Finished release [optimized] target(s) in 1.52s
     Running `target/release/xtask coverage-libs`
Total Time: 76.490934ms
Total Time: 141.972642ms
Total Time: 1.076629ms
Total Time: 69.732935ms
Total Time: 484.809849ms
Total Time: 7.937441ms
Total Time: 1.033502967s
Total Time: 667.98844ms
Total Time: 968.266103ms
```

Before on Windows
```rust
> cargo run -p xtask --release -- coverage-libs --filter=three    
    Finished release [optimized] target(s) in 0.11s
     Running `target\release\xtask.exe coverage-libs --filter=three`
[three.min.js] - using [target\three.min.js]
Tokenizer
        Time
                took 11.7938ms
        Time
                took 2.9642032s
TreeSink
        Time
                took 44.223ms
Total Time: 3.0299302s
end
> cargo run -p xtask --release -- coverage-libs | sls "Total Time:"
    Finished release [optimized] target(s) in 0.11s
[...]

PR #1841, opened 2021-12-01
performance: Remove cloning of parser

MichaReiser:
## Summary

@xunilrj analysis showed that cloning the whole parser considerably reduces the parsing performance. 

The two places where we cloned the parser now have been in

`try_parse_ts` and inside `paren_or_arrow_expr`. 

* `try_parse_ts`: `rewind` used to panic but it seems that the bindings refactoring removed the code that caused the panic. It's now safe to use `rewind`.
* `paren_or_arrow_expr`: No idea what that code was doing. Removing it doesn't regress coverage, nor any test. I also compared the errors when parsing the libraries and there's no difference as well. Let's re-add it when we now what it is used for (with proper tests and such)

Some parse time goodies

* `vue.global.prod`: From 291ms to 12.4ms
* `react.production`: From 754us to 355us
* `dojo`: From 12ms to 2ms
* `d3.min` from 3s to 20ms (wtf?) 
* `pixi.min` from 1.9s to `21ms`
* `three.min` from 2.7s to 28ms

Part of #1837

## Test Plan

Compared coverage, ran unit tests

### DIFF
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -759,36 +759,7 @@ pub fn paren_or_arrow_expr(p: &mut Parser, can_be_arrow: bool) -> CompletedMarke
 	}
 
 	drop(temp);
-	// if we are in a ternary expression, then we need to try and see if parsing as an arrow worked
-	// if it did then we just return it, otherwise it should be interpreted as a grouping expr
-	if p.state.in_cond_expr && p.at(T![:]) && params_marker.is_none() {
-		let func = |p: &mut Parser| {
-			let p = &mut *p.with_state(ParserState {
-				no_recovery: true,
-				..p.state.clone()
-			});
-			p.rewind(checkpoint);
-			parse_parameter_list(p).or_missing_with_error(p, js_parse_error::expected_parameters);
-			if p.at(T![:]) {
-				if let Some(mut ret) = ts_type_or_type_predicate_ann(p, T![:]) {
-					ret.err_if_not_ts(
-						p,
-						"arrow functions can only have return types in TypeScript files",
-					);
-				}
-			}
-			p.expect_no_recover(T![=>])?;
-			parse_arrow_body(p).or_missing_with_error(p, js_parse_error::expected_arrow_body)
-		};
-		// we can't just rewind the parser, since the function rewinds, and cloning and replacing the
-		// events does not work apparently, therefore we need to clone the entire parser
-		let cloned = p.clone();
-		if func(p).is_some() {
-			return m.complete(p, JS_ARROW_FUNCTION_EXPRESSION);
-		} else {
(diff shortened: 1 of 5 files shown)


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -757,36 +757,7 @@ pub fn paren_or_arrow_expr(p: &mut Parser, can_be_arrow: bool) -> CompletedMarke
 	}
 
 	drop(temp);
-	// if we are in a ternary expression, then we need to try and see if parsing as an arrow worked
-	// if it did then we just return it, otherwise it should be interpreted as a grouping expr
-	if p.state.in_cond_expr && p.at(T![:]) && params_marker.is_none() {
```
MichaReiser, line 762: This may be for typescript support... but I don't know for sure.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -673,16 +673,19 @@ pub fn try_parse_ts(
 	p: &mut Parser,
 	func: impl FnOnce(&mut Parser) -> Option<CompletedMarker>,
 ) -> Option<CompletedMarker> {
-	// FIXME: simply rewinding doesnt work for `new A < T;` and it makes the parser enter unreachable
-	// code, and i really cant be bothered to debug incorrect event code because i don't have enough sanity
-	// left in me
```
ematipico, line 678: I really appreciate the honesty of the comment!

ematipico (approved):

IWANABETHATGUY:
how should is use the new efficient parser, i use rslint_lint parser in my codebase like this 
```toml
rslint_parser = { git = "https://github.com/rome/tools", branch = "remove-cloning"}
```
but i can't get clearly speed up from this new perf branch
anything wrong ?

MichaReiser:
> how should is use the new efficient parser, i use rslint_lint parser in my codebase like this
> 
> ```toml
> rslint_parser = { git = "https://github.com/rome/tools", branch = "remove-cloning"}
> ```
> 


Looks about correct. Did cargo pull the new dependency? 

> but i can't get clearly speed up from this new perf branch anything wrong ?

Impossible to say without knowing more about how you use the parser, what you're testing and measuring. I don't expect this to change much for small files because it only improves the parse time but not the other phases as well.

MichaReiser:
To expand on this a bit. Our coverage tests (parsing about 17k very small JS files) doesn't benefit from these improvements at all because the cloning isn't very expensive in these cases (a few events, but not a 100 thousands or millions of events).

IWANABETHATGUY:
> To expand on this a bit. Our coverage tests (parsing about 17k very small JS files) doesn't benefit from these improvements at all because the cloning isn't very expensive in these cases (a few events, but not a 100 thousands or millions of events).

thanks a lot.

PR #1843, opened 2021-12-01
fix(rslint_parser): Fix parsing of multiple tokens

### LINKED ISSUE #1842: 🐛 Incorrect parsing of JsBinaryExpression `>>` and `>>>`
### Environment information

```bash
rustc 1.56.1
commit b716a98
```

### What happened?

parsing 
```
a >> b
```
yields
```
0: JS_BINARY_EXPRESSION@0..7
		0: JS_REFERENCE_IDENTIFIER_EXPRESSION@0..2
			0: IDENT@0..2 "a" [] [Whitespace(" ")]
		1: SHR@2..4 ">>" [] []
		2: JS_REFERENCE_IDENTIFIER_EXPRESSION@4..7
			0: IDENT@4..7 ">> " [] []
```

### Expected result

The second IDENT should be `b`, not `>>`

Same applies to `>>>`.
All other binary operators are fine.
The bug is some where inside [`binary_or_logical_expression_recursive`](https://github.com/rome/tools/blob/main/crates/rslint_parser/src/syntax/expr.rs#L251)? or something about `consume_multiple_tokens`

_No response_

### Code of Conduct

MichaReiser:
## Summary

The parser allows grouping multiple lexer-tokens into a single "parsed" token using `bump_multiple`. However, this path wasn't yet updated to account for leading/trailing trivia.

This PR ensures that trivia is also handled for the multiple token case by unifying the `do_token` implementation.

Fixes #1842

## Test Plan

Added a new test case for bin op.

### DIFF
--- a/crates/rslint_parser/test_data/inline/ok/binary_expressions.rast
+++ b/crates/rslint_parser/test_data/inline/ok/binary_expressions.rast
@@ -67,16 +67,34 @@ JsRoot {
         },
         JsExpressionStatement {
             expression: JsBinaryExpression {
+                left: JsReferenceIdentifierExpression {
+                    name_token: IDENT@87..90 "a" [Whitespace("\n")] [Whitespace(" ")],
+                },
+                operator: SHR@90..93 ">>" [] [Whitespace(" ")],
+            },
+            semicolon_token: missing (optional),
+        },
+        JsExpressionStatement {
+            expression: JsBinaryExpression {
+                left: JsReferenceIdentifierExpression {
+                    name_token: IDENT@94..97 "a" [Whitespace("\n")] [Whitespace(" ")],
+                },
+                operator: USHR@97..101 ">>>" [] [Whitespace(" ")],
+            },
+            semicolon_token: missing (optional),
+        },
+        JsExpressionStatement {
+            expression: JsBinaryExpression {
                 left: JsBinaryExpression {
                     left: JsBinaryExpression {
                         left: JsNumberLiteralExpression {
-                            value_token: JS_NUMBER_LITERAL@87..90 "1" [Whitespace("\n")] [Whitespace(" ")],
+                            value_token: JS_NUMBER_LITERAL@102..105 "1" [Whitespace("\n")] [Whitespace(" ")],
                         },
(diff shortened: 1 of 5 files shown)


crates/rslint_parser/test_data/inline/ok/binary_expressions.rast
```diff
@@ -65,40 +65,58 @@ JsRoot {
             },
             semicolon_token: missing (optional),
         },
+        JsExpressionStatement {
+            expression: JsBinaryExpression {
+                left: JsReferenceIdentifierExpression {
+                    name_token: IDENT@87..90 "a" [Whitespace("\n")] [Whitespace(" ")],
+                },
+                operator: SHR@90..93 ">>" [] [Whitespace(" ")],
```
MichaReiser, line 73: The "right" is missing here because of a bug in the code gen. It doesn't include manual fields in the `debug` output yet. I'm aware of this limitation but don't think it's worth fixing because manual fields should be history once #1724  lands.

crates/rslint_parser/test_data/inline/ok/binary_expressions.rast
```diff
@@ -178,44 +196,60 @@ JsRoot {
...
-        1: PLUS@98..100 "+" [] [Whitespace(" ")]
-        2: JS_NUMBER_LITERAL_EXPRESSION@100..101
-          0: JS_NUMBER_LITERAL@100..101 "1" [] []
+    7: JS_EXPRESSION_STATEMENT@87..94
+      0: JS_BINARY_EXPRESSION@87..94
+        0: JS_REFERENCE_IDENTIFIER_EXPRESSION@87..90
+          0: IDENT@87..90 "a" [Whitespace("\n")] [Whitespace(" ")]
+        1: SHR@90..93 ">>" [] [Whitespace(" ")]
+        2: JS_REFERENCE_IDENTIFIER_EXPRESSION@93..94
+          0: IDENT@93..94 "b" [] []
```
MichaReiser, line 205: Here you go!

Boshen:
Tests also passed on my end.

ematipico (approved):

PR #1845, opened 2021-12-02
 feat(formatter): implement todos in JsAnyExpression 

Boshen:
## Summary
This PR implements most of the todos in `rome_formatter/src/ts/expressions/expression.rs`
And fixes:
* incorrect `import.meta` syntax
* incorrect `/=` `**=` for JsAssignmentExpression

This PR is rebased onto https://github.com/rome/tools/pull/1843 for passing tests for `a >> b`
## Test Plan
See updated formatter tests


Boshen:
btw I can't figure out what syntax `TsAssertion` and `TsConstAssertion` is 😿 , rslint gives no hint either
```
TsAssertion = expr:JsAnyExpression ident: Ident '<' ty:TsType '>'
TsConstAssertion = expr:JsAnyExpression ident: Ident '<' 'const' '>'
```

crates/rome_formatter/src/ts/expressions/any_reference_member.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl ToFormatElement for JsReferenceIdentifierMember {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		formatter.format_token(&self.name_token()?)
+	}
+}
+
+impl ToFormatElement for JsReferencePrivateMember {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		formatter.format_token(&self.name_token()?)
```
MichaReiser, line 27: Private names have the form `#name` where the `#` is its own token inside private member. So you need to visit (format) the `#` token too to make sure it doesn't get lost. I'm a bit surprised that the test is passing

crates/rslint_parser/test_data/inline/ok/import_meta.rast
```diff
@@ -6,6 +6,7 @@ JsRoot {
             expression: ImportMeta {
                 import_token: IMPORT_KW@0..6 "import" [] [],
                 dot_token: DOT@6..7 "." [] [],
+                meta_token: missing (required),
```
MichaReiser, line 9: Seems like the parser isn't parsing the meta token. I'll be working on `import` soon. Should we defer adding the meta token until then?

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+			space_token(),
+			formatter.format_node(self.object()?)?,
+			formatter.format_node(self.arguments()?)?,
+		])
+	}
+}
+
+impl ToFormatElement for JsUnaryExpression {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		// TODO handle `+` `-` `~` `!`
```
MichaReiser, line 163: Why do they need special handling?

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+	}
+}
+
+impl ToFormatElement for JsBinaryExpression {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		let right = if let Some(token) = self.right() {
+			formatter.format_node(token)?
+		} else {
+			empty_element()
+		};
```
MichaReiser, line 178: Oh, the return-type of `right` on `BinaryExpression` is incorrect. `right` should return a `SyntaxResult<JsAnyExpression>` Would you mind changing the implementation?

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+			formatter.format_node(self.test()?)?,
+			space_token(),
+			formatter.format_token(&self.question_mark_token()?)?,
+			space_token(),
+			formatter.format_node(self.consequent()?)?,
+			space_token(),
+			formatter.format_token(&self.colon_token()?)?,
+			space_token(),
+			formatter.format_node(self.alternate()?)?,
+		])
```
MichaReiser, line 201: Can you add a test case that exceeds the 80 character line width?  We then want to wrap the condition over multiple lines. Similar to rome classic

https://github.com/rome/tools/blob/c76ccc0345fe00ea4b3f684424065d05c3ab3abe/internal/formatter/builders/js/expressions/JSConditionalExpression.ts#L48-L68

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+			empty_element()
+		};
+		Ok(format_elements![
+			formatter.format_node(self.left()?)?,
+			space_token(),
+			formatter.format_token(&self.operator()?)?,
+			space_token(),
+			right
+		])
+	}
```
MichaReiser, line 186: This is fine for now but we probably want to print binary expressions across multiple lines if they get very long, similar to what rome classic

https://github.com/rome/tools/blob/8edbed0c3505c8ed3724015df83817c2a92acce8/internal/formatter/builders/js/expressions/JSBinaryExpression.ts#L19-L81

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+impl ToFormatElement for JsAssignmentExpression {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		Ok(format_elements![
+			formatter.format_node(self.left()?)?,
+			space_token(),
+			formatter.format_token(&self.operator_token()?)?,
+			space_token(),
+			formatter.format_node(self.right()?)?,
+		])
+	}
```
MichaReiser, line 214: We may want to make sure that assignments can break if they're very long (or concatenated)

https://github.com/rome/tools/blob/8dcd9dde2b0c3d850f24279c472e0e2fd90bb0af/internal/formatter/builders/js/utils.ts#L271-L285

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+
+impl ToFormatElement for ImportMeta {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		Ok(format_elements![
+			formatter.format_token(&self.import_token()?)?,
+			formatter.format_token(&self.dot_token()?)?,
+			formatter.format_token(&self.meta_token()?)?,
+		])
+	}
+}
```
MichaReiser, line 293: I just checked the grammar and this seems to be a dead node. The `ImportMeta` node is never used on the right-hand side of any other rule. I think we can delete this.

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,362 @@ impl ToFormatElement for JsAnyExpression {
...
+			empty_element()
+		};
+		Ok(format_elements![
+			formatter.format_node(self.left()?)?,
+			space_token(),
+			formatter.format_token(&self.operator()?)?,
+			space_token(),
+			right,
+		])
+	}
```
MichaReiser, line 377: Same as for binary expression, we probably want to indent the right hand side if the expressions are very long

crates/rome_formatter/tests/specs/js/conditional.js
```diff
@@ -0,0 +1,4 @@
+a  ?  b  :  c
+d
+	? (e + f)
+	: (g + h)
```
MichaReiser, line 4: I think it should be fine to print these on a single line because the whole expression isn't exceeding the line width. 

Can we add another example for a conditional that exceeds 80 character line width? Maybe even with nesting?

ematipico:
> btw I can't figure out what syntax `TsAssertion` and `TsConstAssertion` is 😿 , rslint gives no hint either
> 
> ```
> TsAssertion = expr:JsAnyExpression ident: Ident '<' ty:TsType '>'
> TsConstAssertion = expr:JsAnyExpression ident: Ident '<' 'const' '>'
> ```

You ignore anything related to TypeScript at the moment. We need to review it and probably it will change drastically.

crates/rome_formatter/tests/specs/js/expression.js
```diff
@@ -0,0 +1,8 @@
+x = y!
+x = ++y
+x = --y
+x = y++
+x = y--
+x ?? y
+x || y
+x && y
```
MichaReiser, line 8: Can we add some examples that exceed the line width? Maybe even with nesting?

crates/rome_formatter/tests/specs/js/static_member.expected.js
```diff
@@ -0,0 +1,3 @@
+a.b;
+c?.d;
+e.#f
```
MichaReiser, line 3: Shouldn't there be a `;` too?
```suggestion
e.#f;
```

MichaReiser (commented):
This is amazing, thank you!

I believe private names aren't printed correctly yet. 

There's also the question how far you wanna go. There are some expressions that should break if exceeding the line width but I'm fine if we leave this as is for now and tackle it in later PRs.

crates/rome_formatter/tests/specs/js/static_member.expected.js
Boshen: Hah! I thought this was a bug somewhere, and then I printed the diagnostics and saw ”private identifiers are unsupported“. I'll remove this test for now.

crates/rome_formatter/src/ts/expressions/expression.rs
Boshen: We need to handle spacing for `+` `-` `~` `!` vs `void` `delete` `typeof`, fixed in 1936dd0

crates/rome_formatter/src/ts/expressions/expression.rs
Boshen: fixed in 0b01cd5

crates/rslint_parser/test_data/inline/ok/import_meta.rast
Boshen: We can leave it here and fix later I think.

crates/rome_formatter/src/ts/expressions/any_reference_member.rs
Boshen: fixed in e311e73, although the parser hasn't implemented this yet :-( (which is why the test looked strange ... removed the test for now)

Boshen:
@MichaReiser All actionable items are done, we can work on the finer details later, let's put more focus on the parser for now 😀

crates/rslint_parser/test_data/inline/ok/import_meta.rast
MichaReiser: We shouldn't have any parsing errors in any file inside of the `ok` folder and this is a parsing error. The easiest will be to just remove the whole grammar for now since it isn't supported anyway.

crates/rome_formatter/tests/specs/js/static_member.expected.js
```diff
@@ -0,0 +1,2 @@
+a.b;
+c?.d;
```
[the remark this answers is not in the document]
Boshen: "private identifiers are unsupported" in the parser 😢 
https://github.com/rome/tools/blob/main/crates/rslint_parser/src/syntax/expr.rs#L531

I should probably look into outputting all the error diagnostics from the parser when running formatter tests, the formatter currently can give the wrong result which looks correct...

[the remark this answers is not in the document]
ematipico: Makes sense, you have to create a snippet with a class:

```js
class B {
	foo() { #a.b }
}
```

etc.

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+				formatter.format_token(&self.l_curly_token()?)?,
+				block_indent(join_elements(
+					hard_line_break(),
+					formatter.format_nodes(self.members())?
+				)),
+				formatter.format_token(&self.r_curly_token()?)?
+			])
+		])
+	}
+}
```
ematipico, line 275: Would you mind moving this inside a `class.rs` file? Ideally we should move there all formatters related to `class`

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+
+impl ToFormatElement for JsStaticMemberExpression {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		Ok(format_elements![
+			formatter.format_node(self.object()?)?,
+			formatter.format_token(&self.operator()?)?,
+			formatter.format_node(self.member()?)?,
+		])
+	}
+}
```
ematipico, line 148: To move to a `class.rs` file

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+impl ToFormatElement for Template {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		// TODO implement template string, see parser in rslint_parser/src/ast/expr_ext.rs
+		Ok(format_elements![
+			formatter.format_token(&self.backtick_token()?)?,
+			token("TODO"),
+			formatter.format_token(&self.backtick_token()?)?
+		])
+	}
+}
```
ematipico, line 104: Let's move it into a `template.rs` file. This will change and probably might have some custom logic

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+		if let Some(token) = self.id() {
+			tokens.push(formatter.format_node(token)?);
+		}
+		tokens.push(formatter.format_node(self.parameters()?)?);
+		tokens.push(space_token());
+		tokens.push(formatter.format_node(self.body()?)?);
+
+		Ok(concat_elements(tokens))
+	}
+}
```
ematipico, line 244: Let's move it inside a `function.rs` file. There we will have also `JsFunctionDeclaration`

crates/rome_formatter/src/ts/expressions/expression.rs
ematipico: It's worth moving this inside a `import/meta.rs` file

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+impl ToFormatElement for JsImportCallExpression {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		Ok(format_elements![
+			formatter.format_token(&self.import_token()?)?,
+			formatter.format_token(&self.l_paren_token()?)?,
+			formatter.format_node(self.argument()?)?,
+			formatter.format_token(&self.r_paren_token()?)?,
+		])
+	}
+}
```
ematipico, line 306: Inside a `import/mod.rs` file

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+
+impl ToFormatElement for TsNonNull {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		Ok(format_elements![
+			formatter.format_node(self.expr()?)?,
+			formatter.format_token(&self.excl_token()?)?,
+		])
+	}
+}
```
ematipico, line 345: Can we just have a `todo!()` here?

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +18,363 @@ impl ToFormatElement for JsAnyExpression {
...
+
+impl ToFormatElement for JsPostUpdateExpression {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		Ok(format_elements![
+			formatter.format_node(self.operand()?)?,
+			formatter.format_token(&self.operator()?)?,
+		])
+	}
+}
```
ematipico, line 363: We could move these inside a `update_expressions.rs` file. What do you think?

crates/rome_formatter/tests/specs/js/class.expected.js
```diff
@@ -14,3 +14,5 @@ class Foo extends Boar {
 	new_prop = 5;
 	double_semicolon = [5, 3, 4];
 }
+x = class {};
+x = class foo extends Boar {};
```
ematipico, line 18: Let's add a test with a veeeery long names, both the class name and the super class.

If you're not sure what to do, we use Prettier as a reference

crates/rome_formatter/tests/specs/js/function.js
```diff
@@ -0,0 +1,17 @@
...
+}
+x = async function* foo(a) {
+}
+function Foo() {
+  if (!new.target) { }
+}
+function* Foo() {
+  yield
+  yield x
+  yield* x
```
ematipico, line 16: Let's add a couple of tests where `argument` (`x`) is really long

crates/rome_formatter/tests/specs/js/import.js
```diff
@@ -0,0 +1,2 @@
+import(x)
```
ematipico, line 1: Let's add a test with an argument that has a long name

crates/rome_formatter/tests/specs/js/new.expected.js
```diff
@@ -0,0 +1,2 @@
+new a();
+new b(x);
```
ematipico, line 2: Let's add a test with an argument that has a long name. Same for function name

crates/rome_formatter/tests/specs/js/parenthesized.js
```diff
@@ -0,0 +1 @@
+(1)
```
ematipico, line 1: Let's add a test with an argument that has a long name

crates/rome_formatter/tests/specs/js/unary.js
```diff
@@ -0,0 +1,7 @@
+delete  a
+void    b
+typeof  c
+x = +1
+x = -1
+x = ~1
+x = !1
```
ematipico, line 7: Let's test with long names too

ematipico (changes requested):
Thank you for tackling this! :)

I would expand the test base with long names and see how the formatter behaves.

We use Prettier as reference, so use it in case you don't know how to format something!

crates/rome_formatter/tests/specs/js/static_member.expected.js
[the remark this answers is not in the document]
Boshen: Shall we skip `#` for now? I played around a bit and the parser just breaks when it sees `#`

crates/rome_formatter/tests/specs/js/static_member.expected.js
MichaReiser: Should work fine after  #1851 landed

crates/rome_formatter/tests/specs/js/static_member.expected.js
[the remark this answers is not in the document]
MichaReiser: #1851  enables class field support

Boshen:
@MichaReiser @ematipico 
Rebased and updated with e534c7a and f4d5a2e

The formatter tests need a lot more work after looking at the prettier tests,
I'm currently just using the tests for TDD and guiding me through the parser.

crates/rome_formatter/src/ts/expressions/expression.rs
Boshen: done

crates/rslint_parser/test_data/inline/ok/import_meta.rast
Boshen: > We shouldn't have any parsing errors in any file inside of the ok folder and this is a parsing error. The easiest will be to just remove the whole grammar for now since it isn't supported anyway.

I have reverted the change.

crates/rome_formatter/src/ts/expressions/expression.rs
```diff
@@ -8,44 +17,242 @@ impl ToFormatElement for JsAnyExpression {
...
+			formatter.format_node(self.argument()?)?,
+		])
+	}
+}
+
+impl ToFormatElement for TsNonNull {
+	fn to_format_element(&self, _formatter: &Formatter) -> FormatResult<FormatElement> {
+		todo!()
+	}
+}
```
MichaReiser, line 241: Delete, empty implementations are misleading. It's better to get a compile error in that case

crates/rome_formatter/tests/specs/js/module.expected.js
```diff
@@ -0,0 +1 @@
+import.
```
MichaReiser, line 1: This doesn't seem to do any formatting. Can we delete this test (and the format implementation?)

MichaReiser (commented):
Looks good. Would you mind rebasing your changes and address the last two comments?

Boshen:
@MichaReiser rebased and squashed, removed:
* Template
* TsNonNull
* JsStaticMemberExpression (this introduced JsAnyName which is also a todo)

PR #1846, opened 2021-12-02
refactor: parse class

ematipico:
## Summary

This PR moves all methods inside `class.rs` to use the new syntax `ParsedSyntax`

- all methods that return `ParsedSyntax` now have the prefix `parse_` in their name
- applied error recovery inside the creation of the list of members
- split up some errors into two different files because the line of code highlighted was causing issues
	```js
	class extends {}  // <=== this one
	class B {}
	```
	Basically everything after the `{}` was being parsed - as expected - as an expression, causing loads of errors that were not related to what we wanted to test
- renamed boolean function with `is_at`
- added new tests
- fixed some range issue inside existing tests
- changed the grammar of the class modifiers: created a new unions called `JsAnyModifier` which contains, **for now**, only the `declare` modifier. It also contains the unknown node called `JsUnknownModifier` which will be used every time there's an error around them. The plan is to include `static`, `abstract`, etc. too.

## Test Plan

Existing CI


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -97,8 +97,9 @@ fn class(p: &mut Parser, kind: ClassKind) -> ConditionalParsedSyntax {
 			id if id.is_absent() && kind == ClassKind::Declaration && !guard.state.in_default => {
 				let err = guard
 					.err_builder("class declarations must have a name")
-					.primary(guard.cur_tok().range, "");
+					.primary(class_token.start..guard.cur_tok().range.start, "");
```
ematipico, line 100: This error was pointing to a single token, now it shows better

crates/rslint_parser/test_data/inline/err/block_stmt_in_class.rast
```diff
@@ -43,10 +44,10 @@ JsRoot {
         0: IDENT@6..7 "S" [] []
       2: L_CURLY@7..8 "{" [] []
       3: LIST@8..9
-        0: JS_UNKNOWN_MEMBER@8..9
+        0: JS_UNKNOWN_MEMBER@8..8
           0: (empty)
-          1: JS_UNKNOWN_MEMBER@8..9
-            0: L_CURLY@8..9 "{" [] []
+        1: JS_UNKNOWN_MEMBER@8..9
+          0: L_CURLY@8..9 "{" [] []
```
ematipico, line 50: This is still weird and I need to understand what's going on

crates/rslint_parser/test_data/inline/err/class_decl_err.rast
```diff
@@ -27,281 +27,165 @@ JsRoot {
...
   0: (empty)
   1: LIST@0..0
-  2: LIST@0..136
+  2: LIST@0..67
     0: JS_CLASS_DECLARATION@0..8
       0: CLASS_KW@0..6 "class" [] [Whitespace(" ")]
-      1: L_CURLY@6..7 "{" [] []
-      2: LIST@7..7
-      3: R_CURLY@7..8 "}" [] []
+      1: (empty)
```
ematipico, line 80: We are now adding a missing token for the name of the class (which matches our AST)

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -733,31 +737,23 @@ pub(crate) fn parse_equal_value_clause(p: &mut Parser) -> ParsedSyntax {
...
-	}
-
-	p.nth_at(offset, T!['(']) || p.nth_at(offset, T![<])
-}
-
-fn method_class_member(p: &mut Parser, m: Marker) -> CompletedMarker {
-	class_member_name(p).or_missing_with_error(p, js_parse_error::expected_function_body);
-	method_class_member_body(p, m)
+fn parse_method_class_member(p: &mut Parser, m: Marker) -> ParsedSyntax {
+	parse_class_member_name(p).or_missing_with_error(p, js_parse_error::expected_function_body);
```
MichaReiser, line 741: These functions don't strictly comply to the `-> ParsedSyntax` protocol. For example, they don't test if they can parse any valid node and otherwise return `Absent`. That's why I'm not sure if these methods should not instead return a `CompletedMarker` to make this clear (at least all the ones that are internal). 

This would also simplify the code above because you don't need to test if the result is `Present`.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -592,57 +615,38 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
...
-	SingleTokenParseRecovery::with_error(
-		token_set![T![;], T![ident], T![async], T![yield], T!['}'], T![#]],
-		JS_UNKNOWN_MEMBER,
-		err,
-	)
-	.recover(p);
-
-	member_marker.complete(p, JS_UNKNOWN_MEMBER)
+	member_marker.complete(p, JS_UNKNOWN_MEMBER);
+	Absent
```
MichaReiser, line 624: That's not valid. A `parse_*` method should either return `Absent` AND not progress parsing (completing a node progresses parsing) or progress parsing. You can return an `Invalid` ConditionalSyntax hear that the caller then must handle

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -115,11 +116,12 @@ fn class(p: &mut Parser, kind: ClassKind) -> ConditionalParsedSyntax {
 		}
 	}
 
+	// TODO: these two functions should return `ParsedSyntax`, so we can handle possible errors/missing/etc.
```
MichaReiser, line 119: TODO?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -115,11 +116,12 @@ fn class(p: &mut Parser, kind: ClassKind) -> ConditionalParsedSyntax {
...
 	extends_clause(&mut guard);
 	implements_clause(&mut guard);
 
 	guard.expect_required(T!['{']);
-	class_members(&mut *guard);
+	parse_class_members(&mut *guard).or_missing(&mut *guard);
```
MichaReiser, line 124: These can never be missing because it's a list.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -215,25 +217,37 @@ fn extends_clause(p: &mut Parser) {
 	m.complete(p, JS_EXTENDS_CLAUSE);
 }
 
-fn class_members(p: &mut Parser) -> CompletedMarker {
+fn parse_class_members(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 220: I would leave the method return type as a `CompletedMarker` (does it even need to return a marker at all? It isn't used by any caller)

* It's an internal `parse_` method
* It never returns `Absent`

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -215,25 +217,37 @@ fn extends_clause(p: &mut Parser) {
...
+			break;
+		}
 	}
 
-	members.complete(p, LIST)
+	Present(members.complete(p, LIST))
 }
 
-fn class_member(p: &mut Parser) -> CompletedMarker {
+fn parse_class_member(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 244: I think this method should return a `ConditionalParsedSyntax` in case it parsed a member but it turned out to be invalid.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -254,12 +268,12 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
...
 			parse_literal_member_name(p).ok().unwrap(); // bump declare as identifier
-			return method_class_member_body(p, member_marker);
-		} else if is_property_class_member(p, offset) {
+			return parse_method_class_member_body(p, member_marker);
+		} else if is_at_property_class_member(p, offset) {
 			parse_literal_member_name(p).ok().unwrap(); // bump declare as identifier
-			return property_class_member_body(p, member_marker);
+			return parse_property_class_member_body(p, member_marker);
 		} else {
 			let msg = if p.typescript() {
```
MichaReiser, line 278: This `else` branch will eat the `declare` keyword and wrap it in an `Error` for JS. I don't think we should do this and instead should return `Invalid` in that case (at the very end of parsing the member). Not sure how to do that best without having too much repetitive code because this method has so much branching. ... 

Another option could be to create a checkpoint at the begining of the parse method and rewind whenever you find an error and then return `Absent` so that the error recovery of the loop kicks in. 

Downside, we'll end up with two errors: 1) expected a member, 2) declare modifiers not valid.

Same for all the other errors created in this method.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -461,46 +475,55 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
...
 
-				p.error(err);
-			}
+					p.error(err);
+				}
 
-			constructor
+				Present(constructor)
+			} else {
+				Absent
```
MichaReiser, line 504: I believe returning `Absent` here isn't valid because the `parse_` method has already consumed some tokens that are now lost (e.g. all modifiers).

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -592,57 +615,38 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
...
 	member_marker: Marker,
 	member_name_kind: SyntaxKind,
-) -> CompletedMarker {
-	let property = property_class_member_body(p, member_marker);
-
-	if member_name_kind == JS_PRIVATE_CLASS_MEMBER_NAME {
-		let err = p
-			.err_builder("private class properties with `declare` are invalid")
-			.primary(property.range(p), "");
+) -> ParsedSyntax {
```
MichaReiser, line 633: I wouldn't change the return type of these internal methods because

* parse_members expects that they succeed parsing the child. `parse_members` is a special union rule because it "consumes" tokens itself. So, it would be invalid if `parse_members` eat all modifiers and then this method returns `Absent`
* The function also accepts an already started marker
* The method never returns `Absent`

MichaReiser (changes requested):
Thanks for extending the test cases and tackling classes in the first place! 

The parse methods still add many `ERROR` nodes in places they're not permitted. I suggest adding test cases for all of them and then figure out a way how to avoid them

The `parse_members` method often *progresses* parsing and then returns `Absent`. This violates the contract of `ParsedSyntax` that guarantees a caller that a parse method eats no tokens nor produces any errors if it returns `Absent`. This is important because we otherwise loose tokens.

crates/rslint_parser/src/syntax/class.rs
ematipico: I don't know how to reply to this comment

crates/rslint_parser/src/syntax/class.rs
ematipico: Yeah makes sense, I see now that it's a recurring pattern

crates/rslint_parser/src/syntax/class.rs
MichaReiser: haha... I was wondering if this is something you want to address later or if it's something that should be done as part of this PR

ematipico: LOL! It's something I want to address later and added a TODO about what should be done

crates/rslint_parser/src/syntax/class.rs
ematipico: Would it be OK to leave a `TODO` here? It's more related to the typescript parsing and not to the whole refactor to `ParsedSyntax`. I mean, the bug it's been there for some time and can be fixed anytime we want

crates/rslint_parser/test_data/inline/err/class_constructor_err.rast
```diff
@@ -0,0 +1,80 @@
...
+            1: LIST@32..32
+            2: LIST@32..32
+            3: R_CURLY@32..34 "}" [] [Whitespace(" ")]
+      4: R_CURLY@34..35 "}" [] []
+  3: EOF@35..36 "" [Whitespace("\n")] []
+--
+error[SyntaxError]: constructors cannot be static
+  ┌─ class_constructor_err.js:1:11
+  │
+1 │ class B { static constructor() {} }
```
MichaReiser, line 76: Can we add a suggestion, remove the `static` keyword here?

crates/rslint_parser/test_data/inline/err/class_decl_err.rast
```diff
@@ -27,281 +27,167 @@ JsRoot {
...
                     l_paren_token: missing (required),
-                    parameter: JsObjectBinding {
-                        l_curly_token: L_CURLY@69..70 "{" [] [],
-                        properties: [],
-                        r_curly_token: R_CURLY@70..72 "}" [] [Whitespace(" ")],
-                    },
+                    parameter: missing (required),
                     r_paren_token: missing (required),
-                    body: missing (required),
+                    body: JsFunctionBody {
```
MichaReiser, line 46: Nice, Thanks!

MichaReiser (commented):
Can we give this PR a more descriptive title: 

Refine class error recovery (altough, we don't do this completely)

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -215,25 +218,61 @@ fn extends_clause(p: &mut Parser) {
...
+		// an error down the line, we might have already progressed the parser.
+		//
+		// So when there's an error, we mark the whole marker as invalid
+		// and here we abandon the invalid marker and later we try to recover.
+		//
+		//  Not that if we get a [ParsedSyntax::Absent], we want to panic
+		let to_recover = match parsed_member {
+			Valid(marker) => Present(marker),
+			Invalid(conditional_syntax) => {
+				conditional_syntax.abandon(p);
```
MichaReiser, line 247: This doesn't work as you expect. It will expect all that has currently been parsed to the list instead of wrapping it inside a `UNKNOWN_MEMBER`. See my comment above on how to do error recovery in this case.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -215,25 +218,61 @@ fn extends_clause(p: &mut Parser) {
 	m.complete(p, JS_EXTENDS_CLAUSE);
 }
 
-fn class_members(p: &mut Parser) -> CompletedMarker {
+/// Parses a list of class members
+///
+/// # Panics
+///
+/// The method can panic if [parse_class_member] returns [ParsedSyntax::Absent]
```
MichaReiser, line 225: Is this still the case?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -215,25 +218,61 @@ fn extends_clause(p: &mut Parser) {
...
+
+		// In this case we want to mark members that have errors as invalid.
+		//
+		// This is because the parsing of members is incremental and when we encounter
+		// an error down the line, we might have already progressed the parser.
+		//
+		// So when there's an error, we mark the whole marker as invalid
+		// and here we abandon the invalid marker and later we try to recover.
+		//
+		//  Not that if we get a [ParsedSyntax::Absent], we want to panic
```
MichaReiser, line 243: Is this still true? 

What you could do instead (instead of the complicated error recovery I suggest above) is to create a checkpoint right before parsing the member and then restore it in the case the parsed member is invalid (to rewind the parser and throw away any errors). The next step is then to trigger the error recovery as you're used to.

crates/rslint_parser/src/syntax/class.rs
ematipico: > I don't feel super comfortable about this because it is not TypeScript specific.

Actually, it is. All the errors where we have `ERROR` here are due to TypeScript syntax, and the use of particular keyword like `declare` inside a JavaScript file.  

In order to properly fix all these errors, we have to rethink few stuff:

- proper use of `ParsedSyntax<ConditionalSyntax>` inside logic where we expect TS or not;
- re think the grammar of the modifiers. I think we can use the same logic we proposed for class names, and introduce something like `JsAnyModifier` (`static` falls under the modifier definition). In this way, we can better invalidate the green without invalidating the whole class member. I did this change and it will be included in this PR for handling the `declare` keyword;

> My issue with that is that it is difficult to know what progress we make.

We're both in the same both, but this PR has already loads of feedback and it requires more changes, in the logic too.

MichaReiser: Agree on the outcome. It's just that the current recovery will "break" js files as well.

xtask/js.ungram
```diff
@@ -554,7 +555,7 @@ JsAnyConstructorParameter =
 // class Test { a; b = 1; static c; }
 //              ^^ ^^^^^^ ^^^^^^^^^
 JsPropertyClassMember =
-	'declare'?
+	modifiers: JsAnyModifier?
```
MichaReiser, line 558: Should modifiers be a list? If so, should access, abstract, static and so on be modifiers too?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -873,56 +936,117 @@ fn ts_access_modifier<'a>(p: &'a Parser) -> Option<&'a str> {
...
+			} else {
+				p.bump_remap(kind);
+			}
+		}
+		if self.declare {
+			let m = p.start();
+			p.bump_remap(T![declare]);
+			if let Some(err) = self.declare_err {
+				p.error(err);
+				m.complete(p, JS_UNKNOWN_MEMBER);
```
MichaReiser, line 1020: Why does this create a new member?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -873,56 +936,117 @@ fn ts_access_modifier<'a>(p: &'a Parser) -> Option<&'a str> {
...
-		} else {
-			p.bump_remap(kind);
+impl Modifiers {
+	pub fn new(
+		declare: bool,
+		accessibility: bool,
+		is_static: bool,
+		dont_remap_static: bool,
+		declare_err: Option<Diagnostic>,
+	) -> Self {
```
MichaReiser, line 972: Nit: Initializing (some/most) fields with defaults and instead have `setters` may simplify the code a bit? Mainly because it's hard to guess the parameters when reading something like: `true, false, false, true`

xtask/js.ungram
MichaReiser: I just saw your explanation in the summary. 

I'm a bit undecided if a `modifiers` list is a good idea. Certainly, there's the advantage that we can recover from invalid modifiers but there's the downside that we can't enforce the correct order, a transformer could construct `static declare public test()`; 

What's the main benefit of changing the declaration here even if we don't support ts?

Why is this only for properties and not all other members?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -250,34 +287,42 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
 	if declare {
 		offset += 1;
 	}
+	let mut declare_remapped = false;
+	let mut declare_diagnostic = None;
```
MichaReiser, line 291: Nit. It seems `declare_remapped` always seems to be true when `declare_diagnostic` is some. Is it sufficient to only track one state variable?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -250,34 +287,42 @@ fn class_member(p: &mut Parser) -> CompletedMarker {
...
 				"`declare` modifiers can only be used in TypeScript files"
 			};
 			let err = p.err_builder(msg).primary(p.cur_tok().range, "");
 
-			p.error(err);
-			let m = p.start();
-			p.bump_any();
-			m.complete(p, ERROR);
+			declare_diagnostic = Some(err);
+			declare_remapped = true;
```
MichaReiser, line 317: It doesn't seem like `declare` gets remapped here. What's the meaning of `remapped?`

xtask/js.ungram
ematipico: > the downside that we can't enforce the correct order

I don't think it will be an issue. I think there aren't rules around ordering. If so, we can have a lint rule for that.

> What's the main benefit of changing the declaration here even if we don't support ts?

I wanted to address your previous feedback and ended up starting this solution, which I didn't want to revert. My next PR will apply this new grammar to all the modifiers.

> Why is this only for properties and not all other members?

It will be for all members

crates/rslint_parser/src/syntax/class.rs
ematipico: Good catch, it should be a `JS_UNKNOWN_MODIFIER`

crates/rslint_parser/src/syntax/class.rs
ematipico: That was not needed anymore, removed it.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -873,56 +931,154 @@ fn ts_access_modifier<'a>(p: &'a Parser) -> Option<&'a str> {
...
+impl Default for Modifiers {
+	fn default() -> Self {
+		Self {
+			declare: Default::default(),
+			accessibility: Default::default(),
+			is_static: Default::default(),
+			dont_remap_static: Default::default(),
+			declare_err: Default::default(),
 		}
 	}
```
MichaReiser, line 969: You can derive `default` with `#[derive(Default)]` because all type implement default

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -873,56 +931,154 @@ fn ts_access_modifier<'a>(p: &'a Parser) -> Option<&'a str> {
...
+	// 	declare_err: Option<Diagnostic>,
+	// ) -> Self {
+	// 	Self {
+	// 		declare,
+	// 		accessibility,
+	// 		is_static,
+	// 		dont_remap_static,
+	// 		declare_err,
+	// 	}
+	// }
```
MichaReiser, line 987: Delete?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -873,56 +931,154 @@ fn ts_access_modifier<'a>(p: &'a Parser) -> Option<&'a str> {
...
-	if static_ && !dont_remap_static {
-		p.bump_remap(STATIC_KW);
-	} else if static_ && dont_remap_static {
-		// Guaranteed to be at the static keyword, parsing a class member must succeed
-		class_member_name(p).ok().unwrap();
+
+	pub fn dont_remap_static(mut self, should_remap: bool) -> Self {
+		self.dont_remap_static = should_remap;
+		self
+	}
```
MichaReiser, line 1007: Nit: I would probably create two methods, because `dont_remap_static(true)` is a bit confusing. Or just name all of them `set_remap_static`, `set_accessibility`, etc (`has_accessibility` is also a bit confusing because `has` is often used to query a `boolean` value, not to set one)

MichaReiser (approved):
Let's go!

Would you mind adding a TODO to the modifier in the grammar to make it clear, this isn't finalized yet?

PR #1849, opened 2021-12-03
feat(xtask): Generate detailed coverage report for `xtask compare`

Boshen:
## Summary
HashSet is used to properly get all the diffs, the original implementation only compared from base - new, which can miss out some tests

## Test Plan
```
cargo xtask coverage --json > base_results.json 
```
Break the parser (comment some random code out), then
```
cargo xtask coverage --json > new_results.json
cargo xtask compare ./base_results.json ./new_results.json  --markdown
```

### DIFF
--- a/xtask/src/compare/results.rs
+++ b/xtask/src/compare/results.rs
@@ -1,7 +1,9 @@
 use crate::coverage::files::{Outcome, TestResult, TestResults};
 use ascii_table::{AsciiTable, Column};
 use colored::Colorize;
-use std::{collections::HashMap, ffi::OsStr, fs::File, path::Path};
+use std::collections::HashSet;
+use std::iter::FromIterator;
+use std::{fs::File, path::Path};
 
 pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
 	let base_results: TestResults =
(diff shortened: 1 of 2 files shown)


xtask/src/compare/results.rs
```diff
@@ -214,62 +191,87 @@ pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
...
+	let failed_to_panic = base_failed.intersection(&new_panicked);
+
+	// PanicToFailed: panicked on main, failing on the feature branch (kind of good?)
+	let panic_to_failed = base_panicked.intersection(&new_failed);
+
+	// New: Newly added tests
+	let added_tests = new_all.difference(&base_all);
+
+	// Removed: Removed tests
+	let removed_tests = base_all.difference(&new_all);
```
MichaReiser, line 267: Nit: This works and I like its simplicity but it means we have to copy quite a lot of data. 

An alternative would be to 

1. Create tuples of `(main: Option<Outcome>, feature: Option<Outcome>)` .
2. Iterate over the tuples and push the path into the right vector depending on the comparison of the two outcomes.

MichaReiser:
How does the result look now?

Boshen:
@MichaReiser  I pushed a changed to not clone everything. Example output for 969b05d:
| Total | 17608 | 17608 | 0 |
| Passed | 17084 | 17074 | ❌ ⏬ **-10** |
| Failed | 523 | 533 | ❌ ⏫ **+10** |
| Panics | 1 | 1 | 0 |
| Coverage | 97.02% | 96.97% | **-0.06%** |

<details><summary><b>:fire: Regression (10):</b></summary>

```
xtask/src/coverage/test262/test/language/expressions/arrow-function/syntax/early-errors/arrowparameters-bindingidentifier-no-arguments.js
xtask/src/coverage/test262/test/language/reserved-words/await-module.js
xtask/src/coverage/test262/test/language/statements/class/class-name-ident-await-module.js
xtask/src/coverage/test262/test/language/statements/labeled/value-await-module.js
xtask/src/coverage/test262/test/language/expressions/arrow-function/syntax/early-errors/arrowparameters-bindingidentifier-no-eval.js
xtask/src/coverage/test262/test/language/expressions/class/class-name-ident-let.js
xtask/src/coverage/test262/test/language/statements/class/class-name-ident-let.js
xtask/src/coverage/test262/test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-strict.js
xtask/src/coverage/test262/test/language/expressions/class/class-name-ident-await-module.js
xtask/src/coverage/test262/test/language/future-reserved-words/let-strict.js
```
</details>

Boshen:
Sorted the results:

<details><summary><b>:fire: Regression (10):</b></summary>

```
xtask/src/coverage/test262/test/language/expressions/arrow-function/syntax/early-errors/arrowparameters-bindingidentifier-no-arguments.js
xtask/src/coverage/test262/test/language/expressions/arrow-function/syntax/early-errors/arrowparameters-bindingidentifier-no-eval.js
xtask/src/coverage/test262/test/language/expressions/class/class-name-ident-await-module.js
xtask/src/coverage/test262/test/language/expressions/class/class-name-ident-let.js
xtask/src/coverage/test262/test/language/future-reserved-words/let-strict.js
xtask/src/coverage/test262/test/language/reserved-words/await-module.js
xtask/src/coverage/test262/test/language/statements/class/class-name-ident-await-module.js
xtask/src/coverage/test262/test/language/statements/class/class-name-ident-let.js
xtask/src/coverage/test262/test/language/statements/labeled/value-await-module.js
xtask/src/coverage/test262/test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-strict.js
```
</details>

ematipico (approved):

MichaReiser (commented):
I'm a bit concerned about the `Eq` and `Hash` overrides in a way that is probably not what people expect. I think it should be possible to implement the same without having to override `Eq` and `Hash` that is also closer to what we currently have.

MichaReiser (changes requested):
> I'm a bit concerned about the Eq and Hash overrides in a way that is probably not what people expect. I think it should be possible to implement the same without having to override Eq and Hash that is also closer to what we currently have.

xtask/src/compare/results.rs
```diff
@@ -125,61 +127,40 @@ pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
...
+					"<details><summary><b>{} ({}):</b></summary>",
+					title,
+					tests.len()
+				);
+				println!("\n```");
+				let mut paths = tests
+					.iter()
+					.map(|test| test.path.as_os_str().to_str().unwrap())
+					.collect::<Vec<&str>>();
+				paths.sort();
```
MichaReiser, line 143:
```suggestion
				paths.sort_unstable();
```

Boshen:
I'll try the old implementation

xtask/src/compare/results.rs
```diff
@@ -215,61 +196,79 @@ pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
...
+		.partition(|result| result.outcome == Outcome::Passed);
+	let (new_passed, new_not_passed): (Vec<&TestResult>, Vec<&TestResult>) = new_results
+		.details
+		.iter()
+		.partition(|result| result.outcome == Outcome::Passed);
+
+	let base_all: HashSet<&TestResult> = HashSet::from_iter(base_results.details.iter());
+	let base_passed: HashSet<&TestResult> = HashSet::from_iter(base_passed);
+	let base_not_passed: HashSet<&TestResult> = HashSet::from_iter(base_not_passed);
+	let base_failed: HashSet<&TestResult> = base_results
```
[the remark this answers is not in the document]
MichaReiser: ... `all_paths` should probably be a set or otherwise guarantee `uniqueness`.

MichaReiser:
Oh, you don't have to create a new PR. The proposed changes are local to the "diffing" and the rest is amazing!

ematipico:
I would personally stick to the old implementation. The `match` was amazing and it makes **more clear** of what we are doing - the new implementation is clear too much it needs some comment to understand what we are actually doing.

PR #1853, opened 2021-12-03
feat: parse lists

### LINKED ISSUE #1836: 📎 Create generic way to parse lists
> @ematipico There's a lot of repetitive code when it comes to parsing lists. Do you want to take it on to look into a more generic way of parsing lists?

_Originally posted by @MichaReiser in https://github.com/rome/tools/issues/1834#issuecomment-982594834_

ematipico:
## Summary

This PR closes #1836 

I created two traits - for now - that implement two different kind of lists:
- `ParseNodeList`: normal lists, nothing fancy
- `ParseSeparatedList`: lists separated by a recurring separating element, which will be always the same, e.g.: `["foo", 1, true]`, where `,` is the recurring separating element

`ParseNodeList` and `ParseSeparatedList` have the same API and they are not shared.
Myself and @MichaReiser agreed on the fact that it's better to make the developer experience easier instead of making maintainability a priority.

The PR moves one of the lists to this new methodology and there should not be regressions. It also shows how would they work. 

## Test Plan

Current CI


crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,183 @@
...
+use rslint_errors::Diagnostic;
+use rslint_syntax::SyntaxKind;
+use std::ops::Range;
+
+/// An utility that gives some control on how to parse a list of elements
+pub struct ParseList<Condition, CreateElement> {
+	/// This is a condition used inside the loop. When it's false, it exists the loop
+	condition: Option<Condition>,
+	/// A closure that is charge of creating the element inside the loop
+	create_element: Option<CreateElement>,
```
MichaReiser, line 13: Nit: `create` sounds the element is created out of the blue (just a new element) but what it's doing is to `parse` an element. That's why I would rename it to `parse_element`.

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,183 @@
...
+{
+	pub fn new() -> Self {
+		Self {
+			condition: None,
+			create_element: None,
+			list_kind: None,
+		}
+	}
+
+	#[must_use = "You need to provide the condition to exit from the loop"]
```
MichaReiser, line 31: [`must_use`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute) specifies that the returned value must be used, but not that a certain method must be called. 

So, the `must_use` protects` here from the case where the returned builder is not used, but that's already guaranteed by the compiler, because the method "consumes" the builder.

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,183 @@
...
+
+		let list_kind = self
+			.list_kind
+			.expect("`list_kind` can't be `None`, use set_condition");
+
+		let elements = p.start();
+		let mut progress = ParserProgress::default();
+		while (condition)(p) {
+			progress.assert_progressing(p);
+			(create_element)(p).or_missing(p);
```
MichaReiser, line 72: A list shouldn't contain "missing" elements, except if it is a separated list. 

It's also unsafe to call `or_missing` without any recovery because the parser didn't progress when parsing the element but won't progress in the next loop as well because it's still at the same position.

That's why I don't think we should offer a `parse_list` function without any recovery and a parse function that returns `ParsedSyntax`. No recovery and a `parse` method that always returns an item or recovers.

MichaReiser (commented):
I saw that you requested feedback which is why I did a review. I hope that's what you intended. 

I'm not sure what the rustiest way of implementing `parse_list` is but using traits worked well when implementing the `ParsePattern` helpers. 

The reason why traits could work well is that we may be able to implement a generic `ParseList` trait that implements the "loop" but delegates most to methods that different trait implementations can override if they want. 

The main downside is that it now requires a struct implementing that trait, which is slightly more code but we could get away with helper methods like `parse_array` that use good defaults and have `setter` methods to customise the behaviour. 

```rust
trait ParseList<T> {
 fn parse_list(&mut self, p: &mut Parser) -> ParsedSyntax {
  // TODO should this also eat the open / close brackets? 
  let list = self.start_list(p); // Or just call p.start()? 
  let mut progress = ParserProgress:default();
  
 while (!p.at(EOF) && !self.is_at_end(p)) {
     progress.assert_progressing(p);
     let item = self.parse_item(p);
     
     if item.is_absent() && self.recover(p).is_err()  {
       break;
     }
  }
  
  self.finish_list(p, list)
 }
 
 fn recover(&mut self, p: &mut Parser) -> RecoveryResult;
 fn start_list(&mut self, p: &mut Parser) -> Marker;
 fn finish_list(&mut self, p: &mut Parser) -> ParsedSyntax<CompletedMarker>;
 fn is_at_end(&mut self, p: &mut Parser) -> bool;
}

```

I believe that this could be flexible enough to even implement parsing for a separated list because the list is allowed to track some state internally. 

`T` so that we can implement a `ParseList<ParsedSyntax>` and `ParseList<ConditionalSyntax>`?

ematipico:
> I saw that you requested feedback which is why I did a review. I hope that's what you intended.
> 

Yes, that's exactly what I wanted, thank you! :) 

> I'm not sure what the rustiest way of implementing `parse_list` is but using traits worked well when implementing the `ParsePattern` helpers.
> 
> The reason why traits could work well is that we may be able to implement a generic `ParseList` trait that implements the "loop" but delegates most to methods that different trait implementations can override if they want.
> 

That was my other solution 😅

> The main downside is that it now requires a struct implementing that trait, which is slightly more code but we could get away with helper methods like `parse_array` that use good defaults and have `setter` methods to customise the behaviour.
> 
> ```rust
> trait ParseList<T> {
>  fn parse_list(&mut self, p: &mut Parser) -> ParsedSyntax {
>   // TODO should this also eat the open / close brackets? 
>   let list = self.start_list(p); // Or just call p.start()? 
>   let mut progress = ParserProgress:default();
>   
>  while (!p.at(EOF) && !self.is_at_end(p)) {
>      progress.assert_progressing(p);
>      let item = self.parse_item(p);
>      
>      if item.is_absent() && self.recover(p).is_err()  {
>        break;
>      }
>   }
>   
>   self.finish_list(p, list)
>  }
>  
>  fn recover(&mut self, p: &mut Parser) -> RecoveryResult;
>  fn start_list(&mut self, p: &mut Parser) -> Marker;
>  fn finish_list(&mut self, p: &mut Parser) -> ParsedSyntax<CompletedMarker>;
>  fn is_at_end(&mut self, p: &mut Parser) -> bool;
> }
> ```
> 
> I believe that this could be flexible enough to even implement parsing for a separated list because the list is allowed to track some state internally.
> 
> `T` so that we can implement a `ParseList<ParsedSyntax>` and `ParseList<ConditionalSyntax>`?

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	#[must_use = "You need to provide the condition to exit from the loop"]
+	#[inline]
+	pub fn set_condition(mut self, condition: Condition) -> Self {
+		self.condition = Some(condition);
+		self
+	}
+
+	#[must_use = "You need provide the function that creates the single element list"]
```
xunilrj, line 38: Same thing here about @MichaReiser comment.

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,183 @@
...
+		self.list_kind = Some(list_kind);
+		self
+	}
+
+	/// This function is in charge to create a list without any sort of recovering.
+	///
+	/// When using it, if [create_element] returns a [ParsedSyntax::Absent], it will
+	/// mark it as a missing element
+	pub fn create_list(self, p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
+		let condition = self
```
xunilrj, line 57: If you use the builder pattern I suggested above, you will probably not need to check this, because this field will not be ```Option<>```.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1105,14 +1105,13 @@ fn parse_switch_clause(
...
-			let mut progress = ParserProgress::default();
 
-			while !p.at_ts(token_set![T![default], T![case], T!['}'], EOF]) {
-				progress.assert_progressing(p);
-				parse_statement(p, None);
-			}
-			cons_list.complete(p, LIST);
+			ParseList::new()
+				.set_condition(|p| !p.at_ts(token_set![T![default], T![case], T!['}'], EOF]))
+				.set_create_element(|p| Present(parse_statement(p, None).unwrap()))
```
xunilrj, line 1111: We can do

```rust
.set_create_element(|p| parse_statement(p, None).into())
```

and avoid the ```.unwrap()```, right?

xunilrj:
> I'm not sure what the rustiest way of implementing `parse_list` ...

What I see other parser combinators libs doing is:

https://docs.rs/nom/latest/nom/multi/fn.separated_list1.html
```rust
pub fn separated_list1<I, O, O2, E, F, G>(
    sep: G, 
    f: F
) -> impl FnMut(I) -> IResult<I, Vec<O>, E> 
where
    I: Clone + InputLength,
    F: Parser<I, O, E>,
    G: Parser<I, O2, E>,
    E: ParseError<I>
{
...
}
```

lovely if you hate one-letter-variables.... :stuck_out_tongue:


https://docs.rs/chumsky/0.6.0/chumsky/trait.Parser.html#method.separated_by

```rust
fn separated_by<U, P: Parser<I, U>>(self, other: P) -> SeparatedBy<Self, P, U>
where
    Self: Sized, 
{
    ...
}
```

I have the impression that ```nom``` style is closer to ours.

ematipico:
I implemented the proposed solution using `trait`

@MichaReiser : your proposed solution as `ParseList<T>` at the moment is not implemented. I need to review the architecture of the trait first.

If we decided to support `ParseList<ParsedSyntax>` and `ParseList<ConditionalSyntax>`, we'll need to implement an internal `type` and some functions (like `finish_list`) can't have a default implementation.

ematipico:
> I implemented the proposed solution using `trait`
> 
> @MichaReiser : your proposed solution as `ParseList<T>` at the moment is not implemented. I need to review the architecture of the trait first.
> 
> If we decided to support `ParseList<ParsedSyntax>` and `ParseList<ConditionalSyntax>`, we'll need to implement an internal `type` and some functions (like `finish_list`) can't have a default implementation.

Just implemented it using the `Self::FinishedSyntax` internal type.

Now the trait must implement `finish_list` method. Do you think we should implement it for `parse_element` too?

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,103 @@
+use super::{ParsedSyntax, ParserProgress, RecoveryResult};
+use crate::{CompletedMarker, Marker, ParseRecovery, Parser};
+use rslint_errors::Diagnostic;
+use rslint_syntax::SyntaxKind;
+use std::ops::Range;
+
+/// An utility that gives finer control on how to parse a list of elements
+pub trait ParseList {
+	/// The type contained inside [ParsedSyntax]
+	type FinishedSyntax;
```
MichaReiser, line 10: Oh, that's neat. I was mainly thinking to support `ParseList<T>` where `T` is the return type of `parse_element`.

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,103 @@
...
+
+	/// This method is used to check the current token inside the loop. When this method return [false],
+	/// the trait will exit from the loop.
+	fn is_at(&mut self, p: &mut Parser) -> bool;
+
+	/// When calling [parse_separated_list], this method checks, inside the loop, if the parser
+	/// is at a position where the current token is element that will separate the list
+	fn is_at_separating_element(&mut self, _p: &mut Parser) -> bool {
+		unimplemented!("When calling `parse_separated_list`, you need to implement this method.");
+	}
```
MichaReiser, line 87: I would probably create a separate trait for parsing separated lists and then not implement this method.

crates/rslint_parser/src/parser/parse_list.rs
```diff
@@ -0,0 +1,103 @@
...
+	fn start_list(&mut self, p: &mut Parser) -> Marker {
+		p.start()
+	}
+
+	/// It creates a [ParsedSyntax] that will contain the list
+	fn finish_list(&mut self, p: &mut Parser, m: Marker) -> ParsedSyntax<Self::FinishedSyntax>;
+
+	/// This method is used to check the current token inside the loop. When this method return [false],
+	/// the trait will exit from the loop.
+	fn is_at(&mut self, p: &mut Parser) -> bool;
```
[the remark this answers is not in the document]
ematipico: The first time I read this proposal I thought it was syntax sugar to check if we are at the end of the file, which is not what you meant. Although I don't have other naming alternatives 😢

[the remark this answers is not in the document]
MichaReiser: `is_at_list_end`?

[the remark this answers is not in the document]
ematipico: Sounds good

crates/rslint_parser/src/lib.rs
```diff
@@ -337,13 +339,13 @@ pub trait SyntaxFeature: Sized {
 	/// Creates a syntax that is only valid if the current parsing context doesn't support this syntax feature,
 	/// and adds a diagnostic if it does.
 	///
-	/// Returns [Present(Valid)] if the parsing context doesn't support this syntax feature and `syntax` is [Present].
+	/// Returns [Present(Valid)] if the parsing context doesn't support this syntax feature and `syntax` is [ParsedSyntax::Present].
 	///
 	/// Returns [Present(Invalid)], calls the `error_builder` to create a diagnostic,
 	/// adds the diagnostic to the parsing diagnostics, and returns [Present(Invalid)]
-	/// if the parsing context does support this syntax feature and the `syntax` is [Present].
+	/// if the parsing context does support this syntax feature and the `syntax` is [ParsedSyntax::Present].
```
MichaReiser, line 346: Is this the proper syntax to use? Sad, that rust isn't telling me about syntax errors in docs.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
+///! A set of traits useful to parse various types of lists
+use super::{ParsedSyntax, ParserProgress, RecoveryResult};
+use crate::{Marker, ParseRecovery, Parser};
+use rslint_errors::Diagnostic;
+use rslint_syntax::SyntaxKind;
+use std::ops::Range;
+
+/// A generic that defines a generic behaviour for all the possible lists.
+///
+pub trait List {
```
MichaReiser, line 10: Nit: Rust traits are often verbs, which is why I would call this `ParseList`. Or from another perspective, the trait itself isn't a list, it only knows how to parse a list.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
...
+	/// The type returned when calling the function [Self::parse_element]
+	type ParsedElement;
+
+	/// Parses a single element of the list
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement>;
+
+	/// The [SyntaxKind] used to name the list
+	fn list_kind(&mut self) -> SyntaxKind {
+		SyntaxKind::LIST
+	}
```
MichaReiser, line 20: We may need to differentiate here. All our lists always contain a `LIST` node. For example, array expression contains a list with all the elements. I'm not sure if it ever makes sense to override this except if the `ParseList` also parses the start/end tokens of e.g. an `ArrayExpression` in which case the "list" would be the whole array expression node.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
...
+
+	/// This method is used to recover the parser in case [Self::parse_element] returns [ParsedSyntax::Absent]
+	fn recover(
+		&mut self,
+		p: &mut Parser,
+		parsed_element: ParsedSyntax<Self::ParsedElement>,
+	) -> RecoveryResult;
+
+	/// [Diagnostic] thrown in case the parser is not able to recover
+	fn expected_element_error(p: &Parser, range: Range<usize>) -> Diagnostic;
```
MichaReiser, line 41: Is this still needed since we don't have a default `recovery` implementation?

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
...
+			progress.assert_progressing(p);
+
+			let parsed_element = self.parse_element(p);
+
+			if self.recover(p, parsed_element).is_err() {
+				break;
+			}
+		}
+		self.finish_list(p, elements)
+	}
```
MichaReiser, line 86: Should we unify the "normal" list with the "list" trait. Specialisations can still override `parse_list` to e.g. account for separators etc.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
...
+
+	/// It creates a [ParsedSyntax] that will contain the list
+	fn finish_list(&mut self, p: &mut Parser, m: Marker) -> ParsedSyntax<Self::ParsedList>;
```
MichaReiser, line 89: What's the reason that this isn't implemented, considering that the trait knows the list kind? 

Are there use cases where we need to override `start/finish_list`

It would be nice if `finish_list` creates a `missing` marker for empty lists. The facade should be able to handle empty lists.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
...
+/// }
+/// ```
+pub trait ParseSeparatedList: List {
+	/// The type of syntax that will be returned by [Self::parse_list]. The type will be the generic for [ParsedSyntax]
+	type ParsedList;
+
+	/// Tells the parser to mark the current token as missing, continuing the loop.
+	/// This function is used for [Self::parse_list].
+	fn parse_separating_element(&mut self, p: &mut Parser) {
+		p.missing();
```
MichaReiser, line 116: This shouldn't add a missing marker but instead bump the separator token. 

We can probably unify the `parse_separator` and `is_at_separator` methods by introducing `expect_separator` which expects the separator token (and adds the missing marker)

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,152 @@
...
+	/// It panics if the parser doesn't advance at each cycle of the loop
+	fn parse_list(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedList> {
+		let elements = self.start_list(p);
+		let mut progress = ParserProgress::default();
+		while !p.at(SyntaxKind::EOF) && !self.is_at_list_end(p) {
+			progress.assert_progressing(p);
+
+			if self.is_at_separating_element(p) {
+				self.parse_separating_element(p);
+				continue;
```
MichaReiser, line 141: You should continue here. Each separator should be followed by  an element except if we're at the end of the list, in which case you want to break
```suggestion
				if self.is_at_list_end(p) { 
				  break;
				}
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1054,6 +1054,47 @@ pub fn parse_for_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
+		match parse_statement(p, None) {
+			Some(stmt) => Present(stmt),
+			None => Absent,
+		}
+	}
+
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		p.at_ts(token_set![T![default], T![case], T!['}']])
```
MichaReiser, line 1071: `default` / `case` don't mark the end of the list?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1054,6 +1054,47 @@ pub fn parse_for_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+	}
+
+	fn expected_element_error(p: &Parser, range: Range<usize>) -> Diagnostic {
+		js_parse_error::expected_case(p, range)
+	}
+
+	fn recovery() -> ParseRecovery {
+		ParseRecovery::new(JS_UNKNOWN_STATEMENT, STMT_RECOVERY_SET)
```
MichaReiser, line 1079: An `unknown` statement is no valid child for a switch clause.

crates/rslint_parser/src/lib.rs
ematipico: You need to run `cargo doc` if you want to see the warnings. Yes, its' the correct syntax.

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: `PreparedList` or `PrepareList`?

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: > I'm not sure if it ever makes sense to override

Thinking about cross parser/language, I thought that it would make sense to give the possibility to change the syntax kind. 

Are you saying that we would return the `Marker` instead of the `CompletedMarker` and let the consumer to wrap it into a `LIST` based on the situation?

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: This is where it gets tricky. If not implemented, the `recover` method would panic. Same if `recovery()` is not implemented. How do we enforce this using the compiler? Is the documentation enough?

crates/rslint_parser/src/parser/parse_lists.rs
MichaReiser: I personally would go with `ParseList`, also matches the main method which is called `parse_list`.

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: > What's the reason that this isn't implemented, considering that the trait knows the list kind?

We don't know the type of `Self::ParsedList`, this means that we can't have a default implementation. 

We can't do

```rust
Present(m.complete(p, self.list_kind()))
```

Because the types don't match

> Are there use cases where we need to override start/finish_list

Probably in cases where the list is a `ConditionalSyntax`

> It would be nice if finish_list creates a missing marker for empty lists

We could enforce it but at the moment lists are not empty in the current implementations. When they are empty, they are still there.

crates/rslint_parser/src/parser/parse_lists.rs
MichaReiser: LIST is probably going to be a language-agnostic kind, the same as EOF

A list can never be absent (it's either empty or has elements). This is why I'm not sure if it even makes sense to return anything from the `parse_list` method.

crates/rslint_parser/src/parser/parse_lists.rs
MichaReiser: Not sure what you mean. I would remove this method as it's never needed by any implemented method of `ParseList`. Define the method in a specialised trait if any of them need it

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Sorry I didn't understand the comment

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: > A list can never be absent (it's either empty or has elements). This is why I'm not sure if it even makes sense to return anything from the parse_list method.

Oh that. I returned something for testing. I can revert the change.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: `is_at_list_end` tests if the list is at the end... But a `default` token doesn't signal the end of the list, it signals that the next element in the list is a default case.

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: I'd like to keep it separate for now. There will be cases where we will have different separated list, I think.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1054,6 +1054,43 @@ pub fn parse_for_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+	}
+
+	fn recover(
+		&mut self,
+		p: &mut Parser,
+		parsed_element: ParsedSyntax<Self::ParsedElement>,
+	) -> parser::RecoveryResult {
+		parsed_element.or_recover(
+			p,
+			&ParseRecovery::new(JS_UNKNOWN_EXPRESSION, STMT_RECOVERY_SET),
```
MichaReiser, line 1081: `JS_UNKOWN_STATEMENT`... and I finally understood that this is parsing the statements, not the switch cases... sorry

ematipico: no problem!

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,156 @@
...
+	/// If present, it [parses](Self::parse_separating_element) it and continues with loop.
+	/// If not present, it adds a missing marker.
+	fn expect_separator(&mut self, p: &mut Parser) -> bool {
+		if p.expect_required(self.separating_element_kind()) {
+			self.parse_separating_element(p);
+			true
+		} else {
+			false
+		}
+	}
```
MichaReiser, line 122: Do we need all three +  `is_at_separating_element`? `p.expect_required` eats the separator element. and also returns `true`/`false`. I think having just `expect_separator` should be enough that implementations of this trait must provide (they may decide to have some more complicated logic, for example, to re-map the kind.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,156 @@
...
+				break;
+			}
+
+			let parsed_element = self.parse_element(p);
+
+			if self.recover(p, parsed_element).is_err() {
+				break;
+			}
+		}
+		self.finish_list(p, elements).unwrap();
```
MichaReiser, line 154: `finish_list` shouldn't return `ParsedSyntax` if you call `unwrap` on it anyway. Make it so that it returns `CompletedMarker` or `InvalidSyntax` instead

crates/rslint_parser/src/parser/parse_lists.rs
MichaReiser: OK, so how would this work. Does `recover` set an internal boolean flag to `invalid` if a syntax turned out to be invalid and then returns `Invalid`/`Valid` depending on that flag inside of `finish_list`? 

I'm still not sure if we should have `start/finish` list because all lists ever only are of the `kind` `LIST` and it makes it difficult to implement the optimisation to insert a `missing` slot for lists that contain no elements. 

There's further the point that `parse_list` returns `()`, so the caller has no access to the return value of `finish_list`. One could then argue that `parse_list` should return a `ParsedSyntax` but that enforces all callers to handle the `ParsedSyntax` which isn't necessarily for lists (as long as they are not `ConditionalSyntax`. 

I would go with the following:
* Change the return type of `parse_list` to `Option<TList>`
* Replace the `start_list`/`finish_list` with a `finish_list(Option<CompletedMarker>) -> Option<T>` that can decide if it wants to wrap the `CompletedMarker`. 

Now this approach has the downside that the compiler won't create a warning if `finish_list` returns `Option<ConditionalSyntax>`. 

How many use cases do we have where the parser returns a `ConditionalSyntax of List`? Is it worth adding all the complexity for this edge case?

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -0,0 +1,156 @@
...
+	/// Usually here you want to check the current token.
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool;
+
+	/// This method is used to recover the parser in case [Self::parse_element] returns [ParsedSyntax::Absent]
+	fn recover(
+		&mut self,
+		p: &mut Parser,
+		parsed_element: ParsedSyntax<Self::ParsedElement>,
+	) -> RecoveryResult;
+}
```
MichaReiser, line 32: Should we remove this trait? It doesn't provide any useful methods on its own.

MichaReiser (commented):
I realised that I had a misunderstanding of how `traits` work in rust and am now wondering if `ParseNormaleList` and `ParseSeparatedList` should be structs instead... 

```rust
trait ParseList {
  type TList;
  type TItem;

  fn parse_list(&mut self, p: &mut Parser) -> Option<self::TList> {
    // default implementation
  }

  fn parse_item(&mut self, p: &mut Parser) -> self::TItem;

  fn recover(&mut self, p: &mut Parser, item: self::TItem) -> RecoveryResult<self::TItem>;
}

struct NormalList<TItem, Parse> {
  parse_fn: Parse
};

impl<Parse> NormalList<CompletedMarker, P> where P: Fn(&mut Parser) -> ParsedSyntax<CompletedMarker> { ... }

impl<Parse> NormalList<ConditionalSyntax, P> where P: Fn(&mut Parser) -> ParsedSyntax<ConditionalSyntax> { 
}

fn parse_list<P>(p: &mut Parser, parse_item: P, recovery: ParseRecovery) -> Option<CompletedMarker> where P: Fn(&mut Parser) -> Option<CompletedMarker> {
   NormalList { parse: parse_item }.parse_list(p)
 }
}

#[must_use])
fn parse_conditional_list<P>(p: &mut Parser, parse_item: P, recovery: ParseRecovery) -> Option<ConditionalSyntax> where P: Fn(&mut Parser) -> Option<ConditionalSyntax> {
   NormalList { parse: parse_item }.parse_list(p)
 }
}
```

The structs would then become implementation details of the `parse_list` functions. 

I'm not sure of what the downside of this are compared to use multiple traits.

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: It does, it provides shared behaviour among traits

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: I would suggest to keep it simple for now and make changes as we see fit - if we see fit - based on the usage in the future.

At the moment we don't have lists that have `ConditionalSyntax` (I haven't seen them). In case we will have the case, we will change the code base accordingly.

So we can return nothing - `()` - from `parse_list`. We remove `type ParsedList` and we always return a `Present(CompletedMarker)` inside `finish_list`, which will have a default implementation now.

MichaReiser (approved):
Amazing! It will allow me to delete so much code from the module implementation :)

Issue #1819, opened 2021-11-25
📎  [WIP] AST updates


jamiebuilds:
### Description

Reviewing the new Ungrammar AST carefully there are a number of changes I think should be made for these reasons (in this order):

- Accuracy: Where possible we should represent _all_ valid programs and _only_ valid programs.
- Clarity: When you see a node in isolation, you should have a clear picture of what syntax it represents and where it is valid.
- Predictability: When you're building the AST or just interacting with the API, you should be able to guess what the API is for the task at hand.
- Consistency: This is more an element of predictability, but it's worth calling out specifically, consistency creates habits and things become second nature.

There are some places where being accurate isn't always _reasonable_:

- When it would create lots of overly-specific node types that would burden developers having to translate between two very similar nodes just to move syntax around.
- When it very closely ties the AST to the _current state_ of the spec and would likely lead to breaking changes. It's tempting to just copy how the spec defines things (although that's not always possible), sometimes it's worth giving a little wiggle room in order to be defensive.

When it comes to naming there's a couple high-level _guidances_ (not strict rules) that I've come up with:

- When you introduce a term that represents a category, define its purpose clearly, and make sure all the nodes that fall in that category are using it in the same way. (i.e. "declaration" is currently being used in several different syntactic contexts).
- Prefer to name things after their relationship to one another in syntax, rather than their relationship to another node's "type" (although if you do this everywhere they'll start to be the same). This helps keep things consistent, but also is defensive against changing syntax. 
- Never use abbreviations, even where it seems obvious. (i.e. `Parameter` not `Param`)
- Things that accept many items should be pluralized words. (i.e. `items` not `body`)

There are a couple "category" words that I've defined. In all cases, any node that falls into these categories should be swappable with any other node that falls into the same category. They are:

- **Expression:** An expression produces a value and can be written wherever a value is expected.
- **Statement:** A statement represents a discrete step in evaluation, it does not produce a value and therefore cannot be used in any position a value is expected.
- **Type:** A type represents a constraint
- **Binding:** A binding represents a name that is created in a scope and can be referenced later on.
- **Assignment:** An assignment represents an operation that mutates a value either in scope or within another value.
- **{Binding,Assignment}Pattern:** A pattern represents the destructuring of a value to create a binding or assignment.
- **Name:** [WIP]
- **{Object,Class}Member:** [WIP: A named item of an object]
- **Element:** [WIP: An item of an array]
- **Parameter:** [WIP: An item of signature]
[...]

PR #1854, opened 2021-12-06
Binding, Patterns, and Identifier naming

MichaReiser:
## Summary

Part of the naming improvements #1819 suggested by @jamiebuilds and a preparation to implement `import`/`export`

* Refines naming of `Binding`s:
  * `JsAnyBinding` is only a `JsIdentifierBinding` or a `JsUnknownBinding`. 
  * Change `id`s to use `JsAnyBinding` instead of `JsUnknownBinding` to allow local error-recovery
  * Rename object and array bindings to `JsArrayBindingPattern` and `JsObjectBindingPattern`
* Refines naming of assignment targets
  * `JsAnyAssignment` now is a union over `StaticMemberAssignment`, `ComputedMemberAssignment`, `UnknownAssignment`, and `IdentifierAssignment`
  * Rename the array and object patterns to `JsArrayAssignmentPattern` and `JsObjectAssignmentPattern`
* Introduce `JsReferenceIdentifier` for identifiers referencing a variable from the current scope. Main reason is to avoid using `IdentifierExpression` in `ShorthandProperties` and `export { name, name2 }` because these names aren't expressions
* Replace `MemberReference` with `JsName`. `JsName` will be used anywhere where an identifier is neither a binding nor a reference (member expression, `import {x as name}`, ...).


crates/rslint_parser/src/syntax/assignment.rs
```diff
@@ -0,0 +1,407 @@
...
+}
+
+fn try_expression_to_assignment(
+	p: &mut Parser,
+	target: CompletedMarker,
+	checkpoint: Checkpoint,
+) -> Result<CompletedMarker, ()> {
+	if !matches!(
+		target.kind(),
+		JS_PARENTHESIZED_EXPRESSION
```
MichaReiser, line 275: Hmm, unclear why git doesn't recognise this as the same file as `assignment_target`. I mainly changed this method. I only did renames in the other methods.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -222,50 +221,39 @@ fn parse_expression_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
-					kind: T![ident],
-					range,
-				} => Some(range),
-				_ => None,
-			})
-			.expect(
-				"Tried to get the ident of a name node, but there was no ident. This is erroneous",
-			);
+	// Labelled statement
+	if p.nth_at(1, T![:]) {
```
MichaReiser, line 225: No need to first parse an expression and then test if this is a labelled statement. We can simply look if the next token is the colon and if so, treat it as a labelled statement.

xunilrj (approved):

crates/rslint_parser/src/parser.rs
```diff
@@ -383,7 +388,9 @@ impl<'t> Parser<'t> {
 			errors_pos,
 		} = checkpoint;
 		self.tokens.rewind(token_pos);
-		self.drain_events(self.cur_event_pos() - event_pos);
+		if self.cur_event_pos() > event_pos {
+			self.drain_events(self.cur_event_pos() - event_pos);
+		}
```
ematipico, line 393: What are we fixing here?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1089,14 +1099,18 @@ pub(crate) fn parse_identifier(
...
-				"yield" | "let" if StrictMode.is_supported(p) => Some(
-					p.err_builder(&format!(
-						"Illegal use of `{}` as an identifier in strict mode",
-						name
-					))
-					.primary(p.cur_tok().range, ""),
-				),
-
+				"yield" | "let" | "public" | "protected" | "private" | "package" | "implements"
+				| "interface" | "static"
```
ematipico, line 1103: `"protected"`, for example, will emit a false negative. Meaning that the usage of that keyword is illegal only in a TypeScript context but JavaScript can use it and it should emit no errors.

We have to revisit the logic and reduce the check only to JS keywords if TS is not enabled.

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -252,8 +244,23 @@ pub(crate) fn parse_object_member_name(p: &mut Parser) -> ParsedSyntax<Completed
 	}
 }
 
+fn is_nth_at_object_member_name(offset: usize, p: &Parser) -> bool {
```
ematipico, line 247:
```rust
fn is_nth_at_object_member_name(p: &Parser, offset: usize) -> bool {
```

nit: I would swap the arguments; most of the times we put the parser in the first place

ematipico (changes requested):

crates/rslint_parser/src/parser.rs
MichaReiser: Let's revert this change. I thought it's needed to use `split_off` in the `rewrite_events` but that's not the case (I must have had another bug in the implementation).

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: `protected` is not a valid JS identifier in strict mode

>  It is a Syntax Error if this phrase is contained in strict mode code and the StringValue of IdentifierName is: "implements", "interface", "let", "package", "private", "protected", "public", "static", or "yield". 

Source: https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-identifiers-static-semantics-early-errors

You can also see this from the number of fixed tests

```
xtask/src/coverage/test262/test\language\future-reserved-words\implements-strict.js
xtask/src/coverage/test262/test\language\future-reserved-words\interface-strict.js
xtask/src/coverage/test262/test\language\future-reserved-words\package-strict.js
xtask/src/coverage/test262/test\language\future-reserved-words\private-strict.js
xtask/src/coverage/test262/test\language\future-reserved-words\protected-strict.js
xtask/src/coverage/test262/test\language\future-reserved-words\public-strict.js
xtask/src/coverage/test262/test\language\future-reserved-words\static-strict.js
```

> We have to revisit the logic and reduce the check only to JS keywords if TS is not enabled.

It should be fine even in TS context. We only need to make sure to first eat all accessibility modifiers before calling `parse_identifier` (if we then expect any accessibility modifiers).

ematipico: OK this is new to me 😱

ematipico (approved):

PR #1855, opened 2021-12-07
Split `JsRoot` into `JsModule` and `JsScript`

MichaReiser:
## Summary

#1819 proposes that we split `JsRoot` into `JsScript` and `JsModule` so that we can enforce that module syntax is only permitted inside `Module`s. 

This PR promotes `JsModule` and `JsScript` to "roots" and defines a `JsAnyRoot`. It gives us the same flexibility and has the advantage that the `parse_module` method can be typed as `Parse<JsModule>` instead of `Parse<JsRoot>` 

This approach also avoids an additional union type compared to:

```
JsRoot = program: JsAnyProgram

JsAnyProgram = JsModule | JsScript
```

## Test Plan

Verified rast changes


xtask/js.ungram
```diff
@@ -48,11 +48,20 @@ JsUnknownModifier = SyntaxElement*
 
 Ident = 'ident'
 
-JsRoot =
+JsAnyRoot =
+	JsScript | JsModule
+
+JsScript =
```
MichaReiser, line 54: @jamiebuilds I promoted `JsScript` and `JsModule` to roots rather than having

```
JsRoot = program: JsAnyProgram
JsAnyProgram = JsScript | JsModule
```

`JsRoot` could contain the "shared" `interpreter` and `directives` but I think that it makes sense to have these insides of the "program". 

This further allows to type `parse_module` to return `Parse<JsModule>` so that callers don't need to cast the root's program to a module or script.

ematipico (approved):

PR #1856, opened 2021-12-07
Parse Import module item

MichaReiser:
## Summary

Implements the parsing of `import` module items as specified in #1819.


xtask/js.ungram
```diff
@@ -43,16 +43,26 @@ JsUnknownStatement = SyntaxElement*
 JsUnknownExpression = SyntaxElement*
 JsUnknownMember = SyntaxElement*
 JsUnknownBinding = SyntaxElement*
-JsUnknownAssignmentTarget = SyntaxElement*
+JsUnknownAssignment = SyntaxElement*
```
MichaReiser, line 46: @jamiebuilds any feedback on the grammar changes

xtask/js.ungram
```diff
@@ -44,7 +44,9 @@ JsUnknownExpression = SyntaxElement*
 JsUnknownMember = SyntaxElement*
 JsUnknownBinding = SyntaxElement*
 JsUnknownAssignment = SyntaxElement*
-JsUnknownModifier = SyntaxElement*
+// Generic Unknown node used in other places where it's none of the above.
+// Useful for less common cases
+JsUnknown = SyntaxElement*
```
MichaReiser, line 49: I introduced the generic `JsUnknown` because it feels like having an `Unknown` node for every case where we have a list feels a bit overkill but not having an unknown node otherwise results in invalid syntax bubbling up quite far up the tree (for example, an invalid specifier inside of a named import invalidates the specifier list, the named import clause and the import statement).

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+	if !p.at(T!['{']) {
+		return Absent;
+	}
+
+	let m = p.start();
+
+	parse_default_import_specifier(p).or_missing(p);
+	parse_named_import(p).or_missing_with_error(p, expected_named_import);
+	expect_keyword(p, "from", T![from]);
+	parse_module_source(p).or_missing_with_error(p, expected_module_source);
```
ematipico, line 196: Should we add an error test case for `import "a" from 4`?

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+			first = false;
+		} else {
+			p.expect_required(T![,]);
+			if p.at(T!['}']) {
+				break;
+			}
+		}
+
+		let recovered = parse_import_assertion_entry(p, &mut assertion_keys)
+			.or_invalid_to_unknown(p, JS_UNKNOWN)
```
ematipico, line 380: Why not `JS_UNKNOWN_EXPRESSION`?

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+// test_err import_assertion_err
+// import "foo" assert { type, "json" };
+// import "foo" \u{61}ssert { type: "json" };
+// import { foo } assert { type: "json" };
+// import "foo"
+// assert { type: "json" }
+// import foo from "foo.json" assert { "type": "json", type: "html", "type": "js" };
+// import "x" assert;
+// import foo from "foo.json" assert { type: "json", lazy: true, startAtLine: 1 };
+fn parse_import_assertion(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
```
ematipico, line 352: I think the whole assertion syntax should be `ParsedSyntax<ConditionalSyntax>` because it's still a language feature that is not supported by all runtimes/browsers.

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+		JS_STRING_LITERAL | T![ident] => {
+			p.bump_any();
+		}
+		t if t.is_keyword() => {
+			p.bump_remap(T![ident]);
+		}
+		T![:] => {
+			p.missing();
+			p.error(
+				expected_any(&["identifier", "string literal"], p.cur_tok().range).to_diagnostic(p),
```
ematipico, line 433: We can move this error inside `js_parse_errors.rs`

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+	if !p.at(JS_STRING_LITERAL) {
+		Absent
+	} else {
+		let m = p.start();
+		p.bump_any();
+		Present(m.complete(p, JS_MODULE_SOURCE))
+	}
+}
+
+fn expect_keyword(p: &mut Parser, keyword_name: &str, kind: SyntaxKind) {
```
ematipico, line 492: I think this function can be moved inside `util.rs`, it seems a function that other parts of the parser can benefit from

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+	expected_any(&["namespace import", "named imports"], range).to_diagnostic(p)
+}
+
+fn expected_export_name(p: &Parser, range: Range<usize>) -> Diagnostic {
+	expected_any(&["string literal", "identifier"], range).to_diagnostic(p)
+}
+
+fn expected_named_import_specifier(p: &Parser, range: Range<usize>) -> Diagnostic {
+	expected_node("identifier", range).to_diagnostic(p)
+}
```
ematipico, line 529: Should we move these little fellas inside `js_parse_errors.rs`

crates/rslint_parser/test_data/inline/err/import_assertion_err.rast
```diff
@@ -0,0 +1,541 @@
...
+
+--
+error[SyntaxError]: Duplicate statement labels are not allowed
+  ┌─ import_assertion_err.js:5:10
+  │
+2 │ import "foo" \u{61}ssert { type: "json" };
+  │                            ---- `type` is first used as a label here
+  ·
+5 │ assert { type: "json" }
+  │          ^^^^ a second use of `type` here is not allowed
```
ematipico, line 485: I think this error is not correct... or is it? 🤔

crates/rslint_parser/test_data/inline/err/import_assertion_err.rast
```diff
@@ -0,0 +1,541 @@
...
+  ┌─ import_assertion_err.js:7:18
+  │
+7 │ import "x" assert;
+  │                  ^ unexpected
+
+--
+error[SyntaxError]: expected `JS_STRING_LITERAL` but instead found `true`
+  ┌─ import_assertion_err.js:8:57
+  │
+8 │ import foo from "foo.json" assert { type: "json", lazy: true, startAtLine: 1 };
```
ematipico, line 516: This error is not correct, we are showing an internal `SyntaxKind`

crates/rslint_parser/test_data/inline/err/import_assertion_err.rast
```diff
@@ -0,0 +1,541 @@
...
+  │                                                             ^ unexpected
+
+--
+error[SyntaxError]: expected `JS_STRING_LITERAL` but instead found `1`
+  ┌─ import_assertion_err.js:8:76
+  │
+8 │ import foo from "foo.json" assert { type: "json", lazy: true, startAtLine: 1 };
+  │                                                                            ^ unexpected
```
ematipico, line 531: We're printing an internal `SyntaxKind`

crates/rslint_parser/test_data/inline/err/import_err.rast
```diff
@@ -0,0 +1,466 @@
...
+  │          -  ^ a second declaration of `a` is not allowed
+  │          │   
+  │          `a` is first declared here
+
+--
+error[SyntaxError]: `default` imports must be aliased
+  ┌─ import_err.js:6:10
+  │
+6 │ import { default } from "c";
+  │          ^^^^^^^- add `as identifier` here
```
ematipico, line 440: The backtick \` should finish after `as`

ematipico (changes requested):

xtask/js.ungram
```diff
@@ -879,16 +881,132 @@ JsAnyModuleItem =
...
+    | JsImportDefaultClause
+    | JsImportNamespaceClause
+//    | TsImportRequireClause
+//    | TsImportAliasClause
+
+// import "abcd"
+// import "abcd" assert ...
+JsImportBareClause =
+	source: JsModuleSource
+	assertion: JsImportAssertion?
```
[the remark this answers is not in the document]
MichaReiser: The semicolon is part of the `JsImport` rule

crates/rslint_parser/src/syntax/module.rs
MichaReiser: Because `expressions` arent' allowed in an assertion entry. That's why I added the `JS_UNKNOWN` so that I don't need to add a `JsUnknownAssertionEntry` and `JsUnknownNamedImportSpecifier`

crates/rslint_parser/src/syntax/module.rs
MichaReiser: Let's leave it as is until we revisit our syntax features. There's currently no way to enable / disable features in inline tests.

crates/rslint_parser/test_data/inline/err/import_assertion_err.rast
MichaReiser: This is actually correct, I mean, as correct as error recovery can be. We could probably still do better by adding custom recovery from `\u{61}assert` or when the assert is on the next line. 

The reason why we get this error is that the assert expressions for these two statements are parsed as block statements, each containing a labelled statement. 

```
import "foo" \u{61}ssert { type: "json" };
assert { type: "json" }
```

crates/rslint_parser/src/syntax/module.rs
MichaReiser: I so far added errors to `js_parse_errors` if they appeared in different places throughout the program, potentially in different rust files but kept other errors local in the same rust file (if used multiple times) or inline if used only once.

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -0,0 +1,529 @@
...
+
+	let end = p.cur_tok().range.start;
+
+	semi(p, start..end);
+
+	Present(import.complete(p, JS_IMPORT))
+}
+
+// test import_default_clause
+// import foo from "test";
```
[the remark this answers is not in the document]
MichaReiser: I have some generic test cases on `parse_import` where I added this example too. The parser recovers but not very nice. The problem is that we can't stuff the number literal 4 somewhere into the import statement. This will only be possible if we introduce something like a `GarbageTrivia`

MichaReiser:
> Would you mind revert this change, please? It doesn't seem this change is related to this PR and it also goes in conflict with what I've done in #1859

It is related to this change because I unified all these one-off `Unknown` nodes into one generic `Unkown` . This diff otherwise needs to introduce a `JsUnknownNamedImportSpecifier` and `JsUnknownImportAssertionEntry` and I believe that a generic `Unknown` is as good in these places as having separate ones (you can still see in the grammar where they might appear). 

But I'm open to go back to specific unknown nodes but would love to understand your reasoning behind it.

crates/rslint_parser/src/syntax/module.rs
ematipico: In old RomeJS we had all diagnostics stored in one single place and I kind liked it.

xtask/js.ungram
MichaReiser: @jamiebuilds is off. Let's move on with what we have and revisit if needed

ematipico:
> > Would you mind revert this change, please? It doesn't seem this change is related to this PR and it also goes in conflict with what I've done in #1859
> 
> It is related to this change because I unified all these one-off `Unknown` nodes into one generic `Unkown` . This diff otherwise needs to introduce a `JsUnknownNamedImportSpecifier` and `JsUnknownImportAssertionEntry` and I believe that a generic `Unknown` is as good in these places as having separate ones (you can still see in the grammar where they might appear).
> 
> But I'm open to go back to specific unknown nodes but would love to understand your reasoning behind it.

I think having specialized unknown nodes help:

- to understand what kind grammar was supposed to be there (expression, statement, binding, etc.)
- to understand what grammar we are parsing while we write code
- analyzers might, in the future, to act differently based on the type of unknown nodes (visit children VS do nothing)


I think having a broad unknown node could increase uncertainty of when using using a generic unknown node or a specialized one. For example you used it here in this PR but suggested its use in another one. Where's bar? When should we use it? I really liked specialized unknown nodes.

MichaReiser:
> > > Would you mind revert this change, please? It doesn't seem this change is related to this PR and it also goes in conflict with what I've done in #1859
> > 
> > 
> > It is related to this change because I unified all these one-off `Unknown` nodes into one generic `Unkown` . This diff otherwise needs to introduce a `JsUnknownNamedImportSpecifier` and `JsUnknownImportAssertionEntry` and I believe that a generic `Unknown` is as good in these places as having separate ones (you can still see in the grammar where they might appear).
> > But I'm open to go back to specific unknown nodes but would love to understand your reasoning behind it.
> 
> I think having specialized unknown nodes help:
> 
>     * to understand what kind grammar was supposed to be there (expression, statement, binding, etc.)
> 
>     * to understand what grammar we are parsing while we write code
> 
>     * analyzers might, in the future, to act differently based on the type of unknown nodes (visit children VS do nothing)
> 
> 
> I think having a broad unknown node could increase uncertainty of when using using a generic unknown node or a specialized one. For example you used it here in this PR but suggested its use in another one. Where's bar? When should we use one or the other? I really liked specialized unknown nodes.

I reverted the change. Mainly because it shouldn't be part of this PR. We can decide to introduce a generic `Unknown` if we see that things go over board and introduce it in a dedicated PR

ematipico (approved):

PR #1858, opened 2021-12-08
test(formatter): fail js tests with parser errors

Boshen:
## Summary
Fail js tests with parser errors.

When implementing the formatter todos, I encountered numerous parsing errors that produced undefined behavior, erroring out with errors should aid implementing the formatter. 

## Test Plan
```
cargo test -p rome_formatter js
```

Result
```
failures:

---- formatter::js::block_stmt_err stdout ----
thread 'formatter::js::block_stmt_err' panicked at 'without error: ParserError([Diagnostic { file_id: 0, severity: Error, code: Some("SyntaxError"), title: "Expected a statement or declaration, but found none", tag: None, primary: Some(SubDiagnostic { severity: Error, msg: "Expected a statement or declaration here", span: FileSpan { file: 0, range: 65..66 } }), children: [], suggestions: [], footers: [] }, Diagnostic { file_id: 0, severity: Error, code: Some("SyntaxError"), title: "expected a statement but instead found 'let recovered     = \"no\"'", tag: None, primary: Some(SubDiagnostic { severity: Error, msg: "Expected a statement here", span: FileSpan { file: 0, range: 68..92 } }), children: [], suggestions: [], footers: [] }, Diagnostic { file_id: 0, severity: Error, code: Some("SyntaxError"), title: "expected `'}'` but instead the file ends", tag: None, primary: Some(SubDiagnostic { severity: Error, msg: "the file ends here", span: FileSpan { file: 0, range: 93..93 } }), children: [], suggestions: [], footers: [] }])', crates/rome_formatter/tests/spec_test.rs:31:26

---- formatter::js::if_stmt_err stdout ----
thread 'formatter::js::if_stmt_err' panicked at 'without error: ParserError([Diagnostic { file_id: 0, severity: Error, code: Some("SyntaxError"), title: "expected `'('` but instead found `test`", tag: None, primary: Some(SubDiagnostic { severity: Error, msg: "unexpected", span: FileSpan { file: 0, range: 36..40 } }), children: [], suggestions: [], footers: [] }, Diagnostic { file_id: 0, severity: Error, code: Some("SyntaxError"), title: "expected `')'` but instead found `{`", tag: None, primary: Some(SubDiagnostic { severity: Error, msg: "unexpected", span: FileSpan { file: 0, range: 41..42 } }), children: [], suggestions: [], footers: [] }, Diagnostic { file_id: 0, severity: Error, code: Some("SyntaxError"), title: "expected `'}'` but instead the file ends", tag: None, primary: Some(SubDiagnostic { severity: Error, msg: "the file ends here", span: FileSpan { file: 0, range: 168..168 } }), children: [], suggestions: [], footers: [] }])', crates/rome_formatter/tests/spec_test.rs:31:26


failures:
    formatter::js::block_stmt_err
    formatter::js::if_stmt_err
```

**Note**: These two tests are expected to produce errors, I don't know what should be the correct way to handle these.

### DIFF
--- a/crates/rome_formatter/src/lib.rs
+++ b/crates/rome_formatter/src/lib.rs
@@ -71,7 +71,7 @@ pub use printer::PrinterOptions;
 use rome_core::file_handlers::Language;
 use rome_core::App;
 use rome_path::RomePath;
-use rslint_parser::parse_text;
+use rslint_parser::{parse_text, ParserError};
 
 use std::io::Read;
 use std::str::FromStr;
@@ -95,6 +95,9 @@ pub enum FormatError {
 
 	/// When the ability to format the current file has been turned off on purpose
 	CapabilityDisabled,
+
+	/// ParserErrors from rslint_parser
+	ParserError(Vec<ParserError>),
 }
 
 impl From<SyntaxError> for FormatError {
@@ -176,7 +179,6 @@ impl Formatted {
 	}
 }
 
-// TODO: implement me + handle errors
 /// Main function
 pub fn format(rome_path: &mut RomePath, options: FormatOptions) -> FormatResult<Formatted> {
 	// we assume that file exists
@@ -191,10 +193,15 @@ pub fn format(rome_path: &mut RomePath, options: FormatOptions) -> FormatResult<
 			let result = match handler.language() {
 				Language::Js => {
 					let parsed_result = parse_text(buffer.as_str(), 0);
-					Formatter::new(options).format_root(&parsed_result.syntax())
+					if parsed_result.errors().is_empty() {
+						Formatter::new(options).format_root(&parsed_result.syntax())
+					} else {
+						Err(FormatError::ParserError(parsed_result.errors().to_vec()))
+					}
 				}
 				Language::Json => {
 					let element = tokenize_json(buffer.as_str());
+					// TODO: add error handling
 					Ok(format_element(&element, options))
 				}
 				Language::Ts | Language::Unknown => Err(FormatError::UnsupportedLanguage),
--- a/crates/rome_formatter/tests/spec_test.rs
+++ b/crates/rome_formatter/tests/spec_test.rs
@@ -1,5 +1,6 @@
 use rome_core::create_app;
-use rome_formatter::{format_file, FormatOptions};
+use rome_formatter::{format, FormatOptions};
+use rome_path::RomePath;
 use std::fs;
 use std::path::Path;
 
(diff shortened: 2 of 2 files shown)


MichaReiser:
Thanks for looking into this.

We actually do want to format files that contain errors. We don't know how well it will work and what fine-tuning is necessary to do so well but we shouldn't just bail out (especially because `errors` is more `diagnostics` and may only contain warnings). The way the formatter is currently handling this is that the formatter skips over statements that have missing mandatory children. I don't think we've yet decided what to do about unknown * nodes but we may want to do something similar there.

Boshen:
> Thanks for looking into this.
> 
> We actually do want to format files that contain errors. We don't know how well it will work and what fine-tuning is necessary to do so well but we shouldn't just bail out (especially because `errors` is more `diagnostics` and may only contain warnings). The way the formatter is currently handling this is that the formatter skips over statements that have missing mandatory children. I don't think we've yet decided what to do about unknown * nodes but we may want to do something similar there.

Since the main goal of this is to get parser errors from valid syntaxes, I'll try and setup two test suites to differentiate valid / invalid parsing. Probably after the snapshot PR is merged.

MichaReiser:
> Since the main goal of this is to get parser errors from valid syntaxes, I'll try and setup two test suites to differentiate valid / invalid parsing

Valid syntax should never have parse errors and these would be parser bugs. We shouldn't have to change how `format` works to accommodate the test needs. 

What we could do (but unfortunately, isn't straightforward) is to have a `format_node(node: &SyntaxNode, options: FormaterOptions)` function that operates on an already parsed node. The spec tests can then call directly into `parse_text` or `parse_module`, assert that there are no parsing errors for the "ok" tests and then call into `format_node`. 

I would be OK if we add such a method that currently simply assumes that the node is `JavaScript` since we have no way to get the `RomePath` or handler from a `SyntaxNode`.

MichaReiser:
> I'll keep the changes locally for now. I needed this because the parser is incomplete at the moment, so I actually got some parser errors inside formatting, and it returned the wrong output :-/
> 
> I'll reopen this when something better comes up.

Would you mind reporting issues where the parser incorrectly creates syntax errors for correct syntax so that we can take a look at it?

Boshen:
> > I'll keep the changes locally for now. I needed this because the parser is incomplete at the moment, so I actually got some parser errors inside formatting, and it returned the wrong output :-/
> > I'll reopen this when something better comes up.
> 
> Would you mind reporting issues where the parser incorrectly creates syntax errors for correct syntax so that we can take a look at it?

Of course, I already made one with `>>` ;-) Probably more when I start implementing more `todos` in the formatter.

PR #1859, opened 2021-12-08
feat: fix parse class members/properties with new API

ematipico:
## Summary

This PR fixes all the issues that we were having inside error handling inside classes.

The majority of errors were around TypeScript syntax, although the handling wasn't correct and that was resulting in tokens that were skipped and green tree incorrect.

In order to fix all these errors, I had to change a bit the current grammar, here's the list of changes:

- now all class members and properties, inside the grammar, will have a `JsAnyModifier` that will include `static`, `declare` and `abstract`. This will allow us to invalidate only those words without invalidating the whole member/property. This change of grammar includes a `JsUnknownModifier` node. The fix node includes a `missing` token when the modifier is not there;
- `extends` and `implements` clauses are now correctly marked as missing in case they are not there; in case of errors we mark them as `JS_UNKNOWN_EXPRESSION`, this is because we don't have specific nodes. Although this results in a decent solution because the parser invalidates only the portion `extends A, B, C`, leaving the rest of the class in a correct state; 
- the operators `?` and `!` are now correctly marked as `missing` when they are not there;
- the access modifiers `readonly` and `private` are now wrapped inside a node called `TsAnyAccessibility`, which follows the same principle of `JsAnyModifier`. This will allow us to invalidate only those words. Because of the fact that these modifiers belong to TS only, errors are wrapped into an unknown node called `TsUnknownAccessibilityModifier`;
- update the formatter implementation;

## Test Plan

Current CI


xtask/src/codegen/kinds_src.rs
```diff
@@ -383,13 +383,15 @@ pub const KINDS_SRC: KindsSrc = KindsSrc {
...
-		// unknown nodes
+		// unknown nodes JS
 		"JS_UNKNOWN_EXPRESSION",
 		"JS_UNKNOWN_STATEMENT",
 		"JS_UNKNOWN_MEMBER",
 		"JS_UNKNOWN_BINDING",
 		"JS_UNKNOWN_ASSIGNMENT",
 		"JS_UNKNOWN_MODIFIER",
+		// unknown nodes TS
+		"TS_UNKNOWN_ACCESSIBILITY_MODIFIER",
```
MichaReiser, line 394: I introduce da `JS_UNKOWN` in #1856 so that we don't need an unknown for every single list

xtask/js.ungram
```diff
@@ -610,13 +606,11 @@ JsSetterClassMember =
 
 JsEmptyClassMember = ';'
 
-// TODO: apply this node to all class members and properties
 JsAnyModifier =
 	JsModifier
 	| JsUnknownModifier
```
MichaReiser, line 611: What if we have `static declare abstract a() {}`? This won't be a valid `modifier` because the modifiers are out of order, meaning, we would still need to wrap the whole member as an "UnknownMember".

xtask/js.ungram
```diff
@@ -610,13 +606,11 @@ JsSetterClassMember =
...
 
-// TODO: move here also the rest of the modifiers
-JsModifier = 'declare'?
+JsModifier = 'declare'? 'static'? 'abstract'?
```
MichaReiser, line 613: Should we have different modifiers for each kind to avoid an empty modifier?

MichaReiser (commented):
I only reviewed the grammar changes and will need to spend some time thinking about what we're trying to solve here. 

There are still cases where this change won't be sufficient to correctly recover from incorrectly ordered modifiers and the API allows for code fixes to build up invalid ASTs without any error.

xtask/js.ungram
```diff
@@ -586,9 +584,8 @@ JsMethodClassMember =
 // class Test { get a() {} }
 //              ^^^^^^^^^^
 JsGetterClassMember =
-	access_modifier: TsAccessibility?
-	'abstract'?
-	'static'?
+	modifiers: JsAnyModifier?
+	access_modifier: TsAnyAccessibility?
```
MichaReiser, line 588: Should we merge these? Like, making `access_modifier` part of the `modifiers`?

xtask/js.ungram
```diff
@@ -1260,10 +1254,13 @@ TsImport = 'import' type_args:TsTypeArgs '.'? '(' qualifier:TsEntityName ')'
 TsIndexSignature = 'readonly'? '[' pat:JsAnyBinding ':' ty:TsType ']'
 
 // typescript type of access
-TsAccessibility = 'private' | 'readonly'
+TsAnyAccessibility =
+	TsUnknownAccessibilityModifier
+	| TsAccessibility
 
+TsAccessibility = 'private' | 'readonly'
```
MichaReiser, line 1261: `private` and `readonly` can be used together, which this grammar wouldn't support?

```
class Test {
  private readonly test;
}
```

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -116,9 +115,25 @@ fn parse_class(p: &mut Parser, kind: ClassKind) -> ParsedSyntax<CompletedMarker>
...
-	// TODO: these two functions should return `ParsedSyntax`, so we can handle possible errors/missing/etc.
-	extends_clause(&mut guard);
-	implements_clause(&mut guard);
+	match extends_clause(&mut guard) {
+		Absent => {
+			guard.missing();
+		}
+		Present(Invalid(syntax)) => {
+			syntax.or_to_unknown(&mut guard, JS_UNKNOWN_EXPRESSION);
+		}
```
MichaReiser, line 124: Will this not invalidate the AST because it generates

```
- class
  - identifier: A
  - UNKNOWN EXPRESSION
  - {
```

Same goes for implements.

You can use `extends_clause(&mut guard).or_invalid_to_unknown(guard, JS_UNKNOWN_EXPRESSION).or_missing(guard)`

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -688,34 +714,44 @@ fn property_declaration_class_member_body(
...
 /// Parses the body of a property class member (anything after the member name)
 fn parse_property_class_member_body(p: &mut Parser, member_marker: Marker) -> CompletedMarker {
-	let optional_range = optional_member_token(p);
-	if p.at(T![!]) {
-		let range = p.cur_tok().range;
+	let parsed_syntax = optional_member_token(p);
+	match parsed_syntax {
+		Absent => p.missing(),
+		Present(Invalid(syntax)) => {
+			syntax.or_to_unknown(p, JS_UNKNOWN_EXPRESSION);
```
MichaReiser, line 721: Same here. The parent of the `proprty_class_member_body` is a member and a member doesn't allow expressions as it's direct children.

crates/rslint_parser/test_data/inline/err/class_member_bang.rast
```diff
@@ -0,0 +1,64 @@
...
+      3: (empty)
+      4: L_CURLY@8..10 "{" [] [Whitespace(" ")]
+      5: LIST@10..16
+        0: JS_PROPERTY_CLASS_MEMBER@10..16
+          0: (empty)
+          1: (empty)
+          2: JS_LITERAL_MEMBER_NAME@10..13
+            0: IDENT@10..13 "foo" [] []
+          3: (empty)
+          4: JS_UNKNOWN_EXPRESSION@13..14
```
MichaReiser, line 50: As you can see... this expression doesn't show up in the AST because it's not a valid member of `JS_PROPERTY_CLASS_MEMBER`

crates/rslint_parser/src/syntax/class.rs
MichaReiser: I spend some time thinking about what should happen in the case that TS syntax is used inside of an `extends` clause or if there's an `implements` clause. 

I believe the right thing to do is to mark the whole class as `invalid` because it contains invalid syntax. This isn't as bad as it may sound. Analysers can still retrieve all class members by using `root.descendants().of_type::<JsAnyClassMember>()` 

This also aligns the behavior with what we do in case a function contains any TS syntax outside of a TS file.

MichaReiser:
Regarding `declare`. I just played around with typescript and `declare` and it seems `declare` is only valid for property members. That's why I would apply the same rule as for `implements` and `extends`

* Use of `declare` for property in a `JS` file -> becomes an unknown member
* Use of `declare` for any other member -> becomes an unknown member.

MichaReiser:
Regarding the `modifiers`. I would keep explicit fields on the members and bail-out to an unknown member if the syntax contains a TS only modifier or if the modifiers are out of order.  This aligns the behaviour with what we do in other places and:

* We can restrict the modifiers to exactly the set of modifiers supported by this member
* We can enforce the ordering of the modifiers in our AST mutation API
* It makes it clear that the member is invalid. Let's say we have `private readonly m(` The `readonly` tells us that this is a `field` but the `(` suggests that this is a method. Whatever assumption isn't correct and `UnknownMember` represents this the most accurately.

xtask/src/codegen/kinds_src.rs
ematipico: Personally, I don't see the value of having a generic unknown node. It doesn't hint anything and doesn't help to understand what's the supposed type in that position.

xtask/js.ungram
ematipico: Isn't the ordering something that we could fix with fixed slots? #1721

xtask/js.ungram
ematipico: What do you mean when you say "to avoid an empty modifier"?

xtask/js.ungram
ematipico: I don't know to be honest 🤔 let me see what RomeJS does. It might make sense merge them.

xtask/js.ungram
ematipico: I don't know. It might have been a mistake. `abstract` and `readonly` can't be used together.

We will fix it when we will review the TS parsing

crates/rslint_parser/src/syntax/class.rs
ematipico: OK, let's do that

xtask/js.ungram
MichaReiser: We can use fixed slots but what if the `abstract` appears in the position of `static`? We can't just swap the slots because re-printing the parsed CST would then result in a different tree than what we originally parsed.

xtask/js.ungram
MichaReiser: `""` is a valid production of the `JsModifier` rule because all modifiers can be absent.

xtask/js.ungram
ematipico: Just checked Rome. It keeps them separate. Normal modifiers are cumulative (can be one or all of them at the same time), but the accessibility is only one (`private`, `protected` or `public`)

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -688,34 +744,44 @@ fn property_declaration_class_member_body(
 
 /// Parses the body of a property class member (anything after the member name)
 fn parse_property_class_member_body(p: &mut Parser, member_marker: Marker) -> CompletedMarker {
-	let optional_range = optional_member_token(p);
-	if p.at(T![!]) {
-		let range = p.cur_tok().range;
+	let parsed_syntax = optional_member_token(p);
```
MichaReiser, line 747: This method must return a `ConditionalSyntax` as well because it is invalid as soon as it uses any typescript syntax. 

* `abandoning` isn't valid because it then puts the token into the member, in a place where it's not expected
* Using a `JS_UNKNOWN_EXPRESSION` is not a valid solution here because a property has no direct `child` that accepts an expression.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -737,23 +803,31 @@ fn parse_property_class_member_body(p: &mut Parser, member_marker: Marker) -> Co
...
 				.primary(p.cur_tok().range, "");
 
+			is_invalid = true;
 			p.error(err);
-			// TODO: remap error to an unknown node
-			p.bump_remap(ERROR);
+		}
+		p.bump_any();
+
+		let completed_marker = Present(m.complete(p, QUESTION));
```
MichaReiser, line 822: Question is a token kind, isn't it? I don't think it's valid to create a node with that kind. The AST Facade expects a token and not a node.

MichaReiser (changes requested):
Much better. I believe there are still a few cases where we construct an invalid AST structure. 

I would further try to move the `member_is_valid` as close as possible to where it isn't guaranteed to always be true.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -233,9 +275,9 @@ fn parse_class_members(p: &mut Parser) {
...
 			//
 			// So when there's an error, we mark the whole marker as invalid
-			// and restore the state of the parser to a previous [Checkpoint].
-			Invalid(_) => {
-				p.rewind(checkpoint);
+			// and the whole member becomes an unknown node.
+			Invalid(invalid_syntax) => {
+				invalid_syntax.or_to_unknown(p, JS_UNKNOWN_MEMBER);
 				Absent
 			}
```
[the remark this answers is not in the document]
ematipico: Unfortunately no, we still need the `Absent` in order to recover. There are still cases where the parser doesn't advance.

If we remove it, the parser doesn't progress and it doesn't recover either.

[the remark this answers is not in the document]
MichaReiser: Then this is a problem with the error recovery. The contract of `ParsedSyntax` is that we return `Present` if the parser progresses, or return `Absent` otherwise. Can you investigate why recovery isn't working in that case? I mean, the parser should progress because it adds an unknown member containing all processed tokens thus far and we then continue whatever tokens follows immediately after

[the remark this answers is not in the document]
ematipico: The recovery kicks in if `ParsedSyntax` is `Absent`. That's why I had to return `Absent` from my `match`. 
https://github.com/rome/tools/blob/de73a963c897962b6e633d16d47146b58cba16cd/crates/rslint_parser/src/parser/parsed_syntax.rs#L224-L239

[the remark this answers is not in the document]
MichaReiser: Can you return `ConditionalSyntax` then? Because it means that the node is invalid and we did parse something.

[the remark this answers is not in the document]
ematipico: No, there are cases where we don't parse anything.

Take this example that is currently what's failing:

```js
class B {{}}
```

We don't parse anything and the parser gets stuck at the second `{`. I am looking how to avoid that.

crates/rslint_parser/test_data/inline/err/class_extends_err.rast
```diff
@@ -2,41 +2,47 @@ JsModule {
...
+                JsIdentifierBinding {
+                    name_token: IDENT@6..8 "A" [] [Whitespace(" ")],
+                },
+                JsExtendsClause {
+                    extends_token: EXTENDS_KW@8..16 "extends" [] [Whitespace(" ")],
+                    super_class: JsIdentifierExpression {
+                        name: JsReferenceIdentifier {
+                            value_token: IDENT@16..20 "bar" [] [Whitespace(" ")],
+                        },
                     },
```
MichaReiser, line 17: Can we convert the `JsExtendsClause` to an `Error` node (in the future `UNKNOWN)` if there's more than one super class. As you can see, the `extends foo` disappeared.

crates/rslint_parser/test_data/inline/err/class_extends_err.rast
```diff
@@ -2,41 +2,47 @@ JsModule {
...
-                super_class: JsIdentifierExpression {
-                    name: JsReferenceIdentifier {
-                        value_token: IDENT@51..54 "bar" [] [],
+        JsUnknownStatement {
+            items: [
+                CLASS_KW@34..41 "class" [Whitespace("\n")] [Whitespace(" ")],
+                JsIdentifierBinding {
+                    name_token: IDENT@41..43 "A" [] [Whitespace(" ")],
+                },
+                JsExtendsClause {
```
MichaReiser, line 32: Same here, the `,foo` is part of the `JsExtendsClause` but not accessible through the Ast facade. Can we convert it to an `Error` node instead

crates/rslint_parser/test_data/inline/err/class_member_bang.rast
```diff
@@ -0,0 +1,67 @@
...
+        JsClassDeclaration {
+            class_token: CLASS_KW@0..6 "class" [] [Whitespace(" ")],
+            id: JsIdentifierBinding {
+                name_token: IDENT@6..8 "B" [] [Whitespace(" ")],
+            },
+            implements_clause: missing (optional),
+            extends_clause: missing (optional),
+            l_curly_token: L_CURLY@8..10 "{" [] [Whitespace(" ")],
+            members: [
+                JsPropertyClassMember {
```
MichaReiser, line 14: This should become an `UnknownMember` in JS because the `excl_token` is otherwise missing.

crates/rslint_parser/test_data/inline/err/class_member_modifier.rast
```diff
@@ -0,0 +1,64 @@
...
+      2: (empty)
+      3: (empty)
+      4: L_CURLY@8..10 "{" [] [Whitespace(" ")]
+      5: LIST@10..24
+        0: JS_UNKNOWN_MEMBER@10..24
+          0: (empty)
+          1: (empty)
+          2: (empty)
+          3: TS_ACCESSIBILITY@10..19
+            0: ABSTRACT_KW@10..19 "abstract" [] [Whitespace(" ")]
```
MichaReiser, line 49: Should `abstract` really be part of `TS_ACCESSIBILITY`? Isn't it it's own field?

crates/rslint_parser/test_data/inline/err/optional_member.rast
```diff
@@ -0,0 +1,66 @@
...
+        JsClassDeclaration {
+            class_token: CLASS_KW@0..6 "class" [] [Whitespace(" ")],
+            id: JsIdentifierBinding {
+                name_token: IDENT@6..8 "B" [] [Whitespace(" ")],
+            },
+            implements_clause: missing (optional),
+            extends_clause: missing (optional),
+            l_curly_token: L_CURLY@8..10 "{" [] [Whitespace(" ")],
+            members: [
+                JsPropertyClassMember {
```
MichaReiser, line 14: This should become an `unknown` node (see other comments inside the code)

MichaReiser (changes requested):
The overall handling looks good but there are some unaddressed comments from previous reviews related to what we return from `parse_class_*` and how to handle members with TS syntax.

crates/rslint_parser/test_data/inline/err/class_member_modifier.rast
ematipico: No, this is wrong. Accessibility modifiers are: `private`, `protected` and `public`

I will fix it later in another PR. I don't want to change the grammar for now

crates/rslint_parser/test_data/inline/err/class_extends_err.rast
ematipico: Doing so would break:

```
'Tried to cast node ERROR@8..32 as `"rslint_parser::ast::generated::nodes::AnyNode"` but was unable to cast',
```

crates/rslint_parser/test_data/inline/err/class_extends_err.rast
ematipico: Doing so would break:

```
'Tried to cast node ERROR@8..32 as `"rslint_parser::ast::generated::nodes::AnyNode"` but was unable to cast',
```

crates/rslint_parser/test_data/inline/err/class_constructor_parameter_readonly.rast
```diff
@@ -0,0 +1,86 @@
...
+          4: JS_LITERAL_MEMBER_NAME@10..21
+            0: IDENT@10..21 "constructor" [] []
+          5: (empty)
+          6: JS_CONSTRUCTOR_PARAMETER_LIST@21..34
+            0: L_PAREN@21..22 "(" [] []
+            1: LIST@22..32
+              0: TS_CONSTRUCTOR_PARAM@22..32
+                0: JS_UNKNOWN_EXPRESSION@22..31
+                  0: READONLY_KW@22..31 "readonly" [] [Whitespace(" ")]
+                1: IDENT@31..32 "b" [] []
```
ematipico, line 69: Fixing this one...

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -237,11 +237,13 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
...
 		if p.at(T![class]) {
-			parse_class_declaration(&mut *p.with_state(ParserState {
+			let p = &mut *p.with_state(ParserState {
 				in_default: true,
 				..p.state.clone()
-			}))
-			.unwrap();
+			});
+			parse_class_declaration(p)
+				.or_invalid_to_unknown(p, JS_UNKNOWN_EXPRESSION)
```
MichaReiser, line 245: Can we add a todo here. An `expression` isn't valid in this position.

crates/rslint_parser/src/syntax/program.rs
```diff
@@ -261,7 +263,9 @@ pub fn export_decl(p: &mut Parser) -> CompletedMarker {
 	}
 
 	if !only_ty && p.at(T![class]) {
-		parse_class_declaration(p).unwrap();
+		parse_class_declaration(p)
+			.or_invalid_to_unknown(p, JS_UNKNOWN_EXPRESSION)
```
MichaReiser, line 267: Same here. Can we add a todo. An expression isn't valid in this position.

MichaReiser (approved):
Thanks for working on this. This turned out to be far more complicated than expected. We need to keep in mind to revisit the implementation once we have a generic `UNKNOWN` node to mark TS syntax errornous.

PR #1864, opened 2021-12-09
fix(rslint_parser): Member of computed assignment targets

### LINKED ISSUE #1863: 🐛 Parser: Broken AST lookup for `JsComputedMemberAssignment`
### Environment information

```bash
main branch
```

### What happened?

While implementing the formatter for `JsComputedMemberAssignment`,

`a[b] = c[d]` will give you `a[a] = c[d]`

#### Exploration

given our ungram:
```
// a['b'] = a[expr] = "test"
// ^^^^^^   ^^^^^^^
JsComputedMemberAssignment =
	object: JsAnyExpression
	'['
	member: JsAnyExpression
	']'
```

`object` and `member` will end up with the same node as `object`,
```rust
impl JsComputedMemberAssignment {
	pub fn object(&self) -> SyntaxResult<JsAnyExpression> { support::required_node(&self.syntax) }
	pub fn member(&self) -> SyntaxResult<JsAnyExpression> { support::required_node(&self.syntax) }
}
```

because required_node just looks for the first `JsAnyExpression` it encounters
```rust
	pub(super) fn node<N: AstNode>(parent: &SyntaxNode) -> Option<N> {
		parent.children().find_map(N::cast)
	}

	pub(super) fn required_node<N: AstNode>(parent: &SyntaxNode) -> SyntaxResult<N> {
		node(parent).ok_or_else(|| SyntaxError::MissingRequiredChild(parent.clone()))
	}
```

### Expected result

`member` gives the correct node.

_No response_

### Code of Conduct

MichaReiser:
A computed assignment target contains two expressions, one for the object and one for the member and the facade didn't manually implement the `member` field. The result (until we use fixed slots) is that both `member` and `object` return the same `JsAnyExpression`.

This went unnoticed because the AST code gen didn't add manual fields to the `Debug` implementation of a node. This diff changes the code gen to also include manual fields to surface such issues in the future.

Adding the `manual` field support to the code gen surfaced some typing mismatches between our ungrammar file and the implementations that I fixed as well.

Fixes #1863

## Test Plan

Verified the changes in the RAST files.


crates/rslint_parser/test_data/inline/ok/assignment_target.rast
```diff
@@ -176,16 +174,57 @@ JsModule {
...
+            expression: JsAssignmentExpression {
+                left: JsComputedMemberAssignment {
+                    object: JsIdentifierExpression {
+                        name: JsReferenceIdentifier {
+                            value_token: IDENT@125..127 "a" [Whitespace("\n")] [],
+                        },
+                    },
+                    l_brack_token: L_BRACK@127..128 "[" [] [],
+                    member: JsStringLiteralExpression {
+                        value_token: JS_STRING_LITERAL@128..131 "'b'" [] [],
```
MichaReiser, line 193: Now correctly returns `b`

ematipico (approved):

PR #1866, opened 2021-12-10
 feature(rslint_parser): Debug print for Unknown* node children

MichaReiser:
## Summary

The children of `Unknown` nodes are simply a sequence of `SyntaxElement`s. That's why the items currently print without field names nor validation if the CST matches the format of our AST facade.

This PR adds an `AnyNode` union that is a union over all nodes defined in the grammar, allowing to cast any `SyntaxNode` to a Node and, from there, call the correct `Debug` implementation for a node. So that `Unknown` nodes now print their children using their AST representation.

The motivation behind this is that it's currently difficult to verify if a tree containing `Unknown` nodes is conforming to our AST facade or not. 
## Test Plan

Verified the RAST output

### DIFF
--- a/crates/rslint_parser/test_data/inline/err/function_decl_err.rast
+++ b/crates/rslint_parser/test_data/inline/err/function_decl_err.rast
@@ -134,11 +134,9 @@ JsModule {
         JsExpressionStatement {
             expression: JsUnknownExpression {
                 items: [
-                    Node(
-                        0: JS_REFERENCE_IDENTIFIER@118..125
-                          0: IDENT@118..125 "yield" [Whitespace("\n")] [Whitespace(" ")]
-                        ,
-                    ),
+                    JsReferenceIdentifier {
+                        value_token: IDENT@118..125 "yield" [Whitespace("\n")] [Whitespace(" ")],
+                    },
                 ],
             },
             semicolon_token: missing (optional),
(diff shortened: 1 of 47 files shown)


crates/rslint_parser/test_data/inline/err/function_decl_err.rast
```diff
@@ -134,11 +134,9 @@ JsModule {
...
             expression: JsUnknownExpression {
                 items: [
-                    Node(
-                        0: JS_REFERENCE_IDENTIFIER@118..125
-                          0: IDENT@118..125 "yield" [Whitespace("\n")] [Whitespace(" ")]
-                        ,
-                    ),
+                    JsReferenceIdentifier {
+                        value_token: IDENT@118..125 "yield" [Whitespace("\n")] [Whitespace(" ")],
+                    },
```
MichaReiser, line 139: Hmm, I need to revisit this. I don't think the `reference` identifier should remain a "valid" node because it's referencing an invalid identifier.

xtask/js.ungram
```diff
@@ -1051,7 +1052,7 @@ TsExportAssignment = 'export' '=' expr: JsAnyExpression ';'?
 // typescript namespace export
 TsNamespaceExportDecl = 'export' 'as' 'namespace' ident: Ident? ';'?
 
-Specifier = name:Ident 'as'? manual__alias:Name?
+Specifier = name:Ident 'as'? manual__alias: JsName?
```
MichaReiser, line 1055: Ohoh, this must have been a merge error from my side when I rebased parse import statements.

ematipico (approved):

PR #1869, opened 2021-12-10
refactor: use parse list

ematipico:
## Summary

This PR is part of #1725

It implements the traits `ParseNodeList` and `ParseSeparatedList` for the majority of lists we have inside our parser.

Some lists where not refactored because it involived great changes or were part of some code that will be changed anyway.

The changes of `.rast` files were because now the lists are marked as empty if no children were found.

## Test Plan

No regressions.

### DIFF
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -1,4 +1,4 @@
-use crate::parser::{ParsedSyntax, ParserProgress};
+use crate::parser::{ParsedSyntax, ParserProgress, RecoveryResult};
 use crate::syntax::binding::parse_binding;
 use crate::syntax::decl::{parse_formal_param_pat, parse_parameter_list, parse_parameters_list};
 use crate::syntax::expr::expr_or_assignment;
@@ -12,8 +12,8 @@ use crate::syntax::typescript::{
 };
 use crate::ParsedSyntax::{Absent, Present};
 use crate::{
-	CompletedMarker, ConditionalSyntax, Event, Invalid, Marker, ParseRecovery, Parser, ParserState,
-	StrictMode, TokenSet, Valid,
+	CompletedMarker, ConditionalSyntax, Event, Invalid, Marker, ParseNodeList, ParseRecovery,
+	Parser, ParserState, StrictMode, TokenSet, Valid,
 };
 use rslint_errors::Diagnostic;
 use rslint_syntax::SyntaxKind::*;
@@ -121,7 +121,7 @@ fn parse_class(p: &mut Parser, kind: ClassKind) -> ParsedSyntax<CompletedMarker>
 	implements_clause(&mut guard);
 
 	guard.expect_required(T!['{']);
-	parse_class_members(&mut *guard);
+	ClassMembersList.parse_list(&mut *guard);
 	guard.expect_required(T!['}']);
 
 	Present(m.complete(&mut *guard, kind.into()))
@@ -214,18 +214,16 @@ fn extends_clause(p: &mut Parser) {
 	m.complete(p, JS_EXTENDS_CLAUSE);
 }
 
-/// Parses a list of class members
-fn parse_class_members(p: &mut Parser) {
-	let members = p.start();
+struct ClassMembersList;
 
-	let mut progress = ParserProgress::default();
-	while !p.at(EOF) && !p.at(T!['}']) {
-		progress.assert_progressing(p);
+impl ParseNodeList for ClassMembersList {
+	type ParsedElement = CompletedMarker;
 
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
 		let checkpoint = p.checkpoint();
 		let parsed_member = parse_class_member(p);
 
-		let to_recover = match parsed_member {
+		match parsed_member {
 			Valid(marker) => Present(marker),
 			// In this case we want to mark members that have errors as invalid.
 			//
@@ -238,23 +236,27 @@ fn parse_class_members(p: &mut Parser) {
 				p.rewind(checkpoint);
 				Absent
 			}
-		};
+		}
+	}
 
-		let member_recovered = to_recover.or_recover(
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		p.at(T!['}'])
+	}
+
+	fn recover(
+		&mut self,
+		p: &mut Parser,
+		parsed_element: ParsedSyntax<Self::ParsedElement>,
+	) -> RecoveryResult {
+		parsed_element.or_recover(
 			p,
 			&ParseRecovery::new(
 				JS_UNKNOWN_MEMBER,
 				token_set![T![;], T![ident], T![async], T![yield], T!['}'], T![#]],
 			),
 			js_parse_error::expected_class_member,
-		);
-
-		if member_recovered.is_err() {
-			break;
-		}
+		)
 	}
-
-	members.complete(p, LIST);
 }
 
 fn parse_class_member(p: &mut Parser) -> ConditionalSyntax {
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -9,6 +9,7 @@ use super::typescript::*;
 use super::util::*;
 #[allow(deprecated)]
 use crate::parser::single_token_parse_recovery::SingleTokenParseRecovery;
+use crate::parser::RecoveryError::AlreadyRecovered;
 use crate::parser::{ParserProgress, RecoveryResult};
 use crate::syntax::assignment::{
 	expression_to_assignment, expression_to_assignment_pattern, parse_assignment,
@@ -637,11 +638,11 @@ impl ParseSeparatedList for ArgumentsList {
 		p: &mut Parser,
 		parsed_element: ParsedSyntax<Self::ParsedElement>,
 	) -> RecoveryResult {
-		parsed_element.or_recover(
-			p,
-			&ParseRecovery::new(JS_UNKNOWN_EXPRESSION, STARTS_EXPR).enable_recovery_on_line_break(),
-			js_parse_error::expected_case_or_default,
-		)
+		if parsed_element.is_absent() {
+			Err(AlreadyRecovered)
+		} else {
+			Ok(parsed_element.unwrap())
+		}
 	}
 
 	fn separating_element_kind(&mut self) -> SyntaxKind {
@@ -657,7 +658,7 @@ impl ParseSeparatedList for ArgumentsList {
 ///
 /// `"(" (AssignExpr ",")* ")"`
 
-// test_err invalid_arg_LIST
+// test_err invalid_arg_list
 // foo(a,b;
 // foo(a,b var
 pub fn args(p: &mut Parser) -> CompletedMarker {
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -1193,10 +1193,10 @@ impl ParseSeparatedList for ArrayElementsList {
 	type ParsedElement = CompletedMarker;
 
 	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
-		if p.at(T![...]) {
-			Present(spread_element(p))
-		} else {
-			expr_or_assignment(p).into()
+		match p.cur() {
+			T![...] => Present(spread_element(p)),
+			T![,] => Present(p.start().complete(p, JS_ARRAY_HOLE)),
+			_ => expr_or_assignment(p).into(),
 		}
 	}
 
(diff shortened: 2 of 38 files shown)


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -238,23 +236,27 @@ fn parse_class_members(p: &mut Parser) {
 				p.rewind(checkpoint);
 				Absent
 			}
-		};
+		}
+	}
 
-		let member_recovered = to_recover.or_recover(
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		p.at(T!['}'])
```
MichaReiser, line 243: Unrelated to this PR. I'm wondering if we should use the `STMT_RECOVERY_SET` as well. So that it recovers when it sees a 

```
class Foo {

function test() {}
```

Because the function likely doesn't belong to the class.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -615,6 +615,47 @@ fn is_at_identifier_name(p: &Parser) -> bool {
...
+	type ParsedElement = CompletedMarker;
+
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
+		if p.at(T![...]) {
+			Present(spread_element(p))
+		} else {
+			match expr_or_assignment(p) {
+				Some(marker) => Present(marker),
+				None => Absent,
+			}
```
MichaReiser, line 630: Nit: `expr_or_assignment(p).into()`

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -615,6 +615,47 @@ fn is_at_identifier_name(p: &Parser) -> bool {
...
+		}
+	}
+
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		p.at(T![')'])
```
MichaReiser, line 635: Nit: Should we include `STMT_RECOVERY_SET` here to recover from

```js
function test(a

class Test {}
```

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1167,6 +1189,78 @@ pub fn template(p: &mut Parser, tag: Option<CompletedMarker>) -> CompletedMarker
...
+
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
+		if p.at(T![...]) {
+			Present(spread_element(p))
+		} else {
+			match expr_or_assignment(p) {
+				Some(marker) => Present(marker),
+				None => Absent,
+			}
+		}
```
MichaReiser, line 1205: Nit
```suggestion
			expr_or_assignment(p).into()
	}
```

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1167,6 +1189,78 @@ pub fn template(p: &mut Parser, tag: Option<CompletedMarker>) -> CompletedMarker
...
+	}
+
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		p.at(T![']'])
```
MichaReiser, line 1209: Same with `STMT_RECOVERY_SET`

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1167,6 +1189,78 @@ pub fn template(p: &mut Parser, tag: Option<CompletedMarker>) -> CompletedMarker
...
+
+			if self.is_at_list_end(p) {
+				break;
+			}
+
+			self.expect_separator(p);
+		}
+		elements_list.complete(p, LIST);
+	}
+}
```
MichaReiser, line 1262: What's the reason for overriding `parse_list`. Can't we handle `JS_ARRAY_HOLE` inside of `parse_element`?

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -18,49 +18,50 @@ use rslint_syntax::T;
...
-		if p.at(T![,]) {
-			p.missing();
-			continue;
-		}
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
+		parse_object_member(p)
+	}
+
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		p.at(T!['}'])
```
MichaReiser, line 31: Nit: Include `STMT_RECOVERY_SET`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -576,24 +585,45 @@ pub(crate) fn directives(p: &mut Parser) -> Option<ParserState> {
...
+		p: &mut Parser,
+		parsed_element: ParsedSyntax<Self::ParsedElement>,
+	) -> RecoveryResult {
+		parsed_element.or_recover(
+			p,
+			&ParseRecovery::new(JS_UNKNOWN_STATEMENT, STMT_RECOVERY_SET),
+			js_parse_error::expected_directive,
+		)
+	}
+}
```
MichaReiser, line 620: I'm not sure if it makes sense to use `ParseList` for directives because we won't need any error recovery for directives (we only parse as long as it's a string literal... if not, we abort. 
```suggestion
	fn recover(
		&mut self,
		p: &mut Parser,
		parsed_element: ParsedSyntax<Self::ParsedElement>,
	) -> RecoveryResult {
		Err(RecoveryError::AlreadyRecovered)
	}
}
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -549,13 +550,18 @@ pub(super) fn parse_block_impl(
...
-	let mut old_state: Option<ParserState> = None;
+impl Default for DirectivesList {
+	fn default() -> Self {
+		Self { old_state: None }
+	}
+}
 
-	fn is_directive(p: &Parser) -> bool {
+impl DirectivesList {
+	pub fn at_directives(&mut self, p: &mut Parser) -> bool {
```
MichaReiser, line 564: Can we make this private? 
```suggestion
	fn is_at_directive(&mut self, p: &mut Parser) -> bool {
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -806,6 +836,45 @@ pub fn variable_declaration_statement(p: &mut Parser) -> ParsedSyntax<CompletedM
...
+}
+
+impl ParseNodeList for VariableDeclaratorList {
+	type ParsedElement = CompletedMarker;
+
+	fn start_list(&mut self, p: &mut Parser) -> Marker {
+		let marker = p.start();
+		variable_declarator(p, &self.const_range, self.no_semi, self.is_let);
+		marker
+	}
```
MichaReiser, line 852: This isn't correctly handling the comma after the first declarator, isn't it? 

You can also leave the variable declarations, I rewrote them to use `ParseList` in #1870

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1145,6 +1211,33 @@ impl ParseNodeList for SwitchClausesList {
...
 
+struct ConsList;
+impl ParseNodeList for ConsList {
+	type ParsedElement = CompletedMarker;
+
+	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
+		match parse_statement(p, None) {
+			None => Absent,
+			Some(marker) => Present(marker),
+		}
```
MichaReiser, line 1222:
```suggestion
		parse_statement(p, None).into()
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1239,52 +1389,60 @@ pub fn parse_switch_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+	// 	let mut temp = p.with_state(ParserState {
+	// 		break_allowed: true,
+	// 		..p.state.clone()
+	// 	});
+	//
+	// 	let clause = parse_switch_clause(&mut *temp, &mut first_default);
+	//
+	// 	if let Present(marker) = clause {
+	// 		if marker.kind() == JS_DEFAULT_CLAUSE && first_default == None {
+	// 			first_default = Some(marker);
```
MichaReiser, line 1415: Delete?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1207,6 +1293,70 @@ fn parse_switch_clause(
 		}
 	}
 }
+#[derive(Default)]
+struct SwitchStatementList {
+	first_default: Option<CompletedMarker>,
+}
+
+impl ParseNodeList for SwitchStatementList {
```
MichaReiser, line 1301: Nit: `SwitchCaseList`... It's parsing the cases, not the statements.

crates/rslint_parser/src/tests.rs
```diff
@@ -13,9 +13,13 @@ use std::path::{Path, PathBuf};
 #[test]
 fn parser_smoke_test() {
 	let src = r#"
-    console.log("hello world");
+switch (foo) {
+ case bar:
+ default:
+}
```
MichaReiser, line 19: Revert?

crates/rslint_parser/test_data/inline/err/invalid_arg_list.rast
```diff
@@ -24,11 +24,17 @@ JsModule {
                                 value_token: IDENT@6..7 "b" [] [],
                             },
                         },
+                        missing separator,
+                        JsUnknownExpression {
+                            items: [
+                                SEMICOLON@7..8 ";" [] [],
+                            ],
+                        },
```
MichaReiser, line 32: What we had before seemed more accurate..

crates/rslint_parser/test_data/inline/err/invalid_arg_list.rast
```diff
@@ -73,59 +78,74 @@ JsModule {
...
 1 │ foo(a,b;
   │        ^ unexpected
 
 --
-error[SyntaxError]: expected `')'` but instead found `var`
+error[SyntaxError]: expected a default, or a case but instead found 'foo'
+  ┌─ invalid_arg_list.js:2:1
+  │
+2 │ foo(a,b var
+  │ ^^^ Expected a default, or a case here
```
MichaReiser, line 133: why do we now have these additional errors?

crates/rslint_parser/test_data/inline/err/object_expr_err.rast
```diff
@@ -14,110 +14,133 @@ JsModule {
...
+  │          ^ Expected a property, a shorthand property, a getter, a setter, or a method here
+
+--
+error[SyntaxError]: Expected a statement or declaration, but found none
+  ┌─ object_expr_err.js:1:15
+  │
+1 │ let a = {, foo}
+  │               ^ Expected a statement or declaration here
+
+--
```
MichaReiser, line 128: The old implementation seemed to have better recovery than the new one.

MichaReiser (changes requested):
Nice! There's a lot of code that simply is gone when adopting the new parse list. 

Ideally, this PR shouldn't change the way the parser recovers and it seems that it now recovers worse than before. Can you have a look at it?

crates/rslint_parser/src/syntax/class.rs
ematipico: I would create a different const for these cases. It might work though

crates/rslint_parser/src/syntax/object.rs
ematipico: I won't do it in this PR. This will change some logic that is not related to this PR. As you said, this PR should not change the way we do stuff.

crates/rslint_parser/src/syntax/expr.rs
ematipico: It's the custom logic that we have inside the parsing of the list. I tried normal means but it wouldn't work. Feel free to take stab at it.

MichaReiser: I managed to get it working by moving the `progressing` assertion before `parse_element`. So that empty elements still progress if they're followed by a separator.I'll push soon.

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -18,49 +18,88 @@ use rslint_syntax::T;
...
+					break;
+				}
+			}
+			progress.assert_progressing(p);
+
+			// missing member
+			if p.at(self.separating_element_kind()) {
+				p.missing();
+				continue;
+			}
```
MichaReiser, line 76: I added support for missing members to the separated list parsing.

crates/rslint_parser/test_data/inline/err/invalid_arg_list.rast
```diff
@@ -113,14 +124,21 @@ JsModule {
       1: (empty)
   3: EOF@20..21 "" [Whitespace("\n")] []
 --
-error[SyntaxError]: expected `')'` but instead found `;`
+error[SyntaxError]: expected `,` but instead found `;`
   ┌─ invalid_arg_list.js:1:8
   │
 1 │ foo(a,b;
   │        ^ unexpected
```
MichaReiser, line 132: The old behaviour looked more correct. an we restore it?

MichaReiser (approved):

PR #1870, opened 2021-12-10
feature(rslint_parser): For in/of and Variable facade

MichaReiser:
## Summary

- Renames `EqualValueClause` to `JsInitializerClause`
- Renames `JsVariableDeclarationStatement` to `JsVariableStatement`
- Renames `JsVariableDeclaration` to `JsVariableDeclarationList`
- Rename `JsVariableDeclarator` to `JsVariableDeclaration`
- Rename `ForInStmt` to `JsForInStatement`
- Rename `ForOfStmt` to `JsForOfStatement`
- Delete `ForStmtInit`
- Introduce `JsForVariableDeclaration` that only allows a single variable declaration.

Deletes the `parse_marker` from the `Parser`.

This PR doesn't change the error recovery for `for` statements nor for variable declarations because of the pending discussion in #1867. I'll spend some time now to prototype an approach that uses a generic `Unknown` to remove conditional syntax.

## Test Plan

The parser's performance seems to improve a bit. I suspect this is stemming from the removal of the `parse_marker` to verify if the variable declaration of a for of/in loop only contains a single variable.

* jquery: 6.98ms -> 6.49ms
* vue: 13.5788ms -> 12.194ms
* react: 496.9µs -> 461.5µs
* react-dom: 8.6984ms -> 7.7618ms
* compiler: 26.9218ms -> 22.8185ms
* dojo:  1.8681ms -> 1.6019ms
* d3: 23.1232ms -> 19.6941ms
* pixi: 27.6754ms -> 23.5175ms
* three: 31.8306ms -> 28.224ms
* tex: 73.1209ms -> 46.4471ms


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -614,58 +616,7 @@ pub(crate) fn parse_statements(
...
-						.primary(m.range(p), "move this declaration to the top level");
-
-					p.error(err);
-				}
-				m.change_kind(p, JS_UNKNOWN_STATEMENT);
-			}
-			_ => {
-				parse_statement(p, recovery_set);
-			}
-		};
```
MichaReiser, line 668: It seems I really messed up the merging in #1856

xtask/js.ungram
```diff
@@ -144,27 +144,49 @@ JsElseClause =
...
 	manual__second_semi: ';'
 	update:ForStmtUpdate?
 	')'
 	cons:JsAnyStatement
 
-ForStmtInit = inner: ForHead
-ForHead = JsVariableDeclaration | JsAnyExpression
+JsAnyForInitializer = JsVariableDeclarationList | JsAnyExpression
+
+// TODO Inline when AST facade uses fixed offsets
```
ematipico, line 157: Can we add the issue number of the fixed offsets?

crates/rome_formatter/src/ts/declarators/variable_declaration_statement.rs
```diff
@@ -2,34 +2,33 @@ use crate::{
...
-		let declarators = formatter.format_separated(self.declarators())?;
-
 		Ok(format_elements![
 			formatter.format_token(&self.kind_token()?)?,
 			space_token(),
-			join_elements(space_token(), declarators),
+			join_elements(
+				space_token(),
+				formatter.format_separated(self.declarations())?
+			),
```
ematipico, line 24: I don't think we have at test where have long lists of declarations and by looking at the code, it doesn't seem we handle it. We should add a `TODO` as reminder or fix it now.

crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -107,8 +107,8 @@ pub trait ParseSeparatedList {
 
 	/// It creates a [ParsedSyntax] that will contain the list
 	/// Only called if the list isn't empty
-	fn finish_list(&mut self, p: &mut Parser, m: Marker) {
-		m.complete(p, SyntaxKind::LIST);
+	fn finish_list(&mut self, p: &mut Parser, m: Marker) -> CompletedMarker {
+		m.complete(p, SyntaxKind::LIST)
```
ematipico, line 111: What's the reason of change the API signature? Personally I would leave the API like this and I would introduce some new flag and API to retrieve the list marker on demand. 

Doing so would have a waterfall effect where every list needs a `.unwrap`, which goes against what we designed before.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -825,48 +794,106 @@ fn parse_variable_declaration(p: &mut Parser, no_semi: bool) -> ParsedSyntax<Com
...
 	}
 
-	declared_list.complete(p, LIST);
-	p.state.name_map.clear();
-	Present(m.complete(p, JS_VARIABLE_DECLARATION))
+	fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+		if self.first {
+			false
+		} else {
+			!p.at(T![,])
```
ematipico, line 847:
```suggestion
			!p.at(self.separating_element_kind())
```

ematipico: Isn't `;` the list end?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -825,48 +794,106 @@ fn parse_variable_declaration(p: &mut Parser, no_semi: bool) -> ParsedSyntax<Com
...
+	}
 }
 
 // A single declarator, either `ident` or `ident = assign_expr`
-pub(crate) fn variable_declarator(
+pub(crate) fn parse_variable_declaration(
 	p: &mut Parser,
 	is_const: &Option<Range<usize>>,
-	for_stmt: bool,
+	for_in_or_of: bool,
```
ematipico, line 878: The name of the variable and the type are misleading. Is should start with `is_*` or be a `enum`. 

```rs
enum For {
	In,
	Of
}
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -966,57 +983,69 @@ pub fn parse_do_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+			});
 
-		if p.at(T![in]) || p.cur_src() == "of" {
-			if let Some(err) = p.state.for_head_error.take() {
-				p.error(err);
+			parse_variable_declarations(&mut *guard, true).unwrap()
+		};
+
+		let is_in = p.at(T![in]);
+		let is_of = p.cur_src() == "of";
```
ematipico, line 1010: Can't we use `T![of]`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -966,57 +983,69 @@ pub fn parse_do_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+
+		if is_in || is_of {
+			if let Some(declarations_list) = declarations {
+				declarations_list.undo_completion(p).abandon(p);
```
ematipico, line 1014: What's happening here?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -966,57 +983,69 @@ pub fn parse_do_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
-			} else {
-				p.bump_remap(T![of]);
+
+			if let Some(additional_declarations_range) = additional_declarations {
+				p.error(
+					p.err_builder(&format!(
+						"Only a single declaration is allowed in a `for...{}` statement.",
+						if is_of { "of" } else { "in" },
+					))
+					.primary(additional_declarations_range, "additional declarations"),
```
ematipico, line 1023: If with `declaration` we mean **variable declaration**, then it's wrong. We can have multiple declarations (I tried various parsers and they don't emit errors). And I think the changes we applied to the grammar reflect that.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1113,7 +1168,21 @@ pub fn parse_for_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
 		None,
 	);
-	Present(m.complete(p, kind))
+
+	let mut completed = m.complete(p, kind);
+
+	if let Some(await_range) = await_range {
+		if kind != JS_FOR_OF_STATEMENT {
+			p.error(
+				p.err_builder("await can only be used in conjunction with for of statements")
```
ematipico, line 1177:
```suggestion
				p.err_builder("await can only be used in conjunction with `for..of` statements")
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1113,7 +1168,21 @@ pub fn parse_for_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+					.primary(await_range, "Remove the await here")
+					.secondary(completed.range(p), "or convert this to a for of statement"),
```
ematipico, line 1179:
```suggestion
					.secondary(completed.range(p), "or convert this to a `for..of` statement"),
```

crates/rslint_parser/test_data/inline/err/for_stmt_err.rast
```diff
@@ -249,7 +498,59 @@ error[SyntaxError]: expected `')'` but instead found `{`
...
+  │
+4 │ for (in []) {}
+  │           ^ unexpected
+
+--
+error[SyntaxError]: Only a single declaration is allowed in a `for...of` statement.
+  ┌─ for_stmt_err.js:5:13
+  │
+5 │ for (let i, j = 6 of []) {}
+  │             ^^^^^ additional declarations
```
ematipico, line 527: We should not have this error

ematipico (changes requested):
I left some comments on the PR.

I would personally revert the changes on the `ParseSeparatedList`. We can retrieve the list marker in other ways and would prefer the way where we change the API signature, just for one case.

crates/rslint_parser/src/parser/parse_lists.rs
MichaReiser: What do you have in mind? 

There's no waterfall effect because there's no need to call `unwrap` on `CompletedMarker` nor `Option`.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: `T![,]` seems cleaner to me... one less interaction. It's testing if we're currently add a comma, which separates two declarations.

```
let a, b
      ^
```

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: The `JsForVariableDeclaration` directly contains the variable declaration without an intermediate `JsVariableDeclarationList`. That's why I need to abandon it. I can add a comment.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: `for of` and `for in` can only have one... Try

```
for (let a, b of c) {}
for (let a, b, in c) {}
```

This is different for `for in`, where `for (let a, b = 1;;) {}` works perfectly fine.

crates/rslint_parser/src/parser/parse_lists.rs
ematipico: Then we should apply this change to the other list too.

crates/rslint_parser/test_data/inline/err/for_stmt_err.rast
MichaReiser: See comment above.

crates/rslint_parser/src/parser/parse_lists.rs
MichaReiser: This change is automatically applied for other lists as long as they don't override `finish_list`. I can update the documentation to mention that it now returns the completed marker if the list isn't empty

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Oh yeah, I used the normal `for()` lol. Sorry about that!

crates/rslint_parser/src/syntax/stmt.rs
ematipico: Yes please, it's a strange path. A comment would help to understand

crates/rslint_parser/src/syntax/stmt.rs
ematipico: I know, but we wouldn't need to maintain `T![,]` twice :) 
I think using the API is better

MichaReiser: I prefer the current solution because I doubt there's ever anything to maintain. It's super unlikely that the separator for variable declarations will ever change and I find the inline `T![,]` easier to read than having to jump to the method definition.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: Nope, `of` is a contextual keyword and the lexer returns it as an identifier.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: I introduced a `VariableDeclarationContext` that can either be a `for` or `variable statement`

crates/rslint_parser/src/syntax/stmt.rs
ematipico: The separator of the variable won't change, but one day the macro might. 

> easier to read than having to jump to the method definition.

Then, let's remove the `separating_element_kind` :)

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -825,48 +809,112 @@ fn parse_variable_declaration(p: &mut Parser, no_semi: bool) -> ParsedSyntax<Com
...
-pub(crate) fn variable_declarator(
+fn parse_variable_declaration(
 	p: &mut Parser,
 	is_const: &Option<Range<usize>>,
-	for_stmt: bool,
 	is_let: bool,
-) -> Option<CompletedMarker> {
+	is_first: bool,
+	declaration_context: VariableDeclarationContext,
+) -> ParsedSyntax<CompletedMarker> {
```
ematipico, line 901: The quantity of arguments here is too much I think. Wouldn't it better to use a `struct` as this point? Doing so, we can also document the arguments. `is_first` doesn't say much to me

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -966,57 +1009,72 @@ pub fn parse_do_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
...
+			parse_variable_declarations(&mut *guard, VariableDeclarationContext::For).unwrap()
+		};
+
+		let is_in = p.at(T![in]);
+		let is_of = p.cur_src() == "of";
+
+		if is_in || is_of {
+			if let Some(declarations_list) = declarations {
+				// remove the intermediate list node created by parse variable declarations that is needed
+				// for a ForInOrOfInitializer
```
ematipico, line 1042: I was hoping to have some explanation of _why_ we are removing the list.

ematipico (changes requested):

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: Lol sorry, there's a *not* missing... *that is not needed for a `ForInOrOfInitializer`

ematipico (approved):

PR #1874, opened 2021-12-14
chore: add license for internal forks

ematipico:
## Summary

When we locally copied the code from `rowan` and `rslint`, we forgot to copy over their licenses.

This PR:
- adds the license files of the [`rowan` repo](https://github.com/rust-analyzer/rowan) (they are two);
- adds the [license file of `rslint` project](https://github.com/rslint/rslint/blob/master/LICENSE); the same very file was copied inside the internal forks that we intentionally selected;

## Test Plan

None

### DIFF
--- a/crates/rome_rowan/LICENSE-APACHE
+++ b/crates/rome_rowan/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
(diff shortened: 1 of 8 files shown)


yassere (commented):
I don't know much about licensing requirements, so I'll leave that for someone else to review. Only thing I noticed was the "lisence" typo in the PR title that we should fix before merging.

Waye254222 (approved):

xunilrj (approved):

PR #1876, opened 2021-12-15
test(formatter): reorganize snapshots

Boshen:
## Summary

* copied rome-classic snapshots over
* grouped tests into structured directories

Some noticeable things we can improve on:
* print line breaks
* print regex
* print comments 

## Test Plan
`cargo test -p rome_formatter` passes


crates/rome_formatter/tests/specs/js/arrow/call.js
```diff
@@ -0,0 +1,44 @@
+const testResults = results.testResults.map(testResult =>
+    formatResult(testResult, formatter, reporter)
+  );
+
+it('mocks regexp instances', () => {
+	expect(
+		// () => moduleMocker.generateFromMetadata(moduleMocker.getMetadata(/a/)),
```
Boshen, line 7: commented out because regex is not implemented yet

ematipico:
> print line breaks

What do you have in mind?

Boshen:
> > print line breaks
> 
> What do you have in mind?

I don't have any thing in mind yet, but prettier preserves original new line, and spaces things out nicely. 

We should probably have this conversation somewhere else, since I don't know the guide line for Rome formatter yet (is there one?)

ematipico:
> > > print line breaks
> > 
> > 
> > What do you have in mind?
> 
> I don't have any thing in mind yet, but prettier preserves original new line, and spaces things out nicely.
> 

Feel free to open a Github discussion

> We should probably have this conversation somewhere else, since I don't know the guide line for Rome formatter yet (is there one?)

For the time being, we want to keep Prettier as reference.

Boshen:
Great! Anything I should change on this pr before I move onto the code coverage task?

crates/rome_formatter/tests/specs/js/arrow/call.js.snap
```diff
@@ -0,0 +1,77 @@
...
+	url: "/test-endpoint-but-with-a-long-url",
+})).not.toThrowError();
+const a = Observable.fromPromise(axiosInstance.post("/carts/mine")).map((
+	response
+) => response.data);
+const b = Observable.fromPromise(axiosInstance.get(url)).map((response) => response.data);
+func(veryLoooooooooooooooooooooooongName,(veryLooooooooooooooooooooooooongName) => veryLoooooooooooooooongName.something());
+const composition = (ViewComponent, ContainerComponent) => class extends React.Component {
+	static propTypes = {};
+};
```
MichaReiser, line 75: Lol... is our formatter a minifier or formatter xD

Boshen: A transition from a minifier to a formatter 😄

ematipico (approved):
Now that we have the tests from the old Rome, I think one important thing that we should do is to actually match its formatting style to the new formatter.

Rome's old formatter was pretty stable (exception for comments and some JSX case), so I think it's worth using its formatting style whereas possible.

PR #1879, opened 2021-12-15
feature(rslint_parser): Class Modifiers

MichaReiser:
## Summary

I noticed that the class parsing doesn't conform to our AST structure when working on the automatic AST shape validation (marks nodes as `UNKNOWN` if they don't match the expected AST shape).

The reason for that is that some members have `missing` slots for modifiers they don't support or a `missing` slot for an optional modifier is missing.

I further noticed that the parsing didn't correctly support members like `private()` because we only had special handling for `declare` and `static`.

This PR rewrites the modifiers handling and adds support for all valid member names (including `declare`, `static`, `private`, ...).

## Test Plan

Added new test cases. Verified that my prototype no longer marked members as invalid/unknown


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -493,62 +363,70 @@ fn parse_class_member(p: &mut Parser) -> ParsedSyntax<ConditionalSyntax> {
...
+		// undo invalid modifiers for methods
+		modifiers.undo_if_missing(&mut *guard, ModifierKind::Declare);
+		modifiers.undo_if_missing(&mut *guard, ModifierKind::Readonly);
+
 		return Present(parse_method_class_member(&mut *guard, member_marker))
 			.into_conditional(member_is_valid);
 	}
 
+	// Insert the missing markers for async and generator for the case this turns out to
+	// be a method member. The marker must be `undo`n if it turns out this is any other member.
```
ematipico, line 387: I don't understand this part of the comment: " The marker must be `undo`n "

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -976,115 +862,274 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax<Co
...
+	let mut progress = ParserProgress::default();
+	loop {
+		progress.assert_progressing(p);
+
+		// set_range doesn't work because this is already too late..
+		if let Some(modifier) = parse_modifier(p, &mut modifiers) {
+			if let Some(existing) = modifiers.get_range(modifier.kind) {
+				p.error(
+					p.err_builder(&format!(
+						"`{}` modifier already seen.",
```
ematipico, line 939: I think we can have a better error message than typescript :)

"{} modifier already used, please use {} only once"

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -976,115 +862,274 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax<Co
...
+		"static" => (ModifierKind::Static, STATIC_KW),
+		"readonly" => (ModifierKind::Readonly, READONLY_KW),
+		"abstract" => (ModifierKind::Abstract, ABSTRACT_KW),
+		_ => {
+			return None;
+		}
+	};
+
+	// Fill in missing placeholders for all modifiers preceding this modifier
+	modifiers.create_missing_for_modifiers_preceding(p, modifier_kind);
```
ematipico, line 1022: I may understand this bit but I think it would be great if we can have a better a comment 
that explains why we are doing that. It's the core/trick of the whole algorithm, especially because this function is called from a loop and by itself it's hard to understand.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -976,115 +862,274 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax<Co
...
-		} else {
-			p.missing();
-		}
-		if self.declare {
-			p.bump_remap(T![declare]);
-			if let Some(err) = self.declare_err {
-				p.error(err);
-				member_is_valid = false;
+
+			if let Some(last) = &last {
```
ematipico, line 949: What's the heuristic of when and what modifier must preceede another? 
Maybe it's worth adding a comment by asking the developer to check the enum `ModifierKind` where we explain that.

Or maybe it's worth moving the enums up, before the functions. After all we read the code, usually, from top to bottom. Or we can do both.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -976,115 +862,274 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax<Co
...
+/// Inserts `missing` marker for all possible class modifiers. These must be undone if a member
+/// doesn't support a specific modifier.
+///
+/// Returns [Ok] if the modifiers are in the correct order, no typescript modifiers are used, or this
+/// is a typescript file
+/// Returns [Err] otherwise
+fn parse_class_member_modifiers(
+	p: &mut Parser,
+) -> Result<ClassMemberModifiers, ClassMemberModifiers> {
+	let mut last: Option<Modifier> = None;
```
ematipico, line 926: It should be `latest`, not `last`. Or `previous_modifier`, even better.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -976,115 +862,274 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax<Co
...
+	let mut valid = true;
+	let mut modifiers = ClassMemberModifiers::default();
+
+	let mut progress = ParserProgress::default();
+	loop {
+		progress.assert_progressing(p);
+
+		// set_range doesn't work because this is already too late..
+		if let Some(modifier) = parse_modifier(p, &mut modifiers) {
```
ematipico, line 935: I would rename `modifier` with `current_modifier` which is a good pair against `previous_modifier`, espcially when we compare them.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -493,62 +363,70 @@ fn parse_class_member(p: &mut Parser) -> ParsedSyntax<ConditionalSyntax> {
...
-				.primary(range, "");
-
-			p.error(err);
-		}
-
-		// test_err class_constructor_err
+		// test class_static_constructor_method
 		// class B { static constructor() {} }
 		return if is_constructor {
+			async_missing_marker.undo(p);
```
ematipico, line 403: Maybe we can add a comment explaining why we are undoing these missing markers.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -493,62 +363,70 @@ fn parse_class_member(p: &mut Parser) -> ParsedSyntax<ConditionalSyntax> {
...
+			modifiers.undo_if_missing(p, ModifierKind::Static);
+			modifiers.undo_if_missing(p, ModifierKind::Readonly);
+			modifiers.undo_if_missing(p, ModifierKind::Declare);
+
 			let constructor = parse_constructor_class_member_body(p, member_marker);
 			let mut constructor_has_error = false;
-			if let Present(Valid(constructor)) = constructor {
-				if is_static {
-					let err = p
-						.err_builder("constructors cannot be static")
```
ematipico, line 535: LOL are you telling me that we were having a false error here? XD Damn!

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -561,15 +439,31 @@ fn parse_class_member(p: &mut Parser) -> ParsedSyntax<ConditionalSyntax> {
...
+			modifiers.undo_if_missing(p, ModifierKind::Declare);
+			modifiers.undo_if_missing(p, ModifierKind::Readonly);
+
 			Present(parse_method_class_member_body(p, member_marker))
 				.into_conditional(member_is_valid)
 		};
 	}
 
-	if let NodeMarker(member) = member {
+	async_missing_marker.undo(p);
```
ematipico, line 460: Small comment to explain the undoing

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -775,23 +668,25 @@ fn parse_property_class_member_body(
 }
 
 /// Eats the ? token for optional member. Emits an error if this isn't typescript
-fn optional_member_token(p: &mut Parser) -> Result<Range<usize>, ()> {
-	if p.at(T![?]) {
+fn optional_member_token(p: &mut Parser) -> Result<RangeOrMissingMarker, ()> {
```
ematipico, line 671: I don't quite understand why we used `RangeOrMissingMarker`. 
Other than a change in the enums, I struggle to see the diffeerence, especially because we don't use `RangeOrMissingMarker::Missing`

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -976,115 +862,274 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax<Co
...
+enum RangeOrMissingMarker {
+	Missing(CompletedMissingMarker),
+	Range(Range<usize>),
+}
+
+impl RangeOrMissingMarker {
+	/// Undoes this if it a missing marker
+	fn undo_if_missing(self, p: &mut Parser) {
+		if let RangeOrMissingMarker::Missing(missing) = self {
+			missing.undo(p);
```
ematipico, line 1065: Now that I see more usage of `undo`, I must say that the name of the API doesn't click.

I struggle to make a connection between "undoing" and "something missing". This is probably due the naming of the internals.

I think a better fit might be: `marker.mark_missing()` (in the future, `mark_empty`) and `marker.unmark_missing` (in the future `unmark_empty`). More explicit, I think.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -102,11 +106,19 @@ fn parse_class(p: &mut Parser, kind: ClassKind) -> ParsedSyntax<ConditionalSynta
...
 			}
 		}
+	} else {
+		if kind == ClassKind::Declaration && !guard.state.in_default {
+			let err = guard
+				.err_builder("class declarations must have a name")
+				.primary(class_token_range.start..guard.cur_tok().range.start, "");
+			guard.error(err);
+		}
+		guard.missing();
```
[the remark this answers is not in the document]
MichaReiser: We do... we should probably start splitting tests up again and moving them closer to the lines affecting them.

https://github.com/rome/tools/blob/810a1c8907f8e9ccb785e049494e6dbc8b778d00/crates/rslint_parser/src/syntax/class.rs#L38-L39

crates/rslint_parser/src/syntax/class.rs
MichaReiser: I changed the wording. I want to avoid having the modifier in the message twice.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: I expanded the documentation.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: Hehe, seems we have different takes on the ordering :) I normally hide "details" by moving them below. For example, for react components: At the very top is the react component itself, any helpers (or sub components) are further down because these are only important if you want to dig deeper. I added a comment.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: Kind of. It seems to be an error in TS.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: `missing()` is called on the parser. I intend to rename this to `p.empty_slot()` or we can do `p.mark_empty()` as well. 

I don't like `unmark_missing` because it repeats the name. The name is `CompletedMissingMarker` and you call `undo` on it. So the type is already giving you `missing.undo()`. 

Wdyt of `drop` or `drop_slot`?

MichaReiser:
I go ahead and merge this PR considering that the only pending discussion is around the naming of the `missing` marker method that is likely going away with #1881 anyway. Let's revisit the name if we decide not to go with. #1881

PR #1880, opened 2021-12-15
refactor: expressions to use the new `ParsedSyntax` API

ematipico:
## Summary

This PR presents lot of commits because I decided to refactor each function at the time, instead of doing everything in one go. It's delicate part of our parser so I wanted to make sure that each change didn't break the existing implementation.

I still think there are some missing markers that we don't mark yet, and this is probably because I'm not familiar yet with this part of the grammar.

Here's the list of things:

- changed the grammar name of `optional_chain_token` to `optional_chain`. The `_token` suffix is automatically added by our codegen
- all functions that return `ParsedSyntax` now are prefixed with `parse_` in their name
- error handling or missing expressions
- converted the usage of `Option<CompletedMarker>` around the code base to fit the new `ParsedSyntax<CompletedMarker>`
- added new error test cases to show missing markers

**This PR doesn't handle error recovery yet**, in fact some of the new tests and existing tests show incorrect recovery because that will be tackled in another PR.

## Test Plan

Current CI


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1071,14 +1112,14 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
 				// test import_call
 				// import("foo")
 				p.expect_required(T!['(']);
-				if expr_or_assignment(p).is_none() {
+				if parse_expr_or_assignment(p).is_absent() {
```
MichaReiser, line 1115: `or_missing_with_error`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1087,6 +1128,7 @@ pub fn primary_expr(p: &mut Parser) -> Option<CompletedMarker> {
 		// test_err primary_expr_invalid_recovery
 		// let a = \; foo();
 		_ => {
+			// TODO: remove it and delegate recovery to parent
```
MichaReiser, line 1131: Can we already delete it?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1242,9 +1288,9 @@ impl ParseSeparatedList for ArrayElementsList {
 
 	fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax<Self::ParsedElement> {
 		match p.cur() {
-			T![...] => Present(spread_element(p)),
+			T![...] => spread_element(p),
```
MichaReiser, line 1291: `parse_spread_element`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1285,135 +1331,168 @@ impl ParseSeparatedList for ArrayElementsList {
 // [foo,];
 // [,,,,,foo,,,,];
 // [...a, ...b];
-pub fn array_expr(p: &mut Parser) -> CompletedMarker {
+pub fn array_expr(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
```
MichaReiser, line 1334: Nit: `parse_array_expression`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1285,135 +1331,168 @@ impl ParseSeparatedList for ArrayElementsList {
...
-	m.complete(p, JS_SPREAD)
+	p.bump(T![...]);
+	parse_expr_or_assignment(p)
+		.or_missing_with_error(p, js_parse_error::expected_expression_assignment);
+	Present(m.complete(p, JS_SPREAD))
 }
 
 /// A left hand side expression, either a member expression or a call expression such as `foo()`.
-pub fn lhs_expr(p: &mut Parser) -> Option<CompletedMarker> {
+pub fn lhs_expr(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
```
MichaReiser, line 1360: Can we make this private? 

`parse_lhs_expr`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1285,135 +1331,168 @@ impl ParseSeparatedList for ArrayElementsList {
...
-		super_marker
+		super_syntax
 	} else {
-		member_or_new_expr(p, true)?
+		parse_member_or_new_expr(p, true)
 	};
 
-	if lhs.kind() == JS_ARROW_FUNCTION_EXPRESSION {
-		return Some(lhs);
+	if let Present(m) = lhs {
```
MichaReiser, line 1380: Nit:

if lhs.kind() == Some(JS_ARROW_FUNCTION_EXPRESSION) {
    return lhs;
}

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1285,135 +1331,168 @@ impl ParseSeparatedList for ArrayElementsList {
...
-		let checkpoint = p.checkpoint();
-		let mut complete = try_parse_ts(p, ts_type_args);
-		if !p.at(T!['(']) {
-			p.rewind(checkpoint);
-			None
-		} else {
-			if let Some(ref mut comp) = complete {
-				comp.err_if_not_ts(p, "type arguments can only be used in TypeScript files");
+	match lhs {
+		Present(lhs) => {
```
MichaReiser, line 1387: Use `map`

lhs.map(|lhs| {
    ...
})

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1285,135 +1331,168 @@ impl ParseSeparatedList for ArrayElementsList {
...
-	m.abandon(p);
-	Some(lhs)
 }
 
 /// A postifx expression, either `LHSExpr [no linebreak] ++` or `LHSExpr [no linebreak] --`.
 // test postfix_expr
 // foo++
 // foo--
-pub fn postfix_expr(p: &mut Parser) -> Option<CompletedMarker> {
+pub fn postfix_expr(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
```
MichaReiser, line 1430: `parse_postfix_expr` (and make it private?)

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1285,135 +1331,168 @@ impl ParseSeparatedList for ArrayElementsList {
...
-				let complete = m.complete(p, JS_POST_UPDATE_EXPRESSION);
-				Some(complete)
-			}
-			T![--] => {
-				let assignment_target = expression_to_assignment(p, lhs?, checkpoint);
-				let m = assignment_target.precede(p);
-				p.bump(T![--]);
-				let complete = m.complete(p, JS_POST_UPDATE_EXPRESSION);
-				Some(complete)
+	match lhs {
```
MichaReiser, line 1433: Use `map`

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1422,27 +1501,27 @@ pub fn unary_expr(p: &mut Parser) -> Option<CompletedMarker> {
...
 	if p.at_ts(UNARY_SINGLE) {
 		let m = p.start();
 		let op = p.cur();
 		p.bump_any();
 
-		let res = if let Some(unary) = unary_expr(p) {
+		let res = if let Present(unary) = unary_expr(p) {
 			unary
 		} else {
 			m.abandon(p);
```
MichaReiser, line 1523: We're not allowed to return `Absent` here because we parsed a token on line 1518.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1457,8 +1536,18 @@ pub fn unary_expr(p: &mut Parser) -> Option<CompletedMarker> {
...
 			}
 		}
-		return Some(m.complete(p, JS_UNARY_EXPRESSION));
+		return Present(m.complete(p, JS_UNARY_EXPRESSION));
 	}
 
 	postfix_expr(p)
 }
+
+pub(super) fn is_at_name(p: &Parser) -> bool {
```
MichaReiser, line 1545: This is the same as `is_at_identifier_name`?

crates/rslint_parser/src/syntax/object.rs
```diff
@@ -172,7 +173,7 @@ fn parse_object_member(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
 					// let b = { a: true }
 					// If the member name was a literal OR we're at a colon
 					p.expect_required(T![:]);
-					if expr_or_assignment(p).is_none() {
+					if parse_expr_or_assignment(p).is_absent() {
```
MichaReiser, line 176: `or_missing_with_error`

crates/rslint_parser/src/syntax/pattern.rs
```diff
@@ -30,7 +30,7 @@ pub(crate) trait ParseWithDefaultPattern {
 				pattern.precede_or_missing_with_error(p, Self::expected_pattern_error);
 			p.bump_any(); // eat the = token
 
-			if expr_or_assignment(p).is_none() {
+			if parse_expr_or_assignment(p).is_absent() {
```
MichaReiser, line 33: `or_missing_with_error`

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -348,7 +349,8 @@ pub fn parse_throw_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
 		p.error(err);
 		p.missing();
 	} else {
-		p.expr_with_semi_recovery(false);
+		// TODO: review this part when error recovery is fixed
```
MichaReiser, line 352: What part of error recovery must be fixed?

I think we can just mark it as `missing` if the recovery failed.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -460,7 +462,7 @@ pub fn parse_return_statement(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
 	p.bump_any(); // return keyword
 	if !p.has_linebreak_before_n(0) && p.at_ts(STARTS_EXPR) {
 		// TODO: review this part and make sure it plays well with the new recovery logic
-		p.expr_with_semi_recovery(false);
+		p.expr_with_semi_recovery(false).ok();
```
MichaReiser, line 465: We can probably just mark it as `missing` if the recovery failed`.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -634,7 +636,7 @@ pub fn parenthesized_expression(p: &mut Parser) {
 
 	let pos = p.token_pos();
 	// Remove the pos check once `expr` has been converted to `ParsedSyntax`
-	if expr(p).is_none() && pos == p.token_pos() {
+	if parse_expression(p).is_absent() && pos == p.token_pos() {
```
MichaReiser, line 639: Can we remove the `pos` hack now and just use `parse_expression(p).or_missing_with_error`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1128,7 +1130,7 @@ fn parse_for_head(p: &mut Parser) -> SyntaxKind {
 			return parse_for_of_or_in_head(p);
 		}
 
-		if init_expr.is_none() && p.token_pos() == checkpoint.token_pos {
+		if init_expr.is_absent() && p.token_pos() == checkpoint.token_pos {
```
MichaReiser, line 1133: Can we remov the `token_pos` hack now and just use `p.or_missing_with_error`

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1167,15 +1169,15 @@ fn parse_for_of_or_in_head(p: &mut Parser) -> SyntaxKind {
 
 	if is_in {
 		p.bump_any();
-		if expr(p).is_none() {
+		if parse_expression(p).is_absent() {
```
MichaReiser, line 1172: `or_missing_with_error`

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1167,15 +1169,15 @@ fn parse_for_of_or_in_head(p: &mut Parser) -> SyntaxKind {
...
 			p.missing();
 		}
 
 		JS_FOR_IN_STATEMENT
 	} else {
 		p.bump_remap(T![of]);
 
-		if expr_or_assignment(p).is_none() {
+		if parse_expr_or_assignment(p).is_absent() {
```
MichaReiser, line 1180: `or_missing_with_error`

crates/rslint_parser/src/parse.rs
```diff
@@ -237,7 +237,7 @@ pub fn parse_expr(text: &str, file_id: usize) -> Parse<JsAnyExpression> {
 	let (tokens, mut errors) = tokenize(text, file_id);
 	let tok_source = TokenSource::new(text, &tokens);
 	let mut parser = crate::Parser::new(tok_source, file_id, Syntax::default());
-	crate::syntax::expr::expr(&mut parser);
+	crate::syntax::expr::parse_expression(&mut parser).or_missing(&mut parser);
```
MichaReiser, line 240: The `missing` here won't work. Because building the tree will fail on line `245`. I think it should be fine to just use `unwrap` for now (I introduced a `ExpressionSnipped` which fixes this problem).

MichaReiser (changes requested):
Overall looking good. There are some cases where we are still advancing the parser even though we return `Absent`. Would you mind having a look at it.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -128,26 +129,27 @@ pub(crate) fn expr_or_assignment(p: &mut Parser) -> Option<CompletedMarker> {
...
 	}
-	assign_expr_base(p)
+	parse_assign_expr_base(p)
 }
 
-fn assign_expr_base(p: &mut Parser) -> Option<CompletedMarker> {
+fn parse_assign_expr_base(p: &mut Parser) -> ParsedSyntax<CompletedMarker> {
 	if p.state.in_generator && p.at(T![yield]) {
-		return Some(yield_expr(p));
+		return Present(yield_expr(p));
```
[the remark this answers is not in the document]
ematipico: Yeah, it's fine. It's an internal function used only once and there's a check on the `yield` keyword before calling it. There are no risks

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -156,8 +158,11 @@ fn assign_expr_base(p: &mut Parser) -> Option<CompletedMarker> {
 	});
 
 	let checkpoint = guard.checkpoint();
-	let target = conditional_expr(&mut *guard)?;
-	assign_expr_recursive(&mut *guard, target, checkpoint)
+	if let Present(target) = parse_conditional_expr(&mut *guard) {
+		parse_assign_expr_recursive(&mut *guard, target, checkpoint)
+	} else {
+		Absent
+	}
```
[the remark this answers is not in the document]
ematipico: This suggestion is not correct. The `mapper` has to return `CompletedMarker` or `ConditionalSyntax` because `.map` already applies `Present`.. and `parse_assign_expr_recursive` already returns `ParsedSyntax`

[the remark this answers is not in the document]
MichaReiser: Oh, you can use `and_then` in that case

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -328,12 +344,12 @@ fn binary_or_logical_expression_recursive(
             .primary(p.cur_tok().range, "But this operator was encountered instead");
 
 		p.error(err);
-		None
+		Absent
```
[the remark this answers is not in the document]
ematipico: I think the whole logic is wrong and it needs to be reviewed. Not sure if we have to do it now or not. 
Inside the tests we have this errored test `foo + * 2` and generate the following AST/tree:

```
  0: JS_BINARY_EXPRESSION@11..21
    0: JS_IDENTIFIER_EXPRESSION@11..16
      0: JS_REFERENCE_IDENTIFIER@11..16
        0: IDENT@11..16 "foo" [Whitespace("\n")] [Whitespace(" ")]
    1: PLUS@16..18 "+" [] [Whitespace(" ")]
    2: JS_BINARY_EXPRESSION@18..21
      0: (empty)
      1: STAR@18..20 "*" [] [Whitespace(" ")]
      2: JS_NUMBER_LITERAL_EXPRESSION@20..21
        0: JS_NUMBER_LITERAL@20..21 "2" [] []
```

```
left: JsIdentifierExpression {
    name: JsReferenceIdentifier {
        value_token: IDENT@11..16 "foo" [Whitespace("\n")] [Whitespace(" ")],
    },
},
operator: PLUS@16..18 "+" [] [Whitespace(" ")],
right: JsBinaryExpression {
    left: JsNumberLiteralExpression {
        value_token: JS_NUMBER_LITERAL@20..21 "2" [] [],
    },
    operator: STAR@18..20 "*" [] [Whitespace(" ")],
    right: missing (required),
},

```

As you can see they are out of sync. I also think we should mark a missing right hand.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -549,7 +582,7 @@ pub fn subscripts(p: &mut Parser, mut lhs: CompletedMarker, no_call: bool) -> Co
 // foo?.for
 // foo?.bar
 // foo.#bar
-pub fn static_member_expression(
+pub fn parse_static_member_expression(
```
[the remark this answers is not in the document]
ematipico: Good catch!  Yes this should return `ParsedSyntax` too

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -842,7 +874,7 @@ pub fn paren_or_arrow_expr(p: &mut Parser, can_be_arrow: bool) -> CompletedMarke
 	async_missing_marker.undo(p);
 	type_params_missing_marker.undo(p);
 
-	if let Some(params) = params_marker {
+	if let Some(NodeMarker(params)) = params_marker {
```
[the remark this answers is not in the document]
ematipico: > GitHub should really allow you to comment on unchanged lines:

If you press "." while opening a PR, github will enter in editor mode and it allows you to see the **whole** file with its differences. And you can add comments in any part of the file :)

This is really handy for big PRs. Give it a try!

> Do we need to add a missing marker in the case the parenthesized_expression is empty on line 892?

I really don't know to be honest. We don't have test error cases so it's hard to tell for me

crates/rslint_parser/src/syntax/expr.rs
ematipico: We're not ready yet. I will look into it later

crates/rslint_parser/src/syntax/stmt.rs
ematipico: The missing marker is already created inside function `expr_with_semi_recovery`. This will create an additional empty slot that won't match the AST

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -169,11 +174,18 @@ fn assign_expr_base(p: &mut Parser) -> Option<CompletedMarker> {
 // [,,,foo,bar] = baz;
 // ({ bar, baz } = {});
 // ({ bar: [baz = "baz"], foo = "foo", ...rest } = {});
-fn assign_expr_recursive(
+
+// test_err assign_expr_right
+// (foo = );
+
+// TODO: review the left missing expression, the green tree might be better
```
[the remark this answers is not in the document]
ematipico: Check the test `assign_expr_left.rast`. Even though we have a correct `JsParenthesizedExpression`, we don't actually arrive at the end of the expression. Instead we mark `=` as unknown expression. 

Not sure if it's possible, but I think we should have a `JsAssignmentExpression` instead where the left hand is missing. Do you think it's possible to achieve that?

crates/rslint_parser/src/syntax/assignment.rs
```diff
@@ -90,11 +92,11 @@ pub(crate) fn parse_assignment(
 	// TODO remove the rewind inside of the error handle once the `unary_expr` returns a ParsedSyntax
 	let assignment_expression = match expr_kind {
 		AssignmentExprPrecedence::Unary => unary_expr(p),
```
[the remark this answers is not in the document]
ematipico: `unary_expr` returns a `JsUnknownExpression` in case of error. This will result into a panic because `JsUnknownExpression` can't be casted in `JsAnyArrayAssignmentPatternElement`.

How do you think we should proceed?

crates/rslint_parser/src/syntax/assignment.rs
```diff
@@ -90,11 +92,11 @@ pub(crate) fn parse_assignment(
 	// TODO remove the rewind inside of the error handle once the `unary_expr` returns a ParsedSyntax
```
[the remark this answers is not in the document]
ematipico: Not yet

crates/rslint_parser/src/parser.rs
```diff
@@ -494,29 +494,21 @@ impl<'t> Parser<'t> {
...
-
-		func(self)
+		func(self).or_recover(
+			self,
+			&ParseRecovery::new(SyntaxKind::JS_UNKNOWN_EXPRESSION, token_set![T![;]]),
+			|p, range| {
+				p.err_builder("expected an expression, but found `;` instead")
+					.primary(range, "")
+			},
+		)
```
MichaReiser, line 511: This is no longer the same behaviour as before and the message no longer matches. The old implementation crated a diagnostic only in the case if the parser currently is at a semicolon. The new recovery logic creates a diagnostic whenever the expression is absent and only recovers if there's a semicolon (but e.g. not if we're at the start of a new statement, e.g. this will happily eat over the function in 

```
a + 
function test() {}
```

which is not what we want.

we should revisit this logic and improve the error message

crates/rslint_parser/src/syntax/expr.rs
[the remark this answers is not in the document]
MichaReiser: We should improve the comment here and explain that.

crates/rslint_parser/src/syntax/expr.rs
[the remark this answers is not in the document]
MichaReiser: I would argue we should look into this now since this PR is trying to "fix" the expr parsing or we should at least add a todo

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -624,55 +639,49 @@ pub fn computed_member_expression(
 	}
 
 	p.expect_required(T!['[']);
-	if expr(p).is_none() {
-		p.missing();
-	}
+	parse_expression(p).or_missing(p);
```
MichaReiser, line 642:
```suggestion
	parse_expression(p).or_missing_with_error(p);
```

MichaReiser:
I'll look into the remaining issues so that we can then merge the changes

PR #1881, opened 2021-12-16
Feature/tree shape validation

MichaReiser:
## Motivation

Writing a correct parser is difficult to work and we could do better to remove some of the burdens from parser authors by moving them into the tree builder.  See #1867 for an outline of the discussed approach.

### empty slots
Our current infrastructure requires the parser authors to add the exact right amount of empty slots for each parsed node. This introduces a lot of branching (`ParsedSyntax` helped a bit) and sometimes even requires to preemptively add empty slots to then later undo them again if a node turned out to be of a different type after all (e.g. a class `getter` member has no slot for the `async` token but a method does). 

### syntax errors

Our parse methods return `ConditionalSyntax` if it may parse a node with a syntax error (that isn't a missing required child). The problem with a `ConditionalSyntax` is that it often invalidates the parent as well, resulting in a lot of `is_valid = parse_child` branching.

## This PR 

This PR proposes an approach to reduce the complexity in the parser at a runtime cost in the tree sink step. It introduces a new step in the `TreeBuilder` that uses information from the grammar to:

* Insert empty slots for children that are not present in the source code (and only if the node isn't an `UNKNOWN` node)
* Convert nodes to `Unknow*` when the parsed children don't match the children defined in the grammar.

To achieve this, the following changes were necessary:

* Introduces a new `SyntaxKind` trait that provides methods to convert a kind to an `UnknownKind`.
* Introduces a new `UNKNOWN`  kind which is used when there's no more specific `UNKNOWN_` kind (e.g. `UNKNOWN_STATEMENT)`. See #1867 for more details. 
* Code gen a `SyntaxFactory` that validates if a parsed node matches the AST schema and inserts missing empty slots.  
* Return a cache entry from the `NodeCache` implementation to allow the `TreeBuilder` to customize how a node should be built (call into `SyntaxFactory::make_syntax`)
* Introduces a `RawSyntaxNode` which is a thin wrapper around `GreenNode`. The main motivation is to hide away `GreenNode`s but give other libraries the option to customize the node construction in places where we deliberately allow it (using `SyntaxNode` would be too expensive)

This PR does not convert nodes with missing required children to 'Unknown'. We can do so in a different PR if we decide to go for that approach. 

This PR further includes some renamings and cleanup work:
* Rename `rowan::SyntaxKind` to `RawSyntaxKind`.
* Rename `rslint_parser:SyntaxKind` to `JsSyntaxKind`
* Remove `p.eat_optional`, `p.expect_required` as they are no longer required (use `p.eat` and `p.expect)
* Removes `ConditionalSyntax` as this is no longer required
* Remove generic parameter from `ParsedSyntax`

## How to Review

I would start reviewing the changes in the parsing logic to get a glimpse of what was necessary before and is now taken care of by the parser infrastructure (files in the `rslint_parser/syntax` directory). 

I would then take a look at the `TreeBuilder` and the `JsSyntaxFactory` to understand the newly built infrastructure. 

## Test

I manually verified the RAST updates. Most changes are because:

* We no longer insert `empty` slots if a node turns out to be an `Unknown` node
* Changes in the grammar to make our AST structure fit
* now correctly invalidated parent (or nodes) in case of a syntax error

### Cache Hits

~2 up to 4 times increase because it now takes into account all nodes that have up to 3 actual children, ignoring empty slots. 

I experimented with increasing the limit but going over 3 only brings minimal cache improvements but comes at a significant performance overhead.

* jquery.min.js: 9'158 -> 27'469
* vue.global.prod.js: 14'470 -> 45'100
* react.production.min.js: 966 -> 1'369
* react-dom.production.min.js: 13'962 -> 35'563
* target/compiler.js: 59'383 -> 104'577
* target/dojo.js: 4'421 -> 4'257 (Our parser generates incorrect sequence expressions that are now turned into unknown nodes, and thus, are not cached)
* target/d3.min.js: 29'198 -> 102'956
* target/pixi.min.js: 36'265 -> 119'888
* three.min.js: 42'295 -> 145'365
* tex-chtml-full.js: 60'844 -> 270'134

### Performance

Validating the AST schema against the actual parsed nodes adds an additional `O(N)` traversal in the tree sink step. The initial version was about 5-10% slower because of that and because it now cached more nodes. That's why I spent some time improving the cache performance by avoiding re-computing the hash of nodes (which requires re-computing the hash of the whole sub-tree) during resizes of the hash map.

The result is that the performance has improved 
 

* jquery.min.js
	* main: 
		* fastest: total: 15.093ms (tokenization: 2.2114ms, parsing: 6.4851ms, tree_sink: 6.3965ms)
		* slowest: total: 16.1349ms (tokenization: 2.3131ms, parsing: 6.2157ms, tree_sink: 7.6061ms)
	* this branch:
		* fastest: total: 14.3275ms (tokenization: 2.4944ms, parsing: 5.5857ms, tree_sink: 6.2474ms)
		* slowest: total: 15.3482ms (tokenization: 2.3973ms, parsing: 6.8029ms, tree_sink: 6.148ms)
* vue.global.prod.js
	* main
		* fastest: total: 27.75ms (tokenization: 3.5263ms, parsing: 12.9922ms, tree_sink: 11.2315ms)
		* slowest: total: 31.6103ms (tokenization: 3.9298ms, parsing: 14.583ms, tree_sink: 13.0975ms)
	* this branch
		* fastest: total: 26.2003ms (tokenization: 3.4204ms, parsing: 12.9212ms, tree_sink: 9.8587ms)
		* slowest: total: 27.9678ms (tokenization: 3.5374ms, parsing: 13.1475ms, tree_sink: 11.2829ms)
* react.production.min.js
	* main
		* fastest: total: 1.1663ms (tokenization: 223.9µs, parsing: 468.4µs, tree_sink: 474µs)
		* slowest: total: 1.4846ms (tokenization: 225.3µs, parsing: 642µs, tree_sink: 617.3µs)
	* this branch
		* fastest: total: 1.2379ms (tokenization: 230.7µs, parsing: 579.2µs, tree_sink: 428µs)
		* slowest: total: 1.4386ms (tokenization: 303.9µs, parsing: 594.6µs, tree_sink: 540.1µs)
* react-dom.production.min.js
	* main 
		* fastest: total: 21.4139ms (tokenization: 3.3816ms, parsing: 8.7606ms, tree_sink: 9.2717ms)
		* slowest: total: 24.9795ms (tokenization: 3.9028ms, parsing: 10.3919ms, tree_sink: 10.6848ms)
	* this branch
		* fastest: total: 18.3145ms (tokenization: 2.9719ms, parsing: 7.6095ms, tree_sink: 7.7331ms)
		* slowest: total: 18.7549ms (tokenization: 3.0384ms, parsing: 8.0506ms, tree_sink: 7.6659ms)
* compiler.js
	* main
		* fastest: total: 87.5119ms (tokenization: 15.3712ms, parsing: 28.866ms, tree_sink: 43.2747ms)
		* slowest: total: 92.8007ms (tokenization: 15.2893ms, parsing: 33.4572ms, tree_sink: 44.0542ms)
	* this branch
		* fastest: total: 75.7518ms (tokenization: 14.49ms, parsing: 26.4512ms, tree_sink: 34.8106ms)
		* slowest: total: 79.4099ms (tokenization: 17.2467ms, parsing: 28.7847ms, tree_sink: 33.3785ms)
* dojo.js
	* main
		* fastest: total: 4.5574ms (tokenization: 1.0468ms, parsing: 1.4499ms, tree_sink: 2.0607ms)
		* slowest: total: 5.0395ms (tokenization: 1.2505ms, parsing: 1.6681ms, tree_sink: 2.1209ms)
	* this branch
		* fastest: total: 4.1652ms (tokenization: 921.1µs, parsing: 1.4257ms, tree_sink: 1.8184ms)
		* slowest: total: 4.6712ms (tokenization: 1.237ms, parsing: 1.5797ms, tree_sink: 1.8545ms)
* d3.min.js
	* main
		* fastest: total: 52.4153ms (tokenization: 7.3116ms, parsing: 22.2979ms, tree_sink: 22.8058ms)
		* slowest: total: 59.7842ms (tokenization: 7.9862ms,
		parsing: 27.5101ms, tree_sink: 24.2879ms)
	* this branch
		* fastest: total: 48.3781ms (tokenization: 7.2178ms, parsing: 21.6863ms, tree_sink: 19.474ms)
		* slowest: total: 56.447ms (tokenization: 7.6625ms, parsing: 27.1659ms, tree_sink: 21.6186ms)
* pixi.min.js
	* main 
		* fastest: total: 60.6618ms (tokenization: 8.6254ms, parsing: 25.6991ms, tree_sink: 26.3373ms)
		* slowest: total: 70.3036ms (tokenization: 9.0011ms, parsing: 33.1792ms, tree_sink: 28.1233ms)
	* this branch
		* fastest: total: 56.2068ms (tokenization: 8.5876ms, parsing: 24.0831ms, tree_sink: 23.5361ms)
		* slowest: total: 62.4494ms (tokenization: 8.617ms, parsing: 28.6664ms, tree_sink: 25.166ms)
* three.min.js
	* main
		* fastest: total: 73.4015ms (tokenization: 10.5409ms, parsing: 30.5531ms, tree_sink: 32.3075ms)
		* slowest: total: 80.3063ms (tokenization: 14.6462ms, parsing: 34.5074ms, tree_sink: 31.1527ms)
	* this branch
		* fastest: total: 68.1116ms (tokenization: 10.6749ms, parsing: 30.2379ms, tree_sink: 27.1988ms)
		* slowest: total: 72.1357ms (tokenization: 9.7205ms, parsing: 34.9193ms, tree_sink: 27.4959ms)
tex-chtml-full.js
	* main 
		* fastest: total: 128.8105ms (tokenization: 19.4729ms, parsing: 51.6658ms, tree_sink: 57.6718ms)
		* slowest: total: 134.8473ms (tokenization: 20.9036ms, parsing: 55.8733ms, tree_sink: 58.0704ms)
	* this branch
		* fastest: total: 123.1956ms (tokenization: 18.5511ms, parsing: 53.8234ms, tree_sink: 50.8211ms)
		* slowest: total: 126.1946ms (tokenization: 18.6072ms, parsing: 56.1538ms, tree_sink: 51.4336ms)

	> AMD Ryzen 5800x, fastest/slowest out of 5 runs

## Next Steps

* Finish #1724 . This should be straightforward because the schema validation guarantees that all children are in the right slots
* Replace `ERROR` with `UNKNOWN`. Should be a straightforward search & replace 
* Decide if we want to convert a node into an `UNKNOWN` node if a required child element is missing (see #1867  discussion). Straightforward to make the codegen changes but will require a bit of work to route the `SourcePrecense::Absent` flag from the parser to the `GreenToken`.


MichaReiser:
Feel free to have a look at the changes. I still need to address some regressions, re-add some commented out tests and write some more documentation. But I would appreciate any feedback you may have.

crates/rome_rowan/src/green.rs
```diff
@@ -10,10 +10,11 @@ pub(crate) use self::{
 };
 
 pub use self::node_cache::NodeCache;
+pub(crate) use self::node_cache::NodeCacheNodeEntryMut;
 
 /// SyntaxKind is a type tag for each token or node.
```
xunilrj, line 15:
```rust
/// RawSyntaxKind is a type tag for each token or node.
```

crates/rome_rowan/src/green/node_cache.rs
```diff
@@ -15,7 +15,15 @@ use super::token::GreenTokenTrivia;
 type HashMap<K, V> = hashbrown::HashMap<K, V, BuildHasherDefault<FxHasher>>;
 
 #[derive(Debug)]
-struct NoHash<T>(T);
+struct NoHashToken(GreenToken);
+
+#[derive(Debug)]
+struct NoHashNode {
```
xunilrj, line 21: ```PreCalculatedHashNode``` makes more sense?

crates/rome_rowan/src/raw_language.rs
```diff
@@ -0,0 +1,169 @@
...
+	UNKNOWN = 7,
+	FOR_KW = 8,
+	L_PAREN_TOKEN = 9,
+	SEMICOLON_TOKEN = 10,
+	R_PAREN_TOKEN = 11,
+	EQUAL_TOKEN = 12,
+	LET_TOKEN = 13,
+	CONDITION = 14,
+	PLUS_TOKEN = 15,
+	__LAST,
```
xunilrj, line 37: We can use to avoid having marker variants:

https://github.com/XX/variant_count  
https://github.com/Folyd/variant-counter  
https://github.com/magiclen/enum-ordinalize

crates/rome_rowan/src/raw_language.rs
```diff
@@ -0,0 +1,169 @@
...
+	fn to_unknown(&self) -> Self {
+		RawLanguageKind::UNKNOWN
+	}
+
+	fn to_raw(&self) -> RawSyntaxKind {
+		RawSyntaxKind(*self as u16)
+	}
+
+	#[allow(unsafe_code)]
+	fn from_raw(raw: RawSyntaxKind) -> Self {
```
xunilrj, line 54: Why not ```Option<Self>``` or ```Result<Self, ...>``` instead of panicking?

crates/rome_rowan/src/raw_language.rs
```diff
@@ -0,0 +1,169 @@
...
+
+				slots.into_node(kind, children)
+			}
+			_ => unreachable!("{:?} is not a node kind", kind),
+		}
+	}
+}
+
+#[doc(hidden)]
+pub type RawSyntaxTreeBuilder = TreeBuilder<'static, RawLanguage, RawLanguageSyntaxFactory>;
```
xunilrj, line 169: Do we need to "lock" ```'static``` here? if ```RawSyntaxTreeBuilder``` has just one generic parameter, the compiler will be able to deduce it most times.


```rust
#[doc(hidden)]
pub type RawSyntaxTreeBuilder<'a> = TreeBuilder<'a, RawLanguage, RawLanguageSyntaxFactory>;

```

xunilrj:
> The result is that the performance has improved

One nice option to display total time is (https://github.com/sharkdp/hyperfine)

```
> cargo install hyperfine
> cargo build -p xtask --release
> hyperfine './target/release/xtask.exe coverage-libs --filter=three'
```

I have the impression that this also "fix" the problem with ```println!``` being super slow.

xunilrj (approved):

crates/rome_rowan/src/green/node_cache.rs
MichaReiser: I kept the existing name (only added the node prefix) because the main intention is that this node doesn't implement `Hash`

[```
// So here we use *raw* API of hashbrown and provide the hashes manually,
// instead of going via a `Hash` impl. Our manual `Hash` and the
// `#[derive(Hash)]` are actually different! At some point we had a fun bug,
// where we accidentally mixed the two hashes, which made the cache much less
// efficient.
//
// To fix that, we additionally wrap the data in `NoHash` wrapper, to make sure
// we don't accidentally use the wrong hash!

```](https://github.com/rome/tools/blob/37304ba5dc46c03ff30f67f89ac0633b80aae8e4/crates/rome_rowan/src/green/node_cache.rs#L39-L46)

Let me rename it to `CachedNode`.

File: crates/rome_rowan/src/raw_language.rs (hunk 3 above)
Comment (MichaReiser) [reply crates/rome_rowan/src/raw_language.rs:37]: I want to wait `std::mem::variant_count` to avoid adding new dependencies.

File: crates/rome_rowan/src/raw_language.rs (hunk 4 above)
Comment (MichaReiser) [reply crates/rome_rowan/src/raw_language.rs:54]: This would make handling `RawSyntaxKind` rather painful and callers often can't do anything else than panic themselves.

For example, the `from_raw` is used whenever you call `SyntaxNode::kind`. Returning `Result` or `Option` there would make the API difficult to use and, in my view, bring little value because most our APIs require you to pass in a typed syntax node that gets then internally converted to a `RawSyntaxKind` (hence, converting it back should be safe). 

Ideally, we could enforce it at compile time that it's not possible to construct a `RawSyntaxKind` except if you go over a `SyntaxKind` implementation.

**Comment by MichaReiser** [comment]
> > The result is that the performance has improved
> 
> One nice option to display total time is (https://github.com/sharkdp/hyperfine)
> 
> ```
> > cargo install hyperfine
> > cargo build -p xtask --release
> > hyperfine './target/release/xtask.exe coverage-libs --filter=three'
> ```
> 
> I have the impression that this also "fix" the problem with `println!` being super slow.

Oh that's nice. Let me undo my changes that run the benchmarks multiple time (and added sleeps).

File: crates/rslint_parser/src/lossless_tree_sink.rs [hunk 6]
```diff
@@ -88,15 +73,29 @@ impl<'a> LosslessTreeSink<'a> {
...
 		if kind == JsSyntaxKind::EOF {
 			self.needs_eof = false;
 		}
 
-		self.do_tokens(kind, len, 1)
+		self.do_tokens(kind, 1)
 	}
 
-	fn do_tokens(&mut self, kind: JsSyntaxKind, len: TextSize, token_count: u8) {
+	fn do_tokens(&mut self, kind: JsSyntaxKind, token_count: u8) {
```
xunilrj, line 84: Nice!

xunilrj (approved):

crates/rome_rowan/src/syntax_factory.rs
```diff
@@ -0,0 +1,275 @@
...
+	inner: Peekable<ParsedChildrenIntoIterator<'a, K>>,
+	missing_count: usize,
+	next_node: bool,
+	separator: K,
+}
+
+impl<'a, K: SyntaxKind> Iterator for SeparatedListWithMissingNodesOrSeparatorSlotsIterator<'a, K> {
+	type Item = Option<RawSyntaxElement<K>>;
+
+	#[cold]
```
ematipico, line 139: Does it make sense to flag a function of a trait as cold? Just wondering

MichaReiser:
> I will probably ask you for a one to one session. There are too many parts of this new code that require lot of questions. I couldn't finish the review and will leave it for now. I just some feedback and will leave more once I will come back

Feel free to put something on my calendar.

crates/rome_rowan/src/api.rs
```diff
@@ -1,33 +1,31 @@
...
 };
 
-pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Hash {
-	type Kind: fmt::Debug + PartialEq;
-
-	fn kind_from_raw(raw: SyntaxKind) -> Self::Kind;
-	fn kind_to_raw(kind: Self::Kind) -> SyntaxKind;
-}
+/// Type tag for each node or token of a language
+pub trait SyntaxKind: fmt::Debug + PartialEq + Copy {
```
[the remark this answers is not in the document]
MichaReiser: So that we can have methods like `to_unknown()` and use these in language-independent code.

crates/rome_rowan/src/green/node_cache.rs
```diff
@@ -186,9 +180,72 @@ impl NodeCache {
 	}
 }
 
+pub(crate) enum NodeCacheNodeEntryMut<'a> {
```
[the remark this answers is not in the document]
MichaReiser: We only know if a node is cacheable when it gets inserted, not when querying the cache. 
I used this naming because it resembles the same enum from the underlying `HashMap` so that it becomes a recognizable pattern.

PR #1882, opened 2021-12-16
test(formatter): add tests to cover the whole syntax

### LINKED ISSUE #1878: 📎 Formatter: add tests to cover the whole syntax by using code coverage
### Description

We should try 100% code coverage. The current plan is to use https://github.com/mozilla/grcov to find missing coverage and add snapshots to cover them.

Getting the code coverage is quite easy, here are a few screen shots by using grcov:
image
image

Once we have everything covered, we should be able to look at all the snapshots and make bigger adjustments for the formatter.

Boshen:
closes #1878

## Summary
I used [grcov](https://github.com/mozilla/) to generate the coverage, and filled in all uncovered lines, it's currently sitting at 92% (increased from 90%). The missing pieces are TypeScript, Unknown* and Holes.

## Test Plan
`cargo test` passes

## What's next
https://github.com/rome/tools/pull/1876#pullrequestreview-833798471
> Now that we have the tests from the old Rome, I think one important thing that we should do is to actually match its formatting style to the new formatter.
>
> Rome's old formatter was pretty stable (exception for comments and some JSX case), so I think it's worth using its formatting style whereas possible.

I'll work on this branch until everything conforms to old Rome.


crates/rome_formatter/src/ts/assignment/array_assignment_pattern.rs
```diff
@@ -30,7 +30,9 @@ impl ToFormatElement for JsAnyArrayAssignmentPatternElement {
 			JsAnyArrayAssignmentPatternElement::JsArrayAssignmentPatternRestElement(
 				array_assignment_pattern_rest_element,
 			) => array_assignment_pattern_rest_element.to_format_element(formatter),
-			JsAnyArrayAssignmentPatternElement::JsArrayHole(_array_hole) => todo!(),
+			JsAnyArrayAssignmentPatternElement::JsArrayHole(array_hole) => {
+				array_hole.to_format_element(formatter)
+			}
```
xunilrj, line 35: silly but we can do: 

```rust
JsAnyArrayAssignmentPatternElement::JsArrayHole(array_hole) => array_hole.to_format_element(formatter)
```

crates/rome_formatter/tests/specs/js/arrow/arrow_function.js
```diff
@@ -0,0 +1,3 @@
+() => {}
+async () => {}
+(foo) => {}
```
xunilrj, line 3: Can we test more than one parameter?

```rust
(foo,bar) => {}
```

crates/rome_formatter/tests/specs/js/binding/array_binding.js
```diff
@@ -0,0 +1,2 @@
+[a="b"]=c
+let [a="b"]=c
```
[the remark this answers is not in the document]
ematipico: This test covers the array bindings, so this test is not needed here (which is already covered somewhere else)

crates/rome_formatter/tests/specs/js/object/getter_setter.js.snap
```diff
@@ -0,0 +1,32 @@
...
+
+}
+let b = 
+{
+	set;
+	foo();
+	{}
+	set;
+	foo(a);
+	{}
```
ematipico, line 30: Found a bug

ematipico (changes requested):

crates/rome_formatter/tests/specs/js/arrow/arrow_function.js
ematipico: Function arguments are already covered inside `function/function_args.js`

crates/rome_formatter/src/ts/assignment/array_assignment_pattern.rs
ematipico: I think this is the formatter 🤔

Boshen:
@ematipico I am in the right direction for this: https://github.com/rome/tools/pull/1882/commits/d6a1c94082d4ea0e090d800b32a04a3c96d22b26

To align with prettier, we collapse new lines appropriately, but I couldn't find a general place to add the code, so I added it to JsStatementList to test it out, can you point me in the right direction on where to collapse new lines?

crates/rome_formatter/tests/specs/js/object/getter_setter.js.snap
Boshen: This is syntax error, I have removed it.

crates/rome_formatter/tests/specs/js/class/class.js.snap
```diff
@@ -47,7 +46,9 @@ x = class aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa extends bbbbbbb
 ---
 # Output
 class Foo extends Boar {
-	constructor() {
+	constructor(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
+	bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,
+	c = d) {
```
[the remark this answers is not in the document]
Boshen: Fixed in 6877d63, I see there are few places that have this pattern, perhaps I can refactor them.

crates/rome_formatter/src/ts/assignment/array_assignment_pattern.rs
Boshen: Yep it's the formatter :-/

ematipico:
> @ematipico Am I in the right direction for this: [d6a1c94](https://github.com/rome/tools/commit/d6a1c94082d4ea0e090d800b32a04a3c96d22b26)
> 
> To align with prettier, we collapse new lines appropriately, but I couldn't find a general place to add the code, so I added it to JsStatementList to test it out, can you point me in the right direction on where to collapse new lines?

I think this fix requires its own PR and its own issue, mainly because this change involves `rome_rowan` as well. Myself and Micha were discussing about having a `new_line_trivia` but this still up in the air and not sure if this is the right direction. That's why it's worth a small discussion first.

Generally, I think this logic should live inside the formatter (or even inside the printer) BUT the single nodes should not be aware of this. We might also check what old Rome was doing.

Boshen:
I see, I'll drop the commit and clean up this PR and leave the formatter as is for now.

Boshen:
Fixed some spacing issues, ready to review and merge

crates/rome_formatter/tests/specs/js/expression/new_expression.js.snap
```diff
@@ -6,11 +6,11 @@ expression: new_expression.js
...
 new b(x)
-new c(dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd)
+new c(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,bbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,cccccccccccccccccccccccccccccc)
 
 ---
 # Output
 new a();
 new b(x);
-new c(dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd);
+new c(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, bbbbbbbbbbbbbbbbbbbbbbbbbbbbbb, cccccccccccccccccccccccccccccc);
```
Boshen, line 15: This needs special treatment, adding normal breaking logic will break things, for example passing arrow functions.

crates/rome_formatter/tests/specs/js/object/getter_setter.js.snap
```diff
@@ -0,0 +1,18 @@
...
+	get foo() {}
+}
+let b = {
+	set foo(a) {}
+}
+
+---
+# Output
+let a = {get foo() {}};
+let b = {set foo(a) {}};
```
MichaReiser, line 17: Can we add spaces after the `{` and right before the `}` of the object expressoin?

crates/rome_formatter/tests/specs/js/script.js.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: crates/rome_formatter/tests/spec_test.rs
+expression: script.js
+
+---
+# Input
+#!/usr/bin/env node
```
MichaReiser, line 7: It's impossible to tell if this is indeed working. Can we add a `"use strict"` directive as well and some additional statements?

```
"use strict"

let a =    "b";
```

MichaReiser:
Nice, thanks for adding all these new tests. I've found two small improvements, and we should then be ready to merge this.

crates/rome_formatter/tests/specs/js/script.js.snap
Boshen: done in 89315cd

crates/rome_formatter/tests/specs/js/object/getter_setter.js.snap
Boshen: Good catch! Fixed in 31fceb9. We only need special treatment for single line object expression :-)

crates/rome_formatter/tests/specs/js/object/property_key.js.snap
```diff
@@ -15,5 +15,8 @@ const foo = {
 
 ---
 # Output
-const foo = {"foo-bar": true, "bar": {"lorem_ispsum": {"lorem-ipsum": true}}};
+const foo = {
+	"foo-bar": true,
+	"bar": { "lorem_ispsum": { "lorem-ipsum": true } },,
```
Boshen, line 20: This extra comma comes from the original input + `if_group_breaks(token(","))`.
I think this issue should be addressed in another place.

Boshen:
@MichaReiser Ready to merge.

There are some issues and API improves that can be triaged, I'll add them to the "Formatter API improvements" umbrella.

crates/rome_formatter/tests/specs/js/script.js.snap
```diff
@@ -0,0 +1,15 @@
...
+"use strict"
+var express = require("express")
+
+---
+# Output
+#!/usr/bin/env node
+var express = require("express");
```
MichaReiser, line 14: Whoops, the `use strict` is gone...

crates/rome_formatter/tests/specs/js/object/property_key.js.snap
MichaReiser: `object_expression` uses `format_separated` which already inserts the trailing `,` if it breaks across multiple lines. 

I think the right thing to do here is to remove it from `format_separated` and leave it up to the `format_separated` call sites to decide whatever they want to have a trailing separator (some of which may not even support it, for example, a variable declaration statement).

MichaReiser: These are the relevant lines

https://github.com/rome/tools/blob/497abd0a74a0f1dd656b0e357e6d2ca686db919d/crates/rome_formatter/src/formatter.rs#L141-L143

 I think we can just change it to 

```
		for (index, element) in list.elements().enumerate() {
			let node = self.format_node(element.node()?)?;
			if let Some(separator) = element.trailing_separator()? {
				let formatted_separator = self.format_token(&separator)?;
				if index == list.len() - 1 {
					result.push(node)
				} else {
					result.push(format_elements![node, formatted_separator]);
				}
			} else {
				result.push(node);
			}
		}
```

Boshen: Fixed in 5c1b0e7, so nice to see all the extras trailing commas gone!

crates/rome_formatter/tests/specs/js/script.js.snap
Boshen: Fixed in 3f028d4, JsDirectiveList was never rendered 🤣

crates/rome_formatter/src/ts/directive.rs
```diff
@@ -0,0 +1,21 @@
...
+		let semicolon = if let Some(token) = self.semicolon_token() {
+			formatter.format_token(&token)?
+		} else {
+			empty_element()
+		};
+
+		Ok(format_elements![
+			formatter.format_token(&self.value_token()?)?,
+			semicolon,
+			hard_line_break()
```
MichaReiser, line 18: Won't this add two hard-line breaks if joined together with what we have in `scripts.rs`. I just had a look at expression statement and we don't add any line break there.

MichaReiser (approved):
Amazing, thank you so much for working on this

crates/rome_formatter/tests/specs/js/array/spaces.js.snap
```diff
@@ -13,7 +13,7 @@ let e = [2,2,1,3];
 ---
 # Output
 let a = [];
-let b = [, ];
+let b = [,];
```
Boshen, line 16: @MichaReiser Super annoying edge case here, the parser should give 3 `JsArrayHole`s for the input `let b = [,,];`, but we were given 2:
```
                  1: JS_ARRAY_EXPRESSION@8..12
                    0: L_BRACK@8..9 "[" [] []
                    1: JS_ARRAY_ELEMENT_LIST@9..11
                      0: JS_ARRAY_HOLE@9..9
                      1: COMMA@9..10 "," [] []
                      2: JS_ARRAY_HOLE@10..10
                      3: COMMA@10..11 "," [] []
                    2: R_BRACK@11..12 "]" [] []
```

MichaReiser: Two holes should be correct for `let b = [,,];` 

```
let b = [,,];
undefined
b
Array [ <2 empty slots> ]
```

array holes are inserted if there was no element between two commas. The last comma is a *trailing comma, and thus, doesn't create an array hole.

crates/rome_formatter/src/ts/directive.rs
```diff
@@ -0,0 +1,21 @@
...
+	ToFormatElement,
+};
+use rslint_parser::ast::JsDirective;
+
+impl ToFormatElement for JsDirective {
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		let semicolon = if let Some(token) = self.semicolon_token() {
+			formatter.format_token(&token)?
+		} else {
+			empty_element()
```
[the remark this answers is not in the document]
Boshen: Both issues fixed in d3995f1

crates/rome_formatter/tests/specs/js/array/spaces.js.snap
Boshen: Oh I was confused with who prettier is handling with "manual" vs "auto" inserted trailing commas ...

The formatting here is correct then, except we are not handling the manual case for now

Boshen:
@MichaReiser Ready to merge again 😁 

I think I'll start comparing prettier outputs and jot down all the things we should add to the formatter in the umbrella issue.

PR #1884, opened 2021-12-21
feature(rslint_parser): Use offsets to retrieve children in facade

### LINKED ISSUE #1724: ☂️ Use fixed slots for children

MichaReiser:
## Summary

This PR changes the AST format to lookup child nodes by their fixed slot index instead of iterating over all children and finding the first element with the right kind.

This fixes a few issues where manual implementations were incorrect or missing. It will further allow us to model some nodes more to our liking.

Closes #1724 

## Test Plan

Verified the changes in the snapshot tests


crates/rslint_parser/src/ast.rs
```diff
@@ -359,66 +336,56 @@ impl std::fmt::Display for SyntaxError {
...
 	use crate::{SyntaxError, SyntaxResult};
+	use rome_rowan::SyntaxSlot;
 	use std::fmt::{Debug, Formatter};
 
-	pub(super) fn node<N: AstNode>(parent: &SyntaxNode) -> Option<N> {
-		parent.children().find_map(N::cast)
+	pub(super) fn node<N: AstNode>(parent: &SyntaxNode, slot_index: usize) -> Option<N> {
+		match parent.slots().nth(slot_index)? {
+			SyntaxSlot::Empty => None,
+			SyntaxSlot::Node(node) => Some(node.to()),
```
MichaReiser, line 349: Calling `to` or not matching on the `token` kind in the token methods is safe because the `SyntaxFactory` guarantees that nodes are well-formed. 

The only cases where this can throw is if users manipulate `SyntaxNode`s direct which is unsafe territory. 

However, we'll need to uphold the same guarantees when designing our mutation API.

crates/rslint_parser/test_data/inline/err/binary_expressions_err.rast
```diff
@@ -38,11 +38,11 @@ JsModule {
                 },
                 operator: PLUS@16..18 "+" [] [Whitespace(" ")],
                 right: JsBinaryExpression {
-                    left: JsNumberLiteralExpression {
+                    left: missing (required),
```
MichaReiser, line 41: The expression is `* 2, which is why the new tree is correct.

crates/rslint_parser/test_data/inline/err/import_assertion_err.rast
```diff
@@ -21,7 +21,7 @@ JsModule {
                         JsImportAssertionEntry {
                             key: JS_STRING_LITERAL@28..35 "\"json\"" [] [Whitespace(" ")],
                             colon_token: missing (required),
-                            value_token: JS_STRING_LITERAL@28..35 "\"json\"" [] [Whitespace(" ")],
+                            value_token: missing (required),
```
MichaReiser, line 24: The old facade missed a manual implementation for value which is why it used the key as value if there was no value present.

xunilrj (approved):

PR #1889, opened 2021-12-22
feature(rslint_parser): Synthesized Tokens

MichaReiser:
## Summary
#1867 proposes to convert any node to an `Unknown*` node if any required child is missing, not present in the source code.

This allows typing the field accessors for required children as `T` rather than `Result<T>`, which is a nice ergonomics improvement.

However, this would result in many nodes turning into `Unknown` nodes if only a single token is missing. For example, `if (test) {` turns into an `Unknown` node because the closing `}` is missing.

At this point, the parser can be fairly certain that the user is about to complete the `if` statement, thus, it would be safe to assume the closing `}` is present in the source code.

This PR adds support for synthesized tokens. A synthesized token is a token that isn't present in the original source text (isn't written) but has been inserted by the parser to form a valid tree.

Inserting synthesized punctuation tokens should be safe as most analyzers will just be skipping over them. The only use case where checking for the source presence is relevant right now is the formatter where we don't want to insert not written tokens.

This PR doesn't change the AST facade yet and only inserts a few synthesized nodes. We'll need to play around in which cases it makes sense to insert additional synthesized tokens.

## Test Plan

Verified the RAST changes


crates/rome_rowan/src/green/token.rs
```diff
@@ -61,9 +61,23 @@ impl From<Vec<TriviaPiece>> for GreenTokenTrivia {
 	}
 }
 
+/// Signals whatever a token is present or written in the source text or if it was inserted by the
+/// parser to construct a valid tree.
+#[repr(u8)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub(crate) enum SourcePresence {
```
MichaReiser, line 68: I thought about creating a `GreenTokenFlags` bitflags structure but it felt premature. We only have this one flag for now and nothing is stopping us from migrating it to bitflags in the future (as it's internal).

MichaReiser:
Let's hold back on this for now. I'll post more details in #1867

PR #1892, opened 2021-12-23
newline token kind when lexing

xunilrj:
PS: This PR is outdated. See https://github.com/rome/tools/issues/1901.

We were treating whitespace and newline as ```WHITESPACE```. Which is fine, but it was generating the problem that we were peeking into the ```WHITESPACE``` token text to find new lines. Now we don't need to do this anymore, which is making the parser a lit bit faster.

# Example

Before
```rust
assert_lex! {
	"\n ",
	WHITESPACE:2
}
```

After
```rust
assert_lex! {
	"\n ",
	NEWLINE:1
	WHITESPACE:1
}
```

Or with ```\r\n```

Before 
```rust
assert_lex! {
	" \r\n",
	WHITESPACE:1
	WHITESPACE:2
}
```

After
```rust
assert_lex! {
	" \r\n",
	WHITESPACE:1
	NEWLINE:2
}
```

# Multiple lines

Today, when we find multiple lines we merge them together in one ```WHITESPACE``` token. Now we are splitting them in multiple ```NEWLINE``` tokens, for example:

```rust

#[test]
fn all_whitespace() {
	...
	assert_lex! {
		"\n\n",
		NEWLINE:1
		NEWLINE:1
	}
        ...
}
```

# Performance

Performance gains came from:
1 - we don't iterate the token text for newlines anymore;
2 - we don't create token_offset_pair and tokens_after_linebreaks; and obviously, we don't search them anymore.
2.1 - We pre-calculate token offset and if it is the next non-trivia token after newlines;
3 - We don't copy needlessly tokens when parsing;
4 - Token struct is smaller (12 bytes versus 16 bytes), even though we have "more data" now.
5 - Faster allocator

|lib|CPU Before|CPU After|Memory Before| Memory After|
--- | --- | ---| ---| ---|
|jquery|best 25ms|average 16.381ms|  Max Bytes: 8.71 MB; Total Bytes: 16.65 MB | Max Bytes: 8.57 MB; Total Bytes: 16.65 MB |
|threejs|best 117ms|average 89.533ms| Max Bytes: 35.75 MB; Total Bytes: 562.57 MB | Max Bytes: 35.04 MB; Total Bytes: 562.55 MB |

Before
```
Total Time: 43.6018ms
Total Time: 98.2213ms
Total Time: 2.6924ms
Total Time: 65.8382ms
Total Time: 236.5217ms
Total Time: 10.5943ms
Total Time: 146.1323ms
Total Time: 142.3762ms
Total Time: 175.1149ms
Total Time: 268.3024ms
```

Output is different but the libs follow the same order

After(Windows)
```
[jquery.min.js] Total Time: 16.484ms (tokenization: 1.9288ms, parsing: 7.568ms, tree_sink: 6.9872ms)
[vue.global.prod.js] Total Time: 36.8386ms (tokenization: 4.1165ms, parsing: 19.4657ms, tree_sink: 13.2564ms)
[react.production.min.js] Total Time: 1.2815ms (tokenization: 206.9┬╡s, parsing: 549.2┬╡s, tree_sink: 525.4┬╡s)  
[react-dom.production.min.js] Total Time: 26.9426ms (tokenization: 2.5768ms, parsing: 13.0159ms, tree_sink: 11.3499ms)
[compiler.js] Total Time: 90.265ms (tokenization: 15.4077ms, parsing: 30.0471ms, tree_sink: 44.8102ms)
[dojo.js] Total Time: 4.714ms (tokenization: 791.5┬╡s, parsing: 1.4962ms, tree_sink: 2.4263ms)
[d3.min.js] Total Time: 57.3223ms (tokenization: 6.2511ms, parsing: 26.2195ms, tree_sink: 24.8517ms)
[pixi.min.js] Total Time: 66.207ms (tokenization: 7.7532ms, parsing: 29.5057ms, tree_sink: 28.9481ms)
[three.min.js] Total Time: 77.0848ms (tokenization: 8.9038ms, parsing: 35.8389ms, tree_sink: 32.3421ms)
[tex-chtml-full.js] Total Time: 166.347ms (tokenization: 17.9924ms, parsing: 64.6204ms, tree_sink: 83.7342ms)
[lodash.js] Total Time: 24.9849ms (tokenization: 4.2597ms, parsing: 8.6357ms, tree_sink: 12.0895ms)
```

After (WSL)
```
[jquery.min.js] Total Time: 12.199781ms (tokenization: 1.682611ms, parsing: 4.70618ms, tree_sink: 5.81099ms)     
[vue.global.prod.js] Total Time: 26.576707ms (tokenization: 3.532758ms, parsing: 12.469799ms, tree_sink: 10.57415ms)
[react.production.min.js] Total Time: 838.385µs (tokenization: 137.062µs, parsing: 288.282µs, tree_sink: 413.041µs)
[react-dom.production.min.js] Total Time: 17.357789ms (tokenization: 3.02271ms, parsing: 5.770853ms, tree_sink: 8.564226ms)
[compiler.js] Total Time: 66.397749ms (tokenization: 12.879798ms, parsing: 18.502785ms, tree_sink: 35.015166ms)  
[dojo.js] Total Time: 3.155883ms (tokenization: 571.429µs, parsing: 1.01304ms, tree_sink: 1.571414ms)
[d3.min.js] Total Time: 40.242459ms (tokenization: 4.844913ms, parsing: 15.089494ms, tree_sink: 20.308052ms)     
[pixi.min.js] Total Time: 48.257906ms (tokenization: 7.138972ms, parsing: 17.613233ms, tree_sink: 23.505701ms)   
[three.min.js] Total Time: 60.235828ms (tokenization: 9.597417ms, parsing: 24.153288ms, tree_sink: 26.485123ms)  
[tex-chtml-full.js] Total Time: 103.596197ms (tokenization: 14.693265ms, parsing: 39.13065ms, tree_sink: 49.772282ms)
[lodash.js] Total Time: 15.36596ms (tokenization: 3.387414ms, parsing: 4.313899ms, tree_sink: 7.664647ms) 
```

The difference between Windows and WSL (I think) comes from the allocator. For example, see https://www.reddit.com/r/rust/comments/qnfcdx/very_slow_deallocation_of_boxt_on_windows.

If we use a faster allocator, we can squeeze even more performance:

After (now Windows and WSL have similar numbers)
```
[jquery.min.js] Total Time: 10.182497ms (tokenization: 1.370557ms, parsing: 3.806857ms, tree_sink: 5.005083ms)   
[vue.global.prod.js] Total Time: 18.897428ms (tokenization: 2.309826ms, parsing: 8.105518ms, tree_sink: 8.482084ms)
[react.production.min.js] Total Time: 608.614µs (tokenization: 87.132µs, parsing: 223.124µs, tree_sink: 298.358µs)
[react-dom.production.min.js] Total Time: 15.292781ms (tokenization: 1.727245ms, parsing: 5.567364ms, tree_sink: 
7.998172ms)
[compiler.js] Total Time: 61.509581ms (tokenization: 9.736649ms, parsing: 19.206972ms, tree_sink: 32.56596ms)    
[dojo.js] Total Time: 2.731947ms (tokenization: 512.855µs, parsing: 861.773µs, tree_sink: 1.357319ms)
[d3.min.js] Total Time: 36.161736ms (tokenization: 4.403448ms, parsing: 13.652004ms, tree_sink: 18.106284ms)     
[pixi.min.js] Total Time: 47.889442ms (tokenization: 6.162049ms, parsing: 19.753283ms, tree_sink: 21.97411ms)    
[three.min.js] Total Time: 59.604656ms (tokenization: 6.875095ms, parsing: 24.566829ms, tree_sink: 28.162732ms)  
[tex-chtml-full.js] Total Time: 95.775742ms (tokenization: 13.025206ms, parsing: 34.555772ms, tree_sink: 48.194764ms)
[lodash.js] Total Time: 13.902207ms (tokenization: 2.923941ms, parsing: 4.667398ms, tree_sink: 6.310868ms) 
```

It is very hard to get comparable numbers, but this is how long Chrome/Brave is taking with jquery:

image


crates/rslint_lexer/src/lib.rs
```diff
@@ -124,15 +124,21 @@ impl<'src> Lexer<'src> {
...
 
-	fn consume_newlines(&mut self) {
+	fn consume_newline(&mut self) {
 		while self.current().is_some() {
 			let chr = self.get_unicode_char();
+
+			if chr == '\r' {
+				self.cur += chr.len_utf8();
+				continue;
+			}
```
MichaReiser, line 134: I think we need a special treatment for line breaks, potentially with lookahead because two consequent `\r\r` would currently parse as a single newline but this is supposed to be two line breaks when using `\r` as newline character. 

This code also incorrectly groups `\r\u{2029}` into a single new line.

crates/rslint_lexer/src/lib.rs
```diff
@@ -1250,8 +1250,14 @@ impl<'src> Lexer<'src> {
...
-				self.consume_whitespace();
-				tok!(WHITESPACE, self.cur - start)
+				self.consume_newline();
+				let len = self.cur - start;
+				if len > 0 {
+					tok!(NEWLINE, len)
+				} else {
+					self.consume_whitespace_until_newline();
+					tok!(WHITESPACE, self.cur - start)
+				}
```
MichaReiser, line 1260: NIT: Would it make sense to extract this into a method `consume_newline_or_whitespace` which returns the token kind because the exact same code is repeated below?

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -199,10 +199,7 @@ impl<'a> LosslessTreeSink<'a> {
 				break;
 			}
 
-			let pos: u32 = self.text_pos.into();
-			let pos = pos as usize;
-			let text = &self.text[pos..(pos + token.len)];
-			if break_on_newline && text.chars().any(rslint_lexer::is_linebreak) {
```
MichaReiser, line 205: Is the speed up mainly coming from this change? Or what's the reason that introducing `NEWLINE` improvement performance that much?

crates/rslint_parser/src/lossless_tree_sink.rs
```diff
@@ -212,6 +209,7 @@ impl<'a> LosslessTreeSink<'a> {
 			length += len;
 
 			let current_trivia = match token.kind {
+				NEWLINE => TriviaPiece::Whitespace(token.len),
```
MichaReiser, line 212: Should we introduce a `NewLine` `TriviaPiece`? It seems unclear why the `WHITESPACE` gets split into two trivia when looking at the CST because both pieces are of the same kind. 

Having a `NEWLINE` trivia piece will also be useful inside of the formatter

crates/rslint_parser/src/lossy_tree_sink.rs
```diff
@@ -226,6 +226,7 @@ impl<'a> LossyTreeSink<'a> {
 
 			let current_trivia = match token.kind {
 				SyntaxKind::WHITESPACE => continue,
+				SyntaxKind::NEWLINE => continue,
```
MichaReiser, line 229: NIT:
```suggestion
				SyntaxKind::WHITESPACE | SyntaxKind::NEWLINE => continue,
```

crates/rslint_parser/src/token_source.rs
```diff
@@ -63,11 +62,15 @@ impl<'t> TokenSource<'t> {
...
-				if !has_linebreak && src.chars().any(is_linebreak) {
+				if !has_linebreak && token.kind == rslint_lexer::SyntaxKind::NEWLINE {
 					has_linebreak = true;
+				} else if !has_linebreak && token.kind == rslint_lexer::SyntaxKind::COMMENT {
+					let src = source
+						.get(len.into()..(usize::from(len) + token.len))
+						.expect("src and tokens do not match");
+					if src.chars().any(rslint_lexer::is_linebreak) {
+						has_linebreak = true;
+					}
```
MichaReiser, line 73: Nit: It's a bit difficult to reason about this code. can we remove some of the nesting?

```rust
if token.kind.is_trivia() && !has_line_break {
  has_line_break = match token.kind {
    SyntaxKind::NEWLINE => true,
    SyntaxKind::COMMENT => { ... },
    _ => false
  }
}
```

xtask/src/coverage/mod.rs
```diff
@@ -270,9 +270,13 @@ fn exec_test(mut code: String, strict: bool, module: bool) -> (String, ExecRes)
 
 	let result = std::panic::catch_unwind(|| {
 		if module {
-			parse_module(&code, 0).ok().map(drop)
+			let p = parse_module(&code, 0);
+			// dbg!(p.syntax());
+			p.ok().map(drop)
```
MichaReiser, line 275: Revert?

xtask/src/libs/mod.rs
```diff
@@ -89,20 +90,17 @@ pub fn run(filter: String) {
...
-					#[cfg(feature = "dhat-on")]
 					let stats = dhat::get_stats().unwrap();
 					let tokenizer_timing = timing::start();
-					let (tokens, mut errors) = rslint_parser::tokenize(text.as_str(), 0);
-					let tok_source = rslint_parser::TokenSource::new(text.as_str(), &tokens);
-					println!("\tTime");
-					println!("\t\ttook {:?}", tokenizer_timing.stop());
+					let (tokens, mut diags) = rslint_parser::tokenize(text, 0);
+					let tok_source = rslint_parser::TokenSource::new(text, &tokens);
+					println!("\tTime\n\t\ttook {:?}", tokenizer_timing.stop());
```
MichaReiser, line 103: Oh, I should have merged my `mod` changes... Now we'll have two completely different versions.

I suggest you to remove the `print` statement because they completely screw your measurements. Feel free to copy over the `BenchmarkResult` changes I made in #1881 or I can split it out into its own PR if you want.

xtask/src/libs/mod.rs
```diff
@@ -116,29 +114,47 @@ pub fn run(filter: String) {
...
+						println!("\t{:?}: {}", severity, items.count());
+					}
+
+					// Do all steps again
+					// for a more precisa timing
+
+					let t = timing::start();
+					let _ = rslint_parser::parse_module(text, 0);
+					let dur = t.stop();
+					println!("Total Time: {:?}", dur);
```
MichaReiser, line 146: Does it make sense to print the total time if we run criterion right after?

xtask/src/libs/mod.rs
MichaReiser: Oh, I see how you solved it. That works too.

xtask/Cargo.toml
```diff
@@ -34,6 +34,8 @@ timing = "0.2.3"
 ansi_rgb = "0.2.0"
 dhat = { version = "0.2.4", optional = true }
 humansize = "1.1.1"
+itertools = "0.10.3"
```
MichaReiser, line 37: What do we need itertools for?

MichaReiser (commented):
That's a neat improvment. More functionality and improved runtime. 

Do you see any memory regression since the `GreenTriviaPiece::Whitespace` now works in fewer places (never when there's whitespace + new lines). 

Should we introduce the `NEWLINE` `GreenTriviaPiece` as part of this PR too or do you plan to do so in a followup PR?

crates/rslint_lexer/src/lib.rs
xunilrj: I think it makes sense.

crates/rslint_parser/src/lossless_tree_sink.rs
xunilrj: Yes. I think this is the heart of the improvement. But real gains are probably smaller than what I put in the comments.

crates/rslint_parser/src/token_source.rs
xunilrj: I actually want to remove this entirely. We can discuss this.

xtask/src/libs/mod.rs
xunilrj: I like this total time because I do

```
> cargo run -p xtask --release -- coverage-libs | sls "Total Time"
```

But we can move it to the ```else``` clause of the if.

xtask/Cargo.toml
xunilrj: To display diagnostics using ```group_by```:

```rust
println!("Diagnostics");
for (severity, items) in &diags.iter().group_by(|x| x.severity) {
println!("\t{:?}: {}", severity, items.count());
}
```

crates/rslint_lexer/src/lib.rs
```diff
@@ -1474,12 +1474,20 @@ impl Iterator for Lexer<'_> {
...
 			self.lex_token()
 		};
 
+		let after_newline = match token.0.kind {
+			SyntaxKind::NEWLINE => true,
+			_ => false
+		};
+
+		token.0.offset = self.cur - token.0.len;
+		token.0.after_newline = std::mem::replace(&mut self.state.after_newline, after_newline);
```
MichaReiser, line 1493: Is there any performance benefit of using `mem::replace` for a boolean? feels a bit overkill

crates/rslint_parser/src/token_source.rs
```diff
@@ -4,46 +4,20 @@ use std::collections::HashSet;
...
-				.last()
-				.map(|x| {
-					let start: usize = x.1.into();
-					let end = start + x.0.len;
-					start..end
-				})
-				.unwrap_or(0..0)
-		});
-
+fn mk_token2(pos: usize, raw_tokens: &[rslint_lexer::Token]) -> Token {
```
MichaReiser: Rename to `mk_token`

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
 			source,
-			token_offset_pairs,
-			cur: (first, 0),
-			tokens_after_linebreaks,
+			cur: (first, pos),
 			raw_tokens,
 		}
 	}
 
+	fn next_non_trivia(&self, pos: usize, dir: isize) -> Option<usize> {
```
MichaReiser, line 33: what's `dir`? is it `direction` or maybe `steps`?

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
+		let mut pos = pos as isize + dir;
+		if (pos < 0) || ((pos as usize) >= self.raw_tokens.len()) {
+			return None
+		} 
+		while self.raw_tokens[pos as usize].kind.is_trivia() {
+			pos += dir;
```
MichaReiser, line 39: This only seems to work if `dir` is either `1` or `-1`. I suggest changing `dir` to a boolean or enum.

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
+			if (pos < 0) || ((pos as usize) >= self.raw_tokens.len()) {
+				return None
+			}
+		}
```
MichaReiser, line 43: Can this be re-written as a `do...while` instead to avoid the repetition of the out of bound check.

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
+		Some(pos as usize)
+	}
+
 	/// Rewind the current position to a former position.
 	pub fn rewind(&mut self, pos: usize) {
-		self.cur = (mk_token(pos, &self.token_offset_pairs), pos);
+		//println!("rewind: {}", pos);
```
MichaReiser: Delete

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
+		self.cur = (mk_token2(pos, &self.raw_tokens), pos);
 	}
 
 	pub fn last_tok(&self) -> Option<Token> {
-		if self.cur.1 == 0 {
-			return None;
-		}
-		Some(mk_token(self.cur.1 - 1, &self.token_offset_pairs))
+		//println!("last_tok");
```
MichaReiser: Delete, as well as all other commented out println statements.

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
 
 	pub fn source(&self) -> &str {
+		//println!("source");
 		self.source
 	}
 
+	fn raw_lookahead_nth(&self, n: usize) -> usize {
+		let mut idx = self.cur.1;
+		for _ in 0..n {
+			idx = self.next_non_trivia(idx, 1).unwrap();
```
MichaReiser: The old implementation was more lenient than panicking. I don't think it's feasible to require all calls to `lookahead_` to first check if they're not out of bound. The way this has been implemented before is that `mk_token` returned `EOF` if out of bound. 

I believe `raw_lookaehaed_nth` should also return an `Option` and `mk_token` or whatever is calling it must cracefully handle the case that we may be out of bounds.

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
+		//println!("current");
 		self.cur.0.to_owned()
 	}
 
 	pub fn source(&self) -> &str {
+		//println!("source");
 		self.source
 	}
 
+	fn raw_lookahead_nth(&self, n: usize) -> usize {
```
MichaReiser: I struggled to understand what the `raw` part is compared to `next_non_trivia`. But it seems these method are the same except that `next_non_trivia` advances by 1 whereas this method advances by `n`. How about `nth_non_trivia`?

crates/rslint_parser/src/token_source.rs
```diff
@@ -54,112 +28,115 @@ impl<'t> TokenSource<'t> {
...
+		//println!("cur_pos");
+		self.raw_tokens[self.cur.1].offset
 	}
 
 	pub fn cur_token_idx(&self) -> usize {
+		//println!("cur_token_idx");
 		self.cur.1
 	}
 
 	pub fn size_hint(&self) -> usize {
```
MichaReiser, line 117: Should we override `Iterator::size_hint`? and possibly implement `FusedIterator` ?

crates/rslint_lexer/src/token.rs
```diff
@@ -10,17 +10,30 @@ pub struct Token {
 	/// How long the token is in bytes. For tokens with escape sequences
 	/// like strings with `\uXXXX` escapes, the length is the raw length, not considering the char backed by the escape.
 	pub len: usize,
+
+	pub offset: usize,
+	pub after_newline: bool
```
MichaReiser: Can we document the meaning of these two new fields?

I'm wondering if we should use two different `Token` structs. One used by the lexer (`RawToken`) and another by the token source. 

* `RawToken`: only stores kind and len
* `Token`: public API, gives access to `len`, `offset`, and `kind`. Stores `after_newline` internally to implement `has_linebreak_before`. 

This would avoid storing `offset` + `after_newline` for every token on the heap and instead, only require storing it for the current token.

crates/rslint_lexer/src/lib.rs
xunilrj: For performance, no. I used just because is a oneliner.

xunilrj:
This PR got bigger than planned (and much more complex). I will convert it into a issue and break it into multiple PRs, that is going to be easier to review.

ematipico:
Should we close this PR?

xunilrj:
> tokens_after_linebreaks

Yes. I will.

PR #1896, opened 2022-01-05
fix(rslint_parser): Fix parsing of TS keywords

MichaReiser:
## Summary
declare, abstract, namespace, type, enum, module, global, interface, private, protected, and public are typescript keywords but valid identifiers in JS.

This PR removes the TS parsing logic for now and ensures these identifiers are correctly parsed.

## Test Plan

`compiler.js` no longer reports any errors.

### DIFF
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -50,7 +50,7 @@ pub const ASSIGN_TOKENS: TokenSet = token_set![
 	T![**=],
 ];
 
-pub const STARTS_EXPR: TokenSet = token_set![
+const STARTS_EXPR: TokenSet = token_set![
 	T![!],
 	T!['('],
 	T!['['],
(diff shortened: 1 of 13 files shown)


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -959,6 +960,10 @@ pub(crate) fn parse_expression(p: &mut Parser) -> ParsedSyntax {
 	})
 }
 
+pub(crate) fn is_at_expression(p: &Parser) -> bool {
```
xunilrj, line 963: Is this function specific for Typescript? If it is, would be nice to have a ```_ts``` at the end, or a comment.

xunilrj (approved):

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: This is a non TS specific function that tests if the parser is currently at the start of an expression.

PR #1899, opened 2022-01-06
fix(rslint_parser): Panic when parsing named exports

MichaReiser:
## Summary

The named exports parsing didn't always complete all started markers. I then noticed, that the parse method sometimes returns `ExportNamed` nested in an `ExportDecl` and sometimes not.

This PR fixes the `ExportNamed` parsing to not nest it inside an `ExportDecl` and ensures that all markers are correctly completed or abandoned. We can revisit this when working on the final structure for AST nodes.

Fixes most remaining panics.


xtask/js.ungram
```diff
@@ -1139,7 +1140,7 @@ TsNamespaceExportDecl = 'export' 'as' 'namespace' ident: Ident? ';'?
 Specifier = name:JsName 'as'? alias: JsName?
 
 // @ematipico this one is not entirely correct I think..
```
Boshen, line 1142: Is it still "not entirely correct"?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -636,22 +636,6 @@ fn parse_static_member_expression(
 	Present(m.complete(p, JS_STATIC_MEMBER_EXPRESSION))
 }
 
-pub(super) fn parse_name(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 639: Duplicate with `parse_identifier_name`. Keep `parse_name` naming to match the grammar rule name.

ematipico (approved):

PR #1902, opened 2022-01-06
xtask coverage-libs using criterion to better assess performance

xunilrj:
How to run

```
cargo run -p xtask --release -- coverage-libs --filter=jquery
```

Result

```
[jquery.min.js] - using [target/jquery.min.js]
Gnuplot not found, using plotters backend
Benchmarking https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js
Benchmarking https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js: Warming up for 3.0000 s
Benchmarking https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js: Collecting 100 samples in estimated 6.1669 s (500 iterations)
Benchmarking https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js: Analyzing
https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js
                        time:   [11.744 ms 11.977 ms 12.241 ms]
                        change: [-2.3855% -0.0651% +2.6054%] (p = 0.97 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe

Benchmark: https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js
        Tokenization: 1.470025ms
        Parsing:      4.849397ms
        Tree_sink:    6.092388ms
                      ----------
        Total:        12.41181ms
        Diagnostics
                Warning: 2

Summary
-------
jquery.min.js,Total Time,12.41181ms,tokenization,1.470025ms,parsing,4.849397ms,tree_sink,6.092388ms
```

The time returned by ```criterion``` and  the time split by parts sometimes varies a lot, like the example above.

Criterion can be disabled with

```
cargo run -p xtask --release -- coverage-libs --filter=jquery --criterion=false
```

The "Total Time" line is interesting to run with the oneliners below and gather a summary.

Poweshell
```
cargo run -p xtask --release -- coverage-libs --criterion=false|sls "Total Time"
```

Linux
```
cargo run -p xtask --release -- coverage-libs --criterion=false|grep "Total Time"
```

PR #1903, opened 2022-01-06
xtask coveragelibs summary at the end

xunilrj:
Continuation of: https://github.com/rome/tools/pull/1902
This is just the summary at the end.


ematipico:
I think it would be beneficial if we have all these commands/options written in some readme

Issue #1901, opened 2022-01-06
☂️ Lexer Improvements 


xunilrj:
### Description

We can improve the parser performance by making some improvements on the lexer. First when parsing we need to peeking into ```WHITESPACE``` and ```COMMENT` texts to find newlines. We can avoid this having specific tokens.

With these specific tokens, and the lexer knowing that we had a newline, we can completely remove ```tokens_after_linebreaks```. The lexer will know when a non-trivia token is after a newline. We can attach this info in the ```Token``` struct and completely avoid creating and searching this ```HashSet```.

We can also improve the lexer by already calculating the token offset and remove ```token_offset_pair```. Avoiding a whole iteration when creating the ```Parser```.

The lexer is also returning ```LexerReturn``` or for that matter ```Option<Diagnostics>``` on the stack. Although the super vast majority of the lexing never return any diagnostic, this struct is huge and generate a huge cost by copying data. We can avoid this.

The previous implementation of trivia was focused on optimizing for just one trivia. This change breaks this because we split new lines from whitespace. We may need to refactor this part.  

# Tasks

- [ ] Improve "xtask coverage-libs"
    - [x] Run https://github.com/bheisler/criterion.rs for best results
    - [ ] Flag to display diagnostics
    - [ ] Option to try multiple allocators
- [x] Make https://github.com/matklad/drop_bomb configurable (default only on Debug)
    - [x] ```drop_bomb``` already exclude itself on release builds. We just have to use ```DebugDropBomb```.
- [x] Implement new NEWLINE and MULTILINE_COMMENT token kinds
- [x] Remove token_offset_pair and tokens_after_linebreaks
    - [x] ```next_non_trivia``` ```dir``` as an enum
    - [x] ```raw_lookahead_nth``` must return the last token if it goes out of bounds 
- [x] Improve performance of the lexer, decreasing size of LexerReturn
- [ ] Improve trivia
    - [x] ```NewLine``` and multi line comments on the trivia API
    - [x] Allocate only when necessary inside ```get_trivia```
    - [ ] ```GreenTokenTrivia``` is optimized for zero/one trivia. Does this still make sense?

# NEWLINE Token

Today to know with a WHITESPACE token has a newline, we iterate the token when parsing. We can do better by allowing the lexer to create specific tokens with newline. With this we know that WHITESPACE never has newlines.

Before
```rust
assert_lex! {
	"\n ",
	WHITESPACE:2
}
```

After
```rust
assert_lex! {
	"\n ",
[...]

PR #1904, opened 2022-01-06
making marker dropbomb optional on release

xunilrj:
We added https://github.com/matklad/drop_bomb to ```rslint_parser::Marker``` to, very politely, panic if you forget to deal correctly with ```Marker```s. 

It is really helpful when developing but it has a performance impact that we can avoid in release mode. This PR makes the bomp do nothing in release, letting the optimizer to completely scrap it, but also allow the bomb to be opted-in by the ```force-markerbomb``` feature.

The performance gain is smaller than when I investigated this for the first time, but still nice to have.

Before
```
                        time:   [84.660 ms 86.023 ms 87.527 ms]
```

After
```
                        time:   [82.035 ms 82.676 ms 83.358 ms]
```

The bomb can be forced using

```
cargo run -p xtask --features "rslint_parser/force-markerbomb" --release -- coverage-libs --filter=three
```

for more see: https://github.com/rome/tools/issues/1901

### DIFF
--- a/crates/rslint_parser/Cargo.toml
+++ b/crates/rslint_parser/Cargo.toml
@@ -14,7 +14,10 @@ rslint_lexer = { path = "../rslint_lexer", version = "0.2", features = ["highlig
 rome_rowan = { path = "../rome_rowan", version = "0.0.0" }
 num-bigint = "0.3.0"
 lexical = { version = "5.2.0", features = ["radix"] }
-drop_bomb = "0.1.5"
+drop_bomb = { version = "0.1.5", optional = true }
 
 [dev-dependencies]
 expect-test = "1.0"
+
+[features]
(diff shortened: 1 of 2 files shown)


MichaReiser (commented):
I believe the crate has already built in support for disabling drop bombs with a feature flag

https://docs.rs/drop_bomb/latest/drop_bomb/#notes


https://github.com/matklad/drop_bomb/blob/master/src/lib.rs#L86-L87

Boshen:
Splitting out a "debug release" profile with custom overrides will make things easier, i.e. https://doc.rust-lang.org/cargo/reference/profiles.html#overrides

xunilrj:
> I believe the crate has already built in support for disabling drop bombs with a feature flag

You are right. Still don't know why I saw better performance when I commented these bombs. Will close this.

PR #1907, opened 2022-01-07
feature(xtask): add encoding detection to compare

leops:
## Summary

On PowerShell the default behavior for pipes it to output an UTF-16 encoded file, so `cargo xtask coverage --json > results.json` will create a file that can't be decoded directly by `cargo xtask compare` since serde_json and Rust in general assumes an UTF-8 encoding.

This adds a simple encoding detection logic to the loading of test results file that detects (and skips over) Unicode Byte Order Mark, and re-encodes UTF-16 content into UTF-8 before handing it to serde_json

### DIFF
--- a/xtask/src/compare/results.rs
+++ b/xtask/src/compare/results.rs
@@ -1,17 +1,15 @@
 use crate::coverage::files::{Outcome, TestResult, TestResults};
 use ascii_table::{AsciiTable, Column};
 use colored::Colorize;
+use std::char::decode_utf16;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
+use std::io::Read;
 use std::path::{Path, PathBuf};
 
 pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
-	let base_results: TestResults =
-		serde_json::from_reader(File::open(base).expect("Can't read the file of the base results"))
-			.expect("Can't parse the JSON file of the base results");
-	let new_results: TestResults =
-		serde_json::from_reader(File::open(new).expect("Can't read the file of the new results"))
-			.expect("Can't parse the JSON file of the new results");
+	let base_results = read_test_results(base, "base");
+	let new_results = read_test_results(new, "new");
 
 	let base_total = base_results.summary.tests_ran as isize;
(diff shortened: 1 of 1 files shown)


Boshen:
Perhaps it is better to configure your Powershell to output utf-8 instead? I see a lot of people do this on [stackoverflow](https://stackoverflow.com/questions/40098771/changing-powershells-default-output-encoding-to-utf-8) 😃

ematipico:
I'm torned here, both arguments are valid. Although, so far we haven't had this issue even though we have people that use Windows as their primary OS. So my question is, how likely people would encounter this issue?

We should also highlight the fact that these commands are there for us maintainers/contributors. If this error might occur often, especially for contributors, maybe it's worth fixing it.

Although this is also an information that we can add inside the `CONTRIBUTING.md` section. What do you think, team?

MichaReiser:
I'm in favour of this change. This is an issue for everyone using the "classic" Windows PowerShell (not the cross-platform version) as explained in the stack overflow article that @Boshen linked. 

It only adds little complexity and eases the setup of many. I also don't think that it should require engineers to change their dev tooling to work on Rome because they may then end up in a situation where they need UTF8 for Rome but UTF16 for some other project they're working on.

ematipico (approved):

xtask/src/compare/results.rs
```diff
@@ -195,6 +193,74 @@ pub fn emit_compare(base: &Path, new: &Path, markdown: bool) {
...
 
+fn read_test_results(path: &Path, name: &'static str) -> TestResults {
+	let mut file = File::open(path)
+		.unwrap_or_else(|err| panic!("Can't read the file of the {} results: {:?}", name, err));
+
+	let mut buffer = Vec::new();
+	file.read_to_end(&mut buffer)
+		.unwrap_or_else(|err| panic!("Can't read the file of the {} results: {:?}", name, err));
+
+	enum FileEncoding {
```
xunilrj, line 204: I am wondering here if we can use something like: https://github.com/kena0ki/aconv

xunilrj:
> I'm in favour of this change [..] eases the setup of many. 

+1

PR #1908, opened 2022-01-07
new NEWLINE and MULTILINE_COMMENT tokens

xunilrj:
Today to know with a WHITESPACE token has a newline, we iterate the token when parsing. We can do better by allowing the lexer to create specific tokens with newline. With this we know that WHITESPACE never has newlines.

see https://github.com/rome/tools/issues/1901 for more details.

Before
```
                        time:   [84.446 ms 85.177 ms 85.955 ms]
```

After
```
                        time:   [82.680 ms 83.598 ms 84.594 ms]
```


crates/rslint_parser/src/tests.rs
```diff
@@ -233,10 +233,11 @@ pub fn test_trivia_attached_to_tokens() {
...
 	let pieces: Vec<_> = second_let.leading_trivia().pieces().collect();
-	assert_eq!(3, pieces.len());
-	matches!(pieces.get(0).map(|x| x.text()), Some("\n "));
-	matches!(pieces.get(1).map(|x| x.text()), Some("/*hey*/"));
-	matches!(pieces.get(2).map(|x| x.text()), Some(" "));
+	assert_eq!(4, pieces.len());
+	matches!(pieces.get(0).map(|x| x.text()), Some("\n"));
+	matches!(pieces.get(1).map(|x| x.text()), Some(" "));
+	matches!(pieces.get(2).map(|x| x.text()), Some("/*hey*/"));
+	matches!(pieces.get(3).map(|x| x.text()), Some(" "));
```
MichaReiser, line 240: Is this asserting anything? It isn't using the return value of `matches`. Should these be wrapped with `assert`?

crates/rslint_parser/test_data/inline/err/break_stmt.rast
```diff
@@ -41,7 +41,7 @@ JsModule {
                 l_curly_token: L_CURLY@39..40 "{" [] [],
                 statements: JsStatementList [
                     JsBreakStatement {
-                        break_token: BREAK_KW@40..49 "break" [Whitespace("\n  ")] [Whitespace(" ")],
+                        break_token: BREAK_KW@40..49 "break" [Whitespace("\n"), Whitespace("  ")] [Whitespace(" ")],
```
MichaReiser, line 44: Do you plan to introduce the NewLine `TriviaPiece` in a different PR. It feels odd, that there are two separate trivia, both for whitespace. 

This would also speed up the formatter, because determining the number of line breaks between two tokens becomes then finding all new line trivia pieces.

MichaReiser (approved):

crates/rslint_parser/test_data/inline/err/break_stmt.rast
xunilrj: Yes. Probably in the last one.

crates/rslint_parser/src/tests.rs
xunilrj: Fixed!

PR #1911, opened 2022-01-08
chore(ci): Run test262 in release mode to speed CI

Boshen:
Let's see if there is a overall speed up


Boshen:
Forks don't get a bot comment, [source](https://github.com/peter-evans/create-or-update-comment#action-inputs)
> Note: In public repositories this action does not work in pull_request workflows when triggered by forks. Any attempt will be met with the error, Resource not accessible by integration. This is due to token restrictions put in place by GitHub Actions.

.github/workflows/main.yml
```diff
@@ -93,7 +102,10 @@ jobs:
           profile: minimal
       - name: Cache
         uses: Swatinem/rust-cache@v1
+      - name: Compile
+        run: cargo build --release --locked -p xtask
```
Boshen, line 106: `--locked` Require Cargo.lock is up to date

Cargo.toml
```diff
@@ -4,6 +4,13 @@
 resolver = "2"
 members = ["xtask/", "crates/*"]
 
-# TODO: Possibly setup profiles for dev to speed up tests/etc
+[profile.dev]
+# Disabling debug info speeds up both local and CI build,
+# enable only if neccessary
+debug = 0
```
MichaReiser, line 10: What are the effects of this? Does this also apply if I create a debug build locally? 

What happens with test failures in CI? Will they miss a detailed stack trace if debug is set to false?

MichaReiser (commented):
> Previously we tried to turn off the windows build but somehow the linux build kept failing on writing comments, it turns out we were overriding the file in cross builds, resulting the linux build getting "Added 40000 new tests" each time.

This is an amazing finding! Thank you so much for this. 

looks good to me. There seems to be a quiet a bit of repetition but I'm not familiar with github actions, so I don't know if there are ways to minimize it.

Cargo.toml
Boshen: `debug = 0` is already the default setting, so this addition has no effect. I'm just copying this from rust-analyzer's Config.toml, to make it explicit and prevent future misconfigurations.

Boshen:
> There seems to be a quiet a bit of repetition but I'm not familiar with github actions, so I don't know if there are ways to minimize it.

All CI pipeline config files are like this, just repetition everywhere 🥲

Boshen:
Added two more changes to speed up things further

## Merge coverage comparison workflow
We had two workflows for uploading and downloading comparisons. We can make this simpler by checking out the main branch and run it again. This halves our CI build time.

## Remove default toolchain
`rust-toolchain.toml` has `targets` set to 4 platforms. This is unnecessary, and it has the downside of downloading unneeded platforms in both local and CI development.
Removing this field makes speeds up rust-toolchain installation.

Boshen:
All is done, final improvement from 20~30min down to ~3min, 10x improvement 🎉

.github/workflows/main.yml
```diff
@@ -22,9 +22,10 @@ jobs:
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.57.0
+          toolchain: stable
```
Boshen, line 25: using `stable` is preferred, a convention from the community.

Cargo.toml
Boshen: I have reverted the change, it doesn't really belong in this PR.

.github/workflows/test262.yml
```diff
@@ -13,66 +13,50 @@ jobs:
...
+
       - name: Compare results on ${{ matrix.os }}
         if: github.event_name == 'pull_request'
         id: comparison
         shell: bash
         run: |
-          comment="$(cargo xtask compare ./base_results.json ./new_results.json  --markdown)"
+          echo "### Test262 comparison coverage results on ${{ matrix.os }}" > output
+          cargo run -p xtask --release -- compare ./base_results.json ./new_results.json --markdown >> output
+          cat output
```
Boshen, line 58: Forks doesn't comment, so print it here if we need it.

Boshen:
@ematipico mentioned that the rust version in CI doesn't sync with `rust-toolchain.toml`, it turns out that `actions-rs/toolchain@v1` is useless, we can just use rustup to get all the benefits. 
See commit da5b5c0 for the change.

MichaReiser:
Would it make sense to use the minimal profile, for example, the t262 tests? This is something we got for free with the actions. 
https://rust-lang.github.io/rustup/concepts/profiles.html

Boshen:
> Would it make sense to use the minimal profile, for example, the t262 tests? This is something we got for free with the actions. https://rust-lang.github.io/rustup/concepts/profiles.html

That plugin didn't do what is intended. It ran `rustup show` and installed everything specified in `rust-toolchain.toml`, and then set the minimal profile, which is pretty useless ... adding rustfmt/clippy/docs is just a few extra seconds and some IO, so it doesn't really matter.

Here is the output from before:
```
/home/runner/.cargo/bin/rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/runner/.rustup

info: syncing channel updates for '1.57.0-x86_64-unknown-linux-gnu'
info: latest update on 2021-12-02, rust version 1.57.0 (f1edd0429 2021-11-29)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std' for 'wasm32-unknown-unknown'
info: downloading component 'rust-std' for 'x86_64-apple-darwin'
info: downloading component 'rust-std' for 'x86_64-pc-windows-msvc'
info: downloading component 'rust-std'
info: downloading component 'rust-std' for 'x86_64-unknown-linux-musl'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std' for 'wasm32-unknown-unknown'
info: installing component 'rust-std' for 'x86_64-apple-darwin'
info: installing component 'rust-std' for 'x86_64-pc-windows-msvc'
info: installing component 'rust-std'
info: installing component 'rust-std' for 'x86_64-unknown-linux-musl'
info: installing component 'rustc'
info: installing component 'rustfmt'
installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-apple-darwin
x86_64-pc-windows-msvc
x86_64-unknown-linux-gnu
x86_64-unknown-linux-musl

active toolchain
----------------

1.57.0-x86_64-unknown-linux-gnu (overridden by '/home/runner/work/tools/tools/rust-toolchain.toml')
rustc 1.57.0 (f1edd0429 2021-11-29)

/home/runner/.cargo/bin/rustup -V
rustup 1.24.3 (ce5817a94 2021-05-31)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.57.0 (f1edd0429 2021-11-29)`
Installed rustup 1.24.3 support profiles
/home/runner/.cargo/bin/rustup set profile minimal
info: profile set to 'minimal'
/home/runner/.cargo/bin/rustup toolchain install 1.57.0
info: syncing channel updates for '1.57.0-x86_64-unknown-linux-gnu'

  1.57.0-x86_64-unknown-linux-gnu unchanged - rustc 1.57.0 (f1edd0429 2021-11-29)

info: checking for self-updates
warning: tool `rustfmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
warning: tool `cargo-fmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
```

Boshen:
@ematipico This is ready, can you review the changes and its CI output?

xunilrj:
Can we use https://github.com/mozilla/sccache?
https://github.com/marketplace/actions/sccache-action-speedup-rust-github-actions

It may improve things quite a bit....

Boshen:
> Can we use https://github.com/mozilla/sccache? https://github.com/marketplace/actions/sccache-action-speedup-rust-github-actions
> 
> It may improve things quite a bit....

I tried it on my own project, and see no improvements.

There are only about 10 projects using this across github, so I guess its not working as intended (again, like most of the other github action pipelines)

xunilrj:
> I tried it on my own project, and see no improvements.

So let us ignore it for now.

.github/workflows/test262.yml
ematipico: I don't think we need to print here with `cat output`. I mean, it's just for debug... we have the comment for actually see the end result.

ematipico (approved):
Looks good! Just a minor comment

.github/workflows/test262.yml
Boshen: My PR doesn't comment because its from a fork, the comment bot just errors out because it can't access the token required to make a comment.
I need this to copy the output and make a manual comment 😿 

This could have prevented the long standing issue of the linux build not being able to comment, since we couldn't see the output.

Boshen:
My PR doesn't comment because it's from a fork, the comment bot just errors out because it can't access the token required to make a comment.
I need this to copy the output and make a manual comment 😿

This could have prevented the long standing issue of the linux build not being able to comment, since we couldn't see the output.

Can this be merged now 😄 ?

PR #1914, opened 2022-01-10
feat: add prototype lsp server, analyzers, and vscode extension

yassere:
## Summary

This PR adds prototypes of an LSP server, syntax analyzers, and a VS Code extension. They’re all very much a WIP but merging this now will facilitate requested collaboration with @NicholasLYang 

Some unfinished/experimental work was omitted from this PR and can be added in incrementally, as needed.

Notable omissions:

- Incremental computation using `salsa`. Some form of memoization will be important in the future (especially for semantic analysis) but we should evaluate possible approaches once we have more to benchmark.
- A `replace_nodes` method that directly transforms syntax trees using multiple (non-overlapping) `SyntaxEdits` rather than converting them to text edits. This isn’t crucial for LSP, but it will be useful for CI in order to apply multiple autofixes sequentially.  There had been a working implementation prior to the trivia/slots changes, but it hasn’t been updated.
- A thread-safe `SyntaxElementPtr` and a thread-safe version of `SyntaxEdit`. These may not end up being necessary, depending on our approach to concurrency and incremental computation.

Future Work
- Add tests and documentation. Improve code quality.
- Add more analyzers to explore the limitations of the current APIs and make enhancements.
- Replace the hacky `make` module with a proper set of factory functions. Maybe they can (partially) use codegen and validation similar to the new `syntax_factory`


crates/rome_analyze/src/signals.rs
```diff
@@ -0,0 +1,170 @@
...
+#[derive(Default, Debug)]
+pub struct Analysis {
+	pub signals: Vec<Signal>,
+}
+
+impl Analysis {
+	pub fn into_actions(self) -> impl Iterator<Item = Action> {
+		self.signals
+			.into_iter()
+			// TODO: There must be a better way to do this.
```
Boshen, line 157: Either do some heavy `impl Iterator` logic on Signal, or use `flatmap`, and `vec![a]` -> `std::iter::once(a)`?

crates/rome_analyze/src/analyzers.rs
```diff
@@ -0,0 +1,58 @@
...
+	pub(crate) analyze: fn(&AnalyzerContext) -> AnalyzerResult,
+}
+
+pub struct AnalyzerContext<'a> {
+	pub file_id: FileId,
+	analysis: &'a AnalysisServer,
+}
+
+impl<'a> AnalyzerContext<'a> {
+	pub fn new(analysis: &'a AnalysisServer, file_id: FileId) -> Self {
```
MichaReiser, line 32: Nit: Should `AnalysisServer` be called `Analysis` because it reflects the lifetime of an analysis (multiple ones, but single run). Or should the variable be renamed to `analysis_server?`

crates/rome_analyze/src/analyzers/no_var.rs
```diff
@@ -0,0 +1,19 @@
...
+	Analyzer {
+		name: "noVar",
+		action_categories: vec![],
+		analyze,
+	}
+}
+
+fn analyze(ctx: &AnalyzerContext) -> Result<Analysis> {
+	ctx.query_nodes::<ast::JsVariableDeclarations>()
+		.filter(|n| n.kind().map(|k| k.kind()).ok() == Some(JsSyntaxKind::VAR_KW))
```
MichaReiser, line 16: Nit: you can use the `variable_kind` extension method on `VariableDeclarations` so that you don't need to match on the keyword

Nit: The `ok()` isn't needed, you can compare the result directly: `k.kind() == Ok(JsSyntaxKind::VAR_KW)`

crates/rome_analyze/src/analyzers/use_single_case_statement.rs
```diff
@@ -0,0 +1,36 @@
...
+
+fn analyze(ctx: &AnalyzerContext) -> Result<Analysis> {
+	ctx.query_nodes::<ast::JsCaseClause>()
+		.filter(|n| n.consequent().len() > 1)
+		.filter_map(|node| {
+			let message = "rome: A switch case should only have a single statement.";
+			let action_title = "rome: Wrap case body in block";
+
+			let test = node.test().ok()?;
+			let block = ast::JsBlockStatement::quick(node.consequent());
```
MichaReiser, line 22: Where's the `quick` coming from? How about `single`? or `make_with_single_statement()`

crates/rome_analyze/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "rome_analyze"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rslint_parser = { path = "../rslint_parser" }
+salsa = "0.17.0-pre.1"
```
Boshen, line 10: It seems `salsa` is not being used? Is this going to be used in the future?

crates/rslint_parser/src/make.rs
```diff
@@ -0,0 +1,113 @@
...
+	};
+	let node = node.clone_subtree();
+	assert_eq!(node.syntax().text_range().start(), 0.into());
+	node
+}
+
+fn token_from_text(kind: JsSyntaxKind, text: &str) -> SyntaxToken {
+	let mut builder = SyntaxTreeBuilder::new();
+	builder.start_node(JsSyntaxKind::JS_SCRIPT);
+	builder.token(kind, text);
```
MichaReiser, line 32: This is super hacky and only works because you're only querying a token (the scrip itself likely becomes an unknown node). 

The way this should probably work is that `JsSyntaxFactory` has `make_ident` etc. methods as you suggest in the title. But yeah, this requires codegen support.

Can we crate a task to track the pending issues (probably across the whole PR)?

crates/rome_analyze/Cargo.toml
Boshen: Nevermind, 
> Incremental computation using salsa. Some form of memoization will be important in the future (especially for semantic analysis) but we should evaluate possible approaches once we have more to benchmark.

crates/rslint_parser/src/make.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl JsStatementList {
+	pub fn wrap(statement: JsAnyStatement) -> Self {
+		ast_from_text(&statement.text())
+	}
+}
+
+impl JsWhileStatement {
+	pub fn quick(test: JsAnyExpression, body: JsAnyStatement) -> Self {
```
MichaReiser, line 53: Can we rename these `quick` to `make`? Quick isn't really telling much. 

We could then have a `make_with_tokens` or provide modifier methods like `with_l_paren` on the `WhileStatement` to replace specific tokens.

crates/rslint_parser/src/make.rs
```diff
@@ -0,0 +1,113 @@
...
+		let text = format!("while ({}) {}", test.text(), body.text());
+		ast_from_text(&text)
+	}
+}
+
+impl Default for JsEmptyStatement {
+	fn default() -> Self {
+		ast_from_text(";")
+	}
+}
```
MichaReiser, line 63: I'm not sure if it's a good idea to implement `Default`. We'll likely want to cache these nodes as well and, therefore, need access to the `SyntaxFactory` in some way or another. `Default` doesn't provide the means for that. I would stick with `make` that simply doesn't accept any arguments.

crates/rslint_parser/src/make.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl JsBooleanLiteralExpression {
+	pub fn quick(value: bool) -> Self {
+		match value {
+			true => ast_from_text("true"),
+			false => ast_from_text("false"),
+		}
+	}
```
MichaReiser, line 71: All `make` methods should accept token or nodes. I can see how this might be convenient but it also prevents authors from providing their own tokens (that have, for example, comments attached).

Boshen:
> Add more analyzers to explore the limitations of the current APIs and make enhancements.

Can you list out some for us to experiment with? So we can get a feel of this architecture.

crates/rslint_parser/src/make.rs
```diff
@@ -0,0 +1,113 @@
...
+			}
+			text.push_str(&node.text().to_string())
+		}
+		text.push('\n');
+		if indent > 1 {
+			text.push_str(&"\t".repeat(indent - 1));
+		}
+		text.push('}');
+
+		ast_from_text(&text)
```
MichaReiser, line 100: Wouldn't it be easier to use the `TreeBuilder` here rather than converting to text first? Probably requires a method to append an already constructed node (statements could then be a reference as well)

```
let mut builder = SyntaxTreeBuilder::default();
builder.start_node(JS_BLOCK_STATMENT);
builder.token(JS_L_PAREN, "{");
for stmt in statements {
 builder.append_node(stmt);
}
builder.token(JS_R_PAREN, "}");
builder.finish_node();
builder.finish()
```

Boshen:
I followed the code path a couple times. Everything is understandable until the point on authoring code fixes (authoring lints is fine). And from @MichaReiser's code review, maybe it's better to just leave a placeholder for authoring code fixes, so we can focus on the lsp instead of bikeshedding the analyzer.

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -0,0 +1,99 @@
...
+	suppressions::{self, Suppressions},
+	Analysis, AnalyzerContext, AnalyzerResult, AssistContext,
+};
+
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, PartialEq, Eq, Hash)]
+pub struct FileId(pub usize);
+
+#[derive(Default)]
+pub struct AnalysisServer {
+	file_map: HashMap<FileId, Arc<String>>,
```
leops, line 18: Since the data pointed to in `Arc` is immutable anyway I think these can be replaced with `Arc<str>` to remove an indirection level (dereferencing the string becomes Arc ->str instead of Arc -> String -> str)

crates/rome_analyze/src/signals.rs
leops: `flat_map` would help however `std::iter::once` wouldn't work directly since the closure needs to return the same type on all branches (`std::vec::IntoIter` in this case), unless both iterator types were wrapped as `Box<dyn Iterator<Item = Action>>`

crates/rslint_parser/src/make.rs
leops: Additionally we could have a `quote` macro to wrap syntax tree generation in a readable way, so we could just write `quote!( switch true { case #test: #consequent } )` in the `JsCaseClause` implementation for instance, but it will probably have to be a procedural macro if it needs to emit nodes and not just a stream of token

crates/rome_analyze/Cargo.toml
yassere: Yeah, I forgot to remove the `salsa` dependency from this branch. Thanks for catching it.

crates/rome_analyze/src/analyzers.rs
yassere: An `AnalysisServer` should eventually last longer than a single run. I'll rename the variable.

crates/rome_analyze/src/analyzers/use_single_case_statement.rs
yassere: The `quick` comes from the [original rome](https://github.com/rome/tools/blob/ecf6d3a10334195a258631a6adb3548d28ddfb38/internal/compiler/lint/rules/js/useSingleCaseStatement.ts#L22). I'll rename these to `make` as suggested in your other comment. Eventually, these might just end up as methods on a syntax factory. e.g. `SyntaxFactory::js_block_statement(...)`

crates/rslint_parser/src/make.rs
yassere: This whole module is super hacky. Probably a better idea to just remove it from the PR (along with the autofixes that depend on it) and then start working on a proper solution.

crates/rslint_parser/src/make.rs
yassere: I had that as a `new()` and changed it to a `Default` impl to satisfy `clippy` but you're right that it's not a good idea.

crates/rslint_parser/src/make.rs
yassere: I think it's useful to have a shortcut for making a minimal version of something in addition to a complete creation method that would require you pass in all nodes and tokens. Do you think that's a bad idea?

crates/rslint_parser/src/make.rs
yassere: This function especially shouldn't have made it into this PR. I did originally start to use the tree builder but I didn't want to augment it with an `append_node` method in the midst of my sloppy hacking spree.

A custom proc macro could be useful, but hopefully we can get pretty far with a syntax factory codegen.

crates/rome_analyze/src/analyzers/no_var.rs
yassere: The comment, method name, and implementation are all different in this [JsVariableDeclarations impl](https://github.com/rome/tools/blob/df6d5d52e74513d1ba5fdf1a37a00ca83398839c/crates/rslint_parser/src/ast/stmt_ext.rs#L23-L26) :grin:

crates/rome_analyze/src/analysis_server.rs
yassere: Swapping this didn't immediately compile because of unsatisfied trait bounds, so I'll take a closer look at this later. The `AnalysisServer` is very incomplete, so I'll keep `Arc<str>` in mind when I rewrite and expand it.

crates/rome_lsp/src/server.rs
```diff
@@ -0,0 +1,309 @@
+use std::sync::Arc;
+
+use lspower::jsonrpc::Result;
+use lspower::lsp::*;
+use lspower::{Client, LanguageServer, LspService, Server};
+use rome_analyze::{AnalysisServer, FileId, Signal, TextAction};
+use rslint_parser::TextRange;
+use tokio::io::{Stdin, Stdout};
+use tokio::sync::Mutex;
```
leops, line 9: It doesn't look like the state lock is ever retained across an await, so it should be okay to use `std::sync::Mutex` for this and it might have better performance characteristics without the async overhead

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+impl AnalysisServer {
+	pub fn new() -> Self {
+		Self {
+			file_map: Default::default(),
```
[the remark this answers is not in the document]
yassere: The idea is that an `AnalysisServer` never does any of its own I/O. It should be "driven" by something else and then queried for an analysis. (This is based on the `AnalysisHost` from `rust-analyzer`).

So in the case of a CLI, it would create a new `AnalysisServer` and then call `set_file_text` for each file.

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -0,0 +1,99 @@
...
+	pub fn find_node_at_range<T: AstNode>(&self, file_id: FileId, range: TextRange) -> Option<T> {
+		trace!("Find {:?} range: {:?}", std::any::type_name::<T>(), range);
+		let tree = self.parse(file_id);
+		tree.covering_element(range).ancestors().find_map(T::cast)
+	}
+
+	pub fn assists(&self, file_id: FileId, cursor_range: TextRange) -> AnalyzerResult {
+		trace!("Assists range: {:?}", cursor_range);
+		let a_ctx = AnalyzerContext::new(self, file_id);
+		let ctx = AssistContext::new(&a_ctx, cursor_range);
```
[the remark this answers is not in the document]
yassere: Yeah, I'm not super happy about any of the names in this entire module. This particular bit of awkwardness will be gone in my next PR though.

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -0,0 +1,99 @@
...
+
+		let mut signals = vec![];
+
+		for provider in assists::all() {
+			let analyze_fn = provider.analyze;
+			if let Some(signal) = analyze_fn(&ctx) {
+				signals.extend(signal.signals.into_iter())
```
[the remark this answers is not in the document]
yassere: Originally, an analyzer returned a single `Signal`. Now it returns an `Analysis` which contains multiple signals, but I forgot to change the variable name here. Thanks for catching it.

crates/rome_analyze/src/analyzers.rs
```diff
@@ -0,0 +1,61 @@
...
+	vec![
+		no_double_equals::create(),
+		no_var::create(),
+		use_while::create(),
+		use_single_case_statement::create(),
+	]
+});
+
+pub struct Analyzer {
+	pub name: &'static str,
```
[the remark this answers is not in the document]
yassere: Right now, an analyzer name is just a string literal that's part of an analyzer's `create` method. This could turn into a `String` or something later.

PR #1915, opened 2022-01-10
feat(rslint_parser): export syntax

MichaReiser:
## Summary

Rewrite of the JS export syntax in the rslint parser as described in [this discussion](https://github.com/rome/tools/discussions/1868#discussioncomment-1785155) but with slightly different naming. 

The new AST allows

* for optional `id`s in `export default` class or function declarations
* JS `assert` for `export ... from` 
* Correctly groups the optionality of child nodes. For example, the `as` keyword with the `alias` are grouped together into a single node.
* Uses our new naming

## Test Plan
I added plenty of tests to cover the new syntax. 

### Regressions

There are 5 regressions. These tests passed before because of unrelated parsing errors. I suggest that we fix these errors separately. The problem is, that we need to verify if the identifiers only use valid unicode escape sequences and error otherwise. The escape sequences are currently handled as part of the lexer and not exposed in the parser.


crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -105,298 +102,6 @@ pub(crate) fn maybe_ts_type_annotation(p: &mut Parser) -> Option<Range<usize>> {
 	}
 }
 
-pub(crate) fn ts_declare(p: &mut Parser) -> Option<CompletedMarker> {
```
MichaReiser, line 108: These methods were no longer used, that's why I deleted them. We can re-add them when working on TS support.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -50,31 +55,68 @@ pub(super) fn parse_class_expression(p: &mut Parser) -> ParsedSyntax {
...
+	}
+
+	let m = p.start();
+	parse_class(p, m, ClassKind::Statement)
+}
+
+// test export_class_clause
+// export class A {}
+// export class A extends B {}
+pub(super) fn parse_export_class_clause(p: &mut Parser) -> ParsedSyntax {
```
xunilrj, line 69: Any reason why this method is not merged with the ```parse_export_default_class_case```? 

```parse_class``` seems generic enough already.

Comment (xunilrj) [reply crates/rslint_parser/src/syntax/class.rs:69]: And I would expect this method to consume the "export" token.

**Review by xunilrj — APPROVED** [review]

**Comment by xunilrj** [comment]
I linked this issue on https://github.com/rome/tools/issues/1909 as a remainder.

File: crates/rslint_parser/src/syntax/class.rs (hunk 2 above)
Comment (MichaReiser) [reply crates/rslint_parser/src/syntax/class.rs:69]: `export class A` requires an id (`parse_export_class_clause`) but `export default class {}` does not (`parse_export_default_class_case`).

The export keyword is consumed by `JsExport` so that we have one single export statement.

**Comment by MichaReiser** [comment]
@xunilrj the tests are now failing because of the `unwrap` call in `raw_lookahead_nth`. I don't think we should call `unwrap` but instead `None`. Would you mind having a look at it?

File: crates/rslint_parser/src/syntax/expr.rs [hunk 3]
```diff
@@ -1134,10 +1141,14 @@ fn parse_identifier_expression(p: &mut Parser) -> ParsedSyntax {
 // test_err identifier
 // yield;
 // await;
-fn parse_reference_identifier(p: &mut Parser) -> ParsedSyntax {
+pub(crate) fn parse_reference_identifier(p: &mut Parser) -> ParsedSyntax {
 	parse_identifier(p, JS_REFERENCE_IDENTIFIER)
 }
 
+pub(crate) fn is_nth_at_reference_identifier(p: &Parser, n: usize) -> bool {
```
ematipico, line 1148: Should we `#[inline]` this too? Or it is not inline'd because of `n`?

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -446,7 +461,362 @@ fn parse_import_assertion_entry(
...
+	let m = p.start();
+
+	p.bump(T!['{']);
+	ExportNamedFromSpecifierList.parse_list(p);
+	p.expect(T!['}']);
+
+	expect_keyword(p, "from", T![from]);
+
+	parse_module_source(p).or_add_diagnostic(p, expected_module_source);
+	parse_import_assertion(p).ok();
```
ematipico, line 683: Why the import assertion is `.ok()` here?

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -446,7 +461,362 @@ fn parse_import_assertion_entry(
...
+	}
+
+	let start = p.cur_tok().range().start;
+	let m = p.start();
+	p.expect(T![*]);
+
+	parse_export_as_clause(p).ok();
+	expect_keyword(p, "from", T![from]);
+	parse_module_source(p).or_add_diagnostic(p, expected_module_source);
+	parse_import_assertion(p).ok();
```
ematipico, line 649: Why the import assertion is `.ok()` here?

ematipico (approved):
Not much to add, just few questions. Looks good to me!

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: I'm starting to become a bit more defensive about using `inline`. We probably have it in many cases without it having any effect.

crates/rslint_parser/src/syntax/module.rs
MichaReiser: The syntax factory will add the missing slot.

xunilrj:
> @xunilrj the tests are now failing because of the `unwrap` call in `raw_lookahead_nth`. I don't think we should call `unwrap` but instead `None`. Would you mind having a look at it?

I will fix this.

> I also noticed that `next_non_trivia` still accepts an `int` for dir, even though it doesn't work with any value other than `1` and `-1`. Should we change it to a bool?

Yes.

xunilrj:
I tested this change in your branch and it seems to pass all testes now: https://github.com/rome/tools/pull/1928

crates/rslint_parser/src/syntax/expr.rs
xunilrj: I was analyzing the generated assembly yesterday and... the compiler is aggressively inlining a lot of functions... I was a little bit surprised. Even functions without ```#[inline(always)]```.

PR #1916, opened 2022-01-10
feat(rome_formatter): retain comments in the formatter output

leops:
## Summary

This reimplements the formatting of comments from Rome Classic on the new CST model, where comments are attached as either leading or trailing trivias on tokens instead or leading / inner / trailing comments on nodes.

In general this means for a document to be formatted correctly the formatter needs to see every token in the input: this is enforced by the formatter now being stateful and storing a set of all printed comments and tokens and checking all the input tokens were actually processed in debug mode. To help with this I introduced the `format_or_create_token` and `format_replaced_token` to help formatters handle optional tokens or replace an input token with a custom representation while retaining all the attached comments respectively.

I also introduced a `format_delimited_group` specifically to handle constructs of `<opening token> <indent group> <closing tokens>` as the trailing comment of the opening token and the leading comment of the closing token both need to be pushed inside the indent group while the tokens themselves are not. It ended up being quite verbose as many formatters handle this slightly differently so maybe I could introduce shortcut methods for the common cases like `format_delimited_block_indent` or `format_delimited_soft_indent`.

Finally, I changed the semantics of `space_token` and `hard_line_break` for accommodate for individual node formatters not knowing whether and where comments are printed. These were previously implemented as "print a space or newline character unconditionally" but now are interpreted by the printer as "the previous and next element should be separated by a whitespace / newline". Functionally this means multiple consecutive occurrences of these tokens will only print once: newlines are only printed if the "insertion cursor" is not already on an empty line, and spaces are printed if the cursor is not preceded by a whitespace (space or newline). Possible additional tokens to supplement this would be `empty_line()` (force the previous and next element to be separated by an empty line) and `align(column)` (force the printer to insert enough spaces for the next element to be printed on the given column)

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -864,6 +864,12 @@ impl From<Indent> for FormatElement {
     }
 }
 
+impl From<Option<FormatElement>> for FormatElement {
+    fn from(element: Option<FormatElement>) -> Self {
+        element.unwrap_or_else(empty_element)
+    }
+}
+
 #[cfg(test)]
 mod tests {
 
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -155,6 +155,17 @@ pub fn token<S: Into<String>>(text: S) -> FormatElement {
     }
 }
 
+/// Push a [FormatElement] to the end of the current line
+///
+/// ## Examples
+///
+/// ```
+/// use rome_formatter::{FormatOptions, token, format_element, line_suffix, format_elements};
+///
+/// let elements = format_elements![token("a"), line_suffix(token("c")), token("b")];
+///
+/// assert_eq!("abc", format_element(&elements, FormatOptions::default()).code());
+/// ```
 #[inline]
 pub fn line_suffix(element: impl Into<FormatElement>) -> FormatElement {
     FormatElement::LineSuffix(Box::new(element.into()))
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -155,6 +155,11 @@ pub fn token<S: Into<String>>(text: S) -> FormatElement {
     }
 }
 
+#[inline]
+pub fn line_suffix(element: impl Into<FormatElement>) -> FormatElement {
+    FormatElement::LineSuffix(Box::new(element.into()))
+}
+
 /// Inserts a single space. Allows to separate different tokens.
 ///
 /// ## Examples
@@ -608,6 +613,8 @@ pub enum FormatElement {
 
     /// A token that should be printed as is, see [token] for documentation and examples.
     Token(Token),
+
+    LineSuffix(Box<FormatElement>),
 }
 
 /// Inserts a new line
@@ -766,6 +773,7 @@ impl FormatElement {
                 FormatElement::List(List::new(content))
             }
             FormatElement::Token(s) => token(s.trim_start()),
+            FormatElement::LineSuffix(s) => FormatElement::LineSuffix(Box::new(s.trim_start())),
         }
     }
 
@@ -810,6 +818,7 @@ impl FormatElement {
                 }
             }
             FormatElement::Token(s) => token(s.trim_end()),
+            FormatElement::LineSuffix(s) => FormatElement::LineSuffix(Box::new(s.trim_end())),
         }
     }
 }
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -254,12 +254,10 @@ impl Formatter {
                 } else {
                     self.format_token(&separator)?
                 }
+            } else if index == last_index {
+                if_group_breaks(separator_factory())
             } else {
-                if index == last_index {
-                    if_group_breaks(separator_factory())
-                } else {
-                    separator_factory()
-                }
+                separator_factory()
             };
 
             result.push(format_elements![node, separator]);
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -47,6 +47,29 @@ impl Formatter {
         ]))
     }
 
+    /// Formats a group delimited by an opening and closing token,
+    /// such as a function body delimited by '{' and '}' tokens
+    ///
+    /// Calling this method is required to correctly handle the comments attached
+    /// to the opening and closing tokens and insert them inside the group block
+    pub(crate) fn format_delimited_group(
+        &self,
+        open_token: &SyntaxToken,
+        content: impl FnOnce(FormatElement, FormatElement) -> FormatResult<FormatElement>,
+        close_token: &SyntaxToken,
+    ) -> FormatResult<FormatElement> {
+        Ok(format_elements![
+            self.print_trivia(open_token.leading_trivia()),
+            token(open_token.text_trimmed()),
+            content(
+                self.print_trivia(open_token.trailing_trivia()),
+                self.print_trivia(close_token.leading_trivia()),
+            )?,
+            token(close_token.text_trimmed()),
+            self.print_trivia(close_token.trailing_trivia()),
+        ])
+    }
+
     /// Recursively formats the ast node and all its children
     ///
     /// Returns `None` if the node couldn't be formatted because of syntax errors in its sub tree.
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -1,9 +1,12 @@
+use crate::format_element::line_suffix;
 use crate::printer::Printer;
 use crate::{
-    concat_elements, format_elements, token, FormatElement, FormatOptions, FormatResult, Formatted,
-    ToFormatElement,
+    concat_elements, empty_element, format_elements, hard_line_break, if_group_breaks,
+    if_group_fits_on_single_line, join_elements, space_token, token, FormatElement, FormatOptions,
+    FormatResult, Formatted, ToFormatElement,
 };
-use rome_rowan::SyntaxElement;
+use rome_rowan::api::{SyntaxTrivia, SyntaxTriviaPieceComments};
+use rome_rowan::{Language, SyntaxElement};
 use rslint_parser::{AstNode, AstSeparatedList, SyntaxNode, SyntaxToken};
 
 /// Handles the formatting of a CST and stores the options how the CST should be formatted (user preferences).
@@ -52,25 +55,27 @@ impl Formatter {
         &self,
         node: T,
     ) -> FormatResult<FormatElement> {
-        Ok(concat_elements(vec![
-            self.format_node_start(node.syntax()),
+        let leading = self.format_node_start(node.syntax());
+        let trailing = self.format_node_end(node.syntax());
+        Ok(format_elements![
+            leading,
             node.to_format_element(self)?,
-            self.format_node_end(node.syntax()),
-        ]))
+            trailing,
+        ])
     }
 
     /// Helper function that returns what should be printed before the node that work on
     /// the non-generic [SyntaxNode] to avoid unrolling the logic for every [AstNode] type.
     fn format_node_start(&self, _node: &SyntaxNode) -> FormatElement {
-        // TODO: Set the marker for the start source map location, add leading comments, ...
-        concat_elements(vec![])
+        // TODO: Set the marker for the start source map location, ...
+        empty_element()
     }
 
     /// Helper function that returns what should be printed after the node that work on
     /// the non-generic [SyntaxNode] to avoid unrolling the logic for every [AstNode] type.
     fn format_node_end(&self, _node: &SyntaxNode) -> FormatElement {
-        // TODO: Sets the marker for the end source map location, add trailing comments, ...
-        concat_elements(vec![])
+        // TODO: Sets the marker for the end source map location, ...
+        empty_element()
     }
 
     /// Formats the passed in token.
@@ -100,7 +105,30 @@ impl Formatter {
     /// assert_eq!(Ok(token("'abc'")), result)
     /// ```
     pub fn format_token(&self, syntax_token: &SyntaxToken) -> FormatResult<FormatElement> {
-        Ok(token(syntax_token.text_trimmed()))
+        Ok(format_elements![
+            self.print_leading_trivia(syntax_token),
+            token(syntax_token.text_trimmed()),
+            self.print_trailing_trivia(syntax_token),
+        ])
+    }
+
+    pub fn format_or_create_token(
+        &self,
+        token: Option<SyntaxToken>,
+        token_factory: impl FnOnce() -> FormatElement,
+    ) -> FormatResult<FormatElement> {
+        if let Some(token) = token {
+            self.format_token(&token)
+        } else {
+            Ok(token_factory())
+        }
+    }
+
+    pub fn format_comment<L: Language>(
+        &self,
+        trivia: SyntaxTriviaPieceComments<L>,
+    ) -> FormatElement {
+        token(trivia.text().trim())
     }
 
     /// Formats each child and returns the result as a list.
@@ -133,11 +161,15 @@ impl Formatter {
         for (index, element) in list.elements().enumerate() {
             let node = self.format_node(element.node()?)?;
             if let Some(separator) = element.trailing_separator()? {
-                let formatted_separator = self.format_token(&separator)?;
                 if index == list.len() - 1 {
-                    result.push(node)
+                    result.push(node);
+                    // Print the trivia for the last trailing separator token
+                    // without printing the token itself since that depends
+                    // on the outer group being broken or not
+                    result.push(self.print_leading_trivia(&separator));
+                    result.push(self.print_trailing_trivia(&separator));
                 } else {
-                    result.push(format_elements![node, formatted_separator]);
+                    result.push(format_elements![node, self.format_token(&separator)?]);
                 }
             } else {
                 result.push(node);
@@ -145,6 +177,124 @@ impl Formatter {
         }
 
         Ok(result.into_iter())
+    }
+
+    fn print_leading_trivia(&self, token: &SyntaxToken) -> FormatElement {
+        let is_leading_token = token
+            .parent()
+            .and_then(|parent| parent.first_token())
+            .map_or(false, |first_token| *token == first_token);
+
+        if !is_leading_token {
+            return self.print_inner_trivia(token.leading_trivia());
+        }
+
+        // False positive: the trivias need to be collected in a vector as they
+        // are iterated on in reverse order later, but SyntaxTriviaPiecesIterator
+        // doesn't implement DoubleEndedIterator (rust-lang/rust-clippy#8132)
+        #[allow(clippy::needless_collect)]
+        let pieces: Vec<_> = token.leading_trivia().pieces().collect();
+
+        let mut line_count = 0;
+        let mut elements = Vec::new();
+
+        for piece in pieces.into_iter().rev() {
+            if let Some(comment) = piece.as_comments() {
+                let is_single_line = comment.text().trim_start().starts_with("//");
+
+                let comment = self.format_comment(comment);
+                if !comment.is_empty() {
+                    let line_break = if is_single_line {
+                        hard_line_break()
+                    } else {
+                        match line_count {
+                            0 => space_token(),
+                            1 => hard_line_break(),
+                            _ => format_elements![hard_line_break(), hard_line_break()],
+                        }
+                    };
+
+                    elements.push(format_elements![comment, line_break]);
+                }
+
+                line_count = 0;
+            }
+
+            if piece.as_newline().is_some() {
+                line_count += 1;
+            }
+        }
+
+        concat_elements(elements.into_iter().rev())
+    }
+
+    fn print_trailing_trivia(&self, token: &SyntaxToken) -> FormatElement {
+        let is_trailing_token = token
+            .parent()
+            .and_then(|parent| parent.last_token())
+            .map_or(false, |last_token| *token == last_token);
+
+        if !is_trailing_token {
+            return self.print_inner_trivia(token.trailing_trivia());
+        }
+
+        let mut line_count = 0;
+        let mut elements = Vec::new();
+
+        for piece in token.trailing_trivia().pieces() {
+            if let Some(comment) = piece.as_comments() {
+                let is_single_line = comment.text().trim_start().starts_with("//");
+
+                let comment = self.format_comment(comment);
+                if !comment.is_empty() {
+                    elements.push(if line_count >= 1 {
+                        line_suffix(format_elements![
+                            if line_count > 1 {
+                                hard_line_break()
+                            } else {
+                                space_token()
+                            },
+                            hard_line_break(),
+                            comment
+                        ])
+                    } else if !is_single_line {
+                        format_elements![
+                            if_group_breaks(line_suffix(format_elements![
+                                space_token(),
+                                comment.clone()
+                            ])),
+                            if_group_fits_on_single_line(format_elements![space_token(), comment]),
+                        ]
+                    } else {
+                        line_suffix(format_elements![space_token(), comment])
+                    });
+                }
+
+                line_count = 0;
+            }
+
+            if piece.as_newline().is_some() {
+                line_count += 1;
+            }
+        }
+
+        concat_elements(elements)
+    }
+
+    fn print_inner_trivia<L: Language>(&self, trivia: SyntaxTrivia<L>) -> FormatElement {
+        join_elements(
+            hard_line_break(),
+            trivia.pieces().filter_map(|piece| {
+                piece.as_comments().and_then(|comments| {
+                    let comments = self.format_comment(comments);
+                    if !comments.is_empty() {
+                        Some(format_elements![space_token(), comments, hard_line_break()])
+                    } else {
+                        None
+                    }
+                })
+            }),
+        )
     }
 
     /// "Formats" a node according to its original formatting in the source text. Being able to format
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -75,7 +75,7 @@ impl Formatter {
     ///
     /// Calling this method is required to correctly handle the comments attached
     /// to the opening and closing tokens and insert them inside the group block
-    pub(crate) fn format_delimited_group(
+    pub(crate) fn format_delimited(
         &self,
         open_token: &SyntaxToken,
         content: impl FnOnce(FormatElement, FormatElement) -> FormatResult<FormatElement>,
@@ -153,24 +153,18 @@ impl Formatter {
     ///
     /// assert_eq!(Ok(token("'abc'")), result)
     /// ```
-    pub fn format_token(&self, syntax_token: &SyntaxToken) -> FormatResult<FormatElement> {
-        debug_assert!(self
-            .printed_tokens
-            .borrow_mut()
-            .insert(syntax_token.clone()));
-
-        Ok(format_elements![
-            self.print_leading_trivia(syntax_token),
-            token(syntax_token.text_trimmed()),
-            self.print_trailing_trivia(syntax_token),
-        ])
+    pub fn format_token<T>(&self, syntax_token: &T) -> FormatResult<T::Output>
+    where
+        T: token::FormattableToken,
+    {
+        syntax_token.format(self)
     }
 
     /// Print out a token from the original source with a different content
     ///
     /// This will print the associated trivias from the token as well as mark
     /// it as having been consumed by the formatter
-    pub fn format_replaced_token(
+    pub fn format_replaced(
         &self,
         token: &SyntaxToken,
         content: FormatElement,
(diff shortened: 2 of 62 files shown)


Boshen:
If you look closely, some of the snapshots are a bit off. 
You can copy the snapshots into the Prettier Playground to test things out.

I did a small commit with this and here is the discussion: https://github.com/rome/tools/pull/1882#issuecomment-996828249

crates/rome_formatter/tests/specs/js/array/empty_lines.js.snap
```diff
@@ -19,5 +19,14 @@ let a = [
...
-let a = [1, 2, 3, 4];
+let a = [
+
+	1,
+	2,
+
+	3,
+
+	4,
+
```
MichaReiser, line 30: In my view, the empty line after the last element should be removed.

crates/rome_formatter/tests/specs/js/arrow/arrow_nested.js.snap
```diff
@@ -41,18 +41,19 @@ Seq(typeDef.interface.groups).forEach((group) => Seq(group.members).forEach((
...
 }));
-runtimeAgent.getProperties(objectId, false, false, false, (
-	error,
-	properties,
-	internalProperties,
-) => {
+
+runtimeAgent.getProperties(objectId, false,// ownProperties
+ false,// accessorPropertiesOnly
+ false,// generatePreview
```
MichaReiser, line 55: It seems we're losing the whitespace between the token and the comment.

MichaReiser (commented):
This is a big improvement to what we had so far. 

I'm not sure if we can implement the collapsing of new lines generically, or at least, if we would need yet another IR element to do the collapsing. 

The problem is, there are some new lines that we want to remove completely, for example, any new line before the closing parentheses (so that these get deleted when I removed 1-2 elements). 

It probably makes sense to split this PR into two: One for comments and another for new lines

crates/rome_formatter/src/formatter.rs
```diff
@@ -165,3 +172,28 @@ impl Formatter {
...
+				.count();
+
+			if newlines > 1 {
+				elements.push(empty_line());
+			}
+		}
+
+		if let Some(comments) = piece.as_comments() {
+			let comments_text = comments.text();
+			let is_single_line = comments_text.starts_with("//");
```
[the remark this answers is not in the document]
leops: With this version comments are pretty much dropped in place in the output. I'm currently revising this behavior, and for "multiline comments" it will depend on whether they are actually multiline (contain at least one linebreak):

- single line comments will be preceded with a `soft_line_break_or_space` and followed by a `hard_line_break`
- single line block comments will be preceded and followed by a `soft_line_break_or_space`
- multi line block comments will be preceded and followed by a `hard_line_break`

(Also I'd probably want this check to rely on the new `MULTILINE_COMMENT` instead of looking up the text content)

crates/rome_formatter/tests/specs/js/number/number.js.snap
```diff
@@ -10,6 +10,6 @@ expression: number.js
 ---
 # Output
 1.23e4;
-1000e3// FIXME handle number with scientific notation #1294
+1000e3 // FIXME handle number with scientific notation #1294
 ;
```
Boshen, line 14: This `;` ended up on a new line

crates/rome_formatter/src/formatter.rs
```diff
@@ -165,3 +171,27 @@ impl Formatter {
 		}))
 	}
 }
+
+fn print_trivia<L: Language>(elements: &mut Vec<FormatElement>, trivia: SyntaxTrivia<L>) {
+	for piece in trivia.pieces() {
+		if let Some(comments) = piece.as_comments() {
+			let comments_text = comments.text().trim();
+			let is_single_line = comments_text.starts_with("//");
+			let is_multi_line = !is_single_line && comments_text.contains("\n");
```
ematipico, line 180: What about `/** **/` comments? How should we treat them? Consider that such comment can also on a single line

crates/rome_formatter/src/ts/class/property_class_member.rs
```diff
@@ -18,11 +18,17 @@ impl ToFormatElement for JsPropertyClassMember {
 			empty_element()
 		};
 
+		let semicolon = if let Some(semicolon) = self.semicolon_token() {
+			formatter.format_token(&semicolon)?
+		} else {
+			token(";")
+		};
```
ematipico, line 25: We have this pattern which is repeated in different places. Can't we have a function `format_semicolon` inside `formatter`?

crates/rome_formatter/tests/specs/js/arrow/currying.js.snap
```diff
@@ -34,5 +34,6 @@ const bar = (a) => (b) => (c) => a + b + c;
 const mw = (store) => (next) => (action) => {
 	return next(action);
 };
-const middleware = (options) => (req, res, next) => {};
+const middleware = (options) => (req, res, next) => { // ...
+};
```
ematipico, line 38: I think, in this case, we should move the comment in a newline and indent it.

ematipico (changes requested):

crates/rome_formatter/src/formatter.rs
```diff
@@ -44,6 +69,32 @@ impl Formatter {
 		]))
 	}
 
+	/// Formats a group delimited by an opening and closing token,
+	/// such as a function body delimited by '{' and '}' tokens
```
ematipico, line 73:
```suggestion
	/// such as a function body delimited by '{' and '}' tokens.
```

crates/rome_formatter/src/formatter.rs
```diff
@@ -100,7 +153,52 @@ impl Formatter {
...
+	) -> FormatResult<FormatElement> {
+		debug_assert!(self.printed_tokens.borrow_mut().insert(token.clone()));
+		Ok(format_elements![
+			self.print_leading_trivia(token),
+			content,
+			self.print_trailing_trivia(token),
+		])
+	}
+
+	pub fn format_or_create_token(
```
ematipico, line 185: Doc missing

crates/rome_formatter/src/formatter.rs
```diff
@@ -133,12 +231,16 @@ impl Formatter {
 		for (index, element) in list.elements().enumerate() {
 			let node = self.format_node(element.node()?)?;
 			if let Some(separator) = element.trailing_separator()? {
-				let formatted_separator = self.format_token(&separator)?;
-				if index == list.len() - 1 {
-					result.push(node)
+				let separator = if index == list.len() - 1 {
+					// Print the last trailing separator token is replaced
+					// with an empty element since the logic of printing the
+					// last token depends on the group being borken or not
```
ematipico, line 237: This comment is really hard to read, is there some punctuation missing?

crates/rome_formatter/src/formatter.rs
```diff
@@ -161,7 +381,36 @@ impl Formatter {
...
+				token(syntax_token.text())
+			}
 		}))
 	}
 }
+
+pub struct FormatterSnapshot {
+	#[cfg(debug_assertions)]
+	printed_tokens: HashSet<SyntaxToken>,
+}
```
ematipico, line 399: Will this struct have 0 size when it's compiled with the `--release` flag?

ematipico (approved):

crates/rome_formatter/src/formatter.rs
leops: I think I edited the content of this comment a few times and it got really confusing, I'll just completely rewrite it. The idea is that in the current design, the formatter for a node is responsible for adding for example a `if_group_breaks(token(","))` after the separated elements, so if there was a trailing separator at the end it gets discarded but we still want to print the comments that were attached to it

crates/rome_formatter/src/formatter.rs
leops: Yes that's the general idea, the snapshoting system is only needed to handle the logic backtracking on invalid nodes so that tokens don't get processed twice, but it will compile to nothing in release mode and the optimizer should remove it entirely

crates/rome_formatter/src/format_element.rs
```diff
@@ -155,6 +155,11 @@ pub fn token<S: Into<String>>(text: S) -> FormatElement {
 	}
 }
 
+#[inline]
+pub fn line_suffix(element: impl Into<FormatElement>) -> FormatElement {
```
MichaReiser, line 159: Documentation

crates/rome_formatter/src/formatter.rs
```diff
@@ -161,7 +381,36 @@ impl Formatter {
...
+		}
+	}
+
+	pub fn restore(&self, snapshot: FormatterSnapshot) {
+		cfg_if::cfg_if! {
+			if #[cfg(debug_assertions)] {
+				*self.printed_tokens.borrow_mut() = snapshot.printed_tokens;
+			}
+		}
+	}
```
MichaReiser, line 415: For what do we need the `snapshot` and `restore` when formatting? Can you add some documentation explaining its use case?

crates/rome_formatter/src/formatter.rs
```diff
@@ -100,7 +153,52 @@ impl Formatter {
...
+		token_factory: impl FnOnce() -> FormatElement,
+	) -> FormatResult<FormatElement> {
+		if let Some(token) = token {
+			self.format_token(&token)
+		} else {
+			Ok(token_factory())
+		}
+	}
+
+	pub fn format_comment<L: Language>(
```
MichaReiser, line 197: Must this be public or can it be private?

crates/rome_formatter/src/formatter.rs
MichaReiser: I'm wondering if this is something that we could implement as an extension to `Option<SyntaxToken>`

crates/rome_formatter/src/formatter.rs
```diff
@@ -44,6 +69,32 @@ impl Formatter {
...
+		debug_assert!(self.printed_tokens.borrow_mut().insert(open_token.clone()));
+		debug_assert!(self.printed_tokens.borrow_mut().insert(close_token.clone()));
+
+		Ok(format_elements![
+			self.print_leading_trivia(open_token),
+			token(open_token.text_trimmed()),
+			content(
+				self.print_trailing_trivia(open_token),
+				self.print_leading_trivia(close_token),
+			)?,
```
MichaReiser, line 92: Should the content always be ut inside a `group`? At least the method name says it's a group? Or should the outer content be grouped?

I need to review this again tomorrow. It's still not entirely clear to me why we need this method vs just using `groups`. It could help to explain in the comment why we need this method.

leops: Most of the time this will be inside a group but this isn't always the case (for instance `JsFunctionBody` doesn't create a group)

Generally this is needed for cases like this:
```js
while (true) { // comment
	break;
}
```

If the `curly_token` are left outside the indent group the comment is printed exactly where it was in the input,
while if both `curly_token` are pushed into the indent group the result becomes this:
```js
while (true)
  { // comment
  break;}
```

With this change only the trailing comment for the opening `curly_token` will be pushed inside the indent group giving:
```js
while (true) {
  // comment
  break;
}
```

crates/rome_formatter/tests/specs/js/arrow/call.js.snap
```diff
@@ -53,13 +53,17 @@ romise.then(result => result.veryLongVariable.veryLongPropertyName > someOtherVa
 # Output
 const testResults = results.testResults.map((testResult) => formatResult(testResult, formatter, reporter));
 it("mocks regexp instances", () => {
-	expect().not.toThrow();
+	expect( // () => moduleMocker.generateFromMetadata(moduleMocker.getMetadata(/a/)),
+	).not.toThrow();
 });
 expect(() => asyncRequest({ url: "/test-endpoint" }));
+// .toThrowError(/Required parameter/);
```
MichaReiser, line 60: Delete?

crates/rome_formatter/src/printer.rs
```diff
@@ -178,9 +182,24 @@ impl Printer {
...
 			FormatElement::Line { .. } => {
-				self.print_str("\n");
-				self.state.pending_spaces = 0;
-				self.state.pending_indent = args.indent;
+				if !self.state.line_suffixes.is_empty() {
+					self.state
+						.line_suffixes
+						.drain(..)
+						.chain(once(PrintElementCall::new(element, args)))
+						.collect()
```
MichaReiser, line 190: You need to explain this line to me. Where's the side effect happening? All we do is to empty the `line_suffices` (`.clear()` might be faster) and then add a `PrintElementCall`. But I don't see where these events are being processed.

crates/rome_formatter/src/printer.rs
```diff
@@ -272,62 +295,62 @@ impl Printer {
 		Ok(next_calls)
 	}
 
-	fn print_str(&mut self, content: &str) {
-		self.state.buffer.reserve(content.len());
```
MichaReiser, line 276: Any reason you removed the `reserve` call?

crates/rome_formatter/src/formatter.rs
```diff
@@ -147,6 +249,124 @@ impl Formatter {
 		Ok(result.into_iter())
 	}
 
+	fn print_leading_trivia(&self, token: &SyntaxToken) -> FormatElement {
+		let is_leading_token = token
+			.parent()
+			.and_then(|parent| parent.first_token())
+			.map_or(false, |first_token| *token == first_token);
```
MichaReiser, line 256: Nit: This is probably a bit more expensive because it calls into `descendants`

```
NodeA [
  NodeB [
    token
  ] 
] 
```

meaning, it would return `token` even if called on `NodeA`. 

`let is_leading_token = token.prev_sibling_or_token().is_none()`

But it depends on the behaviour you want. Almost all nodes have at least one token as its children, not so `JsArrayHole`

```
JsArrayExpression [
  JsArrayHole
  ','
  ]
```

Should `is_leading_token` be true for the `,` or not?

crates/rome_formatter/src/formatter.rs
MichaReiser: I'm wondering if we can replicate the same with `if_group_breaks` and `if_group_fits_on_single_line` (which we probably should rename because it also gets printed if it isn't inside a group).

crates/rome_formatter/tests/specs/js/comments.js.snap
```diff
@@ -0,0 +1,54 @@
...
+statement();
+
+statement(); // inline
+---
+# Output
+import {
+    func, // trailing comma removal
+} from 'module';
+expression( /* block comment */
+);
```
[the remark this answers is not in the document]
ematipico: > It isn't really changing any semantic.

That's because they don't have editor in mind. Try to think about this situation:

```js
expression(  // line comment |
);
```

The pipe is where is the position of the cursor of the IDE. If we format it as Prettier does, we basically changing the whole semantic! The user wanted to continue write after the comment, assuming that they were inside the parenthesis. But nope, they save, they continue to write code and it is... wrong.

crates/rome_formatter/src/printer.rs
leops: The difference between drain and clear is that drain returns an iterator containing the removed elements, this iterator gets chained with a single PrintElementCall and collected into the vector that's returned from `print_element` to be added back on the stack. So these events will eventually be processed by the next few calls to `print_element` when they're popped off the stack

crates/rome_formatter/src/printer.rs
```diff
@@ -272,62 +295,62 @@ impl Printer {
...
+	fn print_str(&mut self, mut content: &str) {
+		while let Some(newline) = content.find('\n') {
```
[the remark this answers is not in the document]
leops: I should just revert this bit, it was part of a test I did on to allow comments to be pushed off to the previous line but I deemed this too complex. The idea was that it would not be too inefficient because find is implemented on top of memchr and printing the line content boils down to a reserve + memcpy, all of which are highly optimized

crates/rome_formatter/src/printer.rs
MichaReiser: Ah, it gets added to the stack... I missed that. Sorry

crates/rome_formatter/src/formatter.rs
```diff
@@ -147,6 +249,124 @@ impl Formatter {
...
+						}
+					};
+
+					elements.push(format_elements![comment, line_break]);
+				}
+
+				line_count = 0;
+			}
+
+			if piece.as_newline().is_some() {
```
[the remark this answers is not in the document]
leops: Currently multiline comments are represented as `TriviaPiece::Comments` so all types of comments are covered by the previous condition

crates/rome_formatter/src/ts/statements/mod.rs
```diff
@@ -29,9 +29,14 @@ pub fn format_statements(stmts: JsStatementList, formatter: &Formatter) -> Forma
 	join_elements(
 		hard_line_break(),
 		stmts.iter().map(|stmt| {
-			formatter
-				.format_node(stmt.clone())
-				.unwrap_or_else(|_| formatter.format_raw(stmt.syntax()).trim_start().trim_end())
+			let snapshot = formatter.snapshot();
```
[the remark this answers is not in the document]
leops: In the current state of things the formatter state is only used to keep track of the tokens that were seen during the current formatting process in debug mode to detect cases where a token is skipped over by a formatter. In practice forgetting to snapshot and restore the state isn't too dangerous as it will not break anything, but the formatter will likely panic in debug mode because it asserts that tokens do not get formatted twice (a first time in format_node and a second time in the fallback call to format_raw)

crates/rome_formatter/src/formatter.rs
leops: That's possible but it might look inconsistent to have `formatter.format_token(token)` and `token.format_or_create(formatter, factory)`

crates/rome_formatter/src/formatter.rs
leops: That as the case because some callers may never want the trailing separator (variable declarators) or may want to conditionally print it (array expression with holes). But I think it's probably better to have `format_element` handle the general case, and have special cases implement their own logic as required

crates/rome_formatter/src/formatter.rs
```diff
@@ -147,6 +249,124 @@ impl Formatter {
...
+						]
+					} else {
+						line_suffix(format_elements![space_token(), comment])
+					});
+				}
+
+				line_count = 0;
+			}
+
+			if piece.as_newline().is_some() {
```
leops, line 346: I'm not sure if this code path is actually useful, I think for trailing trivias if there is newline piece it's always the last one so the line count might always be 0

MichaReiser: It can happen if the comments are attached to the EOF token. For example if you have

```
let a = 10;

// Some trailing comma
// and another

// and some more
````

Comment (leops) [reply crates/rome_formatter/src/formatter.rs:346]: Oh I thought these were leading trivias, I was thinking that the EOF token couldn't have trailing trivias since that would be "content after the end of the file"

File: crates/rome_formatter/src/formatter.rs (hunk 19 above)
Comment (leops) [reply crates/rome_formatter/src/formatter.rs:256]: Actually with the changes I've made to the semantics of whitespace tokens this isn't needed anymore, I think the comment formatter can just emit space or newline tokens unconditionally and doesn't need to handle "inner" comments in any special way

File: crates/rome_formatter/src/formatter.rs (hunk 24 above)
Comment (MichaReiser) [reply crates/rome_formatter/src/formatter.rs:349]: Oh right. These are leading trivia.

File: crates/rome_formatter/src/ts/class/method_class_member.rs [hunk 25]
```diff
@@ -11,11 +11,13 @@ impl ToFormatElement for JsMethodClassMember {
 		} else {
 			empty_element()
 		};
+		let star_token = formatter.format_or_create_token(self.star_token(), empty_element)?;
```
MichaReiser, line 14: It seems that this is a normal optional token because `format_or_crate` token with an `empty_element` is a no-op if the element isn't present?

```rs
let star_token = if let Some(star_token) = self.star_token() {
    formatter.format_token(&star_token) 
} else {
    empty_element()
}
```

crates/rome_formatter/src/ts/expressions/array_expr.rs
```diff
@@ -2,30 +2,54 @@ use crate::{
...
+				// formatting since this makes a semantic difference
+				let last_index = elements.len().saturating_sub(1);
+				let results = elements
+					.elements()
+					.enumerate()
+					.map(|(index, element)| {
+						let node = formatter.format_node(element.node()?)?;
+
+						let separator = if node.is_empty() || index != last_index {
+							// If the previous element was empty or this is not the last element, always print a separator
```
MichaReiser, line 28: Using `node.is_empty` works but isn't very semantic. How about matching on `element.node()?` because the case your describing here is if the last element is a `JsArrayHole` for which we could test explicitly.

crates/rome_formatter/src/ts/expressions/call_expression.rs
```diff
@@ -1,11 +1,15 @@
-use crate::{format_elements, FormatElement, FormatResult, Formatter, ToFormatElement};
+use crate::{
+	empty_element, format_elements, FormatElement, FormatResult, Formatter, ToFormatElement,
+};
 use rslint_parser::ast::JsCallExpression;
 
 impl ToFormatElement for JsCallExpression {
 	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
 		let name = formatter.format_node(self.callee()?)?;
+		let option =
```
MichaReiser, line 9: This is an optional token, we shouldn't need to use `format_or_create_token` (a helper on `Option<FormatElement>` or implement `ToFormatElement` for `Option<ToFormatElement>` might be helpful tough)). 

```
let optional = if let Some(optional) = self.optional_chain_token() {
    formatter.format_token(&optional)?
} else { 
    empty_element()
};
```

crates/rome_formatter/src/ts/statements/break_statement.rs
```diff
@@ -12,10 +12,12 @@ impl ToFormatElement for JsBreakStatement {
 			empty_element()
 		};
 
+		let semicolon = formatter.format_or_create_token(self.semicolon_token(), || token(';'))?;
```
MichaReiser, line 15: Lol, I get it now... the `||` is for a lambda function. It reads as if it's the or operator.

crates/rome_formatter/src/ts/statements/debugger_statement.rs
```diff
@@ -5,7 +5,7 @@ impl ToFormatElement for JsDebuggerStatement {
 	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
 		Ok(format_elements![
 			formatter.format_token(&self.debugger_token()?)?,
-			token(";")
+			formatter.format_or_create_token(self.semicolon_token(), || token(';'))?
```
MichaReiser, line 8: Nit: Something we can track in a future PR

Would be nice if the `formatter` doesn't require different methods to support inserting a token or adding a new one if it's missing but that this are helpers defined elsewhere. However, I don't know a good place right now. Ideally, it would be something like `formatter.format_token(self.semicolon_token()).or_else(|| token(';'))` but that would require changing `format_token` to return an `Option` instead and implement `ToFormatElement` for `Option<T> where T: ToFormatElement`. Let's leave this for another PR / improvement.

crates/rome_formatter/src/ts/statements/empty_statement.rs
```diff
@@ -2,7 +2,7 @@ use crate::{empty_element, FormatElement, FormatResult, Formatter, ToFormatEleme
 use rslint_parser::ast::JsEmptyStatement;
 
 impl ToFormatElement for JsEmptyStatement {
-	fn to_format_element(&self, _formatter: &Formatter) -> FormatResult<FormatElement> {
-		Ok(empty_element())
+	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		formatter.format_replaced_token(&self.semicolon_token()?, empty_element())
```
MichaReiser, line 6:
```
if let Some(semicolon) = self.semicolon_token() {
    formatter.format_token(semicolon)
} else {
    empty_element()
}
```

or implement `ToFormatElement` for `Option<T> where T: ToFormatElement`

crates/rome_formatter/src/ts/statements/mod.rs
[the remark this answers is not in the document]
MichaReiser: Yeah. I also don't know of a better approach right now. I considered changing `token` to accept a `SyntaxToken` but it would then require additional IR elements to model `ReplacedToken` and `SkippedToken`. Let's go with what we have for now and re-visit it, for example, when working on CST to CST printing (maybe something you're interested in?)

crates/rome_formatter/src/ts/statements/variable_declaration_statement.rs
```diff
@@ -1,28 +1,66 @@
...
 		])
 	}
 }
 
 impl ToFormatElement for JsVariableDeclarations {
 	fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+		let items = self.items();
+		let last_index = items.len().saturating_sub(1);
+
+		let items = items
```
MichaReiser, line 25: What's the reason we can't use `format_separated` for variable declarations?

crates/rome_formatter/src/formatter.rs
MichaReiser: What I'm thinking of is

```
impl<T> ToFormatElement for Option<T> where T: ToFormatElement {
    fn to_format_element(&self, formatter: &Formatter) -> FormatElement {
        match self {
            Some(item) => item.to_format_element(formatter),
            None => empty_element()
        }
    }
}
```

This is still consistent in my view because formatting an `Option` becomes formatting any other *formattable* content.

crates/rome_formatter/src/formatter.rs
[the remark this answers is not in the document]
MichaReiser: Right, but is it important to test if the comment contains a line breeak and if so, increase `line_count`?

Ah, no, it isn't needed. Can we change the `if piece.as_newline` into an `else if...` to make that clear. Would have been nice if `SyntaxTriviaPiece` is an `enum` so that you could match but this isn't the case.

MichaReiser (commented):
This is a huge improvement from what we had before. Thanks for tackling this!

The overall approach looks good to me. I've a few more comments related to the API and usage inside of the formatter. But we should then be good to go :)

crates/rome_formatter/src/formatter.rs
ematipico: I like Micha's suggestion! And we should be safe because an optional element should always be replaced by an empty element, which is "fixed".

crates/rome_formatter/src/ts/bindings/array_binding_pattern.rs
```diff
@@ -1,21 +1,24 @@
...
 
-		Ok(group_elements(format_elements!(
-			formatter.format_token(&self.l_brack_token()?)?,
-			soft_indent(format_elements![
-				join_elements(soft_line_break_or_space(), elements),
-				if_group_breaks(token(",")),
-			]),
-			formatter.format_token(&self.r_brack_token()?)?,
-		)))
+		Ok(group_elements(formatter.format_delimited_group(
```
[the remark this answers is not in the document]
leops: I agree the naming here is tricky: `group` doesn't necessarily mean an actual `FormatElement::Group` but a number of elements that are "semantically grouped" inside a block delimited by an opening and closing token, and these tokens can be parenthesis (arguments lists), brackets (array expressions) or curlies (object expression)

crates/rome_formatter/src/ts/statements/variable_declaration_statement.rs
leops: Besides never wanting a trailing separator to be printed, the variables declarations after the first one are put in an indent group (the first declaration is printed inline after the var/let/const keyword)

MichaReiser (approved):

crates/rome_formatter/src/ts/statements/debugger_statement.rs
leops: I solved this by introducing a `FormattableToken` trait implemented by `SyntaxToken` and `Option<SyntaxToken>` so that `format_token` returns a `FormatElement` and `Option<FormatElement>` respectively depending on what type it's called with. I also implemented `From<Option<FormatElement>>` for `FormatElement` to turn `None` into `FormatElement::Empty`

crates/rome_formatter/src/ts/statements/break_statement.rs
leops: This one is especially ambiguous since there is an "or" operation at work, hopefully with the new generic `format_token(optional).unwrap_or_else(fallback)` it should be clearer

crates/rome_formatter/src/ts/bindings/array_binding_pattern.rs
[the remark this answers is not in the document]
leops: I ended up renaming it to `format_delimited` (and `format_replaced_token` to `format_replaced`) so that they line up with the existing `format_separated`

PR #1917, opened 2022-01-10
feat(rslint_parser): add an error diagnostic to for loops initializer

leops:
## Summary

~`for..in` and `for..of` loops declarators are not allowed to have an initializer in strict mode except in `for(var ... in ...)` statements~

Edit: While writing the test cases I realized my wording was wrong (although the code is correct): initializers are never allowed in strict mode, and loose mode only allows `for(var ... in ...)` but none of the others

### DIFF
--- a/crates/rslint_parser/src/syntax/stmt.rs
+++ b/crates/rslint_parser/src/syntax/stmt.rs
@@ -965,16 +965,33 @@ fn parse_variable_declaration(
 
 		// Heuristic to determine if we're in a for of or for in loop. This may be off if
 		// the user uses a for of/in with multiple declarations but this isn't allowed anyway.
-		let is_in_for_of_or_in = context.parent == VariableDeclarationParent::For
-			&& context.is_first
-			&& (p.cur_src() == "of" || p.at(T![in]));
+		let is_in_for_loop = context.parent == VariableDeclarationParent::For && context.is_first;
+		let is_in_for_of = is_in_for_loop && p.cur_src() == "of";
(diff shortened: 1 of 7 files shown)


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -979,6 +979,23 @@ fn parse_variable_declaration(
 					p.error(err);
 				}
 			}
+			if let Some(initializer) = initializer {
+				// Initializers are disallowed in strict mode for `for..in`
+				// and `for..of`, except for `for(var ... in ...)`
```
MichaReiser, line 984: Would you mind adding a test case showing that this branch works as expected?

```
test for_in_and_of_initializer_loose_mode
...

test_err for_in_and_of_initializer_strict_mode
for (let i = 0 in []) {}
for (var i = 0 in []) {}
for (let i = 0 of []) {}
for (var i = 0 in []) {}
```

MichaReiser (commented):
Nice! Would you mind adding two tests showing that the diagnostics are correctly added in strict mode but not in loose mode?

crates/rslint_parser/src/syntax/stmt.rs
leops: I'll add those, besides that the loose mode case is also covered by `xtask\src\coverage\test262\test\annexB\language\statements\for-in\nonstrict-initializer.js` which isn't marked as fixed because it never raised a diagnostic in the first place

MichaReiser (approved):

PR #1918, opened 2022-01-10
removing token_offset_pair and tokens_after_linebreaks

xunilrj:
see: https://github.com/rome/tools/issues/1901

With the newly created tokens (NEWLINE, MULTILINE_COMMENT), we can completely remove tokens_after_linebreaks. The lexer will know when a non-trivia token is after a newline. We can attach this info in the Token struct and completely avoid creating and searching this HashSet.

We can also improve the lexer by already calculating the token offset and remove token_offset_pair. Avoiding a whole iteration when creating the Parser.

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -30,16 +30,16 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub enum TriviaPiece {
-	Whitespace(usize),
-	Comments(usize),
+	Whitespace(u32),
+	Comments(u32),
 }
 
 impl TriviaPiece {
 	#[inline]
 	pub fn text_len(&self) -> TextSize {
 		match self {
-			TriviaPiece::Whitespace(n) => (*n as u32).into(),
-			TriviaPiece::Comments(n) => (*n as u32).into(),
+			TriviaPiece::Whitespace(n) => (*n).into(),
+			TriviaPiece::Comments(n) => (*n).into(),
 		}
 	}
 }
--- a/crates/rslint_lexer/src/tests.rs
+++ b/crates/rslint_lexer/src/tests.rs
@@ -38,7 +38,9 @@ macro_rules! assert_lex {
                 tokens[idx].0.kind,
             );
 
-            new_str.push_str($src.get(tok_idx..(tok_idx + tokens[idx].0.len)).unwrap());
+            let tok_start = tok_idx as usize;
+			let tok_end = tok_idx as usize + tokens[idx].0.len as usize;
+            new_str.push_str($src.get(tok_start..tok_end).unwrap());
             tok_idx += tokens[idx].0.len;
 
             idx += 1;
@@ -75,7 +77,9 @@ fn losslessness(string: String) -> bool {
 	let mut idx = 0;
 
 	for token in tokens {
-		new_str.push_str(string.get(idx..(idx + token.len)).unwrap());
+		let start = idx as usize;
+		let end = idx as usize + token.len as usize;
+		new_str.push_str(string.get(start..end).unwrap());
 		idx += token.len;
 	}
 
--- a/crates/rslint_lexer/src/token.rs
+++ b/crates/rslint_lexer/src/token.rs
@@ -7,20 +7,53 @@ use crate::JsSyntaxKind;
 pub struct Token {
 	/// The kind of token this is.
 	pub kind: JsSyntaxKind,
+	/// Indicates if there is a newline between this token and the previous non-trivia token.
+	pub after_newline: bool,
+	/// Offset from the start of the file, in bytes.
+	pub offset: u32,
 	/// How long the token is in bytes. For tokens with escape sequences
 	/// like strings with `\uXXXX` escapes, the length is the raw length, not considering the char backed by the escape.
-	pub len: usize,
+	pub len: u32,
 }
 
 impl Token {
 	/// Create a new token which has an exact length of 1.
-	pub fn single(kind: JsSyntaxKind) -> Self {
-		Self { kind, len: 1 }
+	pub fn single(kind: JsSyntaxKind, offset: usize) -> Self {
+		Self {
+			kind,
+			len: 1,
+			offset: offset as u32,
+			after_newline: false,
+		}
 	}
 
 	/// Create a new token which has a specific length.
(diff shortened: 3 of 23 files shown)


crates/rslint_lexer/src/tests.rs
```diff
@@ -38,7 +38,9 @@ macro_rules! assert_lex {
                 tokens[idx].0.kind,
             );
 
-            new_str.push_str($src.get(tok_idx..(tok_idx + tokens[idx].0.len)).unwrap());
+            let tok_start = tok_idx as usize;
+			let tok_end = tok_idx as usize + tokens[idx].0.len as usize;
```
MichaReiser, line 42: Nit: Formatting

crates/rslint_lexer/src/token.rs
```diff
@@ -7,20 +7,53 @@ use crate::JsSyntaxKind;
...
 	/// The kind of token this is.
 	pub kind: JsSyntaxKind,
+	/// Indicates if there is a newline between this token and the previous non-trivia token.
+	pub after_newline: bool,
+	/// Offset from the start of the file, in bytes.
+	pub offset: u32,
 	/// How long the token is in bytes. For tokens with escape sequences
 	/// like strings with `\uXXXX` escapes, the length is the raw length, not considering the char backed by the escape.
-	pub len: usize,
+	pub len: u32,
```
MichaReiser, line 16: Nit: Use `TextSize`? also for `offset`?

crates/rslint_lexer/src/token.rs
```diff
@@ -7,20 +7,53 @@ use crate::JsSyntaxKind;
 pub struct Token {
```
MichaReiser, line 7: What would you think of using two different `Token`. The `Token` return by the lexer only stores `kind`, and `len` (maybe `after_newline`). 

The token source would use another `Token` type that then tracks the `offset` as well. This would be closer to how the parser worked today with the `Token` definition in the parser crate. We should still have the same performance gains because the offset can easily be computed in the token source.

crates/rslint_parser/src/token_source.rs
```diff
@@ -1,164 +1,131 @@
...
+	#[inline(always)]
+	pub fn last_tok(&self) -> Option<&rslint_lexer::Token> {
+		self.next_non_trivia(self.cur, -1)
+			.and_then(|idx| self.raw_tokens.get(idx))
 	}
 
-	pub fn current(&self) -> Token {
-		self.cur.0.to_owned()
+	#[inline(always)]
+	pub fn current(&self) -> &rslint_lexer::Token {
```
MichaReiser, line 68: Is there a significant performance benefit of returning a reference? Handling the lifetimes may make tokens a bit harder to handle.

MichaReiser (commented):
I love it! 

Do you know what the runtime performance and memory impact of this change is?

crates/rome_rowan/src/api.rs
```diff
@@ -30,16 +30,16 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub enum TriviaPiece {
-	Whitespace(usize),
-	Comments(usize),
+	Whitespace(u32),
```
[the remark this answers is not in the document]
xunilrj: This may be a good thing for other PR. But I sometimes feels the ergonomics of ```TextSize``` is a little bit annoying.

crates/rslint_lexer/src/token.rs
xunilrj: hum... One the things I liked about this solution is exactly to remove the ```Token``` struct on the ```rslint_parser``` crate. For multiple reasons:
1 - It is easier to calculate the offset when lexing; we already have this info. I put this on the "Iterator::next" just to avoid having to specify this on every Token creation, but this would actually be faster;
2 - Having a second token means we need a second ```Vec``` and iterate all tokens; or keep the current offset and create the second ```Token``` on demand, like we are doing now; This creates one of the problems I want to fix in the next PR. Wen we return ```Token``` by value, the compiler generates a lot of ```memcpy```. With just one, we can return a borrow with all the data we have, which is cheaper;
3 - With ```offset``` and ```len``` at the same place, it is cheaper/easier to generate a range for tokens, which also makes it cheaper and easier to access tokens ```&str```.

crates/rslint_parser/src/token_source.rs
xunilrj: Yes. I can try to measure to check if the gains materialize in practice. But we avoid a lot of ```memcpy```.

crates/rslint_lexer/src/token.rs
xunilrj: Mind if I change this in another PR?

MichaReiser (approved):

crates/rslint_lexer/src/tests.rs
xunilrj: Done.

crates/rslint_parser/src/token_source.rs
```diff
@@ -1,164 +1,131 @@
...
-			token_offset_pairs,
-			cur: (first, 0),
-			tokens_after_linebreaks,
+			cur: pos,
 			raw_tokens,
 		}
 	}
 
+	#[inline(always)]
+	fn next_non_trivia(&self, pos: usize, dir: isize) -> Option<usize> {
```
MichaReiser, line 32: Should `dir` be a `bool` since the implementation only works with `1` and `-1`

crates/rslint_parser/src/token_source.rs
```diff
@@ -1,164 +1,131 @@
...
+			}
+		}
+		Some(pos as usize)
+	}
+
+	#[inline(always)]
+	fn raw_lookahead_nth(&self, n: usize) -> usize {
+		let mut idx = self.cur;
+		for _ in 0..n {
+			idx = self.next_non_trivia(idx, 1).unwrap();
```
MichaReiser, line 50: This call is causing issues if the lookahead runs out of bounds as you can see in #1915

Not sure what we should return in this case but the parser shouldn't have to check if a lookahead might go out of bounds.

crates/rslint_parser/src/token_source.rs
xunilrj: will do.

crates/rslint_parser/src/token_source.rs
xunilrj: the old implementation used to return the last token. Will keep the same behaviour.

PR #1920, opened 2022-01-10
fix(rslint_parser): fix JsVariableDeclarations is_var method

yassere:
## Summary

Fixes `JsVariableDeclarations::is_var`. Looks like just a copy/paste mistake.


MichaReiser (approved):
whoops

PR #1924, opened 2022-01-10
feat: box lexerreturn to avoid memcpy

xunilrj:
see https://github.com/rome/tools/issues/1901

# LexerReturn

```LexerReturn``` is huge! Doing a ```-Zprint-type-sizes``` we find that Diagnostic alone is 192 bytes.

```
> cargo +nightly rustc -- -Zprint-type-sizes
...
print-type-size type: `rslint_errors::Diagnostic`: 192 bytes, alignment: 8 bytes
print-type-size     field `.file_id`: 8 bytes
print-type-size     field `.code`: 24 bytes
print-type-size     field `.title`: 24 bytes
print-type-size     field `.primary`: 56 bytes
print-type-size     field `.children`: 24 bytes
print-type-size     field `.suggestions`: 24 bytes
print-type-size     field `.footers`: 24 bytes
print-type-size     field `.severity`: 1 bytes
print-type-size     field `.tag`: 1 bytes
print-type-size     end padding: 6 bytes
...
```

Given that ```LexerReturn``` is returned by value in multiple functions, we find the compiler having to copy everything multiple times, for example.

```
>  cargo asm "rslint_parser::parse::tokenize" --rust
...
mov     r13, qword, ptr, [rip, +, memcpy@GOTPCREL]
jmp     .LBB378_2
...
.LBB378_2:
lea     rdi, [rsp, +, 224]
lea     rsi, [rsp, +, 64]
call    qword, ptr, [rip, +, _ZN78_$LT$rslint_lexer..Lexer$u20$as$u20$core..iter..traits..iterator..Iterator$GT$4next17hde555c7f27552293E@GOTPCREL]
movzx   eax, word, ptr, [rsp, +, 232]
cmp     eax, 393
je      .LBB378_4
mov     ebx, dword, ptr, [rsp, +, 232]
mov     rbp, qword, ptr, [rsp, +, 224]
mov     edx, 196
lea     rdi, [rsp, +, 624]
mov     rsi, r12
call    r13
...
```

Relevant here is

```
mov     r13, qword, ptr, [rip, +, memcpy@GOTPCREL]
mov     edx, 196
call    r13
```

```jquery.min``` for example has 41,812 tokens. So just in this function we are coping 8,195,152 bytes, or 8MB. Given that ```LexerReturn``` is moved through two or three function calls, I would not be surprise to see that we are copying at least three times this value.

And to make matters worst, the very next line we throw away this ```Diagnostic``` on 99.99% of the time.

Running Valgrind we see that we spend 7% of the time parsing ```jquery.min``` memcpying stuff. To improve on that, I will box all ```Diagnostic```, hoping that the alloc cost in the end will be much smaller than copying stuff.

```
> valgrind --tool=callgrind --dump-instr=yes --simulate-cache=yes --collect-jumps=yes target/release/xtask coverage-libs --filter=jquery --criterion=false
> kcachegrind
```

image

We are not alone:

```https://github.com/rust-lang/rust/issues/64301```
> I'm going to modify this issue to be about all types that cause many calls to memcpy. These types are all > 128 bytes in size.

One way to count the number of ```memcpy```s is using: https://github.com/jrmuizel/memcpy-find

```
> RUSTFLAGS='-g --emit=llvm-ir' cargo build --release
> memcpy-find target/release/deps/rslint_lexer-10b324ec6226da3e.ll
```

For me this does not work. Not with llvm-13 nor with llvm-12, so I gave up and went bruteforce.

```
> cat target/release/deps/rslint_lexer-10b324ec6226da3e.ll | grep memcpy | wc -l
3184
```

# Performance

Before
```
https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js
                        time:   [15.582 ms 15.706 ms 15.828 ms]
                        time:   [80.301 ms 81.004 ms 81.711 ms]
```

After
```
https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js
                        time:   [14.031 ms 14.190 ms 14.352 ms]
                        time:   [73.313 ms 74.036 ms 74.780 ms]
```

```memcpy``` went from 7% to 3.5%.

image

And now we are calling ```memcpy``` "just":

```
> cat target/release/deps/rslint_lexer-10b324ec6226da3e.ll | grep memcpy | wc -l
2710
```

### DIFF
--- a/crates/rslint_lexer/src/lib.rs
+++ b/crates/rslint_lexer/src/lib.rs
@@ -36,7 +36,7 @@ use state::LexerState;
 use tables::derived_property::*;
 
 pub use rslint_syntax::*;
-pub type LexerReturn = (Token, Option<Diagnostic>);
+pub type LexerReturn = (Token, Option<Box<Diagnostic>>);
 
 // Simple macro for unwinding a loop
 macro_rules! unwind_loop {
@@ -334,7 +334,7 @@ impl<'src> Lexer<'src> {
 
 	// Validate a `\x00 escape sequence, this expects the current char to be the `x`, it also does not skip over the escape sequence
 	// The pos after this method is the last hex digit
-	fn validate_hex_escape(&mut self) -> Option<Diagnostic> {
+	fn validate_hex_escape(&mut self) -> Option<Box<Diagnostic>> {
 		debug_assert_eq!(self.bytes[self.cur], b'x');
 
 		let diagnostic =
@@ -346,8 +346,8 @@ impl<'src> Lexer<'src> {
 
 		for _ in 0..2 {
 			match self.next_bounded() {
-				None => return Some(diagnostic),
-				Some(b) if !(*b as u8).is_ascii_hexdigit() => return Some(diagnostic),
+				None => return Some(Box::new(diagnostic)),
+				Some(b) if !(*b as u8).is_ascii_hexdigit() => return Some(Box::new(diagnostic)),
 				_ => {}
 			}
 		}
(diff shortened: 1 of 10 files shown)


leops:
Currently it seems like all the consumers of `Lexer` end up writing the Diagnostics in a vector or discarding them. It's probably a more complex change than boxing the return types but given that Diagnostics have associated source location informations we don't necessarily have to tie them together with tokens into a single stream, it seems to me that the lexer could have an `Option<&mut Vec<Diagnostic>>` and write its errors in there directly

xunilrj:
> it seems to me that the lexer could have an `Option<&mut Vec<Diagnostic>>` and write its errors in there directly

Agreed.

I think the initial idea of the ```Lexer``` being lazy was to avoid storing tokens in a Vec in the first place. But this is not happening now.

Maybe in the future we manage to make the whole process lazy; then we can "inject" a ```Vec<Diagnostic>``` and each phase. I would love to see this specifically to avoid storing so many parser events.

MichaReiser (approved):

PR #1925, opened 2022-01-11
feat(rslint_parser): Avoid cloning parser state

MichaReiser:
## Summary

The parser has a neat `with_state` function that takes a snapshot of the parser state and restores the state to its previous state when it goes out of scope. However, the downside is that it copies the state at least twice, once inside the `with_state` function and once in the call-site, when constructing the new state which results in a serious performance penalty. 

This PR introduces a new `ChangeParserState` trait that allows making specific changes to the `ParserState` that can later be undone, only keeping a copy of the data they changed (and need to be restored later). Ideally, these changes would have semantic names. However, this PR re-uses the name of the fields. 

I hope to be able to reduce a lot of the boilerplate by introducing generic fields, for example if we're inside a function, constructor, async, etc... that then can all be changed with a single `ChangeParserState` implementation.

## Performance

The parse time (excluding tree sink and lexing) or across all libs improves by 25-33%, in some cases even close to 50%

* jquery.min.js: 3.244868ms -> 2.395019ms
* vue.global.prod.js: 6.825166ms -> 3.757078ms
* react.production.min.js: 246.53µs -> 191.09µs
* react-dom.production.min.js: 4.974296ms -> 3.006538ms
* compiler.js: 12.915072ms -> 9.110865ms
* dojo.js: 784.669µs -> 617.38µs
* d3.min.js: 11.771383ms -> 8.440315ms
* pixi.min.js: 12.437823ms -> 10.148374ms
* three.min.js:  19.487128ms -> 11.145823ms
* tex-chtml-full.js: 29.200492ms  -> 21.564177ms

## Test Plan

Verified that coverage doesn't change.

### DIFF
--- a/crates/rslint_parser/src/parser.rs
+++ b/crates/rslint_parser/src/parser.rs
@@ -138,23 +138,11 @@ pub struct Parser<'t> {
 impl<'t> Parser<'t> {
 	/// Make a new parser
 	pub fn new(tokens: TokenSource<'t>, file_id: usize, syntax: Syntax) -> Parser<'t> {
-		// TODO(RDambrosio016): Does TypeScript imply Module/Strict?
-		let strict = if syntax.file_kind == FileKind::Module {
-			Some(StrictMode::Module)
-		} else {
-			None
-		};
-		let state = ParserState {
-			is_module: syntax.file_kind == FileKind::Module,
-			strict,
-			..ParserState::default()
-		};
-
 		Parser {
 			file_id,
 			tokens,
 			events: vec![],
-			state,
+			state: ParserState::new(syntax),
 			syntax,
 			errors: vec![],
 		}
@@ -471,6 +459,11 @@ impl<'t> Parser<'t> {
 			.map(|x| usize::from(x.range(self).end()))
 			.unwrap_or_default();
 		start..end
+	}
+
+	/// Whether the code we are parsing is a module
+	pub fn is_module(&self) -> bool {
+		self.syntax.file_kind == FileKind::Module
 	}
 }
 
--- a/crates/rslint_parser/src/state.rs
+++ b/crates/rslint_parser/src/state.rs
@@ -1,5 +1,6 @@
+use crate::{FileKind, Parser, Syntax};
 use std::collections::HashMap;
-use std::ops::Range;
+use std::ops::{Deref, DerefMut, Range};
 
 /// State kept by the parser while parsing.
 /// It is required for things such as strict mode or async functions
@@ -9,32 +10,30 @@ pub struct ParserState {
 	/// inside an expression.
 	///
 	/// Also applies for object patterns
-	pub allow_object_expr: bool,
+	allow_object_expr: bool,
 	/// Whether `in` should be counted in a binary expression
 	/// this is for `for...in` statements to prevent ambiguity.
-	pub include_in: bool,
+	include_in: bool,
 	/// Whether the parser is in an iteration statement and `continue` is allowed.
-	pub continue_allowed: bool,
+	continue_allowed: bool,
 	/// Whether the parser is in an iteration or switch statement and
 	/// `break` is allowed.
-	pub break_allowed: bool,
+	break_allowed: bool,
 	/// A list of labels for labelled statements used to report undefined label errors
 	/// for break and continue, as well as duplicate labels
 	pub labels: HashMap<String, Range<usize>>,
 	/// Whether the parser is in a generator function like `function* a() {}`
-	pub in_generator: bool,
+	in_generator: bool,
 	/// Whether the parser is inside of a function
-	pub in_function: bool,
+	in_function: bool,
 	/// Whatever the parser is inside of a constructor
-	pub in_constructor: bool,
+	in_constructor: bool,
 	/// Whether we potentially are in a place to parse an arrow expression
-	pub potential_arrow_start: bool,
+	potential_arrow_start: bool,
 	/// Whether we are in an async function
-	pub in_async: bool,
+	in_async: bool,
 	/// Whether we are in strict mode code
-	pub strict: Option<StrictMode>,
-	/// Whether the code we are parsing is a module
-	pub is_module: bool,
+	strict: Option<StrictMode>,
 	/// The exported default item, used for checking duplicate defaults
 	pub default_item: Option<Range<usize>>,
 	/// If set, the parser reports bindings with identical names. The option stores the name of the
@@ -42,11 +41,9 @@ pub struct ParserState {
 	pub duplicate_binding_parent: Option<&'static str>,
 	pub name_map: HashMap<String, Range<usize>>,
 	/// Whether the parser is in a conditional expr (ternary expr)
-	pub in_cond_expr: bool,
-	pub in_case_cond: bool,
+	in_cond_expr: bool,
 	pub(crate) no_recovery: bool,
-	pub in_declare: bool,
-	pub in_binding_list_for_signature: bool,
+	in_binding_list_for_signature: bool,
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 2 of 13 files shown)


ematipico:
> However, I plan to group many of the flags fields in a bit-flag and then introduce a with_flags function that implements the old behaviour but in a way so that it only snapshots the flags which is the same as storing an integer.

Would you mind create an issue and explain there, in more detail, what's the approach to plan to design?

ematipico (approved):
Even though we save in performance, I think this approach is really unstable. I get that this is a temporary solution, but I wonder if this change can be implemented with a generic function? (probably not)

Boshen:
From my understanding of Rust, I think the pattern we need to explore here is the concept of Interior Mutability, i.e.`std::cell::Cell`, and use `std::cell::swap/replace`. We can then remove &mut on parser to make it clear what can be mutated. 

```rust
pub struct ParserState {
	pub strict: Rc<Option<StrictMode>>,
}
```

```rust
	let mut last_strict = Some(StrictMode::Class(p.cur_tok().range()));
	std::mem::swap(&mut p.state.strict, &mut last_strict);

becomes

	let last_strict = Cell::new(Some(StrictMode::Class(p.cur_tok().range())));
	p.state.strict.swap(last_strict);
```

leops:
I think there's a workable design in-between both the previous drop guard based design and the new save and restore behavior:

```rust
trait ParserModifier {
    fn enter(state: &mut ParserState) -> Self;
    fn exit(self, state: &mut ParserState);
}

struct ParserStateGuard<'state, T: ParserModifier> {
    modifier: T,
    state: &mut ParserState,
}

impl<'state, T: ParserModifier> ParserStateGuard<'state, T> {
    fn enter(state: &mut ParserState) -> Self {
        Self { modifier: T::enter(state), state }
    }
}

impl<'state, T: ParserModifier> Drop for ParserStateGuard<T> {
    fn drop(&mut self) {
        T::exit(self.modifier, self.state);
    }
}

struct InGenerator(bool);

impl ParserModifier for InGenerator {
    fn enter(state: &mut ParserState) -> Self {
        Self(replace(&mut state.in_generator, true))
    }

    fn exit(self, state: &mut ParserState) {
        state.in_generator = self.0;
    }
}
```

This is an approximate example, I haven't included the implementation of `Deref` for `ParserStateGuard`. Also it would be possible to have generic implementations of `ParserModifier` over tuples of modifiers to facilitate merging multiple changes at once

MichaReiser:
> > However, I plan to group many of the flags fields in a bit-flag and then introduce a with_flags function that implements the old behaviour but in a way so that it only snapshots the flags which is the same as storing an integer.
> 
> Would you mind create an issue and explain there, in more detail, what's the approach to plan to design?

I don't know yet in detail but it's the next I plan to tackle. I was thinking of grouping some of the flags like `in_async`, `in_generator` together as well as `in_function` and `in_constructor` but I still need to work out the details because JS allows for some weird constructs. 

> I think there's a workable design in-between both the previous drop guard based design and the new save and restore behavior:

That's actually something I wanted to look in next once I figure out how we want to restructure the parser state. But let me give it a try and fine tune it from their. One difference is that this design doesn't allow to restore the state on drop but I think that's fine.

xunilrj:
> I think there's a workable design in-between both the previous drop guard based design and the new save and restore behavior:

I liked your approach. Remind me of "mutable lens" with undo.

The only crate that tries to do "mutable lens", as far as I know, is https://github.com/nsmryan/myopic. But there are not a lot of examples and its use seems harder than necessary. All other "lens" (linked at myopic readme) focus on "immutability with clone", what we trying to avoid here.

From the undo/redo crates, I think this one is closer to what we need. Apart from the "Point!" macro in the example is very close to what we want.
https://github.com/M-Adoo/rundo/blob/master/docs/quickstart.md

MichaReiser:
I implemented @leops proposal but had to make two changes:

* The `apply` method making the change to the `ParserState` must be an instance method so that it can be parametrized. For example, setting the `strict` field requires knowing the range of the `use strict` directive. 
* A consequence of the above is that there are now two structs for each change. The struct implementing the change logic and a new-type wrapper for storing the previous state. 

@xunilrj this crate looks interesting but also feels a bit magic. There are also fields that are tracked and untracked at the same time. For example, the `name_map` field must not be tracked when items are added or removed to it but there are places where we replace the whole map, which should be tracked. I further hope that we can come up with more semantic names for the changes, so that these become more than simple "mutation calls".

crates/rslint_parser/src/parser.rs
```diff
@@ -472,6 +460,11 @@ impl<'t> Parser<'t> {
 			.unwrap_or_default();
 		start..end
 	}
+
+	/// Whether the code we are parsing is a module
+	pub fn is_module(&self) -> bool {
+		self.syntax.file_kind == FileKind::Module
```
MichaReiser, line 466: Not worth tracking state that we can very cheaply compute.

crates/rslint_parser/src/state.rs
```diff
@@ -1,56 +1,52 @@
...
-	pub strict: Option<StrictMode>,
-	/// Whether the code we are parsing is a module
-	pub is_module: bool,
+	strict: Option<StrictMode>,
 	/// The exported default item, used for checking duplicate defaults
 	pub default_item: Option<Range<usize>>,
 	/// If set, the parser reports bindings with identical names. The option stores the name of the
 	/// node that disallows duplicate bindings, for example `let`, `const` or `import`.
 	pub duplicate_binding_parent: Option<&'static str>,
 	pub name_map: HashMap<String, Range<usize>>,
```
MichaReiser, line 42: These have more complicated "reverting" logic that goes beyond reverting when going out of scope. Directly changing `p.state` feels easier in those cases.

MichaReiser:
> ### Test262 comparison coverage results on ubuntu-latest
> Total 	44951 	44951 	0
> Passed 	42923 	42923 	0
> Failed 	2027 	2027 	0
> Panics 	1 	1 	0
> Coverage 	95.49% 	95.49% 	0.00%

Thanks @Boshen for fixing the ubuntu test reports!

xunilrj:
I did a quick test here and it seems to be working. What do you think?
It alleviates the boilerplate needed, and it makes more obvious that all behave the same.

```rust
macro_rules! gen_parser_mutate {
	($name:ident { $field:ident: $type:path} ) => {
		/// Changes the [ParserState] `in_generator` field
		#[derive(Debug)]
		pub struct $name(pub (crate) $type);

		impl ChangeParserState for $name {
			type Snapshot = $type;

			fn apply(self, state: &mut ParserState) -> Self::Snapshot {
				std::mem::replace(&mut state.$field, self.0)
			}

			fn restore(state: &mut ParserState, value: Self::Snapshot) {
				state.$field = value
			}
		}
	};
}

gen_parser_mutate!{InGenerator{ in_generator: bool}}
gen_parser_mutate!{InFunction{ in_function: bool}}
gen_parser_mutate!{InAsync{ in_async: bool}}
```

It is only a little bit different from yours.
example:


```rust
let p = &mut *p.with_state(
    InFunction(true)
	.and(InGenerator(in_generator))
	.and(InAsync(in_async))
);
```

NicholasLYang:
Perhaps this is a dumb question, but is there a way to have the API use a closure? I've had functions that allow you to run code in a specific state, then restore once the closure finishes

MichaReiser:
> Perhaps this is a dumb question, but is there a way to have the API use a closure? I've had functions that allow you to run code in a specific state, then restore once the closure finishes.
> 
> The reason I ask is that having the change occur on drop would likely be a little confusing. If we can have an API like:
> 
> ```
> run_in_state!(in_generator: true, |p| {  ... })
> ```

That would work as well. I guess, the main question is how you type the ‚in_generator‘. We could use the same approach as introduced in this diff so that ‚run_with_state‘ is just a different api to ‚with_state‘. 

Another API would be a parserstate with all fields being options but this then requires finding the initialized fields, which may be a bit cumbersome.

xunilrj (approved):

NicholasLYang:
This is where macros could work quite well no?  Like we can have:

```
macro_rules! run_in_state
(($field:ident: $value:expr)+, $func:expr) => {
  // swap fields
  let fn = $func;
  fn();
  // swap back fields
}
```

MichaReiser:
> macro_rules! run_in_state
> (($field:ident: $value:expr)+, $func:expr) => {
>   // swap fields
>   let fn = $func;
>   fn();
>   // swap back fields
> }

This is a cool idea and I shortly tried it out. The main economics shortcoming that I see is that IDE's aren't good at providing auto-completion into macros (at least, CLion doesn't).  Also, finding for the places where a state field gets changed is no longer possible, because the tooling doesn't correctly support that (yet?). 

It's a pity, because that would have been a super elegant solution with minimal boilerplate and compiles almost to the same code as I wrote by hand in the initial version. 

That's the code I played around with in clion

```
macro_rules! run_in_state {
	($p:expr, $field:ident: $value:expr, $do:expr) => {{
		let snapshot = std::mem::replace(&mut $p.state.$field, $value);
		let result = $do;
		$p.state.$field = snapshot;
		result
	}};
	// ($p:expr, $field:ident:$value:expr, $($next_field:ident: $next_value:expr),+ $do:expr) => {{
	    // 	run_in_state!(
	    // 		$p,
	    // 		$field,
	    // 		$value,
	    // 		run_in_state!($p, ($next_field, $next_value),+ $do)
	    // 	)
	    // }};
}

fn test(p: &mut Parser) {
	run_in_state!(p, in_async: true, {
		p.eat(T!['{']);
	});
	// run_in_state!(p, in_async: true, in_generator: true, {
	// 	p.eat(T!['{']);
	// })
}
```

> I did a quick test here and it seems to be working. What do you think? It alleviates the boilerplate needed, and it makes more obvious that all behave the same.

I would leave it for now because I'm likely going to collapse exactly those flags, and thus, doesn't seem worth introducing a macro to then remove it again. However, I like the idea of making the struct fields public to remove the need for a `new` function

PR #1928, opened 2022-01-11
fix: return first/last token when index is out of bounds

xunilrj:
Fix ```unwrap``` panic on https://github.com/rome/tools/pull/1915


MichaReiser (approved):

MichaReiser:
Thanks, for looking into this so quickly.

Issue #1909, opened 2022-01-07
☂️ Remaining Parsing Errors from Test262


Boshen:
### Description

This is the umbrella issue for the remaining 2000+ failed test262 parsing tests.

We aim to pass all tests, please create a separate issue and link them here if you wish to tackle some the of errors.

---
The entire test262 diagnostics can be found in this [gist](https://gist.github.com/Boshen/647c7d85b185fab78ea7bfb49a718293).

All the common features and their respective directories are linked below, you can search the directory names in the [gist](https://gist.github.com/Boshen/647c7d85b185fab78ea7bfb49a718293) for their diagnostics.

---
## Language Features

- [ ] [Regex](https://github.com/tc39/test262/tree/main/test/built-ins/RegExp)
  * `built-ins/RegExp/property-escapes/*`
  * https://github.com/rome/tools/pull/1927
- [ ] [Template Literal](https://github.com/tc39/test262/tree/main/test/language/expressions/template-literal)
- [x] [Top Level Await](https://github.com/tc39/test262/tree/main/src/top-level-await) #1936 
- [ ] [Export Assertion](https://github.com/tc39/test262/tree/main/test/language/module-code) `import-assertion-*.js`
- [x] [Dynamic Import Assertion](https://github.com/tc39/test262/blob/main/test/language/expressions/dynamic-import) 
  * https://github.com/rome/tools/pull/1923
- [x] [Export](https://github.com/tc39/test262/tree/main/test/language/module-code) `*-export-*.js`
  * https://github.com/rome/tools/pull/1915
- [x] [Unicode](https://github.com/tc39/test262/tree/main/test/language/identifiers) `*-unicode-*.js`
  * https://github.com/rome/tools/pull/1910
- [x] [For Await Of](https://github.com/tc39/test262/tree/main/test/language/statements/for-await-of) `async-*.js`
- [ ] [Arrow Function Destructuring](https://github.com/tc39/test262/tree/main/test/language/expressions/arrow-function/dstr)
- ... more cases

## Negative Tests
These are mostly case by case errors, you can search for `Expected this test to fail, but instead it passed without errors.`
- [ ] [Block Scope Redeclaration](https://github.com/tc39/test262/tree/main/test/language/statements/switch/syntax/redeclaration) @ematipico #1929 
- [ ] Early Errors `/early-error/`
- [x] https://github.com/rome/tools/pull/1922
- [x] numbers with leading zero https://github.com/rome/tools/pull/1961
- ... more cases

[...]

PR #1929, opened 2022-01-12
fix(rslint_parser): binding redeclaration

ematipico:
## Summary

Part of #1909 

This PR applies various changes to how bindings are tracked during the parsing phase.

### Notable changes

The most notable change is applied to the `ParserState`:
- added a new variable called `binding_context` which is in charge of tracking the scopes, which determine how to track a name (name is a binding);
- created two `HashMap`, responsible to track hoisted names (`var a` is the more common) and lexical names (`let a` is the more common);
- removed `duplicate_binding_parent` in favour of `binding_type`, which better tracks the kind of a binding when is declared. It's not just for `let/const` and `var`, but also functions, modules and - ideally - classes too;
- created a new bit flag called `IN_FUNCTION_BLOCK`, which tells the parser if inside a `function f() {}`. The different from `BindingContext::FunctionBlock` is that inside a function block, it's possible to define other kind of binding context, although any function that is created inside a function block, regardless of its binding context, should behave the same. An example is the following:
	  
```js 
function f() {
	let a;
	{ function a() {} }  // here
}
```
	  
in `// here` the function is has to be treated as a hoisted variable that belongs to the `function f() {}`, this means that we don't care if we are inside a block `{}`, where different rules are applied.

- removed `name_map`, as the new logic takes care of it
- created a state change called `EnterVariableDeclaration` to map the type of parameters list (`var a, a` vs `let a, a`)
- created a state change called `EnterLexicalScope` used to change the type of context of the parser, it also creates a new `lexical_names` map. A change of context usually involves a new set of lexical names. This is used for various situations, where we lexical names live and die inside them: blocks, classes, function blocks, `try/catch/finally`,  statements, `static {}`, etc.
- created a state change called `EnterHoistedScope`, used to create a new `hoisted_names` hashmap. We want to usually use this when we encounter function blocks, function arguments, loops statements, etc.
- created a state change called `EnterFunctionDeclaration`, which explicitly set `binding_type` 
- created a new function called `register_name`, which is responsible to track a binding depending by it's `binding_context` and `variable_type`
- created a new function called `clashes_with_defined_name`, which checks if an existing name has been declared already. The logic changes based on the current `binding_context` and `variable_type`

### Decisions

At the moment we currently have some change state called `EnterParameters` and `EnterFunction`, although their usages is different from what we want we do with the scopes.

- `EnterParameters`  is used, usually, inside a closure where the parameters are parsed. At the end of the parsing, the parser state is restored. Here, we need to create a new hoisted scope for the parameters, although these parameters need to be tracked inside the following block, which means that we need a scoped state:

```rs
let p = &mut *p.with_scoped_state(EnterHoistedScope(BindingContext::Arguments));
p.with_state(EnterParameters(), |p| parse_parameter);
p.parse_block_impl(); // here the new hoisted scope is still alive and we can perform the correct validation
```

- `EnterFunction` has a different semantic and it's used differently from `EnterLexicalScope(BindingContext::FunctionBlock)`

## Test Plan

Created new tests for each case that encountered and gave me headache. Probably some case hasn't been covered by during the commits I took a close look to all the regressions and fixed them one by one.

I also updated lot of existing tests in the parser AND the formatter because the new parsing logic uncovered lot of invalid grammar.

```
cargo format
cargo test
cargo lint
```

I also run a performance test, and it seems that it hasn't changed, more or less. (it's really difficult to tell)


crates/rslint_parser/src/state.rs
```diff
@@ -50,6 +52,54 @@ pub struct ParserState {
 	pub in_binding_list_for_signature: bool,
 }
 
+/// Structure to track the bindings found inside each block
+#[derive(Debug, Clone, PartialEq)]
+pub struct BindingsBlock {
+	/// A series of blocks. Each block tracks the bindings with their identifier name and their range
+	/// Blocks are FILO: First In, Last Out
+	blocks: VecDeque<HashMap<String, Range<usize>>>,
```
MichaReiser, line 60: Do we need to keep a queue of hash maps? Isn't it sufficient to store the existing `HashMap` in the place where we start a new scope, replace it with an empty hash map, and then, when we leave the scope to replace it with the original one?

```
fn block_impl(p: &mut Parser) {
  let last = std::mem::take(&mut p.state.bindings);
  ....
  p.state.bindings = last;
}
```

ematipico: I still need to put more thought into the implementation. Even though your suggestion makes totally sense, new blocks are not confined only confined to `block_impl`. 

We might end up to have repeated code... I will try and see if we can make your solution more generic.

crates/rslint_parser/src/state.rs
```diff
@@ -374,3 +378,224 @@ impl ChangeParserState for EnableStrictMode {
...
+
+/// Data structure responsible to track bindings inside their scopes
+#[derive(Debug, PartialEq)]
+pub struct BindingsTracking {
+	/// Scopes where hoisted variables are tracked. We have two scopes
+	hoisted_scopes: [HoistedVars; 2],
+	/// Whether the parser is inside a block. If so, hoisted vars checked against the second [ScopedVars]
+	is_inside_block: bool,
+	/// Semantic blocks. They follow the FILO order: First-In, Last-Out
+	blocks: VecDeque<HashMap<String, Range<usize>>>,
```
MichaReiser, line 390: I don't think we need to track a `VecDequeue`. It seems all we need are two maps: 

1. To track the hoisted declarations 
2. To track the block-scoped declarations

The 1. must be re-set whenever we enter a function. You can achieve that by storing the existing map in `EnterFunction` and restoring it when we leave the function.

You can define a new `EnterBlock` `ChangeParserState` that does the same for the block-scoped variables: Stores away the existing map when entering and restores it when leaving. You can then add it to `parse_block`, that should be it. 

I would then add helper methods to the `ParserState` to test if a variable has already been declared and to register a new one `register_scoped_variable`, `register_hoisted_variable`. That should be sufficient.

crates/rslint_parser/src/state.rs
```diff
@@ -374,3 +378,224 @@ impl ChangeParserState for EnableStrictMode {
 		state.strict = value.0
 	}
 }
+
+/// Data structure responsible to track bindings inside their scopes
+#[derive(Debug, PartialEq)]
+pub struct BindingsTracking {
+	/// Scopes where hoisted variables are tracked. We have two scopes
+	hoisted_scopes: [HoistedVars; 2],
```
MichaReiser, line 386: Why two? What are the two states?

crates/rslint_parser/src/state.rs
```diff
@@ -374,3 +378,224 @@ impl ChangeParserState for EnableStrictMode {
...
+		}
+	}
+}
+
+impl BindingsTracking {
+	/// Checks whether a binding clashes with one that has been already tracked
+	pub fn clashes(&self, binding_name: &str, is_var: bool) -> Option<&Range<usize>> {
+		if is_var {
+			None
+		} else if self.blocks.len() > 2 {
```
MichaReiser, line 410:
```
let a;
var a;
```

The `var a` clashes with the `let a`;

crates/rslint_parser/src/state.rs
```diff
@@ -374,3 +378,224 @@ impl ChangeParserState for EnableStrictMode {
...
 	}
 }
+
+/// Data structure responsible to track bindings inside their scopes
+#[derive(Debug, PartialEq)]
+pub struct BindingsTracking {
+	/// Scopes where hoisted variables are tracked. We have two scopes
+	hoisted_scopes: [HoistedVars; 2],
+	/// Whether the parser is inside a block. If so, hoisted vars checked against the second [ScopedVars]
+	is_inside_block: bool,
```
MichaReiser, line 388: Shouldn't this be `is_inside_function` in which case we must test against the hoisted vars in the function scope? You can use `in_function`  from the `ParserState`

MichaReiser (commented):
What's the impact on performance?

I'm a bit surprised that this is only fixing 4 tests. Shouldn't this fix much more tests?

ematipico:
> I'm a bit surprised that this is only fixing 4 tests. Shouldn't this fix much more tests?

The majority of redeclarations are around functions and classes.

I will make a separate PR for them

crates/rslint_parser/src/state.rs
ematipico: We have a top-level scope and a scope defined by a block.

This is a valid syntax:

```js
var a;
{
 	let a;
}
```

This is invalid and the redeclaration is around the second `var`:

```js
var a;
{
  	var a;
 	let a;
}
```

And this is also invalid: 

```js
var a;
{
  	{ var a; }
 	let a;
}
```

Even though the second `var` is inside a block, it gets hoisted to the first one and `let` should fire a syntax error.

crates/rslint_parser/src/state.rs
ematipico: A function is not the only one that can create a new scope. There are more cases:

- curly brackets `{  }`
- loops such as `while`, `for..`, etc.

MichaReiser: That's right. These could all use the `EnterScope` `ChangeParserState` helper.

crates/rslint_parser/src/state.rs
ematipico: > The 1. must be re-set whenever we enter a function. You can achieve that by storing the existing map in EnterFunction and restoring it when we leave the function.

The solution proposed is only applicable to functions but not to other cases, for example:

```js
{
  function f() { var b; } 
  let b
}
```

In this case the solution would work and each scope would have its map of hoisted variables.

Unfortunately it's not the same for the following code:

```js
{
  { var b; }
  let b;
}
```

Here `{ var b; }`, the variable gets hoisted to the first block and `let b` can't be declared again.

MichaReiser:
```
{
  { var b; }
  let b;
}
```

The `var b` gets hoisted into the global scope. Which is the default `hoisted_scope` if the parser isn't inside of a function.

ematipico: > The var b gets hoisted into the global scope.
I should have added some variable afterwards 😱 you're right!

crates/rslint_parser/src/state.rs
ematipico: Yes, that was an error in my implantation. One scope for hoisted variables should be enough

crates/rslint_parser/src/state.rs
```diff
@@ -120,6 +128,33 @@ impl ParserState {
 		self.parsing_context
 			.contains(ParsingContextFlags::IN_BINDING_LIST_FOR_SIGNATURE)
 	}
+
+	/// Checks if a binding has been already registered
+	pub fn clashes_with_defined_binding(
+		&self,
+		identifier_name: &str,
+		is_var: bool,
```
MichaReiser, line 136: Should we add an enum with the variants `Lexical` and `Hoisted` and then expose a single `register_name(name, VariableKind)` and `get_existing_name(name, VariableKind)` methods?

crates/rslint_parser/src/state.rs
```diff
@@ -482,3 +532,25 @@ impl ChangeParserState for EnterClassPropertyInitializer {
 		state.parsing_context = value.0;
 	}
 }
+
+#[derive(Debug, Clone, Default)]
+pub struct EnterScopeSnapshot {
+	// parsing_context: ParsingContextFlags,
```
MichaReiser, line 538: Delete?

crates/rslint_parser/src/state.rs
```diff
@@ -418,26 +453,39 @@ impl ChangeParserState for EnterParameters {
...
+		state.scoped_variables = value.scoped_variables;
 	}
 }
 
 #[derive(Debug, Clone, Default)]
 pub struct EnterFunctionSnapshot {
 	parsing_context: ParsingContextFlags,
 	labels: HashMap<String, Range<usize>>,
+	/// scoped variables that belong to the function
+	scoped_variables: HashMap<String, Range<usize>>,
```
MichaReiser, line 488: We need to capture the hoisted variables too.

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -44,6 +44,20 @@ pub(crate) fn parse_binding(p: &mut Parser) -> ParsedSyntax {
...
+// test binding_redeclaration
+// var a; var a;
+// var a; { { var b; } { let b; } }
+// var a; { let a; }
+//
+// test_err binding_redeclaration
+// let a = 7; let a = 8;
+// var a; { { var b; } let b; }
+// var a; { var b; let b; }
+// var a; let a;
```
MichaReiser, line 57: Can we add a test that bindings correctly respect function boundaries?

MichaReiser:
Do you know why your changes arent fixing for example 

```
language/block-scope/syntax/redeclaration/function-name-redeclaration-attempt-with-var.js
```

Is it because we only register `let/const/var` bindings for now?

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -111,9 +125,28 @@ pub(crate) fn parse_identifier_binding(p: &mut Parser) -> ParsedSyntax {
 			}
 
 			let identifier_name = String::from(identifier_name);
+			let id = identifier_name.clone();
```
MichaReiser, line 128: Can we replace the `name_map` above with your solution.

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -111,9 +125,28 @@ pub(crate) fn parse_identifier_binding(p: &mut Parser) -> ParsedSyntax {
...
 			p.state
 				.name_map
 				.insert(identifier_name, identifier.range(p).as_range());
+
+			let is_var = parent == "var";
```
MichaReiser, line 133: It may make sense to introduce an enum for `duplicate_binding_parent`, something like `BindingContext`. Because there can be more (`function`, `class`, ...)

MichaReiser:
I really like the implementation :)

crates/rslint_parser/src/state.rs
```diff
@@ -456,13 +504,15 @@ impl ChangeParserState for EnterFunction {
 		EnterFunctionSnapshot {
 			parsing_context: std::mem::replace(&mut state.parsing_context, new_flags),
 			labels: std::mem::take(&mut state.labels),
+			scoped_variables: std::mem::take(&mut state.scoped_variables),
```
MichaReiser, line 507: You'll need to change where we call `EnterFunction`. The problem is, that it's currently only wrapping the body of the function but not its parameters, meaning, parameters get declared in the outer scope which isn't correct. 

You need to change this for functions, methods, arrow functions etc.

crates/rslint_parser/src/syntax/binding.rs
ematipico: What  `duplicate_binding_parent` used for?

ematipico:
> Do you know why your changes arent fixing for example
> 
> ```
> language/block-scope/syntax/redeclaration/function-name-redeclaration-attempt-with-var.js
> ```
> 
> Is it because we only register `let/const/var` bindings for now?

Exactly that. The issue I am facing is that, the very moment we track also bindings for functions, we start having a lot of regressions due to the issue I explained in the description of the PR.

crates/rslint_parser/src/state.rs
```diff
@@ -120,6 +128,33 @@ impl ParserState {
...
+			self.scoped_variables.get(identifier_name)
+		} else {
+			self.hoisted_variables
+				.get(identifier_name)
+				.or_else(|| self.scoped_variables.get(identifier_name))
+		}
+	}
+
+	/// Register a scoped variable
+	pub fn register_scoped_variable(&mut self, identifier_name: String, range: Range<usize>) {
```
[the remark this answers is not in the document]
ematipico: Yeah, let's follow the spec. Thank you!

crates/rslint_parser/src/state.rs
ematipico: If we decided to move `EnterFunction` before the parsing its body, would that break existing functionality?

crates/rslint_parser/src/state.rs
MichaReiser: We could even decide to re-use the same enum in the this weird `binding_parent` state (forgot its exact name). The enum could then have methods like `is_lexical()` and `is_hoisted` to determine how the binding must be declared. This way, the API become as simple as

```
get_binding(name, BindingParent::Function)
get_binding(name, BindingParent::Parameter)
register_binding(name, BindingParent::Function)
```

crates/rslint_parser/src/syntax/binding.rs
MichaReiser: You use it here :D To track if we're inside a `var`, `let` or `const` declaration. But we should extend it to support any possible binding parent / context.

crates/rslint_parser/src/state.rs
MichaReiser: Using `EnterScope` might be the best solution because arguments are always declared as lexical scoped, right?

ematipico: Yes

> Using `EnterScope` might be the best solution because arguments are always declared as lexical scoped, right?

Yes they are:

```js
var a = "3";

function f(a = "f") {
    console.log(a)
}

const d = (a = "33333") => {
    console.log(a)
}
f(); // logs "f"
d(); // logs "33333"
```

crates/rslint_parser/src/state.rs
```diff
@@ -134,10 +197,213 @@ impl ParserState {
...
+                        match binding_type {
+                            NameType::Hoisted => self
+                                .strict()
+                                .and_then(|_| self.lexical_names.get(identifier_name)),
+
+                            NameType::Function => {
+                                // At the top level of a function or script, inner function declarations are treated like var declarations.
+                                //
+                                // https://tc39.es/ecma262/multipage/syntax-directed-operations.html#sec-static-semantics-toplevelvardeclarednames
+                                if self.in_function_block() || self.strict().is_none() {
```
leops, line 235: If I'm reading this correctly if self.strict() returns None this will always enter the first branch (and always return None in that case), and the program can only enter the second branch if strict returns Some(_) so the two gets could be directly chained without calling self.strict() again

crates/rslint_parser/src/state.rs
```diff
@@ -55,19 +115,22 @@ impl Default for ParserState {
 impl ParserState {
     pub fn new(syntax: Syntax) -> Self {
         let mut state = ParserState {
-            parsing_context: ParsingContextFlags::ALLOW_OBJECT_EXPRESSION
+            parsing_context: (ParsingContextFlags::ALLOW_OBJECT_EXPRESSION
                 | ParsingContextFlags::INCLUDE_IN
-                | ParsingContextFlags::TOP_LEVEL,
+                | ParsingContextFlags::TOP_LEVEL)
+                - ParsingContextFlags::IN_FUNCTION_BLOCK,
```
yassere, line 121: What's the purpose of subtracting the flag here? Isn't that a no-op since the `IN_FUNCTION_BLOCK` flag isn't part of the union?

crates/rslint_parser/src/state.rs
```diff
@@ -32,11 +35,68 @@ pub struct ParserState {
...
+
+impl Default for BindingContext {
+    fn default() -> Self {
+        Self::Hoisted
+    }
+}
+
+/// Identifies the types of bindings encountered
+#[derive(Debug, Clone, PartialEq)]
+pub enum NameType {
```
MichaReiser, line 78: I get the impression that `NameType` and `BindingContext` are overlapping. 

My understanding is that there it's relevant to know the context in which a binding appears because there are different rule depending if you are:

* Top-Level
* Inside a Function body
* Inside a block statement

I believe that's what we want to represent with `BindingContext`.

Then, there's `NameType` (or `BindingType`?). It specifies what kind of binding it is. The different bindings seem to be

```
  Var
  Let,
  Const,
  Argument,
  FunctionId,
  CatchDeclaration,
  Import,
  // Exports?
```

I think it would make sense to restructures the enum in that way and pass `NameType` directly to the register/clash functions

(We can also use `Lexical(LexicalType)` for all cases where the binding always is a lexical binding and only list the ones where the binding type depends on the binding context).

crates/rslint_parser/src/state.rs
```diff
@@ -134,10 +197,213 @@ impl ParserState {
...
     }
+
+    pub fn in_function_block(&self) -> bool {
+        self.parsing_context
+            .contains(ParsingContextFlags::IN_FUNCTION_BLOCK)
+    }
+
+    /// Checks if a binding has been already registered
+    pub fn clashes_with_defined_name(&self, identifier_name: &str) -> Option<&Range<usize>> {
+        if let Some(binding_context) = self.binding_context() {
```
MichaReiser, line 220: I recommend calling `expect()` on the `binding_context` to prevent that we miss any registrations. Not setting the binding context and then calling this method is a logic error.

crates/rslint_parser/src/state.rs
```diff
@@ -134,10 +197,213 @@ impl ParserState {
...
+                            }
+
+                            NameType::Lexical(_) => self.lexical_names.get(identifier_name),
+                            _ => self.lexical_names.get(identifier_name),
+                        }
+                    } else {
+                        None
+                    }
+                }
+                BindingContext::Arguments => self
```
MichaReiser, line 252: I find this whole block hard to read. It seems that a lot of the validation depends if strict mode is enabled or not. 

It may help to split the logic then: 

* Validation that must always be done regardless if it's strict mode or not
* Validation that only must be done in strict-mode. 

Or use ` NameType::Hoisted if self.strict.is_some() => ` instead of the `and_then` might help as well.

It could also improve readability if you collapse match arms that have the same behaviour.

MichaReiser (changes requested):
Wow, this turned out to be a lot of changes. 

I spotted a few logical errors and believe that we can simplify the design even further. Mainly, I have the feeling that there's a lot of overlap between `BindingContext` and `NameType`. My understanding is that `NameType` says what kind of binding it is. Ideally, we would only have `Hoisted` and `Lexical` but, unfortunately, the type of binding sometimes differs depending on the `BindingContext`. One such example are function ids, where the id either gets lexically bound or is hoisted. 

That's why I think that `BindingContext` should only hold the values `TopLevel`, `Block`, `FunctionBlock`. 

`NameType` then lists the different bindings. The generic ones `Hoisted` and `Lexical` for the cases where the binding type doesn't depend on the `BindingContext`. But it also has variants for when it is less clear, like `FunctionId` (and maybe others).

crates/rslint_parser/src/state.rs
ematipico: Yes, that change was not meant to be there. Fixed!

crates/rslint_parser/src/state.rs
ematipico: Good catch, it was a copy/paste gone wrong. Do you agree with me that we can write like this?

```rs
if self.in_function_block() || self.strict().is_none() {
    self.lexical_names.get(identifier_name)
} else {
    self.lexical_names
        .get(identifier_name)
        .or_else(|| self.hoisted_names.get(identifier_name))
}
```

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -227,7 +282,9 @@ fn parse_function(p: &mut Parser, m: Marker, kind: FunctionKind) -> CompletedMar
 // }
 pub(super) fn parse_function_body(p: &mut Parser, flags: SignatureFlags) -> ParsedSyntax {
     p.with_state(EnterFunction(flags), |p| {
-        parse_block_impl(p, JS_FUNCTION_BODY)
+        p.with_state(EnterLexicalScope(BindingContext::FunctionBlock), |p| {
```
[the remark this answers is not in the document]
ematipico: I explained in the description why it's not possible to join the two:

- https://github.com/rome/tools/blob/main/crates/rslint_parser/src/syntax/function.rs#L248-L249
- https://github.com/rome/tools/blob/main/crates/rslint_parser/src/syntax/function.rs#L380-L382

We can't apply a new lexical scope when we are inside the body, we have to do that before we enter the parameters. When we enter the function body, we need to create a new name scope.

crates/rome_formatter/tests/specs/js/array/binding_pattern.js
```diff
@@ -1,3 +1,3 @@
 let [a,b]=c;
-let [a,...b]=c;
+let [d,...e]=c;
```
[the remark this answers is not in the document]
ematipico: How would you suggest to do that? All tests will fail if I don't update the snapshots

crates/rslint_parser/src/state.rs
yassere: Another possible readability improvement here is to reduce rightward drift by using the `?` operator in match expressions and replacing constructs like:
```rs
if let Some(binding_type) = self.binding_type() {
    match binding_type {
        // …
    }
} else {
    None
}
```
with
```rs
match self.binding_type()? {
    // …
}
```

crates/rslint_parser/src/syntax/binding.rs
```diff
@@ -73,49 +93,44 @@ pub(crate) fn parse_identifier_binding(p: &mut Parser) -> ParsedSyntax {
...
-                        &format!(
-                            "a second declaration of `{}` is not allowed",
-                            identifier_name
-                        ),
-                    );
-                p.error(err);
-                identifier.change_to_unknown(p);
-                return identifier;
-            }
+        let err = if binding_context.is_some() {
```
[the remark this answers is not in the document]
ematipico: Small error of mind, we should always have a binding context, I will remove the `Option<>`.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -622,6 +640,8 @@ fn parse_class_member_impl(
                         member_marker.complete(p, JS_GETTER_CLASS_MEMBER)
                     } else {
                         let has_l_paren = p.expect(T!['(']);
+                        let p =
```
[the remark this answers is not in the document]
ematipico: > The BindingContext::Arguments is misleading because the body won't be parsed in an Arguments context.

It's actually the whole point, but I can see why it can be misleading. The thing is that they must be parsed in two different context because parameters behave differently: they are hoisted but can't clash in strict mode.


> I propose to remove the BindingContext arguments from EnterHoistedScope and instead set the BindingContext as part of EnterParameters.

Please read the description of the issue, where I explained why we can't do that, unless we can make the snapshot changes of `EnterParamters` last until the function body is parsed.

> Nit: Can we rename BindingContext::Arguments to Parameters to align it with EnterParameters. Arguments are what we pass in a call expressions. Parameters are the variables defined in a function signature (parametrises the functions behaviour)

Makes sense!

crates/rome_formatter/tests/specs/js/array/binding_pattern.js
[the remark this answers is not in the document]
MichaReiser: Split out the change in another PR and land that first

crates/rslint_parser/src/state.rs
```diff
@@ -450,6 +731,9 @@ bitflags! {
         /// Whether we potentially are in a place to parse an arrow expression
         const POTENTIAL_ARROW_START = 1 << 11;
 
+        /// Whether the parser is inside a function body
+        const IN_FUNCTION_BLOCK = 1 << 12;
```
MichaReiser, line 735: How's that different from function body

Is it possible to derive it from in_function and BindingContext::Blick

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -94,20 +98,16 @@ pub(crate) fn parse_import(p: &mut Parser) -> ParsedSyntax {
...
-    debug_assert!(p.state.name_map.is_empty());
-    p.state.duplicate_binding_parent = Some("import");
-
-    parse_import_clause(p).or_add_diagnostic(p, |p, range| {
-        expected_any(
-            &["default import", "namespace import", "named import"],
-            range,
-        )
-        .to_diagnostic(p)
+    p.with_state(EnterVariableDeclaration(NameType::Module), |p| {
```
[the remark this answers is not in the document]
ematipico: Yeah this will be gone

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -183,7 +184,9 @@ pub(crate) fn parse_statement(p: &mut Parser, context: StatementContext) -> Pars
             export
         }),
         T![;] => parse_empty_statement(p),
-        T!['{'] => parse_block_stmt(p),
+        T!['{'] => p.with_state(EnterLexicalScope(BindingContext::Block), |p| {
+            parse_block_stmt(p)
```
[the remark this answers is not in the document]
ematipico: We can't, because this function is also used inside to parse the function body, where we need to create a `BindingContext::FunctionBlock` instead. Do you want to add a second argument to `parse_block_stmt`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1264,9 +1260,13 @@ fn parse_for_head(p: &mut Parser) -> JsSyntaxKind {
         return JS_FOR_STATEMENT;
     }
 
+    let p = &mut *p.with_scoped_state(EnterHoistedScope(BindingContext::LoopStatements));
```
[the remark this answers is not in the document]
ematipico: Probably you executed some other code before which conflicted with that snippet. That is valid code

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -181,13 +191,21 @@ fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
...
+    p.with_state(EnterLexicalScope(BindingContext::Block), |p| {
+        extends_clause(p).ok();
+    });
 
     if implements_clause(p).is_present() && TypeScript.is_unsupported(p) {
         class_is_valid = false;
     }
 
     p.expect(T!['{']);
+    let p = &mut *p.with_scoped_state(EnterLexicalScope(BindingContext::Block));
```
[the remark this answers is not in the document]
ematipico: Good catch, not needed!

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1465,6 +1468,7 @@ fn parse_switch_clause(
     first_default: &mut Option<CompletedMarker>,
 ) -> ParsedSyntax {
     let m = p.start();
+    let p = &mut *p.with_scoped_state(EnterHoistedScope(BindingContext::Block));
```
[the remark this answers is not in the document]
ematipico: Yes, `switch` creates a new hoisted scope. You probably run some other snippet before, which caused some conflict.

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1621,7 +1633,9 @@ fn parse_catch_clause(p: &mut Parser) -> ParsedSyntax {
     p.bump_any(); // bump catch
 
     parse_catch_declaration(p).ok();
-    parse_block_stmt(p).or_add_diagnostic(p, js_parse_error::expected_block_statement);
+    p.with_state(EnterLexicalScope(BindingContext::Block), |p| {
```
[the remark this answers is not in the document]
ematipico: `parse_block_stmt` is used for normal blocks and function blocks too. This won't work. Do you want to add a new parameter to `parse_block_stmt`?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -1634,7 +1648,7 @@ fn parse_catch_declaration(p: &mut Parser) -> ParsedSyntax {
     let declaration_marker = p.start();
 
     p.bump_any(); // bump (
-
+    let p = &mut *p.with_scoped_state(EnterHoistedScope(BindingContext::Block));
```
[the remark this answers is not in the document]
ematipico: `try/catch/finally` all create a new lexical scope. `catch` also creates a hoisted scope:

This code will fail:

```js
try {} catch (e) { let e; }
```

Because we can't redeclare `e`

crates/rslint_parser/src/syntax/stmt.rs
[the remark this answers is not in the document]
ematipico: See https://github.com/rome/tools/pull/1929#discussion_r788011748

crates/rslint_parser/src/state.rs
```diff
@@ -32,11 +35,68 @@ pub struct ParserState {
     strict: Option<StrictMode>,
     /// The exported default item, used for checking duplicate defaults
     pub default_item: Option<Range<usize>>,
+    pub(crate) no_recovery: bool,
+    /// Tracks the binding variable inside a declaration list: "var a, b, c;"
+    binding_type: Option<NameType>,
```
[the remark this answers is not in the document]
ematipico: See https://github.com/rome/tools/pull/1929#discussion_r788011748

crates/rslint_parser/src/state.rs
```diff
@@ -32,11 +35,68 @@ pub struct ParserState {
...
+    /// import modules bindings `import/from`
+    Module,
+    /// For generic blocks  { }
+    Block,
+    /// A block of a function, which applies different rules compared to [BindingContext::Block]
+    FunctionBlock,
+    /// Track statements that involve a loop such as `for` loops
+    LoopStatements,
+    /// Function arguments
+    Arguments,
```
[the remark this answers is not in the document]
ematipico: Unfortunately not, because parameters need a special treatment: even though they are hoisted, they can't be duplicated. Whilst normal hoisted names can be re-declared without issue.

crates/rslint_parser/src/state.rs
ematipico: I like the suggestion, I think we can cover all the cases with this naming convention. I knew e different pair of eyes would have helped :) 

Thank you

crates/rslint_parser/src/state.rs
ematipico: This won't be needed anymore, I will remove the `Option`

crates/rslint_parser/src/syntax/function.rs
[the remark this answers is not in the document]
MichaReiser: It may then make sense to split EnterFunction and EnterFunctionBody

crates/rslint_parser/src/state.rs
ematipico: `EnterFunction` is the snapshot that sets the `IN_FUNCTION` bitflag, and `EnterFunction` at the moment, it's not what I need.

ematipico:
I am going to pause the works around this PR. It needs some decisions because:
- there still, probably some false positives and some false negatives
- it requires some more engineering works to the whole parsing infrastructure
- this solution could be achieved easily with a second pass

MichaReiser:
@ematipico is it OK to close this PR now that @xunilrj is working on scope resolution?

PR #1932, opened 2022-01-12
chore: add `unicode` task, remove python script

### LINKED ISSUE #1930: Move internal python script to Rust
### Discussed in https://github.com/rome/tools/discussions/1926

<div type='discussions-op-text'>

<sup>Originally posted by **ematipico** January 11, 2022</sup>
Hi @rome/staff 

[This recent PR](https://github.com/rome/tools/pull/1910) showed us that we inherited some internal script that requires `python`.

This, as a consequence, requires for core contributors and staff to have python installed in their machines. This *could* create some friction, even though it's not the end of the world.

I wanted to ask you what would prefer:

1. keep `python` inside our project and update the `CONTRIBUTING.md` with mention of this script and saying that developers require having python installed

2. move the script to Rust

I vote for `2.`
</div>

The result of discussion is that the team would like to [move the python script](https://github.com/rome/tools/blob/main/crates/rslint_lexer/scripts/gen.py) to Rust.

The steps to move this script to Rust are:
- move the script inside the `xtask` folder, where we have all our internal scripts
- add a new command called `unicode`, which can be called with the following command `cargo xtask unicode`
- the new command will call the new script, which can live inside a single file or folder. I don't think it matters much as it's a small script (probably)
- the script will have to update the [`tables.rs` file](https://github.com/rome/tools/blob/main/crates/rslint_lexer/src/tables.rs)

It would be great if only new contributors would tackle this issue, as it's not a high priority task and it's really self contained. 

Any questions or doubts, please ask here!

malobre:
## Summary
- Closes #1930.

## Test Plan
- [x] `cargo check`
- [x] `cargo fmt`
- [x] `cargo test --package rslint_lexer`
- [x] Compare generated tables: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=912a6a48a2c98a9cc6b338e9888cda6e)


xtask/src/unicode.rs
```diff
@@ -0,0 +1,112 @@
...
+				else { Equal }
+			}).is_ok()
+		}
+
+		pub mod derived_property {
+			#(#properties)*
+		}
+	};
+
+	let pretty = prettyplease::unparse(&syn::parse2::<syn::File>(tokens)?);
```
MichaReiser, line 82: You can use the `reformat` function that all other codegens are using (avoids yet another dependency that)

https://github.com/rome/tools/blob/7f555aa275b3629cc6cbc9a74f7dc2e9b88e6f68/xtask/src/lib.rs#L52

malobre: I thought `prettyplease` would be better because it wouldn't rely on invoking `rustfmt` binary (feels hacky), plus it only depends on `syn` and `proc-macro2`, according to the readme it's also faster than `rustfmt`.

However I understand this might bloat the PR and would be better discussed in its own issue / PR.

malobre:
I cleared the code a bit and added comments.

malobre:
There seems to be some discrepancies in the generated code, I'm looking into it.

malobre:
I've updated the code to match the behaviour of the python script regarding the collapsing of sequential ranges.
The generated arrays are now [identical](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=782b8c0a596346c881cb4316fc90b0c5).

malobre:
There, I'm happy with the code. No more changes I swear ! :laughing:

malobre:
Rebased on `main`.

Cargo.lock
```diff
@@ -1087,6 +1087,16 @@ dependencies = [
...
+[[package]]
+name = "prettyplease"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f3ebd569945e276e2c8000677484f50f5c242b3495929ae1c7d42d89778939a"
+dependencies = [
+ "proc-macro2",
+ "syn",
+]
+
```
MichaReiser, line 1099: I would favour it if we swap to `prettyplease` as part of another PR.

crates/rslint_lexer/src/tables.rs
```diff
@@ -1,468 +1,1427 @@
...
-
-fn bsearch_range_table(c: char, r: &[(char,char)]) -> bool {
+fn bsearch_range_table(c: char, r: &[(char, char)]) -> bool {
     use core::cmp::Ordering::{Equal, Less, Greater};
-    r.binary_search_by(|&(lo,hi)| {
-        // Because ASCII ranges are at the start of the tables, a search for an
-        // ASCII char will involve more `Greater` results (i.e. the `(lo,hi)`
-        // table entry is greater than `c`) than `Less` results. And given that
-        // ASCII chars are so common, it makes sense to favor them. Therefore,
-        // the `Greater` case is tested for before the `Less` case.
```
MichaReiser, line 15: Should we keep this comment?

crates/rslint_lexer/src/tables.rs
```diff
@@ -1,468 +1,1427 @@
...
+        ('𞺫', '𞺻'),
+        ('🯰', '🯹'),
+        ('𠀀', '𪛟'),
+        ('𪜀', '𫜸'),
+        ('𫝀', '𫠝'),
+        ('𫠠', '𬺡'),
+        ('𬺰', '𮯠'),
+        ('丽', '𪘀'),
+        ('𰀀', '𱍊'),
+        ('\u{e0100}', '\u{e01ef}'),
```
MichaReiser, line 769: What's the reason that this entry uses unicode escape but you migrated all other to unicode characters

xtask/src/unicode.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /// Save the properties on disk for later usage, see [`Self::from_cache`].
+    /// # Errors
+    /// Return an error if saving to disk fails.
+    pub fn save_cache(&self) -> Result<()> {
+        std::fs::write(Self::path(), &self.raw)?;
+
+        Ok(())
+    }
```
MichaReiser, line 109: I would make these private methods and expose a single public method that retrieves the properties from cache or downloads them if missing. 

We may want to print to stdout in the case the script uses a cached file. Otherwise someone may run the script to update the unicode table but see no changes and wonder why.

xtask/src/unicode.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    let tokens = quote! {
+        //! Autogenerated file, do not edit by hand.
+        //! Run `cargo xtask unicode` and recommit this file when Unicode support has changed.
+        #![allow(missing_docs, non_upper_case_globals, non_snake_case)]
+
+        fn bsearch_range_table(c: char, r: &[(char,char)]) -> bool {
+            use core::cmp::Ordering::{Equal, Less, Greater};
+            r.binary_search_by(|&(lo,hi)| {
+                // Because ASCII ranges are at the start of the tables, a search for an
```
MichaReiser, line 50: Ah, the comment is still here. That's good enough.

xtask/src/unicode.rs
```diff
@@ -0,0 +1,168 @@
...
+            let fn_ident = quote::format_ident!("{}", property);
+            let table_ident = quote::format_ident!("{}_table", property);
+
+            Ok(quote! {
+                pub const #table_ident: &[(char, char)] = &[
+                    #(#ranges),*
+                ];
+
+                pub fn #fn_ident(c: char) -> bool {
+                    super::bsearch_range_table(c, #table_ident)
```
MichaReiser, line 36: The implementation uses a binary search. How does this script guarantee that the properties are sorted. Is this guaranteed by the unicode table? Would it make sense to re-sort it or at least assert that the entries are sorted?

MichaReiser (commented):
I'm not familiar with the python script but only have some overall comments:

* Did you run your new script. Does it generate the exact same output?

xtask/src/unicode.rs
malobre: The unicode table is sorted, but I couldn't find any mention of it in the [Format Convention](https://www.unicode.org/reports/tr44/#Format_Conventions).
I'll add a call to `slice::sort_unstable`.

crates/rslint_lexer/src/tables.rs
malobre: This is not by choice, I wondered about this too and I'm not sure why but only some char are escaped. This might be related to https://github.com/rust-lang/rust/issues/60495.

malobre:
I have addressed all comments (except for the `char` escaping situation).
@MichaReiser The generated tables are identical: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=912a6a48a2c98a9cc6b338e9888cda6e

MichaReiser (approved):
Excellent work! Thank you so much for working on this.

PR #1933, opened 2022-01-12
NEWLINE and MULTILINE_COMMENT tokens on the trivia api

xunilrj:
PS: The changes are quite small actually. But the new token changed every ```.rast``` file.

Exposing ```NEWLINE``` token as ```GreenTokenTrivia::NewLine``` and ```SyntaxTriviaPieceNewline```; and ```MULTILINE_COMMENT``` as ```SyntaxTriviaPieceComments::has_newline()```.

@ematipico and @Boshen and @leops Any chance our formatter can take advantage of these changes?

### Less allocs

Now ```get_trivia``` does not allocate memory needlessly.

Before:

```
> cargo run -p xtask --release --features "dhat-on" -- coverage-libs --filter=jquery
...
dhat: Total:     33,774,984 bytes in 50,388 blocks
...
```

After
```
> cargo run -p xtask --release --features "dhat-on" -- coverage-libs --filter=jquery
...
dhat: Total:     33,703,768 bytes in 48,100 blocks
...
```

This difference is exactly the numbers of allocs we are doing according ```valgrind```.

Before 
image

After
image

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -30,8 +30,8 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub enum TriviaPiece {
-	Newline(u32), // length
-	Whitespace(u32), // length
+	Newline(u32),        // length
+	Whitespace(u32),     // length
 	Comments(u32, bool), // length, has_newline
 }
 
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -30,22 +30,39 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub enum TriviaPiece {
-	Whitespace(u32),
-	Comments(u32),
+	Newline(u32), // length
+	Whitespace(u32), // length
+	Comments(u32, bool), // length, has_newline
 }
 
 impl TriviaPiece {
 	#[inline]
 	pub fn text_len(&self) -> TextSize {
 		match self {
+			TriviaPiece::Newline(n) => (*n).into(),
 			TriviaPiece::Whitespace(n) => (*n).into(),
-			TriviaPiece::Comments(n) => (*n).into(),
+			TriviaPiece::Comments(n, _) => (*n).into(),
 		}
 	}
 }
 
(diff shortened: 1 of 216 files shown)


crates/rome_rowan/src/api.rs
```diff
@@ -30,23 +30,40 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub enum TriviaPiece {
-	Whitespace(u32),
-	Comments(u32),
+	Newline(u32),        // length
+	Whitespace(u32),     // length
+	Comments(u32, bool), // length, has_newline
```
MichaReiser, line 35: Nit: Rename to Comment? all other variants are singular. 

Would it make sense to split `Comment` into `SingleLine` and `MultiLineComment`. Because that's exactly what the bool represents. @leops would that simplify your logic that determines the number of new lines between two tokens?

leops:
The logic for counting lines between tokens isn't entirely decided yet but I think it will as least have a safe path that scans over the trivias to count newlines so having them as separate trivia types will help yes

crates/rome_rowan/src/api.rs
xunilrj: I thought about that... in the ```kind``` we split because all "data" need to be conveyed in the ```kind``` itself; but here we have more freedom. I think like this is rust-natural model. What do you think?

MichaReiser: I mainly find it a bit strange that we separate `Newline` and `Whitespace` but don't make the same separation for `Comment`. It also increases the size of `TriviaSize` by probably 2bytes because of alignment. 

But I don't have a strong opinion on this.

xunilrj: Many is still the prevalent... mainly because of pointer alignment.

```
>cargo +nightly rustc -p rome_rowan -- -Zprint-type-sizes
...
print-type-size type: `green::token::GreenTokenTrivia`: 16 bytes, alignment: 8 bytes
print-type-size     discriminant: 1 bytes
print-type-size     variant `Many`: 15 bytes
print-type-size         padding: 7 bytes
print-type-size         field `.0`: 8 bytes, alignment: 8 bytes
print-type-size     variant `Whitespace`: 7 bytes
print-type-size         padding: 3 bytes
print-type-size         field `.0`: 4 bytes, alignment: 4 bytes
print-type-size     variant `Comment`: 7 bytes
print-type-size         field `.1`: 1 bytes
print-type-size         padding: 2 bytes
print-type-size         field `.0`: 4 bytes, alignment: 4 bytes
```

MichaReiser (approved):

ematipico (commented):
She should really start adding tests for comments, because we can't actually the results of this changes on the green trees that we have now.

PR #1935, opened 2022-01-13
feature(rslint_parser): Introduce JSParameter node

MichaReiser:
## Summary

Adds a new `JsParameter` node as discussed in #1868 . The main reason for the `JSParameter` node is to have a node where we can attach the typescript type annotation (which makes it different from a binding for which type annotations are not allowed).

## Test Plan

Verified the RAST files.


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -901,11 +900,15 @@ fn parse_constructor_parameter(p: &mut Parser) -> ParsedSyntax {
 			}
 		}
 
-		parse_formal_param_pat(p).or_add_diagnostic(p, expected_parameter);
+		let parameter = parse_parameter(p).or_add_diagnostic(p, expected_parameter);
+
+		if let Some(parameter) = parameter {
+			parameter.undo_completion(p).abandon(p);
+		}
```
ematipico, line 907: Why do we abandon the node when we have something? 🤔

ematipico (approved):
It looks good to me. Why are there changes around `test262` sub module?

crates/rslint_parser/src/syntax/class.rs
MichaReiser: `TsConstructorParam` has the same structure as `JsParameter` except that it has two additional fields. Abandoning allows to reuse the `JsParameter` logic but we discard the node, so that the children are directly attached to the `TsConstructorParam` (the rule isn't `TsConstructorParam = .. js_param: JsParameter` but `TsConstructorParam = ... binding: JsAnyBindingPattern...`

MichaReiser:
> It looks good to me. Why are there changes around `test262` sub module?

Uh, I probably updated them by accident. I reverted the change.

PR #1936, opened 2022-01-13
Top-level await

MichaReiser:
## Summary

Enables top-level await for modules and implements the `Async` and `Yield` parametrization of the `FunctionDeclaration`, `FunctionExpression`, and `ArrowFunctionExpression` according to the specification. 

## Test Plan

The only remaining top-level await tests that are failing now are related to regex parsing.


crates/rslint_parser/src/syntax/decl.rs
```diff
@@ -1,178 +0,0 @@
-//! Class and function declarations.
-
-use super::binding::parse_binding_pattern;
-use super::expr::parse_expr_or_assignment;
```
MichaReiser, line 4: Moved to `function.rs`.

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -231,3 +273,216 @@ pub(super) fn is_at_async_function(p: &mut Parser, should_check_line_break: Line
...
+	list_kind: JsSyntaxKind,
+) {
+	let mut first = true;
+	let has_l_paren = p.expect(T!['(']);
+
+	p.with_state(
+		EnterParameters {
+			signature_flags: flags,
+			allow_object_expressions: has_l_paren,
+		},
```
MichaReiser, line 373: This is the main change. The rest is copy pasta from `decl.rs`

MichaReiser:
The regression is again a test that previously didn't parse at all but now fails because we're aren't correctly handling escaped contextual keywords. This is already tracked in #1909 and I don't intend to fix this as part of this PR.

xunilrj (approved):

PR #1940, opened 2022-01-13
feat: use rslint_errors Diagnostic in rome_analyze

yassere:
## Summary

- An `AnalyzeDiagnostic` now uses an `rslint_errors::Diagnostic` internally. That `Diagnostic` format may not perfectly match our needs, but it makes sense to unify around a single core diagnostics type for all of our crates. We can either modify it incrementally or replace it later, as needed.

- Analyzers now return `Option<Analysis>` instead of `Result<Analysis>`.
    - As analyzers are meant to inspect the syntax tree of an entire file, it’s often not ideal for them to bail early using `?` on fallible functions (e.g. when attempting to access the children of an AST node). It’s preferable for errors to be collected and included as part of the normal return value (similar to how our parser works). TODO: Add error-handling infrastructure for `Analysis`
    - The return type is an `Option` as an ergonomic way of returning an “empty” analysis. An analyzer might only be applicable at all under certain conditions and would want to return early if not. E.g. `context.maybe_get_foo_details()?`

- A new `AnalyzerContext` is now (cheaply) created before calling each analyzer and contains a reference to the analyzer being called. This allows for context methods to access information about the analyzer (like its name) under the hood. Same for `AssistContext`

- Published LSP diagnostics include the analyzer name as the `code` and “rome” as the `source`

### DIFF
--- a/crates/rome_analyze/src/analyzers.rs
+++ b/crates/rome_analyze/src/analyzers.rs
@@ -31,7 +31,7 @@ pub struct AnalyzerContext<'a> {
 }
 
 impl<'a> AnalyzerContext<'a> {
-	pub fn new(
+	pub(crate) fn new(
 		analysis_server: &'a AnalysisServer,
 		file_id: FileId,
 		analyzer: &'a Analyzer,
@@ -43,18 +43,24 @@ impl<'a> AnalyzerContext<'a> {
 		}
 	}
 
+	/// Get the root [SyntaxNode] for the file being analyzed
 	pub fn tree(&self) -> SyntaxNode {
 		self.analysis_server.parse(self.file_id)
 	}
 
+	/// Iterate over syntax nodes in this file that can be cast to T
 	pub fn query_nodes<T: AstNode>(&self) -> impl Iterator<Item = T> {
 		self.analysis_server.query_nodes(self.file_id)
 	}
 
+	/// Find the deepest AST node of type T that covers a TextRange
 	pub fn find_node_at_range<T: AstNode>(&self, range: TextRange) -> Option<T> {
 		self.analysis_server.find_node_at_range(self.file_id, range)
 	}
 
+	/// Create an error [Diagnostic] which includes the [FileId] of the file being analyzed
+	/// and the analyzer's name as a "code". The provided [Span] is recorded as the primary
+	/// label for the diagnostic.
 	#[must_use]
 	pub fn error(&self, span: impl Span, message: impl Into<String>) -> Diagnostic {
 		let code = self.analyzer.name;
--- a/crates/rome_analyze/src/analyzers.rs
+++ b/crates/rome_analyze/src/analyzers.rs
@@ -4,9 +4,10 @@ pub(crate) mod use_single_case_statement;
 pub(crate) mod use_while;
 
 use once_cell::sync::Lazy;
+use rslint_errors::{Diagnostic, Span};
 use rslint_parser::{AstNode, SyntaxNode, TextRange};
 
-use crate::{analysis_server::AnalysisServer, ActionCategory, AnalyzerResult, FileId};
+use crate::{analysis_server::AnalysisServer, ActionCategory, Analysis, FileId};
 
 static ALL_ANALYZERS: Lazy<Vec<Analyzer>> = Lazy::new(|| {
 	vec![
@@ -20,19 +21,25 @@ static ALL_ANALYZERS: Lazy<Vec<Analyzer>> = Lazy::new(|| {
 pub struct Analyzer {
 	pub name: &'static str,
 	pub action_categories: Vec<ActionCategory>,
-	pub(crate) analyze: fn(&AnalyzerContext) -> AnalyzerResult,
+	pub(crate) analyze: fn(&AnalyzerContext) -> Option<Analysis>,
 }
 
 pub struct AnalyzerContext<'a> {
 	pub file_id: FileId,
 	analysis_server: &'a AnalysisServer,
+	analyzer: &'a Analyzer,
 }
 
 impl<'a> AnalyzerContext<'a> {
-	pub fn new(analysis_server: &'a AnalysisServer, file_id: FileId) -> Self {
+	pub fn new(
+		analysis_server: &'a AnalysisServer,
+		file_id: FileId,
+		analyzer: &'a Analyzer,
+	) -> Self {
 		Self {
 			analysis_server,
 			file_id,
+			analyzer,
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


crates/rome_analyze/src/analyzers.rs
```diff
@@ -47,6 +54,12 @@ impl<'a> AnalyzerContext<'a> {
 	pub fn find_node_at_range<T: AstNode>(&self, range: TextRange) -> Option<T> {
 		self.analysis_server.find_node_at_range(self.file_id, range)
 	}
+
+	#[must_use]
+	pub fn error(&self, span: impl Span, message: impl Into<String>) -> Diagnostic {
```
ematipico, line 59: Documentation?

crates/rome_analyze/src/signals.rs
```diff
@@ -102,22 +60,63 @@ impl From<Action> for Analysis {
...
+		Analysis {
+			signals: iter.into_iter().map(Signal::Diagnostic).collect(),
+		}
 	}
 }
 
 #[derive(Debug, Clone)]
+/// Combines an rslint_errors Diagnostic with [SyntaxEdit] actions.
+///
+/// The suggestions on a [Diagnostic] are only suitable for text edits.
```
ematipico, line 91:
```suggestion
/// The suggestions on a [rslint_errors::Diagnostic] are only suitable for text edits.
```

crates/rome_analyze/src/signals.rs
```diff
@@ -102,22 +60,63 @@ impl From<Action> for Analysis {
...
+	pub fn with_actions(diagnostic: Diagnostic, actions: Vec<Action>) -> Self {
+		Self {
+			diagnostic,
+			actions,
+		}
+	}
+
+	pub fn range(&self) -> Option<TextRange> {
+		self.diagnostic.primary_text_range()
+	}
```
ematipico, line 117: Documentation?

crates/rome_analyze/src/signals.rs
```diff
@@ -168,3 +168,20 @@ impl From<Vec<Signal>> for Analysis {
 		Self { signals }
 	}
 }
+
+pub trait DiagnosticExt {
```
ematipico, line 172: I am not sure about this name. What `Ext` stands for? Extension? Extend? Can we add some documentation to the trait and its methods? Doing so might help to give a better name

crates/rome_lsp/src/utils.rs
```diff
@@ -72,3 +72,16 @@ pub(crate) fn text_action_to_lsp(
 		data: None,
 	}
 }
+
+pub(crate) fn diagnostic_to_lsp(
```
ematipico, line 76: Documentation?

ematipico: Maybe we can call it `send_diagnostic_to_lsp`?

crates/rome_analyze/src/assists.rs
```diff
@@ -22,34 +22,46 @@ pub struct AssistContext<'a> {
...
 
-	pub(crate) fn query_nodes<T: AstNode>(&self) -> impl Iterator<Item = T> {
-		self.analyzer_context.query_nodes()
+	pub fn query_nodes<T: AstNode>(&self) -> impl Iterator<Item = T> {
+		self.analysis_server.query_nodes(self.file_id)
 	}
 
-	pub(crate) fn find_node_at_cursor_range<T: AstNode>(&self) -> Option<T> {
-		self.analyzer_context.find_node_at_range(self.cursor_range)
+	pub fn find_node_at_range<T: AstNode>(&self, range: TextRange) -> Option<T> {
```
ematipico, line 58: Documentation?

crates/rome_analyze/src/assists.rs
```diff
@@ -22,34 +22,46 @@ pub struct AssistContext<'a> {
...
+		self.analysis_server.find_node_at_range(self.file_id, range)
+	}
+
+	pub fn find_node_at_cursor_range<T: AstNode>(&self) -> Option<T> {
```
ematipico, line 62: Documentation?

crates/rome_analyze/src/analyzers/use_single_case_statement.rs
```diff
@@ -1,22 +1,21 @@
...
 }
 
-fn analyze(ctx: &AnalyzerContext) -> Result<Analysis> {
+fn analyze(ctx: &AnalyzerContext) -> Option<Analysis> {
 	ctx.query_nodes::<ast::JsCaseClause>()
 		.filter(|n| n.consequent().len() > 1)
 		.map(|node| {
-			let message = "rome: A switch case should only have a single statement.";
-			Signal::diagnostic(node, message)
+			let message = "A switch case should only have a single statement.";
```
ematipico, line 17: I wonder if this string can be a `const` instead.

crates/rome_analyze/src/analyzers/no_double_equals.rs
```diff
@@ -30,8 +29,8 @@ fn analyze(ctx: &AnalyzerContext) -> Result<Analysis> {
 				return None;
 			}
 
-			let message = format!("rome: do not use the {} operator", op.text_trimmed());
-			let signal = Signal::diagnostic(op, message);
+			let message = format!("Do not use the {} operator", op.text_trimmed());
+			let signal: Signal = ctx.error(op, message).into_signal();
```
ematipico, line 33: Is `: Signal` needed here? Strange that the compiler is not able to infer the type

crates/rome_analyze/src/signals.rs
leops: For context, this is a convention of Rust since adding methods to a type is not allowed outside of the crate that declares it, a common way to do this is to create an "extension trait" owned by the extending crate that's then implemented on the external struct. For instance the standard library has `std::ascii::AsciiExt` (although this one is being deprecated since it ended up being implemented only on types owned by `std`), the `futures` library has `FutureExt` and `StreamExt`. The -Ext suffix is a generally accepted convention for these (https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html) although since it's not actually enforced by the language some libraries use different names (for instance `anyhow` chose to use `Context` instead of `ResultExt`)

yassere: This is the convention for defining an [Extension Trait](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html)

In this case, this trait might be temporary since we control `Diagnostic` and can modify it, but I didn't want to commit to making these changes outside of `rome_analyze` yet. I'll add some documentation.

crates/rome_analyze/src/analyzers/no_double_equals.rs
yassere: No, the explicit type shouldn't be necessary here. I forgot to remove it after I added the `into_signal` method.

yassere:
> Documentation?

@ematipico Yeah, good point. I should start transitioning this code out of WIP-mode and begin documenting it properly as it changes.

yassere:
I added some basic doc comments. Once I start to stabilize this and add tests, I'll flesh out the documentation more and add some doctests too.

PR #1941, opened 2022-01-14
chore: switch from tabs to spaces

### LINKED ISSUE #1689: 📎 Consider removing `hard_tab = true` from rustfmt config
### Description

The Rust community seems to be largely in favor of spaces for indentation. There are even `rust-analyzer` completions that assume space-based indentation and produce incorrect formatting when using tabs.

We should consider following convention here.

yassere:
## Summary

Closes #1689 

These are the main steps I took. If I forgot something, please let me know!
- Removed `hard_tab = true` then ran `cargo format`
- Ran `cargo xtask syntax`
- Replaced some tabs that `rustfmt` didn't catch for some reason
- Replaced tabs in comments (including inline parser tests) to avoid mixing tabs and spaces
- Ran `cargo xtask codegen` and updated the `.rast` files

I also updated the `.editorconfig` but I've never really used it, so could someone check if I got that right?

Later, I'll add this commit to a `.git-blame-ignore-revs` but this needs to squash merge first so I can get its hash.

## Test Plan

`cargo test`


.editorconfig
```diff
@@ -10,5 +10,8 @@ indent_size = 2
 
 # YAML doesn't support hard tabs 🙃
 # Templates that will be weird with hard tabs in the website editor
-[{*.yml, .github/**.md}]
+[{*.yml, .github/**.md, *.rs}]
 indent_style = space
+
+[*.rs]
+indent_size = 4
```
MichaReiser, line 17: Nooo! Why 4... I want two spaces... Just kidding :P

ematipico (approved):

PR #1943, opened 2022-01-14
Disallow declarations in single statement context

MichaReiser:
## Summary

[Declarations](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#prod-Declaration) are not allowed in a single line statement context:

```
if (true) const a = 10
```

However, there are some exceptions:

* `IfStatement` and `LabelledStatement` allow `FunctionDeclaration`s (not async nor generators) in a single line statement context
* Labelled function declarations are not permitted in a single line statement context (`if (true) label1: function a() {}`) 

This PR extends `parse_statement` to respect if it's in a single line statement context or not and only allow the permitted productions. 

## Test Plan

I added multiple spec tests to verify these rules

### DIFF
--- a/crates/rslint_parser/src/tests.rs
+++ b/crates/rslint_parser/src/tests.rs
@@ -11,6 +11,7 @@ use std::panic::catch_unwind;
 use std::path::{Path, PathBuf};
 
 #[test]
+#[ignore]
 fn parser_smoke_test() {
 	let src = r#"
 var asyncFn = async function () {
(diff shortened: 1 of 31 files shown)


crates/rslint_parser/src/tests.rs
```diff
@@ -11,6 +11,7 @@ use std::panic::catch_unwind;
 use std::path::{Path, PathBuf};
 
 #[test]
+#[ignore]
```
ematipico, line 14: Is it intentional?

ematipico (approved):

crates/rslint_parser/src/tests.rs
MichaReiser: nope

PR #1948, opened 2022-01-14
fix(rslint_parser): Avoid parsing assignment expr three times

MichaReiser:
## Summary

Reduces the time an assignment pattern is parsed to two:

1. Parse it as an expression to determine if it is a member expression or a single assignment
2. Re-Parse it only if it is a complex array or object pattern

Thanks @ematipico for making me aware of this issue


MichaReiser:
@NicholasLYang I managed to fix the regressions. The only other assertions that are still missing are the remaining assertions from `parse_identifier_binding` (CC: @ematipico )

https://github.com/rome/tools/blob/0a8844c21968e11a45794b914a15d8915351578f/crates/rslint_parser/src/syntax/binding.rs#L78-L111

ematipico (approved):

PR #1949, opened 2022-01-14
fix(rslint_parser): Limit continue to labelled iteration statements

MichaReiser:
## Summary

`continue` statements are only valid inside a loop or if they point to a labelled iteration statement (do, while, for) [spec](https://tc39.es/ecma262/multipage/syntax-directed-operations.html#sec-static-semantics-containsundefinedcontinuetarget).

```js
block_label: {
  loop_label: while true() {
    continue loop_label; // valid
    continue block_label; // invalid, not an iteration statement.
}
```

This PR changes the `label_set` to track what the kind of the labelled statement is (uses lookahead to determine the kind) and validates that any label used with `continue` points to an iteration.

Furthermore, fixes an issue where we didn't respect the negative lookahead of `let [` for `parse_expression`.

## Test Plan

Added new parser tests to cover these cases.


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -469,24 +499,48 @@ fn parse_continue_statement(p: &mut Parser) -> ParsedSyntax {
...
 
-    let end = if !p.has_linebreak_before_n(0) && p.at(T![ident]) {
+    let error = if !p.has_linebreak_before_n(0) && p.at(T![ident]) {
         let label_token = p.cur_tok();
+        let label_name = p.token_src(label_token);
+
+        let error = match p.state.label_set.get(label_name) {
+			Some(LabelledItem::Iteration(_)) => None,
+			Some(LabelledItem::Other(range)) => {
+				Some(p.err_builder("A `continue` statement can only jump to a label of an enclosing iteration statement.")
```
ematipico, line 509:
```suggestion
				Some(p.err_builder("A `continue` statement can only jump to a label of an enclosing iteration statement (`while`, `do/while`, `for`).")
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -469,24 +499,48 @@ fn parse_continue_statement(p: &mut Parser) -> ParsedSyntax {
...
+
         p.bump_any();
-        check_label_use(p, label_token);
 
-        label_token.end()
+        error
+    } else if !p.state.continue_allowed() {
+        Some(
+            p.err_builder(
+                "A `continue` statement can only be used within an enclosing iteration statement.",
```
ematipico, line 532:
```suggestion
                "A `continue` statement can only be used within an enclosing iteration statement (`while`, `do/while`, `for`).",
```

ematipico (approved):

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: I replaced `iteration` with `while`, `do while` and `for` statement

PR #1950, opened 2022-01-14
chore: add .git-blame-ignore-revs file

yassere:
## Summary

Adds a `.git-blame-ignore-revs` file with the hash of the recent "switch tabs to spaces" change so that it can be ignored by git blame.


.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# This file lists revisions of large-scale formatting/style changes so that
+# they can be excluded from git blame results.
+#
+# To set this file as the default ignore file for git blame, run:
+#   $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# chore: switch from tabs to spaces (https://github.com/rome/tools/pull/1941)
+409979ff132efdcb1e9fec3526d641966fd3f2bf
```
MichaReiser, line 8: Should there be a newline at the end of the file?

MichaReiser (approved):

.git-blame-ignore-revs
yassere: Sure, I'll add one.

PR #1951, opened 2022-01-15
feat(rslint_parser): Parser State Checkpoint

MichaReiser:
## Summary

#1929 revealed that rewinding a checkpoint isn't a guarantee that our parser is in its previous state, because it doesn't restore the state as well.

This isn't an issue for most of our state variables because most changes are scoped, automatically restored when exiting the parsing of a section.

However, the `hoisted` and `scoped` variables changes aren't scoped. A variable gets added when its first seen and the change is only undone when entering a new function or block (or class). Thus, re-parsing the same binding after a restore triggers a re-declaration error.

This PR is a preparation for #1929. It adds a `ParserStateCheckpoint` that ensures the `Parser` and its state are correctly rewinded when using checkpoints. The current implementation only adds debug assertions because the current state changes are all scoped, but it provides the infrastructure to rewind some state as well. The additional assertion will hopefully help to catch issues as noticed in #1929 sooner.

Creating a checkpoint must be cheap, that's why this PR changes the `ParserState` to use `IndexMap` that already is a dependency of the LSP crate. An `IndexMap` retains insertion order, thus, restoring to a previous state is as simple as calling `truncate(old_index)` on the map.

### DIFF
--- a/crates/rslint_parser/src/state.rs
+++ b/crates/rslint_parser/src/state.rs
@@ -375,11 +375,11 @@ bitflags! {
     /// Flags describing the context of a function.
     pub(crate) struct SignatureFlags: u8 {
         /// Is the function in an async context
-        const ASYNC 		= 0b00001;
+        const ASYNC 		= 1 << 0;
         /// Is the function in a generator context
-        const GENERATOR 	= 0b00010;
+        const GENERATOR 	= 1 << 1;
         /// Is the function a constructor (or constructor context)
-        const CONSTRUCTOR 	= 0b00100;
+        const CONSTRUCTOR 	= 1 << 2;
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/rslint_parser/src/state.rs
```diff
@@ -262,10 +323,6 @@ gen_change_parser_state!(
     AllowObjectExpression,
     ParsingContextFlags::ALLOW_OBJECT_EXPRESSION
 );
-gen_change_parser_state!(
```
MichaReiser, line 265: No point of setting it if no-one queries the flag

ematipico:
What's the impact, in terms of performance, of using `IndexMap` on our parser now?

MichaReiser:
> What's the impact, in terms of performance, of using `IndexMap` on our parser now?

There's been no significant change and whatever version is faster depends on the specific files. But that's kind of what I expect. Labels aren't frequently used. So switching won't have a significant performance impact. It may be more interesting to see how the performance changes when using it for hoisted and lexical variables. 

An alternative to `IndexMap` would be to use a `VersionedHashMap` that increments its internal `version` whenever the map is mutated. Each entry stores the version it was added. 

Such a map could be reverted by storing the `current` version in the snapshot and on restore compare the version If the versions are identical, then there's nothing to be done. If the version changed, then we must iterate through all entries and remove the entries `>` than the version that the parser snapshot. This would be fast in the majority of use cases but comes at the cost that we need to store the version for each entry (4-8 bytes). 

I don't have a strong preference but `IndexMap` seemed a good candidate, particularly because removing a label now is a simple `pop` call. However, `IndexMap` has the downside that lookups are slightly slower because of its internal data structure.

MichaReiser:
@leops and @xunilrj any thoughts?

crates/rslint_parser/src/state.rs
```diff
@@ -134,16 +134,100 @@ impl ParserState {
...
+/// that state because that's already taken care of by `with_state` and `with_scoped_state`.
+/// But, you can never no and better be safe than sorry. That's why we use some heuristics
+/// to verify that non of the scoped state did change and assert for it when rewinding.
+#[derive(Debug, Clone)]
+#[cfg(debug_assertions)]
+pub(super) struct DebugParserStateCheckpoint {
+    parsing_context: ParsingContextFlags,
+    label_set_len: usize,
+    strict: Option<StrictMode>,
+    default_item: Option<Range<usize>>,
```
leops, line 196: It seems to me that the checkpoints can only be cloned and not copied because of the `Range<usize>` in `StrictMode` and `default_item`, could these be replaced with `TextRange` to let `Checkpoint` implement `Copy` ?

crates/rslint_parser/src/state.rs
```diff
@@ -351,25 +431,24 @@ bitflags! {
         /// Whether `in` should be counted in a binary expression
         /// this is for `for...in` statements to prevent ambiguity.
         const INCLUDE_IN = 0b0000000000100000;
-        const IN_BINDING_LIST_FOR_SIGNATURE = 0b0000000001000000;
         /// Whether the parser is in a conditional expr (ternary expr)
-        const IN_CONDITION_EXPRESSION = 0b0000000010000000;
+        const IN_CONDITION_EXPRESSION = 0b0000000001000000;
```
leops, line 435: These constants could be written as `1 << 6` instead, it might help with readability

leops (approved):

crates/rslint_parser/src/state.rs
MichaReiser: let me add a comment that ideally, the checkpoint wouldn't even implement cloneable. It's unclear what the behaviour is if you rewind the same checkpoint twice (there isn't much harm to it but also, you shouldn't do it if avoidable). We may want to persist more information on the checkpoint in the future so that copying it becomes a considerable performance issue.

PR #1952, opened 2022-01-17
fix(rslint_parser): Fix private name with whitespace

MichaReiser:
## Summary
Trivia isn't allowed between a `#` token and the private name identifier. This PR changes the
*private member name* and *private name* implementation to emit an error if there's a whitespace between the `#` and identifier name.

## Test Plan

Added two new parser tests that handle whitespace between private member names

### DIFF
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -1017,8 +1017,11 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax {
         // 	# test;
         // }
         p.error(
-            p.err_builder("Unexpected space between # and identifier")
-                .primary(hash_end..p.cur_tok().start(), "remove the space here"),
+            p.err_builder("Unexpected space or comment between `#` and identifier")
+                .primary(
+                    hash_end..p.cur_tok().start(),
+                    "remove the space or comment here",
+                ),
         );
         Present(m.complete(p, JS_UNKNOWN))
     } else {
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -14,7 +14,7 @@ use crate::syntax::js_parse_error::expected_binding;
 use crate::syntax::object::{
     is_at_literal_member_name, parse_computed_member_name, parse_literal_member_name,
 };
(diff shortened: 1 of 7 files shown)


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -536,23 +535,135 @@ fn parse_class_member_impl(
...
-        //   ["a" + "b"]
-        //   static staticProperty
-        //   static staticInitializedProperty = 1
-        //   #private
-        //   #privateInitialized = "a"
-        //   static #staticPrivate
-        //   static #staticPrivateInitializedProperty = 1
-        // }
-        if is_at_property_class_member(p, 0) {
+    match member_name {
```
MichaReiser, line 538: This looks like a huge change but actually isn't. All I did is to move the property case to the end of the file because this should be the default when we parsed a member name, because a member name on its own is a member property. 

All the other cases are special cases, either a method, getter, etc...

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1006,9 +1007,24 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax {
...
+    if p.at(T![ident]) && hash_end != p.cur_tok().start() {
+        // test_err private_member_name_with_space
+        // class A {
+        // 	# test;
+        // }
+        p.error(
+            p.err_builder("Unexpected space between # and identifier")
+                .primary(hash_end..p.cur_tok().start(), "remove the space here"),
+        );
+        Present(m.complete(p, JS_UNKNOWN))
```
ematipico, line 1023: Are comments allowed instead?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -662,9 +662,23 @@ fn parse_private_name(p: &mut Parser) -> ParsedSyntax {
...
+
+    if p.at(T![ident]) && hash_end != p.cur_tok().start() {
+        // test_err private_name_with_space
+        // class A {
+        // 	# test;
+        // }
+        p.error(
+            p.err_builder("Unexpected space between # and identifier")
+                .primary(hash_end..p.cur_tok().start(), "remove the space here"),
+        );
```
ematipico, line 676: Repeated error/logic. Should we make it a bit more generic?

PR #1962, opened 2022-01-18
fix(rslint_parser): Allow `in` in sub-expressions

MichaReiser:
## Summary

There are multiple paths where the parser doesn't reset the `include_in` flag when parsing nested expressions. The result is that sub-expressions inside a `for` head incorrectly reject the `in` expression.

This PR moves the expression state out of the `ParserState` into its own `ExpressionContext`. This has the following benefits:

* Calling any `parse_expression*` method requires that engineers think about what should happen to the expression flags (inherit, change some, completely independent).
* Code is easier to follow as it no longer depends on the *global* parser state.


crates/rslint_parser/src/state.rs
```diff
@@ -464,26 +378,17 @@ bitflags! {
 pub(crate) struct ParsingContextFlagsSnapshot(ParsingContextFlags);
 
 /// Enters the parsing of function/method parameters
-pub(crate) struct EnterParameters {
+pub(crate) struct EnterParameters(
     /// Whether async and yield are reserved keywords
```
ematipico, line 382: Should we delete the comment too? Or move it?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -84,18 +85,83 @@ const STARTS_EXPR: TokenSet = token_set![
...
+        } else {
+            self - flag
+        }
+    }
+}
+
+impl Default for ExpressionContext {
+    fn default() -> Self {
+        ExpressionContext::INCLUDE_IN | ExpressionContext::ALLOW_OBJECT_EXPRESSION
+    }
```
ematipico, line 149: I would personally remove the `default` because the developer would not know what are the default flags. If the aim is to make the developer think about the context, I think we should explicitly set the properties of the context every time we create one.

I know that it's a lot of repeated code, but every time I see `ExpressionContext::default()`, I miss lot of information, unless I actually see the implementation.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -84,18 +85,83 @@ const STARTS_EXPR: TokenSet = token_set![
     JS_REGEX_LITERAL
 ];
 
+bitflags! {
+    pub(crate) struct ExpressionContext: u8 {
+        /// Whether `in` should be counted in a binary expression
+        /// this is for `for...in` statements to prevent ambiguity.
```
ematipico, line 91:
```suggestion
        /// Whether `in` should be counted in a binary expression.
        /// This is for `for...in` statements to prevent ambiguity.
```

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -84,18 +85,83 @@ const STARTS_EXPR: TokenSet = token_set![
...
+        const INCLUDE_IN = 0b00000001;
```
ematipico, line 92: Can't we use the `1 << 0` signature?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -84,18 +85,83 @@ const STARTS_EXPR: TokenSet = token_set![
...
+
+    pub(crate) fn is_in_condition_consequent(self) -> bool {
+        self.contains(ExpressionContext::IN_CONDITION_EXPRESSION)
+    }
+
+    pub(crate) fn is_in_included(self) -> bool {
+        self.contains(ExpressionContext::INCLUDE_IN)
+    }
+
+    pub(crate) fn with(self, flag: ExpressionContext, set: bool) -> Self {
```
ematipico, line 137: Maybe we can add a line of documentation.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -268,23 +333,29 @@ fn parse_yield_expression(p: &mut Parser) -> CompletedMarker {
...
             let m = marker.precede(p);
             p.bump_any();
 
-            p.with_state(InConditionExpression(true), parse_expr_or_assignment)
-                .or_add_diagnostic(p, js_parse_error::expected_expression_assignment);
+            parse_expr_or_assignment(
+                p,
+                context
+                    | ExpressionContext::IN_CONDITION_EXPRESSION
+                    | ExpressionContext::INCLUDE_IN,
```
ematipico, line 352: Can't we use the `with*` API here?

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -424,123 +425,124 @@ pub(super) fn parse_parameters_list(
...
+                p.bump_any();
+                parse_binding_pattern(
+                    p,
+                    ExpressionContext::default()
+                        .with_object_expression_allowed(!first || has_l_paren),
+                )
+                .or_add_diagnostic(p, expected_binding);
+
+                // TODO #1725 Review error handling and recovery
+                // rest patterns cannot be optional: `...foo?: number[]`
```
ematipico, line 456: Is the comment still valid? Considering that we closed the issue recently

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: The main idea is that there's a default for parsing an expression in a new context. For example, if you parse an expression statement, then parsing the expression must use the default values for parsing an expression.  T

An alternative is to add a `DEFAULT` const to `ExpressionContext`. We can think about alternative names but there aren't many that come to my mind except: `NEW`. any suggestions?

crates/rslint_parser/src/syntax/function.rs
ematipico: Sorry, I meant if it makes sense to have a comment that references an issue that we closed. Should we open a new one to track the TODO?

crates/rslint_parser/src/syntax/expr.rs
ematipico: So why can't we use `with_object_expression_allowed(true).with_include_in(true)`? My point is, I don't see the value of having some API and ended up not using them. We should always use the API to make the code consistent across the board.

crates/rslint_parser/src/syntax/function.rs
MichaReiser: I can link to the TS one.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: In my view, `Default` is a well-established pattern in Rust code saying that it takes the default value. Clearly, you as an author must know if this is what you need.

ematipico: Can we at least document what/why the default implement certain flags?

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: I made the flags private and wrapped them in an `ExpressionContext` so that all calls must go through the methods.

MichaReiser:
!bench

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -87,18 +88,99 @@ const STARTS_EXPR: TokenSet = token_set![
...
+        self.0
+            .contains(ExpressionContextFlags::IN_CONDITION_EXPRESSION)
+    }
+
+    /// Returns `true` if the expression parsing includes binary in expressions.
+    pub(crate) fn is_in_included(&self) -> bool {
+        self.0.contains(ExpressionContextFlags::INCLUDE_IN)
+    }
+
+    /// Adds the `flat` if `set` is `true`, otherwise removes the `flag`
```
ematipico, line 155:
```suggestion
    /// Adds the `flag` if `set` is `true`, otherwise removes the `flag`
```

ematipico (approved):
Looks good, just a minor typo that I found

PR #1969, opened 2022-01-20
fix(coverage): Fix [raw, module] flag combination of t262

MichaReiser:
## Summary

> `raw` The test source code must not be modified in any way, files from the harness/ directory must not be evaluated, and the test must be executed just once (in non-strict mode, only).

But there are a few tests with flags `[raw, module]`. It's important that we parse these as modules. The `raw` flag has no effect because modules are "strict" by default.


MichaReiser:
!bench

ematipico (approved):

PR #1971, opened 2022-01-20
refactor(xtask): Split up xtask into individual crates

Boshen:
## Background
* The `xtask` folder is bloated
* Having the `xtask` crate as a single entry for all tasks slows compilation speed significantly, since Rust caches dependencies in unit of crates. 
* After Splitting up, running each individual task will no longer compile the entire codebase, this makes codegen a lot easier to work with.

## Summary
This PR untangles all the tasks into individual crates:
```
members = [
	"crates/*",
	"xtask/",
	"xtask/bench",
	"xtask/codegen",
	"xtask/coverage",
]

```
The `xtask` crate is a common utility crate where all the other crates depend on.

We now run the tasks by `cargo run -p xtask_xxx`, we no longer need a `cargo xtask` alias.

TODO:
- [x] move js.ungram into codegen
- [x] add README guide to each crate, fix all mentions of `xtask`
- [x] fix CI

## Test Plan
Run the following commands:
```
cargo codegen # for help message
cargo codegen grammar
cargo codegen test
cargo codegen unicode

cargo coverage # for help message
cargo coverage js
cargo coverage js invalid-regexp --show-rast --show-diagnostics
cargo coverage js --json > base_results.json
cargo coverage compare ./base_results.json ./new_results.json --markdown

cargo bench
cargo run -p xtask_bench --features dhat-on --release
```


yassere:
>  @yassere you initially brought in xtask which proved extremely valuable.

Our `xtask` directory was copied from `rslint` when we forked the other crates. All I did was remark that I thought it was a good idea. 😁 

I believe the idea originally came from `rust-analyzer` and notably, they don't have their [codegen](https://github.com/rust-analyzer/rust-analyzer/blob/df5340386365b2b16c4e9bbae546504b97564c41/crates/syntax/src/tests/sourcegen_ast.rs) as part of xtask. It certainly makes sense to separate out our codegen and anything else that's particularly slow. I don't have much of an opinion on which things to extract, but I do think we want to retain a simple way to create Rust "scripts" with minimal dependencies.

Boshen:
`cargo xtask_codegen` and `cargo xtask` sounds good. 
I'll remove docgen for now, we can bring it back if we need it later.

Boshen:
I think I have refactored this into a pretty good middle ground:
```
[alias]
codegen = "run -p xtask_codegen --"
benchmark = "run -p xtask_bench --release --"
coverage = "run -p xtask_coverage --release --"
```

My reasoning: 
* benchmark and coverage has to be run in release mode, otherwise its just too slow (remember I was frustrated by its speed when I started out?)
* benchmark
  * contains criterion, which has the most impact on compilation (1 minute on my machine), so this needs to be isolated
  * we can add other benchmarks here
* coverage
  * we'll soon be adding TypeScript tests, so isolation is good here

Now all commands run fast without stepping on each others toes.
```
cargo codegen
Run codegen command.
USAGE:
	cargo codegen <SUBCOMMAND> [option]
SUBCOMMANDS:
	test         Transforms parser inline comments into test files
	ast          Transforms js.ungram into AST
	unicode      Generates unicode table inside lexer
```

```
cargo coverage
Run coverage command.
USAGE:
	cargo coverage <SUBCOMMAND> [option]
SUBCOMMANDS:
	js           Runs test262 tests
	compare      Compares output from
OPTIONS
	--markdown   Emits supported output into markdown format. Supported by compare subcommand
	--json       Emits supported output into json format. Supported by js subcommand
```

Adding more categories / subcommands should be intuitive now, as its simply adding a crate or file.

ematipico:
That's amazing!! Great change!!

The only suggestion I have, is to change the name of the command `ast` and its description. Maybe `grammar`? The reasons are:
- the code generation is not related only to the parser, soon we'll add code gen for formatter too
- the command will work on any grammar we'll introduce (HTML, CSS, etc)

xtask/src/ast.rs
```diff
@@ -1,352 +0,0 @@
-//! Definitions for the ECMAScript AST used for codegen
```
Boshen, line 1: We had a duplicated copy for some reason ...

Boshen:
All done, ready to merge.

See Test plan for all the commands.
CI is not working as its running from the `main` branch.

xtask/bench/src/libs.txt
```diff
@@ -1,10 +1,10 @@
...
-https://cdn.jsdelivr.net/npm/three@0.134.0/build/three.min.js
+https://cdn.jsdelivr.net/npm/jquery@3.2.1/dist/jquery.min.js
+https://cdn.jsdelivr.net/npm/vue@3.2.22/dist/vue.global.prod.js
+https://cdn.jsdelivr.net/npm/react@17.0.2/cjs/react.production.min.js
+https://cdn.jsdelivr.net/npm/react-dom@17.0.2/cjs/react-dom.production.min.js
+https://cdn.jsdelivr.net/npm/svelte@3.44.2/compiler.js
+https://cdn.jsdelivr.net/npm/dojo@1.16.4/dojo.js
+https://cdn.jsdelivr.net/npm/d3@7.1.1/dist/d3.min.js
+https://cdn.jsdelivr.net/npm/pixi.js@6.2.0/dist/browser/pixi.min.js
+https://cdn.jsdelivr.net/npm/three@0.134.0/build/three.min.js
```
Boshen, line 9: Some white space newline change from git

xtask/src/lib.rs
```diff
@@ -1,40 +1,28 @@
...
+    Verify,
+}
 
 pub fn project_root() -> PathBuf {
     Path::new(
         &env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| env!("CARGO_MANIFEST_DIR").to_owned()),
     )
     .ancestors()
-    .nth(1)
+    .nth(2)
```
Boshen, line 25: the crates are one level deeper, so we need to navigate up twice.

MichaReiser:
I'm not sure it's ideal to run 'coverage' in release mode by default. I often run coverage on a single test to verify my changes. Having to wait for a release build would slow me down significantly. Also, my impression is that about half of the runtime is spent loading the tests from the file system and not running the tests. 

Maybe all that's needed is to split out the codegen part into it's own crate/command and leave the other as they are?

Boshen:
> I'm not sure it's ideal to run 'coverage' in release mode by default. I often run coverage on a single test to verify my changes. Having to wait for a release build would slow me down significantly. 

I think this is just the problem of using `cargo coverage` alias vs running `cargo run -p xtask_coverage`. Giving the alias `--release` helps out first time comers. Maybe it's best to document both commands. 

> Maybe all that's needed is to split out the codegen part into it's own crate/command and leave the other as they are?

I have been using this branch for development in the last two days, and there are no noticeable delays anymore when running `codegen` and `coverage`. This is purely due to not having to pull in `criterion`, which impacts build time by a major factor. 

I'm on a really old laptop, so the amount of time I have to wait for compilation is probably a lot longer.

The only downside of using Rust is that we have to purposely engineer its build time 😞.

ematipico (approved):
It looks good to me, and I would like to merge it ASAP. One thing though. Would you mind adding a paragraph under `Checks > cargo test`, and add small description for `cargo coverage`? It's nowhere mentioned and we missed it

Boshen:
Updated CONTRIBUTING.md with

#### `cargo coverage`

This command will check and report parser coverage against different test suites.
We currently target the [Official ECMAScript Conformance Test Suite](https://github.com/tc39/test262).

ematipico:
@MichaReiser @xunilrj would you mind to double check and give your 👍 ?

Boshen:
I have rebased the main branch and added the `cargo coverage ts` command.

xunilrj (approved):

MichaReiser:
Looks good. The coverage command is failing on CI. Would you mind taking a look?

Boshen:
> Looks good. The coverage command is failing on CI. Would you mind taking a look?

The CI is running code from the main branch 😄

ematipico:
> Looks good. The coverage command is failing on CI. Would you mind taking a look?

The error is expected. As @Boshen , the code on `main` is outdated. It happened the same to me the  first time I had to implement the comparison feature.

Boshen:
Rebased, ready to merge again.

MichaReiser:
> Rebased, ready to merge again.

The coverage CI job is failing which prevents me from merging the PR. Would you mind taking a look?

PR #1974, opened 2022-01-21
feat(rome_formatter): retain newline trivias in the formatter output

leops:
## Summary

This change was split off from #1916 and lets user-inserted "empty lines" (sequence of 2 or more line breaks) be retained in the formatter output. Unlike comments which can be attached to any token, empty lines are only kept as separators between elements in the following nodes:
- Directive lists
- Statement lists
- Switch statements
- Class declarations
- Array expressions
- Object expressions


crates/rome_formatter/src/format_element.rs
```diff
@@ -254,6 +278,88 @@ where
...
 }
 
+/// Specialized version of [join_elements] for joining SyntaxNodes separated by a space, soft
+/// line break or empty line depending on the input file.
+///
+/// This functions inspects the input source and separates consecutive elements with either
+/// a [soft_line_break_or_space] or [empty_line] depending on how many line breaks were
+/// separating the elements in the original file.
+#[inline]
+pub fn join_elements_soft_line<I, N>(elements: I) -> FormatElement
```
ematipico, line 288: I start to understand why having two methods with almost the same code makes sense in our API, although both methods (`join_elements_soft_line` and `join_elements_soft_line`), but maybe we can DRY this code into an internal function and just pass `soft_line_break_or_space` or `hard_line_break`?

crates/rome_formatter/src/format_element.rs
```diff
@@ -254,6 +278,88 @@ where
...
+                empty_line()
+            } else {
+                hard_line_break()
+            }
+        },
+    ))
+}
+
+/// Get the number of line breaks between two consecutive SyntaxNodes in the tree
+fn get_lines_between(prev_node: &SyntaxNode, next_node: &SyntaxNode) -> usize {
```
ematipico, line 330: Maybe `get_lines_between_nodes` is a better fit? As it is now, it makes me think that this function is more generic, not just for nodes

ematipico, line 330: What would happen if, by mistake, as `prev_node` we actually pass the next node? What would be the end result? Should we also check their width?

crates/rome_formatter/src/intersperse.rs
```diff
@@ -78,3 +80,73 @@ where
...
+            Some((node, elem)) => {
+                self.prev_item = Some(node);
+                Some(elem)
+            }
+            None => None,
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (lo, hi) = self.iter.size_hint();
```
ematipico, line 136: What `lo` stands for here? "lower"?

crates/rome_formatter/tests/specs/js/arrow/params.js.snap
```diff
@@ -331,32 +332,114 @@ foo(
...
 foo(({ a: { b: { c, d } } }) => a);
+
 foo(({ a: { b: { c: { d, e } } } }) => a);
+
 foo(([{ a: { b: { c: { d, e } } } }]) => {});
+
 foo(([...{ a: { b: { c: { d, e } } } }]) => {});
-foo((n = { a: { b: { c: { d, e } } } }) => {});
+
+foo(
```
ematipico, line 366: This is much better now! Our formatter now looks more a formatter and less a minimizer 😝

crates/rome_formatter/tests/specs/js/array/empty_lines.js.snap
```diff
@@ -19,5 +20,12 @@ let a = [
...
 # Output
-let a = [1, 2, 3, 4];
+let a = [
+	1,
+	2,
+
+	3,
+
+	4,
+];
```
ematipico, line 30: This improvement is amazing, even better than Prettier! What prettier spits out is the following:

```js
let a = [
  1, 2,

  3,

  4,
];
```

Which is less consistent

ematipico (approved):
Just some comments and questions!

crates/rome_formatter/src/format_element.rs
```diff
@@ -254,6 +278,88 @@ where
...
+    // Ensure the two nodes are actually siblings on debug
+    debug_assert_eq!(prev_node.next_sibling().as_ref(), Some(next_node));
+    debug_assert_eq!(next_node.prev_sibling().as_ref(), Some(prev_node));
```
MichaReiser, line 333: Should we call the method `get_lines_between_next_node(node)`. That would remove the need for the assertion.

crates/rome_formatter/src/format_element.rs
```diff
@@ -254,6 +278,88 @@ where
...
+
+    // Count the lines separating the two statements,
+    // starting with the trailing trivia of the previous node
+    let mut line_count = prev_node
+        .last_trailing_trivia()
+        .and_then(|prev_token| {
+            // Newline pieces can only come last in trailing trivias, skip to it directly
+            let last_trivia = prev_token.pieces().last()?;
+            let prev_newline = last_trivia.as_newline()?;
+            Some(prev_newline.text().matches('\n').count())
```
MichaReiser, line 343: I'm not sure if it's safe to match on the newline text. What about windows line endings? Why do we need to match on the `\n`.

crates/rome_formatter/src/format_element.rs
```diff
@@ -254,6 +278,88 @@ where
...
+        })
+        .unwrap_or(0);
+
+    // Then add the newlines in the leading trivia of the next statement
```
MichaReiser, line 347: Nit: next node

crates/rome_formatter/src/formatter.rs
```diff
@@ -283,23 +283,22 @@ impl Formatter {
...
                     }
                 };
 
                 elements.push(format_elements![comment, line_break]);
 
                 line_count = 0;
-            } else if piece.as_newline().is_some() {
-                line_count += 1;
+            } else if let Some(newline) = piece.as_newline() {
+                line_count += newline.text().matches('\n').count();
```
MichaReiser, line 294: @xunilrj is there a good way to get the new lines count? Aren't we inserting a `Newline` piece for every new line? In which case counting the new line pieces would be sufficient?

crates/rome_formatter/src/intersperse.rs
```diff
@@ -78,3 +80,73 @@ where
...
     }
 }
+
+/// An iterator adapter that places a separator between all elements.
+/// The separator element is generated by a factory function called
+/// by the iterator with the previous and next SyntaxNode as arguments
+#[derive(Debug, Clone)]
+pub(crate) struct IntersperseFn<I: Iterator, F, N>
+where
+    I: Iterator<Item = (N, FormatElement)>,
```
MichaReiser, line 90: Nit: An alternative would be to make `FormatElement` implement `Clone`

crates/rome_formatter/src/format_element.rs
leops: These ended up with a very similar implementation so I could merge them, I'm not entirely sure how to name the resulting function though: `join_elements_lines` ?

crates/rome_formatter/src/format_element.rs
leops: Passing the wrong nodes would result in an assertion failure on debug so that would be caught by the tests, however this function is module-private on purpose and if I merge the two join_elements_line functions I'd probably move it in there because it's highly specific for this use case and shouldn't be used for anything else.

crates/rome_formatter/src/intersperse.rs
leops: Yes, `size_hint` returns a a lower and higher bound on the size of the iterator (this specific bit of code is copied directly from the implementation of `size_hint` for the `Intersperse` iterator)

crates/rome_formatter/src/format_element.rs
leops: I used matches specifically to handle CRLF newlines instead of just reading the newlines trivia length: in files with LF line endings the text would be a sequence of `\n` so that would yield the correct result but on Windows the text contains a sequence of `\r\n` so counting the occurrences of the `\n` character is correct for all platforms

xunilrj: Not at all. I can do this today.

crates/rome_formatter/src/formatter.rs
leops: I don't know if its the intended behavior but at least in my tests multiple consecutive newlines result a single newline trivia piece covering the entire sequence of line breaks

crates/rome_formatter/src/intersperse.rs
leops: I think `FormatElement` is already clonable

crates/rome_formatter/src/format_element.rs
leops: Oh right yes I forgot about the exotic newline characters. But yes anyway it would be better handled directly in the lexer, then it would also support the line separator and paragraph separator characters correctly

crates/rome_formatter/src/intersperse.rs
MichaReiser: Yeah sorry. The reason you need `IntersperseFn` is that you compute the separator based on the two consecutive nodes.

crates/rome_formatter/src/format_element.rs
ematipico: I think have two different APIs signatures makes more sense (less writing and better documentation), although both these functions can call the same internal function
```rs
pub fn join_elements_soft_line() {
	self.join_elements_with(soft_line_break_or_space())
}

pub fn join_elements_hard_line() {
	self.join_elements_with(hard_line_break())
}

fn join_elements_with(&self, FormatElement) {
	// implementation code here
}
```

crates/rome_formatter/src/format_element.rs
xunilrj: PR created: https://github.com/rome/tools/pull/1991

MichaReiser (approved):

PR #1977, opened 2022-01-21
fix(rslint_parser): Left-recursive sequence expression

MichaReiser:
## Summary

Fixes the sequence expression parsing to be left recursive. I took this as a chance to rewrite `parse_paren_or_arrow_expr`. It now first tries to parse the arguments as sequence expression but bails out early if it encounters any parameter specific syntax.

The performance should be about the same because the old implementation also rewound to re-parse the arrow function parameters. It just did so spread across multiple places.

## Test Plan

Added new parser tests


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -926,22 +922,47 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
+            parse_spread_element(p, context.and_include_in(true))
         } else {
             parse_expr_or_assignment(p, context.and_include_in(true))
-                .or_add_diagnostic(p, js_parse_error::expected_expression_assignment);
+        };
+
+        if argument.is_absent() && p.at(T![,]) {
+            argument.or_add_diagnostic(p, js_parse_error::expected_expression);
+            // missing element
+            continue;
```
MichaReiser, line 951: Adds proper recovery if a call argument is missing.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -961,177 +982,144 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
+    // (a, b, c;
+
+    // First assume that the syntax is a parenthesized expression, potentially with an inner sequence expression
+    // Set this flag to true if any of the expression turns out to be a parameter (e.g., has a type annotation)
+    let mut is_parameters = false;
 
     if !is_empty {
         let context = context
-            .and_potential_arrow_start(true)
+            .and_potential_arrow_start(false)
```
MichaReiser, line 1051: Unclear why this should be true. This method is handling a potential arrow start...

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1211,7 +1215,10 @@ fn parse_primary_expression(p: &mut Parser, context: ExpressionContext) -> Parse
             } else {
                 // `async a => {}` and `async (a) => {}`
                 if context.is_potential_arrow_start()
-                    && ((is_nth_at_name(p, 1) && p.nth_at(2, T![=>])) || p.nth(1) == T!['('])
+                    && ((is_nth_at_name(p, 1)
+                        && p.nth_at(2, T![=>])
+                        && !p.has_linebreak_before_n(2))
```
MichaReiser, line 1220: Linebreaks are not allowed before `=>`

MichaReiser:
!bench

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -926,22 +922,47 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
         }
 
-        if p.at(T![,]) {
-            p.bump_any();
-        } else {
+        if argument
+            .or_recover(
+                p,
+                &ParseRecovery::new(
+                    JS_UNKNOWN_EXPRESSION,
```
ematipico, line 958: If `argument` is result of `parse_spread_element`, shouldn't this become a `JS_UNKNOWN_PARAMETER`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -926,22 +922,47 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
     let m = p.start();
     p.bump(T!['(']);
     let args_list = p.start();
+    let mut first = true;
     let mut progress = ParserProgress::default();
 
     while !p.at(EOF) && !p.at(T![')']) {
+        if first {
```
ematipico, line 929: Time ago, when we refactored the parsing of lists, I deliberately skipped arguments because were a bit tricky. I wonder if we are now in position to use `ParseSeparatedList` or not ?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -926,22 +922,47 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
             // already do a check on "..." so it's safe to unwrap
-            parse_spread_element(p, context.and_include_in(true)).unwrap();
+            parse_spread_element(p, context.and_include_in(true))
         } else {
             parse_expr_or_assignment(p, context.and_include_in(true))
-                .or_add_diagnostic(p, js_parse_error::expected_expression_assignment);
+        };
+
+        if argument.is_absent() && p.at(T![,]) {
+            argument.or_add_diagnostic(p, js_parse_error::expected_expression);
```
ematipico, line 949: expression or parameter? How do we know which diagnostic to fire?

MichaReiser:
> !bench

Hmm, what happened to bench :(

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -961,177 +982,140 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
+    // (a, b, c, d, e, f);
+    // (a, b, c, d, e, f)
+    // (a, b, c)
 
-    let is_empty = p.eat(T![')']);
+    // test_err incomplete_parenthesized_sequence_expression
+    // (a,;
+    // (a, b, c;
+
+    // First assume that the syntax is a parenthesized expression, potentially with an inner sequence expression
```
ematipico, line 1041:
```suggestion
    // First assume that the syntax is a parenthesized expression, potentially with an inner sequence expression.
```

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -961,177 +982,140 @@ fn parse_arguments(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
             }
         }
+    }
 
-        if let Some(sequence) = sequence.take() {
-            sequence.complete(p, JS_SEQUENCE_EXPRESSION);
-        }
+    // test_err empty_parenthesized_expression
+    // ();
+    if is_empty {
```
ematipico, line 1084: Wouldn't make more sense to have this as the `else` branch of `is !is_empty`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1142,26 +1126,42 @@ pub fn parse_expression_snipped(p: &mut Parser) -> ParsedSyntax {
 }
 
 /// A general expression.
+pub(crate) fn parse_expression(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
```
ematipico, line 1129: Wouldn't this function spit out an assignment too, due to `parse_expr_or_assignment`?

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1562,7 +1565,10 @@ impl ParseSeparatedList for ArrayElementsList {
     fn recover(&mut self, p: &mut Parser, parsed_element: ParsedSyntax) -> RecoveryResult {
         parsed_element.or_recover(
             p,
-            &ParseRecovery::new(JS_UNKNOWN_EXPRESSION, EXPR_RECOVERY_SET),
+            &ParseRecovery::new(
+                JS_UNKNOWN_EXPRESSION,
+                EXPR_RECOVERY_SET.union(token_set!(T![']'])),
```
ematipico, line 1570: What's the use case of recovering also at `]`?

crates/rslint_parser/test_data/inline/err/empty_parenthesized_expression.rast
```diff
@@ -0,0 +1,36 @@
...
+  2: JS_MODULE_ITEM_LIST@0..3
+    0: JS_EXPRESSION_STATEMENT@0..3
+      0: JS_PARENTHESIZED_EXPRESSION@0..2
+        0: L_PAREN@0..1 "(" [] []
+        1: (empty)
+        2: R_PAREN@1..2 ")" [] []
+      1: SEMICOLON@2..3 ";" [] []
+  3: EOF@3..4 "" [Newline("\n")] []
+--
+error[SyntaxError]: expected an expression but instead found ')'
```
ematipico, line 29: I think we can safely have an error message that a parenthesized  expression can't be empty

ematipico:
> > !bench
> 
> Hmm, what happened to bench :(

My guess is that these kind of jobs stops when a new commit is pushed.

xtask/js.ungram
```diff
@@ -1428,7 +1428,7 @@ TsNumber = ident: Ident
 TsObject = ident: Ident
 TsBoolean = ident: Ident
 TsBigint = ident: Ident
-TsString = ident: Ident
+TsString = 'ident'
```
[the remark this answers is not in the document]
MichaReiser: That's what we actually parse as part of TS. Printing the ast for `a: string` otherwise panicked.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: Maybe. I didn't want to add more changes to my PR

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: We don't want to run over the closing `]` delimiter. Let's say we have `[<]`. Now, the recovery would eat the `<` and the `]`... and then complain that it expected a `]` :D

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: `parse_spread` should never return `Absent`. And if it does, then the argument (which must be an expression) is some gibberish expression. Thus, a `UNKNOWN_EXPRESSION`. It isn't a `JS_UNKNOWN_PARAMETER` because this code parses arguments and not parameters.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: We're parsing `arguments` and not parameters. Arguments must be expressions.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: We should rename this to `parse_assignment_or_higher`. It's not an or, but it parses an assignment expression or any expression that has higher precedence than an assignment.

MichaReiser:
!bench

Let's try this again

ematipico (approved):

Boshen:
!bench plz

MichaReiser:
bench!

MichaReiser:
manual results

 ### Bench results on ubuntu-latest
```
parser/compiler.js                    1.02     96.4±1.52ms    10.9 MB/sec    1.00     94.8±1.57ms    11.0 MB/sec
parser/d3.min.js                      1.02     61.1±0.84ms     4.3 MB/sec    1.00     60.2±1.06ms     4.4 MB/sec
parser/dojo.js                        1.02      5.3±0.02ms    13.1 MB/sec    1.00      5.2±0.04ms    13.3 MB/sec
parser/jquery.min.js                  1.00     15.7±0.11ms     5.3 MB/sec    1.17     18.3±0.11ms     4.5 MB/sec
parser/pixi.min.js                    1.01     73.8±1.11ms     5.9 MB/sec    1.00     73.1±1.12ms     6.0 MB/sec
parser/react-dom.production.min.js    1.00     21.4±0.18ms     5.4 MB/sec    1.04     22.2±0.79ms     5.2 MB/sec
parser/react.production.min.js        1.00   1171.4±7.46µs     5.3 MB/sec    1.05  1227.3±34.29µs     5.0 MB/sec
parser/tex-chtml-full.js              1.00    164.4±1.33ms     5.5 MB/sec    1.04    171.0±6.56ms     5.3 MB/sec
parser/three.min.js                   1.00     84.1±1.13ms     7.0 MB/sec    1.00     84.1±0.79ms     7.0 MB/sec
parser/vue.global.prod.js             1.01     27.7±1.02ms     4.3 MB/sec    1.00     27.5±0.85ms     4.4 MB/sec
```

PR #1979, opened 2022-01-21
fix(rslint_lexer): Fix lexing of regex literals

MichaReiser:
## Summary

Fixes the remaining 262 tests that incorrectly fail parsing a valid program containing a regex literal.
The lexer uses the `in_expr` state to determine whatever a regexp is valid. This requires, that the lexer correctly tracks in which places an expression is valid. This hasn't been the case for

```
{

}
function // <- smt
```

and some keywords that I added to `before_expr` list.

## Test Plan

Added additional tests to the lexer

### DIFF
--- a/crates/rslint_lexer/src/state.rs
+++ b/crates/rslint_lexer/src/state.rs
@@ -59,8 +59,9 @@ impl LexerState {
             }
 
             T![function] => {
+                // Needed to lex function fn() {}/1/;
                 if self.expr_allowed
-                    && ctx_is_brace_block(
+                    && !ctx_is_brace_block(
                         &self.ctx,
                         self.prev,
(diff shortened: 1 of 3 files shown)


crates/rslint_lexer/src/state.rs
```diff
@@ -59,8 +59,9 @@ impl LexerState {
             }
 
             T![function] => {
+                // Needed to lex function fn() {}/1/;
                 if self.expr_allowed
-                    && ctx_is_brace_block(
+                    && !ctx_is_brace_block(
```
MichaReiser, line 64: A function following a block is a statement is a function statement and not a function expression.

ematipico (approved):

PR #1981, opened 2022-01-21
feat(rslint_parser): Ban template literal after optional chain

Boshen:
## Summary
Template string is not allowed to be passed to tail position of optional chain.

### Test262 comparison coverage results on ubuntu-latest
| Total | 44951 | 44951 | 0 |
| Passed | 43841 | 43845 | ✅ ⏫ **+4** |
| Failed | 1109 | 1105 | ✅ ⏬ **-4** |
| Panics | 1 | 1 | 0 |
| Coverage | 97.53% | 97.54% | **+0.01%** |

<details><summary><b>:tada: Fixed (4):</b></summary>

```
xtask/src/coverage/test262/test/language/expressions/optional-chaining/early-errors-tail-position-null-optchain-template-string-esi.js
xtask/src/coverage/test262/test/language/expressions/optional-chaining/early-errors-tail-position-null-optchain-template-string.js
xtask/src/coverage/test262/test/language/expressions/optional-chaining/early-errors-tail-position-optchain-template-string-esi.js
xtask/src/coverage/test262/test/language/expressions/optional-chaining/early-errors-tail-position-optchain-template-string.js
```
</details>

### DIFF
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -724,15 +724,17 @@ fn subscripts(
     context: ExpressionContext,
 ) -> CompletedMarker {
     // test_err subscripts_err
-    // foo()?.baz[].
+    // foo()?.baz[].;
     // BAR`b
     let mut should_try_parsing_ts = true;
     let mut progress = ParserProgress::default();
+    let mut is_optional_chain = false;
     while !p.at(EOF) {
         progress.assert_progressing(p);
 
         match p.cur() {
             T![?.] if p.nth_at(1, T!['(']) => {
+                is_optional_chain = true;
                 lhs = {
                     let m = lhs.precede(p);
                     p.bump_any();
@@ -752,10 +754,14 @@ fn subscripts(
                 }
             }
             T![?.] if p.nth_at(1, T!['[']) => {
+                is_optional_chain = true;
                 lhs = parse_computed_member_expression(p, lhs, true, context).unwrap()
             }
             T!['['] => lhs = parse_computed_member_expression(p, lhs, false, context).unwrap(),
-            T![?.] => lhs = parse_static_member_expression(p, lhs, T![?.]).unwrap(),
+            T![?.] => {
+                is_optional_chain = true;
+                lhs = parse_static_member_expression(p, lhs, T![?.]).unwrap()
+            }
             T![.] => lhs = parse_static_member_expression(p, lhs, T![.]).unwrap(),
             T![!] if !p.has_linebreak_before_n(0) => {
                 lhs = {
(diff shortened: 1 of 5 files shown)


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -724,15 +724,17 @@ fn subscripts(
     context: ExpressionContext,
 ) -> CompletedMarker {
     // test_err subscripts_err
-    // foo()?.baz[].
+    // foo()?.baz[].;
```
Boshen, line 727:
```
// test_err subscripts_err
// foo()?.baz[].
// BAR`b
```
This wrapped, and triggered the wrong error :-(

crates/rslint_parser/test_data/inline/err/template_after_optional_chain.rast
```diff
@@ -0,0 +1,204 @@
...
+  │
+2 │ obj.val?.[expr]`template`
+  │ ^^^^^^^^^^^^^^^^^^^^^^^^^
+
+--
+error[SyntaxError]: Invalid tagged template on optional chain
+  ┌─ template_after_optional_chain.js:3:1
+  │
+3 │ obj.func?.(args)`template`
+  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ematipico, line 199: Are we able to apply the range of the single template literal?

Boshen: This whole expression is called "Tagged Template", so using the whole range or the template range are both good. But the error message mentions "optional chain" so the whole range is better?

We can apply to the single template literal, but then we have to pass an extra argument to `parse_template_literal` 🤔

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -797,7 +803,19 @@ fn subscripts(
...
-            BACKTICK => lhs = parse_template_literal(p, Present(lhs)),
+            BACKTICK => {
+                lhs = parse_template_literal(p, Present(lhs));
+                // test_err template_after_optional_chain
+                // obj.val?.prop`template`
+                // obj.val?.[expr]`template`
+                // obj.func?.(args)`template`
+                if is_optional_chain {
+                    p.error(
+                        p.err_builder("Invalid tagged template on optional chain")
```
MichaReiser, line 814: Can we use something closer to TS?

> Tagged template expressions are not permitted in an optional chain.

Where it becomes clear of whats

MichaReiser (approved):

crates/rslint_parser/src/syntax/expr.rs
MichaReiser, line 814: Can we change the type of the whole expression to an `UNKNOWN_EXPRESSION`?

crates/rslint_parser/test_data/inline/err/template_after_optional_chain.rast
MichaReiser: Nit: An option would be to capture the `lhs` range before calling into `parse_template_literal` and then add a secondary range to the diagnostic, pointing to the optional chain expression. But I'm not sure if it adds a lot of clarity or makes things even more confusing.

Boshen:
I have updated the error message and changed it to unknown expression.

> Nit: An option would be to capture the lhs range before calling into parse_template_literal and then add a secondary range to the diagnostic, pointing to the optional chain expression. But I'm not sure if it adds a lot of clarity or makes things even more confusing.

Looks like it'll get more confusing :-(

ematipico (approved):

PR #1982, opened 2022-01-21
Feature/coverage for ts

xunilrj:
I had to enable ```git config --system core.longpaths true``` to be able to clone ```Typescript``` repo inside ours. Should we put this somewhere?

Now ```coverage``` accepts "ts", like:

```
cargo run -p xtask --release -- coverage --language=ts   
```

If not specified we fallback to

```
cargo run -p xtask --release -- coverage --language=js
```

Our current results are. We just check if the parser has errors or not. Can be misleading...

```
│   17149   │  7242  │  9481  │  426   │  42.23   │
```


MichaReiser:
Nice work. Should we run all languages if non is specified? That gives us the CI integration for free

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -449,7 +454,13 @@ fn parse_binary_or_logical_expression_recursive(
             m.complete(p, TS_ASSERTION)
         };
         res.err_if_not_ts(p, "type assertions can only be used in TypeScript files");
-        return parse_binary_or_logical_expression_recursive(p, Present(res), min_prec, context);
+        return parse_binary_or_logical_expression_recursive(
```
MichaReiser, line 457: Hmm. Can ee add a todo and track it in the Typescript issue? 

Have you tried larger values? Are all libs still parsing?

Do you know how the typescript parser handles this. Shouldn't we be able to rewrite this into a loop, similar to the sequence expressions parsing (worth a separate PR)

xtask/src/coverage/typescript.rs
```diff
@@ -0,0 +1,156 @@
...
+                let path = file.path().to_string_lossy().replace("\\", "/");
+                path.contains(query)
+            } else {
+                true
+            }
+        })
+        .collect::<Vec<_>>()
+}
+
+pub fn check_file_encoding(path: &std::path::Path) -> Option<String> {
```
MichaReiser, line 38: Nit: This function does more than just checking the encoding. It tests if it's UTF8 and reads the file and otherwise bails out

`read_utf8_file_to_string`?

xtask/src/coverage/typescript.rs
```diff
@@ -0,0 +1,156 @@
...
+    let buffer = std::fs::read(path).unwrap();
+    let bom = buffer.get(0..3);
+    //Utf16Be or // Utf16Le
+    if let Some(&[0xfe, 0xff, _] | &[0xff, 0xfe, _]) = bom {
+        None
+    } else {
```
MichaReiser, line 44: Are there UTF16 tests. What about these?

xtask/src/coverage/typescript.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    pool.scoped(|_| {
+        let pb = &pb;
+
+        let is_detailed = files.len() < 10;
+
+        for file in files {
+            let path = file.path();
+            let code = check_file_encoding(path);
+            if code.is_none() {
```
MichaReiser, line 77: There seems to be a lot of duplicated code. Would it be possible to extract a trait that:

```
trait CoverageTests {
  load_tests(query) -> Option<VecDir>,
  run_test(test) -> TestResult
}
```

xunilrj:
> Nice work. Should we run all languages if non is specified? That gives us the CI integration for free

I was afraid that this could break the comparison somehow.

xunilrj:
> Hmm. Can ee add a todo and track it in the Typescript issue?

Done.

> Have you tried larger values? Are all libs still parsing?

Yes. Makes no difference.

> Do you know how the typescript parser handles this. 

They do it recursively at https://github.com/microsoft/TypeScript/blob/main/src/compiler/parser.ts#L4639. One thing that may be affecting us is https://github.com/rome/tools/blob/main/xtask/src/main.rs#L61.

I did a little bit more tests and dug some strange things.

1 - There is no stackoverflow on WSL;
2 - Which means it may be a limitation on Windows implementation;
3 - Searching online I found that maybe the problem is at https://github.com/rust-lang/rust/blob/master/library/std/src/sys/windows/thread.rs#L39
4 - Someone complains something similar on .NET: ```https://github.com/dotnet/runtime/issues/45971```. Maybe that is what is happening?

I did a quick test here, and Windows and WSL are booming the stack at different depths...

```
fn main() {
    let t = std::thread::Builder::new();
    let handler = t.spawn(|| {
        fn recurse(count: usize) {
            println!("{}", count);
            recurse(count + 1);
        }
        recurse(0);
    }).unwrap();
    handler.join().unwrap();
    println!("Hello, world!");
}
```

I got:

Windows: 12849
WSL: 16336

If I force the stack size

```
    let t = std::thread::Builder::new().stack_size(1024 * 1024);
```

I get:
WSL: 8144
Windows: 6301

So maybe out recursion starts deeper, and we have less recursions available...

> Shouldn't we be able to rewrite this into a loop, similar to the sequence expressions parsing (worth a separate PR)

Yes. Will be much more efficient and safer.

xunilrj:
> Are there UTF16 tests. What about these?

Currently we are ignoring them.
And given that we return ```&str``` in multiple places, I going to say that we don't support non-UTF8 code today.

MichaReiser (approved):

PR #1983, opened 2022-01-22
feat(rslint_parser): Ban `super` keyword in top-level code

Boshen:
## Summary
The `super` keyword cannot appear in top level.

### Test262 comparison coverage results on ubuntu-latest
| Total | 44951 | 44951 | 0 |
| Passed | 43841 | 43842 | ✅ ⏫ **+1** |
| Failed | 1109 | 1108 | ✅ ⏬ **-1** |
| Panics | 1 | 1 | 0 |
| Coverage | 97.53% | 97.53% | **+0.00%** |

<details><summary><b>:tada: Fixed (1):</b></summary>

```
xtask/src/coverage/test262/test/language/global-code/super-prop.js
```
</details>

### DIFF
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -704,13 +704,22 @@ fn parse_member_or_new_expr(
 //   }
 // }
 // super();
+// super.property;
 fn parse_super_expression(p: &mut Parser) -> ParsedSyntax {
     if !p.at(T![super]) {
         return Absent;
     }
     let super_marker = p.start();
     p.expect(T![super]);
-    Present(super_marker.complete(p, JS_SUPER_EXPRESSION))
+    let mut completed = super_marker.complete(p, JS_SUPER_EXPRESSION);
+    if p.state.is_top_level() {
+        p.error(
+            p.err_builder("`super` is only valid inside of a class constructor of a subclass.")
+                .primary(completed.range(p), ""),
+        );
+        completed.change_kind(p, JS_UNKNOWN_EXPRESSION);
+    }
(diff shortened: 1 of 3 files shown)


crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -704,13 +704,22 @@ fn parse_member_or_new_expr(
...
     if !p.at(T![super]) {
         return Absent;
     }
     let super_marker = p.start();
     p.expect(T![super]);
-    Present(super_marker.complete(p, JS_SUPER_EXPRESSION))
+    let mut completed = super_marker.complete(p, JS_SUPER_EXPRESSION);
+    if p.state.is_top_level() {
+        p.error(
+            p.err_builder("`super` is only valid inside of a class constructor of a subclass.")
```
MichaReiser, line 717: The parser already handles calls to the super constructor:

https://github.com/rome/tools/blob/1957619118b689c8d5819eb046157678a8988f77/crates/rslint_parser/src/syntax/expr.rs#L1640-L1650

The message doesn't fit the context. `super.prop` or `super.expr()` is valid anywhere inside of a class method and isn't restricted to class constructors.

There isn't enough information in the `ParserState` at the time being to infer if the parser is inside of a class method or any other function. I think we would need a `in_class` flag in the parser state that can then be used in combination with `in_function`.

For example:

```
// valid
class A {
  method() { 
    super.method()
  }
}

// invalid
function test() {
  super.test();
}
```

MichaReiser (changes requested):

Boshen:
I prefer wait on decision around second pass, all these semantic errors are all nice to haves right now.

PR #1985, opened 2022-01-24
fix(rslint_lexer): Invalid unicode escape sequence in tagged template literal

MichaReiser:
## Summary

Invalid unicode escape sequences are valid inside of tagged template literals, but not template literals. 

This PR suppresses the escape sequence diagnostics for tagged template literals. 

It further fixes an issue with escaped sequences where `validate_escape_sequence` progressed the current position and so did `read_str_literal`. 

It also fixes the span of a diagnostic to be a valid range (and not from pos to 1).

## Test Plan

Added new lexer tests. The `invalid_escape_sequence.js` test is now passing.

### DIFF
--- a/crates/rslint_errors/src/file.rs
+++ b/crates/rslint_errors/src/file.rs
@@ -82,7 +82,17 @@ where
     T: Into<usize>,
 {
     fn as_range(&self) -> Range<usize> {
-        self.start.clone().into()..self.end.clone().into()
+        let start = self.start.clone().into();
+        let end = self.end.clone().into();
+
+        debug_assert!(
+            start <= end,
+            "slice index starts at {} but ends at {}",
+            start,
+            end
+        );
+
+        start..end
     }
 }
 
--- a/crates/rslint_errors/src/file.rs
+++ b/crates/rslint_errors/src/file.rs
@@ -137,10 +137,15 @@ pub struct FileSpan {
 
 impl FileSpan {
     pub fn new(file: FileId, span: impl Span) -> Self {
-        Self {
-            file,
-            range: span.as_range(),
-        }
+        let range = span.as_range();
+        debug_assert!(
+            range.start <= range.end,
+            "slice index starts at {} but ends at {}",
+            range.start,
+            range.end
+        );
+
+        Self { file, range }
     }
 }
 
--- a/crates/rslint_errors/src/file.rs
+++ b/crates/rslint_errors/src/file.rs
@@ -82,17 +82,7 @@ where
     T: Into<usize>,
 {
     fn as_range(&self) -> Range<usize> {
-        let start = self.start.clone().into();
-        let end = self.end.clone().into();
-
-        debug_assert!(
-            start <= end,
-            "slice index starts at {} but ends at {}",
-            start,
-            end
-        );
-
-        start..end
+        self.start.clone().into()..self.end.clone().into()
     }
 }
 
--- a/crates/rslint_lexer/src/lib.rs
+++ b/crates/rslint_lexer/src/lib.rs
@@ -234,7 +234,7 @@ impl<'src> Lexer<'src> {
             // but it may cause a panic for other crates which just consume the diagnostics
             let invalid = self.get_unicode_char();
             let err = Diagnostic::error(self.file_id, "", "expected hex digits for a unicode code point escape, but encountered an invalid character")
-                .primary(self.cur .. invalid.len_utf8(), "");
+                .primary(self.cur .. self.cur + invalid.len_utf8(), "");
             self.cur -= 1;
             return Err(Box::new(err));
         }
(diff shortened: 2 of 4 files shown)


crates/rslint_errors/src/file.rs
```diff
@@ -82,7 +82,17 @@ where
...
-        self.start.clone().into()..self.end.clone().into()
+        let start = self.start.clone().into();
+        let end = self.end.clone().into();
+
+        debug_assert!(
+            start <= end,
+            "slice index starts at {} but ends at {}",
+            start,
+            end
+        );
```
MichaReiser, line 93: I found a diagnostic where the range was `pos..1` instead of `pos..pos + 1`. This should help us find other places where the diagnostic range is invalid and prevent them in the future.

crates/rslint_lexer/src/lib.rs
```diff
@@ -357,21 +357,23 @@ impl<'src> Lexer<'src> {
...
+        self.next(); // eat over the \
+        if let Some(escape) = self.bytes.get(self.cur) {
             match escape {
-                b'u' if self.bytes.get(self.cur + 2) == Some(&b'{') => {
-                    self.advance(2);
-                    self.read_codepoint_escape().err()
-                }
-                b'u' => {
+                // Single escape character
+                b'\\' | b'n' | b'r' | b't' | b'b' | b'v' | b'f' | b'\'' | b'"' => {
```
MichaReiser, line 366: Not strictly necessary but helps because it now matches the structure of the spec.

crates/rslint_lexer/src/lib.rs
```diff
@@ -436,8 +438,10 @@ impl<'src> Lexer<'src> {
         let start = self.cur;
         let mut diagnostic = None;
 
-        while let Some(byte) = self.next_bounded() {
-            match *byte {
+        self.next(); // skip quote;
+
+        while let Some(byte) = self.current().copied() {
```
MichaReiser, line 443: The problem here was that `validate_escape_sequence` progressed the lexer position if it's a valid escape sequence. This resulted in a off by one error when calling `next_bounded` at the beginning of the next loop.

MichaReiser:
bench!

MichaReiser:
GitHub action isn't working for me anymore. Here the manual benchmark results

```
parser/compiler.js                    1.04     53.5±1.30ms    19.6 MB/sec    1.00     51.3±1.71ms    20.4 MB/sec
parser/d3.min.js                      1.00     32.0±0.42ms     8.2 MB/sec    1.00     31.9±0.50ms     8.2 MB/sec
parser/dojo.js                        1.02      2.6±0.05ms    26.4 MB/sec    1.00      2.6±0.03ms    26.8 MB/sec
parser/jquery.min.js                  1.01      9.3±0.11ms     8.9 MB/sec    1.00      9.2±0.09ms     9.0 MB/sec
parser/pixi.min.js                    1.00     40.1±0.58ms    10.9 MB/sec    1.01     40.7±1.26ms    10.8 MB/sec
parser/react-dom.production.min.js    1.05     12.4±0.18ms     9.3 MB/sec    1.00     11.8±0.14ms     9.7 MB/sec
parser/react.production.min.js        1.09  605.4±100.23µs    10.2 MB/sec    1.00    554.4±9.62µs    11.1 MB/sec
parser/tex-chtml-full.js              1.00     85.2±1.37ms    10.7 MB/sec    1.00     85.2±1.69ms    10.7 MB/sec
parser/three.min.js                   1.00     45.0±0.63ms    13.0 MB/sec    1.04     46.7±1.00ms    12.6 MB/sec
parser/vue.global.prod.js             1.01     14.9±0.20ms     8.1 MB/sec    1.00     14.7±0.16ms     8.2 MB/sec
```

MichaReiser:
!bench

crates/rslint_errors/src/file.rs
xunilrj: Should we also put this assert where the range was created? I mean... in the ```Diagnostic``` and/or in ```Token``` constructor?

crates/rslint_lexer/src/lib.rs
```diff
@@ -357,21 +357,23 @@ impl<'src> Lexer<'src> {
 
     // Validate a `\..` escape sequence and advance the lexer based on it
     fn validate_escape_sequence(&mut self) -> Option<Box<Diagnostic>> {
+        debug_assert_eq!(self.current().unwrap(), &b'\\');
         let cur = self.cur;
-        if let Some(escape) = self.bytes.get(self.cur + 1) {
+        self.next(); // eat over the \
```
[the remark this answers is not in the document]
xunilrj: > Eating over the \ makes sure that the lexer always progresses

What do you think of a ```debug_assert!``` to check if the lexer is progressing? I had one problem with the lexer stuck forever. I can create the PR.

xunilrj (approved):

crates/rslint_lexer/src/lib.rs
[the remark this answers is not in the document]
MichaReiser: Having something like `ParserProgress` for the lexer would certainly be nice!

crates/rslint_errors/src/file.rs
MichaReiser: I think that makes sense!

PR #1987, opened 2022-01-24
feat(rome_formatter): initial formatting of module

ematipico:
## Summary

This small PR sets the foundation for formatting modules. This PR will unlock two works that I already started:
- formatting of `import` syntax
- better  testing for the formatter (script / module)

I moved the parsing of `JsScript` inside a folder called `root`, where also `module.rs` will live.

## Test Plan

Current CI for now


MichaReiser (approved):
It would be nice if we could re-use some of the code between script/module and avoid copy pasting all of it. 

How did you test that the changes work? Any tests?

crates/rome_formatter/src/ts/root/mod.rs
```diff
@@ -0,0 +1,27 @@
+use crate::{hard_line_break, join_elements, FormatElement, Formatter};
+use rslint_parser::ast::JsModuleItemList;
+use rslint_parser::AstNode;
+use rslint_parser::AstNodeList;
+
+mod any_module_item;
+mod module;
+mod script;
+
+pub fn format_module_list(list: JsModuleItemList, formatter: &Formatter) -> FormatElement {
```
leops, line 10: Should this function should be called `format_module_item_list` instead ?

crates/rome_formatter/src/ts/root/module.rs
```diff
@@ -0,0 +1,33 @@
...
+            elements.push(hard_line_break());
+        }
+
+        elements.push(format_module_list(self.items(), formatter));
+
+        elements.push(formatter.format_token(&self.eof_token()?)?);
+
+        Ok(format_elements![
+            concat_elements(elements),
+            hard_line_break()
```
leops, line 30: I think the LSP protocol includes a boolean for whether or not the user wants a trailing newline in their file, which makes me think we shouldn't hardcode the last line break in the module or script element and that its printing should be delegated to the `Printer` (based on a new field in `FormatOptions`). That last part is probably work for another PR though ...

ematipico: Would you like to open a issue for that? I exactly copied to code we already have because I didn't want to break the existing code, as it's not the aim of the PR

ematipico:
> It would be nice if we could re-use some of the code between script/module and avoid copy pasting all of it.
> 

Yeah, but that's not the aim of this PR. 

> How did you test that the changes work? Any tests?

I tested them locally with other changes I have been doing, although they are too many so I decided to first open a PR to set the foundations.

crates/rome_formatter/src/ts/root/mod.rs
ematipico: I will address it in the next PR

ematipico:
> > It would be nice if we could re-use some of the code between script/module and avoid copy pasting all of it.
> 
> I don't agree here. You introduced new code by copy pasting the existing implementation. That's why I believe it would have been worthwhile to think about ways to avoid the code duplication,.

I am sorry, I don't want to overlook the issue, I just wanted to get this merged ASAP. I will address the issue when I will implement the syntax for modules.

PR #1988, opened 2022-01-24
test: better testing of the formatter

ematipico:
## Summary

This PR changes the way we test our formatter, by making more granular and more configurable, allowing us to test various situations based on the language of the file and its features.

Before we were stuck to parse each file as a script, preventing to parse specific language features that are available only when a file is a module.

Here's what's changed:

- moved all spec files inside a folder called `module/`
- moved files that have language features that are only allowed in loose mode inside a new folder called `script/`
- added a new parameter to the macro `gen_tests!`, a string that marks if a test is a script or a module (unfortunately this is something that we can't deduce by reading a file)
- changed the paths inside the of `gen_tests!` inside `spec_tests.rs` file
- removed some code from `rome_core::App`, which was written with JS in mind (which is not very Rust friendly)
- when formatting a file, we need to pass now the features that belong to file
- `App` has some API in charge to return information about the features available to each language and the features available to each file (`js` and `json` for now)
- changed the formatter API and created two new ones: `format_js_file` and `format_json_file` which are highly specialized and require less dynamic typing.  
- removed parsing and formatting of JSON, as it was a temporary solution for having a working formatter while the parser was under construction. The JSON parsing was relying on the JS parsing, which wasn't actually correct. Now the parser is stable and we can remove JSON parsing

## Test Plan

Added two tests inside `module/` that contains syntax that belongs to a script and made sure that their snapshot tests emits code that is not  formatted.


crates/rome_formatter/tests/specs/js/module/comments.js
```diff
@@ -1,6 +1,6 @@
-import {
-    func, // trailing comma removal
-} from 'module';
+// import {
+//     func, // trailing comma removal
+// } from 'module';
```
ematipico, line 3: I had to comment this because we don't have a yet its formatted syntax. I will re enable this test in another PR

crates/rome_formatter/src/lib.rs
```diff
@@ -174,55 +174,74 @@ impl Formatted {
...
     let printer = Printer::new(options);
     printer.print(element)
 }
+
+pub fn format_file_and_save(rome_path: &mut RomePath, options: FormatOptions, app: &App) {
+    let language = app.get_language(rome_path.extension().expect("Could not read the file"));
+
+    let result = match language {
+        Language::Js | Language::Ts => {
+            let features = JsFileFeatures::default().module();
```
ematipico, line 231: At the moment we fall back to a module, even though we don't know that yet. We will now that once we will be able to detect a project with a `package.json`

crates/rome_formatter/src/ts/root/mod.rs
```diff
@@ -7,7 +7,7 @@ mod any_module_item;
 mod module;
 mod script;
 
-pub fn format_module_list(list: JsModuleItemList, formatter: &Formatter) -> FormatElement {
+pub fn format_module_item_list(list: JsModuleItemList, formatter: &Formatter) -> FormatElement {
```
ematipico, line 10: cc @leops

crates/rome_formatter/src/lib.rs
```diff
@@ -174,55 +174,74 @@ impl Formatted {
...
-            };
-
-            result
-        } else {
-            Err(FormatError::CapabilityDisabled)
-        }
+/// Formats a JSON file based on its features.
+///
+/// It returns a [Formatted] result, which the user can use to override a file.
+pub fn format_json_file(
```
MichaReiser, line 180: I don't like having even more `format` function. Ideally, there should be less than needing a format function for each language. What's preventing us from keeping these generic?

crates/rome_formatter/tests/specs/js/module/with.js.snap
```diff
@@ -0,0 +1,21 @@
+---
```
ematipico, line 1: @MichaReiser as you can this file is not formatted because it gets evaluated as a module, resulting in parsing errors.

crates/rome_formatter/src/lib.rs
```diff
@@ -174,55 +174,74 @@ impl Formatted {
...
+) -> FormatResult<Formatted> {
+    let handler = app.get_js_features();
+
+    if handler.capabilities().format {
+        let buffer = path.get_buffer_from_file();
+        let parsed_result = if features.module {
+            parse_module(buffer.as_str(), 0).syntax()
+        } else {
+            parse_text(buffer.as_str(), 0).syntax()
+        };
```
MichaReiser, line 213: This isn't supporting typescript. 

I've a local change that supports module script and typescript. I implemented it by adding a `parse_syntax` method that accepts a `Syntax` 

```
pub fn parse_syntax(text: &str, file_id: usize, syntax: Syntax) -> Parse<JsAnyRoot> {
```

We probably want to have a different way of specifying the supported syntax but this seems to be the most flexible with the current architecture. 

The `parse_module` , `parse_expression_snipped`, and `parse_text` (can we rename it to `parse_script`) can then call into `parse_syntax` directly

crates/rome_formatter/src/lib.rs
ematipico: > What's preventing us from keeping these generic?

The features that differ among languages and the fact that we parse the source right inside the formatter.

We could have one single API, but in order to achieve that, we would need to pass to `format` a `SyntaxNode`, which I think we should do.

crates/rome_core/src/file_handlers/javascript.rs
```diff
@@ -23,3 +39,44 @@ impl ExtensionHandler for JsFileHandler {
...
+        self.typescript = false;
+        self.tsx = false;
+        self.jsx = false;
+        self
+    }
+
+    /// Enables TypeScript
+    pub fn typescript(mut self) -> Self {
+        self.script = false;
+        self.module = true;
```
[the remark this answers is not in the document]
ematipico: Yeah it seems that it's possible to have shabang, although TypeScript allows to have module syntax while having the shebang.

This means that we are not a position, yet, to infer this information with what we have. Using module would allow us to actually parse more syntax.

crates/rome_core/src/file_handlers/javascript.rs
```diff
@@ -1,13 +1,29 @@
 use super::{ExtensionHandler, Mime};
+use std::fmt::Debug;
 
 #[derive(Debug, PartialEq, Eq)]
-pub struct JsFileHandler {}
+pub struct JsFileHandler;
+
+/// Features that belong to the language
+#[derive(Debug)]
+pub struct JsFileFeatures {
```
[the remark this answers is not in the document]
ematipico: The description is wrong, sorry. This is a set of information that should be associated to the single file, so when we read it, we know what we should do based on this `struct`.

This is not an information that is exclusive to the parser, but other entities too.

> js can either be script or module

Yeah, that will probably change. It might make sense to have an enum of a bit flag. A file can only be a script, module or a snippet (based on the grammar)

crates/rome_formatter/src/lib.rs
ematipico: > This isn't supporting typescript.

Could you elaborate this please? I don't want to support TS right now

MichaReiser: I want to avoid having a `parse_*` method for `parse_typescript_module`, `parse_typescript_script` and then potentially all other combinations with different JsFeatures turned on or of.

That's why our design should be open to supporting new dialects (with JSX, without JSX) without the need to add a new parse method for each possible combination.

crates/rome_formatter/src/lib.rs
ematipico: Oh, and the JSON is also another reason why we can't split it :)

MichaReiser: > Oh, and the JSON is also another reason why we can't unify it :)

What do you mean by that?

Also, I don't mind deleting the JSON formatting/parsing. The main motivation was to have something running until we have a JS parser in place.

crates/rome_formatter/src/lib.rs
ematipico: Sure, it makes sense. Although, I think we should avoid this change here because it's not related to the initial change. We can change it once we will make the `parse_*` stable?

MichaReiser (approved):

PR #1991, opened 2022-01-24
split newline tokens

xunilrj:
Better handling of new lines to facilitate the work of the formatter. see https://github.com/rome/tools/pull/1974#discussion_r789627340

### DIFF
--- a/crates/rslint_lexer/src/lib.rs
+++ b/crates/rslint_lexer/src/lib.rs
@@ -125,16 +125,15 @@ impl<'src> Lexer<'src> {
 
     fn consume_newlines(&mut self) -> usize {
         let start = self.cur;
-        while self.current().is_some() {
+        if self.current().is_some() {
             let chr = self.get_unicode_char();
             if is_linebreak(chr) {
                 self.state.had_linebreak = true;
                 self.cur += chr.len_utf8();
-                if chr == '\r' && self.current() == Some(&('\n' as u8)) {
-                    self.cur += '\n'.len_utf8(); 
+                if chr == '\r' && self.current() == Some(&b'\n') {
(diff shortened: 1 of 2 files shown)


crates/rslint_lexer/src/lib.rs
```diff
@@ -130,6 +130,9 @@ impl<'src> Lexer<'src> {
             if is_linebreak(chr) {
                 self.state.had_linebreak = true;
                 self.cur += chr.len_utf8();
+                if chr != '\r' {
+                    break;
+                }
```
MichaReiser, line 135: We'll need a lookahead that the next character is a `\n` to correctly handle `\r\r`

MichaReiser (changes requested):
Nice, Thanks for working on this! 

We should either add some more lexer tests or a parser test to verify that the lexer supports all newline characters defined in the [JS spec](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-line-terminators)

crates/rslint_lexer/src/lib.rs
```diff
@@ -130,6 +130,12 @@ impl<'src> Lexer<'src> {
             if is_linebreak(chr) {
                 self.state.had_linebreak = true;
                 self.cur += chr.len_utf8();
+                let next_chr = self.current();
+                if chr == '\r' && next_chr == Some(&('\n' as u8)) {
+                    continue;
```
MichaReiser, line 135: Nit: can we skip the `continue` in that case and instead just move the cursor position? 

```
self.cur += '\n'.len_utf8(); 
break;
```

MichaReiser (approved):

PR #1992, opened 2022-01-24
feat(rome_formatter): format import syntax

ematipico:
## Summary

This PR formats the whole `import ...` syntax.

I took the liberty to create a new `format_list` API inside the `Formatter`, where there was some repeated pattern. The documentation is not the best to be honest, it would be great if someone else would suggest something more to it.

There's an existing issue where a trailing comment is not well positioned in some cases. I reproduced this issue also with existing grammar, so it would be better to fix it with another PR so to not add too much to this PR.

## Test Plan

Added various tests, with long variables and comments.


crates/rome_formatter/src/ts/import/bare_clause.rs
```diff
@@ -0,0 +1,18 @@
+use crate::{
+    empty_element, format_elements, space_token, FormatElement, FormatResult, Formatter,
+    ToFormatElement,
+};
+use rslint_parser::ast::JsImportBareClause;
+
+impl ToFormatElement for JsImportBareClause {
+    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        let source = formatter.format_node(self.source()?)?;
+        let assertion = if let Some(assertion) = self.assertion() {
```
NicholasLYang, line 10: Seems like whenever we format assertions we do this same check and insert space. Why not insert the space when we format the assertions then?

NicholasLYang: Alternatively, we could have a function that formats with a leading space or returns an empty element.

crates/rome_formatter/tests/specs/js/module/import/import_specifiers.js
```diff
@@ -0,0 +1,15 @@
...
+    loooooooooooooooooooong2 as moreeeeeeloooooooooooooooooooong2,
+// some other comment
+} from "test";
+import {
+    loooooooooooooooooooong3 as moreeeeeeloooooooooooooooooooong3
+    // some funky comment
+} from "test";
+
+import {loooooooooooooooooooong3,loooooooooooooooooooong5,loooooooooooooooooooong6 } from "boo"
+import { f as x, default as w, "a-b-c" as y } from "b";
```
NicholasLYang, line 15: Do we want to include some tests for a default + named imports? Like `import a, { b, c, d} from "module"`

crates/rome_formatter/src/formatter.rs
```diff
@@ -270,6 +270,30 @@ impl Formatter {
         Ok(result.into_iter())
     }
 
+    /// It formats a list of nodes that are not separated.
+    pub fn format_list<List, Node: Clone + AstNode + ToFormatElement>(
```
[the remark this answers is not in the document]
ematipico: Whitespace is a trivia, which is not considered a separator. This method is just an utility methods for `AstNodeList`. Where's we also have an utility method for `AstSeparatedList`

crates/rome_formatter/src/ts/import/bare_clause.rs
ematipico: > Seems like whenever we format assertions we do this same check and insert space. Why not insert the space when we format the assertions then?

Because the assertions, per se, don't know where they are placed inside the code and it's responsibility of the parent to decide how they have to be formatted. Because the parent knows the context of the assertions. 

> Alternatively, we could have a function that formats with a leading space or returns an empty element.

We might have a white space long 4, which is not ideal. It might take longer in terms of performance to check its length and decide to print. You suggestion makes sense and it's worth exploring, although this is a pattern we have in the whole formatter so it might make more sense to explore in a separate PR

NicholasLYang: Yeah makes sense! We can note it down for later.

crates/rome_formatter/src/formatter.rs
```diff
@@ -266,6 +266,31 @@ impl Formatter {
...
```
leops, line 270: I wonder if this could have a more explicit name to help distinguish it from `format_nodes`, at least its documentation should explain that it falls back to printing individual nodes verbatim if they fail formatting and that elements are separated by line breaks / empty lines

crates/rome_formatter/src/ts/directives.rs
```diff
@@ -1,25 +1,15 @@
...
-                    .trim_start()
-                    .trim_end()
-            }
-        };
-
-        (directive, elem)
-    }))
+pub fn format_directives_list(directives: JsDirectiveList, formatter: &Formatter) -> FormatElement {
+    if !directives.is_empty() {
+        format_elements![formatter.format_list(directives), hard_line_break()]
```
leops, line 9: It might be better to insert this line break in the parent node using `join_elements_hard_line` to insert either a hard line break or empty line depending on the input

ematipico: `formatted_list` already applies `join_elements_hard_line` at the end. Why should we apply `join_elements_hard_line` again? Also, this function requires a list... and what if the parent doesn't have a list? Or maybe I didn't understand your suggestion

leops: Right I think my explanation wasn't very clear, I meant doing a `join_elements_hard_line` between the directives list and the rest of the script / module, in pseudo-code `join_elements_hard_line([directives.to_format_element(), statements.to_format_element()])` for instance, so that the formatter respects the empty line in inputs like:

```
"use strict";

statement();
```

crates/rome_formatter/src/formatter.rs
```diff
@@ -266,6 +266,37 @@ impl Formatter {
...
     }
 
+    /// It formats a list of nodes that are not separated. It's a ad-hoc function to
+    /// format lists that implement [rslint_parser::AstNodeList].
+    ///
+    /// The elements of the list are joined together using [join_elements_hard_line], which will
+    /// end up separated by hard lines or empty lines.
+    ///
+    /// If the formatter fails to format an element, said element gets printed verbatim.
+    pub fn format_list<List, Node: Clone + AstNode + ToFormatElement>(
```
MichaReiser, line 276: Is it still necessary that `Node` implements `Clone`?

crates/rome_formatter/src/ts/directives.rs
ematipico: Unfortunately I ran into some typing issue while trying to fix the problem. I talked to @leops  and we agreed to revisit it later

crates/rome_formatter/src/intersperse.rs
```diff
@@ -98,6 +99,7 @@ impl<I, F, N> IntersperseFn<I, F, N>
 where
     I: Iterator<Item = (N, FormatElement)>,
     F: FnMut(&N, &N) -> FormatElement,
+    N: AstNode,
```
leops, line 102: There isn't really any reason to constraint the `N` type since `IntersperseFn` doesn't assume anything specific about it, and it will allow an alternative implementation of `join_elements_hard_line` to process an iterator over generic SyntaxNodes instead of a specific type of AstNodes

ematipico: Sorry, old code while I was trying to fix a previous issue. Removed

leops (approved):

PR #1993, opened 2022-01-24
feat: arrow function and assignment line breaking

NicholasLYang:
## Summary

Added line breaking after `=>` and `=` in arrow functions and assignments respectively.

## Test Plan
We already have some tests in place for arrow functions and assignments.


crates/rome_formatter/src/format_element.rs
```diff
@@ -395,6 +395,71 @@ pub fn soft_indent<T: Into<FormatElement>>(content: T) -> FormatElement {
     }
 }
 
+/// A line break and indent if the enclosing [Group] doesn't fit on a single, line, a space
+/// otherwise. Does not insert a new line at the end, as this messes up semicolon placement
```
MichaReiser, line 399: Nit: The `messes up` semicolon placement is to be specific to one use case and out of context in this documentation (what semicolon?). 

Can we describe this more conceptually. It inserts an optional line break and indents the content if the group doesn't fit on a new line. Inserts a space otherwise.

crates/rome_formatter/src/format_element.rs
```diff
@@ -395,6 +395,71 @@ pub fn soft_indent<T: Into<FormatElement>>(content: T) -> FormatElement {
...
+/// ]);
+///
+/// assert_eq!(
+///   "a = 10",
+///   format_element(&elements, FormatOptions::default()).code()
+/// );
+/// ```
+///
+#[inline]
+pub fn soft_indent_or_space<T: Into<FormatElement>>(content: T) -> FormatElement {
```
MichaReiser, line 450: nit: Wdyt of `soft_line_indent`?

crates/rome_formatter/src/ts/expressions/import_meta_expression.rs
```diff
@@ -0,0 +1,13 @@
+use rslint_parser::ast::ImportMeta;
+
+use crate::{format_elements, FormatElement, FormatResult, Formatter, ToFormatElement};
+
+impl ToFormatElement for ImportMeta {
```
leops, line 5: I think the AST nodes that implement `ToFormatElement` must be added to the match in `cst.rs`, as I understand it cannot be automatically generated yet because not every node is supported by the formatter

MichaReiser: We could generate the file if the code gen generates empty boilderplates for every node (that call into `format_verbatim`). It's just that no one implemented the code gen part yet ;)

crates/rome_formatter/src/format_element.rs
NicholasLYang: Sure! I added the `or_space` to match `soft_line_break_or_space`. Do you think we should keep it?

format-folder.patch
```diff
@@ -0,0 +1,278 @@
+From 8a724165569993b942bec7ea63cac375faa54881 Mon Sep 17 00:00:00 2001
```
ematipico, line 1: This file should not be here

ematipico (changes requested):
There are two files: `checker.js` and `format-folder.patch` that don't belong to this PR, I think

crates/rome_formatter/src/format_element.rs
MichaReiser: Probably makes sense to have it explicit: `soft_line_indent_or_space`.

crates/rome_formatter/src/ts/auxiliary/any_function.rs
```diff
@@ -43,7 +46,21 @@ impl ToFormatElement for JsAnyFunction {
             tokens.push(space_token());
         }
 
-        tokens.push(formatter.format_node(self.body()?)?);
+        let body = self.body()?;
+        let body_has_soft_line_break = matches!(
```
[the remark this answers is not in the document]
NicholasLYang: Honestly I ported this from the old Rome, but the general idea seems to be that we don't want to line break too early. We don't know that these nodes have line breaks other than their implementation because there's no real overarching rule about which nodes have line breaks.

I'll try to flesh out the explanation and provide some examples

crates/rome_formatter/tests/specs/js/arrow/params.js.snap
```diff
@@ -329,7 +330,8 @@ foo(
 ---
 # Output
 fooooooooooooooooooooooooooooooooooooooooooooooooooo(
-	(action) => (next) => dispatch(action),
+	(action) => (next) =>
+		dispatch(action),
```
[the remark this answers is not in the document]
NicholasLYang: Oooh yeah good catch! Forgot to group at the end of the function formatting

crates/rome_formatter/src/ts/auxiliary/any_function.rs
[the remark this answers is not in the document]
NicholasLYang: According to prettier, chained arrow functions are not broken until it's so long that they're all line broken. I've noticed that pattern in prettier where there's a bail for extremely long lines where it vertically stacks everything. We can implement that in the future. As for now, we shouldn't break after a `=>` if it's followed by another arrow function.

MichaReiser (approved):

PR #1998, opened 2022-01-25
feat: add a macro for executing an expression on all AST node types

leops:
## Summary

This PR adds a `map_syntax_node` macro with the following syntax: `map_syntax_node!(<SyntaxNode expression>, <AstNode> => <expression>, <JsSyntaxKind> => <expression>)`. The macros expands to a match on the kind of the SyntaxNode in the first argument, and for each JsSyntaxKind that implements AstNode duplicates the second argument as a match branch that casts the SyntaxNode to the appropriate AstNode type before executing the expression. For all JsSyntaxKind that do not implement AstNode the third argument is expanded as a fallback branch in the match expression.

As it stands this is only used to implement `ToFormatElement` for `SyntaxNode`, but I chose to implement this as a macro so that it can be eventually leveraged in other algorithms that implement a tree traversal and may need to "cast" a SyntaxNode into all possible AstNode types.

### DIFF
--- a/crates/rome_formatter/src/cst.rs
+++ b/crates/rome_formatter/src/cst.rs
@@ -3,14 +3,12 @@ use rslint_parser::{map_syntax_node, SyntaxNode};
 
 impl ToFormatElement for SyntaxNode {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        map_syntax_node!(
-            self.clone(),
-            node => node.to_format_element(formatter),
-            _ => Ok(formatter.format_verbatim(self))
-        )
+        map_syntax_node!(self.clone(), node => node.to_format_element(formatter))
     }
 }
 
+// Temporary macro providing a default implementation of ToFormatElement to
+// the node types currently missing one as part of #1997
 macro_rules! impl_format_verbatim {
     ( $( $node:ident, )* ) => {
         $( impl $crate::ToFormatElement for ::rslint_parser::ast::$node {
--- a/crates/rome_formatter/src/cst.rs
+++ b/crates/rome_formatter/src/cst.rs
@@ -1,347 +1,148 @@
 use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
-use rslint_parser::ast::{
-    ImportMeta, JsArrayBindingPattern, JsArrayExpression, JsArrowFunctionExpression,
-    JsBlockStatement, JsBooleanLiteralExpression, JsCallArguments, JsCallExpression, JsCaseClause,
-    JsCatchClause, JsClassStatement, JsConstructorParameters, JsContinueStatement,
-    JsDebuggerStatement, JsDefaultClause, JsDefaultImportSpecifier, JsDoWhileStatement,
-    JsEmptyStatement, JsExport, JsExportAsClause, JsExportClassClause, JsExportDefaultClassClause,
-    JsExportDefaultExpressionClause, JsExportDefaultFunctionClause, JsExportFromClause,
-    JsExportFunctionClause, JsExportNamedClause, JsExportNamedFromClause,
-    JsExportNamedFromSpecifier, JsExportNamedShorthandSpecifier, JsExportNamedSpecifier,
-    JsExportVariableClause, JsExpressionStatement, JsFinallyClause, JsForInStatement,
-    JsForStatement, JsFunctionStatement, JsGetterClassMember, JsIdentifierBinding,
-    JsIdentifierExpression, JsIfStatement, JsImport, JsImportAssertion, JsImportAssertionEntry,
-    JsImportBareClause, JsImportCallExpression, JsImportDefaultClause, JsImportNamedClause,
-    JsImportNamespaceClause, JsLabeledStatement, JsLiteralExportName, JsModule, JsModuleSource,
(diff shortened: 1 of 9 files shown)


crates/rslint_parser/src/ast/generated/macros.rs
```diff
@@ -0,0 +1,928 @@
+//! Generated file, do not edit by hand, see `xtask/src/codegen`
+
+#[macro_export]
+macro_rules! map_syntax_node {
```
MichaReiser, line 4: Looks neat. How does the use of a macro affect compile time?

leops: I haven't seen any noticeable change in compilation time, currently the macro is only used in a single place and expands to a match expression similar to what used to be there so there shouldn't be any major difference

ematipico:
I am not sure about this solution, the part that makes me uncomfortable is the introduction of unsafe code where there wasn't before. Is it really needed?

leops:
Technically the `new_unchecked` doesn't need to be tagged as `unsafe` as it doesn't directly do anything that the Rust language would deem unsafe, although constructing an AstNode from a non-matching SyntaxNode may lead to panics (and maybe undefined behavior depending on how rowan handles that). I'm mostly using as a signal that means "don't use this function" because while it has to be a part of the public API of the `rslint_parser` crate to be called in the `map_syntax_node` macro, it shouldn't be used by anything else. Ultimately if having an `unsafe` function is an issue I could remove it and replace its usage with `cast(syntax).unwrap()`, which should compile to the same code if the Rust compiler is smart enough to detect this code can never panic

xunilrj (approved):

leops:
Unfortunately creating the AstNode struct directly in the macro wouldn't work as the macro expansion is scoped to the crate that calls it, in this case `rome_formatter` which doesn't have access to the `syntax` field as it is private to the `rslint_parser` crate

ematipico:
Between `unsafe` and `.unwrap()`, I am more keen to those `.unwrap()`. The usage of `unsafe` could be to anything. The usage of `unwrap` is narrower

leops:
Actually that was sort of the point. There is no "unsafety" as the Rust languages defines it here, I could leave the code as it is and just remove the unsafe and it would compile and work exactly the same, as constructing an AstNode from the wrong kind of SyntaxNode does not lead to any undefined behavior (only potential missing children errors or panics when using the accessors because the slots don't match). But I chose to manually tag the `new_unchecked` function as unsafe precisely because it sounds scary and will deter or at least prompt a second thought to anyone trying to use it.

ematipico:
> Actually that was sort of the point. There is no "unsafety" as the Rust languages defines it here, I could leave the code as it is and just remove the unsafe and it would compile and work exactly the same, as constructing an AstNode from the wrong kind of SyntaxNode does not lead to any undefined behavior (only potential missing children errors or panics when using the accessors because the slots don't match). But I chose to manually tag the `new_unchecked` function as unsafe precisely because it sounds scary and will deter or at least prompt a second thought to anyone trying to use it.

Makes sense! Thank you for the explanation!

crates/rome_formatter/src/cst.rs
```diff
@@ -1,347 +1,149 @@
...
-                Ok(formatter.format_list(JsStatementList::cast(self.clone()).unwrap()))
-            }
-            JsSyntaxKind::JS_VARIABLE_DECLARATIONS => JsVariableDeclarations::cast(self.clone())
-                .unwrap()
-                .to_format_element(formatter),
-            JsSyntaxKind::JS_IMPORT_ASSERTION_ENTRY => JsImportAssertionEntry::cast(self.clone())
-                .unwrap()
-                .to_format_element(formatter),
+macro_rules! impl_format_verbatim {
+    ( $( $node:ident, )* ) => {
```
MichaReiser, line 15: I assume this part will be replaced by #2031 ?

crates/rslint_parser/src/ast/generated/macros.rs
```diff
@@ -0,0 +1,1185 @@
+//! Generated file, do not edit by hand, see `xtask/codegen`
+
+#[macro_export]
+macro_rules! map_syntax_node {
```
MichaReiser, line 4: Would be good to have some documentation explaining how this macro is to be used. 

I don't know if we need the fallback. The only case where a `kind` has no corresponding `AstNode` is if the kind is a token kind and not a node kind.... I believe it's fine to panic in that case because all kind of weird things can happen if someone starts creating a `SyntaxNode` with a token kind (maybe a reason to separate the two enums?) 

An alternative would be to call the method `map_syntax_element` that accepts `SyntaxNode` and `SyntaxToken`. The macro then has the form `map_syntax_element(element, node_mapper, token_mapper)`

crates/rome_formatter/src/cst.rs
leops: Yes, this is a temporary measure until the codegen part is in place to make sure everything compiles

crates/rslint_parser/src/ast/generated/macros.rs
leops: This is mostly a matter of leaving the choice of whether to panic or return an error to the code using the macro rather than the macro itself, but I could have the fallback branch default to `_ => unreachable!()` when its not specified

MichaReiser (approved):

PR #2000, opened 2022-01-25
feat(rslint_parser): Add `parse` method

MichaReiser:
## Summary

Adds a method that allows parsing a file with the provided syntax. This is useful because in many cases all we want to do is to parse the root of a program and enable certain syntax features. 

It's likely that the `Syntax` struct will change in the future. It's just what already existed.

This PR also adds a way to run TypeScript codegen tests by writing `// TYPESCRIPT`. We probably want to have a more elaborated approach of doing this in the codegen but it unblocks me from working on parsing typescript types. 

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_formatter/src/lib.rs
+++ b/crates/rome_formatter/src/lib.rs
@@ -55,7 +55,7 @@ mod intersperse;
 mod printer;
 mod ts;
 pub use formatter::Formatter;
-use rslint_parser::{parse_module, SyntaxError, SyntaxNode};
+use rslint_parser::{parse, Syntax, SyntaxError, SyntaxNode};
 
 pub use format_element::{
     block_indent, concat_elements, empty_element, group_elements, hard_line_break, if_group_breaks,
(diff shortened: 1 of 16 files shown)


crates/rome_formatter/src/lib.rs
```diff
@@ -183,13 +181,9 @@ pub fn format_element(element: &FormatElement, options: FormatOptions) -> Format
...
     let result = if app.can_format(rome_path) {
         let buffer = rome_path.get_buffer_from_file();
-        let features = JsFileFeatures::default().module();
-        let parsed_result = if features.module {
-            parse_module(buffer.as_str(), 0).syntax()
-        } else {
-            parse_text(buffer.as_str(), 0).syntax()
-        };
-        Some(format(options, &parsed_result))
+        let syntax = Syntax::default().module();
```
MichaReiser, line 184: We can use `JsFileFeatures` if it can be derived from the `rome_path` but feels overkill for now.

MichaReiser:
@ematipico should we delete the `JsFeatures` for now. They aren't really used anywhere or even derived. Feels like code that is in place for sometime in the future but we may end up never using it.

xunilrj (approved):

crates/rome_formatter/src/lib.rs
ematipico: We are not in a position yet where we can produce a `JsFileFeatues` from `rome_path`. It was my initial objective and I hope to have it in the future once we will have a project explorer and a project setting.

I think for now this change is more than enough. Once we will have more languages (JSON for example) we will think to a better solution

ematipico (approved):

PR #2001, opened 2022-01-25
TS-Type Parsing

MichaReiser:
## Summary

Adds parsing for typescript types. I started with the unofficial [TypeScript grammar](https://gray-sea-07a57a41e.azurestaticapps.net/#prod-MappedObjectType) specification but compared it with TypeScript's implementation to ensure we capture the "observed" behaviour of the TypeScript and not what the maybe outdated spec defines. 

The PR is larger but most of it are straightforward implementations of various `parse_ts_*` methods with their corresponding tests. I suggest taking a look at the `js.ungram` file. It contains the most controversial changes. 

This PR focuses solely on parsing the happy path. It has some error recovery logic but it isn't fleshed out nor tested. It also doesn't implement any semantic checks. 

## Test Plan

I added a large number of tests for isolated syntax. 

The TS coverage is now 

```
│   17164   │  9941  │  5845  │  1378  │  57.92   │
```

From before 

```
│   17149   │  9721  │  5476  │  1952  │  56.69   │
```

The PR moves many panicking tests to failed but also fixes a bunch. Many of the type parsing tests are still failing because the test files use `interfaces` or `declare` syntax that the parser doesn't yet support.


MichaReiser:
lol, this fixes the last panic. Nice ;)

crates/rome_rowan/src/syntax_factory.rs
```diff
@@ -126,22 +126,23 @@ pub trait SyntaxFactory: fmt::Debug {
 
 /// Iterator that "fixes up" a separated list by inserting empty slots for any missing
 /// separator or element.
-struct SeparatedListWithMissingNodesOrSeparatorSlotsIterator<'a, K: SyntaxKind> {
+struct SeparatedListWithMissingNodesOrSeparatorSlotsIterator<'a, K: SyntaxKind, S: Separator<K>> {
```
MichaReiser, line 129: TypeScript allows `,` and `;` as separator between object type members: `{ a: string, b: number; c: "test" }`

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1142,16 +1159,11 @@ fn parse_paren_or_arrow_expr(p: &mut Parser, context: ExpressionContext) -> Pars
...
-        if p.at(T![:]) {
-            let complete = ts_type_or_type_predicate_ann(p, T![:]);
-            if let Some(mut complete) = complete {
-                complete.err_if_not_ts(
-                    p,
-                    "arrow functions can only have return types in TypeScript files",
-                );
-            }
-        }
+        TypeScript
```
MichaReiser, line 1162: I didn't add the `TypeScript` gating code everywhere it is needed. That's something we must do when working on the remaining TS syntax.

xtask/codegen/js.ungram
```diff
@@ -1383,218 +1387,383 @@ TsImportEqualsDecl =
...
+	'.'
+	right: TsAnyName
 
-TsTypeName = ident: 'ident'
-TsQualifiedPath = lhs:TsEntityName '.' rhs:TsTypeName
 
+TsArrayType =
+	element_type: TsType
+	'['
+	']'
```
ematipico, line 1695: Arrays types can also be defined like this: `Array<string>`. Should we add this to the grammar?

xtask/codegen/js.ungram
```diff
@@ -1383,218 +1387,383 @@ TsImportEqualsDecl =
...
-	TsTypeName
-	| TsQualifiedPath
+TsImportTypeQualifier =
+	'.'
+	right: TsAnyName
 
-TsTypeName = ident: 'ident'
-TsQualifiedPath = lhs:TsEntityName '.' rhs:TsTypeName
 
+TsArrayType =
```
ematipico, line 1692: Small example please?

xtask/codegen/js.ungram
```diff
@@ -1383,218 +1387,383 @@ TsImportEqualsDecl =
...
+	'['
+	']'
 
-TsTypeArgs = '<' TsTypeArgList '>'
-TsTypeArgList = (TsType (',' TsType)* )
+TsIndexedAccessType =
+	object_type: TsType
+	'['
+	index_type: TsType
+	']'
```
ematipico, line 1701: Small example please?

xtask/codegen/js.ungram
```diff
@@ -1383,218 +1387,383 @@ TsImportEqualsDecl =
...
 
+TsTupleType =
+	'['
+	elements: TsTupleTypeElementList
+	']'
```
ematipico, line 1706: Small example please?

crates/rslint_parser/src/parser.rs
```diff
@@ -373,6 +373,18 @@ impl<'t> Parser<'t> {
         }
     }
 
+    #[inline]
```
ematipico, line 376: Small documentation please?

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -49,6 +49,13 @@ impl ParsedSyntax {
         }
     }
 
+    pub fn marker(&self) -> Option<&CompletedMarker> {
```
ematipico, line 52: Can't find this function. Probable it's not used?

crates/rslint_parser/src/state.rs
```diff
@@ -542,3 +542,20 @@ impl ChangeParserState for WithLabel {
         state.label_set.pop();
     }
 }
+
+pub(crate) struct EnterType;
```
ematipico, line 546: Small documentation please?

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -391,12 +380,38 @@ pub(super) fn parse_parameter(p: &mut Parser, context: ExpressionContext) -> Par
...
+                ts_only_syntax_error(p, "Type annotations", annotation.range(p).as_range())
+            })
+            .ok();
+
         parse_initializer_clause(p, context).ok();
         m.complete(p, JS_PARAMETER)
     })
 }
 
+pub(super) fn skip_parameter_start(p: &mut Parser) -> bool {
```
ematipico, line 395: Can we document the function please? I'm having hard time to understand what's used for

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -38,1143 +44,1390 @@ pub const DISALLOWED_TYPE_NAMES: &[&str] = &[
...
+            IntersectionOrUnionType::Intersection => parse_ts_primary_type(p),
         }
-        type_member_semi(p);
-        m.complete(p, TS_PROPERTY_SIGNATURE)
-    })
+    }
 }
 
-pub(crate) fn try_parse_index_signature(
+#[inline]
```
ematipico, line 183: Why are we ` #[inline]` this function?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -3,26 +3,32 @@
 use super::expr::{
```
ematipico, line 3: `typescript.rs` has grown too much I think. Can we break it down into separate files and make is a folder `typescript/` instead? Not necessarily  now, maybe later

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -21,9 +21,36 @@ pub(crate) fn get_precedence(tok: JsSyntaxKind) -> Option<u8> {
     })
 }
 
-pub(crate) fn expect_keyword(p: &mut Parser, keyword_name: &str, kind: JsSyntaxKind) {
-    if p.at(T![ident]) && p.cur_src() == keyword_name {
+#[inline]
+pub(crate) fn is_at_contextual_keyword(p: &Parser, name: &str) -> bool {
```
ematipico, line 25: Small documentation?

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -21,9 +21,36 @@ pub(crate) fn get_precedence(tok: JsSyntaxKind) -> Option<u8> {
...
+    is_nth_at_contextual_keyword(p, 0, name)
+}
+
+#[inline]
+pub(crate) fn is_nth_at_contextual_keyword(p: &Parser, n: usize, name: &str) -> bool {
```
ematipico, line 30: Small documentation?

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -21,9 +21,36 @@ pub(crate) fn get_precedence(tok: JsSyntaxKind) -> Option<u8> {
...
+    p.nth_at(n, T![ident]) && p.nth_src(n) == name
+}
+
+pub(crate) fn eat_contextual_keyword(
```
ematipico, line 34: Small documentation?

crates/rslint_parser/src/syntax/util.rs
```diff
@@ -21,9 +21,36 @@ pub(crate) fn get_precedence(tok: JsSyntaxKind) -> Option<u8> {
...
+) -> bool {
+    if is_at_contextual_keyword(p, keyword_name) {
         p.bump_remap(kind);
+        true
+    } else {
+        false
+    }
+}
+
+pub(crate) fn expect_contextual_keyword(
```
ematipico, line 47: Small documentation?

ematipico (changes requested):
The PR is too big and I am having hard time to check all of it, especially `typescript.rs`. I left some comments around the grammar. 

I believe the majority of the functions should require some documentation, so other developers can piggyback the code without too much friction

MichaReiser:
> I believe the majority of the functions should require some documentation, so other developers can piggyback the code without too much friction

The functions are all named after the grammar rules. I don't know what else to document other than that they parse the syntax after which they're named. But that's already given by the function name.

xtask/codegen/js.ungram
```diff
@@ -1383,218 +1387,383 @@ TsImportEqualsDecl =
...
 
-// typescript type reference
-TsTypeRef = name:TsEntityName type_args:TsTypeArgs
+TsReferenceType =
+	name: TsAnyName
+	type_arguments: TsTypeArgs?
 
+TsTypeAlias =
+	'type'
+	binding_identifier: JsIdentifierBinding
```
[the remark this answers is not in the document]
MichaReiser: Good catch. That's a bit surprising but makes sense. I can introduce a `TsIdentifierBinding`

xtask/codegen/js.ungram
```diff
@@ -1383,218 +1387,383 @@ TsImportEqualsDecl =
...
+// type OptionsFlags<Type> = {
+//   [Property in keyof Type]: boolean;
+// };
+TsMappedType =
+	'{'
+	readonly_modifier: TsMappedTypeReadonlyModifierClause?
+	'['
+	property_name: TsTypeParameterName
+	'in'
+	keys_type: TsType
```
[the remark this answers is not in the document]
MichaReiser: That's correct but it would requires different union types for all cases where only a limited set of types is needed. This is the same as for `BinaryExpression` where not any expressions are valid left- or right-hand sides. Nevertheless, the expression is typed as any expression. 

Allowing more syntax provides more flexibility and allows to mark nodes as invalid in specific context if we have a second pass. Overall, the parser's recovery stands in direct conflict with a strict AST because it limits it's options. 

I believe that using the more generic type is better than having a over detailed AST which requires a ton of additional nodes and unions and naming is hard.

xtask/codegen/js.ungram
MichaReiser: This is a `TsReferenceType` with type parameters.

xtask/codegen/js.ungram
MichaReiser: I'm not sure if it makes sense to add examples to all rules. Many are fairly obvious from the grammar. It's `type[]`

crates/rslint_parser/src/syntax/typescript.rs
MichaReiser: Could make sense to split it as part of it's own PR

xunilrj: Agreed. I was actually planning on doing this already on my PR.

ematipico: > Could make sense to split it as part of it's own PR

Starting here might help to understand what actually is new and what actually has been modified/refactored. Unless everything was already there 🤔

Anyway, not a blocker

xtask/codegen/js.ungram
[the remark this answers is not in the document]
ematipico: Sure, no problem

ematipico:
I was wondering if we can split this PR in two/three parts? Maybe one with just the new grammar (and auto generated code), one with the new implementation?

crates/rslint_parser/src/syntax/typescript.rs
MichaReiser: That's easy. I deleted all existing type parsing and replaced it with a new one. So all is new :D #

MichaReiser:
> I was wondering if we can split this PR in two/three parts? Maybe one with just the new grammar (and auto generated code), one with the new implementation?

I don't think that splitting it that way adds a lot of value

* Generated code is generated. No need to review it
* The grammar and parser logic go hand in hand or we need to keep the old grammar around as well, which makes the changes more difficult. 

The only option I could see is to split by different parsing rules but the parsing rule code doesn't add anything fancy and are heavily inspired by the TypeScript parser. 

That's why I would focus the review on the grammar changes, as you already did!

crates/rslint_parser/src/syntax/typescript.rs
MichaReiser: To avoid that the parser repetitively calls into the `IntersectionOrUnionType` methods that match on self. There are only two call-sites. So inlining should be fine.

crates/rslint_parser/src/syntax/typescript.rs
MichaReiser: Moved it out into its own `types.rs`

MichaReiser:
@ematipico I addressed your review feedback and wrote additional documentation.

ematipico (approved):

PR #2005, opened 2022-01-26
chore: change vscode extension name and version

yassere:
## Summary

When I added a new extension locally for `rome_lsp`, I used the `rome-dev` name and had the word "Dev" all over the place so that I wouldn't mix it up with the published version. This changes the name to match what's on the marketplace and updates the version to `"0.0.4"`.

Although I noticed that the extension name is `vscode-rome` in the `archived-js` branch and I don't remember why that's the case or if we should do the same thing here. @sebmck?

@leops Also, feel free to close this PR if its easier to incorporate these changes into your #1995 PR.

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -1,9 +1,9 @@
 {
-	"name": "rome-dev",
+	"name": "rome",
 	"publisher": "rome",
-	"displayName": "Rome Dev",
-	"description": "Rome LSP Dev Client",
-	"version": "0.0.1",
+	"displayName": "Rome",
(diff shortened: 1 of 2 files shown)


ematipico (approved):

yassere:
@leops Would merging this PR trigger any new workflows since it bumps the extension version number? I just want to make sure this doesn't attempt to publish anything.

leops:
It should trigger the release workflow (which I see as a good occasion to test that the trigger works correctly), but it won't deploy anything since the deployment environment and associated marketplace token have not been created yet

Issue #2002, opened 2022-01-25
Create helpers for `Formatter`


ematipico:
From #1726 

> - A) Implement extension methods on `SyntaxResult` , `Option`, and `AstNode`/ `SyntaxToken` to format the element (to reduce the number of try operators). You can find some ideas [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3d25881c7e0eb2be701d0fc418d99242) but please spend some time to think about alternatives (for example, should we add helpers to the `FormatResult` type?).
>     - **B) Provide helpers to format optional tokens, for example, `formatter.format_optional_token(token) -> FormatElement` and overload `format_token` to accept an `Option`: `format_token<T: Into<Option<Token>>`? Or implement the `format` methods as extension methods?**
>         
>         ```rust
>         formatter.format_token(stmt.semicolon_token()).unwrap_or_else(|| token(";"));
>         formatter.format_optional_token(element.comma_token());
>         ```
>         
>         Implemented as part of #1916: `format_token` is generic over optional tokens, can take either a `SyntaxToken` or `Option<SyntaxToken>` and return a `FormatElement` or `Option<FormatElement>` respectively
>

PR #2012, opened 2022-01-26
feat(rome_formatter): traits for tokens

ematipico:
## Summary

Part of #2002 

This PR propose the usage of traits to format tokens and reducing the use of the try operator. The try operator has been replaced with the use of `.expect()` inside the traits  themselves.

One thing to notice is that these traits now return `FormatElement`. I thought that there was not point into returning `FormatResult<FormatElement>`, otherwise we need to keep using the try operator somewhere else.

I didn't want to put the new traits inside `formatter.rs` because they are not related.

New APIs for tokens/nodes:
- `.format(formatter)`
- `.format_with(formatter, |node_or_token| { ... })`

New APIs for optional tokens/nodes:
- `.format_or_empty()`
- `.format_or`
- `.format_with_or_empty`
- `.format_with_or` 

Do you think I should rename `or`  in `or_else`? 

The idea is to shift also calls like this:
- from `formatter.format_node(&node)` to `node.format(formatter)`
- from `formatter.format_token(&token)` to `token.format(formatter)`

This would allow avoid to know in advance if we're formatting a token or a node. Although we would need to import the trait needed for that particular case. 

## Test Plan

Current CI, no tests should break,

### DIFF
--- a/crates/rome_formatter/src/formatter_traits.rs
+++ b/crates/rome_formatter/src/formatter_traits.rs
@@ -19,7 +19,9 @@ pub trait FormatOptionalTokenAndNode {
     /// let result = token.format_or_empty(&formatter);
     ///
     /// assert_eq!(Ok(empty_element()), result)
-    fn format_or_empty(&self, formatter: &Formatter) -> FormatResult<FormatElement>;
+    fn format_or_empty(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        self.format_or(formatter, empty_element)
+    }
 
     /// This function tries to format an optional [token](rslint_parser::SyntaxToken). If the token doesn't exist,
     /// an [empty token](crate::FormatElement::Empty) is created. If exists, the utility
@@ -58,13 +60,33 @@ pub trait FormatOptionalTokenAndNode {
         with: With,
     ) -> FormatResult<FormatElement>
     where
-        With: FnOnce(FormatElement) -> FormatElement;
+        With: FnOnce(FormatElement) -> FormatElement,
+    {
+        self.format_with_or(formatter, with, empty_element)
+    }
 
     /// This function tries to format an optional [token](rslint_parser::SyntaxToken) as is. If the token doesn't exist,
     /// it calls the passed closure, which has to return a [create::FormatElement]
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// use rome_formatter::{Formatter, token};
+    /// use rslint_parser::{SyntaxToken};
+    /// use rome_formatter::formatter_traits::{FormatOptionalTokenAndNode};
+    ///
+    /// let formatter = Formatter::default();
+    /// let empty_token: Option<SyntaxToken> = None;
+    ///
+    /// let result = empty_token.format_or(&formatter, || token(" other result"));
+    ///
+    /// assert_eq!(Ok(token(" other result")), result);
     fn format_or<Or>(&self, formatter: &Formatter, op: Or) -> FormatResult<FormatElement>
     where
-        Or: FnOnce() -> FormatElement;
+        Or: FnOnce() -> FormatElement,
+    {
+        self.format_with_or(formatter, |token| token, op)
+    }
 
     /// If the token/node exists, it will call the first closure which will accept formatted element.
     ///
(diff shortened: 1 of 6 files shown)


crates/rome_formatter/src/formatter_traits.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn format(&self, formatter: &Formatter) -> FormatElement;
+}
+
+impl FormatOptionalToken for Option<SyntaxToken> {
+    fn format_or_empty(&self, formatter: &Formatter) -> FormatElement {
+        match self {
+            None => empty_element(),
+            Some(token) => formatter
+                .format_token(token)
+                .expect("Can't format the token"),
```
MichaReiser, line 52: This shouldn't panic. We should return a `FormatResult` here. Same for the other methods.

crates/rome_formatter/src/formatter_traits.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// ## Panics
+    ///
+    /// It panics if the formatting fails.
+    fn format_or<Op>(&self, formatter: &Formatter, op: Op) -> FormatElement
+    where
+        Op: FnOnce() -> FormatElement;
+}
+
+/// Utility trait used to format simple tokens
+pub trait FormatToken {
```
MichaReiser, line 37: What is this used for?

crates/rome_formatter/src/formatter_traits.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn format_or_empty(&self, formatter: &Formatter) -> FormatElement {
+        match self {
+            None => empty_element(),
+            Some(token) => formatter
+                .format_token(token)
+                .expect("Can't format the token"),
+        }
+    }
+
+    fn format_with_empty<With>(&self, formatter: &Formatter, with: With) -> FormatElement
```
MichaReiser, line 56: Nit: `format_with_or_empty`

crates/rome_formatter/src/formatter_traits.rs
```diff
@@ -0,0 +1,90 @@
+use crate::{empty_element, FormatElement, Formatter};
+use rslint_parser::{SyntaxResult, SyntaxToken};
+
+/// Utility trait used to simplify the formatting of optional tokens
+pub trait FormatOptionalToken {
```
leops, line 5: Let's remove the `FormattableToken` trait in `formatter.rs` since they are both implementing the same thing

crates/rome_formatter/src/formatter_traits.rs
leops: In practice this will never panic because formatting a token can never return an error (currently the only errors that can be returned by the formatter are SyntaxErrors propagated from the parser), but I agree the method should either propagate the error or we should modify `format_token` to have it return a `FormatElement` instead of a `Result`

crates/rome_formatter/src/formatter_traits.rs
ematipico: I completely missed that trait. Thank you!

crates/rome_formatter/src/formatter_traits.rs
ematipico: > This shouldn't panic. We should return a FormatResult here. Same for the other methods.

If we returned `FormatResult`, the existence of these helpers would not help at all. What we want to try to do remove the usage of the try operators.

I agree with @leops  that `format_token` should return a `FormatElement`.

Also, why are we returning a `Result`? I mean, I think we don't need it anymore. The only reason we had `Result` is because of `SyntaxResult`, which is a safety net of missing children... BUT, we now have the syntax factory, which validates the green tree. So I know wonder, should we elide the usage of `Result`?

MichaReiser: I don't think the main goal is to remove the need for the try-operator. The main goal is to avoid the `if let Some... else`. 

Writing `if_stmt.else_clause().format_or_empty()?` Feels much better than the extended `if let Some`

>  Also, why are we returning a Result? I mean, I think we don't need it anymore. The only reason we had Result is because of SyntaxResult, which is a safety net of missing children when navigating children inside an AST... BUT, we now have the syntax factory, which validates the green tree. So I now wonder, should we elide the usage of Result?

I don't see a particular reason for keeping it for `FormatToken` but defer to you and @leops to make a decision.

crates/rome_formatter/src/formatter_traits.rs
ematipico: > Let's remove the `FormattableToken` trait in `formatter.rs` since they are both implementing the same thing

I think I will tackle this with another PR. The new traits use, internally, `format_token` and `format_node`, which rely on that said trait. 

I would like first to reach a consensus on the possible shift of how to format tokens and nodes. Later I can remove the old stuff

ematipico:
@leops  I applied your suggestions! That was great, I completely missed the whole thing 😅

leops (approved):

PR #2013, opened 2022-01-27
feat(rome_formatter): add a baseline version of range formatting

### LINKED ISSUE #2009: 📎Create an API to format a range
Essentially we want to be able to give a portion of text to the formatter and return the formatted result (either formatted or keep as is because it's a broken piece of code). Ideally, we could use the same APIs we have now.

leops:
## Summary

This is an initial implementation of the range formatting API. It works by tracking the location of tokens between the input and output of the formatter, and use that information to remap the location of the formatting range from the source document to the result and slice the resulting code accordingly. I consider this an initial baseline implementation since it still reformats the entire document: the next step would be to find the lowest common ancestor for the start and end of the formatting range and reformat only that node, with the printer initialized to the correct indentation level.

This change introduce a layout modification of the `Token` type in the formatter: I added source location information alongside the text content, and took the opportunity to change its type from a struct to an enum and allow the text content to be static string in addition to a heap allocated one. This removes the need for an allocation in the common case of dynamically constructed tokens (eg. `token(",")`) at no additional memory cost (relative to adding a `source: Option<TextRange>` field to the existing `Token` struct) and makes differentiating the tokens created from the input source (with range information) and the ones created by the formatter easier. I also included the normalization of newline characters at token creations, all the line terminators supported in ECMAScript are now represented as "\n" in tokens.

Fixes #2009

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -836,24 +836,35 @@ impl ConditionalGroupContent {
 /// See [token] for documentation
 #[derive(Debug, Eq, Clone)]
 pub enum Token {
+    /// Token constructed by the formatter from a static string
     Static { text: &'static str },
+    /// Token constructed from the input source as a dynamics
+    /// string and a range of the input source
     Dynamic { text: String, source: TextRange },
 }
 
 impl Token {
-    pub(crate) const fn new_static(text: &'static str) -> Self {
+    /// Create a token from a static string
+    const fn new_static(text: &'static str) -> Self {
         Self::Static { text }
     }
 
+    /// Create a token from a dynamic string and a range of the input source
     pub(crate) fn new_dynamic(text: &str, source: TextRange) -> Self {
+        const LINE_SEPARATOR: char = '\u{2028}';
+        const PARAGRAPH_SEPARATOR: char = '\u{2029}';
         Self::Dynamic {
+            // Normalize all line terminators to "\n" since its
+            // the only line break type supported by the printer
             text: text
                 .replace("\r\n", "\n")
-                .replace(&['\r', '\u{2028}', '\u{2029}'], "\n"),
+                .replace(&['\r', LINE_SEPARATOR, PARAGRAPH_SEPARATOR], "\n"),
             source,
         }
     }
 
+    /// Get the range of the input source covered by this token,
+    /// or None if the token was synthesized by the formatter
     pub(crate) fn source(&self) -> Option<&TextRange> {
         match self {
             Token::Static { .. } => None,
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -33,7 +33,7 @@ pub fn empty_element() -> FormatElement {
 ///   token("b"),
 /// ]);
 ///
-/// assert_eq!("a,b", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("a,b", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 /// See [soft_line_break_or_space] if you want to insert a space between the elements if the enclosing
 /// [Group] fits on a single line.
@@ -53,7 +53,7 @@ pub fn empty_element() -> FormatElement {
 ///  ..FormatOptions::default()
 /// };
 ///
-/// assert_eq!("a long word,\nso that the group doesn't fit on a single line", format_element(&elements, options).code());
+/// assert_eq!("a long word,\nso that the group doesn't fit on a single line", format_element(&elements, options).as_code());
 /// ```
 #[inline]
 pub const fn soft_line_break() -> FormatElement {
@@ -76,7 +76,7 @@ pub const fn soft_line_break() -> FormatElement {
 ///   hard_line_break()
 /// ]);
 ///
-/// assert_eq!("a,\nb\n", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("a,\nb\n", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 #[inline]
 pub const fn hard_line_break() -> FormatElement {
@@ -98,7 +98,7 @@ pub const fn hard_line_break() -> FormatElement {
 ///   empty_line()
 /// ]);
 ///
-/// assert_eq!("a,\n\nb\n\n", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("a,\n\nb\n\n", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 #[inline]
 pub const fn empty_line() -> FormatElement {
@@ -119,7 +119,7 @@ pub const fn empty_line() -> FormatElement {
 ///   token("b"),
 /// ]);
 ///
-/// assert_eq!("a, b", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("a, b", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 ///
 /// The printer breaks the lines if the enclosing [Group] doesn't fit on a single line:
@@ -137,7 +137,7 @@ pub const fn empty_line() -> FormatElement {
 ///  ..FormatOptions::default()
 /// };
 ///
-/// assert_eq!("a long word,\nso that the group doesn't fit on a single line", format_element(&elements, options).code());
+/// assert_eq!("a long word,\nso that the group doesn't fit on a single line", format_element(&elements, options).as_code());
 /// ```
 #[inline]
 pub const fn soft_line_break_or_space() -> FormatElement {
@@ -157,7 +157,7 @@ pub const fn soft_line_break_or_space() -> FormatElement {
 /// use rome_formatter::{token, format_element, FormatOptions};
 /// let elements = token("Hello World");
 ///
-/// assert_eq!("Hello World", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("Hello World", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 ///
 /// Printing a string literal as a literal requires that the string literal is properly escaped and
@@ -169,7 +169,7 @@ pub const fn soft_line_break_or_space() -> FormatElement {
 /// // the tab must be encoded as \\t to not literally print a tab character ("Hello{tab}World" vs "Hello\tWorld")
 /// let elements = token("\"Hello\\tWorld\"");
 ///
-/// assert_eq!(r#""Hello\tWorld""#, format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!(r#""Hello\tWorld""#, format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 #[inline]
 pub const fn token(text: &'static str) -> FormatElement {
@@ -189,7 +189,7 @@ pub const fn token(text: &'static str) -> FormatElement {
 ///
 /// let elements = format_elements![token("a"), line_suffix(token("c")), token("b")];
 ///
-/// assert_eq!("abc", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("abc", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 #[inline]
 pub fn line_suffix(element: impl Into<FormatElement>) -> FormatElement {
@@ -206,7 +206,7 @@ pub fn line_suffix(element: impl Into<FormatElement>) -> FormatElement {
 /// // the tab must be encoded as \\t to not literally print a tab character ("Hello{tab}World" vs "Hello\tWorld")
 /// let elements = format_elements![token("a"), space_token(), token("b")];
 ///
-/// assert_eq!("a b", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("a b", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 #[inline]
 pub const fn space_token() -> FormatElement {
@@ -221,7 +221,7 @@ pub const fn space_token() -> FormatElement {
 /// use rome_formatter::{concat_elements, FormatElement, space_token, token, format_element, FormatOptions};
 /// let expr = concat_elements(vec![token("a"), space_token(), token("+"), space_token(), token("b")]);
 ///
-/// assert_eq!("a + b", format_element(&expr, FormatOptions::default()).code())
+/// assert_eq!("a + b", format_element(&expr, FormatOptions::default()).as_code())
 /// ```
 pub fn concat_elements<I>(elements: I) -> FormatElement
 where
@@ -265,7 +265,7 @@ where
 /// let separator = concat_elements(vec![token(","), space_token()]);
 /// let elements = join_elements(separator, vec![token("1"), token("2"), token("3"), token("4")]);
 ///
-/// assert_eq!("1, 2, 3, 4", format_element(&elements, FormatOptions::default()).code());
+/// assert_eq!("1, 2, 3, 4", format_element(&elements, FormatOptions::default()).as_code());
 /// ```
 #[inline]
 pub fn join_elements<TSep, I>(separator: TSep, elements: I) -> FormatElement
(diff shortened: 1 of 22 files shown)


crates/rome_formatter/src/format_element.rs
```diff
@@ -833,21 +834,102 @@ impl ConditionalGroupContent {
 }
 
 /// See [token] for documentation
-#[derive(Debug, PartialEq, Eq, Clone)]
-pub struct Token(String);
+#[derive(Debug, Eq, Clone)]
+pub enum Token {
+    Static { text: &'static str },
+    Dynamic { text: String, source: TextRange },
```
ematipico, line 840: Can we document these new variants? It would help to understand when use one or the other

crates/rome_formatter/src/format_element.rs
```diff
@@ -833,21 +834,102 @@ impl ConditionalGroupContent {
...
-        Self(String::from(content))
+    pub(crate) const fn new_static(text: &'static str) -> Self {
+        Self::Static { text }
+    }
+
+    pub(crate) fn new_dynamic(text: &str, source: TextRange) -> Self {
+        Self::Dynamic {
+            text: text
+                .replace("\r\n", "\n")
+                .replace(&['\r', '\u{2028}', '\u{2029}'], "\n"),
```
ematipico, line 852: Can we add a comment explaining why we are doing this? Also, can we assign `\u{2028}` and `\u{2029}` to two variables? The reason is mostly because with variables we can assign a name and know what chars they are.

crates/rome_formatter/src/format_element.rs
```diff
@@ -833,21 +834,102 @@ impl ConditionalGroupContent {
...
+pub enum Token {
+    Static { text: &'static str },
+    Dynamic { text: String, source: TextRange },
+}
 
 impl Token {
-    pub fn new(content: &str) -> Self {
-        debug_assert!(!content.contains('\r'), "The content '{}' contains a carriage return '\\r' character but string tokens must only use line feeds '\\n' as line separator. Use '\\n' instead of '\\r' and '\\r\\n' to insert a line break in strings.", content);
-        Self(String::from(content))
+    pub(crate) const fn new_static(text: &'static str) -> Self {
```
ematipico, line 844: It would be great we can have some comment docs for each function

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
-        Self { code }
+    fn new(code: String, range: Option<TextRange>, sourcemap: Vec<SourceMarker>) -> Self {
+        Self {
+            code,
+            range,
+            sourcemap,
+        }
+    }
+
+    fn empty() -> Self {
```
ematipico, line 182: maybe `clear` is a better fit for Rust? (`Vec` uses `clear` too)

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+        }
+    }
+
+    /// Range of the input source file covered by this formatted code,
+    /// or None if the entire file is covered in this instance
+    pub fn range(&self) -> Option<TextRange> {
+        self.range
+    }
+
+    pub fn sourcemap(&self) -> &[SourceMarker] {
```
ematipico, line 196: Some documentation? I know it's understandable but it's always good to have some doc

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+        &self.sourcemap
     }
 
     pub fn code(&self) -> &String {
         &self.code
     }
+
+    pub fn into_code(self) -> String {
+        self.code
+    }
```
ematipico, line 206: Probably we want to add some doc. It would help to understand when to use one or the other

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
 /// Formats a JavaScript (and its super languages) file based on its features.
 ///
 /// It returns a [Formatted] result, which the user can use to override a file.
 pub fn format(options: FormatOptions, syntax: &SyntaxNode) -> FormatResult<Formatted> {
-    Formatter::new(options).format_root(syntax)
+    let element = Formatter::new(options).format_root(syntax)?;
+    Ok(Printer::new(options).print(&element))
+}
+
+/// Formats a range withing a JavaScript file
```
ematipico, line 217:
```suggestion
/// Formats a range withing a file, supported by Rome
```

I presume this API will be generic and used to format other files

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+
+        if start_token.is_some() && end_token.is_some() {
+            break;
+        }
+    }
+
+    // If not token were found, this means the input node was empty
+    // or the formatting function was not passed the root SyntaxNode
+    // for the file: default to the first and last token in the root
+    // node or exit early with an empty result if no such token exists
```
ematipico, line 249:
```suggestion
    // If no tokens were found, this means the input node was empty
    // or the formatting function was not passed the root [SyntaxNode]
    // of the file: default to the first and last tokens in the root
    // node, or exit early with an empty result if these tokens don't exist
```

This `or the formatting function was not passed the root [SyntaxNode] of the file` is difficult to understand. Can we reword it?

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+        .peekable();
+
+    let mut common_root = None;
+    while let Some((lhs, rhs)) = iter.peek() {
+        if lhs != rhs {
+            break;
+        }
+
+        common_root = iter.next().map(|(node, _)| node);
+    }
```
ematipico, line 298: What are we doing here? 🤔 
Can we add some comment please?

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+    // Logically this should always return at least the root node,
+    // fallback to said node just in case
+    let common_root = common_root.as_ref().unwrap_or(root);
+
+    // Determine the initial indentation level for the printer by inspecting the trivias
+    // of each token from the first token of the common root towards the start of the file
+    let tokens = std::iter::successors(common_root.first_token(), |token| token.prev_token());
+
+    // From the iterator on tokens, build an iterator on trivia pieces (once again the iterator is
+    // reversed, starting from the last trailing trivia towards the first leading trivia)
```
ematipico, line 309:
```suggestion
    // From the iterator of tokens, build an iterator of trivia pieces (once again the iterator is
    // reversed, starting from the last trailing trivia towards the first leading trivia)
```

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+        common_root = iter.next().map(|(node, _)| node);
+    }
+
+    // Logically this should always return at least the root node,
+    // fallback to said node just in case
+    let common_root = common_root.as_ref().unwrap_or(root);
+
+    // Determine the initial indentation level for the printer by inspecting the trivias
+    // of each token from the first token of the common root towards the start of the file
+    let tokens = std::iter::successors(common_root.first_token(), |token| token.prev_token());
```
ematipico, line 306: I think it's worth adding that this is not a bulletproof heuristic, because trivias might not have the same level of indentation of the rest of the code. Maybe as comment of the generic API, consumers should be aware of it.

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+
+    // From the iterator on tokens, build an iterator on trivia pieces (once again the iterator is
+    // reversed, starting from the last trailing trivia towards the first leading trivia)
+    let trivias = tokens.flat_map(|token| {
+        // We need to build up the trivia pieces into a vector as SyntaxTriviaPieceIterator does
+        // not support reverse iteration, but we're only interested in newline and whitespace
```
ematipico, line 312: We do now support reverse iteration, I guess :)

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+        None => 0,
+    };
+
+    let element = Formatter::new(options).format_root(common_root)?;
+    let formatted = Printer::new(options).print_with_indent(&element, initial_indent);
+
+    // This finds the closests marker to the beginning of the source
+    // starting before or at said starting point, and the closest
+    // marker to the end of the source range starting after or at
+    // said ending point respectively
```
ematipico, line 348:
```suggestion
    // This finds the closest marker to the beginning of the source
    // starting before or at said starting point, and the closest
    // marker to the end of the source range starting after or at
    // said ending point respectively
```

crates/rome_formatter/src/lib.rs
```diff
@@ -152,26 +154,254 @@ impl Default for FormatOptions {
...
+                }
+                None => Some((marker, end_dist)),
+            }
+        }
+    }
+
+    // If no start or end were found this means the edge of the formatting
+    // range was near the edge of the input and no marker was emitted before
+    // the start (or after the end) of the formatting range: in this case
+    // the start/end marker default to the start/end of the input
```
ematipico, line 383:
```suggestion
    // If no start or end were found, this means that the edge of the formatting
    // range was near the edge of the input, and no markers ware emitted before
    // the start (or after the end) of the formatting range: in this case
    // the start/end marker default to the start/end of the input
```

crates/rome_formatter/src/lib.rs
```diff
@@ -194,3 +424,62 @@ pub fn format_file_and_save(rome_path: &mut RomePath, options: FormatOptions, ap
...
+while(
+    true
+) {
+    function func() {
+    func(     /* comment */
+    );
+    
+    let array =
+        [ 1
+    , 2];
```
ematipico, line 449: We should test formatting a range that is at a different indentation level compared to the rest of the code

crates/rome_formatter/src/printer.rs
```diff
@@ -97,10 +99,18 @@ impl<'a> Printer<'a> {
     }
 
     /// Prints the passed in element as well as all its content
-    pub fn print(mut self, element: &'a FormatElement) -> Formatted {
+    pub fn print(self, element: &'a FormatElement) -> Formatted {
+        self.print_with_indent(element, 0)
+    }
+
+    pub(crate) fn print_with_indent(
```
ematipico, line 106: Small documentation?

crates/rome_formatter/src/lib.rs
ematipico: I guess we should also document that it returns something, which is a different behaviour from a normal `empty` function. Or maybe, even better, rename the function.

crates/rome_formatter/src/lib.rs
ematipico: I think we should also mention in which case the API can fail to format some code.

crates/rome_formatter/src/lib.rs
leops: In general the `clear` method mutates an object in place but here this function constructs a new empty `Formatted`. For container objects like `Vec` or `String` the `new` function is usually used for this purpose, but here that name is already used. Another alternative is implementing the `Default` trait, but it has the disadvantage of being public while may not want users of the formatter to be able to construct `Formatted` objects themselves

crates/rome_formatter/src/lib.rs
leops: This is a convention of Rust that I addressed in my followup commit: methods named `as_X` borrow `&self` and return a reference while methods called `into_X` move `self` and return a value. I created an `as_code` method (and changed the return type to the more idiomatic `&str`), but I've only marked the existing `code` method as deprecated since it might be used in other PRs that use the formatter. Would it be acceptable to remove it ?

crates/rome_formatter/src/lib.rs
ematipico: Oh sorry, I didn't see the `Self`.  What about `with_empty`? It seems to be a defined pattern. Or `new_empty`?

crates/rome_formatter/src/lib.rs
ematipico: Yeah I think your new change is actually way better! Yeah, completely agree that we can remove it!

crates/rome_formatter/src/lib.rs
leops: As the failure behavior of the formatter might be subject to change I'm not sure if we want to document it just yet. It's something that could go into the crate-level documentation since it's the same for all entry points of the formatter: currently formatting never directly fails, but upon encountering missing or unknown nodes it may print the input tokens verbatim instead of reformatting them

crates/rome_formatter/src/lib.rs
leops: I think `with` tends to imply a builder pattern with a signature like `with_X(self) -> Self`, but `new_empty` is alright

crates/rome_formatter/src/format_element.rs
```diff
@@ -833,21 +834,113 @@ impl ConditionalGroupContent {
...
 /// See [token] for documentation
-#[derive(Debug, PartialEq, Eq, Clone)]
-pub struct Token(String);
+#[derive(Debug, Eq, Clone)]
+pub enum Token {
+    /// Token constructed by the formatter from a static string
+    Static { text: &'static str },
+    /// Token constructed from the input source as a dynamics
+    /// string and a range of the input source
+    Dynamic { text: String, source: TextRange },
```
MichaReiser, line 843: Would it make sense to pass in a `SyntaxToken`?

crates/rome_formatter/src/format_element.rs
```diff
@@ -833,21 +834,113 @@ impl ConditionalGroupContent {
...
+    /// Create a token from a dynamic string and a range of the input source
+    pub(crate) fn new_dynamic(text: &str, source: TextRange) -> Self {
+        const LINE_SEPARATOR: char = '\u{2028}';
+        const PARAGRAPH_SEPARATOR: char = '\u{2029}';
+        Self::Dynamic {
+            // Normalize all line terminators to "\n" since its
+            // the only line break type supported by the printer
+            text: text
+                .replace("\r\n", "\n")
+                .replace(&['\r', LINE_SEPARATOR, PARAGRAPH_SEPARATOR], "\n"),
```
MichaReiser, line 861: I would prefer if we do this inside of the `format_verbatim` which should be the only place where we copy over the line breaks. Wdyt?

crates/rome_lsp/src/server.rs
```diff
@@ -251,6 +252,98 @@ impl LanguageServer for LSPServer {
...
+                indent_style: if params.options.insert_spaces {
+                    IndentStyle::Space(params.options.tab_size as u8)
+                } else {
+                    IndentStyle::Tab
+                },
+                range: params.range,
+            })
+        });
+
+        // TODO: Clean up this error handling
```
MichaReiser, line 283: TODO?

MichaReiser (commented):
Neat!

crates/rome_formatter/src/format_element.rs
leops: We also need to copy line breaks when printing template literals, although in that case we shouldn't normalize them but at the moment the printer can't handle that

crates/rome_formatter/src/format_element.rs
leops: I'm looking into it, and I can add an additional `SyntaxToken` variant to the `Token` enum however the `Dynamic` variant will still be needed for comments, "normalized" string literals where the quotes have been replaced, and verbatim tokens that must not be trimmed unlike all other tokens

crates/rome_lsp/src/server.rs
leops: This is part of the ongoing cleanup of the LSP, not related to formatting

ematipico (approved):

PR #2014, opened 2022-01-27
chore(git): add .vs to the list of ignored IDE directories

leops:
## Summary

Visual Studio creates a .vs directory when opening the repo with it (for instance to run a profiling session), but much like other editors we don't want that to be committed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ Profile-*.json
 # Editor configuration
 .vscode
 .idea
+.vs
 # Release artifacts
 **/target
 # Ignore test files that contributors could create locally


MichaReiser:
I can see the motivation behind this but our contribution guideline mentions that IDE specific files should be ignored in the user's global gitignore. I don't know the history behind this decision, maybe @ematipico does

https://github.com/rome/tools/blob/main/CONTRIBUTING.md#user-files

ematipico:
Initially that was the intention, but we decided to add VSCode and IntellJ files to the ignore lists because they are the most used by Rome's contributors. That helps to lower the friction. Although, we don't want to do that for **all** IDEs.

leops:
Another thing I could do is to change the `.vscode` line to `.vs*` which would cover both editors at once, but yes I agree this could create a precedent we don't want of

MichaReiser (approved):

xunilrj (approved):

Issue #1999, opened 2022-01-25
TypeScript Parser: Enum declaration


xunilrj:
part of https://github.com/rome/tools/issues/1966:

- [x] Parsing working for Typescript;
    - [x] Simple members
    - [x] Literal members
    - [x] Computed members
- [x] Nice error for Javascript;
- [x] Nice error for broken ```enums```;

PR #2015, opened 2022-01-27
parsing enums in ts

xunilrj:
see https://github.com/rome/tools/issues/1999

Before

```
│   17185   │  9753  │  5478  │  1954  │  56.75   │
```

After
```
│   17185   │  9894  │  5468  │  1823  │  57.57   │
```


xtask/codegen/js.ungram
```diff
@@ -1450,10 +1450,9 @@ TsTypeRef = name:TsEntityName type_args:TsTypeArgs
 
 
 // typescript enum
-TsEnum = 'const'? 'enum'  ident: Ident '{' members: TsEnumMemberList '}'
-
-TsEnumMemberList = TsEnumMember*
-TsEnumMember = ident: Ident '=' value:JsAnyExpression
+TsEnum = 'enum'  ident: JsName '{' members: TsEnumMemberList '}'
+TsEnumMemberList = TsEnumMember (',' TsEnumMember)* ','?
+TsEnumMember = JsName ('=' JsAnyExpression)?
```
MichaReiser, line 1455: Nit: Can we split the declaration across multiple lines as done for most other declarations?

* I would change `ident` to `name` to match the node type
* Prefix the enum member name and expression with `name` and `expression` to avoid field names like `js_name` and `js_any_expression`.
* I'm surprised that the code gen allows you to write `('=' JsAnyExpression)`. I guess, it just ignores the fact that the optionality is for both children. You need to move the whole sub-expression into its own node. Luckily, there is `JsInitializerClause` that wraps `=` and a `JsAnyExpression` (I would then name the field `initializer`)

xtask/coverage/src/typescript.rs
```diff
@@ -80,7 +83,8 @@ pub fn run_ts(
             }
             let code = code.unwrap();
             let result = std::panic::catch_unwind(|| {
-                let r = parse(&code, 0, Syntax::default().typescript());
+                let syntax = Syntax::default().typescript();
```
MichaReiser, line 86: This is fixed in #2019

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -295,59 +295,35 @@ fn type_member_semi(p: &mut Parser) {
...
-        let err_occured = if !p.at_ts(token_set![T![ident], T![yield], T![await]])
-            && !p.cur().is_keyword()
-            && !p.at(JS_STRING_LITERAL)
-        {
-            let err = p
-                .err_builder("expected an identifier or string for an enum variant, but found none")
-                .primary(p.cur_tok().range(), "");
+        if p.at(JsSyntaxKind::IDENT) {
+            let variant = p.start();
+            let _ = parse_name(p);
```
MichaReiser, line 316: Fun one 

```
enum Test {
    ["A"] = 3,
}
```

this is valid xD 

Either introduce a `TsAnyEnumMemberName = 	JsLiteralMemberName
	| JsComputedMemberName` or should we rename `JsAnyObjectMemberName`?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -295,59 +295,35 @@ fn type_member_semi(p: &mut Parser) {
...
 
-        let member = p.start();
-        let err_occured = if !p.at_ts(token_set![T![ident], T![yield], T![await]])
-            && !p.cur().is_keyword()
-            && !p.at(JS_STRING_LITERAL)
-        {
-            let err = p
-                .err_builder("expected an identifier or string for an enum variant, but found none")
-                .primary(p.cur_tok().range(), "");
+        if p.at(JsSyntaxKind::IDENT) {
```
MichaReiser, line 314: The `ParseSeparatedList` trait can be helpful when parsing a list of nodes. It ensures that we have proper error recovery in place.

MichaReiser (commented):
TypeScript supports `const` enums. Do you plan to tackle this as a separate PR?

```
const enum Enum {
  A = 1,
  B = A * 2,
}
```

xtask/coverage/src/typescript.rs
xunilrj: Nice!

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: Let us keep the name as it is. We can decide later how to name things that are used across languages.

xtask/codegen/js.ungram
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -295,63 +297,97 @@ fn type_member_semi(p: &mut Parser) {
...
+    }
+
+    fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+        p.at(T!['}'])
+    }
+
+    fn recover(&mut self, p: &mut Parser, parsed_element: ParsedSyntax) -> RecoveryResult {
+        parsed_element.or_recover(
+            p,
+            &ParseRecovery::new(JS_UNKNOWN_MEMBER, token_set![T![,], T!['}'], T![;], T![:]])
```
MichaReiser, line 356: Is there a valid enum syntax following a colon?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -295,63 +297,97 @@ fn type_member_semi(p: &mut Parser) {
...
-    members_list.complete(p, TS_ENUM_MEMBER_LIST);
+// test ts typescript_enum
+// enum A {}
+// enum B { a, b, c }
+// const enum C { A = 1, B = A * 2, ["A"] = 3, }
+
+// test_err ts invalid_typescript_enum
+// enum A { 1 }
+pub fn ts_enum(p: &mut Parser) -> CompletedMarker {
+    let m = p.start();
```
MichaReiser, line 383: Nit: Does the function need public visiblity?

Can we make this method return `ParsedSyntax` and follow our naming convention

```
fn parse_ts_enum(p: &mut Parser) -> ParsedSyntax {
  if !p.at(T![const]) && !p.at(T![enum]) { return Absent }
  ...
}
```

crates/rslint_parser/test_data/inline/err/invalid_typescript_enum.rast
```diff
@@ -0,0 +1,102 @@
...
+--
+error[SyntaxError]: expected a statement but instead found 'enum'
+  ┌─ invalid_typescript_enum.ts:2:1
+  │
+2 │ enum { A, B }
+  │ ^^^^ Expected a statement here
+
+--
+enum A { 1 }
+enum { A, B }
```
MichaReiser, line 102: Would be nice if this gets parsed as an enum, at least in TypeScript files.

xunilrj:
I will close this PR and create a new one.

MichaReiser:
Can you rename TsEnum to TsEnumStatement so that it matches our naming convention.

PR #2017, opened 2022-01-27
refactor(error): use thiserror to implement the Error trait on FormatError and SyntaxError

leops:
## Summary

Use the procedural `Error` macro from the `thiserror` crate to automatically derive the `std::error::Error` and `std::fmt::Display` trait for the `FormatError` and `SyntaxError` types

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1229,6 +1229,7 @@ dependencies = [
  "rome_rowan",
  "rslint_parser",
  "tests_macros",
+ "thiserror",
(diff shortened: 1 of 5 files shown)


leops:
@yassere I did not remove the unwrap in `handler.rs` as part of this since I imagine it may conflict with your ongoing refactor

yassere (approved):
Thanks! Yeah, I'll just rebase on top of this once it merges.

Issue #1996, opened 2022-01-25
☂️ Formatter stability improvements


ematipico:
### Description

The formatter is in a good shape, but it doesn't have enough stability to be used. In order to reach a point where we can have a good integration with LSP, we still need to improve it and add some more features.

## Top priority

- [x] #2009 

	Essentially we want to be able to give a portion of text to the formatter and return the formatted result (either formatted or keep as is because it's a broken piece of code). Ideally, we could use the same APIs we have now.
- [x] #2030
	
	This is something that we already have with our parser. We could do the same with our formatter. We could use the babel repository for JavaScript and the the old rome repository for typescript. Using the old rome branch can be useful to see if what we format doesn't change

- [x] #2044
- [x] #2042
- [ ] Better error handling for formatter, by using `Diagnostic`
- [x] #2006
## Nice to have

- [ ] #1997
	At the moment, every time we create new grammar, we have to manually add those nodes to our formatter suite. It would be great if we can automate that part by creating the new stuff automatically
- [x] Reach 100% implementation coverage
	- Template literals @ematipico #2021
	- `import` syntax @ematipico #1992 
	- `export` syntax 	 @ematipico #2008
	- etc.
- [x] #2002

PR #2021, opened 2022-01-28
feat(rome_formatter): template literals and static initializers 

ematipico:
## Summary

This PR is part of #1996 

- renamed nodes that belong to templates by adding `JS_` as prefix 
- emitted the new nodes
- implemented formatting for template grammar
- implemented formatting for static block initializers

## Test Plan

Created new tests and updated existing ones


crates/rome_formatter/src/ts/expressions/template_chunk_element.rs
```diff
@@ -0,0 +1,10 @@
+use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
+use rslint_parser::ast::JsTemplateChunkElement;
+
+impl ToFormatElement for JsTemplateChunkElement {
+    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        let chunk = formatter.format_token(&self.template_chunk_token()?)?;
+
+        Ok(chunk)
```
MichaReiser, line 8: Nit: Inline
```suggestion
       formatter.format_token(&self.template_chunk_token()?)
```

crates/rome_formatter/tests/specs/js/module/template/template.js.snap
```diff
@@ -0,0 +1,30 @@
...
+tag`something`
+`something ${ " hello" }`;
+
+`something ${ () => { var hey; const looooooooooong_expression = "loooooooooong_expression" }} something else ${ ehy }`;
+    `something ${ () => { var hey; const looooooooooong_expression = "loooooooooong_expression"; return hey; }} something else ${ ehy }`
+---
+# Output
+`something`;
+
+tag`something``something ${" hello"}`;
```
MichaReiser, line 19: This looks wrong... Shouldn't these be two statements?

crates/rome_formatter/tests/specs/js/module/template/template.js.snap
```diff
@@ -0,0 +1,30 @@
...
+
+`something ${() => {
```
MichaReiser, line 21: What happens if a template spawns multiple lines

```js
`test
  abcd ${input}
output
`;
```
```

File: xtask/codegen/js.ungram [hunk 4]
```diff
@@ -297,27 +297,27 @@ JsAnyExpression =
...
 	| NewTarget
-	| Template
+	| JsTemplate
 	| TsAssertion
 	| TsConstAssertion
 	| TsNonNull
 
 
-Template =
+JsTemplate =
```
MichaReiser, line 306: Thx

crates/rome_formatter/src/ts/expressions/template.rs
```diff
@@ -0,0 +1,24 @@
...
+        } else {
+            empty_element()
+        };
+        let l_tick = formatter.format_token(&self.l_tick_token()?)?;
+        let r_tick = formatter.format_token(&self.r_tick_token()?)?;
+
+        Ok(format_elements![
+            tag,
+            l_tick,
+            join_elements(empty_element(), formatter.format_nodes(self.elements())?),
```
leops, line 20: This could be a call to `concat_elements` instead

crates/rome_formatter/tests/specs/js/module/template/template.js.snap
ematipico: Nope, this is correct! I also thought it was wrong but nope..  I checked with prettier too!

Unless prettier is also wrong 🤔 I think we're safe to assume that it's correct

crates/rome_formatter/tests/specs/js/module/template/template.js.snap
ematipico: I am going to add a test case for that, thank you

crates/rome_formatter/tests/specs/js/module/template/template.js.snap
```diff
@@ -0,0 +1,41 @@
...
+	const looooooooooong_expression = "loooooooooong_expression";
+}} something else ${ehy}`;
+`something ${() => {
+	var hey;
+	const looooooooooong_expression = "loooooooooong_expression";
+	return hey;
+}} something else ${ehy}`;
+
+`test
+  abcd ${input}
```
leops, line 38: This seems like another case where we need to find the indentation level of the source and override it in the printer, currently if this expression breaks I think it will be indented at the same level as the start of the template expression

ematipico: You're right. I am going to add a test for this case. It will be wrong and we can figure out how to do that in a fashion way

leops (approved):

PR #2024, opened 2022-01-28
JS: Parse deeply-nested binary expressions

MichaReiser:
## Summary

Adding the TypeScript conformance tests revealed that our binary expression stack overflows for deeply-nested binary expressions. 

The existing implementation uses a tail-call-recursion but Rust doesn't seem to be able to optimize the call away. 

This PR rewrites the binary expression parsing to use a loop so that the binary-expression parsing recurses at most `N(operator precedence)` levels deep. The PR further replaces the `u8` values for the precedence values with an enum which simplifies reasoning. 

## Test Plan

`binderBinaryExpressionStress.ts` now runs with the default stack size configuration.


MichaReiser:
!bench

MichaReiser:
!bench

ematipico (approved):

PR #2029, opened 2022-01-31
ci: formatter benchmarks

### LINKED ISSUE #2030: Create a workflow where we format some big code bases and report its results into the CI.

ematipico:
## Summary

This PR is part of #1996 
Closes #2030 

I creates a new workflow to run the formatter against the libraries that we use to bench the parser. 

I changed the key comments to trigger the benchmarks:
- `!bench_parser` will trigger  the parser benchmarks
- `!bench_formatter` will trigger the formatter benchmarks 
- the old alias `cargo benchmark` has been removed
- new cargo alias called `cargo bench_parser`
- new cargo alias called `cargo bench_formatter`

## Test Plan

Existing CI and run locally the new command.

### DIFF
--- a/.github/workflows/bench_parser.yml
+++ b/.github/workflows/bench_parser.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   bench:
     name: Bench
-    if: github.event.issue.pull_request && contains(github.event.comment.body, '!bench')
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '!bench_parser')
     runs-on: ${{ matrix.os }}
 
     strategy:
@@ -62,7 +62,7 @@ jobs:
         id: bench_comparison
         shell: bash
         run: |
-          echo "### Bench results on ${{ matrix.os }}" > output
+          echo "### Bench results on ${{ matrix.os }} of the parser" > output
           echo "\`\`\`"  >> output
           critcmp main pr >> output
           echo "\`\`\`" >> output
--- a/.github/workflows/bench_parser.yml
+++ b/.github/workflows/bench_parser.yml
@@ -50,13 +50,13 @@ jobs:
         run: cargo build --release --locked -p xtask_bench
 
       - name: Run Bench on PR Branch
-        run: cargo benchmark --save-baseline pr
+        run: cargo bench_parser --save-baseline pr
 
       - name: Checkout Main Branch
         run: git checkout main
 
       - name: Run Bench on Main Branch
-        run: cargo benchmark --save-baseline main
+        run: cargo bench_parser --save-baseline main
 
       - name: Compare Bench Results on ${{ matrix.os }}
         id: bench_comparison
--- a/xtask/bench/src/features/formatter.rs
+++ b/xtask/bench/src/features/formatter.rs
@@ -1,6 +1,6 @@
 use crate::BenchmarkSummary;
-use rome_formatter::{format, FormatOptions, FormatResult, Formatted};
-use rslint_parser::{parse, Syntax};
+use rome_formatter::{format, FormatOptions};
+use rslint_parser::SyntaxNode;
 use std::fmt::{Display, Formatter};
 use std::time::Duration;
 
@@ -9,9 +9,9 @@ pub struct FormatterMeasurement {
     id: String,
     formatting: Duration,
 }
-pub fn benchmark_format_lib(id: &str, code: &str) -> BenchmarkSummary {
+pub fn benchmark_format_lib(id: &str, root: SyntaxNode) -> BenchmarkSummary {
     let formatter_timer = timing::start();
-    run_format(code).unwrap();
+    run_format(root);
     let formatter_duration = formatter_timer.stop();
 
     BenchmarkSummary::Formatter(FormatterMeasurement {
@@ -20,11 +20,8 @@ pub fn benchmark_format_lib(id: &str, code: &str) -> BenchmarkSummary {
     })
 }
 
-pub fn run_format(code: &str) -> FormatResult<Formatted> {
-    let syntax = Syntax::default().module();
-    let root = parse(code, 0, syntax).syntax();
-
-    format(FormatOptions::default(), &root)
+pub fn run_format(root: SyntaxNode) {
+    format(FormatOptions::default(), &root).unwrap();
 }
 
 impl FormatterMeasurement {
--- a/xtask/bench/src/features/formatter.rs
+++ b/xtask/bench/src/features/formatter.rs
@@ -1,5 +1,5 @@
 use crate::BenchmarkSummary;
-use rome_formatter::{format, FormatOptions};
+use rome_formatter::{format, FormatOptions, FormatResult, Formatted};
 use rslint_parser::{parse, Syntax};
 use std::fmt::{Display, Formatter};
 use std::time::Duration;
@@ -11,7 +11,7 @@ pub struct FormatterMeasurement {
 }
 pub fn benchmark_format_lib(id: &str, code: &str) -> BenchmarkSummary {
     let formatter_timer = timing::start();
-    run_format(code);
+    run_format(code).unwrap();
     let formatter_duration = formatter_timer.stop();
 
     BenchmarkSummary::Formatter(FormatterMeasurement {
@@ -20,11 +20,11 @@ pub fn benchmark_format_lib(id: &str, code: &str) -> BenchmarkSummary {
     })
 }
 
-pub fn run_format(code: &str) {
+pub fn run_format(code: &str) -> FormatResult<Formatted> {
     let syntax = Syntax::default().module();
     let root = parse(code, 0, syntax).syntax();
 
-    let _ = format(FormatOptions::default(), &root);
+    format(FormatOptions::default(), &root)
 }
 
 impl FormatterMeasurement {
--- a/xtask/bench/src/features/formatter.rs
+++ b/xtask/bench/src/features/formatter.rs
@@ -10,17 +10,21 @@ pub struct FormatterMeasurement {
     formatting: Duration,
 }
 pub fn benchmark_format_lib(id: &str, code: &str) -> BenchmarkSummary {
-    let syntax = Syntax::default().module();
-    let root = parse(code, 0, syntax).syntax();
-
     let formatter_timer = timing::start();
-    let _ = format(FormatOptions::default(), &root);
+    run_format(code);
     let formatter_duration = formatter_timer.stop();
 
     BenchmarkSummary::Formatter(FormatterMeasurement {
         id: id.to_string(),
         formatting: formatter_duration,
     })
+}
+
+pub fn run_format(code: &str) {
+    let syntax = Syntax::default().module();
+    let root = parse(code, 0, syntax).syntax();
+
+    let _ = format(FormatOptions::default(), &root);
 }
 
 impl FormatterMeasurement {
--- a/xtask/bench/src/features/formatter.rs
+++ b/xtask/bench/src/features/formatter.rs
@@ -1,5 +1,5 @@
 use crate::BenchmarkSummary;
-use rome_formatter::{format, FormatOptions};
+use rome_formatter::{format, FormatOptions, Formatted};
 use rslint_parser::SyntaxNode;
 use std::fmt::{Display, Formatter};
 use std::time::Duration;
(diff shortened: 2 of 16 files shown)


xtask/bench/src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+                "[{}] - Downloading [{}] to [{}]",
+                filename,
+                lib,
+                file.display()
+            );
+            match ureq::get(lib).call() {
+                Ok(response) => {
+                    let mut reader = response.into_reader();
+
+                    let _ = std::fs::remove_file(&file);
```
leops, line 34: I don't think the file needs to be removed beforehand, `File::create` should truncate it if it already exists

xtask/bench/src/utils.rs
```diff
@@ -0,0 +1,46 @@
...
+                    let mut writer = std::fs::File::create(&file).map_err(err_to_string)?;
+                    let _ = std::io::copy(&mut reader, &mut writer);
```
leops, line 36: Do we really want to ignore this error ? I haven't read the specific implementation of ureq so maybe it buffers the entire response body before returning but if not then the connection might fail as data is being streamed into the file

xtask/bench/src/main.rs
```diff
@@ -32,8 +32,15 @@ fn main() -> Result<()> {
...
     let baseline: Option<String> = args.opt_value_from_str("--save-baseline").unwrap();
-
-    run(filter, criterion, baseline);
+    let feature: Option<String> = args.opt_value_from_str("--feature").unwrap();
+
+    run(
+        filter,
+        criterion,
+        baseline,
+        FeatureToBenchmark::from_str(feature.unwrap_or_else(|| "parser".to_string()).as_str())
```
leops, line 41: This could be written more concisely as `feature.as_deref().unwrap_or("parser").parse().unwrap()`

xtask/bench/src/lib.rs
```diff
@@ -122,7 +108,11 @@ pub fn run(filter: String, criterion: bool, baseline: Option<String>) {
                     rslint_parser::parse_module(code, 0);
                 }
 
-                let result = benchmark_lib(&id, code);
+                let result = match feature {
```
leops, line 111: This PR is about CI so this could be a different task, but the `feature` parameter could also be used to select what is being benchmarked with criterion

xtask/bench/src/lib.rs
```diff
@@ -1,90 +1,76 @@
...
-            heap.curr_bytes.file_size(options::CONVENTIONAL).unwrap()
-        );
-        println!("\t\tMax Blocks: {}", heap.max_blocks);
-        println!(
-            "\t\tMax Bytes: {}",
-            heap.max_bytes.file_size(options::CONVENTIONAL).unwrap()
-        );
+impl Display for FeatureToBenchmark {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let _ = match self {
```
leops, line 36: Why not return the error instead of ignoring it ?

xtask/bench/src/lib.rs
```diff
@@ -1,90 +1,76 @@
...
+}
+
+impl FromStr for FeatureToBenchmark {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "parser" => Ok(Self::Parser),
+            "formatter" => Ok(Self::Formatter),
+            _ => Ok(Self::Parser),
```
leops, line 29: I feel this should be an error, if you pass `--feature formater` having it silently default to the parser can be confusing

.github/workflows/bench_parser.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: Swatinem/rust-cache@v1
+
+      - name: Install critcmp
+        run: cargo install critcmp
+
+      - name: Compile
+        run: cargo build --release --locked -p xtask_bench
+
+      - name: Run Bench on PR Branch
+        run: cargo benchmark --save-baseline pr
```
MichaReiser, line 53: How does it know that its benchmarking the parser. It doesn't seem that you change the `cargo benchmark` call in the `bench_formatter.yaml`

xtask/bench/src/features/formatter.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+impl Display for BenchmarkFormatterResult {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let _ = writeln!(f, "\tFormatting: {:>10?}", self.formatting);
+        let _ = writeln!(f, "\t              ----------");
+        let _ = writeln!(f, "\tTotal:        {:>10?}", self.total());
+
+        // TODO: add diagnostics once the formatter API is able to return them from a formatter error
```
MichaReiser, line 42: Nit: I would remove the comment. We don't know if we'll ever add diagnostics.

xtask/bench/src/lib.rs
MichaReiser, line 29: I would change the command to `cargo benchmark parser` and `cargo benchmark formatter` if the argument is mandatory (which I believe is a good thing).

xtask/bench/src/lib.rs
```diff
@@ -1,90 +1,76 @@
...
-        "\t\tTotal Bytes: {}",
-        (current.total_bytes - before.total_bytes)
-            .file_size(options::CONVENTIONAL)
-            .unwrap()
-    );
-
-    current
+/// If groups the summary by their category and creates a small interface
+/// where each bench result can create their summary
+pub enum BenchSummary {
```
MichaReiser, line 46: Nit: `BenchmarkSummary` to match the naming of the other structs/traits

xtask/bench/src/lib.rs
MichaReiser: We actually use `criterion` to benchmark the libs on CI.

xtask/bench/src/features/parser.rs
```diff
@@ -0,0 +1,133 @@
+use crate::BenchSummary;
+use itertools::Itertools;
+use rslint_errors::Diagnostic;
+use std::fmt::{Display, Formatter};
+use std::ops::Add;
+use std::time::Duration;
+
+#[derive(Debug, Clone)]
+pub struct BenchmarkParseResult {
```
MichaReiser, line 9: Nit: `Result` may be misleading to reader, thinking that this is a Rust `Result`. 

How about `ParserMeasurement`?

xtask/bench/src/lib.rs
ematipico: Unfortunately cargo aliases doesn't support spaces inside keys (they do but what comes after the first key is a subcommand).

I will call them `cargo bech_parser` and `cargo bench_formatter`

ematipico:
@MichaReiser @leops  I update the description of the issue and applied your suggestions

xtask/bench/src/utils.rs
leops: It should at least propagate the error like the other IO calls: `std::io::copy(&mut reader, &mut writer).map_err(err_to_string)?`, but the file could still end up being halfway written before failing. Removing the file on error would be better, something like:
```rust
if let Err(err) = std::io::copy(&mut reader, &mut writer) {
    drop(writer);
    std::fs::remove_file(&file).ok();
    return Err(err_to_string(err));
}
```

xtask/bench/src/lib.rs
MichaReiser: @ematipico I don't think your CI job shows the correct result if you don't call into another function for the criterion profiling as well because the CI uses criterion to retrieve the benchmark numbers.

ematipico: Sorry lads, but I have never used `criterion`, so it would be appreciated if you helped me to understand what I should do in order to fix the issue

ematipico: > We actually use `criterion` to benchmark the libs on CI.

Are you sure? The workflow for benchmarks doesn't pass `--criterion` flag

MichaReiser: I'm fairly certain. The CI task uses criterion compare to compare the results... Isn't criterion the default?

ematipico: > I'm fairly certain. The CI task uses criterion compare to compare the results... Isn't criterion the default?

Yes it is, just checked now! Although now I am confused by the current code... even if we use `criterion`, we still call `benchmark_lib` and we print those measures at the end. Is it intended? 🤔

leops: Right I think criterion is on by default and the CLI argument is used to turn it off with `--criterion false` (it might be more ergonomic to change it to a flag and rename it `--no-criterion`)

MichaReiser: Agree on the `no-criterion`. 

@xunilrj greps the numbers from the console. That's when it's useful to print the summary even when running with criterion. But agree, these results might be misleading. It would certainly be better if we could extract the same numbers from criterion.

leops: As for how it would be implemented, wrapping lines 97 to 104 in a similar match expression should work:

```rust
match feature {
    FeatureToBenchmark::Parser => {
        let mut group = criterion.benchmark_group("parser");
        group.throughput(criterion::Throughput::Bytes(code.len() as u64));
        group.bench_function(&id, |b| {
            b.iter(|| {
                let _ = criterion::black_box(rslint_parser::parse_module(code, 0));
            })
        });
        group.finish();
    }
    FeatureToBenchmark::Formatter => {
        let mut group = criterion.benchmark_group("formatter");
        group.throughput(criterion::Throughput::Bytes(code.len() as u64));
        let syntax = rslint_parser::parse_module(code, 0).syntax();
        group.bench_function(&id, |b| {
            b.iter(|| {
                rome_formatter::format(rome_formatter::FormatOptions::default(), &syntax)
            })
        });
        group.finish();
    }
}
```

xtask/bench/src/lib.rs
```diff
@@ -109,20 +94,38 @@ pub fn run(filter: String, criterion: bool, baseline: Option<String>) {
...
                     group.throughput(criterion::Throughput::Bytes(code.len() as u64));
                     group.bench_function(&id, |b| {
                         b.iter(|| {
-                            let _ = criterion::black_box(rslint_parser::parse_module(code, 0));
+                            match feature {
+                                FeatureToBenchmark::Parser => {
+                                    let _ = criterion::black_box(run_parse(code));
+                                }
+                                FeatureToBenchmark::Formatter => {
+                                    let _ = criterion::black_box(run_format(code).unwrap());
```
leops, line 106: There are a few things to unpack here, the first is that I don't think we need to pass the result of parsing / formatting to `black_box` as criterion will automatically call it on whatever value is returned from the iter closure.

The second thing if that I don't know the performance implication of having the match inside the benchmark function, the branch predictor of the CPU might pick up on it and the Rust compiler might even be able to hoist it but just in case, it might be better to have two calls to iter in two different branches eg. `FeatureToBenchmark::Parser => b.iter(|| run_parse(code))`

Finally `run_format` runs the parser in addition to the formatter, so this is actually benchmarking the two at once instead of just the formatter. I'm not sure if this is the intended behavior, it might be interesting to at least document it somehow.

ematipico: > Finally run_format runs the parser in addition to the formatter, so this is actually benchmarking the two at once instead of just the formatter. I'm not sure if this is the intended behavior, it might be interesting to at least document it somehow.

It seems that there's API for our case: https://docs.rs/criterion/0.3.5/criterion/struct.Bencher.html#method.iter_batched

leops: The formatter is only borrowing the SyntaxNode its formatting so there is no real reason to re-parse the code on every iteration, it can just be parsed once before calling iter and passed as a reference to `run_format`

xtask/bench/src/lib.rs
```diff
@@ -109,20 +96,49 @@ pub fn run(filter: String, criterion: bool, baseline: Option<String>) {
...
+                    let mut group = criterion.benchmark_group(feature.to_string());
                     group.throughput(criterion::Throughput::Bytes(code.len() as u64));
-                    group.bench_function(&id, |b| {
-                        b.iter(|| {
-                            let _ = criterion::black_box(rslint_parser::parse_module(code, 0));
-                        })
+                    group.bench_function(&id, |b| match feature {
+                        FeatureToBenchmark::Parser => b.iter(|| {
+                            #[allow(clippy::unit_arg)]
+                            criterion::black_box(run_parse(code));
```
leops, line 104: Clippy is right about this one though, this call to `black_box` isn't actually doing anything since `run_parse` (or `run_format`) doesn't return a value. I think these function should return the resulting SyntaxNode / Formatted, this way we could also chose whether we want to include the time it takes to deallocate these structures in the benchmark time or not (with `iter_with_large_drop`)

leops (approved):

Issue #1997, opened 2022-01-25
Automatic code generation for formatter


ematipico:
We should be able to create code automatically, every time we run the xtask `syntax`.

This task will be disruptive and will apply the following changes:
- [`cst.rs`](https://github.com/rome/tools/blob/main/crates/rome_formatter/src/cst.rs) will be automatically generated, the whole file. If a new node is created, this file should have a new node created (and imported)
- the folder structure will change drastically and will have to follow the naming convention of the nodes. One file will have to implement one single AST node. No more, no less. This will be the golden rule that would allow us to generate nodes automatically.

This is my proposal:

### Main folders

The root folder will be the language. For now we have all `Js*` nodes, they will live inside a `js/` folder. Once we will introduce new languages, new folders will pop up:
- `ts/` folder for `Ts*` nodes;
- `jsx/` folder for `Jsx*` nodes;
- etc.

### Sub folders

All our nodes contains a "concept" in their name: statement, expression, object member, binding, etc. We could use those concepts as sub folders of the main folders. If a AST node name contains `Expression`, it will be placed inside the `expression/` folder, if  it contains `ObjectMember`, it will be placed inside `object_member/` folder. And so on. 

### `Any*` element,  `JsAnyArrayElement`
1. the file will be inside the `js`folder
2. the file will be inside `js/any/` folder because it contains the `Any` keyword
3. the file name will be `array_element.rs` and it will contain the following code:

```rs
use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
use rslint_parser::ast::JsAnyArrayElement;

impl ToFormatElement for JsAnyArrayElement {
    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
        match self {
            JsAnyArrayElement::JsSpread(element) => element.to_format_element(formatter),
            JsAnyArrayElement::JsAnyExpression(element) => element.to_format_element(formatter),
            JsAnyArrayElement::JsArrayHole(element) => element.to_format_element(formatter),
        }
    }
}
```

### Rest of elements

For the name of the file, I am a bit conflicted. I would personally leave the whole name. 

- `JsDoWhileStatement` => `js/statement/do_while_statement.rs`
- `JsReturnStatement` => `js/statement/return_statement.rs`
- `JsComputedMemberExpression` => `js/expression/computed_member_expression.rs`

Ideally we could remove the "concept" from their name, but I'm not sure if it could create more issues. 

The generated file will contain

```rs

[...]

PR #2031, opened 2022-01-31
feat(rome_formatter): add a codegen xtask for the formatter

leops:
## Summary

This implements a proposal from @ematipico in #1997 to automatically generate a default implementation of `ToFormatElement` for nodes that are currently missing one. The most important idea is to have each the formatter for each node in a different file with a stable path generated from the name of said node, then if that file is missing the codegen task will simply create it (as well as update any mod.rs file that may need to be changed for the new module to be imported in the project). I came up with an initial version of this node name to path conversion which breaks up the nodes into the following directories:

- any (unions)
- list
- unknown
- assignment
- declaration
- expression
- import
- name
- pattern
- statement

Unfortunately not everything falls neatly into place with this classification and it ended up being implemented as huge match expression to handle a bunch of edge case, so I wonder if there is another possible system that would align naturally with how the nodes are currently named.

Additionally, for the initial implementation I included additional bits of logic to migrate the existing impl blocks from where they currently leave to their new place in the module tree, all of that code is marked as deprecated since I intend to remove it before merging this


ematipico:
What's a naming convention change that would simplify the code generation of the formatter?

xtask/codegen/src/formatter.rs
```diff
@@ -0,0 +1,505 @@
...
+            for import in imports {
+                // Clippy complains about child modules having the same
+                // names as their parent, eg. js/name/name.rs
+                if import == stem {
+                    content.push_str("#[allow(clippy::module_inception)]\n");
+                }
+
+                content.push_str("mod ");
+                content.push_str(&import);
+                content.push_str(";\n");
```
MichaReiser: What's your reasoning behind using plain string manipulations over `quote!`?

leops: There isn't any particularly strong reason, these files are just a flat list of module imports so I figured there wasn't any reason to use quote either

xtask/codegen/src/formatter.rs
```diff
@@ -0,0 +1,550 @@
...
+}
+
+#[deprecated]
+struct NodeIndexEntry {
+    use_items: Vec<String>,
+    impl_item: String,
+    file: FileSet,
+}
+
+#[deprecated]
```
ematipico: Can we add some text that says what's the new way? If this is deprecated, there must be a new way to do stuff now

ematipico (approved):

ematipico:
I'd say we can merged it ASAP and address various comments later

xtask/codegen/src/formatter.rs
leops: This is the code that move the existing implementations of `ToFormatElement` so there isn't really a new way of doing it, it's deprecated because it not needed anymore once everything has been moved but I'm still pushing it now for it to be removed later just in case it's needed to rebase an existing PR

Issue #2027, opened 2022-01-31
📎 TypeScript: Expression Parsing


MichaReiser:
### Description

See [expression extensions](https://gray-sea-07a57a41e.azurestaticapps.net/#expressions)

* [x] type arguments #2032 
* [x] Function Expression / Declaration (type parameters, parameter). Also implement for methods
* [x] Non-null-assertion for member expressions #2037 
* [x] `<Type> UnaryExpr` cast operator #2033
* [x] `as` cast operator #2033

PR #2032, opened 2022-02-01
feat(rslint_parser): TS Type Arguments

MichaReiser:
## Summary

Re-implements parsing of type arguments in call and new expressions

```ts
call<A, B, C>()
```

This PR includes a refactoring that splits the `subscripts` parsing into different methods for member vs call subscripts. I found it rather difficult to reason about the code when all `subscripts` were in a single method. 

Part of #2027

## Test Plan

Added new tests with `type arguments`.

Regressions:
* compiler regressions: expect a type checker error. We previously failed parsing the file which is why the test suite assumed we "correctly" fail.
* conformance: contain `<number> t` casts that are currently not supported and is something I'll tackle next


xtask/codegen/js.ungram
```diff
@@ -1760,7 +1765,9 @@ TsTemplateElement =
 	'}'
 
 
-TsExprWithTypeArgs = item: TsAnyName type_params: TsTypeArgs?
+TsExprWithTypeArgs =
```
MichaReiser, line 1768: This will be changed/renamed when working on the `class` parsing.

MichaReiser:
!bench

xtask/codegen/js.ungram
```diff
@@ -1771,8 +1778,12 @@ TsQualifiedName =
...
 	right: JsName
 
-TsTypeArgs = '<' TsTypeArgList '>'
-TsTypeArgList = (TsType (',' TsType)* )
+TsTypeArguments =
+	'<'
+	TsTypeArgumentList
+	'>'
+
+TsTypeArgumentList = (TsType (',' TsType)* )
```
ematipico, line 1786: How are we going to distinguish type arguments when we define a function VS we use call it?

```ts

function a<A, B,>() {} // valid trailing comma

a<string, string,>(); // invalid trailing comma

```

MichaReiser: The grammar defines two types of generic type clauses:

* `TsTypeParameters`: Used by functions, classes, and type aliases. Introduce new type parameters for that particular type
* `TsTypeArguments`: Used when calling a generic function or instantiating a generic class. 

This is roughly the same as for:

* `JsParameters`: List of parameters that a function accepts
* `JsArguments`: List of values that are passed in a function call. 

Thus, it's possible to support trailing commas for one syntax but not the other. Let me double-check that the trailing commas are correctly handled.

ematipico (approved):

PR #2033, opened 2022-02-01
feat(rslint_parser): TS Type assertions

MichaReiser:
## Summary

Implements the parsing of `expr as type` and `<type> expr` type assertion expressions

## Test Plan

Added new tests demonstrating the parsing of the new syntax

Verified that the regressions are tests that pass because our parser doesn't emit type-checking errors. These tests previously seemed to "correctly fail" because of parsing errors in `as` and `cast` expressions.


crates/rslint_parser/src/parse.rs
```diff
@@ -94,7 +94,13 @@ impl<T: AstNode> Parse<T> {
     /// # Panics
     /// Panics if the node represented by this parse result mismatches.
     pub fn tree(&self) -> T {
-        self.try_tree().unwrap()
+        self.try_tree().unwrap_or_else(|| {
+            panic!(
+                "Expected tree to be a {} but root is:\n{:#?}",
```
MichaReiser, line 99: This took me a long time to figure out because the runner reported that the parser panics... but it did only panic after parsing when asserting the right test outcome. 

This adds a nicer error message so that running the test points you right to the problem. is.

xtask/coverage/src/reporters.rs
```diff
@@ -111,7 +111,10 @@ fn print_detailed_test_result(pb: &ProgressBar, result: &TestRunResult) {
 
     match &result.outcome {
         TestRunOutcome::Panicked(panic) => {
-            let msg = panic.downcast_ref::<String>();
+            let msg = panic
+                .downcast_ref::<String>()
+                .map(|s| s.as_str())
+                .or_else(|| panic.downcast_ref::<&'static str>().copied());
```
MichaReiser, line 117: It's a `String` if called `panic(format!(....))` but is a string slice if `unwrap()` panics.

ematipico (approved):

PR #2036, opened 2022-02-01
feat(rome_formatter): normalize line terminators in template literals

leops:
## Summary

Per the ECMAScript specification there is no semantic difference between `\r`, `\n` and `\r\n` in template literals (they all get normalized to `\n`). This means the formatter doesn't need to preserve these and can normalize them all to `\n`, which will get replaced with an OS-appropriate line ending sequence in the printer.

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -922,14 +922,7 @@ impl Token {
 
     /// Create a token from a dynamic string and a range of the input source
     pub(crate) fn new_dynamic(text: String, source: TextRange) -> Self {
-        cfg_if::cfg_if! {
-            if #[cfg(debug_assertions)] {
-                for line_break in text.matches(&['\r', LINE_SEPARATOR, PARAGRAPH_SEPARATOR]) {
-                    panic!("The content '{}' contains an unsupported {:?} line terminator character but string tokens must only use line feeds '\\n' as line separator. Use '\\n' instead of '\\r' and '\\r\\n' to insert a line break in strings.", text, line_break);
-                }
-            }
-        }
-
+        debug_assert!(!text.contains('\r'), "The content '{}' contains an unsupported '\\r' line terminator character but string tokens must only use line feeds '\\n' as line separator. Use '\\n' instead of '\\r' and '\\r\\n' to insert a line break in strings.", text);
         Self::Dynamic { text, source }
     }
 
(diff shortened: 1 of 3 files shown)


crates/rome_formatter/src/ts/expressions/template_chunk_element.rs
```diff
@@ -1,8 +1,41 @@
...
+        // Per https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-static-semantics-trv:
+        // In template literals, the '\r' and '\r\n' line terminators are normalized to '\n'
+        let chunk = self.template_chunk_token()?;
+        formatter.format_replaced(
+            &chunk,
+            FormatElement::from(Token::new_dynamic(
+                normalize_newlines(chunk.text_trimmed(), false),
+                chunk.text_trimmed_range(),
+            )),
+        )
```
MichaReiser, line 18: Nit: Should `normalize_newlines` accept a token so that it can avoid allocating a string if there are now newlines to be replaced?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1003,15 +996,19 @@ const PARAGRAPH_SEPARATOR: char = '\u{2029}';
 
 /// Normalize all line terminators in the text to "\n" since
 /// its the only line break type supported by the printer
-pub(crate) fn normalize_newlines(text: &str) -> String {
-    text.replace("\r\n", "\n")
-        .replace(&['\r', LINE_SEPARATOR, PARAGRAPH_SEPARATOR], "\n")
+pub(crate) fn normalize_newlines(text: &str, separators: bool) -> String {
```
MichaReiser, line 999: It may help explain when it makes sense to set `separators` to true.

MichaReiser (approved):

MichaReiser:
Maybe worth creating a test case for this?

leops:
I added the corresponding test case as the end of `template_chunk_element.rs`, I'm not sure it can reliably be put in a snapshot test without having to fight with git and text editors so they don't normalize the newlines

crates/rome_formatter/src/format_element.rs
leops: I would like to change it to an enum but the best name I came up with was `AllowedNewlines::LineFeedOrSeparators` which might have been too verbose

ematipico: >which might have been too verbose

Better verbose :) always!

ematipico (approved):

crates/rome_formatter/src/format_element.rs
MichaReiser: How about exposing two public methods that call the internal implementation and pass the separators instead?

leops: I ended up rewriting the `normalize_newlines` function to have it perform a single string traversal and allocation instead of two, and opted to simply pass in the array of line terminators to be replaced as the second argument (with a `LINE_TERMINATORS` constant for the common case of replacing everything with `\n`)

MichaReiser:
Thanks for adding the test case

Issue #2035, opened 2022-02-01
Re-Lexing of tokens [Must]


MichaReiser:
### Description

Lexing regular expressions and template literals follow different rules than lexing other grammars

* RegExp: Multiple otherwise valid punctuation tokens must be grouped together into a single token 
* RegExp: Requires tracking if a `/` is the division operator or the start of a regular expression
* Template Literals: Tagged template literals behave slightly different than normal template literals. 
* Backticks outside of an expression context shouldn't be parsed as template literals 

The way our parser currently handles this is that it tracks if it's in a position where an expression is allowed:

https://github.com/rome/tools/blob/64222927caa9705c5ce8762fc14b66f0436232a3/crates/rslint_lexer/src/state.rs#L33-L129

This logic is fragile and changes are hard to test and verify. Furthermore, it duplicates the parser logic that already tracks in which position expressions, or even more specific, regular expressions are valid. 

TypeScript, Babel, and Rome Classic, and SWC use a different approach. They have ways to signal to the lexer that it should re-lex the current token in a particular context:

https://github.com/microsoft/TypeScript/blob/b8f8fd7a3eda6573129f891e76444a6cb4857f68/src/compiler/scanner.ts#L2138-L2189

An example of the TypeScript parser. `rescanSlash` rescans the slash token assuming that it's a regular expression. 

Being able to re-lex from the current token may also become valuable when we start embedding different languages. Let's say we have an HTML file with script tags. We may want to parse the content of the `script` tag as JavaScript by re-lexing the current token as JavaScript (until we find a closing script tag). 

That's why I believe we should restructure our parser to lazily consume the tokens from the lexer. However, this will have implications on how we do look ahead today. 

## Tasks 

* [ ] Incremental lexing #2219 
* [ ] Replace `bump_multiple` with `re_lex` that correctly respects whitespace between tokens #2227

MichaReiser:
@xunilrj you looked into this before. What were your findings?
@NicholasLYang you seemed to be familiar with this kind of problem. Are you aware of alternative solutions?

MichaReiser:
The current approach is based on the work described in https://users.soe.ucsc.edu/~cormac/papers/dls14a.pdf

MichaReiser:
It may be possible to correctly implement the `NonNullAssertionExpression` in TypeScript by recognizing the current context as an expression context only if an expression was allowed by the previous token. 

While this is working for correct problems, it fails to stay in sync if the parser performs error recovery: 

```
[...]

PR #2037, opened 2022-02-02
feat(rslint_parser): TsNonNullAssertionExpression

MichaReiser:
## Summary

* Renames `TsNonNull` to `TsNonNullAssertionExpression` to match our naming scheme.
* Updates the lexer to disallow expressions after a `!`
* Renames `TsTypeAlias` to `TsTypeAliasStatement` to match our naming scheme
* 
## Test Plan

Added a new parser test for non-null assertions.


crates/rslint_lexer/src/state.rs
```diff
@@ -30,12 +30,13 @@ impl LexerState {
         self.prev = Some(next);
     }
 
-    fn update_expr_allowed(&mut self, next: JsSyntaxKind) -> bool {
-        if next.is_keyword() && self.prev == Some(T![.]) {
+    /// Based on https://users.soe.ucsc.edu/~cormac/papers/dls14a.pdf
+    fn update_expr_allowed(&mut self, current: JsSyntaxKind) -> bool {
```
MichaReiser, line 34: `next` felt confusing, particular in combination with `prev` (I would then expect that there are `prev`, `current`, and `next`). The passed in token will become the new `current` token, thus it makes sense to call it `current`.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -681,17 +681,28 @@ fn parse_member_expression_rest(
                 completed
             }
             T![!] if !p.has_linebreak_before_n(0) => {
-                // FIXME(RDambrosio016): we need to tell the lexer that an expression is not
```
MichaReiser, line 684: I created #2035 to track the problem.

ematipico (approved):

PR #2038, opened 2022-02-02
refactor(rome_formatter): new format API, part 2

ematipico:
## Summary

Part of #1996 

This PR refactors more code to use the new format API. 

While doing so, I encountered new edge cases, because of that I created new APIs for these cases:
- `try_format_with_or` which allow the use of the try operator inside the closures
- `try_format_with` which allow the use of the try operator inside the closures

I've also fixed some bug in the process.

## Test Plan

Existing tests should not break

### DIFF
--- a/crates/rome_formatter/src/formatter_traits.rs
+++ b/crates/rome_formatter/src/formatter_traits.rs
@@ -150,7 +150,7 @@ pub trait FormatOptionalTokenAndNode {
     ///
     /// ## Examples
     ///
-    /// ```rust,no_test
+    /// ```rust,ignore
     /// self.declaration().try_format_with_or(
     ///    formatter,
     ///    |declaration| {
@@ -178,7 +178,7 @@ pub trait FormatOptionalTokenAndNode {
         op: Or,
     ) -> FormatResult<FormatElement>
     where
-        With: FnOnce(FormatElement, &Self) -> FormatResult<FormatElement>,
+        With: FnOnce(FormatElement) -> FormatResult<FormatElement>,
         Or: FnOnce() -> FormatResult<FormatElement>;
 }
 
@@ -255,7 +255,7 @@ pub trait FormatTokenAndNode {
         with: With,
     ) -> FormatResult<FormatElement>
     where
-        With: FnOnce(FormatElement, &Self) -> FormatResult<FormatElement>;
+        With: FnOnce(FormatElement) -> FormatResult<FormatElement>;
 }
 
 impl<F: FormatTokenAndNode> FormatTokenAndNode for SyntaxResult<F> {
@@ -275,10 +275,10 @@ impl<F: FormatTokenAndNode> FormatTokenAndNode for SyntaxResult<F> {
         with: With,
     ) -> FormatResult<FormatElement>
     where
-        With: FnOnce(FormatElement, &Self) -> FormatResult<FormatElement>,
+        With: FnOnce(FormatElement) -> FormatResult<FormatElement>,
     {
         match self {
-            Ok(token) => with(token.format(formatter)?, self),
+            Ok(token) => with(token.format(formatter)?),
             Err(err) => Err(err.into()),
         }
     }
@@ -308,7 +308,7 @@ impl FormatTokenAndNode for SyntaxToken {
         with: With,
     ) -> FormatResult<FormatElement>
     where
-        With: FnOnce(FormatElement, &Self) -> FormatResult<FormatElement>,
+        With: FnOnce(FormatElement) -> FormatResult<FormatElement>,
     {
         cfg_if::cfg_if! {
             if #[cfg(debug_assertions)] {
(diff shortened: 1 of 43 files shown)


crates/rome_formatter/src/ts/statements/try_statement.rs
```diff
@@ -12,49 +12,53 @@ impl ToFormatElement for JsTryStatement {
...
-                formatter.format_node(&self.body()?)?
-            ])
-        } else {
-            Ok(format_elements![
-                self.catch_token().format(formatter)?,
-                space_token(),
-                formatter.format_node(&self.body()?)?
-            ])
-        }
+        self.declaration().try_format_with_or(
```
ematipico, line 43: This was a particular case where we need to use the try operator inside the closures.

crates/rome_formatter/src/ts/statements/variable_declaration_statement.rs
```diff
@@ -28,18 +27,27 @@ impl ToFormatElement for JsVariableDeclarations {
...
-                } else {
-                    token(",")
-                };
+                let node = element.node().format(formatter)?;
+                let separator = element.trailing_separator().try_format_with_or(
+                    formatter,
+                    |separator| {
+                        // SAFETY: it's fine to do a double unwrap because both checks on Result and Option
+                        // are already done inside the traits.
+                        let token = element.trailing_separator().unwrap().unwrap();
```
ematipico, line 36: Initially I wanted to pass the token as parameter of the closure, but  it was almost impossible to express it via typing system without incurring in some issue. Even so, passing the token would have required to use the `unwrap`, so it doesn't change the end result.

crates/rome_formatter/tests/specs/js/module/class/class.js
```diff
@@ -14,6 +14,8 @@ class Foo extends Boar {
 
 	}
 
+	* generator (){}
```
ematipico, line 17: This case was uncovered by our formatter and it was buggy. The new trait system worked quite well.

crates/rome_formatter/src/ts/statements/variable_declaration_statement.rs
leops: Maybe there's something I'm missing but this shouldn't actually work because `try_format_with_or` will mark the token as printed in the formatter, then `format_replaced` will try to insert it again and fail. I think the `index == last_index` should just return `Ok(empty_element())` and discard the separator, then the token won't be needed inside the closure.

crates/rome_formatter/src/formatter_traits.rs
```diff
@@ -91,6 +91,13 @@ pub trait FormatOptionalTokenAndNode {
         self.format_with_or(formatter, |token| token, op)
     }
 
+    fn try_format_or<Or>(&self, formatter: &Formatter, op: Or) -> FormatResult<FormatElement>
```
leops, line 94: Technically we don't need to introduce new methods to support this, it's possible to create an `IntoFormatResult` trait implemented for `FormatElement` and `FormatResult<FormatElement>` that converts `Self` into `FormatResult<FormatElement>`, and define the `FnOnce` types with a generic return type eg. `fn format_or<Or, OrRes>() where Or: FnOnce() -> OrRes, OrRes: IntoFormatResult`

MichaReiser (commented):
@ematipico the benchmark output is no longer printed as a table which makes it hard to read the results. Would you mind having a look?

ematipico:
> @ematipico the benchmark output is no longer printed as a table which makes it hard to read the results. Would you mind having a look?

It was never printed as a markdown table, but just as simple block and a mix of tabs/spaces to make the formatting decent. I can change it make a markdown table if you want

MichaReiser:
> > @ematipico the benchmark output is no longer printed as a table which makes it hard to read the results. Would you mind having a look?
> 
> It was never printed as a markdown table ([check the code](https://github.com/rome/tools/pull/2032#issuecomment-1026798140)), but just as simple block and a mix of tabs/spaces to make the formatting decent. I can change it make a markdown table if you want

You're right. It also prints somewhat misaligned locally. I wonder why that is. Maybe there's some way to get the nice formatting back on console & PR comment.

ematipico:
> > > @ematipico the benchmark output is no longer printed as a table which makes it hard to read the results. Would you mind having a look?
> > 
> > 
> > It was never printed as a markdown table ([check the code](https://github.com/rome/tools/pull/2032#issuecomment-1026798140)), but just as simple block and a mix of tabs/spaces to make the formatting decent. I can change it make a markdown table if you want
> 
> You're right. It also prints somewhat misaligned locally. I wonder why that is. Maybe there's some way to get the nice formatting back on console & PR comment.

The output comes form [critcmp](https://github.com/BurntSushi/critcmp), which is installed and run only during the CI. Unfortunately I don't see any markdown option. I guess the only solution would be to build our own.

leops (approved):

PR #2039, opened 2022-02-02
feat(rslint_parser): typescript enum parsing

xunilrj:
see: https://github.com/rome/tools/issues/1999

Parsing enum:

```ts
enum A {}
enum B { a, b, c }
const enum C { A = 1, B = A * 2, ["A"] = 3, }
```

We also try to give nice errors when possible:

```
error[SyntaxError]: `enum` statements must have a name
  ┌─ enum_decl_no_id.ts:1:1
  │
1 │ enum {A,B,C}
  │ ^^^^^

```

```
error[SyntaxError]: invalid `enum` name
  ┌─ enum_decl_no_id.ts:2:6
  │
2 │ enum 1 {A,B,C}
  │      ^

```

```
error[SyntaxError]: `enum` statements must have a name
  ┌─ enum_no_l_curly.ts:1:1
  │
1 │ enum;
  │ ^^^^
```

```
error[SyntaxError]: expected `'{'` but instead found `;`
  ┌─ enum_no_l_curly.ts:2:7
  │
2 │ enum A;
  │       ^ unexpected
```

```
error[SyntaxError]: expected an identifier, a string literal, or a computed name but instead found ';'
  ┌─ enum_no_r_curly.ts:1:7
  │
1 │ enum {;
  │       ^ Expected an identifier, a string literal, or a computed name here
```

```ENUM``` should be enabled only in Typescript:

```
error[SyntaxError]: `enums` can only be declared in TypeScript files
  ┌─ enum_in_js.js:1:1
  │
1 │ enum A {}
  │ ^^^^^^^^^
```

### CODE REVIEW

File: xtask/codegen/js.ungram [hunk 1]
```diff
@@ -1465,10 +1465,17 @@ TsTypeAlias =
...
 
 // typescript enum
-TsEnum = 'const'? 'enum'  ident: Ident '{' members: TsEnumMemberList '}'
-
-TsEnumMemberList = TsEnumMember*
-TsEnumMember = ident: Ident '=' value:JsAnyExpression
+TsEnum = 
+	'const'?
+	'enum'
+	name: JsName
```
ematipico, line 1471: It seems that it's valid to have 

```ts
enum a {}
enum a {}
```

But it's not valid 
```ts
enum a {}
type a = string;
```

The TS compiler doesn't complain about double binding, but something else. I wonder if it makes sense to map the name of the `enum` as binding or not. `JsName` might be too broad

MichaReiser: The enum `id` should be a `JsAnyBinding` because it binds a value in the current scope. For example, the following is valid 


```
enum TestEnum {
    A,
    B
}

console.log(TestEnum);
```

Meaning, enums are different from type aliases because they not only introduce a type but also a name in the current scope. And this is our definition of bindings.

xunilrj: hum... the error ```ts``` gives is about merging: https://www.typescriptlang.org/docs/handbook/declaration-merging.html
This rule is more complex than just name binding. I think this makes more sense in a semantic validation.

xunilrj: > The enum id should be a JsAnyBinding

Done.

xunilrj:
The regression seem to be error edge cases. I think it makes sense to postpone them.

xtask/codegen/js.ungram
```diff
@@ -1470,10 +1470,17 @@ TsTypeAlias =
 TsIdentifierBinding = name: 'ident'
 
 // typescript enum
-TsEnum = 'const'? 'enum'  ident: Ident '{' members: TsEnumMemberList '}'
-
-TsEnumMemberList = TsEnumMember*
-TsEnumMember = ident: Ident '=' value:JsAnyExpression
+TsEnum =
```
MichaReiser, line 1473: Can we rename the node to `TsEnumStatement`

xtask/codegen/js.ungram
```diff
@@ -1470,10 +1470,17 @@ TsTypeAlias =
...
 
 // typescript enum
-TsEnum = 'const'? 'enum'  ident: Ident '{' members: TsEnumMemberList '}'
-
-TsEnumMemberList = TsEnumMember*
-TsEnumMember = ident: Ident '=' value:JsAnyExpression
+TsEnum = 
+	'const'?
+	'enum'
+	name: JsAnyBinding
```
MichaReiser, line 1476: Can we rename the field to `id` to align it with `class`. It also makes it clearer that it doesn't allow arbitrary names, only ids

xtask/codegen/js.ungram
```diff
@@ -1470,10 +1470,17 @@ TsTypeAlias =
...
+	'{' 
+	members: TsEnumMemberList 
+	'}'
+TsEnumMemberList = (TsEnumMember (',' TsEnumMember)* ','?)
+TsEnumMember = 
+	name: JsAnyObjectMemberName
+	(initializer: JsInitializerClause)?
```
MichaReiser, line 1483: Nit: You don't need the parenthesizes here
```suggestion
	initializer: JsInitializerClause?
```

xtask/codegen/src/ast.rs
```diff
@@ -79,6 +79,7 @@ fn make_ast(grammar: &Grammar) -> AstSrc {
                 separator,
                 element_name,
             } => {
+                dbg!(&name, &element_name, &separator);
```
MichaReiser, line 82: Delete?

xtask/codegen/src/ast.rs
```diff
@@ -149,6 +150,7 @@ fn classify_node_rule(grammar: &Grammar, rule: &Rule) -> NodeRuleClassification
             // (T (',' T)* ','?)
             // (T (',' T)*)
             if let Some(comma_list) = handle_comma_list(grammar, rules.as_slice()) {
+                println!("{:?}", comma_list);
```
MichaReiser, line 153: Delete?

xtask/codegen/src/ast.rs
```diff
@@ -248,6 +251,9 @@ struct CommaList<'a> {
 // (T (',' T)*)
 fn handle_comma_list<'a>(grammar: &'a Grammar, rules: &[Rule]) -> Option<CommaList<'a>> {
     // Does it match (T * ',')?
+    // println!("{:?}", rules);
+    // println!("1: {}", matches!(rules, [Rule::Node(node), Rule::Rep(repeat), Rule::Opt(trailing_separator)]));
+    // println!("2: {}", matches!(rules, [Rule::Node(node), Rule::Rep(repeat)]));
```
MichaReiser, line 256: Delete?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -208,3 +147,20 @@ pub(crate) fn maybe_eat_incorrect_modifier(p: &mut Parser) -> Option<CompletedMa
...
+        return Absent;
+    }
+
+    let m = p.start();
+    p.bump(T![@]);
+
+    match p.cur() {
+        JsSyntaxKind::IDENT => {
+            p.bump(JsSyntaxKind::IDENT);
+            Present(m.complete(p, TS_TYPE_ASSERTION_EXPRESSION))
```
MichaReiser, line 162: My understanding is that you can have multiple decorators, pass arguments, etc. We would also need to add the decorators to all nodes that potentially support decorators (and introduce the right node type). 

I would leave this for it's own PR if we decide to support it.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -208,3 +147,20 @@ pub(crate) fn maybe_eat_incorrect_modifier(p: &mut Parser) -> Option<CompletedMa
...
+        }
+        _ => Absent,
```
MichaReiser, line 164: You aren't allowed to return `Absent` here because the function has eaten one token, the `@`. Either lookahead two tokens at the top of the function or simply complete the marker in this case. The `decorator.ident()` then returns `Err`

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
+use crate::parser::{RecoveryResult, ToDiagnostic};
+use crate::syntax::binding::parse_binding;
+use crate::syntax::class::parse_initializer_clause;
+use crate::syntax::expr::ExpressionContext;
+
+use crate::{JsSyntaxKind::*, *};
+
+pub(super) fn parse_literal_member_name(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 8: Nit: Make private?

You can simplify this method to avoid copying most of the `parse_literal_member_name` function from `object.rs`. 

```rs
if p.at(JS_NUMBER_LITERAL) {
   let err = p
                .err_builder("An enum member cannot have a numeric name")
                .primary(p.cur_tok().range(), "");
            p.error(err);
}

parse_literal_member_name(p)
```

Nit: Please rename the method. It's now clashing with the `parse_literal_member_name` from `object.rs`. For example to `parse_ts_enum_member_name`

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+                x
+            })
+        }
+        _ => parse_literal_member_name(p),
+    };
+
+    if name.is_absent() {
+        member.abandon(p);
+        return Absent;
+    }
```
MichaReiser, line 64: You aren't allowed to return `Absent` here because the function may already have parsed a decorator out which now gets attached to the parent. You need to test that neither name nor decorator are present.

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let _ = parse_initializer_clause(p, ExpressionContext::default());
```
MichaReiser, line 66: Nit
```suggestion
    parse_initializer_clause(p, ExpressionContext::default()).ok();
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let mut r#enum = member.complete(p, TS_ENUM_MEMBER);
+    match decorator {
+        Absent => Present(r#enum),
+        Present(_) => {
+            r#enum.change_to_unknown(p);
+            Present(r#enum)
+        }
+    }
```
MichaReiser, line 75: Changing the `enum` to `Unknown` isn't needed. The grammar doesn't specify a decorator field. Thus, the `JsSyntaxFactory` will mark the node as unknown because it detects that the enum has more children than it's supposed to have

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+        p.at(T!['}'])
+    }
+
+    fn recover(&mut self, p: &mut Parser, parsed_element: ParsedSyntax) -> RecoveryResult {
+        parsed_element.or_recover(
+            p,
+            &ParseRecovery::new(
+                JS_UNKNOWN_MEMBER,
+                token_set![JsSyntaxKind::IDENT, T![,], T!['}']],
```
MichaReiser, line 98: You may want to include `STMT_RECOVERY_SET` to recover from

```
enum A {
  test

function a() {
}
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+            | "object"
+            | "any"
+            | "unknown"
+            | "boolean"
+            | "bigint"
+            | "symbol"
+            | "void"
+            | "never"
+    )
+}
```
MichaReiser, line 133: Same as `is_reserved_type_name`?

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+    }
+}
+
+// test ts typescript_enum
+// enum A {}
+// enum B { a, b, c }
+// const enum C { A = 1, B = A * 2, ["A"] = 3, }
+pub fn ts_enum(p: &mut Parser) -> CompletedMarker {
```
MichaReiser, line 182: Nit: pub(crate) 

Can we rename it to `parse_ts_enum_statement` and return a `ParsedSyntax`

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+            | "unknown"
+            | "boolean"
+            | "bigint"
+            | "symbol"
+            | "void"
+            | "never"
+    )
+}
+
+fn parse_name(p: &mut Parser, enum_token_range: Range<usize>) {
```
MichaReiser, line 135: Nit: `parse_enum_id`

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+    let name = p.cur_src();
+
+    let id = if name == "{" {
+        Absent
+    } else {
+        parse_binding(p)
+    };
```
MichaReiser, line 142: This check isn't needed. `parse_binding` returns `Absent` if the parser is positioned at `{`

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+                            "`{}` cannot be used as a enum name because it is already reserved as a enum",
+                            text
+                        ))
+                    .primary(id.range(p), "");
+
+                p.error(err);
+            }
+        }
+        Absent => {
+            if p.nth_at(1, L_CURLY) {
```
MichaReiser, line 159: Please add tests for the error recovery

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+                let range = p.cur_tok().range();
+
+                let m = p.start();
+                p.bump_remap(T![ident]);
```
MichaReiser, line 163: Don't remap the whatever it is as an `ident` or a `JsIdentifierBinding`. You can create a `JsUnknownBinding` if you want to have this kind of error recovery here.

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,197 @@
...
+        parse_binding(p)
+    };
+
+    match id {
+        Present(id) => {
+            let text = p.span_text(id.range(p));
+            if is_reserved_enum_name(text) {
+                let err = p
+                    .err_builder(&format!(
+                            "`{}` cannot be used as a enum name because it is already reserved as a enum",
```
MichaReiser, line 150: already reserved as a type name?

MichaReiser (changes requested):

xtask/codegen/js.ungram
xunilrj: Done.

xtask/codegen/js.ungram
xunilrj: Done.

xtask/codegen/js.ungram
xunilrj: Done.

xtask/codegen/src/ast.rs
xunilrj: Done.

xtask/codegen/src/ast.rs
xunilrj: Done.

xtask/codegen/src/ast.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: Yes. I went with the simplest possible implementation just to start. I can remove it.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: I removed the decorator for now.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Removed the decorator.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: The problem is that ```STMT_RECOVERY_SET``` has `;`.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -140,11 +140,11 @@ fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
...
     // class implements B {}
+    // class extends B {}
 
-    let id = if p.cur_src() == "implements" {
-        Absent
-    } else {
-        parse_binding(p)
+    let id = match p.cur_src() {
+        "implements" | "extends" => Absent,
+        _ => parse_binding(p),
```
ematipico, line 147: We should be careful because when we have `implements`, what comes after can be also a `TsIdentifierBinding`.

For example, this is a valid syntax:

```ts
type A = {};
class B implements A {}
```

We should add a `TODO`, revert the logic or take this into account.

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+    match p.cur() {
+        JS_STRING_LITERAL | T![ident] => {
+            p.bump_any();
+        }
+        t if t.is_keyword() => {
+            p.bump_remap(T![ident]);
+        }
+        JS_NUMBER_LITERAL => {
+            let err = p
+                .err_builder("An enum member cannot have a numeric name")
```
ematipico, line 20: If this is a new error, should we have an error case to cover it?

ematipico:
```suggestion
                .err_builder("An `enum` member cannot have a numeric name")
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// An individual enum member
+fn parse_ts_enum_member(p: &mut Parser) -> ParsedSyntax {
+    let member = p.start();
+
+    let name = match p.cur() {
+        T!['['] => syntax::object::parse_computed_member_name(p),
+        T![#] => {
+            let err = p
+                .err_builder("An enum member cannot be private")
```
ematipico, line 41:
```suggestion
                .err_builder("An `enum` member cannot be private")
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    let _ = parse_initializer_clause(p, ExpressionContext::default());
+
+    Present(member.complete(p, TS_ENUM_MEMBER))
+}
+
+fn expected_ts_enum_member(p: &Parser, range: Range<usize>) -> Diagnostic {
+    parser::expected_any(&["identifier", "string literal", "computed name"], range).to_diagnostic(p)
+}
```
ematipico, line 64: Can we move this function inside a `ts_parse_error.rs` file?

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+        // enum {A,B,C}
+        Absent => {
+            if p.nth_at(1, L_CURLY) {
+                let range = p.cur_tok().range();
+
+                let m = p.start();
+                p.bump_remap(T![ident]);
+                let _ = m.complete(p, JS_IDENTIFIER_BINDING);
+
+                let err = p.err_builder("invalid enum name").primary(range, "");
```
ematipico, line 133:
```suggestion
                let err = p.err_builder("Invalid `enum` name").primary(range, "");
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+                p.error(err);
+            } else {
+                let err = p
+                    .err_builder("enum declarations must have a name")
```
ematipico, line 137:
```suggestion
                    .err_builder("`enum` statements must have a name")
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+    (*t == T![enum] && (is_ident1 || is_l_curly1))
+        || (*t == T![const] && p.nth_at(1, T![enum]) && (is_ident2 || is_l_curly2))
+}
+
+// test ts typescript_enum
+// enum A {}
+// enum B { a, b, c }
+// const enum C { A = 1, B = A * 2, ["A"] = 3, }
+pub(crate) fn parse_ts_enum_statement(p: &mut Parser) -> ParsedSyntax {
+    let m = p.start();
```
ematipico, line 161: Shouldn't this return `Absent` if `enum` is not found? This means that if we find

```ts
const e;
```

The parser should not advance because it's not a enum statement, and return `Absent`.

ematipico (changes requested):

ematipico:
@xunilrj the PR contains a `.txt` file that should not be there

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

xunilrj:
> @xunilrj the PR contains a `.txt` file that should not be there

Fixed.

crates/rslint_parser/src/syntax/class.rs
xunilrj: Did the "TODO".

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Makes more sense.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: I actually prefer the ``let _ =```` because it is easier to find, and it is guaranteed to not generate any code by the compiler.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Makes sense. I am now ```bump_any``` and producing a ```JS_UNKNOWN_BINDING```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+pub(crate) fn is_at_ts_enum_statement(p: &Parser, t: &JsSyntaxKind) -> bool {
+    let is_l_curly1 = p.nth_at(1, JsSyntaxKind::L_CURLY);
+    let is_l_curly2 = p.nth_at(2, JsSyntaxKind::L_CURLY);
+    let is_l_curly3 = p.nth_at(3, JsSyntaxKind::L_CURLY);
+
+    (*t == T![enum] && (is_l_curly1 || is_l_curly2))
+        || (*t == T![const] && p.nth_at(1, T![enum]) && (is_l_curly2 || is_l_curly3))
+}
```
MichaReiser, line 148: Nit: Passing in `t` (which seems to be the current token?) and using offsets is error prone. There's no guarantee that `t` is the current token. 

It may further help readability to split this into a few more if statements

```rust
if p.cur_at(T![enum]) && (is_l_curly1 || is_l_curly2) { 
  return true;
}

if p.cur_at(T![const]) && p.nth_at(T![enum]) && (is_l_curly_2 || p.nth_at(3, T!['{']))
```

Is the test for the curly needed? What if someone wrote `enum id`, shouldn't we parse this as enum too? Isn't it sufficient to test for the `enum` keyword only because neither of these two statements are valid

```
enum;
const enum;
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,176 @@
...
+// const enum C { A = 1, B = A * 2, ["A"] = 3, }
+pub(crate) fn parse_ts_enum_statement(p: &mut Parser) -> ParsedSyntax {
+    let m = p.start();
+
+    p.eat(T![const]);
+
+    let enum_token_range = p.cur_tok().range();
+    if !p.expect(T![enum]) {
+        m.abandon(p);
+        return Absent;
```
MichaReiser, line 162: Returning `Abandon` here violates the contract that a parse method doesn't eat any token when returning `Abandon`. 

The problem is, that the parse method may have eaten the `const` keyword already and now attaches it to an arbitrary parent node. 

I suggest either calling `is_at_ts_enum_statement` first or testing `p.cur_at(T![enum]) || p.nth_at(1, T![enum]))`

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+    Present(m.complete(p, JS_LITERAL_MEMBER_NAME))
+}
+
+/// An individual enum member
+fn parse_ts_enum_member(p: &mut Parser) -> ParsedSyntax {
+    let member = p.start();
+
+    let name = match p.cur() {
+        T!['['] => syntax::object::parse_computed_member_name(p),
```
MichaReiser, line 39: The regressions are all related to computed property names not being valid. But that seems to be something that applies across classes, enums, interfaces, objects... Maybe something for another PR?

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -186,13 +186,7 @@ pub(crate) fn parse_statement(p: &mut Parser, context: StatementContext) -> Pars
...
         T![with] => parse_with_statement(p),
         T![while] => parse_while_statement(p),
-        t if (t == T![enum] && is_nth_at_name(p, 1))
-            || (t == T![const] && p.nth_at(1, T![enum]) && is_nth_at_name(p, 2)) =>
-        {
-            let mut res = ts_enum(p);
-            res.err_if_not_ts(p, "enums can only be declared in TypeScript files");
-            Present(res)
-        }
+        t if is_at_ts_enum_statement(p, &t) => parse_ts_enum_statement(p),
```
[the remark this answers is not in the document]
xunilrj: hum... that is an interesting question...

The generated asm is wild, but it looks like it is first comparing itself with L_CURLY (token 6)

```rust
> cargo asm --rust rslint_parser::syntax::stmt::parse_statement
...
.LBB666_229:
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)] (C:\github\rome\tools\crates\rslint_syntax\src\generated.rs:6)
    cmp     word, ptr, [rbp, +, 638], 6
 t if is_at_ts_enum_statement(p, &t) => parse_ts_enum_statement(p),
    je      .LBB666_320
    cmp     word, ptr, [rbp, +, 424], 6
    jne     .LBB666_436
    jmp     .LBB666_320
...
```

If I help the compiler a little bit we have a comparison with token 73, which I think is ```ENUN_KW```

```rust
self.nth_tok(n).kind == kind (src\parser.rs:196)
     lea     rdx, [rcx, +, 2*rcx]
     #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)] (C:\github\rome\tools\crates\rslint_syntax\src\generated.rs:6)
     cmp     word, ptr, [r14, +, 4*rdx, +, 8], 73
t if is_at_ts_enum_statement(p, &t) => parse_ts_enum_statement(p),
    jne     .LBB666_499
    or      al, bl
    je      .LBB666_499
.LBB666_320:
    mov     rcx, r12
    call    rslint_parser::syntax::typescript::enums::parse_ts_enum_statement
    jmp     .LBB666_53
```

So I think the compiler does need a little bit of help here.
Your suggestion does make sense!

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,178 @@
...
+            let err = p
+                .err_builder("An enum member cannot have a numeric name")
+                .primary(p.cur_tok().range(), "");
+            p.error(err);
+            p.bump_any()
+        }
+        _ => {
+            m.abandon(p);
+            return Absent;
+        }
```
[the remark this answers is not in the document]
MichaReiser: You can do

```
let is_number = p.at(JS_NUMBER_LITERAL);

parse_literal_member_name(p).map(|literal| {
  if is_number {
    ... errror
  }
})
```

It's shorter and avoids a lot of repetition. But go with your preferred approach.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Yes. They also seem to be changing the compiler target ``es`` version. We may need to discuss what to do with these.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Did the changes.
I have also created some more ```test_err``` files.

MichaReiser (approved):

ematipico (approved):

PR #2040, opened 2022-02-02
feature(rslint_parser): TypeScript parameter extensions

MichaReiser:
## Summary

Implements the parsing for TypeScript's function and constructor parameter extensions:

* optional parameters: `function a(b?) {}`
* property members: `class A { constructor(private a, readonly b) {} }` 
* setters: no optional parameters: `set a(b?) {}` or `set a(b = "test") {}` 
* `this` parameter: `function a(this: string) {}`

There are multiple semantic rules that apply to parameters and I haven't found a reasonable approach that covers all (most) of them:

* optional setter params are disallowed in TS but allowed in JS
* array and object patterns aren't allowed in property members
* array and object patterns aren't allowed for optional parameters in implementations (but OK in a function declaration)
* a parameter can either have a `?` or initializer but not both
* `this` parameters aren't allowed in constructors and arrow functions

A second design goal is that TypeScript is an extension to the JS grammar so that a JS only analyzer works correctly if it ignores all TypeScript nodes. 

The different options that I explored:

* add the type annotation to binding: Quiet elegant but TypeScript's grammar specifies that `?`, `!` and the type annotation appear between the binding and the initializer. Thus, the nodes wouldn't act as an extension that just wrap some Js nodes and add some additional optional children at the end
* Define a `JsFormalParameter` and `JsFormalParameterWithDefault`. The `JsFormalParameter` has an optional `?` token but doesn't allow for an `initializer`. The `JsFormalParameterWithDefault` doesn't allow the `?` token but has an initializer. This prevents an AST transform from adding an initializer (or the `?` token) to a node that already has the `?` set. However, it felt overly complicated to prevent this minor issue when the AST design doesn't prevent the other invariants. Furthermore, it adds additional complexity when handling parameters. 
* Inlining the parameter into `TsPropertyParameter`: This would allow to correctly express that property parameter only allow identifier bindings. However, it breaks with the second goal that TypeScript nodes are merely an extension that JS analyzers can safely ignore. Ignoring a `PropertyParameter` would mean that they miss an identifier binding and may then report false positives in the body of the constructor.

That's why I opted for the "simplest" solution for now, accepting the downside that the AST doesn't encode all invariants. 

part of #2046 

## Test Plan

Verified that all regressions are tests that should fail but are now passing because of fixed syntax errors and the expected errors are type checker errors. 

Verified that all fixed tests are indeed intended fixes. 

Added new test cases for parameters.


xtask/codegen/js.ungram
```diff
@@ -1345,18 +1365,32 @@ JsParameters =
...
+//  ^^^^
 JsRestParameter =
 	'...'
 	binding: JsAnyBindingPattern
+	type_annotation: TsTypeAnnotation?
+
+// (this) => {}
+//  ^^^^
+TsThisParameter =
+	'this'
```
MichaReiser, line 1392: I'm torn if this should be a binding identifier or not

Pros: 
* `JsAnyParameter` can implement a `binding` accessor that returns `SyntaxResult`. This isn't possible if `this` is a token
* `this` appears in a binding context

Cons:
* The identifier should ever only be `this`. Thus, it should be safe to encode it
* `this` isn't binding a new variable in the function scope because `this` is a valid binding in all functions. It may drip analyzers if they introduce a `this` binding for a function.

xtask/codegen/js.ungram
```diff
@@ -1345,18 +1365,32 @@ JsParameters =
...
 	initializer: JsInitializerClause?
 
+// (...a) => {}
+//  ^^^^
 JsRestParameter =
 	'...'
 	binding: JsAnyBindingPattern
+	type_annotation: TsTypeAnnotation?
+
+// (this) => {}
```
ematipico, line 1389: Can we change the example? `this` is not allowed in a arrow function (in TS),  it could be misleading.

xtask/codegen/js.ungram
ematipico: Typescript AST marks `this` [as an identifier](https://ts-ast-viewer.com/#code/GYVwdgxgLglg9mABAQwBRQBYwM4C5HZQBOMYA5gDSIBGcc+YIAttQKZECUiA3gL4BQQA).

I suspect that they do the binding check during a second pass.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -955,43 +962,84 @@ fn parse_constructor_parameter_list(p: &mut Parser) -> ParsedSyntax {
...
+                    read_only = true;
+                }
+                ModifierKind::Accessibility | ModifierKind::__LAST => {}
+                ModifierKind::Declare | ModifierKind::Static | ModifierKind::Abstract => {
+                    let name = p.span_text(range);
+                    let error = p.err_builder(
+						&format!("'{}' modifier can only appear on a class, method, or property declaration.",
+								 name)
+					)
+						.primary(range, "");
```
ematipico, line 1007: It seems there's some issue with the formatting here (I see tabs)

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1216,8 +1255,7 @@ enum ModifierKind {
     Static = 2,
     Readonly = 3,
     Abstract = 4,
-    // Update the array length in `ParsedModifiers` when adding/removing modifiers here or you'll
-    // have a bad time :D
+    __LAST = 5,
```
ematipico, line 1258: What does it represent? Does it require some doc?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1250,4 +1288,20 @@ impl ClassMemberModifiers {
...
     }
+
+    /// Iterates over the present modifiers
+    fn iter(&self) -> impl Iterator<Item = (ModifierKind, &Range<usize>)> {
+        self.modifiers
+            .iter()
+            .enumerate()
+            .filter_map(|(index, range)| {
+                if let Some(range) = range {
+                    assert!(index < ModifierKind::__LAST as usize);
```
ematipico, line 1299: What's a situation where we could incur in an error here? Is it safe to use `assert` instead of `debug_assert`?

ematipico (changes requested):
I noticed that there are error case, I think we should add error cases where we attempt to use `this` as parameter inside an arrow function.

xtask/codegen/js.ungram
MichaReiser: TypeScript uses a single `Parameter` node to represent all possible parameters. That kind of forces them to use an `Identifier`. They also don't have the concept of binding identifiers.

MichaReiser:
> I noticed that there are error case, I think we should add error cases where we attempt to use `this` as parameter inside an arrow function. It's not valid in TS.

Argh, why xD I mean... it does make sense but it adds another invariant.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: Our editorconfig isn't correct. It selects `*.rs` instead of `**.rs`.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: The `modifiers` array has a fixed-length set to the value of `ModifierKind::__LAST` (the last variant of the enum). It's, therefore, guaranteed that the index is always less than `ModifierKind::__LAST`. 

We could make this a `debug_assert` but it shouldn't matter much.

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -392,28 +391,227 @@ pub(super) fn parse_arrow_body(p: &mut Parser, mut flags: SignatureFlags) -> Par
...
+            "this parameter",
+            parameter.range(p).as_range(),
+        ));
+    }
+
+    Present(parameter)
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub(crate) enum ParameterContext {
```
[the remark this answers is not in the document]
MichaReiser: Nice catch. I'm wondering if we should split out a `JsAnyArrowFunctionParameter` union that doesn't include the `TsThisParameter` but I'm worried that this is a bit overkill and will lead to a bit of duplication.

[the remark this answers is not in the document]
ematipico: Yeah, we should draw a line for when using a new union. Honestly I don't know 😅 
It surely allows us to handle different cases nicely, but if leads to much duplication, maybe it's not worth it

MichaReiser:
I explored a few options but decided to not split the arrow and function parameters.

* Option a) Add `this parameter` as a field to `JsFunctionParameters`. The main issue is that the generated code of the `SyntaxFactory` is no longer able to verify if a node has the right amount of commas or elements. For example for `function a(this: string a)` where the comma between `string` and `a` is missing. 
* Option b): Having separate `Parameters`, `ParametersList`, and `AnyParameter` types for `functions`/methods, arrow functions, and constructors. The main issue I faced was that it would then be desired to have a `JsAnyParameters` union type but arrow functions are the weird ones when it comes to functions because they support a single parameter and we don't yet have a way to construct an `AstSeparatedList` containing a single element (without a backing syntax list).
 
That's why I would leave it as is for now. It comes at the cost that a transformer or refactor may convert a function expression to an arrow function without noticing the `TsThisParameter` which is something that was also possible in rome classic.

ematipico (approved):

PR #2041, opened 2022-02-02
refactor(rome_formatter): new format API, part 3

ematipico:
## Summary

Part of #1996 

Finished porting the formatter to use the new API.

Doing so, it also removes two APIs that are not needed anymore:
- `formatter.format_node`
- `formatter.format_token`

The logic contained inside these two APIs have been moved inside the new traits.

## Test Plan

Fixed an incorrect test. Existing tests should yield the same results.

### DIFF
--- a/crates/rome_formatter/tests/specs/js/module/export/from_clause.js.snap
+++ b/crates/rome_formatter/tests/specs/js/module/export/from_clause.js.snap
@@ -16,7 +16,7 @@ export * from "hey";
 
 export * as something_bad_will_happen from "something_bad_might_not_happen";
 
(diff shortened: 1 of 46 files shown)


crates/rome_formatter/tests/specs/js/module/export/from_clause.js.snap
```diff
@@ -16,7 +16,7 @@ export * from "hey";
 
 export * as something_bad_will_happen from "something_bad_might_not_happen";
 
-export * as something_bad_will_happen from "something_bad_might_not_happen"assert {
```
ematipico, line 19: One more bug found by the new API :)

PR #2043, opened 2022-02-02
ci: test on different OS

### LINKED ISSUE #2042: Expand testing suite by adding Windows and make sure it doesn't crash

ematipico:
## Summary

Closes #2042 

Expanded the test suite when we push changes into `main`, by adding all the OS that the gitbub actions support

## Test Plan

Existing CI


.github/workflows/pull_request_test.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  test:
+    name: test
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        include:
+          - os: windows-2022
```
MichaReiser, line 23: Rust seems pretty good at abstracting the OS. At least, we haven't had any real OS specific issues as far as I know. 
Do you think it would be sufficient to run Mac/Windows on main only and rely on Linux to check our PRs?

ematipico: Yeah, I think we can start with your suggestion. We don't have code that run on specific OS yet, so it should be fine. When we will, we could change that.

MichaReiser (approved):
Looks good. 

Would it be sufficient to only test on one version per OS?

Issue #1081, opened 2020-08-15
💭 Consider replacing RJSON with TOML or YAML


sebmck:
We could alternatively support both. I can write a parser for either pretty easily. I really despise both formats, I think they're truly awful, but others seem to like them so I'm willing to concede.

They offer the same advantages we have with RJSON, but are actual community standards with great portability and interopability. Supporting both would also allow us to better integrate into the ecosystem of other languages.

ie. In a ruby project all config would be written in YAML, so having your Rome config also in YAML would be much more comfortable. TOML in Rust etc.

ematipico:
Personally, I don't like YAML. It's very sensible to tab/spaces... let's say I haven't had a pleasant experience so far.

But objectively, it's better to support a standard format (or more known/used) because it integrates better with the current ecosystem (IDEs, existing libs, etc.)

yassere:
I'm definitely in favor of switching to support one or more community standards.

We're trying to minimize the amount of configuration that people need to do anyway, so I don't think the ergonomic benefits of a custom format outweigh the lack of interoperability.

ematipico:
At the moment we internally use RJSON for testing our lint rules. If we decided to support YAML/TOML, what would happen to these files?

I can't see them ported into a YAML file. Would we port them into a JSON file?

sebmck:
YAML wouldn't be too bad since you can do multi-line strings like:

```
example: >\n
··Several lines of text,\n
··with some "quotes" of various 'types',\n
··and also a blank line:\n
··\n
··plus another line at the end.\n
```

We could pick whatever config format we would want to use internally.

Himujjal:
I suggested keeping the RJSON format as its very intuitive and I have already written serializer and deserializer for a similar format named [EKON](https://github.com/ekon-org/ekon). TOML is good but not very intuitive.

Here's the discussion:
https://github.com/rome/tools/discussions/1410#discussion-3284874

jamiebuilds:
Just to note, it's likely that we will have to support parsing `JSONC` files, which is the format that typescript uses for `tsconfig.json`. I think that format has been really successful for TypeScript, and I'd like to suggest it as an alternative to formats like YAML or TOML.

https://github.com/Microsoft/node-jsonc-parser

jamiebuilds:
I believe we basically already support `JSONC` parsing as it's effectively a subset of JSON5

ematipico:
Closing, `main` has now full support for `TOML` as configuration file!

PR #2045, opened 2022-02-03
test: custom tests for formatter

### LINKED ISSUE #2044: Expand testing suite with different options: different indent style and indent size.

ematipico:
## Summary

Closes #2044

Updated the test suite of the formatter. Now, it's possible to create test cases with a particular set of options by creating a `options.json` file beside the test file we want to test. 

All files inside a folder that contains `options.json` will be executed with the set of options written inside the file. Meaning that if you wanted to run only one particular file with some particular options, you would need to put the file inside a separate folder (the PR has an example).

I slightly changed the output of the snapshot, that's why there are lot of changes.

## Test Plan

Updated the current snapshots and create a new case with new output and verified that it is written.


crates/rome_formatter/tests/specs/js/module/expression/import_meta_expression/import_meta_expression.js.snap
```diff
@@ -9,10 +9,26 @@ console.log(import.meta);
...
+-----
+Indent style: Tab
+Line width: 80
+
+-----
 console.log(import.meta);
 import.meta.field =
 	obj.aReallyLongVariableName.andAnotherReallyLongVariableName.andAnotherReallyLongVariableName.andAnotherReallyLongVariable;
 import.meta.aReallyLongVariableName.andAnotherReallyLongVariableName.andAnotherReallyLongVariableName.andAnotherReallyLongVariable;
+## Output 2
```
ematipico, line 24: This is the new output

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -1,9 +1,90 @@
...
 use rome_path::RomePath;
 use rslint_parser::{parse, Syntax};
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+
+#[derive(Debug, Eq, PartialEq, Clone, Copy, Deserialize, Serialize)]
+pub enum IndentStyleTest {
```
MichaReiser, line 11: The `Test` suffix is unclear to me. What's the purpose of these types? Do these represent the test configuration? Why isn't it possible to reuse the `IndentStyle` enum?

crates/rome_formatter/tests/specs/js/module/array/array_nested.js.snap
```diff
@@ -46,8 +47,14 @@ let m1 = [{ a, b }, { a }, { a }];
...
 
----
-# Output
+=============================
+# Outputs
+## Output 1
+-----
+Indent style: Tab
+Line width: 80
+
```
MichaReiser, line 56: Does it need this new line?

crates/rome_formatter/tests/spec_test.rs
ematipico: As written in the description:
> I created on purpose a new structure for options.json because I didn't want to apply Deserialize and Serialize to FormatOptions, risking to add another dependency (serde) to the rome_formatter.

ematipico: That should work!

MichaReiser (commented):
Nice!  I like the idea and code looks good to me. Let's wait for @xunilrj or @leops to chime in to see if the `cfg` approach is the right way to go

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -1,60 +1,130 @@
...
+            let test_directory = PathBuf::from(test_directory);
+            let options_path = test_directory.join("options.json");
+            if options_path.exists() {
+                #[cfg(feature = "test")]
+                {
+                    let mut options_path =
+                        RomePath::new(options_path.display().to_string().as_str());
+                    // SAFETY: we checked its existence already, we assume we have rights to read it
+                    let options: Options =
+                        serde_json::from_str(options_path.get_buffer_from_file().as_str()).unwrap();
```
leops, line 112: Any thoughts on using YAML instead of JSON for these ? They could be re-serialized directly in the snapshot file instead of implementing Display

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -1,60 +1,130 @@
...
-        // we ignore the error for now
-        let snapshot = format!("# Input\n{}\n---\n# Output\n{}", input, result.as_code());
-
-        insta::with_settings!({
-            prepend_module_to_snapshot => false,
-            snapshot_path => spec_input_file.parent().unwrap(),
-        }, {
-            insta::assert_snapshot!(file_name, snapshot, file_name);
-        });
+#[cfg(feature = "test")]
```
leops, line 1: It seems this will exclude all tests from compilation when the "test" feature is not self and will make `cargo test` do nothing but still pass. I think it would make more sense to not have conditionals in the tests and simply fail the build when the feature is not set to avoid accidental false positives

crates/rome_formatter/tests/spec_test.rs
ematipico: I don't think we plan to support YAML for configurations. There was a discussion some time ago: https://github.com/rome/tools/issues/1081

Ideally we should try to dogfood whatever we can. When we will create our own JSON parser, we can use that one instead.

crates/rome_formatter/tests/spec_test.rs
ematipico: Yeah, I know it's more code but I had hard time to get the code right with feature attributes; I don't even know if it is actually correct... I think I will reuse the previous solution, as you suggested it's not uncommon to do so.

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -1,9 +1,90 @@
...
+    fn from(test: SerializableIndentStyle) -> Self {
+        match test {
+            SerializableIndentStyle::Tab => IndentStyle::Tab,
+            SerializableIndentStyle::Space(s) => IndentStyle::Space(s),
+        }
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct SerializableFormatOptions {
```
leops, line 28: The "remote derive" feature of serde might be useful for this (the facade types don't have to implement `From` and `TestOptions` can deserialize to `FormatOptions` directly): https://serde.rs/remote-derive.html

ematipico: Thank you!

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -1,9 +1,77 @@
...
+}
+
+impl SnapshotContent {
+    pub fn add_output(&mut self, content: &str, options: FormatOptions) {
+        self.output.push((String::from(content), options))
+    }
+
+    pub fn set_input(&mut self, content: &str) {
+        self.input = String::from(content);
+    }
```
yassere, line 52: If this accepts `content: impl Into<String>` then you could directly pass in the `String` you get from `fs::read_to_string` rather than copying the data into a new one. It probably won't make a noticeable difference in this case, but we might as well avoid unnecessary allocation.

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -1,9 +1,77 @@
...
+}
+
+impl From<SerializableFormatOptions> for FormatOptions {
+    fn from(test: SerializableFormatOptions) -> Self {
+        Self {
+            indent_style: test.indent_style,
+            line_width: test.line_width,
+        }
+    }
+}
```
yassere, line 37: This shouldn't be necessary if you're using the remote derive feature from serde.

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -35,26 +103,48 @@ pub fn run(spec_input_file: &str, _: &str, file_type: &str) {
...
+                // SAFETY: we checked its existence already, we assume we have rights to read it
+                let input = options_path.get_buffer_from_file();
+                let mut de = serde_json::Deserializer::from_str(input.as_str());
+
+                let options: FormatOptions =
+                    SerializableFormatOptions::deserialize(&mut de).unwrap();
+
+                let copy_options = options;
+                let formatted_result = format(options, &root).unwrap();
+                snapshot_content.add_output(formatted_result.as_code(), copy_options);
```
yassere, line 139: What's the reason for creating a `copy_options` variable here? You should be able to just pass `options` directly for both function calls.

CONTRIBUTING.md
```diff
@@ -226,3 +230,19 @@ cargo watch -i '*.new' -x 'test -p rome_formatter formatter'
...
+
+1. create a folder called `arrow_with_spaces/` and move the JS file there;
+2. then create a file called `options.json`
+3. The content would be something like:
+    ```json
+    {
+        "line_width": 120,
+        "indent_style": {"Space": 4}
+    }
+    ````
```
yassere, line 245: If there's ever a need to test multiple combinations of options for a set of input files, it would be nice if `options.json` was an array (or a `SpecTestOptions` type that includes a `Vec<FormatOptions>` and that could be extended later) rather than a single object. Then you could iterate over all the options and `snapshot_content.add_output` for each one.

Deserializing a remote derived type into a vec is a bit trickier because of [this issue](https://github.com/serde-rs/serde/issues/723). There are some workarounds listed there, but it might be easier to just go back to manual conversion between `SerializableFormatOptions` and `FormatOptions`

crates/rome_formatter/tests/spec_test.rs
ematipico: Unfortunately not, at the moment `format` consumes `options` and can't be used later. We could change the API signature but it's not the aim of this PR.

CONTRIBUTING.md
ematipico: Yeah I know. The conversion is the most straight forward solution

ematipico:
I ended up going back to first solution I implemented:
- using serde `with` doesn't work well with vectors, and we would like to keep the possibility to run multiple configurations at once
- using features is tricky and some tests might not be run

crates/rome_formatter/tests/spec_test.rs
ematipico: Thank you! Done

leops:
Ah yes the `#[serde(with = "...")]` attribute can't be directly set on a field of type `Vec<_>` since Vec is its own specific type, instead the elements of the Vec itself would have to to be a newtype like
```rust
#[derive(Serialize, Deserialize)]
#[serde(transparent)]
struct SerializableOptions(#[serde(with = "FormatOptionsDef")] FormatOptions);
```

crates/rome_formatter/tests/spec_test.rs
yassere: `FormatOptions` implements the `Copy` trait so the `options` variable is still accessible after the call to `format`.  I just confirmed locally.

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -35,26 +116,47 @@ pub fn run(spec_input_file: &str, _: &str, file_type: &str) {
...
+            {
+                let mut options_path = RomePath::new(options_path.display().to_string().as_str());
+                // SAFETY: we checked its existence already, we assume we have rights to read it
+                let options: TestOptions =
+                    serde_json::from_str(options_path.get_buffer_from_file().as_str()).unwrap();
+
+                for test_case in options.cases {
+                    let options = test_case;
+                    let formatted_result = format(test_case.into(), &root).unwrap();
+                    snapshot_content.add_output(formatted_result.as_code(), options.into());
```
yassere, line 150: I'm not sure if you saw my other comment because it's in an "Outdated" thread, but the [Copy](https://doc.rust-lang.org/std/marker/trait.Copy.html) trait lets you access a variable even after it would have been moved. So in this case, you can use `test_case.into()` for both function calls because `test_case` isn't moved. This isn't a necessary change, of course, but I figured you'd want to know more about copy semantics.

xunilrj:
> Nice! I like the idea and code looks good to me. Let's wait for @xunilrj or @leops to chime in to see if the `cfg` approach is the right way to go

I would not focus too much on this now, honestly. @leops last suggestion is definitely the best, but I would go with it only on optimizing later.

ematipico:
Other than some optimizations that can be done later on, the PR is read and it implements what it supposed it.
 
Can we merge it please?

crates/rome_formatter/tests/spec_test.rs
MichaReiser: @ematipico have you tried this?

crates/rome_formatter/tests/spec_tests.rs
```diff
@@ -4,11 +4,11 @@ mod formatter {
 
     mod js_module {
         use crate::spec_test;
-        tests_macros::gen_tests! {"tests/specs/js/module/**/**/*.js", spec_test::run, "module"}
+        tests_macros::gen_tests! {"tests/specs/js/module/**/**/**/*.js", spec_test::run, "module"}
```
MichaReiser, line 7: This shouldn't be needed. `**/*.js` should match all JS files regardless of the depth.

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -22,7 +103,7 @@ use std::path::Path;
 ///
 /// * `json/null` -> input: `tests/specs/json/null.json`, expected output: `tests/specs/json/null.json.snap`
 /// * `null` -> input: `tests/specs/null.json`, expected output: `tests/specs/null.json.snap`
-pub fn run(spec_input_file: &str, _: &str, file_type: &str) {
+pub fn run(spec_input_file: &str, _expected_file: &str, test_directory: &str, file_type: &str) {
```
MichaReiser, line 106: Nit: The signature starts to become a bit awkward (so many strings). Could be worth having a `SpecTestFile` struct that has different attributes like `input_file`, `expected_file`, and `file_type`. I'm not sure what `test_directory` is, whatever it belongs to the file or is the same for all files.

crates/tests_macros/src/lib.rs
```diff
@@ -74,9 +75,12 @@ impl Arguments {
         let file_stem = path.file_stem()?;
         let file_stem = file_stem.to_str()?;
         let test_name = file_stem.to_snake();
+        let mut ancestors = path.ancestors();
+        // the first one will yield the path to the current file
+        ancestors.next();
+        let test_directory = ancestors.next().unwrap().display().to_string();
```
MichaReiser, line 81: What exactly is the `test_directory`. Is it the parent directory of the file? Might be easier to understand if you simply call `path.parent()`

MichaReiser (approved):

yassere (approved):

crates/rome_formatter/tests/spec_test.rs
ematipico: Yes, and it's not possible with the data [structure that I want to have](https://github.com/rome/tools/pull/2045#issuecomment-1031209950).

MichaReiser: Have you seen https://github.com/rome/tools/pull/2045#issuecomment-1031313305

But yeah, this is a nit.

PR #2047, opened 2022-02-04
fix(rslint_parser): mark optional type parameters is alias statement

ematipico:
## Summary

This PR fixes a small bug in our grammar. Now the `type_parameters` in `TsTypeAliasStatement` are marked as optional.

## Test Plan

We already have tests for these cases. I updated the current RAST files


MichaReiser (approved):
Thanks

PR #2048, opened 2022-02-04
feat(rslint_parser): Type parameters

MichaReiser:
## Summary

Adds type parameter parsing to function, arrow functions, and methods. Adds graceful recovery in case getters/setters have type parameters.

* Removes the `type_parameters` from the `ConstructorClassMember`. Constructors don't support type parameters
* Added improved diagnostics for getters, setters, or constructors with type parameters (this is a syntax error)
* Added improved diagnostics for setters with return type annotations
* Renamed `JsObjectMethodMember.type_params` to `type_parameters` to align with other nodes

## Test Plan

Added new tests for type parameters in methods, functions, arrow functions.


ematipico (approved):
Looks good. One tiny thing: sometime we use `cannot` and sometimes we use `can not` in the diagnostics

PR #2049, opened 2022-02-04
chore: move format files to JS folder

ematipico:
## Summary

Inside the formatter we have a folder called `ts/`. This is misleading because we have JS formatters (which will contain TS too) and I think it's better to divide the formatters based on their language.

- renamed `ts/` to `js/`
- created a new folder called `ts/` where we will store files that format TS syntax

We will do the same for another languages and super languages (`json`, `jsx`, etc.)

## Test Plan

Current CI


MichaReiser:
Would it be worth waiting for #2031 and do this change as part of the codegen?

ematipico:
It would yes, but I don't know what its status. I wanted to start working on TS formatting and thought that this change was required first.

MichaReiser:
> It would yes, but I don't know what's its status. I wanted to start working on TS formatting and thought that this change was required first.

Me neither. Maybe you can continue working on the codegen if @leops is focusing on the tests for the time being?

PR #2052, opened 2022-02-04
fix(rome_formatter): various minor corrections to the formatter

leops:
## Summary

This fixes a number of issues that were caught when running our formatter on the prettier test suite:

- Directive lists in function bodies were not printed
- The `static` token on class getters was not printed
- The `async` token on class methods was not printed
- The `static` token on class getters was not printed
- The `await` token on for-of statements was not printed
- The debug assertions in `get_lines_between_nodes` were raised when empty statements were being filtered

Most of these are fixed by formatting the missing tokens. For the last point, instead of filtering out empty statements from the iterator they are left in, but the `join_elements_with` function now skips emitting a line break before an empty element.

### DIFF
--- a/crates/rome_formatter/tests/specs/js/module/class/class.js
+++ b/crates/rome_formatter/tests/specs/js/module/class/class.js
@@ -6,6 +6,10 @@ class Foo extends Boar {
 		super();
 	}
 
+	static get sg() {
+
+	}
+
 	get g() {
 
 	}
@@ -18,6 +22,10 @@ class Foo extends Boar {
 
 	lorem() {
 		return "ipsum";
+	}
+
+	async ipsum() {
+
 	}
 
 	static foo;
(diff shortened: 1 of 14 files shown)


crates/rome_formatter/tests/specs/js/module/class/class.js
```diff
@@ -6,6 +6,10 @@ class Foo extends Boar {
 		super();
 	}
 
+	static get sg() {
+
+	}
```
ematipico, line 11: Can we add a test case with `static async *` together?

ematipico (approved):

crates/rome_formatter/tests/specs/js/module/class/class.js
leops: I've rearranged some methods and added a few others, this should now cover all possible combinations of static, async and generator

ematipico (approved):

PR #2054, opened 2022-02-04
feat(rome_formatter): add the prettier test suite

leops:
## Summary

This add 642 tests cases extracted from Prettier's test suite (I tried to only select JS files and ignore TypeScript / JSX and Flow but a few might have passed through), along with their accompanying snapshots.

Besides greatly increasing the test coverage (I included two additional fixes I missed when extracting #2052) this adds the possibility of semi-exhaustively comparing the output of the Rome Formatter with Prettier: when the `prettier_tests` are run with the `REPORT_PRETTIER` environment variable set to 1 this will generate a `report.md` file detailing the difference between the Prettier snapshots and the Rome snapshots.


xunilrj:
Don't know if it is going to be useful or not, but we can have a number on how much the formatter is missing using something like: 

https://github.com/ProbablyClem/cargo-todo
```
cargo todo legacy | grep "crates/rome_formatter" | wc -l
```

crates/rome_formatter/tests/prettier_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+
+            writeln!(report, "```").unwrap();
+        }
+
+        write("report.md", report).unwrap();
+    }
+}
+
+#[ctor::dtor]
+fn print_report() {
```
xunilrj, line 110: hum.... ```ctor``` does not advise printing anything at ```dtor```. Do you think it is safe?

https://github.com/mmastrac/rust-ctor
>         // Using println or eprintln here will panic as Rust has shut down

xunilrj (approved):

crates/rome_formatter/src/ts/expressions/literal_expression.rs
```diff
@@ -14,12 +17,18 @@ impl ToFormatElement for JsStringLiteralExpression {
...
             let s = &quoted[1..quoted.len() - 1];
             let s = format!("\"{}\"", s);
-            Token::new_dynamic(s, value_token.text_trimmed_range())
+            match normalize_newlines(&s, ['\r']) {
+                Cow::Borrowed(_) => s,
+                Cow::Owned(s) => s,
+            }
         } else {
-            Token::from(&value_token)
+            normalize_newlines(quoted, ['\r']).into_owned()
```
MichaReiser, line 25: How can a JS string contain any new line? Shouldn't these already be escaped as `\\r`?

crates/rome_formatter/tests/prettier_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+
+fn test_snapshot(input_file: &'static str, _: &str, _: &str) {
+    if input_file.contains("typescript") {
+        return;
+    }
+
+    let input_file = Path::new(input_file);
+    let input_code = read_to_string(input_file).unwrap();
+
+    let syntax = parse_script(&input_code, 0).syntax();
```
MichaReiser, line 26: Are these all scripts are some of the files modules?

crates/rome_formatter/tests/prettier_tests.rs
leops: The report printing is writing to an explicitly opened file descriptor and not to stdout so it should work alright. In fact while I was implementing this I used `println` inside `print_report` and it didn't panic or otherwise misbehave (on Windows at least), and in general outside of a few specific things (panicking of course, maybe reading the CLI arguments) Rust is designed to be safe to run even outside of `main` (otherwise calling into Rust libraries through FFI wouldn't be possible)

crates/rome_formatter/tests/prettier_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    fn report(&self, file_name: &'static str, rome: String, prettier: String) {
+        self.state.lock().push((file_name, rome, prettier));
+    }
+
+    fn print(&self) {
+        match env::var("REPORT_PRETTIER") {
+            Ok(value) if value == "1" => {}
```
MichaReiser, line 84: What's the purpose of this env variable? Maybe add some documentation of what `1` means?

MichaReiser (commented):
Do we need to copy over prettier's license as we're taking their work?

crates/rome_formatter/tests/prettier_tests.rs
MichaReiser: How about implementing `Drop` on `DiffReporter` instead?

crates/rome_formatter/src/ts/expressions/literal_expression.rs
leops: This handles the case of a string literal containing an escaped newline: https://github.com/rome/tools/blob/365b83070291b85d1cb3ac974cd7c35f8ed0febf/crates/rome_formatter/tests/specs/prettier/quotes/strings.js#L59

MichaReiser: Right! I forgot about those.

crates/rome_formatter/tests/prettier_tests.rs
leops: That doesn't work because Rust doesn't call `Drop` on statics (I think the rationale is that `Drop` is generally used to free memory, and by the time statics would be dropped the program is exiting so any allocated memory is going to be freed by the OS anyway)

crates/rome_formatter/tests/prettier_tests.rs
ematipico: I would personally use `parse_module` as default. The reason is because `parse_script` doesn't support the module syntax.

leops:
Prettier is licensed as MIT so the license and a copyright notice must be copied over somewhere (probably in `rome_formatter/tests/specs/prettier/LICENSE`), but I think we should probably also have an acknowledgments document somewhere to keep track of these ?

crates/rome_formatter/tests/prettier_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+    fs::{read_to_string, write},
+    path::Path,
+};
+
+use rome_formatter::{FormatOptions, IndentStyle};
+use rslint_parser::parse_script;
+
+static REPORTER: DiffReport = DiffReport::new();
+
+tests_macros::gen_tests! {"tests/specs/prettier/**/*.js", test_snapshot, "script"}
```
ematipico, line 16:
```suggestion
tests_macros::gen_tests! {"tests/specs/prettier/**/**/**/*.js", test_snapshot, "script"}
```

I don't know how deep is the folder structure of prettier tests, but maybe it's safer to specify a deeper level?

ematipico (commented):
I wonder, should we update the `CONTRIBUTING.md` section about this change? What are the developers supposed to do? 

I noticed that the tests were just copied instead of using a git submodule, meaning that we don't know how and when to update  them. Or if this is something temporary.

ematipico:
> Do we need to copy over prettier's license as we're taking their work?

I think it's also worth adding a `README.md` in that folder to give some context.

crates/rome_formatter/tests/prettier_tests.rs
leops: Some of these files are modules, unfortunately besides the few snapshots in the `es6modules` directory there are no clear indications of which files are what (Prettier passes everything to Babel with the `sourceType` set to module so replacing this with `parse_module` might be a sensible default)

crates/rome_formatter/tests/prettier_tests.rs
leops: The `**` glob expression will recurse into directories while performing the search so there's no need to specify it multiple times

MichaReiser (approved):

crates/rome_formatter/tests/prettier_tests.rs
```diff
@@ -0,0 +1,112 @@
...
+    let syntax = parse_script(&input_code, 0).syntax();
+    let options = FormatOptions::new(IndentStyle::Space(2));
+    let formatted = rome_formatter::format(options, &syntax).unwrap();
+
+    let mut snapshot = String::new();
+
+    writeln!(snapshot, "# Input").unwrap();
+    writeln!(snapshot, "```").unwrap();
+    writeln!(snapshot, "{}", input_code).unwrap();
+    writeln!(snapshot, "```").unwrap();
```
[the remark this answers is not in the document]
leops: This makes the snapshots valid markdown files, which improves the viewing of individual snapshot files (I'll add a JS language annotation to these code blocks so they will all have the syntax coloring working)

leops:
> I noticed that the tests were just copied instead of using a git submodule, meaning that we don't know how and when to update them. Or if this is something temporary.

This is not a submodule because the tests are not directly copied over from Prettier, they were extracted from Jest snapshot files in the Prettier repository with a node.js script. Since exactly matching with Prettier is a non-goal for now I figured this would be a useful reference but not necessarily one we'll want to update often or track very closely. Still, with a bit of reformatting and documentation I supposed I could commit my JS script along with this.

crates/rome_formatter/tests/prettier_tests.rs
ematipico: That's the thing. I doesn't work as expected then. Sometimes I create a new level of directories and it's not recognized unless I add a new `**/`

crates/rome_formatter/tests/prettier_tests.rs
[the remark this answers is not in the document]
ematipico: But the files are `.snap`, `.md`. So where do you get this improved visibility?

crates/rome_formatter/tests/prettier_tests.rs
leops: I think that's not an issue with the glob search but with how Cargo is caching builds: because it doesn't know the tests depend on these external files it will not re-build the tests (and re-run the search in the proc macro) until the `.rs` file itself is modified

crates/rome_formatter/tests/prettier_tests.rs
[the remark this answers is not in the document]
leops: It's a user setting so it's not committed but I've associated ".snap" files with the "markdown" syntax in my editor. I could mention that in the snapshots README or in CONTRIBUTING, obviously that's entirely optional and just a minor QoL improvement for viewing snapshots

leops:
An additional change I made is adding the errors emitted by the parser in the snapshots, this makes it easier to single out files that might be formatted wrong due to syntax errors

ematipico:
> An additional change I made is adding the errors emitted by the parser in the snapshots, this makes it easier to single out files that might be formatted wrong due to syntax errors

This highlights that prettier has tests that have wrong code. Although, I wonder what we are trying to achieve by showing errors? Maybe let's just show a error label in the snapshot, I don't think we need to show the parsing errors (and their source code) inside the tests formatter.

leops:
Prettier has both test cases for error recovery as well as tests for experimental syntax. I could indeed just add a flag or line somewhere in the test that only specifies that the file has errors, but I thought that if and when we change the error recovery behavior of the parser or introduce new, previously unsupported syntax this will have a ripple effect on the output of the formatter for these files, and this will be better documented in the resulting git diff if it shows the error changing or getting removed alongside the new formatter result. Maybe there's an in-between solution of printing a list of each error with just their locations but not the error message itself ?

ematipico:
Your arguments are valid, totally. I just don't see parsing errors as part of the implementation of a formatter, because don't belong to them.

Although, I can also see the value of the errors because a tree might have errors that are 4/5 levels deep and the formatter can format the rest of the nodes.

As you suggested, if it's possible, we could highlight the location of where we couldn't format the code:

```js
function a() {
	const a = () => {
		return { lorem: "ipsum"  }; 
	}

	async do {} // this statement don't be formatted, we could alight its start and end positions? 
}
```

This is something that is incredibly valuable because sometimes, if and when there's a parsing error, we don't know which code was formatted and which one not.

MichaReiser:
> Your arguments are valid, totally. I just don't see parsing errors as part of the implementation of a formatter, because I think they don't belong to them.
> 
> Although, I can also see the value of the errors because a tree might have errors that are 4/5 levels deep and the formatter can format the rest of the nodes.
> 
> As you suggested, if it's possible, we could highlight the location of where we couldn't format the code:
> 
> ```js
> function a() {
> 	const a = () => {
> 		return { lorem: "ipsum"  }; 
> 	}
> 
> 	async do {} // this statement don't be formatted, we could alight its start and end positions? 
> }
> ```
> 
> This is something that is incredibly valuable because sometimes, if and when there's a parsing error, we don't know which code was formatted and which one wasn't.

I see value in testing error recovery as well because that's a very important aspect when it comes to IDE integration. Ideally, the formatter wouldn't depend on the parser's error recovery behaviour at all but I don't think that's feasible. It will depend on some of the parsers error recovery, at least implicitly. Thus, it makes sense to capture these implicit dependencies by capturing them in a test. 

I don't think this should prevent us from changing the parsers error recovery at any point but it will give us a heads up that it may regress the formatting or allow us to consider an alternative approach.

leops:
I experimented a bit with error printing, I tried a few ways of removing the error messages to only print the error ranges underlined but the result wasn't really convincing. I think highlighting the ranges that were printed verbatim might not be enough as sometimes the parser recognize incomplete syntax as valid enough for these nodes to be formatted when they should not (for example in https://github.com/rome/tools/blob/e0dc81c3041b560be733d2bbc57ffe5c7bdab2e1/crates/rome_formatter/tests/specs/prettier/arrows/newline-before-arrow/newline-before-arrow.js.snap)

ematipico:
OK, let's keep the current implementation! I can see their values now :)

PR #2055, opened 2022-02-07
refactor(rslint_parser): Formal parameter with default

MichaReiser:
## Summary
Split `JsFormalParameter` into `JsFormalParameter` and `JsFormalParameterWithDefault` to correctly model that an initializer is only allowed if the parameter isn't marked as optional (`?`)

```js
function a(x = "test") {} // valid
function b(x?) {} // valid
function c(x? = "test") {} // invalid
```

## Test Plan

There are already tests covering the two patterns. Verified the rast outputs.


crates/rslint_parser/src/ast/generated/nodes.rs
```diff
@@ -8,22 +8,6 @@ use crate::{
     SyntaxNode, SyntaxResult, SyntaxToken,
 };
 #[derive(Clone, PartialEq, Eq, Hash)]
-pub struct Ident {
-    pub(crate) syntax: SyntaxNode,
-}
-impl Ident {
```
MichaReiser, line 14: No `Ident` node existed. The parser only ever pushed a `ident` token.

ematipico (approved):

ematipico:
As per discussion #2051 , this would be a second `WithDefault` node. At this point, should we change its name?

`JsFormalParameterWithDefault` => `JsWithDefaultFormalParameter`

MichaReiser:
> As per discussion #2051 , this would be a second `WithDefault` node. At this point, should we change its name?
> 
> `JsFormalParameterWithDefault` => `JsWithDefaultFormalParameter`

`WithDefaultFormalParameter` sounds a bit odd. But you brought up an issue that this node should indeed end with `FormalParameter` because it is a `FormalParameter`. 

I'm going to abandon this PR because I'm having second thoughts overall if it makes sense to model the AST as strict. This may be desired from an analyzer perspective but I don't see a reason why e.g. a test runner shouldn't be able to run the test just because of some misplaced typescript annotation.

PR #2057, opened 2022-02-07
feature(rslint_parser): TS Optional & Definite Members and Variables

MichaReiser:
## Summary

Implements parsing for optional and definite members and variables

```
let a!: string;
class A {
  a!: string;
  b?: string = "test";
  c?() {}
}
```

This PR introduces the new AST nodes

```
TsAnyVariableAnnotation =
	TsTypeAnnotation
	| TsDefiniteVariableAnnotation

TsDefiniteVariableAnnotation =
	'!'
	type_annotation: TsTypeAnnotation
```

and

```
TsAnyPropertyAnnotation =
	TsTypeAnnotation
	| TsOptionalPropertyAnnotation
	| TsDefinitePropertyAnnotation

TsOptionalPropertyAnnotation =
	'?'
	type_annotation: TsTypeAnnotation?

TsDefinitePropertyAnnotation =
	'!'
	type_annotation: TsTypeAnnotation
```

These model the constraints that type annotations are mandatory if the `!` operator is used. However, the nodes do not model that an initializer clause isn't allowed in combination with `!`.

One reason for this is simply naming. I haven't been able to come up with a good name for such a node.

The second reason is that I believe that an initializer on a definite property shouldn't prevent a bundler/test runner from running the JavaScript file. Thus, the parser must be able to return a valid node if both the `!` and the initializer are present.

## Test Plan

Added new tests for the newly covered syntax. Also added Tests to show how the syntax is parsed if encountered in a JS file.

Verified the RAST outputs.

The regressions are from files that previously failed because of parsing errors when encountering a `?` but expect other compiler diagnostics.


ematipico:
>One reason for this is simply naming. I haven't been able to come up with a good name for such a node.

> The second reason is that I believe that an initializer on a definite property shouldn't prevent a bundler/test runner from running the JavaScript file. Thus, the parser must be able to return a valid node if both the ! and the initializer are present.

I believe you made the right call here and I agree with it. I think the use of a mandatory type annotation is beneficial only during type checking. We can have also a lint rule to catch these inconsistencies.

crates/rslint_parser/src/parser/parse_error.rs
```diff
@@ -14,15 +15,54 @@ pub(crate) fn expected_any(names: &[&str], range: Range<usize>) -> ExpectedNodeD
...
+    fn to_diagnostic(self, p: &Parser) -> Diagnostic;
+}
+
+impl ToDiagnostic for Diagnostic {
+    fn to_diagnostic(self, _: &Parser) -> Diagnostic {
+        self
+    }
 }
 
+pub struct ExpectedToken(JsSyntaxKind);
```
ematipico, line 33: I would append `Diagnostic` to the name of the struct. So we know when to use it and for what.

ematipico: Can this be `pub(crate)`?

crates/rslint_parser/src/parser/parse_error.rs
```diff
@@ -14,15 +15,54 @@ pub(crate) fn expected_any(names: &[&str], range: Range<usize>) -> ExpectedNodeD
     ExpectedNodeDiagnosticBuilder::with_any(names, range)
 }
 
+#[must_use]
+pub(crate) fn expected_token(token: JsSyntaxKind) -> ExpectedToken {
```
ematipico, line 19: I would call it `expected_token_diagnostic`

crates/rslint_parser/src/parser/parse_error.rs
ematipico: Does it have to be `pub`? Don't we use `expected_token` to generate the `struct`?

MichaReiser: I can change `expected_token` so that it returns `impl ToDiagnostic`.

crates/rslint_parser/src/parser/parse_error.rs
MichaReiser: I would prefer keeping the current name so that it aligns with `expected_any`, `expected_node`...

ematipico: Oh. Then, should we not move it inside `js_parse_error.rs`?

MichaReiser: `expected_any` and `expected_node` are also defined in `parse_error`.

ematipico (approved):

ematipico (approved):

PR #2060, opened 2022-02-07
Windows: inline test file ending

MichaReiser:
## Summary

Our windows build is failing because `ts` line endings are converted to windows line endings by git on check out. 

This PR changes the line ending to LF for all files in the test data directory

https://github.com/rome/tools/actions/runs/1807656578

## Test Plan

Test in prod?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 * text=auto
-crates/rslint_parser/test_data/**/*.js text eol=lf
+crates/rslint_parser/test_data/**/*.{js,ts} text eol=lf
 xtask/js.ungram text eol=lf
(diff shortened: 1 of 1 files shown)


MichaReiser:
@ematipico is there a way to trigger the windows build for a PR?

ematipico:
> @ematipico is there a way to trigger the windows build for a PR?

Not with the current infrastructure. We can add special comments for these kind of situations

MichaReiser:
> > @ematipico is there a way to trigger the windows build for a PR?
> 
> Not with the current infrastructure. We can add special comments for these kind of situations

In that case we just need to be bold 😆

leops:
This has me wondering if there would be any issue with just set the entire repo to use LF and be done with this. Prettier seems to be doing that for instance.

ematipico:
> This has me wondering if there would be any issue with just set the entire repo to use LF and be done with this. Prettier seems to be doing that for instance.

That's another solution

MichaReiser:
> This has me wondering if there would be any issue with just set the entire repo to use LF and be done with this. Prettier seems to be doing that for instance.

I wouldn't mind that but let's address this in its own PR

Issue #2059, opened 2022-02-07
Typescript: `abstract` classes


xunilrj:
- [x] Simple abstract class;
    - [x] With concrete members;
    - [x] With abstract members;
        - [x]  Fix ```conformance/classes/classDeclarations/classAbstractKeyword/classAbstractMethodWithImplementation.ts```;
- [x] ```abstract``` is a valid class name (even in TS) and a valid enum variant;
- [x] abstract classes must be exportable;
- Error Handling
    - [x] Abstract member should not have body;
    - abstract members should not appear in interfaces, non abstract classes and enums;
    - [x] private member cannot be abstract
        - [x] private name member cannot be abstract
    - [x] async member cannot be abstract
    - [x] abstract members should not be static
- [x] overloaded methods/constructors do not have body

MichaReiser:
Let me share my thoughts around abstract members here as this will now be tackled as part of this issue. Feel free to follow another approach:

I considered re-using `MethodSignatureTypeMember`, `GetterSignatureTypeMember`, `SetterSignatureTypeMember` for abstract methods because they don't have a body
However, this comes with the downside that:
* These members must have an abstract token that is only valid inside classes but neither for types nor interfaces
* The naming isn't as nice anymore... `JsAnyClassMember` then contains `*TypeMember` variants

The reason why I wouldn't add `abstract` to the `ClassMember` is that the `body` is mandatory for non-abstract members. That's a constraint that we, so far, uphold. A mandatory node is always mandatory in all contexts. Furthermore, we already have many cases where a specific child isn't allowed but is otherwise considered optional (variable initializers, parameter initializers...)

MichaReiser:
@xunilrj can we close this issue. I'll create new nones for the ambient parsing, `override` and the modifier order.

xunilrj:
Needs improvement:

- [ ] Abstract members should not appear in interfaces, non abstract classes and enums;

```
error[SyntaxError]: ';' expected'
  ┌─ ts_interfaces_cannot_have_abstract_members.ts:2:14
  │
2 │     abstract prop: string;
  │              ^^^^ An explicit or implicit semicolon is expected here...

--
error[SyntaxError]: ';' expected'
  ┌─ ts_interfaces_cannot_have_abstract_members.ts:3:14
  │
3 │     abstract method(): string;
  │              ^^^^^^ An explicit or implicit semicolon is expected here...

--
error[SyntaxError]: index signature members cannot be abstract
[...]

PR #2068, opened 2022-02-08
feat(rslint_parser): typescript abstract class parsing

xunilrj:
see https://github.com/rome/tools/issues/2059

This PR implements:
- abstract keyword before class;
- abstract classes can be exported;
- error when abstract members have body.
- error when abstract is used together with private name;
- error when abstract is used together with async;

# Validations

```
error[SyntaxError]: class declarations must have a name
  ┌─ typescript_abstract_classes_incomplete.ts:1:10
  │
1 │ abstract class {};
  │          ^^^^^^
```

```
error[SyntaxError]: abstract members should not have a body
  ┌─ typescript_class_member_body.ts:5:30
  │
5 │     abstract display(): void { console.log(this.name); }
  │                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^ unexpected

--
error[SyntaxError]: abstract members should not have a body
  ┌─ typescript_class_member_body.ts:6:28
  │
6 │     abstract get my_name() { return this.name; }
  │                            ^^^^^^^^^^^^^^^^^^^^^ unexpected

--
error[SyntaxError]: abstract members should not have a body
  ┌─ typescript_class_member_body.ts:7:32
  │
7 │     abstract set my_name(name) { this.name = name; }
  │                                ^^^^^^^^^^^^^^^^^^^^^ unexpected

```

```
error[SyntaxError]: member with private name cannot be abstract
  ┌─ typescript_abstract_classes_invalid_abstract_private_member.ts:1:20
  │
1 │ abstract class A { abstract #name(); };
  │                    ^^^^^^^^
```

```
error[SyntaxError]: async members cannot be abstract
  ┌─ typescript_abstract_classes_invalid_abstract_async_member.ts:1:20
  │
1 │ abstract class B { abstract async a(); }
  │                    ^^^^^^^^
```

# Regressions

All regressions are because we are now parsing correctly the file and not doing runtime/semantic validations.


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -458,16 +458,25 @@ fn parse_class_member_impl(
...
             parse_constructor_class_member_body(p, member_marker, modifiers)
         } else {
+            // test_err ts typescript_abstract_classes_invalid_abstract_async_member
+            // abstract class B { abstract async a(); }
+            if let Some(abstract_range) = modifiers.get_range(ModifierKind::Abstract) {
+                let err = p
+                    .err_builder("async members cannot be abstract")
+                    .primary(abstract_range, "");
+                p.error(err);
+            }
```
MichaReiser, line 471: Nit. Can we move this into `parse_method_class_member`. You can use the `flags` to determine if it is async. But I guess, doing the check here is fine too if it's the only place where it is needed

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -622,8 +631,9 @@ fn parse_class_member_impl(
                         p.expect(T!['(']);
                         p.expect(T![')']);
                         parse_ts_type_annotation_or_error(p).ok();
-                        parse_function_body(p, SignatureFlags::empty())
-                            .or_add_diagnostic(p, js_parse_error::expected_class_method_body);
+
+                        let body = parse_function_body(p, SignatureFlags::empty());
+                        check_body(body, p, modifiers);
```
MichaReiser, line 636: Nit: How about a `parse_class_method_body` that combines `parse_function_body` and `check_body`?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1171,6 +1211,16 @@ pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax {
...
     } else {
         p.expect(T![ident]);
+
+        // test_err ts typescript_abstract_classes_invalid_abstract_private_member
+        // abstract class A { abstract #name(); };
+        if let Some(abstract_range) = modifiers.get_range(ModifierKind::Abstract) {
+            let err = p
+                .err_builder("members with private name cannot be abstract")
+                .primary(abstract_range, "");
+            p.error(err);
```
MichaReiser, line 1221: We should mark the member name or the member as `unknown` in that case.

crates/rslint_parser/src/syntax/typescript/classes.rs
```diff
@@ -0,0 +1,62 @@
...
+pub(crate) fn parse_ts_abstract_class_statement(
+    p: &mut Parser,
+    _ctx: StatementContext,
+) -> ParsedSyntax {
+    if !is_at_ts_abstract_class_statement(p, LineBreak::DoCheck) {
+        return Absent;
+    }
+
+    let m = p.start();
+    p.bump_remap(rslint_lexer::JsSyntaxKind::ABSTRACT_KW);
```
MichaReiser, line 50: Nit: `T![abstract]`?

crates/rslint_parser/src/syntax/typescript/classes.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let mut class =
+        crate::syntax::class::parse_class(p, m, crate::syntax::class::ClassKind::Statement);
+
+    // test_err abstract_class_in_js
+    // abstract class A {}
+    class.err_if_not_ts(
+        p,
+        "`abstract classes` can only be declared in TypeScript files",
+    );
```
MichaReiser, line 60: We normally introduce the checks in the caller side. `parse_ts_abstract_class_statement` parses typescript syntax. It shouldn't throw if I call into it from another typescript only parse method.

crates/rslint_parser/src/syntax/typescript/classes.rs
```diff
@@ -0,0 +1,62 @@
...
+
+// test_err ts typescript_abstract_classes_incomplete
+// abstract class {};
+
+// test_err ts typescript_abstract_classes_invalid_abstract_constructor
+// abstract class A { abstract constructor();};
+
+pub(crate) fn parse_ts_abstract_class_statement(
+    p: &mut Parser,
+    _ctx: StatementContext,
```
MichaReiser, line 43: Remove `ctx` if not needed?

crates/rslint_parser/src/syntax/typescript/classes.rs
```diff
@@ -0,0 +1,62 @@
...
+) -> ParsedSyntax {
+    if !is_at_ts_abstract_class_statement(p, LineBreak::DoCheck) {
```
MichaReiser, line 45: I'm uncertain if we should instead move the abstract logic into the `parse_class_statement` function because there's no such thing as a `TsAbstractClassStatement` node. If you decide to keep the function, can you then move it to `typescript/statement.rs`? 

Can you also add a test for 

```
export abstract class test {}
export default abstract class Test {}
```

crates/rslint_parser/src/syntax/typescript/enums.rs
```diff
@@ -0,0 +1,186 @@
+use crate::parser::RecoveryResult;
```
MichaReiser, line 1: Can you rebase onto main. This is no inside of the `typescript/statement.rs` file.

crates/rslint_parser/src/syntax/typescript/statement.rs
```diff
@@ -54,7 +54,8 @@ fn parse_ts_enum_member(p: &mut Parser) -> ParsedSyntax {
                 .err_builder("An `enum` member cannot be private")
                 .primary(p.cur_tok().range(), "");
             p.error(err);
-            syntax::class::parse_private_class_member_name(p).map(|mut x| {
+            let modifiers = ClassMemberModifiers::default();
+            syntax::class::parse_private_class_member_name(p, &modifiers).map(|mut x| {
```
MichaReiser, line 58: Nit: Can we have two functions instead? One that accept a `modifiers` and one that does not?`

crates/rslint_parser/src/syntax/typescript/ts_parse_error.rs
```diff
@@ -8,3 +8,9 @@ use std::ops::Range;
 pub(crate) fn expected_ts_enum_member(p: &Parser, range: Range<usize>) -> Diagnostic {
     expected_any(&["identifier", "string literal", "computed name"], range).to_diagnostic(p)
 }
+
+pub(crate) fn unexpected_abstract_member_with_body(p: &Parser, range: Range<usize>) -> Diagnostic {
+    dbg!(&range);
```
MichaReiser: Delete

crates/rslint_parser/src/tests.rs
```diff
@@ -112,7 +112,7 @@ fn run_and_expect_no_errors(path: &str, _: &str, _: &str, _: &str) {
 #[cfg(test)]
 fn run_and_expect_errors(path: &str, _: &str, _: &str, _: &str) {
     let path = PathBuf::from(path);
-    let text = std::fs::read_to_string(&path).unwrap();
+    let text = std::fs::read_to_string(dbg!(&path)).unwrap();
```
MichaReiser, line 115: Delete `dbg`

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/ts_parse_error.rs
xunilrj: Done

crates/rslint_parser/src/syntax/typescript/ts_parse_error.rs
```diff
@@ -8,3 +8,9 @@ use std::ops::Range;
...
+    p.err_builder("abstract members should not have a body")
+        .primary(range, "unexpected")
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rslint_parser/test_data/inline/err/abstract_class_in_js.rast
```diff
@@ -0,0 +1,42 @@
...
+      0: ABSTRACT_KW@0..9 "abstract" [] [Whitespace(" ")]
+      1: CLASS_KW@9..15 "class" [] [Whitespace(" ")]
+      2: JS_IDENTIFIER_BINDING@15..17
+        0: IDENT@15..17 "A" [] [Whitespace(" ")]
+      3: L_CURLY@17..18 "{" [] []
+      4: JS_CLASS_MEMBER_LIST@18..18
+      5: R_CURLY@18..19 "}" [] []
+  3: EOF@19..20 "" [Newline("\n")] []
+--
+error[SyntaxError]: `abstract classes` can only be declared in TypeScript files
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/classes.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/classes.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/enums.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/statement.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -651,8 +661,8 @@ fn parse_class_member_impl(
                             ));
                         }
 
-                        parse_function_body(p, SignatureFlags::empty())
-                            .or_add_diagnostic(p, js_parse_error::expected_class_method_body);
+                        let body = parse_function_body(p, SignatureFlags::empty());
+                        check_body(body, p, modifiers);
 
                         member_marker.complete(p, JS_SETTER_CLASS_MEMBER)
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -57,16 +72,46 @@ pub(super) fn parse_class_expression(p: &mut Parser) -> ParsedSyntax {
...
+    let is_at_abstract_class = is_at_ts_abstract_class_declaration(p, LineBreak::DoCheck);
+    if !is_at_class && !is_at_abstract_class {
         return Absent;
     }
 
     let m = p.start();
+
+    if is_at_contextual_keyword(p, "abstract") {
+        p.bump_remap(T![abstract]);
+    }
```
MichaReiser, line 113: Nit: Delete? `parse_class` eats the `abstract` keyword already

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -132,7 +189,11 @@ impl From<ClassKind> for JsSyntaxKind {
     }
 }
 
-fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
+pub(crate) fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
+    if is_at_contextual_keyword(p, "abstract") {
+        p.bump_remap(T![abstract]);
+    }
+
```
MichaReiser, line 196: Nit: 
```suggestion
    eat_contextual_keyword(p, "abstract", T![abstract]);
```

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -458,16 +519,16 @@ fn parse_class_member_impl(
...
             parse_constructor_class_member_body(p, member_marker, modifiers)
         } else {
             parse_method_class_member(p, member_marker, modifiers, flags)
         });
     }
 
     let is_constructor = is_at_constructor(p, &modifiers);
-    let member_name =
-        parse_class_member_name(p).or_add_diagnostic(p, js_parse_error::expected_class_member_name);
+    let member_name = parse_class_member_name_with_modifiers(p, &modifiers)
```
MichaReiser, line 530: `parse_class_member_name` is now unused. I think we can keep `parse_class_member_name` if the one without modifiers is never used.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -522,7 +583,7 @@ fn parse_class_member_impl(
             //   static async* static() {}
             //   static * static() {}
             // }
-            Present(parse_method_class_member_body(
+            Present(parse_everything_after_identifier(
```
MichaReiser, line 586: This method name isn't telling me what it's parsing. Can we keep `parse_method_class_member_body` or TypeScript uses `parse_...._rest`

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -622,8 +683,8 @@ fn parse_class_member_impl(
                         p.expect(T!['(']);
                         p.expect(T![')']);
                         parse_ts_type_annotation_or_error(p).ok();
-                        parse_function_body(p, SignatureFlags::empty())
-                            .or_add_diagnostic(p, js_parse_error::expected_class_method_body);
+
+                        parse_method_class_member_body(p, modifiers, SignatureFlags::empty());
```
MichaReiser, line 687: Nit: `Getters/setters` aren't methods. Can we name this `parse_method_function_body` or `parse_method_body`. The `member` in the name can be misleading because it gives the impression that it's parsing a `method member`

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -928,8 +995,19 @@ fn parse_method_class_member(
...
-    parse_class_member_name(p).or_add_diagnostic(p, js_parse_error::expected_class_member_name);
-    parse_method_class_member_body(p, m, modifiers, flags)
+    let abstract_range = modifiers.get_range(ModifierKind::Abstract);
+    let is_async = flags.contains(SignatureFlags::ASYNC).then(|| true);
+
+    // test_err ts typescript_abstract_classes_invalid_abstract_async_member
+    // abstract class B { abstract async a(); }
+    if let Some((abstract_range, _)) = abstract_range.zip(is_async) {
+        let err = ts_parse_error::abstract_member_cannot_be_async(p, abstract_range.clone());
+        p.error(err);
```
MichaReiser, line 1005: Nit: Mark the member as unknown?

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -475,6 +477,24 @@ pub(super) fn parse_export(p: &mut Parser) -> ParsedSyntax {
...
+                abstract_class.map(|mut abstract_class| {
+                    p.error(p.err_builder(
+                        "`abstract` classes can only be declared in TypeScript files",
+                    ));
+                    abstract_class.change_to_unknown(p);
+                    abstract_class
+                })
+            } else {
+                abstract_class
+            }
```
MichaReiser, line 496: You can use `TypeScript.exclusive_syntax(p, abstract_class, |p, abstract_class| ... )`. It takes care of the change to `unknown` and  adding the error

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -746,6 +766,9 @@ fn parse_export_default_clause(p: &mut Parser) -> ParsedSyntax {
 
     let clause = match p.nth(1) {
         T![class] => parse_export_default_class_case(p),
+        T![ident] if p.nth_src(1) == "abstract" && p.nth_at(2, T![class]) => {
```
MichaReiser, line 769: Nit: I started to prefer using `is_at_contextual_keyword` so that we can correctly handle the unicode sequences when we have incremental parsing 
```suggestion
        T![ident] if is_nth_at_contextual_keyword(p, 1, "abstract") && p.nth_at(2, T![class]) => {
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -209,6 +210,23 @@ pub(crate) fn parse_statement(p: &mut Parser, context: StatementContext) -> Pars
...
+                abstract_class.map(|mut abstract_class| {
+                    p.error(p.err_builder(
+                        "`abstract` classes can only be declared in TypeScript files",
+                    ));
+                    abstract_class.change_to_unknown(p);
+                    abstract_class
+                })
+            } else {
+                abstract_class
+            }
```
MichaReiser, line 228: TypeScript.exclusive_syntax(p, abstract_class, ...)

crates/rslint_parser/src/syntax/typescript/statement.rs
MichaReiser: I'm a bit confused. If there are now two methods, why is it still needed to pass the `modifiers` here?

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs
xunilrj: How about ```parse_method_class_member_parameters_return_and_body```? Just body is a little bit misleading.

crates/rslint_parser/src/syntax/class.rs
xunilrj: Went with ```parse_method_body```.

crates/rslint_parser/src/syntax/class.rs

xunilrj:
> I'm a bit confused. If there are now two methods, why is it still needed to pass the modifiers here?

Hum... we don't. I fixed it.

crates/rslint_parser/src/syntax/module.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/stmt.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -57,12 +73,37 @@ pub(super) fn parse_class_expression(p: &mut Parser) -> ParsedSyntax {
...
+// abstract class A { abstract name: string; };
+
 /// Parses a class declaration if it is valid and otherwise returns [Invalid].
 ///
 /// A class can be invalid if
 /// * It uses an illegal identifier name
 pub(super) fn parse_class_declaration(p: &mut Parser, context: StatementContext) -> ParsedSyntax {
-    if !p.at(T![class]) {
+    let is_at_class = p.at(T![class]);
+    let is_at_abstract_class = is_at_ts_abstract_class_declaration(p, LineBreak::DoCheck);
```
MichaReiser, line 105: Nit: I often make the `parse` method checks less restrictive because most callers will check if it is indeed a class declaration. For example. I would only test for the presence of the `class` or `abstract` keyword and if either is present, parse them out as a class.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -96,7 +144,7 @@ pub(super) fn parse_export_default_class_case(p: &mut Parser) -> ParsedSyntax {
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
-enum ClassKind {
+pub enum ClassKind {
```
MichaReiser, line 147: I didn't find any usages outside of the `class.rs` file. Can we make `ClassKind` and `parse_class` private again?

crates/rslint_parser/src/syntax/class.rs
MichaReiser: But what about the type parameters, return type? I think listing out all fields will be misleading. 
`parse_method_class_member_rest`. Not ideal, but at least it isn't misleading and we can go into more detail in the documentation.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -965,9 +1021,52 @@ fn parse_method_class_member_body(
...
+    // test_err ts typescript_abstract_classes_invalid_abstract_async_member
+    // abstract class B { abstract async a(); }
+    if let Some((abstract_range, _)) = abstract_range.zip(is_async) {
+        let err = ts_parse_error::abstract_member_cannot_be_async(p, abstract_range);
+        p.error(err);
+        member.change_to_unknown(p);
+        member
+    } else {
+        member
+    }
```
MichaReiser, line 1040: Nit
```suggestion
    if let Some((abstract_range, _)) = abstract_range.zip(is_async) {
        let err = ts_parse_error::abstract_member_cannot_be_async(p, abstract_range);
        p.error(err);
        member.change_to_unknown(p);
    }
    member
```

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -965,9 +1021,52 @@ fn parse_method_class_member_body(
...
         .ok();
 
-    parse_function_body(p, flags).or_add_diagnostic(p, js_parse_error::expected_class_method_body);
+    let abstract_range = modifiers.get_range(ModifierKind::Abstract).cloned();
+    let is_async = flags.contains(SignatureFlags::ASYNC).then(|| true);
 
-    m.complete(p, member_kind)
+    parse_method_body(p, modifiers, flags);
+
+    let mut member = m.complete(p, member_kind);
```
MichaReiser, line 1029: Do we need to mark the member as `unknown` if this is a JS file but it uses the `abstract` modifier?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1125,15 +1223,22 @@ fn is_at_class_member_name(p: &Parser, offset: usize) -> bool {
...
         T!['['] => parse_computed_member_name(p),
         _ => parse_literal_member_name(p),
     }
 }
 
 pub(crate) fn parse_private_class_member_name(p: &mut Parser) -> ParsedSyntax {
+    parse_private_class_member_name_with_modifiers(p, &ClassMemberModifiers::default())
+}
+
+pub(crate) fn parse_private_class_member_name_with_modifiers(
```
MichaReiser, line 1238: Nit: Can be private (together with `ClassMemberModifiers`)

MichaReiser (approved):

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs
xunilrj: I am surprised lint did not catch that. We actually need to make ```parse_class``` private.

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/class.rs
xunilrj: We do at https://github.com/rome/tools/pull/2068/files#diff-853a78342876609c9113a3b676b34500dc93c26c63bb781721d4ea303153681aR217

Rast file:  
https://github.com/rome/tools/pull/2068/files#diff-10c88346c8213434b137358b6df139a7fba1560cfda4b2128c5dde4fc33b052c

Issue #2069, opened 2022-02-08
☂️  Support TypeScript formatting 


ematipico:
# Objective

While the works on the [TypeScript parsing are on going](https://github.com/rome/tools/issues/1966), it's possible now starting working on supporting the formatting for TypeScript language.

In order to achieve that, the contributor needs first to check if the feature is actually supported by our parser. Also, the developer needs to check the error paths that aren't covered by the parser and try to avoid them for the time being.

Before working on something, please leave a comment here. Also, check the [`ungram` document](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram) to understand the grammar of the node.

## List of the nodes (and possible children nodes) to format*

- [x] [`TsTypeAliasStatement`](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L1513-L1519), this involves all the types that belong to `TsAnyType` @ematipico  #2070 
- [x] #2080 @ematipico 
- [x] [`TsTypeAssertionExpression`](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L1826-L1830)
- [x] [`TsAsExpression`](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L1832-L1835)
- [x] [`TsThisParameter`](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L1409-L1411)
- [x] [`TsTypeParameters`](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L1613-L1616)
- [x] [`TsAnyObjectTypeMember`](https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L1643-L1650) 

*the position of the source code might change due to the on going works

ematipico:
I am going to start the formatting of `TsEnumStatement`

PR #2070, opened 2022-02-08
feat(rome_formatter): basic formatting for `type`

ematipico:
## Summary

Part of #2069 

~**I plan to merge this PR after https://github.com/rome/tools/pull/2031, otherwise the amount of conflicts would grow exponentially**~
PR was merged. This PR is now ready for review.

This PR implements a basic support for `TsTypeAliasStatement` and `TsType`. Not all the branches of `TsType` are covered, otherwise the PR would have been too big.

The test file `type_expressions.ts` has all the grammar that is supported in this PR.

## Test Plan

This PR introduces a new typescript test.


crates/rome_formatter/src/ts/object_type_members/any_object_type_member.rs
```diff
@@ -0,0 +1,18 @@
...
+    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        match self {
+            TsAnyObjectTypeMember::TsCallSignatureObjectTypeMember(_) => todo!(),
+            TsAnyObjectTypeMember::TsConstructSignatureObjectTypeMember(_) => todo!(),
+            TsAnyObjectTypeMember::TsGetterSignatureObjectTypeMember(_) => todo!(),
+            TsAnyObjectTypeMember::TsIndexSignatureObjectTypeMember(_) => todo!(),
+            TsAnyObjectTypeMember::TsMethodSignatureObjectTypeMember(_) => todo!(),
+            TsAnyObjectTypeMember::TsPropertySignatureObjectTypeMember(node) => {
+                node.to_format_element(formatter)
+            }
```
ematipico, line 14: I left on purpose the rest of the branches, otherwise it would have been too big.

crates/rome_formatter/src/ts/object_type_members/property_signature.rs
```diff
@@ -0,0 +1,31 @@
...
+};
+use rslint_parser::ast::TsPropertySignatureObjectTypeMember;
+
+impl ToFormatElement for TsPropertySignatureObjectTypeMember {
+    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        let readonly = self.readonly_token().format_or_empty(formatter)?;
+        let name = self.name().format(formatter)?;
+        let type_annotation = self.type_annotation().format_or_empty(formatter)?;
+        let optional = self.optional_token_token().format_or_empty(formatter)?;
+        // Here, we can have two kind of separators: `,` or `;`.
```
ematipico, line 14: The comment is outdated. It seems and also ASI is a valid separator, but it's not covered by our parser yet.

crates/rome_formatter/src/ts/auxiliary/property_signature_object_type_member.rs
```diff
@@ -1,7 +1,31 @@
...
-        Ok(formatter.format_verbatim(self.syntax()))
+        let readonly = self.readonly_token().format_or_empty(formatter)?;
+        let name = self.name().format(formatter)?;
+        let type_annotation = self.type_annotation().format_or_empty(formatter)?;
+        let optional = self.optional_token_token().format_or_empty(formatter)?;
+        // Here, we can have two kind of separators: `,`, `;` or ASI.
+        // Because of how the grammar crafts the nodes, the parent to add the separator to the node.
+        // So here, we create - on purpose - an empty node.
+        let separator = if let Some(separator) = self.separator_token() {
+            formatter.format_replaced(&separator, empty_element())?
```
MichaReiser, line 18: Do I understand this correctly that we intentionally strip all delimiters? 

This would only be safe if whatever token comes next allows ASI (a new line, `}`...)

Edit: I see, it's formatted in the parent. I think it would be worth calling this out here. An alternative would be to use `self.syntax().next_sibling()`. which is `None` if this is the last element or `Some(`TypeMember`)`. You can use that knowledge to decide if the separator should be inserted (could be worth extracting a helper function). 

This is mainly beneficial if we decide to do CST to CST formatting to avoid constructing new nodes just because the separator token got replaced.

crates/rslint_parser/src/tests.rs
```diff
@@ -12,16 +12,12 @@ use std::path::{Path, PathBuf};
...
-    for (x = "" || 0; typeof x !== "string"; x = "" || true) {
-        x; // number | boolean
-    }
-}
+type A = string;
     "#;
 
     let module = parse(src, 0, Syntax::default().typescript());
 
+    dbg!(&module.syntax());
```
MichaReiser, line 20: Delete?

crates/rome_formatter/src/ts/auxiliary/ts_any_name.rs
```diff
@@ -0,0 +1,11 @@
+use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
+use rslint_parser::ast::TsAnyName;
+
+impl ToFormatElement for TsAnyName {
+    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        match self {
```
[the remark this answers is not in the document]
ematipico: True. This is the one that I created (and not even used). Good catch!

crates/rome_formatter/src/ts/types/object_type.rs
```diff
@@ -1,7 +1,25 @@
...
+            &self.l_curly_token()?,
+            |open_token_trailing, close_token_leading| {
+                let list = self.members().to_format_element(formatter)?;
+                Ok(format_elements![
+                    soft_block_indent(format_elements![
+                        open_token_trailing,
+                        list,
+                        close_token_leading
+                    ]),
+                    soft_line_break_or_space()
```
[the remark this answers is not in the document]
ematipico: Good catch! It's a bug

crates/rome_formatter/src/ts/auxiliary/property_signature_object_type_member.rs
ematipico: > I think it would be worth calling this out here.

What do you mean? Isn't the comment enough? (I am going to fix the wording)

---

The main reason I decided to leave the responsibility to the parent to the formatting of the separator, is because it's the only one that knows when to print the separator or not. We could have the logic here, as you suggested, but it doesn't know if it should print separator or not.

For example this case doesn't need a separator:
```ts
type a = { a: string, b: string }
```

We do actually have the function `format_separator`, but unfortunately it's a function made only for `AstSeparatedList` and `TsObjectTypeMemberList` is not one of them

MichaReiser: Fixing the wording helps.

You can determine if it's the last element in the list by using `self.syntax().next_sibling()`. It returns `None` if it's the last and `Some(Node)` otherwise. 

But the way you do it now is fine.... We can re-visit it if we ever decide to do CST -> CST printing.

leops (approved):

crates/rome_formatter/src/ts/auxiliary/property_signature_object_type_member.rs
leops: Out of curiosity what's the reason why `TsObjectTypeMemberList` does not implement `AstSeparatedList` ?

ematipico: > Out of curiosity what's the reason why `TsObjectTypeMemberList` does not implement `AstSeparatedList` ?

Because they don't have a fixed separator: it can be `,`, `;` or ASI. Being able to represent it inside the grammar as a `AstSeparatedList` wasn't possible. (@MichaReiser has more context)

MichaReiser: `TypeMember`s are special for a few reasons
* It allows multiple separators vs. just one: `,` or `;`. This can be supported by the `SyntaxFactory` if we add a `Separator` trait that implements a `match` method so that it can compare vs a single kind or a list of kinds
* Not having a separator is valid too in case ASI applies. Now, that's the case that the `AstSeparatedList` can't handle because the rule is that each node must be separated by a token which isn't the case if ASI applies.

PR #2071, opened 2022-02-08
chore: hide generated code

MichaReiser:
## Summary

Changes to the AST create changes in many generated files. These changes can be distracting from the real changes made by the author.

This PR sets the `linguist-generated=true` attribute for the files generated by the codegen so that they are hidden in diffs by default and no longer count to the repo's statistics.

[Source](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)

## Test Plan

You can see in #2072 that the generated files are hidden by default.


ematipico (approved):
Amazing change!

xunilrj (approved):

PR #2073, opened 2022-02-08
chore(vscode): categories and keywords

ematipico:
## Summary

This PR adds few metadata to the `package.json` of the VScode extension.

Here's a link to the page that explains each field: https://code.visualstudio.com/api/references/extension-manifest

## Test Plan

None

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -41,6 +41,8 @@
 	"keywords": [
 		"Formatter"
 	],
+	"license": "MIT",
+	"preview": true,
 	"scripts": {
 		"compile": "tsc -b",
 		"package": "vsce package -o rome_lsp.vsix",
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -35,6 +35,12 @@
 			}
 		}
 	},
+	"categories": [
+		"Formatters"
+	],
+	"keywords": [
+		"Formatter"
+	],
 	"scripts": {
 		"compile": "tsc -b",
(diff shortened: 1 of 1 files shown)


editors/vscode/package.json
```diff
@@ -35,6 +35,14 @@
...
 		}
 	},
+	"categories": [
+		"Formatters"
+	],
+	"keywords": [
+		"Formatter"
+	],
+	"license": "MIT",
+	"preview": true,
```
ematipico, line 45: I decided to  mark it as a preview, as it's still a work a progress. What do you think?

leops: I think it's okay for now, though eventually we'll probably want to patch the `package.json` on the fly to inject or remove this flag in CI before packaging depending on whether it's being built for a pre-release version or not

editors/vscode/package.json
```diff
@@ -35,6 +35,14 @@
 			}
 		}
 	},
+	"categories": [
+		"Formatters"
+	],
+	"keywords": [
+		"Formatter"
```
MichaReiser, line 42: Maybe add "JavaScript"?

ematipico: We have a limit of 5 keywords, we should make them count 😛 We can also replace `Formatter` with `JavaScript`. What do you think?

MichaReiser (approved):

yassere (approved):
Looks good to me. I agree that we can refine our list of keywords later on.

PR #2075, opened 2022-02-08
feat(rome_formatter): `enum` formatting

### LINKED ISSUE #2080: Format `TsEnumStatement`

ematipico:
## Summary

Closes #2080 

## Test Plan

New tests


crates/rome_formatter/src/ts/auxiliary/enum_member.rs
```diff
@@ -1,7 +1,18 @@
...
+
 impl ToFormatElement for TsEnumMember {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let name = self.name().format(formatter)?;
+        let initializer = self
+            .initializer()
+            .format_with_or_empty(formatter, |initializer| {
+                format_elements![space_token(), initializer]
+            })?;
```
MichaReiser, line 14: Nit: This is probably a very common pattern used everywhere where there is an initializer. Would it make sense to introduce a `format_initializer` helper?

crates/rome_formatter/src/ts/statements/enum_statement.rs
```diff
@@ -1,7 +1,41 @@
...
 impl ToFormatElement for TsEnumStatement {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let const_token = self
+            .const_token()
+            .format_with_or_empty(formatter, |const_token| {
+                format_elements![const_token, space_token()]
+            })?;
+        let enum_token = self.enum_token().format_with(formatter, |enum_token| {
+            format_elements![enum_token, space_token()]
```
MichaReiser, line 16: What's the benefit of using `format_with` over `let enum_token = format_elements![formatter.format_token(&self.enum_token)?, space_token()]`

ematipico: The APIs`format_token` and `format_node` don't exist anymore, because they were creating some friction: the developer needed to know in advance whether `self.something` was a token or a node. The traits `format_*`take care of it, thanks to the trait + type system provided by Rust.

Doing so, writing formatting rules becomes easier, and knowing when something is a node or a token is only needed in particular cases.

crates/rome_formatter/src/ts/auxiliary/enum_member.rs
ematipico: I checked the code base and I don't see many `initializer`. In some cases they also have a different semantic (`for` loop) and formatted differently. I ma not sure if it's worth it 🤔

MichaReiser: There should be plenty. You can search for `JsInitializerClause` in the `js.ungram` file.

ematipico: Oh, I see now. Sometimes they are called `init` and sometimes they are called `initializer`. Do you think we should use just one name? Then we can create one single helper to format them

MichaReiser: Aligning the name sounds good! 

But I don't think you need the rename to implement the helper

```
format_initializer(initializer: &JsInitializerClause) -> FormatResult {
  initializer.format_with_or_empty(....)
}
```

crates/rome_formatter/src/lib.rs
```diff
@@ -57,6 +57,7 @@ mod intersperse;
 mod js;
 mod printer;
 mod ts;
+mod utils;
```
leops, line 60: I think `interpreter.rs` could be merged into this new `utils` module, since it only contains a `format_interpreter` function similar to `format_initializer_clause`

ematipico: Done!

leops (approved):

ematipico:
I like the idea of having a `prelude`. I can make a new PR for that!

PR #2076, opened 2022-02-08
feat(rslint_parser): Parse ambient `enum` and `type` declarations

MichaReiser:
## Summary

Implements parsing of enums and type aliases in ambient contexts: 

```ts
declare enum A { X, Y }
declare const enum B { X, Y }
declare type C = A;
```

This PR doesn't implement parsing of enums inside other ambient contexts like:

```ts
declare namespace a {
  enum A { X, Y }
}
```

This PR reuses the `TsEnumStatement` and `TsTypeAliasStatement` and nests them inside of a `TsDeclareStatement`. This has the benefit that we don't need new nodes but comes at the downside that it may feel weird that a `statement` can appear in such a position. 

Another downside is that an analyzer needs to probe the enum's or alias statement's parent to figure out if it's inside of an ambient context or not.

## Test Plan
Added new parser tests.

Regressions: The regressions are around validating what a valid initializer is (ambient/non-ambient definition). These are things that get out of hand inside of the parser and should probably be validated inside a second pass.


crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -186,7 +186,7 @@ pub(crate) fn parse_statement(p: &mut Parser, context: StatementContext) -> Pars
         T![if] => parse_if_statement(p),
         T![with] => parse_with_statement(p),
         T![while] => parse_while_statement(p),
-        _ if is_at_ts_enum_statement(p) => parse_ts_enum_statement(p),
+        T![const] | T![enum] if is_at_ts_enum_statement(p) => parse_ts_enum_statement(p),
```
xunilrj, line 189: Now I am wondering if there is a rationale for the ordering here...
and if we should choose one.

xunilrj (approved):

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: You mean of the `match` branches or between `T![const] | T![enum]` 

The ordering of the match branches can be relevant if there's ambiguity. For example, `let` is either an identifier or the start of the `let` statement. 

I don't think the ordering of the match in a single branch should matter. Rust should pick the best ordering there.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -112,6 +114,18 @@ pub(crate) fn parse_ts_declare_statement(p: &mut Parser) -> ParsedSyntax {
         T![function] => {
             parse_ts_declare_function(p);
         }
+        T![const] | T![enum] => {
+            // test ts ts_ambient_enum_statement
+            // declare enum A { X, Y, Z }
+            // declare const enum B { X, Y, Z }
+            parse_ts_enum_statement(p).unwrap();
```
ematipico, line 121: Can we have a `// SAFETY` comment that explains why it's safe to use `unwrap`?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -112,6 +114,18 @@ pub(crate) fn parse_ts_declare_statement(p: &mut Parser) -> ParsedSyntax {
...
+        }
+        T![ident] if is_at_contextual_keyword(p, "type") => {
+            // test ts ts_declare_type_alias
+            // declare type A = string;
+            // declare type B = string | number & { a: string, b: number }
+            parse_ts_type_alias(p).unwrap();
```
ematipico, line 127: Can we have a `// SAFETY` comment that explains why it's safe to use `unwrap`?

ematipico (approved):

crates/rslint_parser/src/parser/parsed_syntax.rs
```diff
@@ -116,6 +125,20 @@ impl ParsedSyntax {
         }
     }
 
+    /// Returns the contained [ParsedSyntax::Present] value, consuming the self value.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value is an [ParsedSyntax::Absent] with a custom panic message provided by msg.
+    #[inline]
+    #[track_caller]
```
xunilrj, line 134: Nice touch!

Issue #2058, opened 2022-02-07
TS-Parsing: Interfaces, Declare Functions, Function Overloads


MichaReiser:
Implement parsing for 

- [x] Interfaces #2081 
- [x] ~Abstract methods~ #2059 
- [x] declared functions #2066 
- [x] function overload (`TsFunctionOverloadStatement`)

I grouped these together because they all require a representation of members/functions that don't have a function body.

Why add parsing for `declare` functions? The main reason is that many TS tests make use of the syntax... so it helps to support it as well to get a better test signal.

MichaReiser:
Done. Overloads implemented in #2087

PR #2081, opened 2022-02-09
feat(rslint_parser): Parse TS `interface` syntax

MichaReiser:
## Summary

Adds support for parsing TypeScript's `interface` syntax: 

```ts
interface A {
  prop: string;
  method() {}
}

interface B extends A {}
```

The PR renames the `ObjectTypeMember` nodes to `TypeMembers` because `type` and `interfaces` support the exact same members. 

The PR further refactors the class and interface heritage nodes (`extends`/`implements`): 

* `class`: 
  * `extends`: Allows arbitrary expressions with type arguments. Added type arguments to `JsExtendsClause`
  * `implements`: Permits a list of TypeScript names (`A` or `namespace.B`) with optional type arguments. Introduced `TsImplementsClause`. Renamed `TsExprWithTypeArgs` to `TsNameWithTypeArguments`
* `interface`: 
  * `extends`: Same as `class` implements. Introduced `TsExtendsClause`
  * `implements`: Not permitted

Part of #2058 

## Test Plan

Added new parser tests. 

Regressions. I didn't go through all regressions. There are too many. I checked a few and these are all tests that are supposed to fail and used to fail because the parser failed parsing the `interface` syntax.


crates/rslint_parser/src/syntax/typescript/statement.rs
```diff
@@ -0,0 +1,382 @@
...
+    if let Present(body) = parse_function_body(p, SignatureFlags::empty()) {
+        p.error(
+            p.err_builder("A 'declare' function cannot have a function body")
+                .primary(body.range(p), "remove this body"),
+        );
+    }
+
+    semi(p, start_range..p.cur_tok().start());
+    m.complete(p, TS_DECLARE_FUNCTION)
+}
```
MichaReiser, line 305: Moved here from `enum` and `typescript.rs`

crates/rslint_parser/src/parser.rs
```diff
@@ -242,6 +242,7 @@ impl<'t> Parser<'t> {
     }
 
     /// Make a new error builder with `error` severity
+    #[must_use]
```
ematipico, line 245: Good call!

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -190,120 +189,115 @@ fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
...
+// class A extends bar extends foo {}
+// class B extends bar, foo {}
+// class C implements B {}
+//
+// test_err ts ts_class_heritage_clause_errors
+// class A {}
+// interface Int {}
+// class B implements Int extends A {}
+// class B implements Int implements Int {}
+fn parse_class_heritage_clause(p: &mut Parser) {
```
ematipico, line 211: Is it fine that it doesn't return a `ParsedSyntax`? Just curiosity, because its name is `parse_*`.
Also, do you think it's worth adding a small doc? "Heritage" is not reflected on any node. First time I saw it, it took me some time to understand what it was doing

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -190,120 +189,115 @@ fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
...
+    let mut first_extends: Option<CompletedMarker> = None;
+    let mut first_implements: Option<CompletedMarker> = None;
 
-        let err = p
-            .err_builder("classes cannot have multiple `implements` clauses")
-            .primary(start..p.marker_vec_range(&elems).end, "");
+    loop {
+        if p.at(T![extends]) {
+            let current = parse_extends_clause(p)
+                .expect("Expected extends clause because parser is positioned at extends keyword");
```
ematipico, line 218:
```suggestion
                .expect("Expected extends clause because parser is positioned at `extends` keyword");
```

crates/rslint_parser/src/syntax/stmt.rs
```diff
@@ -228,8 +228,12 @@ pub(crate) fn parse_statement(p: &mut Parser, context: StatementContext) -> Pars
...
         }
 
-        T![ident] if p.cur_src() == "type" && p.typescript() => parse_ts_type_alias(p),
-        // TODO: handle `<T>() => {};` with less of a hack
+        T![ident] if p.cur_src() == "type" && p.typescript() => parse_ts_type_alias_statement(p),
+        T![ident] if is_at_ts_interface_statement(p) => {
+            TypeScript.parse_exclusive_syntax(p, parse_ts_interface_statement, |p, interface| {
+                ts_only_syntax_error(p, "interface", interface.range(p).as_range())
+            })
+        }
```
ematipico, line 236: Do you think it would actually make sense to "merge" these two branches?

```rs
T![ident] if p.typescript() => {
	if p.cur_src() == "type" {}
	else if is_at_ts_interface_statement(p) {} 
	else { Absent }
}
```

Just for readability

xtask/codegen/js.ungram
```diff
@@ -100,9 +100,10 @@ JsAnyStatement =
 	| JsVariableStatement
 	| JsWhileStatement
 	| JsWithStatement
-	| TsEnumStatement
   | JsFunctionStatement
+	| TsEnumStatement
```
ematipico, line 104: Formatting issue I think

ematipico (approved):

crates/rslint_parser/src/syntax/class.rs
MichaReiser: You're paying attention to everything :) I think this is fine. It's internal only. I'm using `expect` instead of a safety comment + `unwrap`.

xtask/codegen/js.ungram
MichaReiser: I believe that is because of the `+` sign.

crates/rslint_parser/src/syntax/stmt.rs
MichaReiser: There are more cases further down that must be handled in case this `ident` isn't the contextual `interface` keyword.

PR #2082, opened 2022-02-09
fix(rslint_parser): Template literal error recovery

MichaReiser:
## Summary

The following typescript test panicked.

```ts
for (let i = 0 ; i < 3; ++i) {
    verify.completions({
        marker: `${i + 1}`,
        exact: [
            { name: "foo", replacementSpan: test.ranges()[i] },
            { name: "bar", replacementSpan: test.ranges()[i] },
        ]
    });
}
```

It revealed two problems:

1. The `parse_conditional_expr` tried to parse the `<` as a call expression with type arguments. The problem was that `parse_type_arguments` performed error recovery when it wasn't able to parse a valid type name. Thus, `parse_ts_type_parameters` ended up parsing the `${i + 1}` as a typescript template literal type
2. `${i + 1}` isn't a valid TypeScript template literal. The parser recognised the template element and then tried to parse `i + 1` as a type. This consumed the `i` but left the `+ 1` tokens. Now, the parser panicked because the parse loop assumes that only specific tokens ever come after a `BACKTICK`.

This PR fixes both issues by:
* Not performing error recovery in the branch where the parser speculatively parses type arguments.
* Skip over any invalid tokens inside `TemplateElement`s to ensure the parser doesn't hit the unreachable branch.

This PR further removes the `no_recovery` flag. This is only needed in this specific case. Therefore, passing it down is easier to maintain.

## Test Plan

Added new parser tests covering the error recovery branch.

### DIFF
--- a/crates/rslint_parser/src/parser/single_token_parse_recovery.rs
+++ b/crates/rslint_parser/src/parser/single_token_parse_recovery.rs
@@ -39,12 +39,8 @@ impl SingleTokenParseRecovery {
         let error = self.get_error();
         if let Some(error) = error {
             p.error(error);
-        } else {
-            // the check on state should be done only when there's no error
-            if p.state.no_recovery {
-                return;
-            }
         }
+
         if !self.parsing_is_recoverable(p) {
             let m = p.start();
             p.bump_any();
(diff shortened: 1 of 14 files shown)


ematipico (approved):
Looks good to me

MichaReiser:
!bench_parser

PR #2084, opened 2022-02-09
feat(rome_formatter): format `TsAnyTypeMember`

ematipico:
## Summary

Part of #2069 

This PR implements formatting for all types include in the union `TsAnyTypeMember`:

- `TsCallSignatureTypeMember`
- `TsPropertySignatureTypeMember`
- `TsConstructSignatureTypeMember`
- `TsMethodSignatureTypeMember`
- `TsGetterSignatureTypeMember`
- `TsSetterSignatureTypeMember`
- `TsIndexSignatureTypeMember`

I changed the `optional_token` to be `optional`. This was causing the AST to have a `optional_token_token`, which was redundant. I updated the RAST files too.

## Test Plan

Created new tests to cover long types and variables and make sure that their formatting is similar to what prettier does.


crates/rome_formatter/src/ts/auxiliary/setter_signature_type_member.rs
```diff
@@ -1,7 +1,27 @@
...
+use rslint_parser::ast::TsSetterSignatureTypeMember;
+
 impl ToFormatElement for TsSetterSignatureTypeMember {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let set = self.set_token().format(formatter)?;
+        let name = self.name().format(formatter)?;
+        let l_paren = self.l_paren_token().format(formatter)?;
+        let r_paren = self.r_paren_token().format(formatter)?;
+        let parameter = self.parameter().format(formatter)?;
```
MichaReiser, line 14: Nit: Format the nodes in appearance order: `r_paren` after `parameter`

crates/rome_formatter/src/utils.rs
```diff
@@ -6,6 +6,23 @@ use crate::{
 use rslint_parser::ast::JsInitializerClause;
 use rslint_parser::SyntaxToken;
 
+/// Utility function to format the separators of the nodes that belong to the unions
+/// of [rslint_parser::ast::TsIndexSignatureObjectTypeMember].
```
MichaReiser, line 10: Nit: You can reference the `TsAnyTypeMember` union here instead of one of its variants.

crates/rome_formatter/tests/specs/ts/expression/type_member.ts.snap
```diff
@@ -0,0 +1,152 @@
...
+	b?(): number,
+	c?(): number,
+	d(): string,
+	bvvvvvvvvvvvvvvvvvvvvvv?(
+		loreum: string,
+		ipsum: symbol,
+		lapis: symbol,
+		emerald: symbol,
+	): G,
+};
```
MichaReiser, line 134: It's really nice to see how far our formatter came in the last few weeks. From something close to a minifier to a formatter that produces readable code :)

crates/rslint_lexer/src/tests.rs
```diff
@@ -47,7 +47,6 @@ macro_rules! assert_lex {
         )*
 
         if idx < tokens.len() {
-            dbg!(&tokens);
```
MichaReiser, line 50: Can we keep this. This `dbg` is intentionall. Feel free to replace it with a `println` if you don't like `dbg`. Having the `dbg` helps understand which tokens are missing and it only ever runs if the test fails.

crates/rslint_parser/test_data/inline/ok/ts_declare_function.rast
```diff
@@ -124,7 +124,7 @@ JsModule {
                                             name: JsLiteralMemberName {
                                                 value: IDENT@80..81 "a" [] [],
                                             },
-                                            optional_token_token: missing (optional),
```
MichaReiser, line 127: duh!

crates/rslint_lexer/src/tests.rs
ematipico: Oh, I thought it was a leftover! I can leave it!

crates/rome_formatter/src/ts/bindings/type_parameters.rs
```diff
@@ -1,7 +1,27 @@
...
 impl ToFormatElement for TsTypeParameters {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let items = formatter.format_separated(self.items(), || token(","))?;
+
+        Ok(group_elements(formatter.format_delimited(
+            &self.l_angle_token()?,
+            |open_token_trailing, close_token_leading| {
+                Ok(format_elements![
+                    if_group_breaks(soft_line_break()),
```
leops, line 15: That looks redundant, soft line breaks already don't print anything if the enclosing group doesn't break

leops (approved):
Besides the remaining lint errors it seems good to go

PR #2087, opened 2022-02-10
feat(rslint_parser): TS function overload syntax

MichaReiser:
## Summary
TypeScript allows function overloading by having a function statement without a body followed by a function statement with the same name and a body

```ts
function test();
function test(a: string | undefined) {}
```

This PR re-uses the `TsDeclareFunction` node (and renames it to `TsDeclareFunctionStatement`) to represent function overloads.
This comes at the downside that `declare` functions now have an optional `async` child even though they can't be async.

The PR regresses the *no-optional pattern parameter for function parameters* it's no longer known upfront if a function is an implementation or a function overload.
It would be possible to use a checkpoint and rewind the parser in case it turns out to be a function overload definition, but I want to create a snapshot in such a hot path and believe that such a check should be part of a second-pass.

## Test Plan

Added new parser tests


MichaReiser:
!bench_parser

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -194,7 +193,16 @@ fn parse_function(p: &mut Parser, m: Marker, kind: FunctionKind) -> CompletedMar
         })
         .ok();
 
-    parse_parameter_list(p, ParameterContext::Implementation, flags)
+    let parameter_context = if kind.is_statement() && TypeScript.is_supported(p) {
+        // It isn't known at this point if this a function overload definition (body is missing) or a regular
+        // function implementation. Let's go with the laxer of the two. Ideally, these verifications should be
+        // part of a second compiler pass.
```
ematipico, line 199:
```suggestion
        // It isn't known at this point if this is a function overload definition (body is missing) or a regular
        // function implementation. Let's go with the latter of the two. Ideally, these verification should be
        // part of a second compiler pass.
```

[Plural of `verification` is `verification`](https://www.wordhippo.com/what-is/the-plural-of/verification.html) 😮

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -204,23 +212,50 @@ fn parse_function(p: &mut Parser, m: Marker, kind: FunctionKind) -> CompletedMar
         })
         .ok();
 
-    if kind.is_statement() {
-        function_body_or_declaration(p, flags);
+    let body = parse_function_body(p, flags);
+
+    // test ts ts_function_overload
```
ematipico, line 217: I see now that this is the new way of creating TS files. Is it documented somewhere by any chance?

ematipico (approved):
Now that I see this implementation, I think a second pass is more important than before. More specifically, we need a second pass to create scopes with their bindings, which is something that other compilers do (esbuild does it for example). 

This reason why I say this is basically the fact that **in this particular case**, TS allows a double scoped binding while, in reality, is only one.

Essentially, in JS this throws an error in strict mode:
```js
function test() {}
function test(a) {}
``` 

but TS allows it, IF it is a function overload. I guess we should discuss it ASAP and see what we can do.

crates/rslint_parser/src/syntax/function.rs
MichaReiser: I don't know. @xunilrj did you mention this in the `CONTRIBUTING.md`?

MichaReiser:
> but TS allows it, IF it is a function overload. I guess we should discuss it ASAP and see what we can do.

This is correct but using a `JsIdentifierBinding` is required because `declare function test()` binds the identifier if it's the first declaration. The same is actually true for an overload, it's just that you get a compile error if the implementation is missing. 

I don't think this is problematic because we need a second symbol resolution pass anyway and pass authors shouldn't implement their own symbol resolution but instead use the built in one (or they have to deal with this).

PR #2095, opened 2022-02-11
fix: correct line comments spacing

leops:
## Summary

This corrects the printing of newlines between comments in leading trivias and the token they're attached to by inserting an empty line instead of a hard line break when 2 or more newlines are found between the comment and the token

### DIFF
--- a/crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
+++ b/crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/spec_test.rs
-assertion_line: 157
+assertion_line: 163
 expression: if_else.js
 
 ---
@@ -114,7 +114,6 @@ if (true) {
 
 if (true) {
 	// trailing
-
 } else {
 	// trailing
 }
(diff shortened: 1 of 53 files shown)


crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
```diff
@@ -114,6 +114,7 @@ if (true) {
 
 if (true) {
 	// trailing
+
```
MichaReiser, line 117: This might be one of the weird exceptions to your rule. Seems to me that the previous formatting felt nicer than the new one. We may want to add some more special casing in the future (if we know that the comment is in front of a closing token: `}`, `]` and maybe `)`

MichaReiser (approved):

crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
leops: I think a simple solution for this would be to call trim_start and trim_end on the leading and trailing trivia in format_delimited

leops:
Right so this PR preserves the empty line between the comment and the token in this situation:
```js
// comment

statement();
```
Previously the formatter would have output this:
```js
// comment
statement();
```

ematipico (approved):

crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
leops: `trim_end` ended up no working correctly as it also strips the hard line break normally inserted after a single line comment, so I opted to add an additional trim_mode argument to `print_leading_trivia` to let a caller decided whether the function should print or trim the whitespace separating the token from the first comment

PR #2096, opened 2022-02-11
refactor: move the implementations of list formatting

leops:
## Summary

This moves the formatting of lists (generally calling `format_separated` on it) to the `ToFormatElement` implementation of the list itself instead of the parent node. This does not change anything for full-document formatting but is important for range formatting as the range being formatted can often cover just the list without having to reformat the entire parent element, so the formatter has to be able to pick up the formatting from there.

There are also a few additional changes with the most important being to have the parent nodes check whether the list is empty by looking at the resulting FormatElement + the potential leading and trailing trivias of the group to correctly account for potential comments or newlines inside the group


ematipico (approved):
Looks good to me. CI is failing though

Issue #2090, opened 2022-02-10
Parse TS `namespace` / `module` syntax


MichaReiser:
Parsing of TS `namespace` and `module` syntax ([documentation](https://www.typescriptlang.org/docs/handbook/namespaces.html))

```ts
module test {}

declare module "shorthand";

// creates nested namespaces
declare module test.b.c.d {}

namespace b {}
declare namespace abcd.c.d.e {}


// TsShorthandDeclareModuleStatement 'declare' 'module' id: string ';'?
// TsModuleStatement = 'declare'? 'module' id: (string | TsIdentifierBinding) body: TsAnyModuleBody
// TsAnyModuleBody = JsBlock | .id
// TsNamespaceStatement = 'declare'? namespace  TsIdentifierBinding body: JsBlock
```

MichaReiser:
I'm currently planning to go with the following grammar: 

```
// TsShorthandModuleDeclaration = 'module' id: string ';'?
declare module "test";

// TsModuleDeclaration = 'module' name: TsAnyModuleName body: JsBlock
declare module test {}
declare module a.b.c.d {}
declare module "test" {}
module test {}
module a.b.c.d {}
module "test" {} // Not allowed

// TsNamespaceDeclaration = 'namespace' name: TsAnyName body: JsBlock
namespace test {}
namespace a.b.c.d {}
declare namespace test {}
declare namespace a.b.c.d {}

// TsAnyModuleName = string | TsIdentifierBinding | TsQualifiedName

// TsDeclareStatement = 'declare' declaration: JsAnyDeclaration

// JsAnyDeclaration = JsFunctionDeclaration | JsClassDeclaration | JsVariableDeclaration | TsEnumDeclaration | TsDeclareFunctionDeclaration | TsTypeAliasDeclaration | TsInterfaceDeclaration 
```

MichaReiser:
> [This is how TS represents](https://ts-ast-viewer.com/#code/HYQwtgpgzgDiDGEAEIB0AjV9UBMkG8BfIA):
> 
> ```ts
> declare module a.b.c.d {}
> ```

Jep. I looked at it and find it weird. Making this work with a CST would require another node because the inner node now has a `.` token but misses the `module` token. 

What I describe is the same as how roslyn represents `namespace System.Console`.

ematipico:
> > [This is how TS represents](https://ts-ast-viewer.com/#code/HYQwtgpgzgDiDGEAEIB0AjV9UBMkG8BfIA):
[...]

PR #2098, opened 2022-02-11
refactor: `JsAnyDeclarationClause`

MichaReiser:
## Summary

This PR introduces a new `JsAnyDeclarationClause` for declarations in non-statement positions. The union variants are the same as `JsAnyDeclaration` with the exception of `JsVariableDeclarationClause` that has an optional semicolon (same as `VariableStatement`). 

Can we add the semicolon to the `VariableDeclarations` node? No, I tried that. The problem is that `VariableDeclarations` are also valid inside the initialiser of a `for (let a, b;;) {}` loop and a semicolon isn't allowed in that case (or the parser eats the for separator semicolons. 

This PR further re-introduces the `TsDeclareStatement` for all `declare JsAnyDeclarationClause` declarations and adds `export type`, and `export enum` support. 

Part of #2090

## Test Plan

Added new test cases. Verified rast changes by hand


crates/rome_formatter/src/js/any/export_clause.rs
```diff
@@ -5,15 +5,13 @@ use rslint_parser::ast::JsAnyExportClause;
 impl ToFormatElement for JsAnyExportClause {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
         match self {
-            Self::JsExportFunctionClause(node) => node.to_format_element(formatter),
```
MichaReiser, line 8: These can now all be represented by `JsAnyDeclarationClause`, in the case of a function, the normal `JsFunctionDeclaration`.

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -473,59 +467,49 @@ pub(super) fn parse_export(p: &mut Parser) -> ParsedSyntax {
...
-            parse_export_function_clause(p)
-        }
-        T![*] => parse_export_from_clause(p),
-        T![ident] if p.cur_src() == "from" => parse_export_from_clause(p),
-        _ if p.nth_at(1, T![ident]) && p.nth_src(1) == "from" => parse_export_from_clause(p),
-        _ if is_at_variable_declarations(p) => parse_export_variable_clause(p),
-        _ => Absent,
     };
 
+    dbg!(p.cur());
```
xunilrj, line 506: delete?

ematipico (approved):

PR #2099, opened 2022-02-11
feat: implement destructuring for AST nodes

leops:
## Summary

This PR adds a new `as_slots` method to all generated AST nodes. This method calls all the accessor methods of the node its implemented for, and return their results in an associated `<node>Slots` struct that consumers of this API are expected to bind to a destructuring pattern. For example:

```rust
let JsObjectExpressionSlots { l_curly_token, members, r_curly_token } = expr.as_slots();
```

This is primarily aimed at the formatter to help statically ensure that no child node or token is skipped when formatting an ASTNode, as Rust will emit an error if a field is missing in a destructuring pattern, and a warning if the resulting binding is unused

### DIFF
--- a/xtask/codegen/src/generate_nodes.rs
+++ b/xtask/codegen/src/generate_nodes.rs
@@ -104,6 +104,49 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
 
             let string_name = name.to_string();
 
+            let slots_name = format_ident!("{}Fields", node.name);
+
+            let (slot_fields, slot_constructors): (Vec<_>, Vec<_>) = node
+                .fields
+                .iter()
+                .map(|field| match field {
+                    Field::Token { name, kind, .. } => {
+                        let many = matches!(kind, TokenKind::Many(_));
+
+                        let method_name = if many {
+                            format_ident!("{}", name)
+                        } else {
+                            field.method_name()
+                        };
+
+                        let is_optional = field.is_optional();
+
+                        let field = if is_optional {
+                            quote! { #method_name: Option<SyntaxToken> }
+                        } else {
+                            quote! { #method_name: SyntaxResult<SyntaxToken> }
+                        };
+
+                        (field, quote! { #method_name: self.#method_name() })
+                    }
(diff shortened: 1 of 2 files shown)


xtask/codegen/src/generate_nodes.rs
```diff
@@ -124,9 +167,18 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
...
+                        pub fn as_slots(&self) -> #slots_name {
+                            #slots_name {
+                                #( #slot_constructors, )*
+                            }
+                        }
+
                         #(#methods)*
                     }
 
+                    pub struct #slots_name {
```
xunilrj, line 179: Do you think we can avoid the copying by doing:

```rust
pub struct #slots_name<'a> {
    #( pub #slot_fields<'a>, )*
}
```

xunilrj:
Quite an interesting solution!

xtask/codegen/src/generate_nodes.rs
leops: This isn't really making copies since the struct being returned holds children nodes, so internally what it really does is bump the reference count of `self` a few times and allocates new `SyntaxNodes` to fill the struct.
I should probably make the naming more clear since "slots" is a concept mostly internal to the parser and Rowan, renaming the method to `as_children` could make what it is actually doing easier to understand

ematipico: What do you think of `extract_children`?

MichaReiser: The `term` children is somewhat overloaded because `SyntaxNode.children` returns the present children. The method here would differ as it returns the present as well as the absent "children". 

I do agree with you that `slot` is a rowan internal name and we should probably rename the `slot` method as well to whatever term we agree. 

What about `as_fields`?

ematipico (approved):
Let's merge it, I really like this solution

leops:
I'm splitting this into two PR, this one will just be adding the codegen and I'll perform the actual refactor in a followup

PR #2100, opened 2022-02-14
feat(rslint_parser): Parse TS Namespace syntax

MichaReiser:
## Summary

Adds support for parsing TypeScript's namespace syntaxes: 

```
namespace X {}
module Y {}

declare module "./abcd" {
  global { }
}
```

This PR doesn't cover

* Validation that statements in the body of a namespace are valid in an ambient context (2nd compiler pass?)
* Validation that the namespace syntax is valid in that context.
  * `global` is only valid inside of a `declare module".source" {}` declaration 
  * `export` doesn't allow for ambient declarations... (except if inside of an ambient context)


MichaReiser:
!bench_parser

MichaReiser:
!bench_parser

MichaReiser:
!bench_parser

xtask/codegen/js.ungram
```diff
@@ -1552,6 +1561,50 @@ TsDeclareFunctionDeclaration =
...
+
+// Can appear on statement level if inside of an ambient context
+// declare module "jest";
+//         ^^^^^^^^^^^^^^
+// declare module "jest" { ... }
+//         ^^^^^^^^^^^^^^^^^^^^^
+TsExternalModuleDeclaration =
+	'module'
+	source: JsModuleSource
+	body: TsAnyExternalModuleDeclarationBody?
```
ematipico, line 1600: The example
```ts
declare module "jest" {};
``` 
**can** contain `;` at the end, so wouldn't it be simpler having  
```
TsExternalModuleDeclaration =
	'module'
	source: JsModuleSource
	body: TsModuleBlock?
	comma: ';'?
```

crates/rslint_parser/src/syntax/typescript/statement.rs
```diff
@@ -214,15 +218,17 @@ pub(crate) fn parse_ts_declare_statement(p: &mut Parser) -> ParsedSyntax {
         return Absent;
     }
 
+    let stmt_start_pos = p.cur_tok().start();
     let m = p.start();
     expect_contextual_keyword(p, "declare", T![declare]);
 
-    parse_declaration_clause(p, true)
+    parse_declaration_clause(p, true, stmt_start_pos)
```
ematipico: I think the ambient signal should arrive from elsewhere. Possibly from `Syntax`.

The reason why I say this, is because `.d.ts` files are ambient files (always) but `.ts` **can** have ambient context. This means that when we read `.d.ts` files, we **always** are in ambient context regardless of the grammar.

crates/rslint_parser/src/syntax/auxiliary.rs
```diff
@@ -55,10 +56,18 @@ pub(crate) fn is_nth_at_declaration_clause(p: &Parser, n: usize) -> bool {
...
+        return true;
+    }
+
     false
 }
 
-pub(crate) fn parse_declaration_clause(p: &mut Parser, ambient: bool) -> ParsedSyntax {
+pub(crate) fn parse_declaration_clause(
+    p: &mut Parser,
+    ambient: bool,
```
ematipico: As suggested before, I think `ambient` should be computed via state (using `Syntax`)

crates/rslint_parser/src/syntax/typescript/statement.rs
```diff
@@ -362,3 +369,157 @@ fn parse_ts_extends_clause(p: &mut Parser) -> ParsedSyntax {
...
+    stmt_start_pos: usize,
+) -> ParsedSyntax {
+    if !is_at_contextual_keyword(p, "namespace") && !is_at_contextual_keyword(p, "module") {
+        return Absent;
+    }
+
+    let m = p.start();
+
+    if !eat_contextual_keyword(p, "namespace", T![namespace]) {
+        expect_contextual_keyword(p, "module", T![module]);
```
ematipico, line 406: I found this strange while reading the code. Why not doing this?

```rs
if eat_contextual_keyword(p, "module", T![module]) {}
```

crates/rslint_parser/src/syntax/typescript/statement.rs
```diff
@@ -362,3 +319,157 @@ fn parse_ts_extends_clause(p: &mut Parser) -> ParsedSyntax {
...
+
+        if p.at(JS_STRING_LITERAL) {
+            parse_module_source(p).expect("expected module source to be present because parser is positioned at a string literal");
```
ematipico, line 409: `.expect` here panics. Is it the correct behaviour that we want?

crates/rslint_parser/src/syntax/typescript/statement.rs
MichaReiser: I changed it and moved ambient into the `ParserState`

crates/rslint_parser/src/syntax/typescript/statement.rs
MichaReiser: Because either `namespace` or `module` must be present

crates/rslint_parser/src/syntax/typescript/statement.rs
MichaReiser: Yes. I'm using `expect` rather than `unwrap` plus a safety comment because it at least has a useful error message in case the invariant is violated. `parse_module_source` never returns `Absent` if the parser is at a `JS_STRING_LITERAL`, which we assert the parser is at on line 408,.

ematipico (changes requested):
I think the ambient context should be defined at `Syntax` level. Maybe we can create a new `FileKind::TypeScriptDeclaration` or something similar.

Then, the new `EnterAmbientContext` should change the bigflag only when we are not inside `FileKind::TypeScriptDeclaration`. 

The default value of `AMBIENT_CONTEXT` should be computed from `Syntax`

crates/rslint_parser/src/syntax/typescript/statement.rs
ematipico: Then, I am failing to understand what case we are handling here. Can you give me some example?

MichaReiser:
> I think the ambient context should be defined at `Syntax` level. Maybe we can create a new `FileKind::TypeScriptDeclaration` or something similar.
> 
> Then, the new `EnterAmbientContext` should change the bigflag only when we are not inside `FileKind::TypeScriptDeclaration`.
> 
> The default value of `AMBIENT_CONTEXT` should be computed from `Syntax`

I pushed a new change that adds `InAmbient` to the `ParserState`. I'm reluctant to change `Syntax` when encountering an ambient context and rather use `ParserState` which we already use heavily to store other "global" state.

crates/rslint_parser/src/syntax/typescript/statement.rs
MichaReiser: In practice, it shouldn't make a difference because this function is only ever called if we have `declare module` or `declare namespace` (or `module` and `namespace)`. However, I've written the function in a safe way that it yields a diagnostic if it is called to parse: 

```
declare module abcd {}
```

This gives you a diagnostic that `module` was expected but it found `modul`

ematipico:
> > I think the ambient context should be defined at `Syntax` level. Maybe we can create a new `FileKind::TypeScriptDeclaration` or something similar.
> > Then, the new `EnterAmbientContext` should change the bigflag only when we are not inside `FileKind::TypeScriptDeclaration`.
> > The default value of `AMBIENT_CONTEXT` should be computed from `Syntax`
> 
> I pushed a new change that adds `InAmbient` to the `ParserState`. I'm reluctant to change `Syntax` when encountering an ambient context and rather use `ParserState` which we already use heavily to store other "global" state.

I didn't mean that we should change `Syntax` inside the `Parser`. I meant we should take in consideration `Syntax` when calling `EnterAmbientContext`. If `Syntax::file_kind` is `FileKind::TypeScriptDeclaration` we should not change the bigflag and carry on with parsing while we are in ambient context.

For example, a `.d.ts` file can't have this syntax:

```ts
function test(); // Top-level declarations in .d.ts files must start with either a 'declare' or 'export' modifier.ts(1046)
```

We can apply this change using a different PR, but I think that the implementation is not finished yet.

MichaReiser:
> We can apply this change using a different PR, but I think that the implementation is not finished yet.

I agree, there's much more to ambient parsing. The goal of this PR wasn't to add full ambient parsing support. There are many weird cases that are and are not supported of which most probably fit best into a 2nd compiler pass:

```
global {} // only valid in d.ts files
export declare "./test" // invalid 
... and more
```

ematipico (approved):

xunilrj (approved):

PR #2102, opened 2022-02-15
feat(rslint_parser): TypeScript export extensions

MichaReiser:
## Summary

Adds support for the following export extensions of TypeScript:

```ts
export type { a };
export { type a }
export { type a as b };
export type { a } from "./a";
export { type a } from "./a";
export { type a as aa } from "./a";
export = b;
export as namespace "mathLib";
```

This PR doesn't add any additional checks if `export = b` or `export as namespace` are used outside of an ambient context OR in combination with other export statements. 

This PR doesn't add `export import ...` syntax support. I'll tackle this as part of the import syntax extensions

## Test Plan

Added new parser tests.


MichaReiser:
!bench_parser

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -594,36 +653,127 @@ impl ParseSeparatedList for ExportNamedSpecifierList {
     }
 }
 
-fn parse_export_named_specifier(p: &mut Parser) -> ParsedSyntax {
+fn parse_any_export_named_specifier(p: &mut Parser) -> ParsedSyntax {
+    if !is_nth_at_reference_identifier(p, 0) {
+        return Absent;
+    }
```
ematipico, line 659: We have `is_nth_at_reference_identifier` and `is_at_identifier` (they both call `is_nth_at_identifier`). Should we align their names?

ematipico (approved):
Looks good to me. I just left a tiny comment

crates/rslint_parser/src/syntax/module.rs
MichaReiser: That's intentional. `JsReferenceIdentifier` is its own node and the fact that the parsing is backed by `parse_identifier` is an implementation detail. The same applies for the `is_` method. A caller shouldn't need to know that `is_reference_identifier` has the exact same rules as `is_identifier`.

xunilrj (approved):

PR #2103, opened 2022-02-15
chore: wip

ematipico:


crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,445 @@
...
+                        current_index = index;
+
+                        group.push(item.into())
+                    }
+                } else {
+                    break;
+                }
+            }
+        }
+    }
```
ematipico, line 170: @leops This is the case where it was difficult to use an iterator. I tried to use `Peekable::next_if`, although `Peekable` only allows to check the next item without consuming it. In out case we want to check the next item and the one after. 

I also thought about copying the current array but it seems too much memory consuming. If you're aware of a possible solution using iterators, that would be amazing.

crates/rome_formatter/tests/specs/prettier/comments/issues.js.snap
```diff
@@ -114,11 +114,14 @@ const regex = new RegExp(
 import path from 'path'; // eslint-disable-line nuclide-internal/prefer-nuclide-uri
 
 // Comments disappear in-between MemberExpressions
-Observable.of(process)// Don't complete until we say so!
-.merge(Observable.never())// Get the errors.
-.takeUntil(throwOnError ? errors.flatMap(Observable.throw) : errors).takeUntil(
-  exit,
-);
+Observable
+  .of(process)
```
ematipico, line 118: This is a known issue, we should keep `Observable.of` on the same line. Will be tackled in another PR

crates/rome_formatter/tests/specs/prettier/last-argument-expansion/arrow.js.snap
```diff
@@ -27,22 +27,25 @@ export default function searchUsers(action$) {
...
+          800,
+        ).takeUntil(action$.ofType(ActionTypes.CLEARED_SEARCH_RESULTS)).mergeMap( // debounce
+          () =>
+            Observable.merge(
+              Observable.of(replace(`?q=${q}`)),
+              ajax.getJSON(`https://api.github.com/search/users?q=${q}`).map(
+                (res) => res.items,
+              ).map(receiveUsers),
+            ),
+        ),
```
ematipico, line 46: I will review this formatting in a separate PR

crates/rome_formatter/src/utils/call_expression.rs
leops: For this specific case since the input to the function is a slice, these loops could operate on array indices directly (either by iterating on `0..flatten_items.len()` ranges or mutating current_index in place) which might be easier than using peekable iterators.
On allocations, I think this function could return a `(&[FlattenItem], &[FlattenItem])` (resulting from a call to `flatten_items.split_at(index)`) and avoid allocating a new vector entirely

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    Ok((group, current_index))
+}
+
+/// computes groups coming after the first group
+fn compute_groups(flatten_items: &[FlattenItem], current_index: usize) -> FormatResult<Groups> {
+    let filtered_items = flatten_items
```
leops, line 177: This could be written as `&flatten_items[current_index..].iter()`, or as I mentioned above have the `compute_first_group` split the `flatten_items` slice earlier so this function only gets the items it's supposed to operate on in the first place

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// Holds onto a [rslint_parser::ast::JsCallExpression]
+    CallExpression(JsCallExpression, Vec<FormatElement>),
+    /// Holds onto a [rslint_parser::ast::JsComputedMemberExpression]
+    ComputedExpression(JsComputedMemberExpression, Vec<FormatElement>),
+    /// Any other node that are not  [rslint_parser::ast::JsCallExpression] or [rslint_parser::ast::JsStaticMemberExpression]
+    /// Are tracked using this variant
+    Node(SyntaxNode, FormatElement),
+}
+
+impl From<&FlattenItem> for FlattenItem {
```
leops, line 232: Isn't this just reimplementing `Clone` ?

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,445 @@
...
+            }
+    }
+
+    /// It strictly checks if the current item is a [rslint_parser::ast::JsCallExpression]
+    pub fn is_call_expression(&self) -> bool {
+        matches!(&self, FlattenItem::CallExpression(_, _))
+    }
+
+    /// It returns a [rome_formatter::FormatElement] of the current flatten item
+    pub fn get_format_element(self) -> FormatElement {
```
leops, line 268: Nit: call this `into_format_element` (or just implement `From<FlattenItem> for FormatElement`)

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// adds the passed element to the current group
+    pub fn continue_group<I: Into<FlattenItem>>(&mut self, flatten_item: I) {
+        debug_assert!(!self.current_group.is_empty());
+        self.current_group.push(flatten_item.into());
+    }
+
+    /// clears the current group, and adds a new group to the groups
+    pub fn close_group(&mut self) {
+        if !self.current_group.is_empty() {
+            let mut elements = vec![];
```
leops, line 327: This could be written as `let elements = self.current_group.clone()`, or alternatively:
```rust
let mut elements = vec![];
std::mem::swap(&mut elements, &mut self.current_group);
self.groups.push(elements);
```
(The difference is the first option retains the capacity of `current_group` while the second does not)

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,445 @@
...
+
+/// Formats together the first group and the rest of groups
+fn format_groups(first_group: &[FlattenItem], groups: Groups) -> FormatElement {
+    let first_formatted_group: Vec<FormatElement> = first_group
+        .iter()
+        .map(|flatten_item| flatten_item.clone().get_format_element())
+        .collect();
+
+    let formatted_groups = if groups.groups_should_break() {
+        block_indent(groups.get_formatted_joined_groups())
```
leops, line 376: It might be possible to fix the position of the trailing semicolon by replacing `block_indent` with `indent(format_elements![hard_line_break(), groups.get_formatted_joined_groups()])`

ematipico:
!bench_formatter

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    Ok(flatten_items.split_at(current_index + 1))
+}
+
+/// computes groups coming after the first group
+fn compute_groups(flatten_items: &[FlattenItem]) -> FormatResult<Groups> {
+    let mut has_seen_call_expression = false;
+    let mut groups = Groups::default();
+    for item in flatten_items {
```
MichaReiser, line 174: Can we avoid `cloning` the items by passing in the owned vector instead (or `impl Iterator<FlattenItem>`) because it seems that we aren't using the `FlattenItem`s thereafter. 

```
fn compute_groups(flatten_items: Vec<FlattenItem>) -> FormatResult<Groups> {
 ...
 for item in flatten_items.into_ter() {... }
}
```

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+        self.current_group.push(flatten_item.into());
+    }
+
+    /// clears the current group, and adds a new group to the groups
+    pub fn close_group(&mut self) {
+        if !self.current_group.is_empty() {
+            let mut elements = vec![];
+            std::mem::swap(&mut elements, &mut self.current_group);
+            self.groups.push(elements);
+            self.current_group.clear();
```
MichaReiser, line 301: Nit: Is it still necessary to call `clear` here after `self.current_group` has been swapped out with an empty vec?

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+        }
+    }
+
+    /// It tells if the groups should be break on multiple lines
+    pub fn groups_should_break(&self) -> bool {
+        // TODO: this should have more checks
```
MichaReiser, line 307: TODO?

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+        self.groups.len() > 3
+    }
+
+    fn get_formatted_groups(&self) -> Vec<FormatElement> {
+        self.groups
+            .iter()
+            .map(|group| {
+                concat_elements(group.iter().map(|flatten_item| flatten_item.clone().into()))
```
MichaReiser, line 315: Nit: We could avoid cloning here if the method consumes `self`. It doesn't seem as that the `Groups` object is used after calling `get_formatted_concat_groups` and `get_formatted_joined_groups.

Nit: I would rename `get_formatted_concat_groups` and `get_formatted_joined_groups` as `into_concated_groups()` and `into_joined_groups` to make it clear this consumes self.

crates/rome_formatter/src/lib.rs
```diff
@@ -479,12 +479,25 @@ pub fn format_file_and_save(rome_path: &mut RomePath, options: FormatOptions, ap
...
 
     use super::{format_range, FormatOptions};
-
+    use crate::{format, IndentStyle};
     use rome_rowan::{TextRange, TextSize};
-    use rslint_parser::parse_script;
+    use rslint_parser::{parse, parse_script, Syntax};
+
+    #[test]
+    fn poc() {
```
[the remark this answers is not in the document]
ematipico: It is only for doing quick tests. That's why it's called POC. I can try to find a different name

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+            }
+    }
+
+    /// It strictly checks if the current item is a [rslint_parser::ast::JsCallExpression]
+    pub fn is_call_expression(&self) -> bool {
+        matches!(&self, FlattenItem::CallExpression(_, _))
+    }
+}
+
+impl Display for FlattenItem {
```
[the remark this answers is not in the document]
ematipico: I will change it to Debug

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+        } else {
+            match item {
+                FlattenItem::CallExpression(_, _) | FlattenItem::ComputedExpression(_, _) => {
+                    current_index = index;
+                }
+
+                _ => break,
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ematipico: I would also need the index to split the `flatten_items` in the end. The job of this function is to get the first group out from the rest of the items, not the other way around.

crates/rome_formatter/src/utils/call_expression.rs
ematipico: We can't do it at the moment because the function `compute_first_group` uses, at the end, `split_at`, which returns a tuple of `(&[FlattenItem], &[FlattenItem])`

MichaReiser: `compute_first_group` could return a `Vec` to

```
fn compute_first_group(flatten_items: Vec<FlattenItem>) -> Vec<FlattenItem> {
  ...
  flatten_items.split_off(index)
}
```

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,413 @@
...
+                if let FlattenItem::StaticMember(_, _) = item {
+                    let next_flatten_item = &flatten_items[index + 1];
+                    if matches!(next_flatten_item, FlattenItem::StaticMember(_, _)) {
+                        current_index = index;
+                    }
+                } else {
+                    break;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ematipico: Yes, thank you for the second pair of eyes! 
I will apply the suggestion but I won't apply the `skip_while` because that's not what I am looking for.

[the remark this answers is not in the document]
MichaReiser: Would you then mind explaining the algorithm to me... It seems I don't understand it after all. 

You can get the index of the end by using 

```
match end_of_first_group.next() {
  Some((index, _)) => {
    // index points to the first element of the second group
    flatten_items.split_off(index) // returns a vec containing the elements of the second, third... groups
  },
  None => vec![] // All items belong to the first group
```

crates/rome_formatter/src/utils/call_expression.rs
ematipico: I applied `to_owned()` and changed the signature of the function.

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,411 @@
...
+        // - `something[1]()[3]()`
+        // - `something()[2].something.else[0]`
+        match item {
+            FlattenItem::CallExpression(_, _) | FlattenItem::ComputedExpression(_, _) => {
+                current_index = index;
+            }
+
+            FlattenItem::StaticMember(_, _) => {
+                if !flatten_items[0].is_call_expression() {
+                    let next_flatten_item = &flatten_items[index + 1];
```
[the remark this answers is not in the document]
ematipico: Yeah I can add some comment. Essentially, that's how you explained: the array will always have at least two elements:
Something like `a()` produces these flatten times:
```
[
    Token("a", 0..1),
    CallExpression: [Empty, Empty, Group(List [Token("(", 5..6), Token(")", 2..7)])],
]
```

Hence, it will never enter the branch of this `match`.

When we have something like `a.b.c()`, the flatten items produced are:

```
[
    Token("a", 0..1),
    StaticMember: [Token(".", 1..2), Token("b", 2..3)],
    StaticMember: [Token(".", 3..4), Token("c", 4..5)],
    CallExpression: [Empty, Empty, Group(List [Token("(", 5..6), Token(")", 6..7)])],
]
```

The loop will match against `StaticMember: [Token(".", 3..4), Token("c", 4..5)],` and the next one is a call expression... the `matches` fails and the loop is stopped.

crates/rome_formatter/src/utils/call_expression.rs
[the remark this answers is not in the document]
MichaReiser: The first element returned by this iterator is the index pointing to the start of the remaining groups.

crates/rome_formatter/src/utils/call_expression.rs
MichaReiser: Can we add an issue or at least in more detail what needs to be done. I wouldn't know what needs to be added if I stumble across this todo in the future.

crates/rome_formatter/src/utils/call_expression.rs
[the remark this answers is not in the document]
MichaReiser: Please add at some documentation since it's hard to understand the implementation.

crates/rome_formatter/src/utils/call_expression.rs
MichaReiser: That works too but requires slightly more memory than using `split_off` because the remaining groups are now stored twice (I wonder if Rust can even avoid a new heap allocation because all elements are already in memory and the length is stored on the stack).

leops: It would be theoretically possible since Rust's `Allocator` trait passes the layout of the allocation back to the deallocate method, but in practice I suspect most implementations discard it and use `free(ptr)` which then relies on state internal to the system allocators to find the size of the block to deallocate so passing in a pointer from the middle of a previously allocated block is not allowed.

As for avoiding allocations here, the solution that came to mind would be to have `compute_first_group` return a usize (the index of the split), then in `format_call_expression` do something like:

```rust
let split_index = compute_first_group(&flattened_expression)?;
let mut flattened_expression = flattened_expression.into_iter();
let first_group = concat_elements((&mut flattened_expression).take(split_index).map(FlattenItem::into));
let rest_of_groups = compute_groups(flattened_expression)?;
```

With `compute_groups` now taking an `Iterator<Type = FlattenItem>` as a parameter instead of a vector

crates/rome_formatter/src/utils/call_expression.rs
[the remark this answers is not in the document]
ematipico: I can apply your suggestion and it works as well. Although it has the subtle detail that using `split_off` mutates `flatten_item` under the hoods, I will add a comment to make sure that people understand  that.

crates/rome_formatter/src/utils/call_expression.rs
ematipico: Applied the suggested changes

ematipico:
@MichaReiser  @leops  I applied both of your suggestions, please take a look again

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,440 @@
...
+    // `flattened_items` now contains only the nodes that should have a sequence of
+    // `[ StaticMemberExpression -> AnyNode + JsCallExpression ]`
+    let rest_of_groups = compute_groups(flattened_items)?;
+    Ok(format_groups(first_group, rest_of_groups))
+}
+
+/// Retrieves the index where we want to calculate the first group.
+/// The first group gathers inside it all those nodes that are not a sequence of something like:
+/// `[ StaticMemberExpression -> AnyNode + JsCallExpression ]`
+fn compute_first_group_index(flatten_items: &mut Vec<FlattenItem>) -> usize {
```
leops, line 144: `flatten_items` doesn't need to be mutable and could be a `&[FlattenItem]`

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -0,0 +1,440 @@
...
+    let mut new_group = flatten_items
+        .iter()
+        .enumerate()
+        // the first element will always be part of the first group, so we skip it
+        .skip(1)
+        .skip_while(|(index, item)| match item {
```
leops, line 150: This could be written as `.find_map(|(index, item)| { let skip = match item { ... }; if skip { None } else { Some(index) } }).unwrap_or(1)`

ematipico: I can rewrite as much as you guys want, I would like to understand the **why** and to what extent 😛 

Does it make more readable for you?

crates/rome_formatter/src/utils/call_expression.rs
ematipico: Strange that clippy didn't catch it. Thank you!

crates/rome_formatter/src/utils/call_expression.rs
leops: Well this doesn't make much of a difference functionally, I just find it makes it more explicit that this is a find operation, not an iterator being built to actually be looped upon

ematipico:
!bench_formatter

PR #2104, opened 2022-02-15
feat(rome_formatter): implement fill-printing for array expressions

leops:
## Summary

This PR adds a new `fill` FormatElement inspired by the element of the same name in Prettier: it behaves as a list separated by `soft_line_break_or_space` where the breaking behavior of the soft line isn't determined by the printing mode of the outer group, but instead by its horizontal position relative to the print width to prevent list items from overflowing:

As an example, `fill(["a,"; 5])` would print the following:
```
---------| print width
a, a, a,
a, a,
```

This new element is used to enable "compact printing" of array expressions: when all the elements in the array are literals of 10 characters or less (optionally nested in a binary expression with the `-`, `+`, `!` or `~` operator) and the expression contains no printable trivia (comments or empty lines), the content of the array is fill-printed instead of separating each element on a different line


crates/rome_formatter/src/js/lists/array_element_list.rs
```diff
@@ -41,3 +51,88 @@ impl ToFormatElement for JsArrayElementList {
         Ok(join_elements_soft_line(results))
     }
 }
+
+/// Returns true if this node has comments or empty lines (2 consecutive
+/// newlines only separated by whitespace)
+fn has_trivia(node: &SyntaxNode) -> bool {
```
ematipico, line 57: Should this method part of `rome_rowan` APIs?

crates/rome_formatter/src/js/lists/array_element_list.rs
```diff
@@ -41,3 +51,88 @@ impl ToFormatElement for JsArrayElementList {
...
+}
+
+/// Returns true if the provided JsArrayElementList could
+/// be "fill-printed" instead of breaking each element on
+/// a different line.
+///
+/// The underlying logic only allows lists of literal expressions
+/// with 10 or less characters, potentially wrapped in a "short"
+/// unary expression (+, -, ~ or !)
+fn can_print_fill(list: &JsArrayElementList) -> bool {
```
ematipico, line 98: Should we move this inside `utils.rs`?

crates/rome_formatter/src/js/lists/array_element_list.rs
```diff
@@ -41,3 +51,88 @@ impl ToFormatElement for JsArrayElementList {
...
+                false
+            }
+        }
+        Ok(JsAnyExpression(expr)) => is_short_literal(&expr),
+        _ => false,
+    })
+}
+
+/// Returns true if the provided expression is a literal with 10 or less characters
+fn is_short_literal(expr: &JsAnyExpression) -> bool {
```
ematipico, line 122: Should we move it inside `utils.rs`? This utility might be helpful in other cases

crates/rome_formatter/src/printer.rs
```diff
@@ -284,47 +294,114 @@ impl<'a> Printer<'a> {
...
             | FormatElement::Indent { .. }
-            | FormatElement::List { .. } => self.print_element(element, args),
-        };
+            | FormatElement::List { .. } => self.print_element(queue, element, args),
+        }
+
+        Ok(())
+    }
+
+    /// Print a list in fill mode
```
ematipico, line 350:
```suggestion
    /// Print a list in fill mode.
```

crates/rome_formatter/src/printer.rs
```diff
@@ -284,47 +294,114 @@ impl<'a> Printer<'a> {
...
+                    self.print_all(queue, item, args.clone());
+                }
+            }
+
+            snapshot = Some(self.state.snapshot());
+        }
+    }
+
+    /// Fully print an element (print the element itself and all its descendants)
+    fn print_all(
```
ematipico, line 386: I don't quite understand this API.  What does "fully print an element" mean? Even with what I read inside the parenthesis, I still don't understand it

crates/rome_formatter/src/printer.rs
```diff
@@ -165,52 +173,50 @@ impl<'a> Printer<'a> {
...
             FormatElement::Group(Group { content }) => {
-                match self.try_print_flat(element, args.clone()) {
-                    Err(_) => {
-                        // Flat printing didn't work, print with line breaks
-                        vec![PrintElementCall::new(content.as_ref(), args)]
-                    }
-                    Ok(_) => vec![],
+                if self.try_print_flat(queue, element, args.clone()).is_err() {
+                    // Flat printing didn't work, print with line breaks
+                    queue.enqueue(PrintElementCall::new(content.as_ref(), args));
```
MichaReiser, line 181: I like that change. It actually is easier to understand what's going on.

crates/rome_formatter/src/js/lists/array_element_list.rs
leops: Maybe not into rowan since concepts such as "empty line" as 2 or more consecutive newlines are very specific to the formatter, but it could move it into `utils` along with the other utility functions in this file

MichaReiser: I actually like that these utilities are inside the `array_element_list.rs` file because they are very specific. The functions otherwise will need more specific names because the `has_trivia` has a very specific meaning

crates/rome_formatter/src/printer.rs
leops: The "issue" this method is meant to solve is that a single call `print_element` does not necessarily guarantees the entire content of the element has actually been printed: for instance calling `print_element` on a `List` will only push the content of said list to the queue but not print anything (and `print_element` has to be called in a loop until the queue is empty).
The `print_all` function on the other hand guarantees that when it returns, the entire content of the element has actually been printed.
This makes implementing the fill printing algorithm easier since it works by printing an element then if it went over the print width restore the print state, insert a line break and try again.

ematipico: Thank you for the explanation! It's now clear

crates/rome_formatter/src/js/lists/array_element_list.rs
leops: I ended up moving `has_trivia` to utils renamed to `has_formatter_trivia` (I also considered `has_printable_trivia` but wasn't sure that was very clear), but I left the other 2 helpers in place since they're specific to `JsArrayElementList`

crates/rome_formatter/src/format_element.rs
```diff
@@ -254,6 +254,31 @@ where
...
+/// use rome_formatter::{fill, FormatElement, space_token, token, format_element, FormatOptions};
+/// let a = from_utf8(&[b'a'; 30]).unwrap();
+/// let b = from_utf8(&[b'b'; 30]).unwrap();
+/// let c = from_utf8(&[b'c'; 30]).unwrap();
+/// let d = from_utf8(&[b'd'; 30]).unwrap();
+/// let expr = fill([token(a), token(b), token(c), token(d)]);
+///
+/// assert_eq!(format!("{a} {b}\n{c} {d}"), format_element(&expr, FormatOptions::default()).into_code())
+/// ```
+pub fn fill(elements: impl IntoIterator<Item = FormatElement>) -> FormatElement {
```
ematipico, line 273: nit: should we rename the method like `fill_elements` or `fill_with_elements`? It's inline with `group_elements` and `join_elements`

leops: Good point, I'll do that

PR #2107, opened 2022-02-15
fix(rslint_parser): error on static abstract members

xunilrj:
Errors:

```
error[SyntaxError]: static members cannot be abstract
  ┌─ typescript_abstract_classes_invalid_static_abstract_member.ts:1:20
  │
1 │ abstract class A { abstract static fn1(); }
  │                    ^^^^^^^^

--
error[SyntaxError]: static members cannot be abstract
  ┌─ typescript_abstract_classes_invalid_static_abstract_member.ts:2:27
  │
2 │ abstract class B { static abstract fn1(); }
  │                           ^^^^^^^^
```

### DIFF
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -1023,6 +1023,7 @@ fn parse_method_class_member_rest(
         .ok();
 
     let abstract_range = modifiers.get_range(ModifierKind::Abstract).cloned();
+    let static_range = modifiers.get_range(ModifierKind::Static).cloned();
     let is_async = flags.contains(SignatureFlags::ASYNC).then(|| true);
 
     parse_method_body(p, modifiers, flags);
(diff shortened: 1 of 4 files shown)


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1031,8 +1032,17 @@ fn parse_method_class_member_rest(
...
+    if let Some((abstract_range, _)) = abstract_range.as_ref().zip(is_async) {
+        let err = ts_parse_error::abstract_member_cannot_be_async(p, abstract_range.clone());
+        p.error(err);
+        member.change_to_unknown(p);
+    }
+
+    // test_err ts typescript_abstract_classes_invalid_static_abstract_member
+    // abstract class A { abstract static fn1(); }
+    // abstract class B { static abstract fn1(); }
+    if let Some((abstract_range, _)) = abstract_range.zip(static_range) {
```
MichaReiser, line 1044: Nit: Adding two errors for `abstract static async` may feel overwhelming to the user. We should first warn about one error and then the second.

MichaReiser (approved):

crates/rslint_parser/test_data/inline/err/typescript_abstract_classes_invalid_static_abstract_member.rast
```diff
@@ -0,0 +1,143 @@
...
+  │
+1 │ abstract class A { abstract static fn1(); }
+  │                    -------- ^^^^^^
+
+--
+error[SyntaxError]: static members cannot be abstract
+  ┌─ typescript_abstract_classes_invalid_static_abstract_member.ts:1:20
+  │
+1 │ abstract class A { abstract static fn1(); }
+  │                    ^^^^^^^^
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rslint_parser/src/syntax/class.rs
xunilrj: hum... not sure on this one. I actually prefer to see all errors at once than compiling twice.

PR #2108, opened 2022-02-15
refactor(rome_formatter): format_delimited helpers

leops:
## Summary

This PR adjusts the grouping semantics of the `format_delimited` function, as well as introduce new helper methods to make using it less verbose.

### Group elements

The first change is mostly significant for the formatter output and changes how the `FormatElements` returned by `format_delimited` are placed into groups. Previously most call sites of this method were invoked as `group_elements(format_delimited(open_token, |trailing, leading| *_indent([trailing, content, leading]), close_token))`, which yields the following IR:

```
Group [
    open_token_leading_trivia,
    open_token,
    Indent [
        open_token_trailing_trivia,
        content,
        close_token_leading_trivia,
    ],
    close_token,
    close_token_trailing_trivia,
]
```

With this change the method is now used as just `format_delimited_*_indent(open_token, content, close_token)` and results in the following IR:

```
open_token_leading_trivia,
Group [
    open_token,
    Indent [
        open_token_trailing_trivia,
        content,
        close_token_leading_trivia
    ],
    close_token,
],
close_token_trailing_trivia
```

The leading and trailing trivias are now pushed out of the group, which is important as these can be empty lines or single-line comments that would force the group to break when its actual content could have fit on a single line:

```
// This comment is a leading trivia on the [ token
[1, 2, 3]
```

Previously the comment would have been included in the group and the array expression would have thus been unnecessarily broken on multiple lines.

### Shorthand helpers

The different call sites of `format_delimited` vary slightly in how they handle their inner indentation but can be broadly categorized into 3 cases:
- `block_indent([trailing, content, leading])`
- `soft_block_indent([trailing, content, leading])`
- `[space, soft_block_indent([trailing, content, leading]), space]`

These 3 case are now implemented in `format_delimited_block_indent`, `format_delimited_soft_block_indent` and `format_delimited_soft_block_spaces` respectively, and the `format_delimited` method has been made private.

### DIFF
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -120,7 +120,7 @@ impl Formatter {
     }
 
     /// Formats a group delimited by an opening and closing token, placing the
-    /// content in a block_indent group
+    /// content in a [block_indent] group
     pub(crate) fn format_delimited_block_indent(
         &self,
         open_token: &SyntaxToken,
@@ -137,7 +137,7 @@ impl Formatter {
     }
 
     /// Formats a group delimited by an opening and closing token, placing the
-    /// content in a soft_block_indent group
+    /// content in a [soft_block_indent] group
     pub(crate) fn format_delimited_soft_block_indent(
         &self,
         open_token: &SyntaxToken,
@@ -154,7 +154,7 @@ impl Formatter {
     }
 
     /// Formats a group delimited by an opening and closing token, placing the
-    /// content in an indent group with soft_line_break_or_space tokens at the
+    /// content in an [indent] group with [soft_line_break_or_space] tokens at the
     /// start and end
     pub(crate) fn format_delimited_soft_block_spaces(
         &self,
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -5,11 +5,12 @@ use std::collections::HashSet;
 
 use crate::formatter_traits::FormatTokenAndNode;
 use crate::{
-    concat_elements, empty_element, empty_line,
+    block_indent, concat_elements, empty_element, empty_line,
     format_element::{normalize_newlines, Token, LINE_TERMINATORS},
-    format_elements, hard_line_break, if_group_breaks, if_group_fits_on_single_line,
-    join_elements_hard_line, line_suffix, soft_line_break_or_space, space_token, FormatElement,
-    FormatOptions, FormatResult, ToFormatElement,
+    format_elements, group_elements, hard_line_break, if_group_breaks,
+    if_group_fits_on_single_line, indent, join_elements_hard_line, line_suffix, soft_block_indent,
+    soft_line_break_or_space, space_token, FormatElement, FormatOptions, FormatResult,
+    ToFormatElement,
 };
 use rome_rowan::SyntaxElement;
 #[cfg(debug_assertions)]
@@ -79,10 +80,10 @@ impl Formatter {
     ///
     /// Calling this method is required to correctly handle the comments attached
     /// to the opening and closing tokens and insert them inside the group block
-    pub(crate) fn format_delimited(
+    fn format_delimited(
         &self,
         open_token: &SyntaxToken,
-        content: impl FnOnce(FormatElement, FormatElement) -> FormatResult<FormatElement>,
+        content: impl FnOnce(FormatElement, FormatElement) -> FormatElement,
         close_token: &SyntaxToken,
     ) -> FormatResult<FormatElement> {
         cfg_if::cfg_if! {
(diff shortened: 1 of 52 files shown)


crates/rome_formatter/tests/specs/prettier/comments/function-declaration.js.snap
```diff
@@ -82,24 +82,24 @@ KEYPAD_NUMBERS.map(num => ( // Buttons 0-9
 
 function f1() {} /* f */
 function f2( /* args */ ) {}
-function f3() /* returns */ {}
-function f4( /* args */ ) /* returns */ {} /* f */
+function f3() {} /* returns */
```
ematipico, line 85: Is this intentional?

```js
// input
function f3 () /* returns */ {}
```
In the input, the comment was in a different position. Now, it's been moved after the `}`. If this function had a big body, the comment would be "lost" because moved at the end of the function

crates/rome_formatter/tests/specs/prettier/comments/if.js.snap
```diff
@@ -87,54 +87,34 @@ else if (4) {
...
-  10
-) { /* comment */ // comment
+if (10) { /* comment */ // comment
   true;
-} else if (11) /* comment */ true; else if (
-  12
-) true; else if ( // comment /* comment */ // comment
-  13
-) true; else { /* comment */ /* comment */ // comment /* comment */
+} else if (11) /* comment */ true; else if (12) true; else if (13) true; else { // comment /* comment */ // comment /* comment */ /* comment */ // comment /* comment */
```
ematipico, line 109: Here we collapse comments that belonged to other regions of the code. Is it intentional?

crates/rome_formatter/tests/specs/prettier/method-chain/comment.js.snap
```diff
@@ -101,11 +97,7 @@ const configModel = this.baseConfigurationService.getCache().consolidated.merge(
   this.cachedWorkspaceConfig,
 );
 
-this.doWriteConfiguration(
-  target,
-  value,
-  options,
-).then( // queue up writes to prevent race conditions
+this.doWriteConfiguration(target, value, options).then( // queue up writes to prevent race conditions
```
ematipico, line 100: I preferred the old version. The comment was still in the correct position.

ematipico (commented):
It seems that some of the new snapshots now print comments in the wrong position

crates/rome_formatter/tests/specs/prettier/comments/function-declaration.js.snap
leops: I agree that these don't look right, but this is actually the comment formatting working as intended: block comments are printed "inline" when their parent group doesn't break, but as line suffixes when the outer group is breaking. In this case the comments used to be part of the `( /* args */ )` group but are now correctly outside of it, and end up not being part of any group which means they are printed as part of the implicit "file group": that group is always breaking so the comments are printed as line suffixes.

leops: This isn't really a new functionality, the way comments are printed hasn't changed it's just that they used to be included in groups they didn't actually belong to semantically. For instance, with the old behavior a function would have been printed like like this:
```js
function name() /* comment */ {}
```
But if enough parameters were added to cause a break, it would have printed like this instead:
```js
function name(
	many,
	parameters,
	to,
	make,
	the,
	group,
	break_,
	across,
	lines,
) {} /* comment */
```
Because the parameter group is broken and the block comment is attached to it, it now gets printed as a line suffix despite being semantically unrelated to the parameters

With the new behavior the printing of block comments is consistent whether the parameters group is breaking or not, the actual issue is that this new behavior is also consistent*ly* wrong in terms of formatting ...

ematipico: Sorry, I didn't explain myself as I wanted. What I meant is that this PR changes existing things, as explained in the description, although there are some changes in the snapshots - not only this one, I mean all of them -  that **don't seem correct**, although I don't know if these changes are intended or they are bugs.

That's why I wanted to see new test cases (that should be correct), so I could have a metric to understand when something is a bug or not.

crates/rome_formatter/src/formatter.rs
```diff
@@ -108,13 +109,79 @@ impl Formatter {
...
+            move |trailing_trivia, leading_trivia| {
+                soft_block_indent(format_elements![trailing_trivia, content, leading_trivia])
+            },
+            close_token,
+        )
+    }
+
+    /// Formats a group delimited by an opening and closing token, placing the
+    /// content in an indent group with soft_line_break_or_space tokens at the
+    /// start and end
```
ematipico, line 157:
```suggestion
    /// Formats a group delimited by an opening and closing token, placing the
    /// content in an indent group with [soft_line_break_or_space] tokens at the
    /// start and end
```

crates/rome_formatter/tests/specs/prettier/comments/function-declaration.js.snap
leops: It is both really, these changes are intended and they are bugs. Much like adding TypeScript parsing seemingly regresses the tests, here making the behavior of the formatter more correct exacerbates an existing issue with how comments are printed. So while I can add more tests, any case that's using block comments will not actually be "correct" because of the line suffix problem.

ematipico (approved):

PR #2110, opened 2022-02-15
feat(rslint_parser): diagnostic for class member body presence/absence

xunilrj:
When methods/constructors are overloaded they can be bodiless. Only getter/setter must always have bodies.
I have not created tests for bodiless overloaded methods, because we will only be able to verify them when doing semantic validation, a step that we don't do today.

```ts
class A {
    constructor(a:number);           // ts is happy
    constructor(a:number|string) {
    }

    fn1(a:number);                   // ts is happy
    fn1(a:number|string) {
    }

    #fn1(a:number);                  // ts is happy
    #fn1(a:number|string) {
    }

    get name ();                     // ts is sad
    get name () { return ""; }

    set name1 (a:number);            // ts is sad
    set name1 (a:number|string) {  }
}

```

# Regressions

Some regressions were actually testing this: 

compiler/ClassDeclaration10.ts
```ts
class C {
   constructor();
   foo();
}
```


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1043,29 +1033,26 @@ fn parse_method_class_member_rest(
...
+    }
+    // test_err getter_class_no_body
+    // class Getters {
+    //   get foo()
+
+    // test_err setter_class_no_body
+    // class Setters {
+    //   set foo(a)
+    else if flags.contains(SignatureFlags::GETTER) || flags.contains(SignatureFlags::SETTER) {
         body.or_add_diagnostic(p, js_parse_error::expected_class_method_body);
```
xunilrj, line 1055: @MichaReiser what do you think renaming this to ```add_diagnostic_if_absent```?

MichaReiser: The `or` tried to give a similar feeling to `Result`/`Option` but it's somewhat failing at that because it also returns the `Option`. 

I'm OK with `add_diagnostic_if_absent` if we change the method to `void` and make it so that it doesn't consume `self`.

MichaReiser:
> Only getter/setter must always have bodies.

This is not true in ambient contexts: 

```ts
declare class Test {
    get test(): string;
}
```

crates/rslint_parser/src/state.rs
```diff
@@ -312,6 +312,10 @@ bitflags! {
         const GENERATOR 	= 1 << 1;
         /// Is the function a constructor (or constructor context)
         const CONSTRUCTOR 	= 1 << 2;
+        /// Is the function a getter
+        const GETTER 	= 1 << 3;
+        /// Is the function a setter
+        const SETTER 	= 1 << 4;
```
MichaReiser, line 318: I try to avoid these global flags whenever possible because it makes it harder to reason about the code. For example, `parse_function` isn't handling the `GETTER` and `SETTER` flags. Is this a bug or intentional? Is it even possible that the flags are ever set? 

My understanding is that the two flags are only used when parsing an accessor body. I would, therefore, recommend extracting a `parse_accessor_body` that performs the additional checks and internally delegates to `parse_method_body`.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1043,29 +1033,26 @@ fn parse_method_class_member_rest(
...
+    //     abstract display(): void { }
+    //     abstract get my_name() { }
+    //     abstract set my_name(name) { }
     //     abstract #private_name() { }
     // }
-    let is_abstract = modifiers.has(ModifierKind::Abstract);
-    if !is_abstract {
+    if modifiers.has(ModifierKind::Abstract) {
+        body.add_diagnostic_if_present(p, unexpected_abstract_member_with_body);
+    }
```
MichaReiser, line 1046: I can't see any changes to the grammar, but the grammar requires that the body is required for method members.  A valid "output" thus creates now an AST that is invalid according to our grammar.

MichaReiser:
```
abstract class B implements A {
    abstract prop: string;
    abstract raw: string;
    abstract readonly ro: string;
    abstract get readonlyProp(): string;
    abstract set readonlyProp(val: string);
    abstract m(): void;
}
```

this syntax is also valid. See `abstractPropertyBasics`

xunilrj:
> > Only getter/setter must always have bodies.
> 
> This isn't the case in ambient contexts:
> 
> ```ts
> declare class Test {
>     get test(): string;
> }
> ```

I think I dealt with this now.

crates/rslint_parser/src/syntax/class.rs
xunilrj: I am dealing with AST at https://github.com/rome/tools/pull/2121

crates/rslint_parser/src/state.rs
xunilrj: Done. May you will not like that ```parse_method_body``` returns the ```ParsedSyntax``` now... which led me to some ```let _ =``` that you dislike.

crates/rslint_parser/src/syntax/class.rs
xunilrj: I will do this in another PR.

crates/rslint_parser/src/state.rs
MichaReiser: eh.. I'm not that opinionated. `let _` is fine with me :)

MichaReiser:
> > ```
> > abstract class B implements A {
> >     abstract prop: string;
> >     abstract raw: string;
> >     abstract readonly ro: string;
> >     abstract get readonlyProp(): string;
> >     abstract set readonlyProp(val: string);
> >     abstract m(): void;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this syntax is also valid. See `abstractPropertyBasics`
> 
> Accepting abstract properties now.

Eh... I wasn't aware that the properties can be abstract too. I only noticed that the getter/setters can be abstract.

PR #2111, opened 2022-02-15
feat(rslint_parser): TypeScript `import` extensions

MichaReiser:
## Summary

Adds support for TypeScript's `import` extensions:

```ts 
import { type A, type b as C } from "./mod";
import type foo from "./mod";
import type * as foo2 from "./mod";
import type { foo3 } from "mod";
```

## Test Plan

Added new parser tests

### DIFF
--- a/crates/rslint_parser/src/syntax/js_parse_error.rs
+++ b/crates/rslint_parser/src/syntax/js_parse_error.rs
@@ -141,18 +141,6 @@ pub(crate) fn expected_named_import_specifier(p: &Parser, range: Range<usize>) -
     expected_node("identifier", range).to_diagnostic(p)
 }
 
-pub(crate) fn expected_local_name_for_default_import(
-    p: &Parser,
-    range: Range<usize>,
-) -> Diagnostic {
-    p.err_builder("`default` imports must be aliased")
-        .primary(p.cur_tok().range(), "`default` used here")
-        .secondary(
-            range.end..range.end,
-            "add the `as` keyword followed by an identifier name here",
-        )
-}
-
 pub(crate) fn duplicate_assertion_keys_error(
     p: &Parser,
     key: &str,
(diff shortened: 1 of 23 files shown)


MichaReiser:
!bench_parser

crates/rslint_parser/src/syntax/js_parse_error.rs
```diff
@@ -141,18 +141,6 @@ pub(crate) fn expected_named_import_specifier(p: &Parser, range: Range<usize>) -
     expected_node("identifier", range).to_diagnostic(p)
 }
 
-pub(crate) fn expected_local_name_for_default_import(
```
ematipico, line 144: Why did we remove this error?

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -311,51 +340,67 @@ impl ParseSeparatedList for NamedImportSpecifierList {
...
-            .precede(p)
-            .complete(p, JS_SHORTHAND_NAMED_IMPORT_SPECIFIER)
-    })
+    let specifier = if metadata.has_alias || p.at(JS_STRING_LITERAL) || p.cur().is_keyword() {
+        if metadata.is_local_name_missing {
+            // test_err import_as_identifier_err
+            // import { as c } from "test";
+            p.error(expected_literal_export_name(
+                p,
+                p.cur_tok().start()..p.cur_tok().start(),
```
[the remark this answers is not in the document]
MichaReiser: The identifier was expected before the `as` and the user shouldn't have to replace the `as` but instead write an identifier before. Ideally, the cursor would point right before the `as` keyword but that's nothing our diagnostic system supports today.

crates/rslint_parser/src/syntax/js_parse_error.rs
MichaReiser: It's now handled in a generic way. Let's say you have `import { yield }` then this must be an error that the alias is missing.

MichaReiser:
> Are the regressions expected?

Yes. These are all type checking related errors that our parser doesn't handle yet.

xunilrj (approved):

ematipico (approved):

MichaReiser:
> > > Are the regressions expected?
> > 
> > 
> > Yes. These are all type checking related errors that our parser doesn't handle yet.
> 
> Is there a way to denylist them? They are really misleading

We don't have a way to do this right now. It becomes less misleading when we fixed all remaining parse errors which we track in #2113

PR #2112, opened 2022-02-16
chore(coverage): Adds `--detailed` option to `cargo coverage`

MichaReiser:
## Summary
The coverage reporter prints the progress with all ran tests to `stderr`. This means, that it isn't possible to pipe the results to a text file. 

This PR adds a `--detailed` option that allows printing the failing tests with their diagnostics after running all tests to `stdout` (except if `--json` is passed). This allows piping the test results to a text file for analysis. 

This PR further changes:

* `--json`: Forces that all other reporters print to `stderr` instead of `stdout`. It's now possible to use all options in combinations
* `--filter=<file>` Introduce the filter option that replaces the previous free argument 
* `--detailed=rast`: Replaces `--show-rast`. It prints the rast of all failing tests in the summary

I updated the documentation to reflect the changes.

I had to fix a bug in the diagnostics crate related to incorrectly indexing into the middle of characters. The issue is that the unicode line breaks are longer than 1 byte. Thus, they break
the `+ 1` arithmetic used throughout the diagnostics package to get the next character right before/after a line break.

I don't think we need to handle these line breaks as "proper" line breaks because many editors
fall back to render the unicode line breaks as a special character.

## Test Plan

[Sample Output](https://gist.github.com/MichaReiser/624c57cd9787a82e4a1fe35375930703)

Piping the output in powershell breaks the formatting of the diagnostics for some reason. I believe it's because it uses UTF16 by default and rust prints the characters using UTF8. However, using `cargo coverage  | Out-File -Encoding utf8NoBOM results.txt` isn't working as well...


ematipico (changes requested):
Can we update the `README.md` with a brief explanation of how the reporting works now? It seems there's new flag, instead we use `atty` under the hood, so there are some conditions when `TextReporter` pops up instead of `SummaryReporter`.

xtask/coverage/src/reporters.rs
```diff
@@ -375,19 +371,88 @@ pub(crate) struct DiagnosticsReporter;
...
+                        writeln!(self.0, "Failed to print diagnostic: {}", err).ok();
+                    }
+                }
+            }
+        }
+    }
+
+    fn run_completed(&mut self) {
+        if let Err(err) = std::io::stdout().write_all(self.0.as_slice()) {
+            println!("Failed to print the test results: {}", err)
```
leops, line 455: If `stdout().write()` returned an error there's a very good change println wont be able to print anything either and will panic. This seems like a very uncommon failure case if it can happen at all (maybe if the output is piped to a file and the disk is full ?) so there isn't much that can be done about it, but maybe using eprintln instead might make more sense

leops (commented):
I tried running this on my own instance of PowerShell (globally configured to output UTF-8) and I did get a valid UTF-8 file when piping the output (with a BOM though because PowerShell doesn't Unicode without BOM) so I suspect any encoding issue is coming from PowerShell and this code is working properly

ematipico (approved):

PR #2116, opened 2022-02-16
feature(xtask): git integration for formatter codegen

leops:
## Summary

This is a followup to #2031 and QoL improvement related to #1997:

- It removes the now-unused migration code from the old folder structure to the new
- The formatter codegen is now integrated with git and will automatically stage modified files, as well as avoid to overwrite files that have uncommitted modifications by default (can be overridden with the `--allow-staged` and `--allow-dirty` CLI flags)
- Rust files that are no longer imported into the auto-generated module tree are now automatically removed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,6 +173,9 @@ name = "cc"
 version = "1.0.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "22a9137b95ea06864e018375b72adfb7db6e6f68cfc8df5a04d00288050485ee"
+dependencies = [
+ "jobserver",
+]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ematipico (commented):
In Rome classic we had already something to deal with repos: https://github.com/rome/tools/blob/archived-js/internal/vcs/index.ts

And our CLI had a `--allow-dirty` when using `format` or `lint` or other commands. I think we can safely create a new crate called `rome_vsc` and have your new code. 

This will be useful for cases like: running [`rome auto-config`](https://rome.tools/#rome-autoconfig)

MichaReiser (approved):

ematipico (approved):

Issue #2004, opened 2022-01-25
📎 Improve state management and document sync in LSP server


yassere:
### Description

The LSP server handles `textDocument/didOpen` and `textDocument/didChange` notifications to synchronize the contents of file buffers with a hastily implemented `DocumentStore`. Not much thought was put into what exactly to store for each document, and it requires an overhaul.

The LSP server should also handle `textDocument/didClose` notifications to clear stored documents and related published diagnostics.

ematipico:
I might be able to help here. Let me know what's needs to be done.

PR #2117, opened 2022-02-16
refactor(lsp): improve error handling and state management

### LINKED ISSUE #2003: 📎 Clean up LSP formatting handler and support FormattingOptions
### Description

Currently, the LSP server supports `textDocument/formatting` requests but uses the default option `IndentStyle::Tab`

LSP clients can send a `FormattingOptions` which includes:
```
	/**
	 * Size of a tab in spaces.
	 */
	tabSize: uinteger;

	/**
	 * Prefer spaces over tabs.
	 */
	insertSpaces: boolean;
```
The LSP server should respect those options when formatting. Also, the `rome_lsp` formatting handler should be cleaned up.

yassere:
## Summary

This PR is a refactor of the LSP server with inspiration from [wasm-lsp-server](https://github.com/wasm-lsp/wasm-lsp-server).

- Much of the logic from `server.rs` was extracted to a `handlers` module.
- Document handling was cleaned up, and the LSP server now primarily uses the LSP `Url` as the key for files. The `AnalysisServer` still requires `FileId`s for uniquely identifying files.
- Error-handling was improved with the help of a `spawn_blocking_task` utility
- Code action state is no longer stored as part of the LSP server. Instead, the `AnalysisServer` is always queried for `codeAction` requests. In a future PR, a long-lived `AnalysisServer` will become a part of the `Session` (rather than created before every query).

Future work:
- Long-lived `AnalysisServer`
- Add incremental text document sync
- Send minimal edits when formatting rather than full replacement text.
- Add some tests

Closes #2003 #2004

### DIFF
--- a/crates/rome_analyze/src/analysis_server.rs
+++ b/crates/rome_analyze/src/analysis_server.rs
@@ -1,6 +1,5 @@
 use std::{collections::HashMap, sync::Arc};
 
-use anyhow::anyhow;
 use rslint_parser::{parse_script, AstNode, SyntaxNode, TextRange};
 use tracing::trace;
 
@@ -29,11 +28,8 @@ impl AnalysisServer {
         self.file_map.insert(file_id, text.into());
     }
 
-    pub fn get_file_text(&self, file_id: FileId) -> anyhow::Result<Arc<str>> {
-        self.file_map
-            .get(&file_id)
-            .cloned()
-            .ok_or_else(|| anyhow!("File text missing for FileId {}", file_id))
+    pub fn get_file_text(&self, file_id: FileId) -> Option<Arc<str>> {
+        self.file_map.get(&file_id).cloned()
     }
 
     pub fn parse(&self, file_id: FileId) -> SyntaxNode {
--- a/crates/rome_lsp/src/documents.rs
+++ b/crates/rome_lsp/src/documents.rs
@@ -6,7 +6,7 @@ use rome_analyze::FileId;
 /// Internal representation of supported [language identifiers]
 ///
 /// [language identifiers]: https://code.visualstudio.com/docs/languages/identifiers#_known-language-identifiers
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum Language {
     JavaScript,
     TypeScript,
--- a/crates/rome_lsp/src/handlers/analysis.rs
+++ b/crates/rome_lsp/src/handlers/analysis.rs
@@ -4,7 +4,7 @@ use rome_analyze::{AnalysisServer, FileId};
 use rslint_parser::TextRange;
 
 use crate::line_index::LineIndex;
-use crate::utils::{self, into_lsp_error};
+use crate::utils;
 
 /// Queries the [`AnalysisServer`] for diagnostics of the file matching [`FileId`]
 ///
@@ -15,7 +15,7 @@ pub(crate) fn diagnostics(
 ) -> jsonrpc::Result<Vec<lsp::Diagnostic>> {
     let text = analysis_server
         .get_file_text(file_id)
-        .map_err(into_lsp_error)?;
+        .ok_or_else(jsonrpc::Error::internal_error)?;
     let line_index = LineIndex::new(&text);
 
     let diagnostics: Vec<_> = analysis_server
@@ -36,7 +36,7 @@ pub(crate) fn code_actions(
 ) -> jsonrpc::Result<Vec<lsp::CodeActionOrCommand>> {
     let text = analysis_server
         .get_file_text(file_id)
-        .map_err(into_lsp_error)?;
+        .ok_or_else(jsonrpc::Error::internal_error)?;
     let line_index = LineIndex::new(&text);
 
     let code_actions: Vec<_> = analysis_server


crates/rome_analyze/src/analysis_server.rs
```diff
@@ -70,6 +75,26 @@ impl AnalysisServer {
...
+        file_id: FileId,
+        cursor_range: Option<TextRange>,
+    ) -> impl Iterator<Item = Action> {
+        let analyzer_actions = self.analyzer_actions(file_id);
+        let assist_actions = match cursor_range {
+            Some(range) => self.assists(file_id, range).into_actions(),
+            None => Analysis::default().into_actions(),
+        };
+        analyzer_actions.chain(assist_actions)
+    }
```
ematipico, line 97: Can we document the different between these two methods? They have the same returning signature

crates/rome_lsp/Cargo.toml
```diff
@@ -14,8 +14,11 @@ rome_formatter = { path = "../rome_formatter" }
 rome_analyze = { path = "../rome_analyze" }
 rslint_errors = { path = "../rslint_errors", version = "0.2.0" }
 rslint_parser = { path = "../rslint_parser" } 
-lspower = "1.5.0"
+lspower = { git = "https://github.com/silvanshade/lspower.git" }
```
ematipico, line 17: Is this intentional? Is there are reason?

crates/rome_lsp/src/session.rs
```diff
@@ -0,0 +1,125 @@
...
+use lspower::jsonrpc::Error as LspError;
+use lspower::lsp;
+use parking_lot::{Mutex, RwLock};
+use rome_analyze::AnalysisServer;
+
+use crate::{
+    capabilities::server_capabilities, documents::Document, handlers, url_interner::UrlInterner,
+};
+
+pub struct Session {
```
ematipico, line 12: Can we document this `struct` please? And possibly all the `pub` fields

crates/rome_lsp/src/session.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+pub fn into_lsp_error(err: impl Display) -> LspError {
+    let mut error = LspError::internal_error();
+    error.data = Some(err.to_string().into());
+    error
+}
+
+impl Session {
```
ematipico, line 56: It would be great if we could document all the `pub` methods

crates/rome_lsp/src/session.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    pub async fn update_diagnostics(&self, url: lsp::Url) -> anyhow::Result<()> {
+        let doc = self.document(&url)?;
+
+        let mut analysis_server = AnalysisServer::default();
+        analysis_server.set_file_text(doc.file_id, doc.text);
+
+        let handle = tokio::task::spawn_blocking(move || {
+            handlers::analysis::diagnostics(analysis_server, doc.file_id)
+        });
```
ematipico, line 112: What do we want to achieve using `spawn_blocking`? What does it do? Sorry, I don't know anything of `tokio`

crates/rome_lsp/src/session.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        let diagnostics = handle.await??;
```
ematipico, line 114: Why is there a double try operator?

crates/rome_lsp/src/server.rs
```diff
@@ -1,165 +1,43 @@
...
     }
 
-    async fn file_id(&self, url: Url) -> FileId {
-        self.state.lock().await.url_interner.intern(url)
-    }
-
-    async fn url(&self, file_id: FileId) -> Url {
-        self.state
-            .lock()
+    pub async fn task<F, R>(&self, f: F) -> LspResult<R>
```
ematipico, line 27: Can we document this one please? Also, internally it uses `tokio::task::spawn_blocking`. Should we call the method `task_blocking`?

yassere:
@ematipico Oops, I meant to open this as a draft PR. Yeah, I plan to document all that before marking this ready for review.

crates/rome_lsp/Cargo.toml
yassere: It's because the published version of `lspower` uses `lsp-types 0.91` and the published version of `lsp-text` uses `lsp-types 0.92` so the types were incompatible. This was just a quick local workaround, but I don't plan to merge it like this.

crates/rome_lsp/src/session.rs
yassere: The idea behind [spawn_blocking](https://docs.rs/tokio/0.2.22/tokio/task/fn.spawn_blocking.html) is that it spawns your task on a thread where computation-heavy or blocking tasks are acceptable. I'm not entirely sure that it's the right choice here, but I was planning to flag this as something to look at during review.

crates/rome_lsp/src/session.rs
yassere: Because there are two `Result` layers that need to be removed. One from the task and one from the handler itself. I might take another pass at making the task helper a bit more ergonomic, though.

crates/rome_analyze/src/analysis_server.rs
yassere: Yeah, I added some comments to explain the difference for now. I expect this file will see a lot of changes when I tackle the `rome_analyze` refactor next, so I'll leave additional analyzer-related documentation until then.

crates/rome_lsp/src/session.rs
yassere: @leops Do you think `spawn_blocking` is a good choice here and as part of the `spawn_blocking_task` utility?

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -23,12 +25,15 @@ impl AnalysisServer {
...
     }
 
-    pub fn set_file_text(&mut self, file_id: FileId, text: impl Into<Arc<String>>) {
+    pub fn set_file_text(&mut self, file_id: FileId, text: impl Into<Arc<str>>) {
         self.file_map.insert(file_id, text.into());
     }
 
-    pub fn get_file_text(&self, file_id: FileId) -> Option<Arc<String>> {
-        self.file_map.get(&file_id).cloned()
+    pub fn get_file_text(&self, file_id: FileId) -> anyhow::Result<Arc<str>> {
```
MichaReiser, line 32: Nit: What's the motivation behind using `Result` here. It doesn't seem to encode the fact that the file is present or absent which is exactly what `Option` does (also the `Error` doesn't provide much added detail)

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -55,6 +60,11 @@ impl AnalysisServer {
         tree.covering_element(range).ancestors().find_map(T::cast)
     }
 
+    /// Returns a combined [`Analysis`] from running every [`AssistProvider`] on
```
MichaReiser, line 63: Nit (not needed to be done now): It would be nice if the constraint that the file must have been previously set in the API. For example by providing a method that returns `Option<File>`  that then provides methods like `assists`, `diagnostics` etc. thus forcing the caller to check if the file exists. This is also something that `set_file_text` (or `register_file`) could return.

crates/rome_lsp/src/documents.rs
```diff
@@ -1,86 +1,52 @@
...
-    documents: HashMap<FileId, Arc<Document>>,
+use std::sync::Arc;
+
+use anyhow::bail;
+use rome_analyze::FileId;
+
+/// Internal representation of supported [language identifiers]
+///
+/// [language identifiers]: https://code.visualstudio.com/docs/languages/identifiers#_known-language-identifiers
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
MichaReiser, line 9: Nit: What's the use case for implementing `Ord` and `PartialOrd`. How's the ordering defined?

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -0,0 +1,160 @@
...
+    let indent_style = if params.insert_spaces {
+        IndentStyle::Space(params.tab_size as u8)
+    } else {
+        IndentStyle::Tab
+    };
+    FormatOptions {
+        indent_style,
+        ..FormatOptions::default()
+    }
+}
```
MichaReiser, line 20: Nit: How about implementing `From<FormattingOptions>` for `FormatOptions` or does that not work because both types are from other crates?

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -0,0 +1,160 @@
...
+
+pub fn format(text: &str, file_id: FileId, params: &FormattingOptions) -> Result<Vec<TextEdit>> {
+    let tree = parse_script(text, file_id).syntax();
```
MichaReiser, line 23: For another PR: We should decide on the file type whatever to call `parse_script`/`parse_module` or just `parse` (TS)

crates/rome_lsp/src/session.rs
leops: Yes in general `spawn_blocking` is the right function to use for these kind of things. To give a bit more context most async executors in Rust run their tasks on a fixed-size thread pool (generally with as many threads as there are CPU cores), so running any kind of long non-async code in the middle of an async function will block one of these threads for the duration of the task, during which it wont be able to service other requests. `spawn_blocking` simply spawns a dedicated thread to run the provided closure outside of the tokio thread pool.

Now for this specific case, not going through `spawn_blocking` and running the blocking code in the async function might not make much of a difference since theses tasks aren't that long-running, the overhead of spawning a new thread might be higher than the time it takes to run the task itself, and I think the LSP server will never be under enough load that it makes a difference for the user anyway.

crates/rome_lsp/src/handlers/formatting.rs
ematipico: I tried that before, and it's not possible doing it with crates external crates, that's why I created a function

crates/rome_analyze/src/analysis_server.rs
yassere: Most of the time, this would be called internally by some other query, and you'd generally want to treat it as an error if it needs the text for an interned `FileId` and it's missing. I think the equivalent function in `rust-analyzer` actually just panics because file text is a salsa input.

But I'll revert this for now. I'll reconsider it when I'm working on `rome_analyze` and update the LSP `analyze` handlers accordingly.

crates/rome_analyze/src/analysis_server.rs
yassere: Eventually, a query like `assists` or `diagnostics` would involve several files under the hood. And the idea is that it's the responsibility of the thing "driving" the `AnalysisServer` to make sure it has its input facts. Methods like `set_file_text` would be called by file watchers or the `textDocument` handlers.

I'll definitely put more thought into this whole API when I resume work on the `rome_analyze` crate next week.

crates/rome_lsp/src/documents.rs
yassere: You're right. That doesn't make sense.

crates/rome_lsp/src/session.rs
ematipico: So, if I understood correctly, if we use `spawn_blocking`, we **could** expect a computation-heavy/blocking task. As a consequence, I wonder if using this API "gives" the wrong message? Meaning, inside the LSP, for the tasks that we are executing, **we should not expect blocking or heavy tasks**, otherwise it means that we are doing something wrong. 

Using `spawn_blocking` means that we are giving for granted that parsing, formatting, etc. could take a lot, which should not be the case because we - mostly - execute tasks on single files. I would expect such behavior if we work on entire project. Should we use this API only for tasks that **we know/aware** of their nature (computation-heavy tasks)?

crates/rome_lsp/src/handlers/formatting.rs
ematipico: For the time being, I think using TS is the safest bet, because the parser is able to parse most of the syntax. Probably we could use `rome_path` and have a function called `detect_syntax` which returns the correct `Syntax` based on the extension of the file. 

When we will be able to detect the type of project (AKA parse and analyze `package.json` ), we might be able to have more information around it.

MichaReiser (approved):

crates/rome_lsp/src/session.rs
yassere: I'm reconsidering now if spawning threads should even be the responsibility of the LSP Server itself. Instead, it could just hold a handle to a `RomeServer` and then await queries, entirely unconcerned with the concurrency strategy that's used under the hood. I'll revisit this soon when I work on the `AnalysisServer` (and maybe that will eventually just become an all-purpose server to query/command for analysis, formatting, bundling, etc.)

The original Rome had a similar design with a nice bridge abstraction.

PR #2121, opened 2022-02-16
feat(rslint_parser): abstract members AST nodes

xunilrj:
see https://github.com/rome/tools/issues/2059

Ambient contexts declare, overloads and abstracts have in common that they are bodiless. This PR deals only with ```abstract```. Others will be dealt with in future PRs.

For this, I created the following nodes that can be re-used (with additions later)

- TsPropertySignatureClassMember
- TsMethodSignatureClassMember
- TsSetterSignatureClassMember
- TsSetterSignatureClassMember

These have the ```abstract``` token as mandatory, but we may need to make it optional later.

# Errors

- Only abstract classes can have abstract members
- Abstract properties cannot have definite assignment assertions.
- Abstract properties cannot have initializers
- Only abstract classes can have abstract members


MichaReiser:
I wouldn't name the nodes `Abstract` because class methods in an ambient context or overload method have no function body as well and I think we should avoid introducing extra nodes for all of these. 

* `DeclareFunctionDeclaration`: Function declaration without a body -> `DeclareMethodClassMember`? 
* Reuse `MethodTypeMember`. The main downside there is that `abstract` is only valid inside of classes and not for `interface`s and `type`s. I'm also not sure if it's a good idea if the members are shared across multiple nodes because it then becomes necessary to test for the parent to figure out the context of the node.

> abstract properties

Do they exist? 

> Do we want an abstract class node

I don't think so because classes in ambient context can't have method bodies too, but these aren't abstract.

xunilrj:
> > abstract properties
> 
> Do they exist?

Yeah. And people seem to love it.

> I would go with the existing class node because classes in ambient context can't have method bodies too, but these aren't abstract.

I decided to keep the abstract nodes here for the same reason I went with specific index signature nodes here (https://github.com/rome/tools/pull/2133#discussion_r810504864).

The rationale is: it is safer to roll back if we decide.

ematipico:
> I would go with the existing class node because classes in ambient context can't have method bodies too, but these aren't abstract.

I want to share my point of view, as person not much involved on the parsing but as direct consumer when comes to create the formatting logic.

Usually, when I want to write a forma test, I look at the grammar and the name of the nodes that I am implementing. This resulted, sometimes, in some misunderstandings due to names that were misleading, or grammar that doesn't reflect 100% the behaviour of the parser. (meaning that the parser emits syntax errors but the grammar is not that strict as the parser)

I know that there should be compromises, otherwise there would be 10000 nodes for each variation, but I am also thinking about factories (creating nodes from scratch), which will be generated by the grammar, and I think these decision will lead to some mistakes.

For example, in this case, we could generate a factory that will be able to create something like:

```ts
class A {
    abstract f: string;
}
```

Personally, I would create a new abstract class node.

Feel free to discard my comment if you think it's not the right decision, I just wanted to share my concerns.

xtask/codegen/js.ungram
```diff
@@ -619,15 +619,21 @@ JsAnyClassMemberName =
 	| JsPrivateClassMemberName
 
 JsAnyClassMember =
+	// Concrete Members
```
MichaReiser, line 622: What do you mean by concrete members? Are there any non-concrete members?

xtask/codegen/js.ungram
```diff
@@ -685,12 +691,19 @@ JsPropertyClassMember =
 	access_modifier: ('private' | 'protected' | 'public')?
 	'static'?
 	'readonly'?
-	'abstract'?
 	name: JsAnyClassMemberName
 	property_annotation: TsAnyPropertyAnnotation?
 	value: JsInitializerClause?
 	';'?
 
+TsAbstractPropertyClassMember =
```
MichaReiser, line 699: What's your plan for methods in an ambient context? 

```
declare class Test {
  method();
}
```

Do you want to introduce another 5 nodes or are you planning to reuse the `abstract` members?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -728,7 +782,15 @@ fn parse_class_member_impl(
                         parse_ts_type_annotation_or_error(p).ok();
 
                         parse_getter_or_setter_body(p, modifiers);
-                        member_marker.complete(p, JS_GETTER_CLASS_MEMBER)
+
+                        let kind = if is_member_abstract {
+                            p.eat(T![;]);
+                            TS_ABSTRACT_GETTER_CLASS_MEMBER
```
MichaReiser, line 788: Do we need to add an error here too if this isn't an abstract class? same for setters

MichaReiser:
> > I would go with the existing class node because classes in ambient context can't have method bodies too, but these aren't abstract.
> 
> I want to share my point of view, as person not much involved on the parsing but as direct consumer when comes to create the formatting logic.
> 
> Usually, when I want to write a format test, I look at the grammar and the name of the nodes that I am implementing. This resulted, sometimes, in some misunderstandings due to names that were misleading, or grammar that doesn't reflect 100% the behaviour of the parser. (meaning that the parser emits syntax errors but the grammar is not that strict as the parser)
> 
> I know that there should be compromises, otherwise there would be 10000 nodes for each variation, but I am also thinking about factories (creating nodes from scratch), which will be generated by the grammar, and I think these decision will lead to some mistakes.
> 
> For example, in this case, we could generate a factory that will be able to create something like:
> 
> ```ts
> class A {
>     abstract f: string;
> }
> ```
> 
> Personally, I would create a new abstract class node.
> 
> Feel free to discard my comment if you think it's not the right decision, I just wanted to share my concerns.

Thanks, @ematipico for sharing your perspective. I can understand that it is frustrating if the AST allows for syntax that in practice, isn't allowed. 

I would still go with the existing `class` node for two reasons: 

1) Introducing an `abstract` class would mean that we also need to introduce a `JsAbstractClassDefaultExportDeclaration` + all the members. 
2)  I would re-use the newly introduced nodes to model ambient members (and rename the members to `TsMethodSignatureClassMember`:

```
declare class Test { 
  method();
}
```

Not reusing the members would otherwise require adding another version for each member. This doesn't just add additional generated code but also results in a lot of duplication in the formatter. You can already see today, that the logic for formatting members has been duplicated. Yes, we could mitigate this by introducing unions for each member and implementing formatting on those:

`JsAnyGetterClassMember = TsMethodSignatureClassMember | JsGetterClassMember` 

But even doing so results in additional complexity


I also remember that Seb once mentioned to me that he designed the rome classic AST in a way so that it's easier to move nodes around without requiring a huge amount of mutation code. Encoding all constraints stands in direct conflict with this goal

xtask/codegen/js.ungram
xunilrj: hum... I had "// abstract members" but I decided to remove it. I will also remove this.

xtask/codegen/js.ungram
MichaReiser: Can you explain your reasoning behind using another set of nodes? 

I'm mainly worried about the duplication that it leads to in formatter (all downstream code) and don't see how we can represent all valid syntax:

```
abstract class Test {
  declare abstract method();
}
```

Is `method` an `abstract` or an ambient member?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -170,7 +170,7 @@ impl From<ClassKind> for JsSyntaxKind {
 }
 
 fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
-    eat_contextual_keyword(p, "abstract", T![abstract]);
+    let is_abstract = eat_contextual_keyword(p, "abstract", T![abstract]);
```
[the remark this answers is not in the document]
xunilrj: Yes. It is working.

crates/rslint_parser/src/syntax/class.rs
xunilrj: I do it outside of this method. See https://github.com/rome/tools/pull/2121/files?file-filters%5B%5D=.rast&show-viewed-files=true#diff-798a98bd05af70fb6465283b6e67aca275885f5759a5638d8cc75f49947cd96b

xtask/codegen/js.ungram
```diff
@@ -688,12 +692,19 @@ JsPropertyClassMember =
...
 	'readonly'?
-	'abstract'?
 	name: JsAnyClassMemberName
 	property_annotation: TsAnyPropertyAnnotation?
 	value: JsInitializerClause?
 	';'?
 
+TsPropertySignatureClassMember =
+	access_modifier: ('protected' | 'public')?
+	'abstract'
```
MichaReiser, line 702: I don't think we need this node since `properties` have no-body that must be present for non-abstract properties. 

I could see reason for introducing this node if the `property_annotation` wouldn't allow for an initializer.
If so, I would change it to 

```
...
name: JsAnyClassMemberName
'?'?
type_annotation: TsTypeAnnotation
';'?
```

xtask/codegen/js.ungram
```diff
@@ -722,6 +732,15 @@ JsMethodClassMember =
 	return_type_annotation: TsReturnTypeAnnotation?
 	body: JsFunctionBody
 
+TsMethodSignatureClassMember =
+	access_modifier: ('protected' | 'public')?
+	'abstract'
```
MichaReiser, line 737: Do you plan to support `declare` in another PR or should I take this on when looking into TS modifiers?

What about methods in ambient contexts? Is this supported. If so, then `abstract` must be optinoal.

xunilrj:
> 2. I would re-use the newly introduced nodes to model ambient members (and rename the members to `TsMethodSignatureClassMember`:

I am going with this solution.

crates/rome_formatter/tests/specs/prettier/typescript/conformance/classes/classDeclarations/classAbstractKeyword/classAbstractMixedWithModifiers.ts.snap
```diff
@@ -26,17 +26,30 @@ abstract class A {
 # Output
 ```js
 abstract class A {
+  
     abstract foo_a();
+
+
```
MichaReiser, line 33: Is this file correctly formatted. There's a lot of whitespace in between.

crates/rome_formatter/tests/specs/prettier/typescript/conformance/classes/classDeclarations/classAbstractKeyword/classAbstractOverloads.ts.snap
```diff
@@ -52,10 +52,14 @@ abstract class A {
 }
 
 abstract class B {
+  
     abstract foo() : number;
+  
     abstract foo();
-    x : number;
+  x: number;
```
MichaReiser, line 59: This member doesn't seem to be formatted correctly.

crates/rome_formatter/tests/specs/prettier/typescript/conformance/classes/classDeclarations/classAbstractKeyword/classAbstractOverrideWithAbstract.ts.snap
```diff
@@ -38,17 +38,20 @@ class A {
...
+  foo() {}
+  
     abstract bar();
 }
 
 abstract class BB extends AA {
+  
     abstract foo();
-    bar () {}
+  bar() {}
```
MichaReiser, line 54: Same here. Seems the parser gets confused

crates/rome_formatter/tests/specs/ts/declaration/class.ts.snap
```diff
@@ -109,9 +109,21 @@ abstract class ConcreteMembers {
...
 }
 
+
+## Unimplemented nodes/tokens
+
+"\n    abstract name: string;" => 781..808
+"\n    abstract display();" => 808..832
+"\n    abstract get my_name();" => 832..860
+"\n    abstract set my_name(val);" => 860..891
+
```
MichaReiser, line 129: Whoa, this is cool. @ematipico is this your work?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -450,6 +462,32 @@ fn parse_class_member(p: &mut Parser) -> ParsedSyntax {
     }
 }
 
+fn add_error_if_member_is_abstract(
+    p: &mut Parser,
+    member: ParsedSyntax,
+    member_abstract_token_range: Option<Range<usize>>,
```
MichaReiser, line 473: Nit: You could avoid cloning if this accepts `Option<&Range<usize>>` and that you only clone if there's an error.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -898,7 +978,23 @@ fn parse_property_class_member_body(p: &mut Parser, member_marker: Marker) -> Pa
         p.error(err);
     }
 
-    Present(member_marker.complete(p, JS_PROPERTY_CLASS_MEMBER))
+    let member = Present(member_marker.complete(p, kind));
+
+    // test_err ts ts_abstract_property_cannot_have_initiliazers
+    // abstract class A {
+    //     abstract name: string = "";
```
MichaReiser, line 990: It also doesn't allow for definite assignments but does allow for optional tokens

```
abstract name!: string; // not ok
abstract name?: string; // ok
```

MichaReiser (commented):
Could you please update your PR summary and outline what's the scope of this PR and what work is still left to do (ambient context, declare members)

xtask/codegen/js.ungram
xunilrj: hum... ```ts```compiler is actually complaining about this:

```ts
abstract class Test {
  declare abstract method();
}
```


```
'declare' modifier cannot appear on class elements of this kind.(1031)
```

It is also complaining about all of these

```ts
abstract class Test {
    declare get age(): string;
    declare abstract get age2(): string;

    declare set age3(): string;
    declare abstract set age4(): string;
}
```

Altought, ```ts``` seems to be happy with:

```ts
abstract class Test {
    declare age: string;
    declare abstract age2: string;
}
```

crates/rome_formatter/tests/specs/ts/declaration/class.ts.snap
ematipico: Yes :) It was really difficult to understand whether a node was verbatim or not , so I added some additional check and put these info in the snapshot!

crates/rome_formatter/tests/specs/prettier/typescript/conformance/classes/classDeclarations/classAbstractKeyword/classAbstractMixedWithModifiers.ts.snap
ematipico: If you see something like this, it means that the file is **not** formatted due to a syntax error, most probably (there are a lot below). 

The presence of whitespaces is due to the fact that the new nodes format verbatim

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -450,6 +467,32 @@ fn parse_class_member(p: &mut Parser) -> ParsedSyntax {
     }
 }
 
+fn add_error_if_member_is_abstract(
```
ematipico, line 470: This function returns `ParsedSyntax`, then it should have the `parse_` prefix.

ematipico: Isn't the object of the function to *only* add the error? It should not return `ParsedSyntax`, right?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -860,7 +942,11 @@ fn property_declaration_class_member_body(
 }
 
 /// Parses the body of a property class member (anything after the member name)
-fn parse_property_class_member_body(p: &mut Parser, member_marker: Marker) -> ParsedSyntax {
+fn parse_property_class_member_body(
+    p: &mut Parser,
+    member_marker: Marker,
+    kind: JsSyntaxKind,
```
ematipico, line 948: Could add some documentation of what this `kind` is? I read that it's `JS_PROPERTY_CLASS_MEMBER` but I don't understand why it's passed. Maybe some doc could help?

xtask/codegen/js.ungram
ematipico: I think that the name is also misleading (at the moment). If this node happens to support `abstract` as mandatory, I think its name should contain `Abstract` in it. If not, then I think the name is fine (and as suggested by Micha, it should be optional)

xtask/codegen/js.ungram
```diff
@@ -734,6 +753,15 @@ JsGetterClassMember =
 	return_type: TsTypeAnnotation?
 	body: JsFunctionBody
 
+TsGetterSignatureClassMember =
+	access_modifier: ('protected' | 'public')?
+	'abstract'
```
ematipico, line 758: If `abstract` is mandatory, do you think it makes sense to add `Abstract` to the name of the node?

xtask/codegen/js.ungram
MichaReiser: We'll probably go with it as it is for now and tackle `declare` and ambient context in a separate PR. But the nodes are already designed in a way to support these as well.

xtask/codegen/js.ungram
xunilrj: I did something slightly different to use the current parser code:

```
TsPropertySignatureClassMember =
	access_modifier: ('protected' | 'public')?
	'abstract'
	'readonly'?
	name: JsAnyClassMemberName
	property_annotation: TsAnyPropertySignatureAnnotation?
	';'?

TsAnyPropertySignatureAnnotation =
	TsTypeAnnotation
	| TsOptionalPropertyAnnotation
```

xtask/codegen/js.ungram
xunilrj: Yes. Let us deal with ```declare``` and ambient in another PR.

crates/rslint_parser/src/syntax/class.rs
xunilrj: Definites I am testing here: https://github.com/rome/tools/pull/2121/files?file-filters%5B%5D=.rast&show-viewed-files=true#diff-4323b5853c84df574f080c5efedd6f23a520d1531d07c289cff5db43011ffd84R71

And optional I am testing here: https://github.com/rome/tools/pull/2121/files?file-filters%5B%5D=.rast&show-viewed-files=true#diff-07d12c98fb07e74b3b4b5a99f6c77950f83341c791fa1a3bfbfe6d048ae3b195R55

crates/rome_formatter/tests/specs/prettier/typescript/conformance/classes/classDeclarations/classAbstractKeyword/classAbstractSuperCalls.ts.snap
```diff
@@ -43,9 +43,14 @@ class A {
...
+  
     abstract foo();
-    bar() { super.foo(); }
-    baz() { return this.foo; }
+  bar() {
+    super.foo();
+  }
+  baz() {
+    return this.foo;
+  }
```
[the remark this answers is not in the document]
xunilrj: Done. It is parsing and formatting correctly now.

crates/rslint_parser/src/syntax/class.rs
xunilrj: Yes. But we need to return because we change the green nodes to Unkown inside of it.

xtask/codegen/js.ungram
xunilrj: In a future PR this abstract will be optional because we going to support this in ambient contexts and method overloading.

MichaReiser (approved):

ematipico (approved):

PR #2123, opened 2022-02-17
fix(rslint_parser): Type-arguments in `extends` heritage clause

### LINKED ISSUE #2122: TS: Correctly parse `extends` clause with generic type
Correctly parse `class A extends B<C>` as a generic type and not as a call expression.

MichaReiser:
## Summary

The parser incorrectly parsed `B<A>` in `class T extends B<A>` as the start of a call expression with type arguments (assumed it would be `B<A>()`) instead of an identifier with type arguments.

This PR adds a check when speculatively parsing the type arguments of a call expression to only eat them if they're followed by a token that is valid coming after type arguments (either a template literal or `(`.

We may need to extend the list of valid tokens in the future to get better error recovery.

## Test Plan

* Added a new parser test 
* Verified that the regressions are tests that should fail with errors emitted by a type checker

closes #2122

### DIFF
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -309,6 +309,11 @@ fn eat_class_heritage_clause(p: &mut Parser) {
     }
 }
 
+// test ts ts_extends_generic_type
+// type IHasVisualizationModel = string;
+// class D extends C<IHasVisualizationModel> {
+//     x = "string";
+// }
 fn parse_extends_clause(p: &mut Parser) -> ParsedSyntax {
     if !p.at(T![extends]) {
         return Absent;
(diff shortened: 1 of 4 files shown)


MichaReiser:
> ## Parser conformance results on ubuntu-latest
> ### TS
> Total 	15976 	15976 	0
> Passed 	11908 	11908 	0
> Failed 	4068 	4068 	0
> Panics 	0 	0 	0
> Coverage 	74.54% 	74.54% 	0.00%
> 🔥 Regression (56):
> 🎉 Fixed (56):
> ### T262
> Total 	45250 	45250 	0
> Passed 	44130 	44130 	0
> Failed 	1120 	1120 	0
> Panics 	0 	0 	0
> Coverage 	97.52% 	97.52% 	0.00%

Lol... this isn't satisfying at all...

ematipico (approved):

Issue #2113, opened 2022-02-16
📎 TS Parser Conformance


MichaReiser:
### Description

This issue tracks the remaining errors where the TypeScript parser fails to parse a valid TS script. 

* [ ] Abstract / ambient class members #2059 
  * `abstractPropertyBasics`
  * `ambientClassMergesOverloadsWithInterface`
  * `ambientExternalModuleWithoutInternalImportDeclaration`
  * `ambientExternalModuleWithInternalImportDeclaration`
  * `ambiguousCallsWhereReturnTypesAgree`
  * ...
* [x] #2128
  * `aliasUsageInIndexerOfClass`
  * `classImplementingInterfaceIndexer`
  * `classIndexer`
  * `classIndexer5`
  * `declFileClassWithIndexSignature`
  * `indexerA`
* [x] #2122
  * `aliasUsageInTypeArgumentOfExtendsClause`: `extends C<IHasVisualizationModel>` gets incorrectly parsed as `<IHas...>` to 
be a call expression
   * `baseTypeOrderChecking`
   * `baseTypeWrappingInstantiationChain`
   * .. 
* [x]  arrow parsing #2126 
  * `callOfConditionalTypeWithConcreteBranches`
* [x] async arrow with type parameters (@MichaReiser )
* [x] `cast`
  * `commentOnParenthesizedExpressionOpenParen1`
  * `m7Bugs`
  * `arrayBestCommonTypes`
  * `arrowFunctionWithObjectLiteralBody5`
  * `arrowFunctionWithObjectLiteralBody6`
* ~JSX~ (will be covered as a separate umbrella task - https://github.com/rome/tools/issues/2153)
  * `doubleUnderscoreReactNamespace`
  * `jsxFactoryIdentifierAsParameter`
  * `jsxFactoryQualifiedNameWithEs5`
  * `jsxPreserveWithJsInput`
  * `jsxEmitWithAttributes`
  * `jsxFactoryIdentifier`
  * `jsxFactoryQualifiedName`
*  ~decorators~ (non goal for now)
   * `commentOnDecoratedClassDeclaration`
   * `decoratorMetadataForMethodWithNoReturnTypeAnnotation01`
   * `decoratorMetadataNoStrictNull`
   * `decoratorMetadataPromise`
   * `decoratorMetadataConditionalType`
   * `decoratorMetadataOnInferredType`
   * ... 
* [x] #2129
   * `commonJsImportClassExpression`
[...]

PR #2127, opened 2022-02-17
fix(rslint_parser): cast followed by parenthesized expressions

MichaReiser:
## Summary

A cast expression followed by a parenthesized expression very much looks like the start of an arrow function except that it then misses the `=>` token.

```
<Test>(a)
```

This PR ensures that the speculative arrow function parsing doesn't parse this out as an arrow function but instead as a cast expression.

Part of #2113 

## Test Plan

Added a new parser test. Verified that the regressing tests are expecting type checker errors that we currently don't emit.


crates/rslint_parser/src/syntax/function.rs
```diff
@@ -395,13 +395,37 @@ pub(super) fn is_at_async_function(p: &Parser, should_check_line_break: LineBrea
     }
 }
 
+pub(super) enum Ambiguity {
+    Allow,
+    Disallow,
+}
```
ematipico, line 401: Nit: 

```suggestion
pub(super) enum Ambiguity {
    Allowed,
    Disallowed,
}
```

It sounds better in my head

ematipico: Would you mind add a comment explaining what kind of ambiguity we have to deal with, please? Without the context of the PR, it might be difficult to get it right away

ematipico (approved):

crates/rslint_parser/src/syntax/function.rs
MichaReiser: I documented the `ambiguity` in the method parsing the arrow function. The enum itself could be reused in other places as well. But I can add a little more documentation.

ematipico: That's the thing, too. I read the comment but I didn't understand what's the actual ambiguity. I needed to check the PR again to remember it, because the description of the PR gives more context. 

I am sure if I will stumble across this code again, I won't remember what's the ambiguity about.

MichaReiser: I added some more documentaiton. I hope that helps.

ematipico (approved):

PR #2130, opened 2022-02-17
fix(rslint_parser): Parse out expressions for rhs of `export = <rhs>`

MichaReiser:
## Summary

This PR fixes the definition and parsing of TypeScript `export = <rhs>` to allow for arbitrary expressions and not just identifier references or qualified names to match TypeScript's behaviour.

```ts
export = function () {} // function expression
export = class () { } // class expression
export = a; // identifier reference
export = a.b.c; // static member expression
export = 3 + 3 + 1; // arbitrary expression
```

## Test Plan

Added new parser tests. Verified that the regressions are tests that are expected to fail with type checker errors.


crates/rome_formatter/src/ts/any/module_ref.rs
```diff
@@ -1,12 +0,0 @@
-//! Generated file, do not edit by hand, see `xtask/codegen`
-
-use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
```
MichaReiser, line 3: These are files that I must have forgotten to delete before when removing these nodes from the grammar file.
They now got deleted thanks to @leops improvement to the formatter codegen!

ematipico (approved):

Issue #2128, opened 2022-02-17
Typescript: Class indexer member


xunilrj:
see https://github.com/rome/tools/issues/2113

- [x] Index Signature for
    - [x] Types
    - [x] Interfaces
    - [x] Classes
- Handling error for
    - [x] invalid modifiers

- Microsoft/Typescript tests:
    - [x] aliasUsageInIndexerOfClass
    - [x] classImplementingInterfaceIndexer
    - [x] classIndexer*
        - These need semantic validation: classIndexer2.ts, classIndexer3.ts, classIndexer4.ts 
    - [x] declFileClassWithIndexSignature
    - [x] indexerA

PR #2133, opened 2022-02-17
feat(rslint_parser): index member for classes

xunilrj:
see https://github.com/rome/tools/issues/2128


crates/rslint_parser/src/syntax/typescript/types.rs
```diff
@@ -869,14 +869,17 @@ fn is_at_ts_index_signature_type_member(p: &Parser) -> bool {
 // // not an index signature
 // type C = { [a]: string }
 // type D = { readonly [a]: string }
-fn parse_ts_index_signature_type_member(p: &mut Parser) -> ParsedSyntax {
+pub(crate) fn parse_ts_index_signature_type_member(
+    p: &mut Parser,
+    m: Option<Marker>,
+) -> ParsedSyntax {
     if !is_at_ts_index_signature_type_member(p) {
         return Absent;
```
MichaReiser, line 877: This will throw if the Marker is `Some`. I suggest to change the method to always accept a `Marker` and return `CompletedMarker`. Doing so delegates the responsibility for checking if the parser is indeed at an index signature to the caller (I sometimes called this methods `expect_ts_index_signature_type`)

crates/rslint_parser/test_data/inline/ok/ts_indexer_signature.rast
```diff
@@ -0,0 +1,303 @@
...
+            class_token: CLASS_KW@0..6 "class" [] [Whitespace(" ")],
+            id: JsIdentifierBinding {
+                name_token: IDENT@6..8 "A" [] [Whitespace(" ")],
+            },
+            type_parameters: missing (optional),
+            extends_clause: missing (optional),
+            implements_clause: missing (optional),
+            l_curly_token: L_CURLY@8..9 "{" [] [],
+            members: JsClassMemberList [
+                TsIndexSignatureTypeMember {
```
MichaReiser, line 16: Classes and type members so far didn't share any of the members. I'm somewhat undecided if this is good. The benefit of disjoint members is that a caller knows when something is a class member without probing the parent. However, it obviously leads to many similar nodes. 

What was your reasoning?

crates/rome_formatter/src/js/any/class_member.rs
```diff
@@ -12,6 +12,7 @@ impl ToFormatElement for JsAnyClassMember {
             Self::JsGetterClassMember(node) => node.to_format_element(formatter),
             Self::JsSetterClassMember(node) => node.to_format_element(formatter),
             Self::JsEmptyClassMember(node) => node.to_format_element(formatter),
+            Self::TsIndexSignatureTypeMember(_) => todo!(),
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript/types.rs
xunilrj: Done.

crates/rslint_parser/test_data/inline/ok/ts_indexer_signature.rast
xunilrj: I am going with specific members per context: Type, Class etc...
My reasoning here is that, if we find out this is overwhelming, we can come up with solutions; and in the worst-case scenario we can merge some in common nodes later, and the compiler will help us. 

The other way is less safe. We may end up with consumers "forgetting" about a node.

xtask/codegen/js.ungram
```diff
@@ -743,6 +744,16 @@ JsSetterClassMember =
 	')'
 	body: JsFunctionBody
 
+// class Test { [a: string]: number }
+//
+TsIndexSignatureClassMember =
+	'readonly'?
```
MichaReiser, line 750: It seems that index accessors can be static too

```
staticIndexSignature2
```

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -506,6 +509,29 @@ fn parse_class_member_impl(
...
+        // }
+        // class B {
+        //     [index: string]: { prop }
+        // }
+        // interface C {
+        //     [a: number]: string;
+        // }
+        // interface D {
+        //     [index: string]: { prop }
+        // }
```
MichaReiser, line 526: The tests for `interface`s seem out of place. This logic is related to class parsing.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -116,3 +122,90 @@ pub(crate) fn try_parse(
 
     res
 }
+
+/// Must be at `[ident:]` or `readonly [ident:]`
+pub(crate) fn is_at_ts_index_signature_member(p: &Parser) -> bool {
```
MichaReiser, line 127: There can be more modifiers. Ideally, we would eat away all class modifiers and then see if what comes next is a valid index signature. This should simplify the logic here a bit and also provide better error recovery if someone uses an invalid modifier.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -116,3 +122,90 @@ pub(crate) fn try_parse(
...
+
+bitflags::bitflags! {
+    /// Flags describing possible members separators
+    pub(crate) struct MembersSeparator: u8 {
+        /// Members can be separated by ','
+        const COMMA 		= 1 << 0;
+        /// Members can be separated by ';'
+        const SEMICOLON 	= 1 << 1;
+    }
+}
```
MichaReiser, line 153: Nit: Using `bitflags` here makes the code harder to understand than it must be. There are really only two situations

* `Comma` (for classes)
* `CommaOrSemicolon` 

Or you can even pass more information to the implementation so that it can decide itself to do the right thing by telling it what the parent is

```
enum IndexSignatureContext {
  Class,
  Type,
  Interface,
}
```

The enum could then even implement `expect_separator()`, `Into<JsSyntaxKind>`, ..

crates/rslint_parser/src/syntax/typescript/types.rs
```diff
@@ -709,8 +710,20 @@ impl ParseNodeList for TypeMembers {
...
-        return parse_ts_index_signature_type_member(p);
+    // test ts ts_index_signature_member
+    // type A = { [a: number]: string }
+    // type B = { readonly [a: number]: string }
+    // // not an index signature
+    // type C = { [a]: string }
+    // type D = { readonly [a]: string }
+    if is_at_ts_index_signature_member(p) {
+        let m = p.start();
+        return Present(expect_ts_index_signature_member(
```
MichaReiser, line 721: The error messages for semicolons now differs between index signatures and any other member.

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -116,3 +122,90 @@ pub(crate) fn try_parse(
...
+
+fn eat_members_separator(p: &mut Parser, possible_separators: MembersSeparator) {
+    let separator_eaten = possible_separators.contains(MembersSeparator::COMMA) && p.eat(T![,]);
+    let separator_eaten = separator_eaten
+        || (possible_separators.contains(MembersSeparator::SEMICOLON) && optional_semi(p));
+
+    if !separator_eaten {
+        let qty = possible_separators.bits().count_ones();
+
+        if qty > 1 {
```
MichaReiser, line 192: What does `qty` stand for?

crates/rslint_parser/src/syntax/typescript.rs
```diff
@@ -116,3 +122,90 @@ pub(crate) fn try_parse(
...
+            let mut tokens = vec![];
+            if possible_separators.contains(MembersSeparator::COMMA) {
+                tokens.push(T![,]);
+            }
+            if possible_separators.contains(MembersSeparator::SEMICOLON) {
+                tokens.push(T![;]);
+            }
```
MichaReiser, line 199: Nit: Would be nice if it's possible to `match`

MichaReiser (changes requested):
```
conformance/classes/indexMemberDeclarations/privateIndexer.ts
conformance/classes/indexMemberDeclarations/publicIndexer.ts
```

Can we add a check to disallow accessiblity modifiers for index accessors?

xtask/codegen/js.ungram
xunilrj: Only for classes; not for interfaces and types.
Done.

crates/rslint_parser/src/syntax/class.rs

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: The strange part now is that class members "eat" their modifiers outside of this method. "interfaces" and "types" eat them here.
We can try to unify this later.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: I think it is better now.

crates/rslint_parser/src/syntax/typescript/types.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/typescript.rs
xunilrj: I removed.

xunilrj:
!bench_parser

xunilrj:
> ```
> conformance/classes/indexMemberDeclarations/privateIndexer.ts
> conformance/classes/indexMemberDeclarations/publicIndexer.ts
> ```
> 
> Can we add a check to disallow accessiblity modifiers for index accessors?

Done.

crates/rslint_parser/test_data/inline/err/index_signature_class_member_in_js.rast
```diff
@@ -0,0 +1,84 @@
...
+    directives: JsDirectiveList [],
+    items: JsModuleItemList [
+        JsUnknownStatement {
+            items: [
+                CLASS_KW@0..6 "class" [] [Whitespace(" ")],
+                JsIdentifierBinding {
+                    name_token: IDENT@6..8 "A" [] [Whitespace(" ")],
+                },
+                L_CURLY@8..9 "{" [] [],
+                JsUnknown {
```
MichaReiser, line 12: Have you added the mapping for `IndexSignatureMember` to `JsUnknownMember` in `JsSyntaxKind.to_unknown_member`?

MichaReiser (approved):

crates/rslint_parser/test_data/inline/err/index_signature_class_member_in_js.rast
xunilrj: Done.
Also put ```UNKOWN_MEMBER```as valid interface member.

PR #2136, opened 2022-02-18
fix(rslint_parser): Move `arrow` parsing into `parse_assignment_expression_or_higher`

MichaReiser:
## Summary

Arrow functions are only allowed in [`AssignmentExpression` positions](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-assignment-operators) but the current implementation also parses out arrow functions in `PrimaryExpression` positions. 

This PR moves the parsing of arrow functions into the `parse_primary_expression` function. What makes parsing arrow function expressions challenging is that the head of an arrow function (everything before the `=>`) can as well be a parenthesized expression in many cases. For example:

```
(a, b, c, { d }) // parenthesized
(a, b, c, { d}) => {} // arrow function, because of the `{}`
```

This ambiguity means, the parser must speculatively parse the whole syntax between `(` and `)` to determine if it is a valid arrow function and, if not, re-parse it as a parenthesized expression. This works reasonably well in is also what TS is doing but causes issues if, for example, the parsing of a parameter's binding encounters any invalid syntax and, thus, error recovery kicks in:

```js
(a, 5 + 3) 
```

The parser will try to perform error recovery if this parenthesized expression is parsed as an arrow function because `5 + 3` isn't a valid binding. Our recovery is even good enough that the parser correctly recovers at the `)` which then gives the impression that the parser was able to correctly parse out the arrow function. 

This PR re-introduces the `speculative_parsing` state (previously called `no_recovery`) to disable error recovery when speculatively parsing any expression. Completely disabling error recovery means that our arrow function parameter recovery may not be as good as it could be because error recovery doesn't kick in. This is something that we can further improve in the future if it proves to be a problem. 

An alternative to this implementation would have been to introduce a new `ExpressionContext` flag that indicates if the current expression should be parsed as an arrow function or not. The downside I see there is that 

1. Requires an additional flag whereas moving it into the `parse_assignment_expression_or_higher` gives us the information for free
2. Parsing sub-expressions inherits the `ExpressionContext`. It would, therefore, have been required to reset the `allow_paren` flag for all calls parsing out sub-expressions that don't call into `parse_assignment_expression_or_higher`. Again, this is something we get for free when moving it into the appropriate function. 

Another benefit of the current implementation is that it moves the parsing of `arrow` functions into a single place. The parsing logic used to be spread all over the expression parsing: `async` functions, `type arguments`, parenthesized or not. Unifying the logic fixed issues with `async` functions with type arguments not being correctly parsed and the parser ignoring line breaks before `=>`

## Test Plan

* There are some formatter snapshot tests that now correctly pass. 
* Added new parser tests
* Fixes some parser tests that incorrectly used arrow function in primary expression positions.


MichaReiser:
!bench_parser

MichaReiser:
!bench_parser

MichaReiser:
!bench_parser

crates/rslint_parser/src/state.rs
```diff
@@ -36,6 +37,15 @@ pub(crate) struct ParserState {
     /// node that disallows duplicate bindings, for example `let`, `const` or `import`.
     pub duplicate_binding_parent: Option<&'static str>,
     pub name_map: IndexMap<String, Range<usize>>,
+
+    /// Indicates that the parser is speculatively parsing a syntax. For example, the syntax `(a, b) ...`
+    /// in JavaScript is either a parenthesized expression or an arrow expression if `...` is a `=>`.
+    /// However, the parser must first parse all of `(a, b)` to know if it's one or the other.
```
ematipico, line 43: The description doesn't connect the meaning of "parsing in a speculative way" and the example. First we should define what kind of difference we have in our parser when we have `speculative_parsing = true`, then we should connect it to the example above. Saying "speculatively parsing a syntax" doesn't help much; I would assume that our parser does some assumptions when `speculative_parsing = true` but what are these assumptions/speculations?

crates/rslint_parser/src/parser/parse_recovery.rs
```diff
@@ -74,6 +83,10 @@ impl ParseRecovery {
             return Err(RecoveryError::AlreadyRecovered);
         }
 
+        if p.state.speculative_parsing {
```
xunilrj, line 86: > /// Tries to recover by parsing all tokens into an `Unknown*` node until the parser finds any token
    /// specified in the recovery set, the EOF, or a line break (depending on configuration).

Wouldn't make sense that this "configuration" was passed as an argument?

crates/rslint_parser/src/state.rs
```diff
@@ -36,6 +37,15 @@ pub(crate) struct ParserState {
...
+    pub(crate) speculative_parsing: bool,
+
+    /// Stores the token positions of syntax that looks like arrow expressions but turned out to be none.
```
ematipico, line 46:
```suggestion
    /// Stores the token positions of a syntax that looks like arrow expressions but turned out to be a different one.
```

xunilrj (approved):

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -413,73 +420,298 @@ impl Ambiguity {
...
+///
+/// The header is everything coming before the body: `async (a) =>`.
 ///
 /// Returns `Err` if `ambiguity` is [Ambiguity::Disallowed] and the syntax
 /// is ambiguous. For example, the parser speculatively tries to parse `<string>(test)` as an arrow
 /// function because the start very much looks like one, except that the `=>` token is missing
-/// (it's a TypeScript `<string>` cast followed by a parenthesized expression.
-pub(super) fn parse_arrow_function(
+/// (it's a TypeScript `<string>` cast followed by a parenthesized expression).
+fn parse_parenthesized_arrow_function_head(
```
ematipico, line 436: Should we explain what's `Marker` inside `Ok` and the `Marker` inside `Err`? I doubt they are the same.

Also, this function doesn't return a `ParsedSyntax`, so we should remove the `parse_*` prefix. Maybe `try_parse_*` is a better fit?

crates/rslint_parser/src/parser/parse_recovery.rs
MichaReiser: You mean to `ParseRecovery` or in general?

Passing the speculative parsing as argument requires adding it to many parse methods:

* Parameters 
* Bindings (used in arguments)
* Expressions (because bindings can have initializers)
* Types (all of them)
* ... maybe even more 

That's why I opted for adding it to the `ParserState`. 

Passing it to `ParseRecovery` is an option but it already has access to the `Parser`, so it felt right to just test for it itself.

crates/rslint_parser/src/syntax/function.rs
```diff
@@ -413,73 +420,298 @@ impl Ambiguity {
...
+        // potential start of type parameters
+        T![<] => {
+            // <a...
+            if is_nth_at_identifier(p, n + 1) {
+                IsParenthesizedArrowFunctionExpression::Unknown
+            } else {
+                IsParenthesizedArrowFunctionExpression::False
+            }
+        }
+        _ => unreachable!(),
```
ematipico, line 651: Should we have an explanation of why it's unreachable?

crates/rslint_parser/src/parser/parse_recovery.rs
```diff
@@ -18,6 +18,14 @@ pub enum RecoveryError {
     ///    the array expression triggers another recovery. Handling this as an error ensures that list parsing
     ///    rules break out of the loop the same way as they would at the EOF.
     AlreadyRecovered,
+
+    /// Returned if the parser is speculatively parsing a branch. For example in JavaScript, the syntax
+    /// `(a, b, c) ...` can either be a parenthesized expression or an arrow function, depending on what
+    /// the first token in `...` is. This means, that the parser must try to parse the syntax either as
+    /// parenthesized or arrow function and revert if it turns out it was the other syntax.
+    /// It's important that the parser doesn't perform any error recovery in that case, which is when
+    /// `ParseRecovery` returns `RecoveryDisabled`.
```
ematipico, line 27: When we have "in that case", which case do you mean? When `"parenthesized or arrow function"` or when `"turns out it was the other syntax"`. I presume when  `"parenthesized or arrow function"`. I would suggest some rewording and maybe adding a reference to the example again, to make it more clear

crates/rslint_parser/src/state.rs
ematipico: If I understood correctly, when `speculative_parsing = true`, the recovery is completely disabled. I think it's important to mention it in the documentation because it's an important change on how the parser works. Then we can say why the recovery is "turned off" and then we can mention the example.

crates/rslint_parser/src/syntax/function.rs
MichaReiser: good point. Let's just return `False` in that case. This was an assumption based on where the function is called from which isn't very safe.

MichaReiser:
@ematipico I tried to improve the documentation based on your excellent feedback.

ematipico (approved):
Looks good now! Thank you!!

PR #2138, opened 2022-02-18
fix(xtask): disable the ssh and https features of libgit2

leops:
## Summary

This disables all default features of `libgit2` since they are not being used and removes a unneeded dependency on OpenSSL


MichaReiser:
Have you tested that `cargo codegen formatter` still works?

leops:
Well, "it works on my machine" but I'd be interested to see if it actually fixes the build issue @ematipico encountered

PR #2140, opened 2022-02-18
feat(rome_formatter): introduce hard_group_elements to fix the positioning of block comments

### LINKED ISSUE #2114: 🐛  Fix the formatting of block comments outside of groups
### Description

The current behavior of block comments in the formatter is to be printed at the position they were in the source if their parent group does not break, but as line suffixes if the parent group breaks. This is to support constructions like the following:

```js
[1, 2, /* comment */ 3, 4];

[
    1,
    2,
    3, /* comment */
    4,
];

while(true) { break /* comment */ }

while(true) {
    break; /* comment */
}
```

However in complex constructs that do not create a new group (function or class declarations for instance), block comments will defer their breaking behavior to the closest group they find, generally a block statement of the file itself: since these groups are always breaking, this leads to incorrect positioning of block comments:

```js
// Input
function /* 1 */ Name /* 2 */ () /* 3 */ {}
// Output before #2108
function Name() /* 3 */ {} /* 1 */ /* 2 */
// Output after #2108
function Name() {} /* 1 */ /* 2 */ /* 3 */

// Input
class /* 1 */ Name /* 2 */ extends /* 3 */ Parent /* 4 */ {}
// Output
class Name extends Parent {} /* 1 */ /* 2 */ /* 3 */ /* 4 */

// Input
if /* 1 */ (expr) stmt(); else /* 2 */ if /* 3 */ (expr) stmt(); else /* 4 */ stmt();
// Output
if /* 1 */ (expr) stmt(); else if /* 3 */ (expr) stmt(); else stmt(); /* 2 */ /* 4 */
```

leops:
## Summary

This PR introduces a new FormatElement, HardGroup, that forces elements directly inside it to be printed in flat mode. This behavior can in turn be escaped by introducing an inner Group element that will resume the normal breaking behavior of the printer.

This change is mostly aimed at fixing the printing of block comments inside constructs that have both a "breaking part" and a "non-breaking part". For instance with class declarations:

```js
   class /* comment */ class Example
// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ non-breaking part
{ // <
} // < breaking part
```

Placing the entire class declaration inside a hard group lets the "head" of the class declarations (`class /* comment */ Example`) be printed in "flat mode" while the body of the declaration is printed in "breaking mode" as it creates its own normal group. This would not have been possible by simply wrapping the class in a normal group as the breaking behavior of groups is "propagating" (any line break within a group will force it to break, including in inner groups)

Fixes #2114


crates/rome_formatter/src/format_element.rs
```diff
@@ -665,6 +665,34 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
     }
 }
 
+/// Forces all conditional elements ([if_group_breaks] and [if_group_fits_on_single_line])
+/// to behave as if they were contained in a non-breaking [Group]
```
ematipico, line 669: I really like this new concept. I wonder if we could expand the documentation by quoting the example of your PR. I find the explanation of the PR better to understand (the head group that  break and the body that can break)

ematipico (approved):

crates/rome_formatter/src/format_element.rs
```diff
@@ -665,6 +665,45 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
...
+/// let elements = group_elements(hard_group_elements(format_elements![
+///   if_group_breaks(token("not printed")),
+///   empty_line(),
+///   if_group_fits_on_single_line(token("printed")),
+/// ]));
+///
+/// assert_eq!("\nprinted", format_element(&elements, FormatOptions::default()).as_code());
+/// ```
+#[inline]
+pub fn hard_group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
```
MichaReiser, line 697: I'm somewhat conflicted by the name of the new group. It feels consistent, but at the same time, does not: 

* `hard`: is commonly used to mention that something get's enforced. Therefore, `hard_group` fits into that group of elements
* `hard`: is commonly used to refer that something gets force printed. With groups it's different. It actually forces not to print hard line breaks. 

How about:

* `flat_group_elements`
* `no_line_break_group_elements`

Overall, the name `..._group` elements doesn't work that well anymore because `group_elements` described an action.  But that might be something for another PR

crates/rome_formatter/src/js/statements/while_statement.rs
```diff
@@ -26,7 +27,19 @@ impl ToFormatElement for JsWhileStatement {
                 &r_paren_token?,
             )?,
             space_token(),
-            body.format(formatter)?
-        ])
+        ];
+
+        let body = body?;
+        if matches!(body, JsAnyStatement::JsBlockStatement(_)) {
```
MichaReiser, line 33: Nit: How about a `is_block_statement()` function on `JsAnyStatement`. Could even be something that we could generate. But  probably worth its own PR.

crates/rome_formatter/src/js/statements/with_statement.rs
```diff
@@ -26,7 +27,19 @@ impl ToFormatElement for JsWithStatement {
...
+            Ok(hard_group_elements(format_elements![
+                head,
+                body.format(formatter)?
+            ]))
+        } else {
+            Ok(format_elements![
+                hard_group_elements(head),
+                body.format(formatter)?
+            ])
+        }
```
MichaReiser, line 43: Is it worth introducing a helper method to format a header followed by a block statement?

crates/rome_formatter/src/printer.rs
```diff
@@ -197,20 +208,16 @@ impl<'a> Printer<'a> {
                 ));
             }
 
-            FormatElement::ConditionalGroupContent(ConditionalGroupContent {
-                mode: GroupPrintMode::Multiline,
-                content,
-            }) => {
-                queue.enqueue(PrintElementCall::new(content, args));
+            FormatElement::ConditionalGroupContent(ConditionalGroupContent { mode, content }) => {
+                if args.hard_group == matches!(mode, GroupPrintMode::Flat) {
```
MichaReiser, line 212: Is it possible that the "normal" print behaviour now re-implements some of the behaviour of the "flat" printing mode? Would it be possible to, instead, call into "flat-printing" or merge the two methods?

leops: The normal printing mode does implement flat printing too now, but the main difference between this and the flat printing mode is the control flow: `try_print_flat_element` can abort with an error (and does if it encounters a line break) while `print_element` always succeeds. Since this implies two different method signatures I'm not sure there's an ergonomic way to merge the two

crates/rome_formatter/src/format_element.rs
leops: To be honest I don't really like the `hard_group_elements` either, and I think `flat_group_elements` is better. Alternatively it could also be named `flatten_elements` or `flat_print_elements` which would fit the "describe an action" requirement although I'm not sure those names accurately reflect what the function actually does either

MichaReiser: `flatten_elements` reminds me of `flat_map`.

The benefit we have is that renaming it in the future is straightforward because of Rust. So we shouldn't spent too much time figuring out the right name now. Maybe something better comes up in the future. 

I guess the main problem is that `group_elements` isn't very telling either. OK, it groups the elements but what does that mean? Would it help if `group_elements` had a second argument specifying how the grouping should work? 

Anyhow. Dont let this block your PR. Seems like something that may become more obvious in the near future (or turn out to not be a problem after all)

crates/rome_formatter/src/js/statements/with_statement.rs
leops: I considered moving this logic to a method on the Formatter but ended up not doing it because the formatter already has dozen helper methods, most of them generic while this one is starting to get specific for a given use-case. Maybe this would be a better fit for a free-standing function in utils

MichaReiser (approved):

PR #2141, opened 2022-02-18
refactor(rome_formatter): use node destructuring in JS

leops:
## Summary

This is an automated refactor of `rome_formatter/src/js` to use node destructuring (`as_fields`) instead of individual accessor methods, coupled with a few manual fixes for various and warnings in the second commit (mostly missing typescript nodes and tokens)


ematipico (approved):
If no tests break, I'd say this PR can be merged :)

PR #2146, opened 2022-02-21
feature(rslint_parser): Add TS `export default` extension support

MichaReiser:
## Summary

Adds support for default exporting `interface`s and declare functions:

```ts
export default interface A {}
export default function test(a: string): string;
export default function test(a: string | undefined): string { return "hello" }
```

I decided to refactor the export default clause so that fewer nodes in total are needed:

* Introduce `ExportDefaultDeclarationClause` for any `export default <declaration>` (aligns nicely with `ExportDefaultExpressionClause`)
* Rename `ExportDefaultFunctionClause` and `ExportDefaultClassClause` to `ClassExportDefaultDeclaration` and `FunctionExportDefaultDeclaration`
* Add `TsDeclareFunction` and `TsInterfaceDeclaration` to the enum

## Test Plan

Added new parser tests


ematipico:
Are the regressions expected?

MichaReiser:
> Are the regressions expected?

The regressions are OK. These tests succeeded before because the tests expect an error and we aren't asserting if these are the correct errors (used to fail because of arrow parsing, but they should fail because of some type checking error).

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -129,17 +129,14 @@ pub(super) fn parse_class_declaration(p: &mut Parser, context: StatementContext)
...
 // export default abstract class {}
-pub(super) fn parse_export_default_class_case(p: &mut Parser) -> ParsedSyntax {
-    let is_at_default_class = p.at(T![default]) && p.nth_at(1, T![class]);
-    let is_at_default_abstract_class = p.at(T![default])
-        && is_nth_at_contextual_keyword(p, 1, "abstract")
-        && p.nth_at(2, T![class]);
-    if !is_at_default_class && !is_at_default_abstract_class {
+pub(super) fn parse_class_export_default_declaration(p: &mut Parser) -> ParsedSyntax {
+    let is_at_class = p.at(T![class]);
+    let is_at_abstract_class = is_at_contextual_keyword(p, "abstract");
```
ematipico, line 134: Why do we check two words at the same position? Should we not check `class` at nth `1` if we find the `abstract` keyword? What if we have `export default abstract function() {}`?

The old code was doing it

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -1063,6 +1076,59 @@ fn parse_export_default_clause(p: &mut Parser) -> ParsedSyntax {
...
+            })
+        }
+    };
+
+    declaration.or_add_diagnostic(p, |_, range| {
+        expected_any(
+            &[
+                "class declaration",
+                "function declaration",
+                "interface declaration",
```
ematipico, line 1123: We should  have "interface declaration" only if we are parsing a TS file, otherwise we would risk to give wrong suggestions to the user

ematipico (commented):
Left some question for some clarifications

crates/rslint_parser/src/syntax/class.rs
MichaReiser: I updated the implementation here and for `parse_class_declaration`.

ematipico (approved):

PR #2147, opened 2022-02-21
feat(rome_formatter): track verbatim code inside snapshots

ematipico:
## Summary

This PR implements a way to track code that is formatted using `format_verbatim`, this is useful particularly during the implementation of new nodes.

Now that we have a codegen that automatically creates a default implementation, sometimes it's difficult to track which nodes correctly implements **our** new implementation of `ToFormatElement` against the default one.

This PR does:
- create a new API called `format_unknown` which is specifically created and called by unknown nodes
- change the implementation of `format_verbatim`  , which creates a newly element called `FormatElement::Verbatim`
- create a new  struct `Verbatim` where we store information needed for the snapshots
- write unimplemented nodes/tokens inside the snapshot

The PR creates some noise around the prettier test suite because the line of assertion has changed. You can ignore them.
The PR uncovered few cases which will be fixed in a consecutive PR, I would like to leave them on purpose to show how they work.

## Test Plan

Checked that the updated tests are correctly tracked

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -876,13 +876,18 @@ pub enum FormatElement {
 
     LineSuffix(Content),
 
+    /// A token that tracks tokens/nodes that are printed using [`format_verbatim`](Formatter::format_verbatim) API
     Verbatim(Verbatim),
 }
 
+/// Information of the node/token formatted verbatim
 #[derive(Clone, Eq, PartialEq)]
 pub struct Verbatim {
+    /// the range that belongs to the node/token formatted verbatim
     pub range: TextRange,
+    /// the text that belongs to the node/token formatted verbatim
     pub text: String,
+    /// The [FormatElement] version of the node/token
     pub element: Box<FormatElement>,
 }
 
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -875,6 +875,25 @@ pub enum FormatElement {
     Token(Token),
 
     LineSuffix(Content),
+
+    Verbatim(Verbatim),
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct Verbatim {
+    pub range: TextRange,
+    pub text: String,
+    pub element: Box<FormatElement>,
+}
+
+impl Verbatim {
+    pub fn new(element: FormatElement, text: String, range: TextRange) -> Self {
+        Self {
+            element: Box::new(element),
+            range,
+            text,
+        }
+    }
 }
 
 impl Debug for FormatElement {
@@ -905,6 +924,7 @@ impl Debug for FormatElement {
             FormatElement::LineSuffix(content) => {
                 fmt.debug_tuple("LineSuffix").field(content).finish()
             }
+            FormatElement::Verbatim(verbatim) => verbatim.element.fmt(fmt),
         }
     }
 }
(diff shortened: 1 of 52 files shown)


leops:
Down the line release builds will not ship with any unimplemented nodes, so as this code will never actually get called it will probably not have much if any overhead (it's even possible it will completely get optimized out). On the other hand adding `#cfg(debug_assertions)]` attributes all over the code might make it less readable, so overall I'm okay with just leaving this as it is.

For the assert line number change, I think `insta` will skip updating the snapshot if its the only difference: if you discard all the changes made to the snapshots and regenerate them again only the ones that have unimplemented ranges should get updated.

crates/rome_formatter/src/formatter.rs
```diff
@@ -411,25 +412,44 @@ impl Formatter {
...
+    /// doesn't track nodes/tokens.
+    pub fn format_unknown(&self, node: &SyntaxNode) -> FormatElement {
+        self.format_verbatim_node_or_token(node)
+    }
+
+    fn format_verbatim_node_or_token(&self, node: &SyntaxNode) -> FormatElement {
+        concat_elements(node.children_with_tokens().map(|child| {
+            match child {
+                SyntaxElement::Node(child_node) => {
+                    // TODO: Add source map markers before/after node as well as any additional elements that
```
leops, line 433: This TODO might be out of date, with the current implementation sourcemap data is stored per-token instead of per-node

crates/rome_formatter/src/format_element.rs
```diff
@@ -1260,6 +1280,7 @@ impl FormatElement {
             FormatElement::List(list) => FormatElement::List(list.trim_start()),
             FormatElement::Token(s) => FormatElement::Token(s.trim_start()),
             FormatElement::LineSuffix(s) => FormatElement::LineSuffix(Box::new(s.trim_start())),
+            FormatElement::Verbatim(v) => v.element.trim_start(),
```
leops, line 1283: As I understand this, trimming a `Verbatim` element will return a new non-verbatim element. Since the only place `trim_start` or `trim_end` are currently used is on verbatim elements (in `format_list`) it seems like this code could be erasing some of the issues this PR intends to catch

crates/rome_formatter/src/formatter.rs
```diff
@@ -411,25 +412,41 @@ impl Formatter {
...
+        self.format_verbatim_node_or_token(node)
+    }
+
+    fn format_verbatim_node_or_token(&self, node: &SyntaxNode) -> FormatElement {
+        concat_elements(node.children_with_tokens().map(|child| {
+            match child {
+                SyntaxElement::Node(child_node) => {
+                    // Here we call `format_unknown` because we don't want to track it as [FormatElement::Verbatim]
+                    // all the possible children. The first node to call `format_verbatim` should be node to be tracked
+                    self.format_unknown(&child_node)
```
leops, line 435: The comment explains why this is not calling `format_verbatim` again, but what's the reason for calling back into `format_unknown` instead of immediately recursing into `format_verbatim_node_or_token` ?

crates/rome_formatter/src/printer.rs
```diff
@@ -245,6 +248,13 @@ impl<'a> Printer<'a> {
                     .line_suffixes
                     .push(PrintElementCall::new(&**suffix, args));
             }
+            FormatElement::Verbatim(verbatim) => {
+                self.state
+                    .verbatim_markers
+                    .push((verbatim.text.clone(), verbatim.range));
+                // queue.enqueue(PrintElementCall::new(&verbatim.element, args));
```
leops, line 255: I think it's generally preferred to write back the content to the queue instead of recursing into `print_element` (at least I know profiling data is easier to read with short call stacks)

crates/rome_formatter/src/formatter.rs
ematipico: Good catch! The comment was written before a final refactor

crates/rome_formatter/src/format_element.rs
ematipico: True, and this uncovered a bug and I have now fixed!

leops (approved):
It might be lacking some documentation around the new Verbatim struct and associated functions but on the functional side its good to go

PR #2148, opened 2022-02-21
fix(rslint_parser): TS `asserts` predicate

MichaReiser:
## Summary

TypeScript supports two kind of type refinement functions:

**[Type Predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)**

```ts
function isString(a): a is string {
  return typeof a === "string"
}

function test(a) {
  if isString(a) {
    console.log("a is a string"); // The type of `a` is string
  }
}
```

The function returns a boolean whatever the specified argument fulfills the condition.

**[Type Assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions)**

```ts
function assertString(a): asserts a is string {
  if typeof a !== "string" {
    throw new Error("A is supposed to be a string");
  }
}

function test(a) {
  assertString(a);
  console.log(a); // The type of `a` is string
}
```

Asserts are in that sense special that they also allow omitting the `is ...` part.

```ts
function asserts(condition: any): asserts condition {
  if (!condition) {
    throw new Error("Whoops");
  }
}

function test(a) {
  asserts(a > 10);
  // a is guaranteed to be bigger than 10 and has the type `number`
}
```

This second part is something that our current AST didn't model correctly

This PR introduces a new `TsAssertsReturnType` for `asserts...` refinements and renames `TsTypePredicate` to `TsPredicateReturnType` because these are types (they appear in type positions).

## Test Plan
Extended the existing parser tests with an example that uses `asserts <param_name>`

Regressions: 
* The regressions are because of tests that are supposed to fail and used to fail because of parsing errors instead of the supposed type checking errors.


ematipico (changes requested):
`TsAnyTypePredicateParameterName` includes `TsThisType` which is not a valid syntax inside a function. Which means that we parser should check if the parameter  is not the word `this`.

```ts
// invalid
function a(this: string): asserts this is string {}

// valid
interface A {
    (this: string): asserts this is string
}
```

MichaReiser:
> `TsAnyTypePredicateParameterName` includes `TsThisType` which is not a valid syntax inside a function. Which means that the parser should check if the parameter is not the word `this` and throw an error.
> 
> ```ts
> // invalid
> function a(this: string): asserts this is string {}
> 
> // valid
> interface A {
>     (this: string): asserts this is string
> }
> ```

Nice catch! The current focus is around passing all valid TS syntax but not emitting diagnostic for all possible invalid syntax. For example, the parser also doesn't verify that the parameter name is indeed a valid parameter name.

ematipico:
OK, thank you for the clarification. Sometimes it's also difficult to understand if an error is a `SyntaxError` or a type check error.

ematipico (approved):

PR #2149, opened 2022-02-21
feat(rome_formatter): improve the formatting of if-else chains

leops:
## Summary

This improves the formatting of `if(...) ... else if(...) ...` chains by flattening them and formatting them as a single group: this means the entire chain will either be printed on a single line if it fits, or each branch will be broken into a different line with an indent block

```js
if (1) 1; else if (2) 2; else 3;

if (very_long_condition_1)
	very_long_statement_1();
else if (very_long_condition_2)
	very_long_statement_2();
else
	very_long_statement_3();
```

### DIFF
--- a/crates/rome_formatter/src/js/statements/if_statement.rs
+++ b/crates/rome_formatter/src/js/statements/if_statement.rs
@@ -1,72 +1,93 @@
 use crate::formatter_traits::{FormatOptionalTokenAndNode, FormatTokenAndNode};
+use crate::{concat_elements, group_elements, hard_group_elements, soft_block_indent};
 
 use crate::{
-    format_elements, group_elements, hard_group_elements, space_token, FormatElement, FormatResult,
-    Formatter, ToFormatElement,
+    format_elements, space_token, FormatElement, FormatResult, Formatter, ToFormatElement,
 };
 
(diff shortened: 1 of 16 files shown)


crates/rome_formatter/src/js/statements/if_statement.rs
```diff
@@ -1,72 +1,91 @@
...
+        )?,
+        space_token(),
+        into_block(formatter, consequent?)?,
+    ];
+
+    Ok((head, else_clause))
+}
+
+/// Wraps the statement into a block if its not already a JsBlockStatement
+fn into_block(formatter: &Formatter, stmt: JsAnyStatement) -> FormatResult<FormatElement> {
```
ematipico, line 81: This is an utility that we could use somewhere else, what do you think? Should we move it inside utils?

ematipico (approved):
Looks good to me, I love the new formatting!

crates/rome_formatter/src/js/statements/if_statement.rs
leops: Yes that's the general idea, I was planning to do that in a followup PR to change the formatting of all the other "single statement" statements (while, do-while, for and with)

PR #2150, opened 2022-02-21
fix(rslint_parser): TS nested type arguments

MichaReiser:
## Summary

The lexer incorrectly lexes `<<` as a left shift operator for type arguments that contain a nested call signature:

```ts
type Bar = ReturnType<<T>(x: T) => number>;
```

This PR changes the lexer to lex `<<` as two separate `<` tokens and let the parser decide if it's a left shift operator or not as the parser already does for `>>` and `>>>`.

Part of #2113 

## Test Plan

Added a new parser test.

### DIFF
--- a/crates/rslint_lexer/src/lib.rs
+++ b/crates/rslint_lexer/src/lib.rs
@@ -187,17 +187,23 @@ impl<'src> Lexer<'src> {
         chr
     }
 
-    // Get the current byte
+    /// Get the current byte
     #[inline]
     fn current(&mut self) -> Option<&u8> {
         self.bytes.get(self.cur)
     }
 
-    // Get the next byte and advance the index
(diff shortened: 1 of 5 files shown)


MichaReiser:
The syntax is even more subtle than that and implementing that would need access to trivia inside the parser, something we currently don't support. 

The issue is that this syntax (pay attention to the space between `< <` is valid:

```ts
var r2 = < <T>(x: T) => T>f; // valid
```
but 

```
var r2 = <<T>(x: T) => T>f; // valid
```

isn't. 

The way TypeScript implements this is that they always parse `<<` as a left shift token but have a `reScanLessThan` function on the scanner for cases where a single `<` is valid as well (type arguments, for example). 

We'll be able to do the same when #2035 is implemented CC: @xunilrj

PR #2151, opened 2022-02-21
feat(rome_formatter): add the Prettier Typescript tests

leops:
## Summary

This adds the TypeScript tests from the Prettier test suite to the existing port of the JS tests in the tests for the formatter. It also fixes a few bugs caught by the additional test cases (mostly TS-specific tokens or nodes being skipped over in existing JS formatters)


ematipico:
For thew fixes that you caught, that's because we are not using the `as_fields()`. Should we move those implementations to use it? Probably your automated fixed didn't catch some files.

leops:
The automated refactor I ran to use `as_fields` was only on the JS formatters since the TypeScript implementation is changing a lot so it would have a lot of merge conflicts, and has two notable exceptions in `any/class.rs` and `any/function.rs` as those don't have an auto-generated fields struct due to being `Any*` unions instead of plain AST nodes

PR #2152, opened 2022-02-21
feat(rome_formatter): format typescript, part 3

ematipico:
## Summary

Part of #2069 

This PR adds formatting most of the types we have in TS and some nodes inside classes.

I also changed a bit the grammar as we had a `separator` token which was not really separator but a simple `;`.

This PR doesn't implement any particular logic for the formatting of conditional types and unions/intersections types because they will need a particular logic.

## Test Plan

New tests


crates/rome_formatter/src/ts/types/function_type.rs
```diff
@@ -1,7 +1,26 @@
...
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let TsFunctionTypeFields {
+            parameters,
+            fat_arrow_token,
+            type_parameters,
+            return_type,
+        } = self.as_fields();
+
+        Ok(format_elements![
```
leops, line 17: This should be wrapped in a `hard_group_elements` to ensure the placement of comments is coherent with `JsAnyFunction`

leops (approved):

PR #2158, opened 2022-02-22
chore(coverage): Support multiple test files per test case

MichaReiser:
## Summary

A TypeScript test file can define multiple files:

```
// @Filename: tsconfig.json
{
  "compilerOptions": {
    "allowJs": true,
    "outDir": "foo",
    "isolatedModules": true,
  }
}

// @Filename: index.js
module.exports = {}
var x = 1
```

This test defines 2 files, a `tsconfig.json` and an `index.js`.

This PR adds support for multiple files to our test runner and extracts the different files for the TypeScript test suite. The test runner parses each individual file and aggregates the diagnostics when asserting the outcome.

This will enable us to extract more metadata in the future, for example, respect the `@declaration` flag once the parser supports a TypeScript declaration file kind.

## Regressions

These are tests that are expected to fail and used to fail because they contain a `.json` file (that our parser can't parse). This PR now skips the JSON files so that all files can now be parsed correctly (and thus, no longer fail with an unrelated error).

### DIFF
--- a/xtask/coverage/README.md
+++ b/xtask/coverage/README.md
@@ -23,8 +23,8 @@ You can filter the tests by passing an additional `--filter` so that only the te
 cargo coverage --filter=source
 ```
 
-The `--detailed` parameter enables a more detailed summary at the end of the test run that includes diagnostic and the `rast` (when using--detailed=rast`)
-for all failing tests. `--detailed` can also be useful if you want to pipe the results to a text file.
+The `--detailed` parameter enables a more detailed summary at the end of the test run that includes diagnostic.
+You can use `--detailed-=debug` if you're debugging a test (prints the AST and diagnostics, even for passing tests). `--detailed` can also be useful if you want to pipe the results to a text file.
 
 
 ```bash
--- a/xtask/coverage/README.md
+++ b/xtask/coverage/README.md
@@ -24,7 +24,7 @@ cargo coverage --filter=source
 ```
 
 The `--detailed` parameter enables a more detailed summary at the end of the test run that includes diagnostic.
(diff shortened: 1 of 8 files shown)


xunilrj (approved):

xtask/coverage/README.md
```diff
@@ -23,8 +23,8 @@ You can filter the tests by passing an additional `--filter` so that only the te
 cargo coverage --filter=source
 ```
 
-The `--detailed` parameter enables a more detailed summary at the end of the test run that includes diagnostic and the `rast` (when using--detailed=rast`)
-for all failing tests. `--detailed` can also be useful if you want to pipe the results to a text file.
+The `--detailed` parameter enables a more detailed summary at the end of the test run that includes diagnostic.
+You can use `--detailed-=debug` if you're debugging a test (prints the AST and diagnostics, even for passing tests). `--detailed` can also be useful if you want to pipe the results to a text file.
```
ematipico, line 27: `--detailed-=debug` is it correct? Is there an additional dash `-`?

xtask/coverage/src/reporters.rs
```diff
@@ -122,11 +117,8 @@ pub enum SummaryDetailLevel {
     Coverage,
     /// Prints the coverage table as well as all failing tests with their diagnostics
     Failing,
-    /// Prints the diagnostics of failing tests as well as of tests that are supposed to fail and correct
-    /// emitted diagnostics
-    AllDiagnostics,
-    /// Prints all failing tests with their RAST output
-    FailingWithRast,
+
+    Debug,
```
ematipico, line 121: Can we add a small description?

ematipico (approved):

PR #2159, opened 2022-02-23
feat(rslint_parser): Replace `Syntax` with `SourceType`

MichaReiser:
## Summary

The definition of the existing `Syntax` struct isn't flexible enough to fully support the different TS/JS combinations. For example:

* TypeScript files can be modules or scripts.
* TypeScript supports declaration files
* The support of top-level await depends on the language version

This PR replaces the `Syntax` struct with a `SourceType` struct that consists of:

```
pub struct SourceType {
    language: Language,
    variant: LanguageVariant,
    module_kind: ModuleKind,
    version: LanguageVersion,
}
```

* Language: JS or TS (TS has a flag for declaration file)
* variant: Standard or JSX
* module_kind: Module / Script (maybe CommonJS at some point?)
* version: Not used today but the ECMA Script version of the file

Correctly representing this combination revealed that:

* TypeScript files weren't parsed as a module (or in strict mode)
* A TS file can have multiple `export default` for function overloads, interfaces or if they appear in different ambient contexts (e.g. two `declare module` blocks).
* The coverage didn't strip the BOM from test files, causing issues when parsing out the individual test files of a test case
* `await` is supported in ambient context.
* Files with the option `alwaysStrict` must be parsed as if they have a `"use strict"` directive at the top
* Some coverage tests run with multiple `module` options and the expected errors file is then named `<testname>(module=<module>).errors.txt`

The current implementation infers the `SourceType` from the file extension. TypeScript (and babel with `sourceType: 'unambiguous'`) at the other parse the file as a module if it contains any module syntax and as script otherwise. Babel does so by first parsing it as a script and re-parse it as a module if it encounters any module syntax. This isn't very efficient. I don't know how we want to handle this in the future. If this is something the language server can infer, is tracked in a config, or something else. 

This struct isn't intended to be used to determine which features should be transpiled during a transform. I believe that we'll want a more granular representation so that individual syntax can be toggled on and off.

## Test Plan

Verified that the regressions are tests that are supposed to fail but now pass because the files are now correctly parsed as module/script/in strict mode...


xunilrj:
> variant: Standard or JSX

Would make sense to model JSX and others as "features flags"?

ematipico:
About the data structure, I would suggest the following change:

```
pub struct SourceType {
    language: Language(Variant, Version),
    module_kind: ModuleKind
}
```

I think versions and variations are really tight to the language itself, not to the source type. The proposed structure could hint to have combinations that are technically wrong, e.g. `Language::JS` and `Variant::TSX`.

I know that the fields are private and we decide the combinations via public methods, but the data structure seems off.

MichaReiser:
!bench_parser

MichaReiser:
> About the data structure, I would suggest the following change:
> 
> ```
> pub struct SourceType {
>     language: Language(Variant, Version),
>     module_kind: ModuleKind
> }
> ```
> 
> I think versions and variations are really tight to the language itself, not to the source type. The proposed structure could hint to have combinations that are technically wrong, e.g. `Language::JS` and `Variant::TSX`, which is not correct.
> 
> I know that the fields are private and we decide the combinations via public methods, but the data structure seems off.

I initially had the same model but then realised that `JS` and `TS` implement all variants (`Standard` and `JSX). That's why I moved the variant out of the language since the two are orthogonal (independent of each other).

xtask/coverage/src/util.rs
```diff
@@ -0,0 +1,66 @@
+use std::char::{decode_utf16, DecodeUtf16Error};
+
+pub(crate) enum DecodedString<'a> {
```
leops, line 3: This could be using `Cow<'a, str>` instead of introducing a new type, this would be consistent with the standard library as its the type returned by similar functions like `to_string_lossy` for instance

crates/rome_formatter/tests/specs/prettier/typescript/ambient/ambient.ts.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/prettier_tests.rs
-assertion_line: 124
+assertion_line: 119
```
leops, line 3: Nit: the diff for this file could be discarded

crates/rome_formatter/src/lib.rs
```diff
@@ -488,8 +488,8 @@ pub fn format_element(element: &FormatElement, options: FormatOptions) -> Format
 pub fn format_file_and_save(rome_path: &mut RomePath, options: FormatOptions, app: &App) {
     let result = if app.can_format(rome_path) {
         let buffer = rome_path.get_buffer_from_file();
-        let syntax = Syntax::default().module();
-        let root = parse(buffer.as_str(), 0, syntax).syntax();
+        let source_type = SourceType::from_path(rome_path.as_path()).unwrap_or_else(SourceType::js);
+        let root = parse(buffer.as_str(), 0, source_type).syntax();
```
ematipico, line 492: We have to add `mjs`, `mts`, `cjs` and `cts` to the list of files that we can format: https://github.com/rome/tools/blob/main/crates/rome_core/src/file_handlers/mod.rs#L18-L36


Otherwise `app.can_format` will return `false` even though the parser is ready. If that's the intention, then we should update that file too.

crates/rslint_parser/src/lib.rs
```diff
@@ -127,96 +127,192 @@ pub trait TreeSink {
...
+                "jsx" => Self::jsx(),
+                "ts" | "mts" => Self::ts(),
+                "cts" => Self::ts().with_module_kind(ModuleKind::Script),
+                "tsx" => Self::tsx(),
+                _ => return None,
+            }
+        };
+
+        Some(source_type)
+    }
```
ematipico, line 281: We have [`file_handlers`](https://github.com/rome/tools/blob/main/crates/rome_core/src/file_handlers/mod.rs) that does the same thing, in a less fancy way (and less info).

I wonder how this code will interpolate with what we already have? It seems that there are some repetitions here in terms of logic, so I would like to understand what's your plan after this PR.

ematipico (changes requested):
Could you please revisit the description of the PR? It seems that there are more changes than expected (code coverage, parser, etc.) which are not tracked in the PR.

It would be preferable to split them. But if not possible, would you mind share some context about the changes please?

I also share some thoughts about this new `SourceType`: it makes sense but on the other hand it also deals with file extensions (file system), which, I think, should not be responsible for it. We have `core` and `file_handlers` that do that and should keep doing it. The parser, hence `SourceType` should have knowledge of paths and files. Maybe it would make more sense to move it outside the parser?

Also, I would like to understand how this change **will play** with file handlers

xtask/coverage/src/util.rs
MichaReiser: Nice, I invented my own `Cow` (except that it doesn't implement `Deref` and other niceties). That will help me remember when to use it in the future.

crates/rslint_parser/src/lib.rs
MichaReiser: I agree, the `SourceType`, specifically the part where we construct the `SourceType` from a path is overlapping with `RomePath` and `FileHandler`. 

The reason why I added this only for the parser is that I don't know how this should work for multiple languages because the `SourceType` will probably look different for different languages. I don't feel comfortable tackling this now without knowing how we design other parts of our multi-language architecture.

MichaReiser:
> Could you please revisit the description of the PR? It seems that there are more changes than expected (code coverage, parser, etc.) which are not tracked in the PR.

I'll update the PR description. Many of these changes were necessary to address regressions caused by the fact that TypeScript files now get parsed as `Module`s.

crates/rslint_parser/src/lib.rs
```diff
@@ -127,96 +127,193 @@ pub trait TreeSink {
...
+impl Default for LanguageVersion {
+    fn default() -> Self {
+        Self::latest()
+    }
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum ModuleKind {
+    Script,
+    Module,
```
ematipico, line 151: Did we remove the "snippet kind" on purpose (`JsExpressionSnipped`)? (I don't even know what was its real usage)

crates/rslint_parser/src/lib.rs
```diff
@@ -127,96 +127,193 @@ pub trait TreeSink {
...
+impl Default for ModuleKind {
+    fn default() -> Self {
+        ModuleKind::Module
+    }
+}
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum LanguageVariant {
+    Standard,
+    JSX,
```
ematipico, line 172: Can we add a comment saying that `JSX` also covers `.tsx` files?

crates/rslint_parser/src/lib.rs
```diff
@@ -127,96 +127,193 @@ pub trait TreeSink {
...
+                definition_file: false,
+            },
+            ..Self::default()
         }
-        this
     }
 
-    pub fn top_level_await(mut self) -> Self {
-        self.top_level_await = true;
+    /// language: JS, variant: JSX, module_kind: Module, version: Latest
```
ematipico, line 245:
```suggestion
    /// language: TS, variant: JSX, module_kind: Module, version: Latest
```

crates/rslint_parser/src/syntax/module.rs
```diff
@@ -1050,33 +1050,56 @@ fn parse_export_default_clause(p: &mut Parser) -> ParsedSyntax {
...
-            p.error(err);
-            clause.change_kind(p, JsSyntaxKind::JS_UNKNOWN);
-        } else {
-            p.state.default_item = Some(clause.range(p).into());
+        if let Some(existing_default_item) = p.state.default_item.as_ref().filter(|_| p.is_module())
+        {
+            if existing_default_item.kind.is_overload()
+                && (default_item_kind.is_overload() || default_item_kind.is_function_declaration())
+            {
+                // It's ok to have multiple overload declarations and an implementation
```
ematipico, line 1069:
```suggestion
                // It's ok to have multiple overload declarations and an implementation.
```

ematipico (approved):

MichaReiser:
> I agree the `LanguageVariant` concept could be expressed as a bitset, however it seems unclear to me at the time how it would be structured and what would / would not go into that (for instance TypeScript could also technically be considered a variant / extension of JavaScript) so I think the current enum approach is fine

I also don't consider this to be the final design. I'm fairly certain that we'll re-iterate a couple of times before we get it right but it increases the correctness bar for any new format

crates/rslint_parser/src/lib.rs
MichaReiser: The `parse_expression_snipped` still exists. The method must be used when you want to parse a single expression. Just calling `parse_expression` doesn't work because the `Expression` node doesn't allow any `EOF` token, `ExpressionSNipped` does.

`JsExpressionSnipped` isn't part of this enum because it doesn't change the parsing semantic. An expression parsed with `ModuleKind` will be parsed in strict mode and doesn't allow `await` as an identifier. If you change the `ModuleKind` to `Script`, using `await` as an identifier will be OK.

PR #2160, opened 2022-02-23
feat(rome_lsp): format broken code under flag

ematipico:
## Summary

Our parser is able to format partial broken code, but this feature can be misleading for some users. This PR enables this features behind an option. 

This PR adds a new option to our extension called `format_errored_code` to our VSCode extension.

If this option is disabled, the formatter is called only if there aren't errors during the parsing. If the options is enabled, the formatter is called regardless.

## Test Plan

Checked manually that the settings are loaded and the broken code is formatted only when the option is enabled.


editors/vscode/package.json
```diff
@@ -23,14 +23,24 @@
...
 				"rome.lspBin": {
 					"type": [
 						"string",
 						"null"
 					],
 					"default": null,
-					"description": "The rome lsp server executable."
+					"markdownDescription": "The rome lsp server executable."
+				},
+				"rome.formatter.formatErroredCode": {
```
yassere, line 36: Would something like `"rome.formatter.allowSyntaxErrors"` make sense?

yassere (approved):

editors/vscode/package.json
ematipico: I don't know to be honest, the word `allowSyntaxErrors` doesn't really help me to understand that it really does. Maybe `formatCodeWithErrors`/`formatCodeWithSyntaxErrors`??

yassere: `formatWithSyntaxErrors` sounds fine. Or how about `formatInvalidSyntax`?

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -19,25 +20,36 @@ pub(crate) fn to_format_options(params: &FormattingOptions) -> FormatOptions {
...
+    params: &FormattingOptions,
+    workspace_settings: WorkspaceSettings,
+) -> Result<Option<Vec<TextEdit>>> {
+    let syntax = Syntax::default();
 
-    let options = to_format_options(params);
+    let parse_result = parse(text, file_id, syntax);
 
-    let new_text = rome_formatter::format(options, &tree)?.into_code();
+    if workspace_settings.formatter.format_errored_code || !parse_result.has_errors() {
```
leops, line 33: Nit: Doing an early return of `Ok(None)` here would make the diff more readable and avoid right-shifting the entire function

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -48,51 +60,60 @@ pub(crate) struct FormatRangeParams<'input> {
...
-            let start_loc = line_index.line_col(range.start());
-            let end_loc = line_index.line_col(range.end());
-            Range {
-                start: Position {
-                    line: start_loc.line,
-                    character: start_loc.col,
-                },
+pub(crate) fn format_range(
+    params: FormatRangeParams,
+    workspace_settings: WorkspaceSettings,
```
leops, line 65: The reason I originally introduced these params-structs was because the handler function were taking a lot of parameters, so having them explicitly named and laid out in a struct made the calling code more readable. As such it would be better to add the new `workspace_settings` parameter to the structs too instead of adding more elements to the signature of the functions again

crates/rslint_parser/src/parse.rs
```diff
@@ -86,6 +86,11 @@ impl<T> Parse<T> {
     pub fn source_type(&self) -> JsSourceType {
         self.source_type
     }
+
+    /// Returns [true] if the parser encountered some errors during the parsing
+    pub fn has_errors(&self) -> bool {
+        !self.errors.is_empty()
```
leops, line 92: One tricky thing here is that not having any diagnostics doesn't actually guarantee there is no error, it's perfectly possible for the resulting syntax tree to have unknown nodes or missing required children without the parser emitting a diagnostic. I don't see any reason it should happen, so any case it does should be considered a bug, but the point is that checking for the existence of diagnostics may not be entirely reliable

ematipico: > it's perfectly possible for the resulting syntax tree to have unknown nodes 

Shouldn't this be considered a bug? Having an unknown node without a diagnostic doesn't seem "right" to me.

> missing required children without the parser emitting a diagnostic

Is there a scenario where this is accepted? Isn't it a bug if we encounter a scenario like this? If you think that these scenarios are not bugs, we can a better check later.

editors/vscode/package.json
ematipico: I went for `formatWithSyntaxErrors`

crates/rslint_parser/src/parse.rs
leops: I completely agree these are bugs, but at the same time error recovery is tricky: invalid syntax may throw the parser into the wrong path at any turn, so for one way parsing can go right there's a huge number of ways it can go wrong. It seems pretty hard to account for this perfectly, and I've actually encountered both cases of unknown nodes or missing children without diagnostics already

ematipico: What would you suggest to do for the time being?

crates/rslint_parser/src/tests.rs
```diff
@@ -156,7 +156,7 @@ fn assert_errors_are_present(program: &Parse<JsAnyRoot>, path: &Path) {
 }
 
 fn assert_errors_are_absent<T>(program: &Parse<T>, path: &Path) {
-    if program.errors().is_empty() {
+    if !program.has_errors() {
```
ematipico, line 159: @MichaReiser now all our tests will go under this new function. Each test will fail if there are unknown nodes without diagnostics.

crates/rslint_parser/src/parse.rs
```diff
@@ -63,6 +64,34 @@ impl<T> Parse<T> {
...
+
+        for descendent in node.descendants() {
+            if descendent.kind().is_unknown() {
+                has_unknown_nodes = true;
+                break;
+            }
+        }
+
+        has_unknown_nodes
+    }
```
MichaReiser, line 94: I'm not sure if this should be part of the `Parse` interface. The parser should always emit diagnostics. Not emitting diagnostics is an implementation bug so we shouldn't have to pay the overhead of traversing the whole tree (hidden behind an innocent function called `has_errors`). 

I recommend moving this check into our test code. 

Another note. `errors` contains parse errors as well as any other warnings. The behaviour of `errors` should reflect `ok()` which only returns `Err` if there are any `SyntaxError`.

ematipico: >Another note. errors contains parse errors as well as any other warnings. The behaviour of errors should reflect ok() which only returns Err if there are any SyntaxError.


Uhm, this is really misleading, I wasn't aware of it. Thank you for pointing it out.

MichaReiser: Nit: This can be written as

`node.descendants().any(|descendant| descendant.kind().is_unknown())`

leops (approved):

PR #2164, opened 2022-02-23
feat(rome_formatter): implement suppression comments and improve the formatting of verbatim nodes

leops:
## Summary

This PR introduces two complementary changes. The first one is a revision of how whitespace and newlines are printed around verbatim nodes: as the formatter already inserts line breaks and empty lines between statements based on the trivias attached to the nodes, verbatim nodes are now printed with their first leading and last trailing trivias trimmed up to the first comment or token. This brings the printing behavior of verbatim trivias in line with how trivia pieces are printed for non-verbatim nodes. This also means the use of `trim_start` and `trim_end` in `format_list` has become redundant, and since this was the only usage of these functions and their implementation added an additional cost to the maintenance and evolution of `FormatElement` I opted to remove them.

The second revision is the implementation of suppression comments in the formatter, using the same syntax found in Rome Classic (`// rome-ignore formatter: reason`). The parsing of this comments is currently implemented only for the formatter but it should be possible to move this utility to a common crate for it to be reused in other parts of the project. One limitation that's imposed by the current implementation of the suppression detection code is that these comments must be on a separate line (specifically be a leading trivia for the node they want to suppress formatting on), this is to limit the overhead of the detection and make reasoning about the formatting of these nodes easier but we may be able to lift that limitation in the future. As part of this I also removed the `format_node_start` and `format_node_end` placeholders as there doesn't seems to be any use for them at the moment (trivias and sourcemaps are all handled at the token level) and the cost of adding them back if needed should be minimal.

### DIFF
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -6,6 +6,7 @@ use std::iter::once;
 
 use crate::format_element::Verbatim;
 use crate::formatter_traits::FormatTokenAndNode;
+use crate::utils::has_formatter_suppressions;
 use crate::{
     block_indent, concat_elements, empty_element, empty_line,
     format_element::{normalize_newlines, Token, LINE_TERMINATORS},
@@ -87,13 +88,11 @@ impl Formatter {
     }
 
     fn format_syntax_node(&self, node: &SyntaxNode) -> FormatResult<FormatElement> {
-        let start = self.format_node_start(node);
-        let content = node.to_format_element(self)?;
-        Ok(concat_elements(vec![
-            start,
-            content,
-            self.format_node_end(node),
-        ]))
+        if has_formatter_suppressions(node) {
+            return Ok(self.format_suppressed(node));
+        }
+
+        node.to_format_element(self)
     }
 
     /// Formats a group delimited by an opening and closing token,
@@ -202,20 +201,6 @@ impl Formatter {
             },
             close_token,
         )
-    }
-
-    /// Helper function that returns what should be printed before the node that work on
-    /// the non-generic [SyntaxNode] to avoid unrolling the logic for every [AstNode] type.
-    pub(super) fn format_node_start(&self, _node: &SyntaxNode) -> FormatElement {
-        // TODO: Set the marker for the start source map location, ...
-        empty_element()
-    }
-
-    /// Helper function that returns what should be printed after the node that work on
-    /// the non-generic [SyntaxNode] to avoid unrolling the logic for every [AstNode] type.
-    pub(super) fn format_node_end(&self, _node: &SyntaxNode) -> FormatElement {
-        // TODO: Sets the marker for the end source map location, ...
-        empty_element()
     }
 
     /// Print out a `token` from the original source with a different `content`.
@@ -432,6 +417,16 @@ impl Formatter {
     /// doesn't track nodes/tokens as [FormatElement::Verbatim]. They are just printed as they are.
     pub fn format_unknown(&self, node: &SyntaxNode) -> FormatElement {
         self.format_verbatim_node_or_token(node)
+    }
+
+    /// Format a node having formatter suppression comment applied to it
+    pub fn format_suppressed(&self, node: &SyntaxNode) -> FormatElement {
+        format_elements![
+            // Insert a force a line break to ensure the suppression comment is on its own line
+            // and correctly registers as a leading trivia on the opening token of this node
+            hard_line_break(),
+            self.format_verbatim_node_or_token(node),
+        ]
     }
 
     fn format_verbatim_node_or_token(&self, node: &SyntaxNode) -> FormatElement {
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -2,6 +2,7 @@
 use std::cell::RefCell;
 #[cfg(debug_assertions)]
 use std::collections::HashSet;
+use std::iter::once;
 
 use crate::format_element::Verbatim;
 use crate::formatter_traits::FormatTokenAndNode;
@@ -13,7 +14,8 @@ use crate::{
     soft_line_break_or_space, space_token, FormatElement, FormatOptions, FormatResult,
     ToFormatElement,
 };
-use rome_rowan::SyntaxElement;
+use rome_rowan::api::SyntaxTriviaPiece;
+use rome_rowan::Language;
 #[cfg(debug_assertions)]
 use rslint_parser::SyntaxNodeExt;
 use rslint_parser::{AstNode, AstNodeList, AstSeparatedList, SyntaxNode, SyntaxToken};
@@ -333,8 +335,6 @@ impl Formatter {
                     // Lists that yield errors are formatted as they were unknown nodes.
                     // Doing so, the formatter formats the nodes/tokens as is.
                     self.format_unknown(module_item.syntax())
-                        .trim_start()
-                        .trim_end()
                 }
             };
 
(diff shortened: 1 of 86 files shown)


leops:
!bench_formatter

MichaReiser:
> This PR introduces two complementary changes. The first one is a revision of how whitespace and newlines are printed around verbatim nodes: as the formatter already inserts line breaks and empty lines between statements based on the trivias attached to the nodes, verbatim nodes are now printed with their first leading and last trailing trivias trimmed up to the first comment or token. This brings the printing behavior of verbatim trivias in line with how trivia pieces are printed for non-verbatim nodes. This also means the use of trim_start and trim_end in format_list has become redundant, and since this was the only usage of these functions and their implementation added an additional cost to the maintenance and evolution of FormatElement I opted to remove them.

I love it! Nice work.

crates/rome_formatter/src/formatter.rs
```diff
@@ -434,29 +419,64 @@ impl Formatter {
...
+            if #[cfg(debug_assertions)] {
+                for token in node.tokens() {
+                    assert!(self.printed_tokens.borrow_mut().insert(token.clone()));
                 }
             }
-        }))
+        }
+
+        fn skip_whitespace<L: Language>(piece: &SyntaxTriviaPiece<L>) -> bool {
+            piece.as_newline().is_some() || piece.as_whitespace().is_some()
```
MichaReiser, line 442: Nit: Would it make sense to add a `piece.is_whitespace()` method? Seems to be a concept that may be useful in other sue cases too.

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -104,3 +104,209 @@ pub(crate) fn format_head_body_statement(
...
+
+fn parse_suppression_comment(comment: &str) -> impl Iterator<Item = Suppression> {
+    let (head, mut comment) = comment.split_at(2);
+    let is_block_comment = match head {
+        "//" => false,
+        "/*" => {
+            comment = comment.strip_suffix("*/").unwrap();
+            true
+        }
+        _ => panic!(),
```
MichaReiser, line 122: Nit: maybe some nice message what's happening here?

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -104,3 +104,209 @@ pub(crate) fn format_head_body_statement(
...
+        Some(Suppression { categories, reason })
+    })
+}
+
+pub(crate) fn has_formatter_suppressions(node: &SyntaxNode) -> bool {
+    // Lists cannot have a suppression comment attached, it must
+    // belong to either the entire parent node or one of the children
+    match node.kind() {
+        // Files are a special kind of list
+        JsSyntaxKind::JS_MODULE | JsSyntaxKind::JS_SCRIPT => return false,
```
MichaReiser, line 188: You want to add `JS_EXPRESSION_SNIPPED` to. Ideally, you can use `JsAnyRoot::can_cast(node.kind())`

MichaReiser (approved):

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -104,3 +104,209 @@ pub(crate) fn format_head_body_statement(
...
+    };
+
+    first_token
+        .leading_trivia()
+        .pieces()
+        .filter_map(|trivia| trivia.as_comments())
+        .any(|comment| {
+            for suppression in parse_suppression_comment(comment.text()) {
+                for (category, _) in suppression.categories {
+                    if category == "format" {
```
ematipico, line 205: We could have this `"format"` inside a `const`

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -104,3 +104,209 @@ pub(crate) fn format_head_body_statement(
...
+                reason: "explanation"
+            }],
+        );
+
+        assert_eq!(
+            parse_suppression_comment(
+                "/**
+				  * hello
+				  * rome-ignore parse(wow) parse(fish): explanation
+				  */"
```
ematipico, line 303: 1. This file seems to have tabs
2. Should we test against multiple categories? e.g. `// rome-ignore lint format: explanation`

ematipico (changes requested):

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -104,3 +104,209 @@ pub(crate) fn format_head_body_statement(
...
+
+        // TODO: While we do want to detect these, it should
+        // only be enabled in a special "migration" mode where
+        // they get rewritten as rome-ignore
+        if line.trim_end() == "prettier-ignore" {
+            return Some(Suppression {
+                categories: vec![("format", None)],
+                reason: "",
+            });
+        }
```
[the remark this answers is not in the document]
ematipico: I mean, it's a nice thing that you thought about! I haven't honestly.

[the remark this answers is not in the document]
MichaReiser: I also spent some more time thinking about this and agree. There might not even be reason to keep them because the main reason I added them in the past was when prettier screwed up and e.g. dropped some tokens.

crates/rome_formatter/src/utils/mod.rs
leops: This was actually supposed to be `unreachable!()` instead of `panic!()` but I forgot to change when cleaning up the parsing code. The reasoning behind this is simply that its not currently possible for a comment to not start with `/*` or `//`

crates/rome_formatter/src/utils/mod.rs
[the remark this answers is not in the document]
leops: Honestly I implemented this bit mostly to be able to process the prettier tests for ignore comments without modifying them, and was considering adding `cfg!(test) &&` to the condition.

In my own experience I mostly used prettier-ignore comments when I want a node formatted in a specific way (the matrix I added in the test cases is real code), so from that perspective it makes sense to preserve those comments. One thing I had in mind for the migration would have been to check that the output of the formatter with and without the suppression is actually different, and drop the comment (with a warning) if it isn't

crates/rome_formatter/src/utils/mod.rs
leops: Ah yes these are test cases I ported straight over from Rome Classic, since these are string literals the tabs remained. The `parse_multiple_suppression` test does have categories but not without values, all add an additional test for that

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -104,3 +104,209 @@ pub(crate) fn format_head_body_statement(
         ])
     }
 }
+
+#[derive(Debug, PartialEq, Eq)]
+struct Suppression<'a> {
+    categories: Vec<(&'a str, Option<&'a str>)>,
```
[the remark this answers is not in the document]
leops: There definitely should be some amount of validation around these comments, not only should it emit a diagnostic for categories or values that don't make sense but I think it would also make sense for these to behave like `@ts-expect-error` by default and warn if they are not actually suppressing anything (formatter suppression where the output would be the same without the comment, or lint suppression where the associated lint isn't raising any diagnostic for the associated node). However for now I'm not really sure where these diagnostics would be emitted (at least it seems awkward for the formatter to emit diagnostics) nor what the actual category and values will end up being hence why I'm currently parsing everything as strings instead of narrower enums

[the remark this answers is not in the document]
ematipico: All right then, surely this will have a second iteration!

crates/rome_formatter/src/utils/mod.rs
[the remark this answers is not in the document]
leops: I ended up removing the `prettier-ignore` case for now and implemented the replacement logic in the prettier test suite instead which seems to work well enough (except for `js/ignore/ignore-2.js` but that's because JSX is confusing the parser and the comment ends up being parsed as a regexp instead)

ematipico (approved):

PR #2166, opened 2022-02-24
fix(rslint_parser): Allow private name only in `in` expessions

MichaReiser:
## Summary

#2163 fixed the precedence of private names in `in` expressions, but it doesn't catch the case where a private name is used as the left-hand side of any binary expression other than `in`.

This PR adds a check to verify that the operator is the `in` operator

This PR also fixes an issue with the TypeScript test runner. TypeScript supports running a test for different targets. The name of the `errors` file is then called `<test-case>(target=<target>).errors.txt` and a test is expected to fail if it is present. Our runner doesn't support running tests with different targets but we should assert that a test is expected to fail if an `.errors` file exists for any target (same as we currently do for `module`). This isn't perfect but works reasonably well. 

## Test Plan

Added a new parser test

The regressions come from tests that use multiple targets and the test runner so far expected that they should pass because it didn't test for the presence of `<test-case>(target=target).errors.txt` file.

### DIFF
--- a/crates/rslint_parser/src/syntax/expr.rs
+++ b/crates/rslint_parser/src/syntax/expr.rs
@@ -19,7 +19,8 @@ use crate::syntax::function::{
 use crate::syntax::js_parse_error;
 use crate::syntax::js_parse_error::{
     expected_expression, expected_identifier, expected_parameters,
-    expected_simple_assignment_target, expected_ts_type, ts_only_syntax_error,
+    expected_simple_assignment_target, expected_ts_type,
+    private_names_only_allowed_on_left_side_of_in_expression, ts_only_syntax_error,
 };
 use crate::syntax::object::parse_object_expression;
 use crate::syntax::stmt::{is_semi, STMT_RECOVERY_SET};
@@ -420,7 +421,7 @@ fn parse_binary_or_logical_expression_recursive(
         let op_tok = p.cur_tok();
 
         let mut is_unknown = false;
-        if let Present(left) = &left {
+        if let Present(mut left) = &left {
             // test exponent_unary_parenthesized
             // (delete a.b) ** 2;
             // (void ident) ** 2;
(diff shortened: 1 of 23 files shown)


xtask/coverage/src/typescript.rs
```diff
@@ -113,12 +110,12 @@ fn extract_metadata(code: &str, path: &str) -> TestCaseMetadata {
 
             if option_name == "alwaysstrict" {
                 current_file_content.push_str(&format!("\"use strict\";{}", line_ending))
-            } else if option_name == "module" && files.is_empty() {
-                module_options.extend(
+            } else if matches!(option_name.as_str(), "module" | "target") && files.is_empty() {
```
MichaReiser, line 113: TypeScript tests with multiple `target` options have a different `.errors` file for each `traget` and this is something our test runner didn't support yet but is needed to make `privateNameInInExpressionTransform` pass.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -420,7 +421,7 @@ fn parse_binary_or_logical_expression_recursive(
         let op_tok = p.cur_tok();
 
         let mut is_unknown = false;
-        if let Present(left) = &left {
+        if let Present(mut left) = &left {
```
leops, line 424: I don't think this is doing what it should be doing. Binding a mutable borrow on the left-hand side from an immutable borrow on the right-hand side shouldn't be possible, but because `CompletedMarker` implements `Copy` it's actually creating a mutable clone of the original marker and modifying that. It mostly works because the new `kind` of the `left` marker doesn't get read, and `change_kind` does correctly edit the event stream in the parser so the output is right anyway, but it seems error-prone

MichaReiser: Uh, nice catch. Why the heck does `CompletedMarker` implement `Copy` (or even `Clone`...). There so many things that can go wrong if one starts cloning these...

ematipico (approved):
Looks good, after we fix what Leo pointed out

MichaReiser:
!bench_parser

PR #2170, opened 2022-02-24
fix(rslint_parser): Fix spread call expression arguments

### LINKED ISSUE #2169: 🐛 Spread elements are not allowed in call expressions
### Environment information

```bash
N/A
```

### What happened?

The expression `a(...b)` is parsed as an unknown expression with no diagnostics

### Expected result

The current grammar only allow for a list of `JsAnyExpression` in function calls arguments, I'm guessing it should have an optional JsSpread (or an entirely new node) after the expressions list

_No response_

### Code of Conduct

MichaReiser:
## Summary

Call expressions can contain any number of spread arguments:

```
let a, b, c, d;
function test(...args) { console.log(args); }

test(...a, ...b, c, ...d);
```

The parser already parse these correctly but the grammar didn't allow `JsSpread` nodes in the argument position.

This PR introduces a new `JsAnyCallArgument` that is either a `JsAnyExpression` or a `JsSpread`.

Fixes #2169

## Test Plan

Added new parser tests

### DIFF
--- a/crates/rome_formatter/tests/specs/prettier/js/comments/issues.js.snap
+++ b/crates/rome_formatter/tests/specs/prettier/js/comments/issues.js.snap
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/prettier_tests.rs
-assertion_line: 121
+assertion_line: 119
 expression: issues.js
 
 ---
@@ -96,8 +96,7 @@ throw new ProcessSystemError(
 
 // Missing one level of indentation because of the comment
 const rootEpic = (actions, store) =>
-  (
-  combineEpics(...epics)(actions, store)// Log errors and continue.
+  (combineEpics(...epics)(actions, store)// Log errors and continue.
   .catch(
     (err, stream) => {
       getLogger().error(err);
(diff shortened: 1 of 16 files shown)


crates/rome_formatter/tests/specs/prettier/js/comments/issues.js.snap
```diff
@@ -96,8 +96,7 @@ throw new ProcessSystemError(
 
 // Missing one level of indentation because of the comment
 const rootEpic = (actions, store) =>
-  (
-  combineEpics(...epics)(actions, store)// Log errors and continue.
+  (combineEpics(...epics)(actions, store)// Log errors and continue.
```
MichaReiser, line 99: @leops should we ensure that there's single whitespace before comments?

ematipico (approved):

crates/rome_formatter/tests/specs/prettier/js/comments/issues.js.snap
leops: That's normally the case for comments in trailing position, but this comment was originally a leading trivia on the next token so I think what's actually missing here is a line break. I'll track this separately

leops (approved):

PR #1938, opened 2022-01-13
validate unicode escape on string literals

xunilrj:
see https://github.com/rome/tools/issues/1909#issuecomment-1009414160:

In this case, I think our messages are much more meaningful than VSCode's:

VSCode :
image

Ours:

```

warning: invalid digits after unicode escape sequence
   ┌─ xtask/src/coverage/test262/test/language/module-code/early-export-ill-formed-string.js:21:18
   │
21 │ export {Moon as "\uD83C",} from "./early-export-ill-formed-string.js";
   │                  ------ expected valid unicode escape sequence

warning: invalid digits after unicode escape sequence
   ┌─ xtask/src/coverage/test262/test/language/module-code/early-export-ill-formed-string.js:22:10
   │
22 │ export {"\uD83C"} from "./early-export-ill-formed-string.js";
   │          ------ expected valid unicode escape sequence

warning: invalid digits after unicode escape sequence
   ┌─ xtask/src/coverage/test262/test/language/module-code/early-export-ill-formed-string.js:23:10
   │
23 │ import {'\uD83C' as Usagi} from "./early-export-ill-formed-string.js";
   │          ------ expected valid unicode escape sequence

error[SyntaxError]: expected a valid string literal, or a valid identifier but instead found '"\uD83C"'
   ┌─ xtask/src/coverage/test262/test/language/module-code/early-export-ill-formed-string.js:21:17
   │
21 │ export {Moon as "\uD83C",} from "./early-export-ill-formed-string.js";
   │                 ^^^^^^^^ Expected a valid string literal, or a valid identifier here

error[SyntaxError]: expected a valid string literal, or a valid identifier but instead found '"\uD83C"'
   ┌─ xtask/src/coverage/test262/test/language/module-code/early-export-ill-formed-string.js:22:9
   │
22 │ export {"\uD83C"} from "./early-export-ill-formed-string.js";
   │         ^^^^^^^^ Expected a valid string literal, or a valid identifier here

error[SyntaxError]: expected an identifier but instead found ''\uD83C' as Usagi'
   ┌─ xtask/src/coverage/test262/test/language/module-code/early-export-ill-formed-string.js:23:9
   │
23 │ import {'\uD83C' as Usagi} from "./early-export-ill-formed-string.js";
   │         ^^^^^^^^^^^^^^^^^ Expected an identifier here
```

The warnings come from the lexer; the errors from the parser.

The rast is 

```
0: JS_MODULE@0..758
  0: (empty)
[...]

PR #2171, opened 2022-02-24
chore(coverage): assert no unknown in coverage

MichaReiser:
## Summary

Extends the coverage tests to assert that the parsed result contains no `Unknown` nodes for tests that are expected to pass. 

This revealed that our Unicode escape sequence parsing is too strict. The validation if Unicode escape sequences are valid (form a valid character) is only necessary in export-name positions (see T262 regressions) but not any others. This check was added in #1938 and is reverted in this PR because I believe it's more important that we parse valid files correctly. We can correctly implement the check once the work on #2035 is done. 

This PR further fixes the parsing of `getter` object members to parse out a normal type annotation and not a return type annotation.

## Test Plan

Verified that the regressions are specific to the revert of the Unicode verification.

### DIFF
--- a/crates/rslint_lexer/src/errors.rs
+++ b/crates/rslint_lexer/src/errors.rs
@@ -5,7 +5,7 @@ pub fn invalid_digits_after_unicode_escape_sequence(
     start: usize,
     end: usize,
 ) -> Box<Diagnostic> {
-    let err = Diagnostic::warning(file_id, "", "invalid digits after unicode escape sequence")
+    let err = Diagnostic::error(file_id, "", "invalid digits after unicode escape sequence")
         .primary(start..end, "expected valid unicode escape sequence");
     Box::new(err)
 }
--- a/xtask/coverage/src/test262.rs
+++ b/xtask/coverage/src/test262.rs
@@ -1,7 +1,8 @@
-use crate::runner::{TestCase, TestCaseFiles, TestRunOutcome, TestSuite};
+use crate::runner::{
+    create_unknown_node_in_tree_diagnostic, TestCase, TestCaseFiles, TestRunOutcome, TestSuite,
+};
 use regex::Regex;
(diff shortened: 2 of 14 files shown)


crates/rslint_lexer/src/errors.rs
```diff
@@ -5,7 +5,7 @@ pub fn invalid_digits_after_unicode_escape_sequence(
     start: usize,
     end: usize,
 ) -> Box<Diagnostic> {
-    let err = Diagnostic::warning(file_id, "", "invalid digits after unicode escape sequence")
+    let err = Diagnostic::error(file_id, "", "invalid digits after unicode escape sequence")
```
MichaReiser, line 8: I changed this to an `error` because the `Lexer` pushes an `ERROR_TOKEN` when it encounters an invalid unicode escape character which, in turn, invalidates any parent node (makes it an unknown node). It's important to have an error diagnostic in that case.

xtask/coverage/src/test262.rs
```diff
@@ -85,19 +87,36 @@ impl Test262TestCase {
...
+                if let Some(unknown) = root
+                    .syntax()
+                    .descendants()
+                    .find(|descendant| descendant.kind().is_unknown())
+                {
+                    TestRunOutcome::IncorrectlyErrored {
+                        errors: vec![Diagnostic::new(
+                            0,
+                            Severity::Bug,
+                            "Unknown node in test that should pass",
```
ematipico, line 111: What if we forgot to add diagnostics to unknown nodes? Is it also the case here? Should we rephrase the error message?

ematipico (approved):

xtask/coverage/src/test262.rs
MichaReiser: Good point. I can re-phrase it. It really is that it's a test without any diagnostic but with unknown nodes

PR #2173, opened 2022-02-24
feat(rome_formatter): format binaryish expressions

### LINKED ISSUE #2156: 📎 Format binary and logical expressions expressions
### Description

At the moment we don't do anything for format binary and logical expressions, we should implement an algorithm to format them.

EDIT: also `TsIntersectionType` might benefit from the same logic

ematipico:
## Summary

Closes #2156 

The main idea was inspired from prettier, but the whole implementation is actually me,  because I needed to come up with a decent way to adapt the end result to our formatting strategy.

The main function has a big documentation doc that contains the explanation of what the algorithm tries to achieve. Again, it's a game between recursion and splitting parts between nodes.

## Test Plan

 I used the prettier test suite to check if the end result is what we want. Some tests are a bit off but the majority of those belong to broken code.

### DIFF
--- a/crates/rome_formatter/src/utils/binarish_expression.rs
+++ b/crates/rome_formatter/src/utils/binarish_expression.rs
@@ -1,0 +1,360 @@
+use crate::formatter_traits::FormatTokenAndNode;
+use crate::{
+    concat_elements, empty_element, format_elements, group_elements, hard_line_break,
+    if_group_breaks, if_group_fits_on_single_line, indent, join_elements, soft_line_break,
+    soft_line_break_or_space, space_token, FormatElement, FormatResult, Formatter, ToFormatElement,
+};
+use rslint_parser::ast::{
+    JsAnyExpression, JsBinaryExpression, JsBinaryExpressionFields, JsLogicalExpression,
+    JsLogicalExpressionFields,
+};
+use rslint_parser::{AstNode, JsSyntaxKind, SyntaxNode};
+use std::fmt::Debug;
+
+/// This function is charge to flat binaryish expressions that have the same precedence of their operators
+///
+/// This means that expressions like `some && thing && elsewhere` are entitled to fall in the same group.
+///
+/// Instead, if we encounter something like `some && thing  || elsewhere && thing`, we will great two groups:
+/// `[some, thing]` and `[elsewhere, thing]`, each group will be grouped altogether.
+///
+///
+/// Let's take for example:
+///
+/// ```ignore
+/// some && thing && elsewhere && happy
+/// ```
+///
+/// These expressions have a nested  nodes, which is roughly something like this:
+///  
+/// ```block
+/// JsLogicalExpression {
+///     left: JsLogicalExpression {
+///         left: JsLogicalExpression {
+///             left: "some"
+///             operator: "&&",
+///             right: "thing"
+///         }
+///         operator: "&&"
+///         right: "elsewhere"
+///     }
+///     operator: "&&"
+///     right: "happy"
+/// }
+/// ```
+///
+/// Our final result should be something like this:
+/// ```ignore
+/// some &&
+/// thing &&
+/// elsewhere &&
+/// happy
+/// ```
+///
+/// So what we are going to do here is:
+/// - create a vector of flatten items, where the most nested node is the first one,`left: "some"` in our
+/// example. The last one will be the first that we encounter, in this case the node that contains `right: "happy"`
+/// - each element of the vector will contain two elements. One is the AST node, the other one is its
+/// formatted version
+/// - the formatted elements will be grouped
+///
+///
+/// The flattening of the groups is done with recursion, during the recursions we want to be careful of:
+/// - not printing nodes/token twice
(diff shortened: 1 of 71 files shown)


crates/rslint_parser/src/util.rs
```diff
@@ -160,7 +160,17 @@ pub trait SyntaxNodeExt {
     fn contains_comments(&self) -> bool {
         self.tokens()
             .iter()
-            .any(|tok| tok.has_leading_comments() || tok.has_trailing_comments())
+            .any(|tok| tok.has_trailing_comments() || tok.has_leading_comments())
+    }
+
+    /// Whether the node contains trailing comments.
+    fn has_trailing_comments(&self) -> bool {
```
leops, line 167: I'm not entirely certain these two methods are actually doing what they should be doing, or at least the name of the methods can be misleading: what they actually do is iterate on all the tokens of a node (including those of its descendant nodes) and check whether those tokens have any trailing or leading comment, but based on the method name you could expect them to only check whether the last token of the node has any trailing comment, or the first token has any leading comments

crates/rome_formatter/src/utils/binarish_expression.rs
```diff
@@ -0,0 +1,662 @@
...
+/// `&& happy`. If we follow the logic explained so far, this node doesn't have an operator
+/// anymore because we passed it to its child. And we can't try to add a new operator.
+/// But this is fine! Because this is want we wanted! By removing the operator, we are left with `happy`
+/// which is what we wanted since the beginning!
+///
+pub fn format_binaryish_expression(
+    syntax_node: &SyntaxNode,
+    formatter: &Formatter,
+) -> FormatResult<FormatElement> {
+    let mut flatten_nodes = FlattenItems::new(syntax_node.to_owned());
```
leops, line 102: Nit: `.clone()` is generally used rather than `.to_owned()` (and is shorter)

crates/rome_formatter/src/utils/binarish_expression.rs
```diff
@@ -0,0 +1,662 @@
...
+            previous_operator.has_trailing_comments() || right.syntax().contains_comments(),
+        )
+    } else {
+        (
+            empty_element(),
+            // Here we want to check only leading comments;
+            // trailing comments will be added after the end of the whole expression.
+            // We want to handle cases like `lorem && ipsum // comment`
+            // This part is a signal to the formatter to tell it if the whole expression should break.
+            right.syntax().has_leading_comments(),
```
leops, line 355: Based on my previous comment about `has_leading_comments` and `has_trailing_comments`, I'm not entirely certain this bit is doing what it should be doing either. For instance `has_comments` would be set to true if `right` contained the following call expression:
```js
func({
    // leading comment
    key: 'value',
})
```
Is this what's actually intended ?

crates/rome_formatter/src/utils/binarish_expression.rs
```diff
@@ -0,0 +1,662 @@
...
+            )));
+        }
+
+        let is_inside_parenthesis = is_inside_parenthesis(&self.current_node);
+        let should_not_indent = should_not_indent_if_parent_indents(&self.current_node);
+        let should_ident_if_parent_inlines = should_indent_if_parent_inlines(&self.current_node);
+
+        if is_inside_parenthesis {
+            Ok(format_elements![
+                if_group_breaks(group_elements(format_elements![
```
leops, line 490: The expressions `if_group_breaks(group_elements(hard_line_break()` and `if_group_fits_on_single_line(hard_group_elements(` seem redundant, the first will create a new group that always break if the parent group is already breaking, and the second will create a new group that never breaks if the parent is already non-breaking

MichaReiser:
@ematipico what's the status of this PR. Is it on hold?

ematipico:
> @ematipico what's the status of this PR. Is it on hold?

It's been ready for code review

crates/rslint_parser/src/util.rs
ematipico: True! I updated the code that the tests too, they were misleading!

crates/rome_formatter/src/utils/binarish_expression.rs
ematipico: No, and I will fix the implementation of `has_leading_comments` and `has_trailing_comments`. I will also add a better test to make sure it does what's intended

crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/spec_test.rs
-assertion_line: 163
+assertion_line: 168
```
leops, line 3: Nit: these changes could be discarded

leops (approved):

crates/rome_formatter/tests/specs/js/module/statement/if_else.js.snap
ematipico: Done

Issue #2120, opened 2022-02-16
📎 Serialization for AST/CST


NicholasLYang:
### Description

This will be very useful for the playground and for inspecting trees in general.

MichaReiser:
There's an existing string serialization format but that's hard to inspect. I believe it could help to be more explicit about what the goal is because it may then differ significantly how we approach this. 

* Serializing the whole AST can be slow for large files and may, therefore, not be desired
* It may be sufficient to instead expose the metadata for each node kind. Kind of a reflection (readonly) API for `SyntaxNode`s.  
* Have you looked into if we could integrate Rome into [astexplorer](https://astexplorer.net)

PR #2176, opened 2022-02-24
feat(rslint_parser): check for missing children in parser tests

leops:
## Summary

This adds a `cfg(test)`-only `missing_children` method to AstNode returning an iterator over SyntaxError. It traverses the syntax tree an accumulate errors for each missing required children, this method is then called in the parser tests and the missing children are reported similarly to diagnostics.


xtask/codegen/src/generate_nodes.rs
```diff
@@ -142,10 +154,41 @@ pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
                             quote! { #method_name: SyntaxResult<#ty> }
                         };
 
-                        (field, quote! { #method_name: self.#method_name() })
+                        let missing_children = if is_list {
+                            quote! {
+                                .chain(
+                                    self.#method_name().missing_children()
```
MichaReiser, line 160: I'm having a hard time getting my head around.  Does this correctly handle separated lists?

MichaReiser (commented):
I'm having somewhat conflicting feelings about this change. 

It does achieve the goal but it results in a lot of additional code and start wondering if it instead would be useful to expose the metadata for each node in a different form. How many fields does a `JS_MODULE` have, which fields are mandatory or optional? Is it a separated list? 

Something like this could also be useful when thinking about serializing/presenting nodes #2120 

An alternative approach to this which I once planned to do but never came along is to add a `contains_syntax_errors` flags to `SyntaxNode'`s and that flag gets set to `true` inside of the `JsSyntaxFactory` if any child node has that flag or if any required child is missing. 

This wouldn't just be useful here but also in other places where we want to assert that a tree is valid. Ideally, adding a diagnostic would also toggle that flag but that will require some changes to the way we emit diagnostics today.

crates/rslint_parser/src/tests.rs
```diff
@@ -176,9 +179,15 @@ fn assert_errors_are_absent<T>(program: &Parse<T>, path: &Path) {
         emitter.emit_with_writer(diagnostic, &mut buffer).unwrap();
     }
 
+    let buffer = buffer.into_inner();
+    let mut buffer = String::from_utf8(buffer).unwrap();
+    for syntax_error in syntax_errors {
+        writeln!(buffer, "{}", syntax_error).expect("writing to a String cannot fail");
+    }
+
```
MichaReiser, line 187: It would be neat for debuggability if this would emit `Diagnostic`'s too instead of writing to a string. But I can see why this is difficult. 

A cheap way of implementing the same is to print the AST as a string and then assert that it doesn't contain `missing (required)`.

MichaReiser:
@leops is it OK if we close this PR for now?

leops:
Oh I completely forgot about this one. In the most recent version of this PR it just checked whether there was any "missing (required)" in the debug representation of the tree, so I think it worked as expected without adding too much new code but if I remember correctly I didn't move it out of draft at the time because TS support was incomplete at the time and the test was failing then. Now that's it's stable (and in preparation for the future other languages we want to add) maybe I could rebase it on the latest main ?

MichaReiser:
Oh, didn't notice you updated it. Yeah, that looks useful. Let's rebase it and see if all tests pass.

MichaReiser (approved):
Awesome, all tests are passing :) This will be helpful when working on the JSON parser

PR #2178, opened 2022-02-25
feat(rome_formatter): improve the formatting of call expressions

leops:
## Summary

This PR modifies the breaking logic of call expression chains by relying on the printer's existing breaking behavior, with special cases to disallow breaking when the chain only has a single call expression or to force breaking if the chain contains any hard line break

It also adds additional logic to the printing of call arguments to not create a new group when there are no comments over the whole arguments groups and a single argument that's either an array expression, object expression or "simple" function expression (less than 3 identifier parameters, no type parameters or return type annotations)


crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -203,9 +211,10 @@ fn compute_first_group_index(flatten_items: &[FlattenItem]) -> usize {
                 Some(index)
             }
         })
-        // The first group has always at least one item. If no more items have been skipped,
-        // and we arrive here, we can say that the index where want to split the group is 1
-        .unwrap_or(1)
+        // If the above returns None this means either all items were skipped
+        // or the list was empty. In either case, this means the first group
+        // covers the entire list of FlattenItems
```
ematipico, line 216:
```suggestion
        // covers the entire list of [FlattenItem]
```

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -415,10 +438,26 @@ impl Groups {
...
+            return false;
+        }
+
+        // Break the groups if any hard line break is found
+        let has_line_breaks = self
+            .groups
+            .iter()
+            .flat_map(|group| group.iter())
+            .flat_map(|item| item.as_format_elements())
+            .any(|element| element.has_hard_line_breaks());
```
ematipico, line 453: Could you explain this bit by adding some comment to the example, please? The reason why we should expand this bit because when we say "hard line break", we don't actually refer to the source code. We talk about internal implementation of our printer. Hence, I think we should explain in details what cases we are covering.

crates/rome_formatter/src/utils/simple.rs
```diff
@@ -0,0 +1,184 @@
...
+    } else if let Some(object) = JsObjectExpression::cast(node.syntax().clone()) {
+        is_simple_object_expression(object)
+    } else {
+        Ok(false)
+    }
+}
+
+/// Returns true if the passed [JsAnyFunction] does not have any comment, type
+/// parameters, return type annotation and simple parameters (see [is_simple_function_parameters])
+fn is_simple_function_expression(func: JsAnyFunction) -> SyntaxResult<bool> {
```
ematipico, line 24: Can we explain the concept of "simple"? To us, internally, this simple has a meaning now, then we should explain the meaning. What are we trying to achieve with this concept? I think it's worth a longer explanation. e.g. I read the documentation I don't understand the "threshold" between "simple" and "not simple"

crates/rome_formatter/src/utils/simple.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    if l_curly_token?.has_leading_comments() || r_curly_token?.has_trailing_comments() {
+        return Ok(false);
+    }
+
+    Ok(true)
+}
+
+/// Returns true if the passed [JsParameters] has 2 or less parameters that are
+/// all simple parameters (see [is_simple_parameter]) with no comments trivia
```
ematipico, line 116: Same here, we should explain what's the concept of "simple" and what we want to try to achieve with that. Examples might help to understand that.

crates/rome_formatter/src/utils/simple.rs
```diff
@@ -0,0 +1,184 @@
...
+            }
+            None => return Ok(false),
+        }
+    }
+
+    Ok(true)
+}
+
+/// Returns true if the passed [JsFormalParameter] is a single identifier
+/// with no question mark token, type annotation or initializer
```
ematipico, line 147: Same here too. We should explain what's a simple parameter for us and what we are trying to achieve

ematipico (commented):
It seems good to me. I would improve a bit the documentation of some new concepts and, if possible, I would add more tests (our own) and not do not rely only on the prettier spec (which is more difficult to compare/understand).

crates/rome_formatter/src/utils/call_expression.rs
leops: What this bit of code does is emulate a simplified version of the similar logic found in the printer to force groups to break if they contain any "hard line break": these not only include `hard_line_break` elements but also `empty_line` or tokens containing the `\n` character. The idea is that since any of these will force the group to break when it gets printed, the formatter needs to emit a group element for the call chain in the first place or will not be printed correctly.

One thing to note is that in the context of printing an expression, hard line breaks emitted by the formatter are often the result of such line breaks having been originally present in the input source, for instance after a single line comment. So while this bit of code aims to be generic over any kind of line break, in practice the main problem it aims to solve is improving the printing of comments in call chains

crates/rome_formatter/src/utils/simple.rs
leops: The issue with the name "simple" is that I originally picked to cover one specific case but its definition actually drifted quite a bit while I was writing this PR: the actual definition of "simple" expressions would be expressions that are going to create a single group anyway, so they don't need to be wrapped in a second one: this includes object, array or parenthesized expressions, as well as function declarations that have a high probability of breaking only in their body group. This last bit is defined recursively as functions that only have a few (less than 3) identifier parameters, no type parameter or return type and a block body: technically such a function expression can still break in both the parameters and body group but the small number of parameters makes it unlikely.

The use case for detecting these "simple" expressions is to avoid creating redundant groups in nested delimited expressions when only one would suffice, for instance in call expressions:
```js
// Formatter result before this PR
new Promise(
  (resolve, reject) => {
    resolve();
  }
);

func(
  {
    key: 'value',
  }
);

// Formatter result after this PR
new Promise((resolve, reject) => {
  resolve();
});

func({
  key: 'value',
});
```

ematipico: Thank you! I think it's worth addition your reply to the source code as comment. This will help people to understand what we are trying to achieve. In case a tomorrow I will have to check this code a again, I will know what to do!

crates/rome_formatter/src/utils/simple.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    match func.body()? {
+        JsAnyFunctionBody::JsFunctionBody(body) => {
+            let JsFunctionBodyFields {
+                l_curly_token,
+                directives: _,
+                statements: _,
+                r_curly_token,
+            } = body.as_fields();
```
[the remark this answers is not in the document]
leops: For node destructuring I think it's preferable to manually ignore all the unused fields instead of using a rest token as any future modification of the `JsFunctionBody` node that adds or remove a field will also need to modify this pattern, if only to add another ignored field: in turn this change will show up in the PR diff and "flag" this function for review to check that the code still makes sense after the AST / grammar has been changed

ematipico (approved):
Thank you @leops  for adding more documentation/comments to the code base! I come from an old code base where comments were scarce, so it was really difficult to apply changes or having context without spending hours studying the code. We want to avoid this by adding as much comments as possible. This is really appreciated!

PR #2179, opened 2022-02-28
feat(rome_formatter): implement basic formatting for all the remaining verbatim nodes

leops:
## Summary

This implements an actual formatting logic for all the remaining nodes using `format_verbatim` (for now). For most of these nodes the formatting logic is trivial, the only exception being type unions and intersections as the formatting convention for these is to place the separators in a leading position instead of trailing when the group breaks


crates/rome_formatter/tests/specs/prettier/typescript/intersection/intersection-parens.ts.snap
```diff
@@ -60,43 +60,43 @@ type C = (((number | string))) & boolean;
...
+
+type B2 = C;
+type B3 = (C);
+type B4 = (C);
+type B5 = ((C));
+type B6 = C; /*1*/
+type B7 = (C); /*1*/
+type B8 = (C); /*1*/
+type B9 = ( /*1*/ C);
+type B10 = C; /*1*/
```
ematipico, line 86: Careful here, in `type B10` we lost `/*2*/` comment

crates/rome_formatter/tests/specs/prettier/typescript/prettier-ignore/mapped-types.ts.snap
```diff
@@ -81,60 +81,62 @@ type a= {
...
   [A in // prettier-ignore
-  B ]: C  |  D;
+  B ]: C | D;
 };
 
 type a = {
-  [A in B ]: 
-      // prettier-ignore
-      C  |  D;
+  [A in B ]:
```
ematipico, line 98: Small bug I just saw. We leave the space after `B` and we should not. Unrelated to this PR

crates/rome_formatter/tests/specs/prettier/typescript/union/union-parens.ts.snap
```diff
@@ -153,87 +151,83 @@ class Foo<T extends (string | number)> {}
...
+type A2 = C;
+type A3 = (C);
+type A4 = (C);
+type A5 = ((C));
+type A6 = C; /*1*/
+type A7 = (C); /*1*/
+type A8 = (C); /*1*/
+type A9 = ( /*1*/ C);
+type A10 = C; /*1*/
+type A11 = (C); /*1*/
```
ematipico, line 215: `A11` loses a comment

crates/rome_formatter/src/ts/expressions/template_element.rs
```diff
@@ -1,7 +1,20 @@
...
+        let TsTemplateElementFields {
+            dollar_curly_token,
+            ty,
+            r_curly_token,
+        } = self.as_fields();
+
+        Ok(format_elements![
+            dollar_curly_token.format(formatter)?,
+            ty.format(formatter)?,
+            r_curly_token.format(formatter)?,
```
ematipico, line 17: The formatting of template literals (this one and the JS ones) has be revisited (not in this PR, but it should).

At the moment we don't correctly break if we have a long expression: [link to playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&indentWidth=2&typescript=false&jsx=false#Y29uc3QgYSA9IGBkYWZzO2tsamRmc2FqaztmYWtsZHNqZmRzamtsO2Zkc2prbGZkanNrbGxmamtzZGFmajtrc2RsYTtmc2Rqa2wgICR7IHNvbWV0aGluZyAmJiBlbHNlbHNlZWxzZWxzZWVsc2Vsc2VlbHNlbHNlZWxzZWxzZSAmJiBibGFibGFibGEgPyAiaGVsbG8iIDogIm5vdCBoZWxsbyIgfSA7YXNkZms7ajtzZGZqbGFrZmthamxzZGZkamtsYXNqa2xmc2Rramxmc2RramxkZnM7a2xqc2Rma2pmbHNkO2Zkc2tsajtmYWprZHNsO2Zhc2Rqa2w7a2psc2RmajtsZmthc2RgOw==)

crates/rome_formatter/tests/specs/prettier/typescript/cast/generic-cast.ts.snap
```diff
@@ -87,15 +87,17 @@ function y() {
     never,
   >>someExistingConfigMap.mergeDeep(fallbackOptions);
 
-  const stillTooLong2 = <Immutable.Map<string, boolean, number, object, null, undefined, any, void, never> | undefined>someExistingConfigMap.mergeDeep(
-    fallbackOptions,
-  );
+  const stillTooLong2 = <
+    | Immutable.Map<string,boolean,number,object,null,undefined,any,void,never>
+    | undefined>someExistingConfigMap.mergeDeep(fallbackOptions);
```
ematipico, line 92: This doesn't seem right.

crates/rome_formatter/tests/specs/prettier/typescript/cast/generic-cast.ts.snap
```diff
@@ -146,35 +148,63 @@ function x() {
...
-  };
+  const testObjLiteral2 = <
+    | PermissionsChecker<object>
+    | undefined
+    | number
+    | string
+    | boolean
+    | null
+    | never
+    | object>{ prop1: "myPropVal" };
```
ematipico, line 193: It seems we don't break before `>`

crates/rome_formatter/tests/specs/prettier/typescript/class-comment/class-implements.ts.snap
```diff
@@ -84,71 +84,53 @@ b // comment-b2
...
   constructor() {}
 }
 
-class a3 extends b 
-implements
-// comment
-z,
-y  {
+class a3 extends b implements // comment
+z, y {
```
ematipico, line 97: Would it make sense to indent this line?

ematipico (changes requested):
I see that this PR now does more than before, when it was open. Would mind update the description of the PR?

Some notes:
- I noticed there are some blocking bugs, they be fixed
- I would personally add tests to our test suite too. Tests that we write and that we know that cover certain situations of some code. The prettier test suite has other stuff and it could be difficult to understand what's the "correct format" and which one is the "not implemented one"

crates/rome_formatter/tests/specs/prettier/typescript/prettier-ignore/mapped-types.ts.snap
leops: That was an easy fix so I included it into this batch of improvements

crates/rome_formatter/tests/specs/prettier/typescript/intersection/intersection-parens.ts.snap
leops: Ah yes, sometimes using `if_group_breaks` can be tricky as it doesn't just conditionally remove tokens but also any trivia piece that may be attached to them, I wonder if that's something that could be improved without making the conditional element functions less ergonomic to use ...

crates/rome_formatter/src/ts/types/intersection_type.rs
```diff
@@ -1,7 +1,36 @@
...
+        let TsIntersectionTypeFields {
+            leading_separator_token,
+            types,
+        } = self.as_fields();
+
+        let leading_separator_token = match leading_separator_token {
+            Some(token) => {
+                // The SyntaxToken is converted into a FormatElement using
+                // Token::from to strip the token's trivia pieces which are
+                // then reinserted informat_replaced outside of the
```
ematipico, line 20:
```suggestion
                // then reinserted in format_replaced outside of the
```

crates/rome_formatter/tests/specs/prettier/typescript/intersection/intersection-parens.ts.snap
ematipico: I believe this is a particular case that doesn't require creating a new functionality/API. I think it's the first time that we applying such logic (unions/intersections in TS). I don't there are other similar situations, considering JSX too

ematipico (approved):

PR #2180, opened 2022-02-28
feat(rslint_parser): allow ts property named get and set

xunilrj:
I am creating the Typescript bench and searching for holes in our implementation. This fixes the problem I found at https://github.com/PanayotCankov/nativescript-big-dts/blob/master/ios.d.ts

```
error[SyntaxError]: expected an identifier, a string literal, a number literal, a private field name, or a computed name but instead found ':'
     ┌─ ios.d.ts:2270:5
     │
2270 │     set: NSSet;
     │        ^ Expected an identifier, a string literal, a number literal, a private field name, or a computed name here
```

### DIFF
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -711,7 +711,7 @@ fn parse_class_member_impl(
                     && !is_at_line_break_or_generator
                     && matches!(member_name_text, "get" | "set")
                 {
-                    let is_getter = dbg!(member_name_text == "get");
+                    let is_getter = member_name_text == "get";
 
                     // test getter_class_member
                     // class Getters {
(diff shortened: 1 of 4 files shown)


crates/rslint_parser/src/syntax/class.rs
```diff
@@ -704,9 +704,13 @@ fn parse_class_member_impl(
     match member_name {
         Some(member_name) => {
             if member_name.kind() == JS_LITERAL_MEMBER_NAME {
+                let is_at_semicolon = p.at(T![:]);
```
MichaReiser, line 707: Should this be `is_at_colon` or should the token be a `;`?

MichaReiser (approved):

crates/rslint_parser/src/syntax/class.rs
xunilrj: Typo. Should be ```is_at_colon```.

PR #2182, opened 2022-03-01
feat(rome_formatter): chain of conditionals

### LINKED ISSUE #2157: 📎 Format conditionals
### Description

At the moment we don't do anything fancy to format conditionals.

We should apply an algorithm to format them in a consistent and pleasant way. The implementation should be generic as we have two different nodes that will benefit from it:
- `TsConditionalType`
- `JsConditionalExpression`

Their AST is slightly different but the core of the conditional is essentially the same.

ematipico:
## Summary

Closes #2157 

The algorithm is quite simple and I didn't bother to make more generic. Trying to make it generic could have resulted in something more difficult to implement/understand. I simply copied over the logic and it seems to work file.

## Test Plan

New tests


crates/rome_formatter/src/js/expressions/conditional_expression.rs
```diff
@@ -17,16 +17,38 @@ impl ToFormatElement for JsConditionalExpression {
...
-            test.format(formatter)?,
-            space_token(),
+        let alternate = alternate?;
+        let consequent = consequent?;
+
+        let is_alternate_conditional =
+            alternate.syntax().kind() == JsSyntaxKind::JS_CONDITIONAL_EXPRESSION;
+        let is_consequent_conditional =
+            consequent.syntax().kind() == JsSyntaxKind::JS_CONDITIONAL_EXPRESSION;
+        let parent_is_conditional = self.syntax().parent().map_or(false, |n| {
```
leops, line 27: This code isn't particularly wrong (the SyntaxNodes exist so we can query the parent after all) but in general for other similar case (like call chains or if-else chains) we format the entire chain from for parent node rather than change how child nodes are formatted based on the type of their parent

crates/rome_formatter/src/js/expressions/conditional_expression.rs
```diff
@@ -17,16 +17,38 @@ impl ToFormatElement for JsConditionalExpression {
...
+        let body = if is_alternate_conditional || is_consequent_conditional || parent_is_conditional
+        {
+            indent(format_elements![
+                hard_line_break(),
+                join_elements(soft_line_break(), vec![consequent, alternate],)
+            ])
+        } else {
+            join_elements(space_token(), vec![consequent, alternate])
+        };
+        Ok(format_elements![test, space_token(), body])
```
leops, line 52: I don't think it will change the result of the formatting, but it would make sense to wrap this in a group if `!parent_is_conditional`

crates/rome_formatter/src/js/expressions/conditional_expression.rs
ematipico: Do you think we should flatten the structure instead? Because that's the other alternative I can see. The idea is that if the out-most conditional breaks, all the conditionals should break. That's why I check the parent. This is used to avoid cases like these:

```js
somethingThatsAReallyLongPropName
	? somethingThatsAReallyLongPropName
	: a ? : somethingThatsAReallyLongPropName; // this doesn't break
```

crates/rome_formatter/src/js/expressions/conditional_expression.rs
```diff
@@ -1,32 +1,97 @@
...
-            alternate,
-        } = self.as_fields();
+        format_conditional(self, formatter, false)
+    }
+}
 
-        Ok(format_elements![
-            test.format(formatter)?,
-            space_token(),
+fn format_conditional(
```
leops, line 16: Would it be possible to have this function as a generic utility instead of being declared twice for `JsConditionalExpression` and `TsConditionalType` ?

ematipico: Done!

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,181 @@
...
+        alternate.syntax().kind(),
+        JsSyntaxKind::JS_CONDITIONAL_EXPRESSION | JsSyntaxKind::TS_CONDITIONAL_TYPE
+    );
+
+    let is_consequent_conditional = matches!(
+        consequent.syntax().kind(),
+        JsSyntaxKind::JS_CONDITIONAL_EXPRESSION | JsSyntaxKind::TS_CONDITIONAL_TYPE
+    );
+
+    let consequent = if is_consequent_conditional || parent_is_conditional {
```
leops, line 126: If I'm reading this correctly, the two nested if expressions could be rewritten as `if is_consequent_conditional {} else if parent_is_conditional {} else {}`

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,181 @@
...
+        if is_consequent_conditional {
+            let consequent = cast(consequent.syntax().to_owned());
+            let consequent = format_conditional(&consequent, formatter, true)?;
+            format_elements![question_mark.format(formatter)?, space_token(), consequent]
+        } else {
+            indent(format_elements![
```
leops, line 132: `indent` elements change the current indentation level in the printer but that change doesn't take effect until the next line break: in this case that will only happen if the consequent expression breaks across multiple lines, which is also something that could happen in the second `else` branch so wouldn't it make sens to apply the same formatting there ?

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,181 @@
...
+                        alternate: false_type?,
+                        formatter,
+                    },
+                    parent_is_conditional,
+                    |node| TsConditionalType::cast(node).unwrap().syntax().to_owned(),
+                )?,
+            )
+        }
+
+        _ => unreachable!(
```
leops, line 92: I think `panic` is the correct macro to use here, `unreachable` is generally used to indicate code paths that the program can never actually enter but this branch could be executed if the wrong node type was passed to `format_conditional`

crates/rome_formatter/src/utils/format_conditional.rs
ematipico: > apply the same formatting there
What do you mean by "there"?

leops: I meant in the second `else` branch on line 138, if I understand this correctly when the path on line 132 is taken (`!is_consequent_conditional && parent_is_conditional`) the consequent element will be formatted like this:
```js
parent
    ? child
        ? // line break
            consequent
        : alternate
    : alternate
```
But when the path on line 138 is taken (`!is_consequent_conditional && !parent_is_conditional`) the code will be formatted like this:
```js
parent
    ? // line break
    consequent
    : alternate
```

ematipico: I think it's the other way around. So I want to cover two cases:
1. plain ternaries
2. nested ternaries

Plain ternaries it's easy and should not involve any particular logic:

```js
somethinglong ? somethingistoolong : short
```

Nested ternaries are bit different and the **golden rules** is: if there's at least one level of ternaries, **ALL** ternaries have to break, regardless if  their plain or not.

Now, going back this part that you mentioned before:

> but that change doesn't take effect until the next line break

Surely this is **not** what I want. 


> I meant in the second `else` branch on line 138, if I understand this correctly when the path on line 132 is taken (`!is_consequent_conditional && parent_is_conditional`) the consequent element will be formatted like this:
> 
> ```js
> parent
>     ? child
>         ? // line break
>             consequent
>         : alternate
>     : alternate
> ```

Yeah this is not what's intended I think I now understand what you meant. The code is more complex than intended. That's what happens to me when I rewrite the same thing over and over 😅

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,172 @@
...
+                    },
+                    parent_is_conditional,
+                    |node| TsConditionalType::cast(node).unwrap().syntax().to_owned(),
+                )?,
+            )
+        }
+
+        _ => panic!(
+            "This function should be used only for JsConditionalExpression and TsConditionalType"
+        ),
```
MichaReiser, line 101: Nit: How about an enum that has two variants, ConditionalExpression and ConditionalType

```
enum Conditional {
  Expression(JsConditionalExpression),
  Type(TsConditionalType),
}
```

It's then guaranteed at compile time that no other type is passed ever at compile time.

You could even implement the members that you currently abstract with `FormatConditionalPayload`

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,172 @@
...
+    };
+
+    Ok(format_elements![head, body])
+}
+
+fn format_conditional_body<Node: AstNode + ToFormatElement, Cast>(
+    payload: FormatConditionalPayload<Node>,
+    parent_is_conditional: bool,
+    cast: Cast,
```
MichaReiser, line 110: Nit: Can you instead use `Node::cast(...)` instead of passing `cast` explicitly? or you could add it to the `Conditional` enum

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,172 @@
...
+        JsSyntaxKind::JS_CONDITIONAL_EXPRESSION | JsSyntaxKind::TS_CONDITIONAL_TYPE
+    );
+
+    let is_consequent_conditional = matches!(
+        consequent.syntax().kind(),
+        JsSyntaxKind::JS_CONDITIONAL_EXPRESSION | JsSyntaxKind::TS_CONDITIONAL_TYPE
+    );
+
+    let consequent = if is_consequent_conditional {
+        let consequent = cast(consequent.syntax().to_owned());
```
MichaReiser, line 134: Instead of using `is_consequent_conditional` with `cast` (that calls into unwrap) how about you change `cast` to return `Option<T>` (because `AstNode::cast` already answers if `is_consequent` and then performs the cast).

crates/rome_formatter/src/utils/format_conditional.rs
ematipico: I am failing to understand this part 🤔

MichaReiser: Type arguments in rust are different from TypeScript. They don't get erased at runtime. 

You can do the following inside of the body

```
let condition_node = Node::cast(syntax)
```

This will be transformed into `TsConditionalType::cast(syntax)` if the `Node` type argument is `TsConditionalType`

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,180 @@
...
+        Conditional::Expression(conditional_expression) => {
+            let JsConditionalExpressionFields {
+                consequent,
+                colon_token,
+                test,
+                alternate,
+                question_mark_token,
+            } = conditional_expression.as_fields();
+
+            (
```
MichaReiser, line 68: What are these parentheses for. Or, what kind of syntax is this 😱

crates/rome_formatter/src/utils/format_conditional.rs
```diff
@@ -0,0 +1,180 @@
...
+    };
+
+    Ok(format_elements![head, body])
+}
+
+fn format_conditional_body<Node: AstNode + ToFormatElement, ToConditional>(
+    payload: FormatConditionalPayload<Node>,
+    parent_is_conditional: bool,
+    to_conditional: ToConditional,
+) -> FormatResult<FormatElement>
```
MichaReiser, line 125: Nit: I wonder if we could merge this method with the `format_conditional` by:

* Adding a `as_conditional_payload()` method to `Conditional`. It matches to extract the fields. 
* Adding a `cast_conditional(syntax)` to `Conditional` to replace the `to_conditional` 

That would also remove the need for this method to be generic.

MichaReiser (commented):
I've one more suggestion how we could polish the implementation.

crates/rome_formatter/src/utils/format_conditional.rs
ematipico: the `match` returns a tuple :)

MichaReiser (approved):

Issue #2187, opened 2022-03-01
📎 TS: Class member modifiers [Must]


MichaReiser:
### Description

TypeScript doesn't define a strict order for many class member modifiers. For example, all these orders are valid (and probably more):

```ts
class Test {
    private declare c?: string;
    declare private d?: string;
    declare private readonly e: string;
    private readonly declare f: string;
    protected abstract readonly g: string;
    protected readonly abstract h: string;
}
```

Goal:

* Support different modifier orders
* Add support for `override` modifier
* Add support for `declare` modifier

Issue #2189, opened 2022-03-01
📎TS: Ambient Class Members


MichaReiser:
### Description

Support parsing of class members in an ambient context:

```ts
declare class Test { ... }
// or
declare module {
  class Test { ... }
}
```

Main focus on:
* Verify constructors in ambient context
* Verify AST/Parser output in ambient context

PR #2190, opened 2022-03-01
fix(rslint_parser): Model class member modifiers as array

MichaReiser:
## Summary

TypeScript doesn't enforce a strict class member modifier ordering. For example, all the following
modifier orders are valid:

```ts
class Test {
    private declare c?: string;
    declare private d?: string;
    declare private readonly e: string;
    private readonly declare f: string;
    protected abstract readonly g: string;
    protected readonly abstract h: string;
}
```

`declare` can be the first modifier or come last, same for `abstract`.

This PR replaces the modifier tokens on the individual class members with lists, so that different modifier orders can be supported. This comes at the cost that the AST API can no longer enforce the correct modifier ordering. 

This PR introduces a different modifier list for:

* Index signature members
* Method members
* method signatures
* property members
* property signatures

The advantage of having separate lists is that the AST mutations API can enforce that these lists only contain valid modifiers. However, the downside of this is that the modifier verification logic can't be pushed into a second pass because invalid "modifiers" automatically become `JS_UNKNOWN`, thus, no longer have the correct type as needed in the checker. Having different list types further complicates the parser and introduces plenty of new nodes. But I still believe that this is worth the extra safety when it comes to mutating ASTs (or writing analyzers, you don't need to match on unsupported modifiers).

I decided not to change the modifiers on interface / type members for now. They don't allow arbitrary ordering, thus, having fixed fields is sufficient.

Part of #2187

## Test Plan

Reviewed the RAST file changes. Added one new test for a static initializer block with modifiers

Babel-TS regressions: These are issues with the coverage setup. These test should fail and are now failing but our test coverage set up doesn't handle this correctly. I don't want to test the coverage as part of this PR.

### DIFF
--- a/crates/rslint_parser/src/syntax/class.rs
+++ b/crates/rslint_parser/src/syntax/class.rs
@@ -483,7 +483,7 @@ fn parse_class_member(p: &mut Parser, inside_abstract_class: bool) -> ParsedSynt
             Present(member)
         }
         Absent => {
-            assert!(modifiers.is_empty());
+            debug_assert!(modifiers.is_empty());
             modifiers.abandon(p);
             Absent
         }
@@ -1839,7 +1839,7 @@ impl ClassMemberModifiers {
     /// ## Panics
     /// If the modifier list isn't empty
     fn abandon(mut self, p: &mut Parser) {
-        assert!(self.is_empty());
+        debug_assert!(self.is_empty());
         self.list_marker.undo_completion(p).abandon(p);
         self.bomb.defuse();
     }


crates/rome_formatter/src/js/classes/getter_class_member.rs
```diff
@@ -22,13 +21,14 @@ impl ToFormatElement for JsGetterClassMember {
             body,
         } = self.as_fields();
 
+        let modifiers = if modifiers.is_empty() {
+            empty_element()
+        } else {
+            format_elements![modifiers.format(formatter)?, space_token()]
+        };
+
```
MichaReiser, line 29: @ematipico / @leops is there a better way to format a list and add a white spaces between it and the next token but only if the list wasn't empty?

crates/rome_formatter/tests/specs/prettier/typescript/declare/declare-get-set-field.ts.snap
```diff
@@ -16,7 +16,7 @@ class C {
 # Output
 ```js
 class C {
-  declare get: string;
+  declare get: string
```
MichaReiser, line 19: The `declare` modifier isn't fully supported yet.I'll look into this as part of #2189

crates/rome_formatter/src/js/classes/getter_class_member.rs
ematipico: If `modifiers` is `Option`, you can use `format_with_or_else`. Otherwise you can use `.format_with`

MichaReiser: It's neither an `Option` nor a `Result`. Lists are always present.

The way I understand `format_with` is that it always calls the callback if formatting was successful. It doesn't seem to be testing if the return `FormatElement` is empty.

ematipico: Oh, in that case we don't have an API for this particular situation, but we can create it!

ematipico: Actually, `formatter.format_separated` should work for you. Try to use it

MichaReiser: My list isn't separated :(

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1721,50 +1723,116 @@ enum ModifierKind {
...
-                    Some((kind, range))
-                } else {
-                    None
-                }
-            })
+        self.modifiers.iter()
+    }
+
+    fn is_empty(&self) -> bool {
+        self.modifiers.0.iter().all(|modifier| modifier.is_none())
```
MichaReiser, line 1803: My next PR will change the structure used to store the modifiers so that this operation becomes `O(1)`

crates/rome_formatter/src/js/classes/getter_class_member.rs
leops: The correct combinator to use here would be `join_elements` with `space_token` as a separator. Then its okay to skip emitting the entire element if the list is empty, but it should also be okay to have the final space even if the list was empty as spaces get deduplicated by the printer and skipped at the start of a line

MichaReiser:
!bench_parser

crates/rome_formatter/src/js/classes/getter_class_member.rs
MichaReiser: The list implementations do use `join_elements` with a `space_token` as separator. 

That's a neat detail that should work. Let me change it to always add a space token even if it is empty

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -432,64 +433,60 @@ impl ParseNodeList for ClassMembersList {
...
-                .err_builder("Only abstract classes can have abstract members")
-                .primary(member_abstract_token_range, "");
-            p.error(err);
-            member.change_to_unknown(p);
             Present(member)
         }
-        (Present(member), _) => Present(member),
-        (Absent, _) => Absent,
+        Absent => {
+            assert!(modifiers.is_empty());
```
ematipico, line 486: I would change this to `debug_assert`, it doesn't seem we need this check in production?

xtask/codegen/js.ungram
```diff
@@ -631,29 +631,38 @@ JsAnyClassMember =
...
 	| JsEmptyClassMember
 	| JsUnknownMember
-	
+
+JsStaticModifier = modifier: 'static'
+TsDeclareModifier = modifier: 'declare'
+TsReadonlyModifier = modifier: 'readonly'
+TsAbstractModifier = modifier: 'abstract'
+TsOverrideModifier = modifier: 'override'
+TsAccessibilityModifier = modifier_token: ('private' | 'protected' | 'public')
```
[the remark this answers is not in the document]
MichaReiser: Not for unions.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1721,50 +1743,129 @@ enum ModifierKind {
...
+    fn has(&self, kind: ModifierKind) -> bool {
+        self.0[kind as usize].is_some()
+    }
+
+    /// Iterates over the present modifiers
+    fn iter(&self) -> impl Iterator<Item = (ModifierKind, &Range<usize>)> {
+        self.0.iter().enumerate().filter_map(|(index, range)| {
+            if let Some(range) = range {
+                assert!(index < ModifierKind::__LAST as usize);
+                let kind = unsafe { std::mem::transmute::<u8, ModifierKind>(index as u8) };
```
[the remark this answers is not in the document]
MichaReiser: We do. But I changed out the list implementation in my next PR and it removes the unsafe block

ematipico (approved):
It seems that there will be more changes in the pipeline, so I wait for the next PR for more feedback, if needed.

Regarding the order of the modifiers, I think this will be up to the formatter. Once you'll merge this PR, I will open a task for that unless the grammar will change again.

I

PR #2192, opened 2022-03-02
fix(rslint_parser): Validate TS class member modifiers

MichaReiser:
## Summary

The different class members allow (some) of the following modifiers:

* `private`, `protected`, `public`
* `readonly`
* `override`
* `abstract`
* `static`
* `declare`

This PR adds the necessary violations for the TypeScript modifiers (all except `static`) so that the parser emits an error if:
* If a modifier appears out of order
* If a modifier can't be used with another modifier
* If a modifier can't be used for this member

It further adds support for parsing the `override` modifier.

This PR introduces the new `smallvec` dependency to store the parsed modifiers. This has two benefits compared to the previous data structure. 

* No heap allocations for all valid member modifiers (longest to my knowledge is `private declare readonly static member`
* Smaller footprint then allocating a fixed array with slots for all possible modifiers
* Centralized verification of the modifiers. It's no longer needed to first deduplicate the modifiers when parsing and then validating the combinations in a second pass when it's know what the member kind is. 

Part of #2187 

## Test Plan

Added new integration tests validating the modifier ordering

Regressions:
* babel-ts: Tests that are now failing. Issue with the test runner, see #2193 
* microsoft-ts: Tests that are expected to fail with a type checker diagnostic that a member must or cannot have a `override` modifier (missing in the parent class).

### DIFF
--- a/crates/rslint_parser/test_data/inline/err/ts_readonly_modifier_non_class_or_indexer.rast
+++ b/crates/rslint_parser/test_data/inline/err/ts_readonly_modifier_non_class_or_indexer.rast
@@ -226,7 +226,7 @@ JsModule {
       8: R_CURLY@139..141 "}" [Newline("\n")] []
   3: EOF@141..142 "" [Newline("\n")] []
 --
-error[SyntaxError]: 'readonly cannot appear on a constructor declaration.
+error[SyntaxError]: 'readonly' cannot appear on a constructor declaration.
   ┌─ ts_readonly_modifier_non_class_or_indexer.ts:2:3
   │
 2 │   readonly constructor() {}
--- a/crates/rome_formatter/tests/specs/prettier/typescript/class/constructor.ts.snap
+++ b/crates/rome_formatter/tests/specs/prettier/typescript/class/constructor.ts.snap
@@ -82,7 +82,7 @@ error[SyntaxError]: expected a type parameter but instead found '>'
 17 │   constructor<>() {}
    │               ^ Expected a type parameter here
 
-error[SyntaxError]: constructors cannot have type parameters
+error[SyntaxError]: constructors cannot have type parameters.
    ┌─ constructor.ts:17:14
    │
 17 │   constructor<>() {}
--- a/crates/rome_formatter/tests/specs/prettier/typescript/class/duplicates-access-modifier.ts.snap
+++ b/crates/rome_formatter/tests/specs/prettier/typescript/class/duplicates-access-modifier.ts.snap
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/prettier_tests.rs
-assertion_line: 119
+assertion_line: 125
 expression: duplicates-access-modifier.ts
 
 ---
@@ -30,7 +30,7 @@ class Foo {
 
 # Errors
 ```
-error[SyntaxError]: Accessibility modifier already seen
+error[SyntaxError]: Accessibility modifier already seen.
   ┌─ duplicates-access-modifier.ts:2:10
   │
 2 │   public public a;
@@ -38,7 +38,7 @@ error[SyntaxError]: Accessibility modifier already seen
   │   │       
   │   first modifier
 
-error[SyntaxError]: Accessibility modifier already seen
+error[SyntaxError]: Accessibility modifier already seen.
   ┌─ duplicates-access-modifier.ts:3:11
   │
 3 │   private public b;
@@ -46,7 +46,7 @@ error[SyntaxError]: Accessibility modifier already seen
   │   │        
   │   first modifier
 
-error[SyntaxError]: Accessibility modifier already seen
+error[SyntaxError]: Accessibility modifier already seen.
   ┌─ duplicates-access-modifier.ts:4:13
   │
 4 │   protected private c;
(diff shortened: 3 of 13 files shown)


crates/rome_formatter/tests/specs/prettier/typescript/decorators/inline-decorators.ts.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/prettier_tests.rs
-assertion_line: 119
+assertion_line: 125
```
ematipico, line 3: This change can be discarded

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1719,119 +1531,179 @@ fn parse_modifier(p: &mut Parser, constructor_parameter: bool) -> Option<Modifie
...
+        self.flags.contains(kind.as_flags())
+    }
+
+    /// Returns the range for the passed in modifier or [None] if the modifier isn't set or is a missing marker
+    fn get_first_range(&self, kind: ModifierKind) -> Option<TextRange> {
+        if self.flags.contains(kind.as_flags()) {
+            self.modifiers
+                .iter()
+                .find(|m| m.kind == kind)
+                .map(|m| m.as_text_range())
```
ematipico, line 1689: nit: I think we can use `find_map`?

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1869,4 +1741,401 @@ impl ClassMemberModifiers {
...
+                        self.get_first_range_unchecked(ModifierKind::Readonly),
+                    ));
+                }
+            }
+        }
+
+        None
+    }
+
+    fn check_class_member(
```
ematipico, line 2116: Should it be `check_class_members`? (plural) There's a loop over ALL modifiers

crates/rslint_parser/src/syntax/js_parse_error.rs
```diff
@@ -248,3 +243,54 @@ pub(crate) fn invalid_assignment_error(p: &Parser, range: Range<usize>) -> Diagn
...
+
+pub(crate) fn accessibility_modifier_already_seen(
+    p: &Parser,
+    second_range: TextRange,
+    first_range: TextRange,
+) -> Diagnostic {
+    p.err_builder("Accessibility modifier already seen")
+        .primary(second_range, "duplicate modifier")
+        .secondary(first_range, "first modifier")
+}
```
ematipico, line 266: It seems that "accessibility" is a TS concept (by checking for the bitflag, it groups only TS stuff). Should we move it under `ts_parse_error`?

ematipico (approved):
Overall, it looks good and understandable! Even though the function `check_class_member_modifier` is long, it's very straightforward to read and follow! Definitely a big improvement compared to before! 🚀 

I just left minor comments, not blockers.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: I'll merge it with `complete`. The two methods are always called right after each other.

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -1719,119 +1531,179 @@ fn parse_modifier(p: &mut Parser, constructor_parameter: bool) -> Option<Modifie
...
+    fn as_text_range(&self) -> TextRange {
+        TextRange::new(
+            self.start,
+            self.start.add(TextSize::from(self.length as u32)),
+        )
+    }
 }
 
-// replace length with std::mem::variant_count() when it becomes stable
+// 4 is sufficient to store all valid modifiers without requiring any heap allocations.
```
xunilrj, line 1631: Should we aim for the maximum or for the average?

crates/rslint_parser/test_data/inline/err/ts_readonly_modifier_non_class_or_indexer.rast
```diff
@@ -0,0 +1,262 @@
...
+          5: R_PAREN@135..137 ")" [] [Whitespace(" ")]
+          6: JS_FUNCTION_BODY@137..139
+            0: L_CURLY@137..138 "{" [] []
+            1: JS_DIRECTIVE_LIST@138..138
+            2: JS_STATEMENT_LIST@138..138
+            3: R_CURLY@138..139 "}" [] []
+      8: R_CURLY@139..141 "}" [Newline("\n")] []
+  3: EOF@141..142 "" [Newline("\n")] []
+--
+error[SyntaxError]: 'readonly cannot appear on a constructor declaration.
```
xunilrj, line 229: Why not the same message? "Readonly can only appear on a property declaration or index signature."

xunilrj (approved):

crates/rslint_parser/src/syntax/class.rs
MichaReiser: This is the maximum of valid modifiers. There could be more in case a member has invalid or repeated modifiers. 

I haven't seen much difference when playing around with this number but that may as well just be because we haven't had many TS test cases back then (that use classes as well). However, that also means that a larger value doesn't hurt performance. 

In the end, I had to pick a value and 4 seemed less arbitrary and I honestly don't know what the average is (2? 1? 0?). But that's something we can easily change in the future

crates/rslint_parser/test_data/inline/err/ts_readonly_modifier_non_class_or_indexer.rast
MichaReiser: This is mainly because of the order how the checks are performed. There will always be some inconsistency because the message is `'x' can only appear on a constructor declaration` for all other modifiers but wouldn't for readonly if we make an exception here.

PR #2193, opened 2022-03-02
chore(coverage): Infer should error for ts/babel tests

MichaReiser:
## Summary
Not all tests in the `typescript` directory are expected to pass. A test is expected to fail if the `options.json` has a `throws` property or the `output.json` contains any `SyntaxError`

This PR infers the expected outcome by matching the content of the `options.json` and `output.json` (parsing the JSON feels overkill).

There are still plenty of tests that should pass but fail. However, these tests do contain syntax errors. I'm not sure if it's worth keeping this test set (example; modifiers-override)

## Test Plan

Verified that some tests now reported "Incorrectly passed"

### DIFF
--- a/xtask/coverage/src/ts/ts_babel.rs
+++ b/xtask/coverage/src/ts/ts_babel.rs
@@ -6,26 +6,31 @@ use crate::{
     check_file_encoding,
     runner::{TestCase, TestCaseFiles, TestRunOutcome, TestSuite},
 };
-use std::path::PathBuf;
+use std::path::Path;
 
 const CASES_PATH: &str = "xtask/coverage/babel/packages/babel-parser/test/fixtures/typescript";
 
 struct BabelTypescriptTestCase {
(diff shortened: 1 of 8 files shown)


xtask/coverage/src/ts/ts_babel.rs
```diff
@@ -6,26 +6,31 @@ use crate::{
...
-            .components()
-            .rev()
-            .nth(1)
-            .and_then(|x| x.as_os_str().to_str())
-            .unwrap_or("")
+            .parent()
+            .unwrap()
+            .strip_prefix(CASES_PATH)
+            .unwrap()
+            .display()
```
MichaReiser, line 26: Only having the last component makes it difficult to know which test is failing because some test cases have the same name (but have different parent directories). This change includes the parent directory names as well.

ematipico (approved):

PR #2195, opened 2022-03-02
feat(rome_lsp): correctly detect source type from file

### LINKED ISSUE #2194: 📎 Automatic detection of source type
### Description

At the moment when parsing, we don't do any check on the files and we parse it as module. This is a source of issues for our parser. We should add a way to detect `SourceType` correctly

ematipico:
## Summary

Closes #2194 

Notable changes:

- `SourceType` is now extracted from `RomePath`, which is a tiny wrapper around `PathBuf` ;
- `rslint_parser` implement `TryFrom<RomePath> for SourceType`;
- extracted the old logic inside `SoureType::from_path` inside a private function, which is called by `TryFrom`;
- expanded `RomePath` by adding the `file_id`, which is used inside the LSP server and the parser;
- updated the dependencies of few crates;

This PR **does not** try to merge together the concept of `Language` that we have inside `rome_core` and `rslint_parser`. That part is still separated.

## Test Plan

I checked locally that:
- trying to format a `.js` file, the `SourceType.langauge` is `Language::JavaScript`
- trying to format a `.ts` file, the `SourceType.langauge` is `Language::TypeScript`


xtask/coverage/src/ts/ts_microsoft.rs
```diff
@@ -170,9 +172,9 @@ fn extract_metadata(code: &str, path: &str) -> TestCaseMetadata {
 }
 
 fn add_file_if_supported(files: &mut TestCaseFiles, name: String, content: String) {
-    let path = Path::new(&name);
     // Skip files that aren't JS/TS files (JSON, CSS...)
-    if let Some(mut source_type) = SourceType::from_path(path) {
+    if let Ok(source_type) = RomePath::new(&name).try_into() {
+        let mut source_type: SourceType = source_type;
```
ematipico, line 177: @leops , @xunilrj is there a better way to this? Issue I had was that the compiler wasn't bale to infer the type of `source_type`.

leops: I think `if let Ok(mut source_type) = SourceType::try_from(RomePath::new(&name))` should work

crates/rslint_parser/src/lib.rs
```diff
@@ -337,6 +317,76 @@ impl SourceType {
...
+    type Error = RomeError;
+
+    fn try_from(path: RomePath) -> Result<Self, Self::Error> {
+        let file_name = path.file_name().ok_or(RomeError::CantReadTheFile)?;
+        let file_name = file_name.to_str().ok_or(RomeError::CantReadTheFile)?;
+        let extension = path.extension_as_str();
+
+        compute_source_type_from_path_or_extension(file_name, extension)
+    }
+}
```
MichaReiser, line 354: Do we need all three implementations? Shouldn't `&SourceType` cover the others as it's the most restrictive?

crates/rslint_parser/src/lib.rs
```diff
@@ -337,6 +317,76 @@ impl SourceType {
...
+            "cjs" => SourceType::js_module().with_module_kind(ModuleKind::Script),
+            "jsx" => SourceType::jsx(),
+            "ts" | "mts" => SourceType::ts(),
+            "cts" => SourceType::ts().with_module_kind(ModuleKind::Script),
+            "tsx" => SourceType::tsx(),
+            _ => return Err(RomeError::SourceFileNotSupported(extension.into())),
+        }
+    };
+    Ok(source_type)
+}
```
MichaReiser, line 377: How about implementing `TryFrom<&Path> for SourceType`. You can then call this implementation in the `TryFrom<RomePath>` as well (which has access to `Path`)

crates/rslint_parser/src/lib.rs
```diff
@@ -337,6 +317,76 @@ impl SourceType {
...
+impl Debug for SourceType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Source Type")
+            .field("Language", &self.language)
+            .field("Module kind", &self.module_kind)
+            .field("Variant", &self.variant)
+            .field("Version", &self.version)
+            .finish()
+    }
+}
```
MichaReiser, line 388: How's this different from the default implementation that you get with `#derive(Debug)`?

xtask/bench/src/lib.rs
```diff
@@ -116,7 +117,7 @@ pub fn run(args: RunArgs) {
             Ok((id, code)) => {
                 let code = code.as_str();
 
-                let source_type = SourceType::from_path(Path::new(&id)).unwrap();
+                let source_type: SourceType = RomePath::new(&id).try_into().unwrap();
```
MichaReiser, line 120: Is it worth constructing a `RomePath` Using `Path` gives us the same functionality (without an added dependency)?

crates/rome_core/src/lib.rs
```diff
@@ -29,6 +31,43 @@ impl Default for App {
...
+impl Debug for RomeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RomeError::SourceFileNotSupported(extension) => {
+                write!(f, "Rome doesn't support this {extension} yet")
+            }
+            RomeError::CantReadTheFile => {
+                write!(f, "Rome is not able to read the file")
+            }
+        }
```
MichaReiser, line 52: You can just call into the `Display` implementation.

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -20,23 +21,32 @@ pub(crate) fn to_format_options(params: &FormattingOptions) -> FormatOptions {
...
+        format_options,
+        text,
+        workspace_settings,
+        rome_path,
+    } = params;
+
+    let source_type = rome_path.try_into()?;
 
-    let parse_result = parse(text, file_id, syntax);
+    let file_id = rome_path.file_id().unwrap_or(0_usize);
```
MichaReiser, line 41: I'm not sure if we should use `RomePath` to infer the language inside of the LSP. @yassere told me that VS Code exposes the language ID (TS, TSX, JS, JSX) and I think we better use that than the path.

crates/rslint_parser/src/lib.rs
```diff
@@ -337,6 +317,76 @@ impl SourceType {
     }
 }
 
+impl TryFrom<&mut RomePath> for SourceType {
```
MichaReiser, line 320: It feels a bit overkill to have a dependency on `rome-path` just to be able to implement `TryFrom` for `SourceType`. 

The parser's currently independent from `RomePath` and I don't think that's bad. All the parser must be able to do is to parse a file, this is where the `Path` abstraction is sufficient.

crates/rslint_parser/src/lib.rs
ematipico: Yeah that should be a good compromise!

crates/rome_lsp/src/handlers/formatting.rs
ematipico: The `language_id` is just another [`enum` abstraction](https://github.com/rome/tools/blob/main/crates/rome_lsp/src/documents.rs#L6-L25), which is the third:
1. lsp
2. parser
3. core

The objective is to unify them, right? In the future we would need to do some kind of conversation and use just one `Language` somewhere 😅

MichaReiser: I mean VS code hands us a `Language` id. Shouldn't we use that to detect the file type?

crates/rslint_parser/Cargo.toml
```diff
@@ -8,6 +8,8 @@ description = "An extremely fast ECMAScript parser made for the rslint project"
 repository = "https://github.com/rslint/rslint"
 
 [dependencies]
+rome_core = { path = "../rome_core", version = "0.0.0" }
```
yassere, line 11: I think that a crate called `rome_core` should probably end up as the shared infrastructure used by various frontends (e.g. cli, lsp) and would itself be dependent on the parser, rather than the other way around. This is something we should probably revisit once we start working on the CLI and think more about how everything fits together.

ematipico: Yes, exactly. The only reason why I put it there is because of the `RomeError` enum, which is totally a temporary implementation that will be revisited later.

crates/rome_lsp/src/handlers/formatting.rs
MichaReiser: I can see that we need the capabilities. But let's say someone has mapped the extension `.es6` to `JavaScript`. Rome claims to be able to format `JavaScript`, thus VS Code will call our formatter plugin for that file and we then fail because we say, nah, we don't support this file.

That's why I believe we should respect the user-configured language (with module +JSX by default).

crates/rome_core/src/lib.rs
MichaReiser: I would call from `Debug` into `Display`. `Display` has tighter constraints than `Debug`, it must be something that's presentable to end-users whereas debug is any representation that's good enough for a dev.

crates/rslint_parser/src/lib.rs
```diff
@@ -337,6 +312,46 @@ impl SourceType {
     }
 }
 
+impl TryFrom<&Path> for SourceType {
+    type Error = RomeError;
+
+    fn try_from(path: &Path) -> Result<Self, Self::Error> {
+        let file_name = path.file_name().expect("Can't read the file");
+        let file_name = file_name.to_str().expect("Can't read the file");
```
MichaReiser, line 320: Are the expect messages correct? This code isn't reading the file, is it?

crates/rslint_parser/src/lib.rs
```diff
@@ -337,6 +312,46 @@ impl SourceType {
     }
 }
 
+impl TryFrom<&Path> for SourceType {
+    type Error = RomeError;
```
MichaReiser, line 316: Would it make sense to have a more specific error here than `RomeError` ?

Something like `UnknownSourceTypeErrror`. The caller can then handle it (and convert it to a `RomeError`).

MichaReiser (approved):

PR #2196, opened 2022-03-02
feat(rome_formatter): predictable order of modifiers

ematipico:
## Summary

This PR implements a predictable ordering of modifiers inside the formatter.

## Test Plan

Added few test to cover them. Will need to more once `declare` in stable

### DIFF
--- a/crates/rome_formatter/src/ts/lists/index_signature_modifier_list.rs
+++ b/crates/rome_formatter/src/ts/lists/index_signature_modifier_list.rs
@@ -1,21 +1,12 @@
-use crate::utils::{sort_modifiers_by_precedence, SortedModifiers};
+use crate::utils::sort_modifiers_by_precedence;
 use crate::{join_elements, space_token, FormatElement, FormatResult, Formatter, ToFormatElement};
-use rslint_parser::ast::{TsAnyIndexSignatureModifier, TsIndexSignatureModifierList};
+use rslint_parser::ast::TsIndexSignatureModifierList;
 
 impl ToFormatElement for TsIndexSignatureModifierList {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        let ordered_nodes = sort_modifiers_by_precedence(self, to_sorted_modifier);
-
         Ok(join_elements(
             space_token(),
-            formatter.format_nodes(ordered_nodes)?,
+            formatter.format_nodes(sort_modifiers_by_precedence(self))?,
         ))
-    }
-}
-
-fn to_sorted_modifier(modifier: &TsAnyIndexSignatureModifier) -> SortedModifiers {
-    match modifier {
-        TsAnyIndexSignatureModifier::JsStaticModifier(_) => SortedModifiers::Static,
-        TsAnyIndexSignatureModifier::TsReadonlyModifier(_) => SortedModifiers::Readonly,
     }
 }
--- a/crates/rome_formatter/src/utils/mod.rs
+++ b/crates/rome_formatter/src/utils/mod.rs
@@ -4,10 +4,10 @@ mod simple;
 use crate::formatter_traits::{FormatOptionalTokenAndNode, FormatTokenAndNode};
 use crate::{
     empty_element, format_elements, hard_group_elements, hard_line_break, space_token,
-    FormatElement, FormatResult, Formatter, ToFormatElement,
+    FormatElement, FormatResult, Formatter,
 };
 pub use call_expression::format_call_expression;
-use rslint_parser::ast::{JsAnyRoot, JsAnyStatement, JsInitializerClause};
+use rslint_parser::ast::{JsAnyRoot, JsAnyStatement, JsInitializerClause, Modifiers};
 use rslint_parser::{AstNode, AstNodeList, SyntaxNode, SyntaxNodeExt, SyntaxToken};
 
 pub(crate) use simple::*;
@@ -350,34 +350,16 @@ mod tests {
     }
 }
 
-/// Helpful data structure to make the order modifiers predictable inside the formatter
-#[derive(Debug, Ord, PartialOrd, Eq, PartialEq)]
-#[repr(u8)]
-pub(crate) enum SortedModifiers {
-    // please we wary of the order
-    Accessibility,
-    Declare,
-    Static,
-    Abstract,
-    Override,
-    Readonly,
-}
-
 /// This function consumes a list of modifiers and applies a predictable sorting.
-///
-/// The function needs to accept a closure that maps the node modifier to the correct [SortedModifier].
-pub(crate) fn sort_modifiers_by_precedence<List, Node, MapTo>(
-    list: &List,
-    map_to: MapTo,
-) -> Vec<Node>
+pub(crate) fn sort_modifiers_by_precedence<List, Node>(list: &List) -> Vec<Node>
 where
-    Node: AstNode + ToFormatElement,
+    Node: AstNode + Clone,
     List: AstNodeList<Node>,
-    MapTo: Fn(&Node) -> SortedModifiers,
+    Modifiers: for<'a> From<&'a Node>,
 {
     let mut nodes_and_modifiers = list.iter().collect::<Vec<Node>>();
 
-    nodes_and_modifiers.sort_unstable_by_key(|node| map_to(node));
+    nodes_and_modifiers.sort_unstable_by_key(|node| Modifiers::from(node));
 
     nodes_and_modifiers
 }
--- a/crates/rslint_parser/src/ast/modifier_ext.rs
+++ b/crates/rslint_parser/src/ast/modifier_ext.rs
@@ -1,0 +1,81 @@
+use crate::ast::{
+    JsAnyMethodModifier, JsAnyPropertyModifier, TsAnyIndexSignatureModifier,
+    TsAnyMethodSignatureModifier, TsAnyPropertyParameterModifier, TsAnyPropertySignatureModifier,
+};
+
+/// Helpful data structure to make the order modifiers predictable inside the formatter
+#[derive(Debug, Ord, PartialOrd, Eq, PartialEq)]
+#[repr(u8)]
+pub enum Modifiers {
+    // modifiers must be sorted by precedence.
+    Accessibility,
+    Declare,
+    Static,
+    Abstract,
+    Override,
+    Readonly,
+}
+
+impl From<&TsAnyIndexSignatureModifier> for Modifiers {
+    fn from(modifier: &TsAnyIndexSignatureModifier) -> Self {
+        match modifier {
+            TsAnyIndexSignatureModifier::JsStaticModifier(_) => Modifiers::Static,
+            TsAnyIndexSignatureModifier::TsReadonlyModifier(_) => Modifiers::Readonly,
+        }
+    }
+}
+
+impl From<&JsAnyMethodModifier> for Modifiers {
+    fn from(modifier: &JsAnyMethodModifier) -> Self {
+        match modifier {
+            JsAnyMethodModifier::JsStaticModifier(_) => Modifiers::Static,
+            JsAnyMethodModifier::TsAccessibilityModifier(_) => Modifiers::Accessibility,
+            JsAnyMethodModifier::TsOverrideModifier(_) => Modifiers::Override,
+        }
+    }
+}
+
+impl From<&TsAnyMethodSignatureModifier> for Modifiers {
+    fn from(modifier: &TsAnyMethodSignatureModifier) -> Self {
+        match modifier {
+            TsAnyMethodSignatureModifier::JsStaticModifier(_) => Modifiers::Static,
+            TsAnyMethodSignatureModifier::TsAbstractModifier(_) => Modifiers::Abstract,
+            TsAnyMethodSignatureModifier::TsAccessibilityModifier(_) => Modifiers::Accessibility,
+            TsAnyMethodSignatureModifier::TsOverrideModifier(_) => Modifiers::Override,
+        }
+    }
+}
+
+impl From<&JsAnyPropertyModifier> for Modifiers {
+    fn from(modifier: &JsAnyPropertyModifier) -> Self {
+        match modifier {
+            JsAnyPropertyModifier::JsStaticModifier(_) => Modifiers::Static,
+            JsAnyPropertyModifier::TsAccessibilityModifier(_) => Modifiers::Accessibility,
+            JsAnyPropertyModifier::TsOverrideModifier(_) => Modifiers::Override,
+            JsAnyPropertyModifier::TsReadonlyModifier(_) => Modifiers::Readonly,
+        }
+    }
+}
+
+impl From<&TsAnyPropertyParameterModifier> for Modifiers {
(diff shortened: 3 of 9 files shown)


crates/rome_formatter/src/utils/mod.rs
```diff
@@ -349,3 +349,40 @@ mod tests {
         );
     }
 }
+
+/// Helpful data structure to make the order modifiers predictable inside the formatter
```
ematipico, line 353:
```suggestion
/// Helpful data structure to make the order of modifiers predictable inside the formatter
```

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -349,3 +349,40 @@ mod tests {
...
+#[derive(Debug, Ord, PartialOrd, Eq, PartialEq)]
+#[repr(u8)]
+pub(crate) enum SortedModifiers {
+    Accessibility = 1,
+    Declare = 2,
+    Static = 3,
+    Abstract = 4,
+    Override = 5,
+    Readonly = 6,
+}
```
ematipico, line 363: I wanted to do something like:

```rs
enum B {
	A(C) = 1
}
```

But unfortunately the feature is not ready yet: https://github.com/rust-lang/rust/issues/60553

xunilrj: I don't think you need explicit discriminants in this case. The discriminant order is already used to order different variants
see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=85b7596ea78d1f3cbdc3db7fb45b67cb

```rust
#[derive(PartialEq, Eq, PartialOrd, Ord, Debug)]
enum Something {
    A(i32),
    B(i32),
    C(i32),
}

fn main() {
    let mut v = vec![Something::A(2), Something::C(0), Something::B(1)];
    v.sort();
    println!("{:?}", v);
}
```

```
[A(2), B(1), C(0)]
```

This is what the ```derive``` does:

```
impl PartialOrd for Something {
    fn partial_cmp(&self, other: &Something) -> Option<Ordering> {
    {
         let __self_vi = ::core::intrinsics::discriminant_value(&*self);
         let __arg_1_vi = ::core::intrinsics::discriminant_value(&*other);
         if true && __self_vi == __arg_1_vi {
                ...
         } else {
                ::core::cmp::PartialOrd::partial_cmp(&__self_vi, &__arg_1_vi)
          }
      }
}
```

crates/rome_formatter/src/ts/lists/index_signature_modifier_list.rs
```diff
@@ -1,12 +1,21 @@
...
         ))
     }
 }
+
+fn to_sorted_modifier(modifier: &TsAnyIndexSignatureModifier) -> SortedModifiers {
+    match modifier {
+        TsAnyIndexSignatureModifier::JsStaticModifier(_) => SortedModifiers::Static,
+        TsAnyIndexSignatureModifier::TsReadonlyModifier(_) => SortedModifiers::Readonly,
+    }
+}
```
MichaReiser, line 21: Nit: How about implementing `From<&TsAnyIndexSignatureModifier> for SortedModifiers` and then call `SortedModifiers::from(self)` or accept a `Into<SortedModifiers>`  as argument in `into_sorted_modifiers`


Or as an alternative. What about renaming `SortedModifiers` to `ClassMemberModifier` and move it into the syntax module. You can then add manual implementation on the different AST nodes and unions to get the modifier (`as_class_member_modifier()`) or implementing `From`.  Because I can imagine that this functionality can be useful in other use cases as well.

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -349,3 +349,40 @@ mod tests {
...
+
+/// This function consumes a list of modifiers and applies a predictable sorting.
+///
+/// The function needs to accept a closure that maps the node modifier to the correct [SortedModifier].
+pub(crate) fn into_sorted_modifiers<List, Node, MapTo>(list: List, map_to: MapTo) -> Vec<Node>
```
MichaReiser, line 368: Nit: How about renaming this to `sort_modifiers_by_precedence`. I find `into` confusing because it doesn't return a `SortedModifiers` but a `Vec<Node>` instead.

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -349,3 +349,40 @@ mod tests {
...
+where
+    Node: AstNode + ToFormatElement,
+    List: AstNodeList<Node>,
+    MapTo: Fn(&Node) -> SortedModifiers,
+{
+    let mut nodes_and_modifiers = list
+        .iter()
+        .map(|node| {
+            let sorted_modifier: SortedModifiers = map_to(&node);
```
MichaReiser, line 377: `map_to` is what the `Into` trait gives you

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -349,3 +349,40 @@ mod tests {
...
+            (node, sorted_modifier)
+        })
+        .collect::<Vec<(Node, SortedModifiers)>>();
+
+    nodes_and_modifiers.sort_unstable_by(|(_, left), (_, right)| left.cmp(right));
```
MichaReiser, line 382: YOu won't need all these intermediate vecs. Converting a `node` to a `SortedModifiers` is cheap, thus you can do it ad hoc (should be cheaper than allocating a new vec, sort it, then creat another vec).

```
let node_list = list.iter().collect::<Vec<_>>();

sorted.sort_unstable_by(|left, right| map_to(left).cmp(map_to(right));
sorted
```

MichaReiser (commented):
I have a few small comments regarding the implementation

However, the bigger question is if this is something the formatter should be opinionated about or not. Prettier isn't.

crates/rome_formatter/src/utils/mod.rs
leops, line 368: Nit: this could have a `list: &List` parameter instead to avoid having to clone the list

crates/rome_formatter/src/utils/mod.rs
leops, line 382: Nit: You can use `.sort_unstable_by_key(|(_, modifier)| modifier)` for this, but it probably doesn't change anything

crates/rome_formatter/src/utils/mod.rs
leops: To add on this, the derive documentation for `PartialOrd` explicitly spells out that enum variants are ordered top-to-bottom (https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable), and since Rust has very strong stability guarantees this is a behavior that can be relied upon safely

ematipico:
> I have a few small comments regarding the implementation
> 
> However, the bigger question is if this is something the formatter should be opinionated about or not. Prettier isn't.

I am not sure if I'm seeing doable, but Prettier is quite opinionated: [check their playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgCpyYw7AA6UOlOADgE4CWAbuvDgCZxga1w5gD8SHEQZQA5gG5yVdp2686TFrzaDhMUZOlUOXdDxoNmrHujbRUATxxwhI+uKkUqi471PmoV2XoMAzOw0HLWdKOgh4MHg2HHQAIxF0KJwPC2sxQM0nGXDI6JS4MzTYhI0k4gALTOCnUhhcznzUr2t4xOSiFnowHCr1LPI6hqi4GObvNrKOmC6e3Xle6sdB+toIxtGCopaS9uJOmG6fBYhGOFoGDkX+mvIAXxAAGhAIakPoTGRQfTWAdwAFfQIT4odCoX7oSyfZ5xWhJADWcBgAGV0ABbOAAGQccGQfjBmDgMLhYERKOoSWCyA0AFciSA4Gi4qMOGxMehxDT0GI4AAxCC0NEsQ7iZAgdA0mAQJ4gCowNGoADqFXo8EwFLAcGRwNVTFVljFYGwMochNoMH+cLEQrxBPpACtMAAPZHBVBwACKNPWttQhOeFNoZrFMEs1EIYAYbxlilgivobBgFWQAA4AAwBtaExVw6hiuiEc5nGUAR298EtrxB4swAFooHBRqMZTwy-QeJbuTakPi-fTCWj6NTaHTnpg3Z7y7ie3bnjM4vHE8mkAAmOdw+ioYIAYQgaO7DMwAFYZTTCXg2r7-SBGHSAJJQDiwZGR+hvACCj+RofdV7gdzuIA)

crates/rome_formatter/src/utils/mod.rs
ematipico: I wanted to use `Into` but it **consumes** the node. I don't want to consume the node. We could clone it but I'm not sure what's the difference at the end.. they do the same thing

crates/rome_formatter/src/ts/lists/index_signature_modifier_list.rs
ematipico: > Or as an alternative. What about renaming SortedModifiers to ClassMemberModifier 

About the naming, that wouldn't work. This utility is also used for the constructor modifiers. 

> Because I can imagine that this functionality can be useful in other use cases as well.

Could you name a few? I wouldn't want to make this more abstract and generic if possible

MichaReiser: > About the naming, that wouldn't work. This utility is also used for the constructor modifiers.

Constructors are also class members

> Could you name a few? I wouldn't want to make this more abstract and generic if possible

Any logic that wants to operate on all modifiers, regardless of what the member is. You want to know if a member ris abstract, iterate over all modifiers, call the `as_class_member_modifier` and test if it is abstract. That's easier than having to handle all unions explicitly.

crates/rome_formatter/src/utils/mod.rs
ematipico: Turns out, we can't use into:

```rs
    node_list.sort_unstable_by(|left, right| {
        let left: SortedModifiers = left.clone().into(); // error here*
        let right: SortedModifiers = right.clone().into(); 
        left.cmp(&right)
    });
```

* = 
```
the trait bound `SortedModifiers: From<&Node>` is not satisfied E0277 
the trait `From<&Node>` is not implemented for `SortedModifiers` 
Help: the following implementations were found: <SortedModifiers as From<JsAnyMethodModifier>> 
Note: required because of the requirements on the impl of `Into<SortedModifiers>` for `&Node`
```

I think the callback is better in this case.

MichaReiser: Did you parametrize the implementation to accept any node that implements `Into<SortedModifier`?

leops: The issue here is that the function doesn't have a `where Node: Clone` clause so doing `left.clone()` will just clone the reference and return a new `&Node`. In practice there isn't a need to call `.clone()` anyway, having a `where SortedModifiers: From<&Node>` clause (and doing `let left = SortedModifiers::from(left)`) will suffice

leops: The syntax for specifying a trait bound with references isn't very well known so I'll post an example here, it looks like this:
```rust
fn sort_nodes<Node>(list: &mut [Node])
where
    SortedModifiers: for<'a> From<&'a Node>,
{
    list.sort_unstable_by_key(|node| SortedModifiers::from(node));
}
```

crates/rome_formatter/src/ts/lists/index_signature_modifier_list.rs
ematipico: > Constructors are also class members

I meant `TsPropertyParameterModifierList`: 


From the grammar
```block

// constructor(private a: string)
//             ^^^^^^^^^^^^^^^^^
TsPropertyParameter =
	modifiers: TsPropertyParameterModifierList
	formal_parameter: JsAnyFormalParameter

TsPropertyParameterModifierList = TsAnyPropertyParameterModifier*

TsAnyPropertyParameterModifier =
	TsAccessibilityModifier
	| TsReadonlyModifier
	| TsOverrideModifier
```

They are subject to the same rules, that's why I kept the name generic.

MichaReiser: Feel free to call it `Modifier`? Properties are a bit weird. These are parameters & class members at the same time.

crates/rome_formatter/src/utils/mod.rs
ematipico: >  SortedModifiers: for<'a> From<&'a Node>

Doesn't this mean that we have to implement `impl From<SortedModifier> for &JsAnyMethodModifier`? If so, it's not what we're looking for

crates/rome_formatter/src/ts/lists/index_signature_modifier_list.rs
ematipico: I moved the implementation inside a `modifier_ext.rs` file, inside the `syntax` folder

crates/rslint_parser/src/ast/modifier_ext.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn as_sorted_modifiers(&self) -> Vec<JsAnyPropertyModifier> {
+        sort_modifiers_by_precedence(self)
+    }
+}
+
+impl TsMethodSignatureModifierList {
+    pub fn as_sorted_modifiers(&self) -> Vec<TsAnyMethodSignatureModifier> {
+        sort_modifiers_by_precedence(self)
+    }
+}
```
MichaReiser, line 74: Nit: I would just call `sort_modifiers_by_precedence(self)` in the formatter implementations.

crates/rslint_parser/src/ast/modifier_ext.rs
```diff
@@ -0,0 +1,138 @@
...
+    TsAnyPropertySignatureModifier, TsIndexSignatureModifierList, TsMethodSignatureModifierList,
+    TsPropertyParameterModifierList, TsPropertySignatureModifierList,
+};
+use crate::{AstNode, AstNodeList};
+
+/// Helpful data structure to make the order modifiers predictable inside the formatter
+#[derive(Debug, Ord, PartialOrd, Eq, PartialEq)]
+#[repr(u8)]
+pub(crate) enum Modifiers {
+    // please we wary of the order
```
MichaReiser, line 13:
```suggestion
    // Modifiers must be sorted by precedence. 
```

MichaReiser (approved):

crates/rslint_parser/src/ast/modifier_ext.rs
ematipico: This implementation is inline with how all the rest of the `_ext.rs` files do: extending the APIs of the existing nodes.

MichaReiser: That makes sense but the `sorted` is very specific and probably less useful for not formatter specific code. 

I could see a `as_modifiers` operations on the list but don't think that's that useful because you actually want to have it sorted.  Also, the `as_sorted_modifiers` doesn't do much. The re-usable part really is about converting a `*Modifier` node to `Modifier`. 

That's why I would remove the extension method here (I can see a `as_modifier()` method on the non-list nodes). It also doesn't give you much, it's just a forwarding call.

PR #2197, opened 2022-03-02
feat(rslint_error): Better diagnostics for huge lines

xunilrj:
We have some problems reporting errors on long lines. In this case a huge 4k line. This is not only aesthetically unpleasant but also poses a performance problem, given that we iterate each character (multiple times) to pretty print these diagnostics.

image

This PR limits the reported line to 80 characters and fits as many labels as we can in this space. One problem is that the code starts and ends abruptly. I tried the obvious "..." but, given they are valid JS, I thought they were more confusing than anything.

image

Another problem is when the label range itself is long.

Before
```
error[ERRORCODE]: Can't have function here
  ┌─ file.js:1:1
  │
1 │ export default function some_function_name() { return 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'; }
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ na-na
```

After
```
running 1 test
error[ERRORCODE]: Can't have function here
  ┌─ file.js:1:1
  │
1 │ export default function some_function_name() { return 'aaaaaaaaaaaaaaaaaaaaaaaaa
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ na-na
```


crates/rslint_errors/src/codespan/term/renderer.rs
```diff
@@ -594,6 +590,142 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+        outer_padding: usize,
+        line_number: usize,
+        line_range: Range<usize>,
+        source: &str,
+        severity: Severity,
+        single_labels: &[SingleLabel<'_>],
+        num_multi_labels: usize,
+        multi_labels: &[(usize, LabelStyle, MultiLabel<'_>)],
+    ) -> Result<(), Error> {
+        let max_line_length = 250;
```
xunilrj, line 666: ```250``` is arbitrary.
It does not affect our current examples and is smaller than the minified-file case.

crates/rslint_errors/src/tests.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn set_color(&mut self, _: &termcolor::ColorSpec) -> std::io::Result<()> {
+        Ok(())
+    }
+
+    fn reset(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+#[test]
```
xunilrj, line 37: There are no asserts. It is more just to more easily see results.

MichaReiser: Can't we assert against a snapshot that you take today? Also be aware, these `println` will now show up when we run our tests which can slow down test execution (IO is slow)

crates/rslint_errors/src/codespan/term/renderer.rs
```diff
@@ -594,6 +590,142 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+                line_number,
+                line_candidate,
+                severity,
+                &single_labels,
+                num_multi_labels,
+                &multi_labels,
+            );
+        } else {
+            // This implementation ignores multi_labels
+            debug_assert!(multi_labels.len() == 0);
```
MichaReiser, line 682: What are multi labels. Is this something we use in our code base? 

The issue is, the parser creating the diagnostic can't know if it will point into a long line or not.

crates/rslint_errors/src/tests.rs
MichaReiser: Here a snipped that prints to a buffer instead

https://github.com/rome/tools/blob/006603524ebb4b5d126fde1e9a2f6aa040ada997/crates/rslint_parser/src/tests.rs#L175-L183

MichaReiser (commented):
Have you looked into if there's a way to indicate that the line has been truncated?

How do other libraries solve this problem (miette, babel, TS?)

crates/rslint_errors/src/codespan/term/renderer.rs
xunilrj: Yeah. I am inclined to fall back to the old behaviour if we have ```multi_labels```. It is safer.

xunilrj:
> Have you looked into if there's a way to indicate that the line has been truncated?
> 
> How do other libraries solve this problem (miette, babel, TS?)

# Babel
Babel prints everything. And the caret point to nothing actually.

image

# Miette
miette (https://github.com/zkat/miette) readme example shows a long line trimmed: it also start and end the line abruptly.

image

# TS

ts also prints everything and point to nowhere.

image

MichaReiser:
> > Have you looked into if there's a way to indicate that the line has been truncated?
> > How do other libraries solve this problem (miette, babel, TS?)
> 
> # Babel
> 
> Babel prints everything. And the caret point to nothing actually.
> 
> image
> # Miette
> 
> miette (https://github.com/zkat/miette) readme example shows a long line trimmed: it also start and end the line abruptly.
> 
> image
> # TS
> 
> ts also prints everything and point to nowhere.
> 
> image

Thanks!

crates/rslint_errors/src/codespan/term/renderer.rs
```diff
@@ -594,6 +590,142 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+        max_line_length: usize,
+        line_number: usize,
+        line_range: Range<usize>,
+        severity: Severity,
+        single_labels: &mut Vec<(LabelStyle, Range<usize>, &str)>,
+        outer_padding: usize,
+        source: &str,
+    ) -> Result<(), Error> {
+        let labels_start = single_labels.iter().next().unwrap().1.start;
+        let labels_end = single_labels.iter().last().unwrap().1.end;
```
ematipico, line 604: Is it safe to `unwrap`? Should we have a fallback?

crates/rslint_errors/src/codespan/term/renderer.rs
```diff
@@ -594,6 +590,142 @@ impl<'writer, 'config> Renderer<'writer, 'config> {
...
+
+            // ... if not, we print one single_label per time
+            // showing only the interesting part of the line.
+            let mut candidates = vec![];
+            for single_label in single_labels.iter() {
+                candidates.push(single_label.clone());
+
+                // We need to know which part of the long line we are going to display
+                let last_end = candidates.iter().last().unwrap().1.end;
+                let first_start = candidates.iter().next().unwrap().1.start;
```
ematipico, line 692: Is it safe to `unwrap`?

crates/rslint_errors/src/tests.rs
ematipico: I also think it would be great if we were able to test the result against a snapshot

crates/rslint_errors/src/codespan/term/renderer.rs
xunilrj: Done.

crates/rslint_errors/src/codespan/term/renderer.rs
xunilrj: Done with moving this to a constant.

crates/rslint_errors/src/codespan/term/renderer.rs
xunilrj: It is, but I removed it.

xunilrj:
I decided to remove the test because it is not helping much.

MichaReiser:
> I decided to remove the test because it is not helping much.

Can we add any other test to verify the functionality? How about a parser snapshot test?

xunilrj:
> Can we add any other test to verify the functionality? How about a parser snapshot test?

Done at https://github.com/rome/tools/pull/2197/files#diff-41b338c8e2ae9f1b3118c72c23dc3535da25847fc225bd5ec30ac3fd16dbf995

ematipico:
> > Can we add any other test to verify the functionality? How about a parser snapshot test?
> 
> Done at [#2197 (files)](https://github.com/rome/tools/pull/2197/files#diff-41b338c8e2ae9f1b3118c72c23dc3535da25847fc225bd5ec30ac3fd16dbf995)

Is it the expected behaviour? (I checked the snapshot)

xunilrj:
> Is it the expected behaviour? (I checked the snapshot)

That is a good question. It is what I intended. 😄 We can check if everyone thinks it is better.

Without this PR we would underline the whole function like this:

```
error[SyntaxError]: Illegal use of an export declaration outside of a module
  ┌─ export_huge_function_in_script.js:2:1
  │
2 │ export function A () { return "Kinsmen hot Moria tea serves. Sticky camp spell covering forged they're Oakenshield vines. Admirable relatives march regained wheel Ere eternally on rest parts unhappy? Leave hundreds market's Argonath answered avail grieve doing goodness! Wrong miserable well-wishers wander stood immediately neither Agreed goat poison holes fire? Nobody tosses a Dwarf. Brigands Bilbo Baggins prisoner stinker birthday injuries. Kili's loosened shy spiders till. Gandalf's death was not in vain. Nor would he have you give up hope. Bread kindly ghost Beorn's jelly. Andûril two-faced bitterness biding seemed says drinking splendor feed light unnoticed one! Carven nearest Eärendil fireworks former. Mattress smelling wandering teaching appear taste wise Mithril uprooted winter forebearers wheel. Let's beside Proudfoots succumbed! Excuse Anárion stolen helpless nudge study shown holding form? Changes point Snowbourn material side outer highest eaves flash-flame relic descendant lurking. Thousand death Agreed oppose whole? Glóin head's hurts feasting fight shiny legacy. Thror's broken odds suffice believe well-protected? Rightfully manners begged Maggot's fairer. Unheard-of grog shields sad wondering gardener killed gone Galadriel! Pan Frodo fingers spreads magic parting amount interest idly naked. It's some form of Elvish. I can't read it. Silverwork Wraiths riddled enchantment apple anywhere."; }
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not allowed inside scripts
```

Now we are abruptly trimming the line to 250 characters. We could be checking that we are connected to a terminal and limiting the line length; but we would also need to limit somehow for non terminals.

MichaReiser (approved):

ematipico:
I think we can keep it like this for now! Ideally, the line with should come from the consumer, but I presume it should be implemented in another PR.

ematipico (approved):

PR #2199, opened 2022-03-03
fix(rslint_parser): TS type assertion assignment

MichaReiser:
## Summary

Ensures that the parser correctly parses the assignment in `(<any> x) = test` as an assignment and doesn't mark it as an invalid assignment target.

## Test Plan

Updated existing parser tests to cover the type assertion assignment.


crates/rome_formatter/src/ts/assignments/type_assertion_assignment.rs
```diff
@@ -0,0 +1,25 @@
+use crate::formatter_traits::FormatTokenAndNode;
```
ematipico, line 1: Can we create some tests for this new node, in the formatter?

PR #2202, opened 2022-03-04
ci: change version name for pre-release

ematipico:
## Summary

When releasing the VS Code extension in `nightly` mode, we change version/tag to: `pre-release-{YYYY}-{MM}-{DD}`, e.g: `pre-release-2022-03-04`


leops:
The change looks alright, and I imagine we have to merge it to test it anyway, but I wonder if we should we also inject this information into the version of the vscode package (eg. `0.0.4-prerelease.2022.03.04`) as I don't know how the marketplace will handle uploading a new bundle of the package with the same version ID but different content

ematipico:
> The change looks alright, and I imagine we have to merge it to test it anyway, but I wonder if we should we also inject this information into the version of the vscode package (eg. `0.0.4-prerelease.2022.03.04`) as I don't know how the marketplace will handle uploading a new bundle of the package with the same version ID but different content

Sure, it makes sense. It should be an easy change

PR #2205, opened 2022-03-04
refactor(rome_cli): refactor the CLI around the formatter

leops:
## Summary

Changes in this PR:
- `clap` has been replaced with `pico-args`, the only real difference this makes is that we don't have autogenerated help pages but I think we may want those to be written semi-manually and shared with the website in some form anyway
- formatting of individual files and directories: the CLI accepts any number of these as parameters and will recurse into directories ignoring unsupported files and subdirectories matching a fixed ignore list (`.git` and `node_modules` for now)
- implements `--check` mode, where the CLI will error if the output of the formatter is different from the content of the input instead of overwriting the file
- unified error handling within the CLI using `rslint_error::Diagnostic`
- alternative path for reading files using mmap, will need some research on how it impact performance on different OSes


crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+            }
+        }
+        Ok(Err(err)) => {
+            ctx.push_diagnostic(err);
+        }
+        Err(err) => {
+            let msg = match err.downcast::<String>() {
+                Ok(msg) => format!("formatting panicked: {msg}"),
+                Err(_) => String::from("formatting panicked"),
+            };
```
MichaReiser, line 319: This can either be a `String` or a `&'static str` reference, depending on what was passed to `panic`

crates/rslint_parser/src/parse.rs
```diff
@@ -64,6 +64,11 @@ impl<T> Parse<T> {
         self.errors.as_slice()
     }
 
+    /// Get the errors which occurred when parsing
```
MichaReiser, line 67: Nit: Might be better to call this `into_diagnostics`. The `errors` vec contains diagnostics with different severities

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+        Some(IndentStyle::Tab) => {
+            options.indent_style = IndentStyle::Tab;
+        }
+        Some(IndentStyle::Space(default_size)) => {
+            options.indent_style = IndentStyle::Space(size.unwrap_or(default_size));
+        }
+        None => {}
+    }
+
+    let is_check = session.args.contains("--check");
```
ematipico, line 56: There are a couple reasons why I think we should change the name the name of this options:
- we had a command called `rome check` which we will probably resurrect unless we find a better name
- we had a command called `rome ci` which was in charge of running command with a fixed set of options, like this one

As we don't have `rome ci` yet,  and I think `--check` is a bit misleading, why we don't rename it to `--ci`? It seems more explicit to me.

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+        inputs.push(input);
+    }
+
+    // At this point any remaining command line argument is unknown
+    for arg in session.args.finish() {
+        panic!("unexpected argument {arg:?}");
+    }
+
+    let (tx_files, rx_files) = channel();
+    let (tx_diags, rx_diags) = channel();
```
ematipico, line 79: What `tx` stands for? Just asking because `tx` doesn't remind me of sender

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+                continue;
+            }
+
+            ctx.push_diagnostic(Diagnostic::error(file_id, "IO", "unhandled file type"));
+        }
+    });
+
+    let duration = start.elapsed();
+    let count = formatted.load(Ordering::Relaxed);
+    println!("Formatted {count} files in {duration:?}");
```
ematipico, line 131: We should print it only when `is_silent == false`

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+use rslint_errors::{
+    file::{FileId, Files, SimpleFile},
+    Diagnostic, Emitter,
+};
+use rslint_parser::{parse, SourceType};
+
+use crate::CliSession;
+
+/// Handler for the "format" command of the Rome CLI
+pub(crate) fn format(mut session: CliSession) {
```
ematipico, line 33: What do you think of returning a `Result` from this command, we would avoid having `expect`/`panic` and handling it in the parent?

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    fn line_index(&self, id: FileId, byte_index: usize) -> Option<usize> {
+        self.storage.get(&id)?.line_index(id, byte_index)
+    }
+
+    fn line_range(&self, file_id: FileId, line_index: usize) -> Option<Range<usize>> {
+        self.storage.get(&file_id)?.line_range(file_id, line_index)
+    }
+}
```
ematipico, line 191: Can we document struct and methods please?

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+    app: &'a App,
+    options: FormatOptions,
+    is_check: bool,
+    file_ids: &'a AtomicUsize,
+    formatted: &'a AtomicUsize,
+    files: Sender<(FileId, PathBuf)>,
+    diagnostics: Sender<Diagnostic>,
+}
+
+impl<'a> TraversalContext<'a> {
```
ematipico, line 205: Can we document the methods please?

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+    app: &'a App,
+    options: FormatOptions,
+    is_check: bool,
+    path: &'a Path,
+    file_id: FileId,
+}
+
+/// This function performs the actual formatting: it reads the file from disk
+/// (or mmap it), parse and format it then either write the result back or
+/// compare it to the original content and emit a diagnostic in check mode
```
ematipico, line 336: What does `mmap` mean? It's not standard english, so we should add a link to it (if it's an acronym) or reword to make it simpler to understand.
```suggestion
/// This function performs the actual formatting: it reads the file from disk
/// (or mmap it), parse and format it; then, it either writes the result back or
/// compares it with the original content and emit a diagnostic
```

I removed "in check mode" because if we weren't in check mode, it would imply there was another scenario, which is not. It's either check mode (comparison) or not (write to file)

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+enum FileBuffer {
+    Mmap(memmap2::Mmap),
+    String(String),
+}
+
+impl FileBuffer {
+    fn read(file: &mut fs::File, allow_mmap: bool) -> io::Result<Self> {
+        // TODO: figure out on which platforms this is useful
+        if allow_mmap {
+            if let Ok(mmap) = unsafe { memmap2::Mmap::map(&*file) } {
```
ematipico, line 408: Can we add a `// SAFETY` comment that explains why it's safe to have an `unsafe` block here?

ematipico (commented):
Before merging it, I left some questions. There are few things that are not clear to me and the description of the PR doesn't address them.

For example, there are few panics left unhandled, although the description of the PR says that errors were unified with `Diagnostic`. I would like to understand at what extend this was the goal.

crates/rome_cli/src/commands/format.rs
MichaReiser: `tx` and `rx` stand for `transmit` and `receive`.

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+    {
+        inputs.push(input);
+    }
+
+    // At this point any remaining command line argument is unknown
+    for arg in session.args.finish() {
+        panic!("unexpected argument {arg:?}");
+    }
+
+    let (tx_files, rx_files) = channel();
```
MichaReiser, line 78: Nit: Would it make sense to extract the "running" of the formatting into its own function. Ideally, this would live inside of a language server that coordinates the formatting but it's fine to have this in the CLI for now.

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+    let (tx_diags, rx_diags) = channel();
+
+    let formatted = AtomicUsize::new(0);
+    let file_ids = AtomicUsize::new(0);
+
+    let ctx = TraversalContext {
+        app: &session.app,
+        options,
```
MichaReiser, line 86: Nit: Rename to `formatter_options`. It's otherwise misleading why `is_check` isn't part of `options`. Or introduce a new `FormatCommandOptions` that can be converted into `FormatOptions`?

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+        }
+    });
+
+    let duration = start.elapsed();
+    let count = formatted.load(Ordering::Relaxed);
+    println!("Formatted {count} files in {duration:?}");
+
+    let mut file_ids = HashSet::new();
+    let mut diagnostics = Vec::new();
+    while let Ok(diag) = rx_diags.recv() {
```
MichaReiser, line 135: Nit :Would it make sense if we transmit the `Diagnostic` as well as a `&str` reference to the source to avoid having to reload the diagnostics later?

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+}
+
+#[derive(Clone)]
+/// Context object shared between directory traversal tasks
+struct TraversalContext<'a> {
+    app: &'a App,
+    options: FormatOptions,
+    is_check: bool,
+    file_ids: &'a AtomicUsize,
+    formatted: &'a AtomicUsize,
```
MichaReiser, line 200: What does `file_ids` and `formatted` store that they're `usize` values? Are these the current `file_id` counter and how many files have been formatted? May help to add a bit of documentation.

MichaReiser (approved):
Nice work. Looks good to me. May be worth exploring for future PRs how some of this logic can be independent from the CLI so that it can be used using other bindings (Editor, WASM binding)

crates/rome_cli/src/lib.rs
```diff
@@ -1,74 +1,55 @@
...
+        Ok(None | Some("help")) => {
+            println!("{HELP}");
+        }
 
-            format_file_and_save(&mut file, FormatOptions::new(options), &rome_app);
+        Ok(Some(cmd)) => {
+            panic!("unknown command {cmd:?}")
+        }
+        Err(err) => {
+            panic!("failed to parse command: {err}")
```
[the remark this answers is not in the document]
leops: I probably should clarify that the errors being converted into diagnostics are the ones that are conceptually attached to a given path. So a directory traversal or file read error can be converted to a diagnostic as it is directly related to formatting a specific path, however failing to parse arguments from the command line isn't really related to any specific file (and there is no real way to recover from that anyway) so for the time being its easier to just crash the CLI immediately

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+use crate::CliSession;
+
+/// Handler for the "format" command of the Rome CLI
+pub(crate) fn format(mut session: CliSession) {
+    let mut options = FormatOptions::default();
+
+    let size = session
+        .args
+        .opt_value_from_str("--indent-size")
+        .expect("failed to parse indent-size argument");
```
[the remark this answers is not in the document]
leops: In the general case any argument can fail to parse if the input was no valid UTF-8 (I think this can happen on Windows as the command line is encoded in UTF-16). Additionally an error can be returned by the underlying implementation of `FromStr`, so for `indent-size` it would fail if the input string was not a valid `u8` (with `abc`, `-1` or `512` for instance) while for `indent-style` it would fail on any string that is not `tab` or `space`

crates/rome_cli/src/commands/format.rs
leops: The name `--check` originally came from Prettier where it does a similar thing, but I could rename it to `--ci` (or add `--ci` as an alias of `--check`)

crates/rome_cli/src/commands/format.rs
```diff
@@ -0,0 +1,468 @@
...
+        self.storage.get(&file_id)?.line_range(file_id, line_index)
+    }
+}
+
+#[derive(Clone)]
+/// Context object shared between directory traversal tasks
+struct TraversalContext<'a> {
+    app: &'a App,
+    options: FormatOptions,
+    is_check: bool,
```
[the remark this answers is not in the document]
leops: I think the `--dry-run` mode for the formatter is something we'd add along with the `review` command for previewing and accepting linter auto-fixes as they would rely on the same diffing and display logic internally

crates/rome_cli/src/commands/format.rs
leops: This bit here is tricky because it is unsafe by nature: `mmap` maps a file on disk to a segment of (virtual) memory, this allows us to access the content of the file "as if" it was an `&[u8]` directly without having to explicitly read it to a buffer first. The safety issue is that while the `memmap2` library tries as much as possible to lock the file with the operating system it is not absolutely guaranteed that the file will not actually change while its being read. And unlike a buffer being read once then living entirely within the memory of the process, any change made to the underlying file will be reflected immediately in the content of the slice returned by mmap, however as Rust assumes that the content of a `&[u8]` is immutable it can potentially lead to undefined behavior. In this case though said "undefined behavior" would probably just be the lexer or parser ending in an inconsistent state and failing or panicking, but its still somewhat undesirable and is why I would like to get rid of this path if it doesn't bring any noticeable improvement to the performance of the CLI

crates/rome_cli/src/commands/format.rs
leops: Unfortunately its not possible to directly send the `&str` between threads since its not an owned value, and the underlying buffer can be either a `String` or a `Mmap` and I would like to avoid keeping too many of those around to keep the number of open file descriptors low

crates/rome_cli/src/commands/format.rs
ematipico: Just to give you some background on how things were: rome classic has a command `rome format` but it doesn't actually save and format, it only formats the file and print the output on terminal. The real format and save, was part of the `rome check` command, which is a command that format, lint (with autofixes), dependency check, license check, etc. Although this command did more then the users expected.

Also, the reason why we didn't have a `format` that does actually saves the file, was mostly because linter and format were quite tight to each other. This is good and bad at the same time.

We can change things with this rewrite. Although rome has more awareness of how things are run and having a command `rome ci` would allow us to basically set the `is_check` to the format API.

I think we could do something like:
- `rome format --ci` launches the formatter in `--ci` mode. 
- `rome ci`, launches various checks in `--ci` mode (these checks can be formatter, linter, dependency check, audit, etc.). In the documentation we can reference to the single commands (`rome format --ci`, `rome lint --ci`, etc.)

leops: I renamed the `--check` CLI argument to `--ci`

crates/rome_cli/src/commands/help.rs
```diff
@@ -0,0 +1,38 @@
...
+USAGE:
+    rome format [OPTIONS] <INPUTS...>
+
+    INPUTS can be one or more filesystem path, each pointing to a single file or an entire directory to be searched recursively for supported files
+
+OPTIONS:
+    --ci                        Enable CI mode, lock files and exit with an error if the formatter would modify them
+    --skip-errors               Skip over files containing syntax errors instead of returning an error
+    --indent-style <tabs|space> Determine whether the formatter should use tabs or spaces for indentation (default: tabs)
+    --indent-size <u8>          If the indentation style is set to spaces, determine how many spaces should be used for indentation (default: 2)
```
ematipico, line 22: nit: `u8` is not a wide known type in the web world and we can't assume if people know it. I would just put `number` and deal with its size ourselves

ematipico (approved):
Amazing work and thank you for addressing feedback and answering questions! I just left a minor comment, not a blocker

PR #2206, opened 2022-03-04
feat(rome_formatter): format simple expressions in templates

### LINKED ISSUE #2203: Formatter: Template literals
Sometimes we should break inside template literals that contains "complex" expressions:

```js
// should not break
const bar = tag`but whral? ${foo.bar[2]} long long long long longg long long long`;

// should brek
const bar = tag`but whral? ${() => {
	let b = 3;
	return b + 3;
}} long long long long longg long long long`;

// should break
const bar = tag`but whral? ${
	(somethingBig || somethingSmall) && checkSomethingElse
} long long long long longg long long long`;

```

ematipico:
## Summary

Closes #2203 

I created a unified function to print both template literals and template literal types. 

There, I introduced the concept of "plain" expression. The main reason why I didn't include that check with `is_simple_expression` is because the checks needed template literals a different (even though some overlap, which is where I used a `simple::` function.

## Test Plan

New tests

### DIFF
--- a/crates/rome_formatter/src/utils/mod.rs
+++ b/crates/rome_formatter/src/utils/mod.rs
@@ -6,14 +6,20 @@ mod simple;
 use crate::formatter_traits::{FormatOptionalTokenAndNode, FormatTokenAndNode};
 use crate::{
     empty_element, format_elements, hard_group_elements, hard_line_break, space_token,
-    FormatElement, FormatResult, Formatter,
+    FormatElement, FormatResult, Formatter, Token,
 };
 pub use binarish_expression::format_binaryish_expression;
 pub(crate) use call_expression::format_call_expression;
 pub(crate) use format_conditional::{format_conditional, Conditional};
-use rslint_parser::ast::{JsAnyRoot, JsAnyStatement, JsInitializerClause, Modifiers};
+use rslint_parser::ast::{
+    JsAnyExpression, JsAnyFunction, JsAnyRoot, JsAnyStatement, JsInitializerClause,
+    JsTemplateElement, JsTemplateElementFields, Modifiers, TsTemplateElement,
+    TsTemplateElementFields, TsType,
+};
 use rslint_parser::{AstNode, AstNodeList, SyntaxNode, SyntaxNodeExt, SyntaxToken};
+use thiserror::private::DisplayAsDisplay;
 
+use crate::format_element::normalize_newlines;
 pub(crate) use simple::*;
 
 /// Utility function to format the separators of the nodes that belong to the unions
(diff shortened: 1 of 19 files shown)


crates/rome_formatter/tests/specs/js/module/template/template.js.snap
```diff
@@ -58,3 +81,42 @@ output
...
+
+`<div>${
+	this.set && this.set.artist
+	/* avoid console errors if `this.set` is undefined */
+}</div>`;
+
+`<div>${
+	/* avoid console errors if `this.set` is undefined */
+	this.set && this.set.artist
+}</div>`;
```
ematipico, line 121: The formatter, at the moment, is not able to understand tabs inside the template literal, that's why the trailing `}` is not aligned with opening `${`

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -367,3 +372,153 @@ where
...
+                    expression,
+                    r_curly_token,
+                    dollar_curly_token,
+                } = template_element.as_fields();
+
+                let has_comments = expression?.syntax().contains_comments()
+                    || r_curly_token?.has_leading_comments()
+                    || dollar_curly_token?.has_trailing_comments();
+
+                Ok(has_comments)
```
leops, line 507: It's probably ok to shorten this to just `template_element.syntax().contains_comments()` as I don't think its possible for the `dollar_curly_token` and `r_curly_token` to have leading or trailing trivia respectively

crates/rome_formatter/tests/specs/prettier/js/line-suffix-boundary/boundary.js.snap.new
```diff
@@ -0,0 +1,96 @@
+---
```
leops, line 1: This snapshot change hasn't been accepted

crates/rome_formatter/tests/specs/prettier/js/strings/template-literals.js.snap.new
```diff
@@ -0,0 +1,335 @@
+---
```
leops, line 1: This snapshot isn't accepted too

crates/rome_formatter/src/utils/mod.rs
ematipico: Even though your comment is logically correct, removing these checks completely messes up the formatting.
Probably there's a bug in `.contains_comments`. 

Take this snippet:

```js
${// $FlowFixMe found when converting React.createClass to ES6
ExampleStory.getFragment('story')}
`;
```

And this is the check:

```block
[crates/rome_formatter/src/utils/mod.rs:500] expression?.syntax().contains_comments() = false
[crates/rome_formatter/src/utils/mod.rs:501] r_curly_token?.has_leading_comments() = false
[crates/rome_formatter/src/utils/mod.rs:502] dollar_curly_token?.has_trailing_comments() = true
```

PR #2207, opened 2022-03-05
refactor: Move JS syntax to `rome_js_syntax`

MichaReiser:
## Summary

Moves the AST Facade and the Syntax nodes into `rome_js_syntax` because:

* Compiling the generated syntax files takes quite a bit of time but they rarely change. Moving the generated files into their own crates speeds up the development on the parser and lexer because the generated files no longer get recompiled on parser/lexer changes.
* Having the syntax in its own crate removes the dependency on the parser for crates that only traverse the CST.

Single line change compilation time for `rslint_parser`:

* Syntax in `rslint_parser` crate: 2.12s (Debug), 9.32 (Release)
* Syntax in `rome_js_syntax`: 0.49 (Debug), 3.30 (Release)

**Implications**

`rslint_lexer` now depends on the AST facade: 
* Must be recompiled for every change in the facade. This may slow down the iteration speed when making changes on the syntax.
* I don't believe that having a dependency on the facade is an architectural problem because the lexer is tightly tied to the Parser anyway and the main motivation for it being in its own crate is improved compilation time.

## Test Plan

`cargo test`

Verified the codegen commands.


MichaReiser:
!bench_parser

leops (approved):

xunilrj (approved):

ematipico:
While you're at it , I was wondering if it could make sense to rename the crate to `rome_synxtax`?

ematipico (approved):

MichaReiser:
> While you're at it , I was wondering if it could make sense to rename the crate to `rome_synxtax`?

Renamed the crate to `rome_js_syntax`

PR #2208, opened 2022-03-07
refactor(rslint_parser): Lex contextual keywords as keywords

MichaReiser:
## Summary

This PR is part of #2035. The motivation for tackling this now is that I want to reduce the `Parser`s API. Currently, the parser exposes the following API:

* `cur`: Returns kind of current token
* `cur_src`: Returns the `string` representation of the current token
* `cur_tok`: Returns the `Range` of the current token.
* `nth`/`nth_src`/`nth_tok`: Same as the `cur` pendants but for the `nth` token (lookahead).

The `cur_src` and `cur_tok` methods are necessary for diagnostics (to get the range or print the actual token that was present in the source). However, `nth` should be limited to only allow tests on the `kind` because the only use should be to do lookahead. The main use of `nth_src` is the lookahead for contextual keywords. 

This PR changes the lexer to lex contextual keywords as keywords (for example, lex `yield` as `YIELD_KW` instead of `IDENT`) so that `nth_src` and later hopefully `nth_tok` can be removed. This involves some widespread changes because

* We now need to match on the keyword kind instead of `IDENT` in all match expressions
* Replace all usages of `eat_contextual` and `expect_contextual`
* The parser now needs to remap contextual keywords to identifiers (or names) when in places where they aren't keywords. 

This change has the nice side effect that we can now emit diagnostics if:
* a keyword uses any escape sequences (not allowed): `\u0061wait 0;`
* an identifier matches the text of a keyword in an identifier position and uses escape sequences (not allowed): `var x = ({ l\u0065t }) => {};`

Part of #2035 

## Bench Results

Local benchmark results 

```
group                                 contextual                             main
-----                                 ----------                             ----
parser/checker.ts                     1.00     85.7±2.47ms    30.5 MB/sec    1.05     90.1±3.62ms    29.0 MB/sec
parser/compiler.js                    1.00     53.0±2.08ms    19.8 MB/sec    1.02     54.0±1.28ms    19.4 MB/sec
parser/d3.min.js                      1.00     33.3±1.23ms     7.9 MB/sec    1.01     33.8±0.71ms     7.8 MB/sec
parser/dojo.js                        1.00      2.6±0.05ms    26.2 MB/sec    1.01      2.6±0.04ms    25.9 MB/sec
parser/ios.d.ts                       1.00     83.4±2.55ms    22.4 MB/sec    1.06     88.5±1.98ms    21.1 MB/sec
parser/jquery.min.js                  1.00      9.4±0.41ms     8.8 MB/sec    1.01      9.5±0.15ms     8.7 MB/sec
parser/math.js                        1.00     64.0±2.13ms    10.1 MB/sec    1.01     64.8±1.36ms    10.0 MB/sec
parser/parser.ts                      1.00  1748.1±21.70µs    26.4 MB/sec    1.09  1903.4±87.88µs    24.2 MB/sec
parser/pixi.min.js                    1.00     41.5±1.67ms    10.6 MB/sec    1.01     41.8±0.92ms    10.5 MB/sec
parser/react-dom.production.min.js    1.00     12.1±0.41ms     9.5 MB/sec    1.01     12.1±0.19ms     9.5 MB/sec
parser/react.production.min.js        1.00   561.2±11.70µs    11.0 MB/sec    1.01    566.4±7.56µs    10.9 MB/sec
parser/router.ts                      1.00  1461.2±19.11µs    41.5 MB/sec    1.03  1509.9±29.41µs    40.2 MB/sec
parser/tex-chtml-full.js              1.00     87.5±2.66ms    10.4 MB/sec    1.02     89.0±1.68ms    10.2 MB/sec
parser/three.min.js                   1.00     46.3±1.39ms    12.7 MB/sec    1.01     46.9±0.82ms    12.5 MB/sec
parser/vue.global.prod.js             1.00     15.0±0.63ms     8.0 MB/sec    1.01     15.1±0.27ms     8.0 MB/sec

```

## Test Plan

Verified RAST changes. The majority of these is because the parser's now able to recover on `let` / `abstract` etc. tokens when parsing out statements. Something that wasn't possible before (would have required to add `ident` which is too detailed).


crates/rslint_parser/src/parser.rs
```diff
@@ -200,6 +200,22 @@ impl<'t> Parser<'t> {
         true
     }
 
+    /// Eats a keyword that doesn't allow unicode escape characters
+    pub fn eat_keyword(&mut self, kind: JsSyntaxKind, name: &str) -> bool {
```
MichaReiser, line 204: Ideally, keywords could use `expect` and `eat` but that will require a way to ask the `Lexer` if the last token contained a unicode escape sequence. This is something we can add once we have re-lexing of tokens in place.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1286,59 +1308,66 @@ pub(crate) fn is_nth_at_reference_identifier(p: &Parser, n: usize) -> bool {
...
-            Present(identifier)
         }
-        _ => Absent,
+        // t if t.is_non_contextual_keyword() => Some(
+        //     p.err_builder(&format!(
+        //         "Illegal use of reserved keyword `{}` as an identifier",
+        //         p.cur_src(),
+        //     ))
+        //     .primary(p.cur_tok().range(), ""),
+        // ),
```
ematipico, line 1357: To delete?

ematipico (approved):

PR #2209, opened 2022-03-07
feat(rome_formatter): verify the stability of the formatter output

leops:
## Summary

This PR adds two additional checks to the existing formatter snapshot tests:
- It verifies that source code that used to parse without error before formatting still does after it's been reformatted
- It verifies that reformatting this code a second time yields the exact same output string

It also includes a number of changes to the formatting and printing logic of some nodes to try and ensure they get formatted in a stable and reliable way

Outline of the changes:
- Adjustments to line break insertion around comments to try to ensure that leading comments remain in leading position, and trailing comments remain in trailing position (as part of this, `SyntaxTriviaPiecesIterator` now implements `ExactSizeIterator`)
- One exception is trailing comments crossing an `indent` block (due to being wrapped in a `line_suffix`): those will instead always be move to a leading position inside the indent block
- Comments are now identified as a special `FormatElement` variant that's not printed in any specific way, but allows for special formatting logic to be implemented around them: for instance calling `group_elements([comment(), token()])` will now result in the leading comment getting pushed out of the group (`[Comment(), Group(Token())]`)
- Disallow trailing separators after rest elements in function parameter and binding or assignment patterns, or in type argument lists
- Force semicolon insertion instead of dropping empty statements when they're the body of a block statement (if, for, while, labeled ...)
- Insert spaces in nested unary and pre-update expressions
- Replicate the formatting logic of parameter groups when inserting a pair or parenthesis around single identifier arrow function parameters
- Add `hard_group_element` around type alias declarations, the `id` and `eq` tokens of variable declarations, and template expressions
- Only stop the parsing of separated lists at the EOF token for list types that explicitly handle it in their implementation of `is_at_list_end`
- Fix operators of mixed precedence being dropped in binary-ish expressions
- Call chain flattening items with a trailing comment will now always close their group (and indirectly create a new one for the following elements)


crates/rslint_parser/src/parser/parse_lists.rs
```diff
@@ -132,7 +132,7 @@ pub trait ParseSeparatedList {
         let elements = self.start_list(p);
         let mut progress = ParserProgress::default();
         let mut first = true;
-        while !p.at(JsSyntaxKind::EOF) && !self.is_at_list_end(p) {
+        while !self.is_at_list_end(p) {
```
MichaReiser, line 135: I don't think this is the appropriate fix for the problem you're facing. It may even emit incorrect diagnostics, for example:

```
let a = [
```

My understanding is that this will now emit two diagnostics:

1. Expected an expression but reached the end of the file (incorrect), no expression is required
2. Expected `]` but reached the end of the file

I believe the problem at hand is that `ParseList` (and `ParseSeparatedList`) don't support lists that must contain a single element. Therefore, I recommend adding a way to customise if it should emit a Diagnostic if the list is empty. 

The problem is, this may not be straightforward because of the `ParseList` design that has no method to just generate the diagnostic. 



I'm not convinced this is the right fix. The implementation is now also no longer consistent with `ParseList` (that still uses the old behaviour). 

Have you tried what happens with an empty file. I would expect that it now creates a diagnostic saying, that it expected a statement but there was none

crates/rome_formatter/tests/specs/prettier/js/method-chain/issue-4125.js.snap
```diff
@@ -268,20 +263,21 @@ function HelloWorld() {
 action$
   .ofType(ActionTypes.SEARCHED_USERS)
   .map((action) => action.payload.query)
-  .filter((q) => !!q)
+  .filter((q) => ! !q)
```
MichaReiser, line 266: The space between the two `!` feels weird to me.

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -136,12 +138,25 @@ pub fn run(spec_input_file: &str, _expected_file: &str, test_directory: &str, fi
...
+        let has_errors = parsed.has_errors();
+        let root = parsed.syntax();
+
         let formatted_result = format(FormatOptions::default(), &root);
+
         let file_name = spec_input_file.file_name().unwrap().to_str().unwrap();
         // we ignore the error for now
         let result = formatted_result.unwrap();
 
+        if !has_errors {
```
MichaReiser, line 151: Nit: Would it be possible to assert that there are the same number of errors before/after?

crates/rome_formatter/src/utils/call_expression.rs
```diff
@@ -248,7 +251,13 @@ fn compute_groups(flatten_items: impl Iterator<Item = FlattenItem>) -> FormatRes
             }
             FlattenItem::Node(_, _) => groups.continue_group(item),
         }
+
+        // Close the group immediately if the node had any trailing comments
```
MichaReiser, line 255: why?

crates/rome_formatter/tests/specs/prettier/js/assignment/unary.js.snap
```diff
@@ -25,8 +25,8 @@ const loooooooooooooooooooooooooong3 = !looooooooooooooong.looooooooooooooong.lo
 const loooooooooooooooooooooooooong4 = !"looooooooooooooooooooooooooooooooooooooooooog";
 const loooooooooooooooooooooooooong5 = void void looooooooooooooong.looooooooooooooong.loooooong;
 const loooooooooooooooooooooooooong6 = void void "looooooooooooooooooooooooooooooooooooooooooog";
-const loooooooooooooooooooooooooong7 = !!looooooooooooooong.looooooooooooooong.loooooong;
-const loooooooooooooooooooooooooong8 = !!"looooooooooooooooooooooooooooooooooooooooooog";
+const loooooooooooooooooooooooooong7 = ! !looooooooooooooong.looooooooooooooong.loooooong;
+const loooooooooooooooooooooooooong8 = ! !"looooooooooooooooooooooooooooooooooooooooooog";
```
ematipico, line 29: This is a regression, we should not have spaces

crates/rome_formatter/tests/specs/prettier/js/empty-statement/body.js.snap
```diff
@@ -20,10 +20,10 @@ do; while(1);
...
 with (a);
 if (1) {} else if (2) {} else {}
-for (;;)
-while (1)
-for (var i in o)
-for (var i of o)
+for (;;) ;
+while (1) ;
+for (var i in o) ;
+for (var i of o) ;
```
ematipico, line 26: There's a space before `;`. Is it intentional?

crates/rslint_parser/src/parser/parse_lists.rs
leops: I wasn't really sure about this fix, which is why I didn't touch the implementation of `ParseList` as I was already surprised this worked in the first place. Interestingly even by adding "unexpected EOF" test cases for every type that implements `ParseSeparatedList` I didn't get any major change in the errors reported (the only difference is the diagnostic reporting a missing element instead of a missing closing token) but I agree a better fix is explicitly checking for empty lists in the places that don't allow them (probably in the implementation of `finish_list` for `VariableDeclaratorList` here)

crates/rome_formatter/src/format_element.rs
```diff
@@ -656,12 +675,28 @@ pub fn soft_line_indent_or_space<T: Into<FormatElement>>(content: T) -> FormatEl
 /// ```
 #[inline]
 pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
-    let content = content.into();
+    let content: FormatElement = content.into();
 
-    if content.is_empty() {
-        content
+    // If content is a list, split off the leading Comment elements
+    let (leading, content) = if let FormatElement::List(list) = content {
```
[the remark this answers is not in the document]
leops: This change is generally related to how the presence of line breaks in comment can cause groups to break unpredictably. Not including trailing trivias was an oversight as the only way for trailing comments to contain line break is to write a multiline comment (`type T = A | B /*\n*/`) so there are only very few cases that can actually impact the formatting of groups but they do exist.

Overall the intention was for the new `comment` element to be mostly an internal detail: it's only really supposed to be emitted by the formatter itself when processing trivia pieces, and this "comment extraction" logic exists specifically so that individual `ToFormatElement` implementations don't have to know or apply special care about potential comments in their groups.

crates/rome_formatter/tests/spec_test.rs
leops: This is something we should be checking for in the future, but the error recovery behavior of the formatter is not stable enough yet

crates/rome_formatter/src/utils/binarish_expression.rs
```diff
@@ -328,11 +328,16 @@ fn split_binaryish_to_flatten_items(
     let right_kind = &right.syntax().kind();
     let right_expression_should_group = right_expression_should_group(right_kind);
 
-    let formatted_left = concat_elements([
+    let formatted_left = format_elements![
         left.format(formatter)?,
         space_token(),
         operator.format(formatter)?,
-    ]);
+        if operator.has_trailing_comments() {
```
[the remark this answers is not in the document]
leops: Unfortunately I think by the time `FlattenItems::into_format_element` is run some of the information needed for this logic to happen isn't accessible anymore

crates/rome_formatter/tests/specs/prettier/js/method-chain/issue-4125.js.snap
leops: I changed the behavior from inserting a space to parenthesizing the inner expression (this is what Prettier is doing) and reduced the logic that trigger this path to just the cases that might be ambiguous to parse (like `a + ++b` or `a - - - b`)

crates/rome_formatter/tests/spec_test.rs
```diff
@@ -169,3 +194,35 @@ pub fn run(spec_input_file: &str, _expected_file: &str, test_directory: &str, fi
...
+
+        panic!(
+            "formatter output had error diagnostics where input had none:\n{}",
+            std::str::from_utf8(buffer.as_slice()).expect("non utf8 in error buffer")
+        )
+    }
+
+    let re_format = format(format_options, &re_parse.syntax()).unwrap();
+
+    similar_asserts::assert_str_eq!(text, re_format.as_code());
```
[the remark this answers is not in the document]
leops: Importing modules in integration tests is unfortunately convoluted so I had put it off until now, but I managed to work around it and have the `check_reformat` function moved to its own module used by the other tests

crates/rome_formatter/src/format_element.rs
```diff
@@ -1196,10 +1219,44 @@ impl FormatElement {
...
             }
             FormatElement::Token(token) => token.contains('\n'),
-            FormatElement::LineSuffix(suffix) => suffix.has_hard_line_breaks(),
+            FormatElement::LineSuffix(_) => true,
+            FormatElement::Comment(content) => content.has_hard_line_breaks(),
             FormatElement::Verbatim(verbatim) => verbatim.element.has_hard_line_breaks(),
         }
     }
+
+    pub fn split_trivia(self) -> (FormatElement, FormatElement, FormatElement) {
```
ematipico, line 1228: Amazing! This API will be ton useful! It would be great if we could add some documentation, especially on the returned tuple, just to mention what we have in each position.

ematipico (approved):

crates/rome_formatter/src/js/expressions/unary_expression.rs
```diff
@@ -1,30 +1,76 @@
...
+            ]);
+        }
+
+        // Parenthesize the inner expression if it's a binary or pre-update
+        // operation with an ambiguous operator (+ and ++ or - and --)
+        let is_ambiguous_expression = match &argument {
+            JsAnyExpression::JsUnaryExpression(expr) => {
+                let inner_op = expr.operator()?;
+                matches!(
+                    (operator.kind(), inner_op.kind()),
```
ematipico, line 36: Let's use `operation` API, so we can match against `JsUnaryOperation::Plus`, etc.

leops: As part of this I added a similar `operation()` method to `JsPreUpdateExpression`

PR #2213, opened 2022-03-08
feat(rome_formatter): omit or add parenthesis 

### LINKED ISSUE #2204: Formatter: Apply parenthesis
Sometimes, should apply parenthesis to some statements, to help the readability of the code.

Inside a return that has a statement that breaks
```js
// before
function f() {
	return  something_long_looooooooooooong +  something_long_looooooooooooong
}

// after
function f() {
	return  ( 
		something_long_looooooooooooong +  something_long_looooooooooooong
	)
} 
```

Inside conditional expressions that have nested conditionals:
```js
// before
a  && b && c || d && c && f

// after
(a && b && c) || (d && c && f);
```

Arrow functions with single parameter:

```js
// before
const f = a => true

// after
const f = (a) => true
```

ematipico:
## Summary

Closes #2204 

This PR adds some optimizations and compatibility traits around the application of parenthesis around certain expressions.

In particular:
- apply parenthesis inside nested logical expressions that have difference precedence;
- apply parenthesis to sequence expressions inside a `return` statement
- omit parenthesis inside "simple" expressions  

## Test Plan

New test cases and checked the existing ones


crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js.snap
```diff
@@ -214,46 +222,51 @@ const a = x + y;
...
-		something &&
-		elsewhere &&
-		happy &&
-		thoughts &&
-		somethingsomethingsomethingsomething;
-
-(
-	lorem && ipsum // foo
-);
+	(
```
ematipico, line 237: At the moment prettier adds parenthesis but doesn't group the expressions inside it, result in something like:

```js
const a =
  (aa &&
    bb &&
    something &&
    elsewhere &&
    happy &&
    thoughts &&
    somethingsomethingsomethingsomething) ||
  (aa &&
    bb &&
    something &&
    elsewhere &&
    happy &&
    thoughts &&
    somethingsomethingsomethingsomething);

```

Personally, I don't like it and I think that grouping the expression is nicer to see

crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js.snap
```diff
@@ -214,46 +222,51 @@ const a = x + y;
...
 (
-	lorem && call_function(1, 3) // foo
-);
-
-(
-	lorem &&
-		// foo
-		(3 + 6 == 9)
-);
+	// Implicitly parenthesized && and || requires parens
```
ematipico, line 269: We don't have the infrastructure to fix this, yet. #2209 will set the foundation of fixing these cases

Issue #2183, opened 2022-03-01
☂️ Content changes before release


ematipico:
### Description

- [x] Update `README.md` with the current status of the project
- [x] #2184
- [x] #2279
- [x] Document formatter differences against most popular formatters (prettier, dprint)
- [x] #2280
- [x] Getting started guide with our VSCode extension

PR #2215, opened 2022-03-08
feature(playground): Adding playground to tools repo

NicholasLYang:
## Summary
Part of #2183

- Added playground from [rome/playground](http://github.com/rome/playground) in rome_playground
- Added boilerplate to deploy playground as Cloudflare Worker in `workers-site`
- Added workflow to deploy playground


Cargo.toml
```diff
@@ -4,6 +4,7 @@
 resolver = "2"
 members = [
 	"crates/*",
+	"playground",
```
xunilrj, line 7: Any reason to not have this inside crates?

playground/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "rome-playground"
```
xunilrj, line 2: Maybe we could match folder and package name?

playground/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "name": "playground",
+  "version": "0.0.0",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@tailwindcss/forms": "^0.4.0",
+    "@types/prettier": "^2.4.3",
```
ematipico, line 11: This should be a dev dependency

playground/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "react-tabs": "^3.2.3"
+  },
+  "devDependencies": {
+    "@types/react": "^17.0.33",
+    "@types/react-dom": "^17.0.10",
+    "@types/react-tabs": "^2.3.4",
+    "@vitejs/plugin-react": "^1.0.7",
+    "autoprefixer": "^10.4.2",
+    "postcss": "^8.4.6",
+    "tailwindcss": "^3.0.19",
```
ematipico, line 25: Isn't `tailwindcss` supposed to be a direct dependency? 🤔 If yes, it should be moved. If not, then also `@tailwindcss/forms` should be a dev dependency

playground/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+        Ok(())
+    }
+}
+
+#[wasm_bindgen]
+pub fn run(
+    code: String,
+    line_width: u16,
+    is_tab: bool,
+    indent_width: u8,
```
ematipico, line 79: This should be an `Option<u8>`.  If we choose tabs, we aren't supposed to have this value

playground/src/LineWidthInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+  lineWidth: number;
+  setLineWidth: (lineWidth: number) => void;
+}
+
+export default function LineWidthInput({ lineWidth, setLineWidth }: Props) {
+  return (
+    <div className="w-[300px] p-5 flex items-end">
+      <div className="pr-4">
+        <label
+          htmlFor="number"
```
ematipico, line 11: This should be `lineWidth`

playground/src/IndentStyleSelect.tsx
```diff
@@ -0,0 +1,54 @@
...
+export default function IndentStyleSelect({
+  indentStyle,
+  setIndentStyle,
+  indentWidth,
+  setIndentWidth,
+}: Props) {
+  return (
+    <div className="pl-5 flex">
+      <div>
+        <label className="block text-sm font-medium text-gray-700">
```
ematipico, line 19: `htmlFor` is missing

playground/src/IndentStyleSelect.tsx
```diff
@@ -0,0 +1,54 @@
...
+          Indent Type
+        </label>
+        <select
+          id="location"
```
ematipico, line 23: `name` is missing

playground/src/LineWidthInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+      >
+        80
+      </button>
+      <button
+        onClick={() => setLineWidth(120)}
+        disabled={lineWidth === 120}
+        className="bg-slate-500 m-2 text-sm w-[80px] p-1 rounded text-slate-50 disabled:bg-slate-300 transition"
+      >
+        120
+      </button>
```
ematipico, line 42: Let's have an aria label on these buttons. Having just numbers on these buttons doesn't help to understand what they do

Cargo.toml
ematipico: I agree too, I think we should move the this crate under the folder `crates/`

playground/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
```
ematipico, line 3: As internal standard, crates that are not published should have `0.0.0` as version

playground/src/App.css
```diff
@@ -0,0 +1,42 @@
...
+
+@media (prefers-reduced-motion: no-preference) {
+  .App-logo {
+    animation: App-logo-spin infinite 20s linear;
+  }
+}
+
+.App-header {
+  background-color: #282c34;
+  min-height: 100vh;
```
ematipico, line 18: Why is the header so high?

playground/src/App.css
```diff
@@ -0,0 +1,42 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  font-size: calc(10px + 2vmin);
```
ematipico, line 23: We should really use `rem` for font-size

playground/src/App.css
```diff
@@ -0,0 +1,42 @@
...
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
+
+button {
+  font-size: calc(10px + 2vmin);
```
ematipico, line 41: Let's use `rem`

ematipico (changes requested):
Overall it looks good to me. One important thing that I would do before merging it, would be to move the playground crate inside the `crates/` directory and change its version to `0.0.0`

playground/package.json
NicholasLYang: Let's remove this entirely. Gonna try to use Rome now

NicholasLYang:
Ahh, we might have to cancel this PR because we can't have multiple sites on a single repo for Cloudflare Pages

ematipico:
> Ahh, we might have to cancel this PR because we can't have multiple sites on a single repo for Cloudflare Pages

We can still have the playground web app in a different folder (not `website/`) and do a ad-hod deployment

NicholasLYang:
Yeah I think I have a solution. We can just use an action to manually publish to a Cloudflare Worker site. Not too hard.

ematipico:
Could you please update the template of the PR and add some description of that we're trying to achieve? It seems that we are now doing more than just adding the playground to the main repository.
Also, could you please link this PR to the relative umbrella issue?

One question: is there a reason why the web app and the crate are together? Can they be separated?

NicholasLYang:
@ematipico the crate and the web app are connected because the crate is just a tiny wrapper around a few Rome APIs. It's really not worth separating in my opinion. We gain more by colocating it with the web app code than we do by having it separate. We could put the JS in a different folder from the Rust within the same crate I suppose

ematipico (approved):
Let's merge it and give it a shot! :)

PR #2216, opened 2022-03-08
fix(lsp): minor logic bug with formatting with errors setting

NicholasLYang:

### DIFF
--- a/crates/rome_lsp/src/handlers/formatting.rs
+++ b/crates/rome_lsp/src/handlers/formatting.rs
@@ -79,7 +79,7 @@ pub(crate) fn format_range(params: FormatRangeParams) -> Result<Option<Vec<TextE
         .workspace_settings
         .formatter
         .format_with_syntax_errors
-        || !parse_result.has_errors()
+        && !parse_result.has_errors()
     {
         return Ok(None);
     }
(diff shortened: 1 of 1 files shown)


yassere:
Looks like there's a similar mistake in the `format_range` and `format_on_type` functions too.

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) fn format_range(params: FormatRangeParams) -> Result<Option<Vec<TextE
         .workspace_settings
         .formatter
         .format_with_syntax_errors
-        || !parse_result.has_errors()
+        && !parse_result.has_errors()
```
yassere, line 82: This would mean we return early if "format_with_syntax_errors" is enabled but there aren't any syntax errors. I assume we only want to bail for the same condition you changed the `format` handler to:
`!workspace_settings.formatter.format_with_syntax_errors && parse_result.has_errors()`

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -146,7 +146,7 @@ pub(crate) fn format_on_type(params: FormatOnTypeParams) -> Result<Option<Vec<Te
...
```
yassere, line 149: Same here. I don't think the condition here should be any different than what you changed the `format` handler to.

yassere (changes requested):

ematipico (approved):

ematipico (approved):

PR #2217, opened 2022-03-09
ci: update `package.json` for pre release

ematipico:
## Summary

This PR does two things:

- in case of prerelease, we update the `"version"` field of the `package.json`. It seems that while publishing the extension, the the tool read the version from the `package.json`;
- updates how we construct the version number in case of prerelease, which is now something like this: `pre-release-2022-03-09-1646822342`
	1. `pre-release` is always the same
	2. the current date
	3. the current milliseconds from `1970-01-01 00:00:00`, this would allow use to push multiple prerelease at the same date

## Test Plan

To merge it and kick a prerelease

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,9 +40,12 @@ jobs:
       - name: Check prerelease status
         id: prerelease
         if: env.nightly == 'true' || steps.version.outputs.type == 'prerelease' || steps.version.outputs.type == 'prepatch' || steps.version.outputs.type == 'premajor' || steps.version.outputs.type == 'preminor'
+        working-directory: editors/vscode
         run: |
-          echo "prerelease=true" >> $GITHUB_ENV
-          echo "version=$(date +"pre-release-%F")" >> $GITHUB_ENV
+          CURRENT_VERSION=$(cat package.json | grep version |  awk '{print $2}' | sed 's/"//g' | sed 's/,//g')
+          NEW_VERSION=${{ steps.version.outputs.version }}-$(date +"pre-release-%F-%s")
+          cat package.json | sed "s/"${CURRENT_VERSION}"/"${NEW_VERSION}"/g" >  temp
+          mv temp package.json
 
       - name: Check version status
         if: steps.version.outputs.changed == 'true'


.github/workflows/release.yml
```diff
@@ -40,9 +40,12 @@ jobs:
       - name: Check prerelease status
         id: prerelease
         if: env.nightly == 'true' || steps.version.outputs.type == 'prerelease' || steps.version.outputs.type == 'prepatch' || steps.version.outputs.type == 'premajor' || steps.version.outputs.type == 'preminor'
+        working-directory: editors/vscode
         run: |
-          echo "prerelease=true" >> $GITHUB_ENV
-          echo "version=$(date +"pre-release-%F")" >> $GITHUB_ENV
+          CURRENT_VERSION=$(cat package.json | grep version |  awk '{print $2}' | sed 's/"//g' | sed 's/,//g')
+          NEW_VERSION=${{ steps.version.outputs.version }}-$(date +"pre-release-%F-%s")
+          cat package.json | sed "s/"${CURRENT_VERSION}"/"${NEW_VERSION}"/g" >  temp
```
leops, line 47: Wouldn't that also override the version number of dependencies ? For instance if the current version were `2.6.7` it looks like this would override the version of `vsce` too

leops (commented):
As an alternative to printing out the full timestamp, we could also use the short hash of the commit being being deployed

.github/workflows/release.yml
ematipico: Good catch!

ematipico: Applied the change. I tested locally by simulating the use case you mentioned and it works as expected.

ematipico:
> we could also use the short hash of the commit being being deployed

Done, I used [`github.sha`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

ematipico:
> > we could also use the short hash of the commit being being deployed
> 
> Done, I used [`github.sha`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

I reverted this change. I wouldn't work for workflows that are manually dispatched. For now I would like to keep the timestamp. Fetching the "latest" commit is something that also changes depending from which branch the workflow is triggered.

leops (approved):

leops:
> > > we could also use the short hash of the commit being being deployed
> > 
> > 
> > Done, I used [`github.sha`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)
> 
> I reverted this change. I wouldn't work for workflows that are manually dispatched. For now I would like to keep the timestamp. Fetching the "latest" commit is something that also changes depending from which branch the workflow is triggered.

Interesting I didn't think it was possible for the workflow to be dispatched without a commit. But yeah it doesn't make much of a difference, the timestamp works just as well

ematipico:
> > > > we could also use the short hash of the commit being being deployed
> > > 
> > > 
> > > Done, I used [`github.sha`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)
> > 
> > I reverted this change. I wouldn't work for workflows that are manually dispatched. For now I would like to keep the timestamp. Fetching the "latest" commit is something that also changes depending from which branch the workflow is triggered.
> 
> Interesting I didn't think it was possible for the workflow to be dispatched without a commit. But yeah it doesn't make much of a difference, the timestamp works just as well

Yeah, the thing that made rever this part is that `github.sha` is the commit that triggers the workflow. Although, `workflow_dispatch` allows to manually trigger the workflow and to choose the branch too. So I thought that the commit, using this env variable, might not be present in these cases. Which means that we have to manually retrieve the latest commit of the branch. Too much hassle for now 😅

PR #2218, opened 2022-03-09
feat(rome_formatter): omit or add parenthesis 

### LINKED ISSUE #2204: Formatter: Apply parenthesis

ematipico:
## Summary

Closes #2204 

This PR adds some optimizations and compatibility traits around the application of parenthesis around certain expressions.

In particular:
- apply parenthesis inside nested logical expressions that have difference precedence;
- apply parenthesis to sequence expressions inside a `return` statement
- omit parenthesis inside "simple" expressions  

## Test Plan

New test cases and checked the existing ones


crates/rome_formatter/tests/specs/prettier/js/optional-chaining/chaining.js.snap
```diff
@@ -164,13 +164,13 @@ a?.[b ? c : d];
...
 let f = () => ({}?.());
 let g = () => ({}?.b);
-a = () => ({}?.() && a);
-a = () => ({}?.()() && a);
-a = () => ({}?.().b && a);
-a = () => ({}?.b && a);
-a = () => ({}?.b() && a);
-(a) => ({}?.()?.b && 0);
-(a) => ({}?.b?.b && 0);
+a = () => {}?.() && a;
```
leops, line 167: The parenthesis should not be removed here as it creates an ambiguity for the parser (`() => {}` can be interpreted as a function with an empty body)

crates/rome_formatter/tests/specs/prettier/js/optional-chaining/chaining.js.snap
```diff
@@ -119,7 +119,7 @@ a?.b[3].c?.(x).d.e?.f[3].g?.(y).h;
 (a ? b : c)?.d;
 
 (list || list2)?.length;
-(list || list2)?.[(list || list2)];
+list || list2?.[list || list2];
```
leops, line 122: This doesn't look correct, the `||` operation has a higher precedence than `?.[]` so it will get parsed back as `list || (list2?.[list || list2])`

crates/rome_formatter/src/utils/binarish_expression.rs
```diff
@@ -377,33 +413,109 @@ fn split_binaryish_to_flatten_items(
...
+        None
+    };
+
+    let operation_is_higher = if let Some(compare_to) = compare_to {
+        match (previous_operation, compare_to) {
+            (Operation::Logical(previous_operation), Operation::Logical(compare_to)) => {
+                compare_to > previous_operation
+            }
+
+            // Prettier, for some reason, decided that binary expressions with different precedence don't need parenthesis.
```
leops, line 479: If this something we really want to be compatible with, or could we choose to parenthesize those expressions anyway and disregard compatibility for this ?
Also Prettier does seem to insert parenthesis for some of those at least, like `1 << 2 + 3` becoming `1 << (2 + 3)`

crates/rome_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -48,3 +67,76 @@ fn is_simple_parenthesized_expression(node: &JsParenthesizedExpression) -> Synta
 
     Ok(true)
 }
+
+fn parenthesis_can_be_omitted(node: &JsParenthesizedExpression) -> SyntaxResult<bool> {
+    let expression = node.expression()?;
+    let parent = node.syntax().parent();
+
+    if let Some(parent) = parent {
+        if matches!(
```
leops, line 76: I wonder if this could instead be implemented by having a generic "precedence" value associated with certain nodes so we could simply compare the precedence of the parent with the one of the parenthesized expression instead of having to exhaustively match a huge set of node kinds

crates/rome_formatter/src/utils/binarish_expression.rs
ematipico: > If this something we really want to be compatible with, or could we choose to parenthesize those expressions anyway and disregard compatibility for this ?

I really don't mind discard the compatibility and add the parenthesis based on the precedence. It would make things a bit consistent across the board ( why `1 + 4 * 3` is formatted with no parenthesis while `1 << 2 + 3` yes?)

crates/rome_formatter/src/js/expressions/parenthesized_expression.rs
ematipico: That's a nice a idea, I will play around with it

crates/rome_formatter/src/utils/binarish_expression.rs
ematipico: Done

crates/rome_formatter/src/js/expressions/parenthesized_expression.rs
ematipico: Done

crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js.snap
```diff
@@ -255,6 +266,15 @@ const a =
 (
 	lorem &&
 		// foo
-		(3 + 6 == 9)
+		((3 + 6) == 9)
```
leops, line 269: The parenthesis aren't needed here, the binary expression has a higher precedence over the logical expression and would get parsed correctly without the parenthesis, and this seems like a common enough pattern (eg. `if((i + 1) === length)`) that we probably don't want to rewrite it "too eagerly"

crates/rome_formatter/tests/specs/prettier/js/binary_math/parens.js.snap
```diff
@@ -33,25 +33,25 @@ const result = (a + b) >>> 1;
...
+if (rawIndex < (1 << (list._level + SHIFT))) {
 }
 var res = size < SIZE ? 0 : (((size - 1) >>> SHIFT) << SHIFT);
 sign = 1 - (2 * (b[3] >> 7));
 exponent = (((b[3] << 1) & 0xff) | (b[2] >> 7)) - 127;
 mantissa = ((b[2] & 0x7f) << 16) | (b[1] << 8) | b[0];
 
-2 / 3 * 10 / 2 + 2;
-const rotateX = ((RANGE / rect.height) * refY - RANGE / 2) * getXMultiplication(
+((2 / 3) * 10 / 2) + 2;
```
leops, line 43: This seems oddly arbitrary, multiplication and division operations have the same precedence but only some divisions are getting parenthesized

crates/rome_formatter/tests/specs/prettier/js/babel-plugins/bigint.js.snap
```diff
@@ -242,9 +242,11 @@ view[0];
 //   ^ negative because of overflow
 
 1n + 2;
-// ↪ TypeError: Cannot mix BigInt and other types, use explicit conversions
+(
+  // ↪ TypeError: Cannot mix BigInt and other types, use explicit conversions
```
leops, line 246: Could `split_trivia` be used here to move the comment out of the parenthesis group (I imagine the empty line should be moved as well but that's not implemented yet) ?

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -70,7 +70,7 @@ impl JsLiteralMemberName {
 }
 
 /// A binary operation applied to two expressions
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
```
leops, line 73: Maybe the documentation comment (or a normal comment) should explain the ordering logic for these enums, to clearly state that the variants are to be declared in a specific order

crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js.snap
ematipico: In ALL cases updated in this PR, the parenthesis are not needed, because they work on correct code. For example:
`a && b || c` and `(a && b) || c` are the same, the parenthesis are not needed.

That's why I wanted to understand if we wanted to be consistent or not **across the board**.

crates/rome_formatter/tests/specs/prettier/js/binary_math/parens.js.snap
ematipico: This is the result of how our parser represents the expression, for `2 / 3 * 10 / 2`

```block
JsModule {
    interpreter_token: missing (optional),
    directives: JsDirectiveList [],
    items: JsModuleItemList [
        JsExpressionStatement {
            expression: JsBinaryExpression {
                left: JsBinaryExpression {
                    left: JsBinaryExpression {
                        left: JsNumberLiteralExpression {
                            value_token: JS_NUMBER_LITERAL@0..2 "2" [] [Whitespace(" ")],
                        },
                        operator: SLASH@2..4 "/" [] [Whitespace(" ")],
                        right: JsNumberLiteralExpression {
                            value_token: JS_NUMBER_LITERAL@4..6 "3" [] [Whitespace(" ")],
                        },
                    },
                    operator: STAR@6..8 "*" [] [Whitespace(" ")],
                    right: JsNumberLiteralExpression {
                        value_token: JS_NUMBER_LITERAL@8..11 "10" [] [Whitespace(" ")],
                    },
                },
                operator: SLASH@11..13 "/" [] [Whitespace(" ")],
                right: JsNumberLiteralExpression {
                    value_token: JS_NUMBER_LITERAL@13..14 "2" [] [],
                },
            },
            semicolon_token: missing (optional),
        },
    ],
    eof_token: EOF@14..14 "" [] [],
}
```

[Link to playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&indentWidth=2&typescript=false&jsx=false#MiAvIDMgKiAxMCAvIDI=)

crates/rome_formatter/tests/specs/prettier/js/babel-plugins/bigint.js.snap
ematipico: Yes, definitely. I thought it has a private visibility. I will update the PR!

crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js.snap
leops: As I understand it, the solution Prettier adopted seems mostly empirical: it expects that readers know the precedence of common mathematical operations like multiplication or addition and those don't need to be parenthesized, but for less used operators (bit-shift) or ones whose precedence may not be obvious (or vs. and) these will be parenthesized. In this case, this is a common enough pattern that Prettier seems to consider the relationship between the addition and equality operators to be "well-known", and the expression is not parenthesized

ematipico: Exactly that. I think this goes against the project philosophy, where it says that we should not assume that people know stuff in advance. Prettier approach relies on the fact that people already knows the precedence of some operators and this creates inconsistent formatting (e.g. why some operations are parenthesized while others are not?)

crates/rome_js_syntax/src/expr_ext.rs
ematipico: Yeah, this is greatly wrong. Unfortunately `Ord` and `PartialOrd` don't fit our cases, because we have operators here that have the same precedence, so we can't rely on the default implementation.

We will have to manually implement those traits.

crates/rome_formatter/tests/specs/prettier/js/binary_math/parens.js.snap
ematipico: The issue was also due to an incorrect implementation of the precedence of the binary operations. Now it should be fine

ematipico:
While implementing the ordering of the operations, I found out that `instaceof` and `in` are not parsed as `JsBinaryExpressions`. I will fix that part in a subsequent PR, otherwise I risk to bloat the goal of this current PR. In the meanwhile, I removed the variants that are impossible.

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -112,10 +115,128 @@ pub enum JsBinaryOperation {
...
+            3
+        } else if self.is_comparison_operator() {
+            2
+        } else {
+            1
+        }
+    }
+}
+
+impl PartialOrd for JsBinaryOperation {
```
leops, line 179: It would probably make more sense to implement an explicit `fn compare_precedence(&self, other: &Self) -> std::cmp::Ordering` method on `JsBinaryOperation`, since this implementation `Plus > Minus`, `Plus < Minus` and `Plus == Minus` all return `false` which isn't correct

ematipico: I didn't quite understand your comment. I implemented the suggested function but I don't see any relevant changes unless there was a specific case not covered, which it would be better to test... but I don't know which one it is. What was the main concern?

leops: Oh right, my concern was that `JsBinaryOperation` should not actually implement `Ord` and `PartialOrd` at all and the comparison should be performed by using the `compare_precedence` method instead to make it explicit that the precedence of the operators is what's being compared instead of comparing the operators directly, and also because the implementation of these traits are expected to agree with the implementation of `Eq` and `PartialEq`

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -112,10 +115,135 @@ pub enum JsBinaryOperation {
...
+        } else {
+            1
+        }
+    }
+
+    pub fn compare_precedence(&self, other: &Self) -> Ordering {
+        let self_precedence = self.get_precedence();
+        let other_precedence = other.get_precedence();
+
+        #[allow(clippy::comparison_chain)]
```
leops, line 182: The reason Clippy is complaining here is because this could be written as `self_precedence.cmp(&other_precedence)`

Issue #2153, opened 2022-02-21
☂️ JSX Parsing Support


xunilrj:
See specification at: https://facebook.github.io/jsx/

# Lazy Lexer

- [x] Is it possible to implement JSX parsing with the lexer as it is?
    - If, yes. Do we want to implement the parser as fast as possible and use "relexing"?
    - If, not. We need to make the lexer be driven by the parser (lazy lexing)
        - [x] Make ```TokenSource``` to be lazy. #2219 
        - [x] ```Parser``` will have new methods to drive the lexer. 

see:
https://github.com/rome/tools/issues/2035

# JSX Parser

- [ ] Implement JSX parser 
    - Where we can expect JSX?
    - Is it when "is_object_expression_allowed" is true?
    - [x] JSXElement (open/close and self closing)
    - [x] Fragments #2243
    - [x] Add JSX to function lookahead #2240 
    - [x] Constant attributes #2224 
    - [x] Expression attributes #2224 
    - [x] Spread attributes #2242 
    - [x] Hierarchical JSXElement #2229 
    - [x] Expression inside JSXElement #2229 
    - [x] JSXText #2237 
    - [x] JsxStringLiteral (doesn't allow escape sequences) #2241 
    - [x] Spread content
    - [x] Namespaces #2231 
    - [x] ASI
    - [x] Support trivia betweens tokens "< div > < div / > < > < / > < / div >" #2233 
    - [x] TS type arguments for elements <GenericComponent<string>> (call to parse_ts_type_arguments for TS only)
    - [ ] Correctly deal with jsx and type assertions" [discussion](https://github.com/rome/tools/pull/2224#discussion_r827696340)
    - [ ] Mature the checkpoint/rewind and node abandoning when parsing JSX [discussion](https://github.com/rome/tools/pull/2224#discussion_r826656207)
    - [ ] Error Handling
      - Add diagnostic for sequence expression in JSX attribute
      - Opening / closing element name mismatch

# Testing

- [ ] Test Suites    
    - [x] From Microsoft/Typescript test suite:
        - https://github.com/microsoft/TypeScript
        - [x] doubleUnderscoreReactNamespace
        - [x] jsxFactoryIdentifierAsParameter
        - [x] jsxFactoryQualifiedNameWithEs5
        - [x] jsxPreserveWithJsInput
        - [x] jsxEmitWithAttributes
        - [x] jsxFactoryIdentifier
        - [x] jsxFactoryQualifiedName
    
[...]

PR #2224, opened 2022-03-11
Feature/jsx simple attributes

xunilrj:
see https://github.com/rome/tools/issues/2153

Support for attributes

```jsx
function f() {
     return <div string_literal="a" expression={1} novalue el=<a/>></div>;
}
```


crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +173,66 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+    ParsedSyntax::Present(m.complete(p, JsSyntaxKind::JSX_ATTRIBUTE_INITIALIZER_CLAUSE))
+}
+
+fn expect_jsx_attribute_value(p: &mut Parser) -> ParsedSyntax {
+    let m = p.start();
+
+    p.bump(JsSyntaxKind::JS_STRING_LITERAL);
+
+    ParsedSyntax::Present(m.complete(p, JsSyntaxKind::JSX_STRING_LITERAL))
+}
```
MichaReiser, line 238: I would expect that any `expect_` method would add an error if the node isn't present. 

Why not call this `parse_` and return `Absent` if they're missing. For example, the initializer is optional.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +173,66 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+
+    fn list_kind() -> JsSyntaxKind {
+        JsSyntaxKind::JSX_ATTRIBUTE_LIST
+    }
+}
+
+fn expect_jsx_attribute_name(p: &mut Parser) -> ParsedSyntax {
+    let m = p.start();
+
+    p.bump(JsSyntaxKind::IDENT);
```
MichaReiser, line 218: Nit: The majority of the code uses `T![ident]` for tokens and the `JsSyntaxKind::*` names only for nodes.

crates/rslint_parser/src/syntax/expr.rs
```diff
@@ -1742,10 +1742,9 @@ pub(super) fn parse_unary_expr(p: &mut Parser, context: ExpressionContext) -> Pa
 
     // if we are at "<"; or we have JSX or Typescript type assertions
     if p.at(T![<]) {
-        let jsx = JSX.parse_exclusive_syntax(p, maybe_parse_jsx_expression, |p, assertion| {
+        let jsx = Jsx.parse_exclusive_syntax(p, maybe_parse_jsx_expression, |p, assertion| {
```
MichaReiser, line 1745: Is it safe to parse out `JSX` even if `JSX` is disabled?

crates/rome_formatter/tests/specs/prettier/js/ternaries/nested.js.snap
```diff
@@ -197,6 +200,7 @@ a
 # Errors
 ```
 error[SyntaxError]: JSX elements are a JSX only feature. Convert your file to a JSX file or remove the syntax.
+<<<<<<< HEAD
```
MichaReiser, line 203: You have a merge issue here.

xunilrj: Ah sorry. You were faster than me.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: that is what I am doing now.

crates/rome_js_syntax/src/generated/kind.rs
```diff
@@ -457,6 +457,10 @@ pub enum JsSyntaxKind {
     TS_SETTER_SIGNATURE_CLASS_MEMBER,
     TS_INDEX_SIGNATURE_CLASS_MEMBER,
     TS_INDEX_SIGNATURE_MODIFIER_LIST,
+    JSX_NAME,
+    JSX_NAMESPACE_NAME,
+    JSX_REFERENCE_IDENTIFIER,
+    JSX_STRING_LITERAL,
     JSX_ELEMENT_EXPRESSION,
```
MichaReiser, line 464: Please add the `JSX_ELEMENT_EXPRESSION` to `JS_UNKNOWN_EXPRESSION` in `SyntaxKind.to_unknown`

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -84,7 +90,7 @@ fn parse_jsx_expression(p: &mut CheckpointedParser<'_, '_>) -> ParsedSyntax {
 // function f() {
 //     return <div />
 // }
-fn parse_jsx_element(p: &mut CheckpointedParser<'_, '_>) -> ParsedSyntax {
+fn parse_jsx_element(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 93: Does it still make sense to have the `CheckpointParser` abstraction as it's no longer used here (which was the main use case)? 

Please add some documentation to this method when it's safe to call it and when not. I think it would make sense to do something as in `try_parse_arrow_function_head` where the signature makes it clear that parsing might fail and that it's the responsiblity of the caller to abandon the marker in that case. 

https://github.com/rome/tools/blob/dc0ef1ef9e38e929f8ac64c31fe3af9ea46e01a8/crates/rslint_parser/src/syntax/function.rs#L445-L500

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -111,12 +117,18 @@ fn parse_jsx_element(p: &mut CheckpointedParser<'_, '_>) -> ParsedSyntax {
...
-fn parse_jsx_element_head(p: &mut CheckpointedParser<'_, '_>, m: Marker) -> ParsedSyntax {
-    if !p.parser.eat(T![<]) {
+fn parse_jsx_element_head(p: &mut Parser, m: Marker) -> ParsedSyntax {
+    if !p.eat(T![<]) {
         return ParsedSyntax::Absent;
     }
 
-    let _ = parse_jsx_any_element_name(p);
+    let name = parse_jsx_any_element_name(p);
+    if name.is_absent() {
```
MichaReiser, line 126: Abandoning here isn't safe if you call in from `expect_jsx_attribute_value`. I believe you need to rewrite all these functions to have a `try_parse` version and a normal version.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +179,91 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+    fn is_at_list_end(&mut self, p: &mut Parser) -> bool {
+        p.at(T![>]) || p.at(T![/])
+    }
+
+    fn recover(&mut self, p: &mut Parser, parsed_element: ParsedSyntax) -> RecoveryResult {
+        parsed_element.or_recover(
+            p,
+            &ParseRecovery::new(
+                JsSyntaxKind::JS_UNKNOWN_MEMBER,
+                token_set![T![/], T![>], T![<], T!['{'], T!['}'],],
```
MichaReiser, line 215: You probably want to recover on identifiers too, because they mark the start of a  new attribute

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +179,91 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+    }
+
+    let m = p.start();
+
+    p.bump(JsSyntaxKind::IDENT);
+
+    ParsedSyntax::Present(m.complete(p, JsSyntaxKind::JSX_NAME))
+}
+
+fn expect_jsx_attribute_initializer_clause(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 238: `expect` methods on the parser generate a diagnostic if the parser isn't at that position and return a bool. However, this method panics if the parser is positioned anywhere but an initializer. Also, returning `ParsedSyntax` only makes sense if the parser may not be at that given position, which can never happen here. 

That's why I would change this method to `parse_jsx_attribute_initializer_clause` and add the `if !p.at(T![=]) { return Absent }` at a check. You can then remove the `p.at(T![=])` check in the attribute parsing.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +179,91 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+    let m = p.start();
+
+    p.bump(T![=]);
+    let _ = expect_jsx_attribute_value(p);
+
+    ParsedSyntax::Present(m.complete(p, JsSyntaxKind::JSX_ATTRIBUTE_INITIALIZER_CLAUSE))
+}
+
+fn expect_jsx_attribute_value(p: &mut Parser) -> ParsedSyntax {
```
MichaReiser, line 247: Rename to `parse_jsx_attribute_value`.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +179,91 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+    p.bump(JsSyntaxKind::IDENT);
+
+    ParsedSyntax::Present(m.complete(p, JsSyntaxKind::JSX_NAME))
+}
+
+fn expect_jsx_attribute_initializer_clause(p: &mut Parser) -> ParsedSyntax {
+    let m = p.start();
+
+    p.bump(T![=]);
+    let _ = expect_jsx_attribute_value(p);
```
MichaReiser, line 242: You need to add a diagnsotic if the attribute value is missing after the `=` or is `<abcd a= />` valid?

MichaReiser (changes requested):
The parsing now abandons nodes even if outside of a `CheckpointedParser`. This isn't valid. I think we have to rewrite the speculative parsing mode similar to what we do with parenthesized functions.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Done.

crates/rome_js_syntax/src/generated/kind.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/expr.rs
MichaReiser: I just checked how this is implemented in TS and see the following issues

* JSX expressions are part of the primary expression rule and not of the unary expression rule. Thus, they shouldn't be parsed out here or we risk incorrect precedence handling (Typescript implements it as part of [`parseUpdateExpression`](https://github.com/microsoft/TypeScript/blob/4013271cd84f2766c0e8aaa813e719fc4a355ec8/src/compiler/parser.ts#L4900-L4903)
* `<test> <b>a</b>` would parse as a valid TS type assertion if we support parsing out type assertions and JSX at the same time because `<type>` would parse as the type and `<b>a</b>`. Yes, we would emit an error that one of them isn't allowed but the error message should really be that `<test>` has no corresponding close tag.
* Overall, our error recovery isn't good and this is mainly due to the fact that we do speculative JSX parsing where it isn't needed. We know that anything starting with a `<a` in a primary expression context should be parsed as a JSX expression and we should provide the best possible error recovery.
* I don't think we should even parse out JSX in JS only mode or we risk parsing `a<b>c` as invalid syntax tough it's perfectly valid  (we probably shouldn't parse out type assertions for the very same reason)

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Yeah. I plan to deal with diagnostics in another PR.

MichaReiser: I think it would make sense to do it now for this case. Because you're intentionally suppressing the compiler warning about the `must_use` of the `ParsedSyntax` which isn't correct here.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Done.

xtask/codegen/src/kinds_src.rs
```diff
@@ -485,7 +489,12 @@ pub const KINDS_SRC: KindsSrc = KindsSrc {
         "JSX_CLOSING_ELEMENT",
         "JSX_ANY_ELEMENT_NAME",
         "JSX_MEMBER_EXPRESSION",
-        "JSX_REFERENCE_IDENTIFIER",
+        "JSX_ATTRIBUTE_LIST",
+        "JSX_ATTRIBUTE",
+        "JSX_ATTRIBUTE_INITIALIZER_CLAUSE",
+        "JSX_ANY_ATTRIBUTE_VALUE",
```
MichaReiser, line 495: You don't need to add kinds for any nodes because these are just unions and not any real "nodes". Same for `JsxAnyAttributeName`

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -167,3 +179,91 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
+    if p.at(JsSyntaxKind::JS_STRING_LITERAL) {
+        let m = p.start();
+        p.bump(JsSyntaxKind::JS_STRING_LITERAL);
+        ParsedSyntax::Present(m.complete(p, JsSyntaxKind::JSX_STRING_LITERAL))
+    }
+    // expression values
+    else if p.at(T!['{']) {
+        let m = p.start();
+        p.bump(T!['{']);
+        let _ = super::expr::parse_expression(p, ExpressionContext::default());
```
MichaReiser, line 259: this should call into `parse_assignment_expression_or_higher` according to the [spec](https://facebook.github.io/jsx/#prod-JSXAttributeValue)

The following isn't allowed: `a={test, more}` maybe worth adding a test

xunilrj: Yeah. I prefer parsing the whole expression and emitting an error.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Done.

xtask/codegen/src/kinds_src.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: done.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: I will address this in another PR. See https://github.com/rome/tools/issues/2153.

crates/rslint_parser/src/syntax/expr.rs
xunilrj: I will address this in another PR. See https://github.com/rome/tools/issues/2153

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Done.

xunilrj:
> The parsing now abandons nodes even if outside of a `CheckpointedParser`. This isn't valid. I think we have to rewrite the speculative parsing mode similar to what we do with parenthesized functions.

THis will be addressed in another PR. See https://github.com/rome/tools/issues/2153.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -144,16 +148,19 @@ fn parse_jsx_element(p: &mut CheckpointedParser<'_, '_>, in_expression: bool) ->
...
-    in_expression: bool,
-) -> ParsedSyntax {
+fn parse_jsx_element_head(p: &mut Parser, m: Marker, in_expression: bool) -> ParsedSyntax {
     if !p.eat(T![<]) {
         return ParsedSyntax::Absent;
     }
 
-    let _ = parse_jsx_any_element_name(p);
+    let name = parse_jsx_any_element_name(p);
+    dbg!(&name);
```
MichaReiser, line 157: Please delete the `dbg` here

crates/rslint_parser/src/syntax/jsx/mod.rs
MichaReiser: Not really done, but will be dealt as part of another PR.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -241,13 +251,12 @@ fn parse_jsx_any_element_name(p: &mut CheckpointedParser<'_, '_>) -> ParsedSynta
...
 // <if />;
 // <namespace:name></namespace:name>;
-// <dashed-namespaced:dashed-name />;
 fn parse_jsx_any_name(p: &mut Parser) -> ParsedSyntax {
     parse_jsx_reference_identifier(p).map(|identifier| {
         if p.at(T![:]) {
             let m = identifier.precede(p);
             p.bump(T![:]);
-            parse_jsx_name(p).or_add_diagnostic(p, expected_identifier);
+            dbg!(parse_jsx_name(p)).or_add_diagnostic(p, expected_identifier);
```
MichaReiser, line 259: Remove `dbg`

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -279,3 +288,88 @@ fn parse_jsx_reference_identifier(p: &mut Parser) -> ParsedSyntax {
...
+
+fn parse_jsx_attribute_initializer_clause(p: &mut Parser) -> ParsedSyntax {
+    if !p.at(T![=]) {
+        return ParsedSyntax::Absent;
+    }
+
+    let m = p.start();
+
+    p.bump(T![=]);
+    let _ = parse_jsx_attribute_value(p);
```
MichaReiser, line 348: Please add a TODO here to add a diagnostic if it is missing or add it now.

crates/rslint_parser/test_data/inline/err/jsx_namespace_member_element_name.jsx
```diff
@@ -1 +1,3 @@
+<namespace:a></namespace:a>
 <namespace:a.b></namespace:a.b>
+<dashed-namespace:a.b></dashed-namespace:a.b>
```
MichaReiser, line 3: You need to add semicolons at the end of every statement or it will parse this as binary expressions
```suggestion
<namespace:a></namespace:a>;
<namespace:a.b></namespace:a.b>;
<dashed-namespace:a.b></dashed-namespace:a.b>;
```

crates/rslint_parser/test_data/inline/ok/jsx_any_name.jsx
```diff
@@ -2,4 +2,3 @@
 <a-b-c-d-e />;
 <if />;
 <namespace:name></namespace:name>;
-<dashed-namespaced:dashed-name />;
```
MichaReiser, line 5: Why did you remove this test case?

crates/rslint_parser/test_data/inline/err/jsx_namespace_member_element_name.jsx
xunilrj: Done.

crates/rslint_parser/test_data/inline/ok/jsx_any_name.jsx
xunilrj: I got confused and thought we could not have dashed namespaces here.... but the problem is different.
Done.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Done.

crates/rslint_parser/src/syntax/jsx/mod.rs
xunilrj: Done.

MichaReiser (approved):

PR #2226, opened 2022-03-13
refactor(rome_analyze): remove dependency on once_cell

nerdypepper:
By using `const` over lazily built statics, we can remove the dependency
on `once_cell`.

## Summary

We do not require `once_cell` in this scenario because Rust's `const` can achieve better results: compile-time evaluation, in-built language feature, no extra dependency.

## Test Plan

Public API remains the same.


yassere (approved):
Thank you! It's always nice to remove unnecessary dependencies.

PR #2228, opened 2022-03-14
fix(rslint_lexer): `read_regex` binary size

MichaReiser:
## Summary

The lexer uses a `unwind_loop` macro that unwinds loops by repeating the same statement 5 times, then repeating it 5 times inside a `loop` statement:

https://github.com/rome/tools/blob/3a22852e9b129a59e0d3b8eb4dbc48bb9675be7a/crates/rslint_lexer/src/lib.rs#L41-L57

```
unwind_loop!(self.next_byte())
```

becomes

```
self.next_byte();
self.next_byte();
self.next_byte();
self.next_byte();
self.next_byte();

loop {
  self.next_byte();
  self.next_byte();
  self.next_byte();
  self.next_byte();
  self.next_byte();
}
```

I don't know how much this improves performance, but it causes a significant binary bloat in the case of `read_regex` that uses a nested `unwind_loop`: The outer for parsing the regex, the inner for parsing the flags.

This PR removes the inner unwind because:

- Regular expressions aren't that common
- Not all of them have flags

This reduces the lexer size almost to 1/5th of its previous size:

* before: 473.9KiB
* after: 108.8KiB

This will probably also help with compilation time.

## Test Plan

Compared `bench_parser`. No significant change in performance (may be different for a library that heavily uses regex).

### DIFF
--- a/crates/rslint_lexer/src/lib.rs
+++ b/crates/rslint_lexer/src/lib.rs
@@ -1072,67 +1072,68 @@ impl<'src> Lexer<'src> {
                     if !in_class {
                         let (mut g, mut i, mut m, mut s, mut u, mut y, mut d) = (false, false, false, false, false, false, false);
 
-                        unwind_loop! {
-                            let next = self.next_bounded().copied();
+                        while let Some(next) = self.next_bounded().copied() {
                             let chr_start = self.cur;
+
                             match next {
-                                Some(b'g') => {
+                                b'g' => {
                                     if g && diagnostic.is_none() {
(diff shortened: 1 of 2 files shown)


xunilrj (approved):

xunilrj:
My vote would be to completely remove ```unwind_loop``` and assess if we have any significant performance regression.

MichaReiser:
> My vote would be to completely remove `unwind_loop` and assess if we have any significant performance regression.

Yeah, that makes sense or to see if there are alternatives.

PR #2230, opened 2022-03-15
refactor(rslint_parser): Use re-lexing for JSX multi-tokens

MichaReiser:
## Summary
Use the new re-lexing infrastructure to lex the `/>` and `</`
as a single token in a JSX context. This guarantees that
`/     >` isn't incorrectly parsed as `/>`

### DIFF
--- a/crates/rslint_parser/src/syntax/jsx/mod.rs
+++ b/crates/rslint_parser/src/syntax/jsx/mod.rs
@@ -1,8 +1,8 @@
 pub mod jsx_parse_errors;
 
-use rslint_lexer::{JsSyntaxKind, T};
+use rslint_lexer::{JsSyntaxKind, ReLexContext, T};
 
-use crate::{Checkpoint, Marker, ParsedSyntax, Parser};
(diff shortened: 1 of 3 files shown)


crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -134,24 +140,24 @@ fn parse_jsx_element_head(p: &mut CheckpointedParser<'_, '_>, m: Marker) -> Pars
...
     }
 
     let m = p.start();
 
-    if p.at(T![<]) && p.nth_at(1, T![/]) {
-        p.bump_multiple(2, JsSyntaxKind::L_ANGLE_SLASH);
-    } else {
+    p.bump(T![</]);
+
+    if parse_jsx_any_element_name(p).is_absent() {
```
MichaReiser, line 153: This would be a fragment which we currently don't support parsing, that's why we should abandon here

MichaReiser:
I misread the spec. The spec uses different tokens for `/` and `>` because for example this code is valid: `<a /*comment */ />` This only works if we keep the tokens separate, what I'll go for then.

PR #2232, opened 2022-03-15
chore(website): restore philosophy.md section and update packages

ematipico:
## Summary

This PR restores the "Philosophy" section of the website and removes a bullet point from it.

Also, updated few libraries. 

## Test Plan

None

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -1,7481 +1,7 @@
 {
 	"name": "rome-website",
-	"lockfileVersion": 2,
 	"requires": true,
-	"packages": {
-		"": {
-			"name": "rome-website",
-			"devDependencies": {
-				"@11ty/eleventy": "^0.11.0",
-				"@11ty/eleventy-navigation": "^0.1.6",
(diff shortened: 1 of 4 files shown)


yassere:
We probably need to reword the "Transparency" section. For practical reasons, we sometimes make project decisions while video conferencing. We definitely want to maintain a commitment to transparency, but it's difficult for an engineering startup to restrict its internal communication to only moderation decisions.

If we want to merge this right away, I'd suggest we just keep the "Technical" section for now and leave out the "Project Management" section until we have time to put more thought into the exact commitments we want to make.

ematipico:
> We probably need to reword the "Transparency" section. For practical reasons, we sometimes make project decisions while video conferencing. We definitely want to maintain a commitment to transparency, but it's difficult for an engineering startup to restrict its internal communication to only moderation decisions.
> 
> If we want to merge this right away, I'd suggest we just keep the "Technical" section for now and leave out the "Project Management" section until we have time to put more thought into the exact commitments we want to make.

@yassere done

yassere (approved):

PR #2234, opened 2022-03-15
feat(rome_cli): add tracing metrics collection and printing

leops:
## Summary

This PR introduces a `MetricsLayer` that collects timing metrics from the execution of `tracing` spans. Spans with a level of `DEBUG` and below only have their total (wall time) duration calculated while spans with a level of `TRACE` track both their busy (self time) and idle (child span) time. When the execution of a span completes, the timing data is accumulated into a histogram and this data is later retrieved in the `print_metrics` function to display a graphical visualization of the statistical data associated with each metric.

Both the `init_metrics` (injection of the collection layer) and `print_metrics` functions are only called when the `--show-metrics` flags is passed to the CLI (and the `format` subcommand is being used), so the overhead of metrics collection should be negligible when the feature is not being used.

# Example output

```
parse: mean = 333.2µs, stdev = 10.1ms
  8.3µs | *                                        |   0.0%
 60.5µs | *********************                    |  50.0%
148.2µs | ***********                              |  75.0%
300.8µs | *****                                    |  87.5%
619.0µs | ***                                      |  93.8%
  1.1ms | **                                       |  96.9%
  1.8ms | *                                        |  98.4%
  2.6ms | *                                        |  99.2%
  3.8ms | *                                        |  99.6%
  5.3ms | *                                        |  99.8%
  7.4ms | *                                        |  99.9%
 13.7ms | *                                        | 100.0%
 22.3ms | *                                        | 100.0%
745.0ms | *                                        | 100.0%
745.0ms |                                          | 100.0%
```

### DIFF
--- a/crates/rome_cli/src/metrics.rs
+++ b/crates/rome_cli/src/metrics.rs
@@ -9,7 +9,12 @@ use std::{
 use hdrhistogram::Histogram;
 use parking_lot::{Mutex, RwLock};
 use tracing::{span, subscriber::Interest, Level, Metadata, Subscriber};
-use tracing_subscriber::{layer::Context, prelude::*, registry::LookupSpan, Layer};
+use tracing_subscriber::{
+    layer::Context,
+    prelude::*,
+    registry::{LookupSpan, SpanRef},
+    Layer,
+};
 
 /// Implementation of a tracing [Layer] that collects timing information for spans into [Histogram]s
 struct MetricsLayer;
@@ -51,6 +56,37 @@ enum CallsiteEntry {
     },
 }
 
+impl CallsiteEntry {
+    fn from_level(level: &Level) -> Self {
+        /// Number of significant figures retained by the histogram
+        const SIGNIFICANT_FIGURES: u8 = 3;
+
+        match level {
+            &Level::TRACE => Self::Trace {
+                // SAFETY: Histogram::new only returns an error if the value of
+                // SIGNIFICANT_FIGURES is invalid, 3 is statically known to work
+                total: Histogram::new(SIGNIFICANT_FIGURES).unwrap(),
+                busy: Histogram::new(SIGNIFICANT_FIGURES).unwrap(),
+                idle: Histogram::new(SIGNIFICANT_FIGURES).unwrap(),
+            },
+            _ => Self::Debug {
+                total: Histogram::new(SIGNIFICANT_FIGURES).unwrap(),
+            },
+        }
+    }
+
+    fn into_histograms(self, name: &str) -> Vec<(Cow<str>, Histogram<u64>)> {
+        match self {
+            CallsiteEntry::Debug { total } => vec![(Cow::Borrowed(name), total)],
+            CallsiteEntry::Trace { total, busy, idle } => vec![
+                (Cow::Borrowed(name), total),
+                (Cow::Owned(format!("{name}.busy")), busy),
+                (Cow::Owned(format!("{name}.idle")), idle),
+            ],
+        }
+    }
+}
+
 /// Extension data attached to tracing spans to keep track of their idle and busy time
 ///
 /// Most of the associated code is based on the similar logic found in `tracing-subscriber`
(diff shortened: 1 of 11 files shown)


MichaReiser:
Would you mind rebasing and then run `!bench_parser` to see if there are no perf regressions (or ideally, run the benchmarks locally and comment the results).

leops:
!bench_parser

crates/rome_cli/src/metrics.rs
```diff
@@ -0,0 +1,254 @@
+use std::{
+    borrow::Cow,
+    collections::HashMap,
+    hash::Hash,
+    ptr,
+    time::{Duration, Instant},
+};
+
+use hdrhistogram::Histogram;
+use parking_lot::{Mutex, RwLock};
```
MichaReiser, line 10: What's the reason for using `parking_lot` vs the built-in `Mutex` and `RwLock`?

crates/rome_cli/src/metrics.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Static pointer to the metadata of a callsite, used as a unique identifier
+/// for collecting spans created from there in the global metrics map
+struct CallsiteKey(&'static Metadata<'static>);
+
+impl PartialEq for CallsiteKey {
+    fn eq(&self, other: &Self) -> bool {
+        ptr::eq(self.0, other.0)
+    }
+}
```
MichaReiser, line 30: Nit: Shouldn't it be possible to derive `Eq`, `Hash` and `PartialEq` because `Metadata` implements `Eq`?

crates/rome_cli/src/metrics.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// and create new histogram for the spans created by this callsite
+    fn register_callsite(&self, metadata: &'static Metadata<'static>) -> Interest {
+        if !metadata.is_span() {
+            return Interest::never();
+        }
+
+        let entry = match metadata.level() {
+            &Level::TRACE => CallsiteEntry::Trace {
+                total: Histogram::new(3).unwrap(),
+                busy: Histogram::new(3).unwrap(),
```
MichaReiser, line 89: What's the meaning of the 3?

crates/rome_cli/src/metrics.rs
```diff
@@ -0,0 +1,254 @@
...
+        METRICS
+            .write()
+            .insert(CallsiteKey(metadata), Mutex::new(entry));
+
+        Interest::always()
+    }
+
+    /// When a new span is created, attach the timing data extension to it
+    fn on_new_span(&self, _attrs: &span::Attributes<'_>, id: &span::Id, ctx: Context<'_, S>) {
+        let span = ctx.span(id).expect("Span not found, this is a bug");
```
MichaReiser, line 106: Would be helpful to know why this is a bug. How about: "Span not found, should have been registered by register_callsite`

crates/rome_cli/src/metrics.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        if extensions.get_mut::<Timings>().is_none() {
+            extensions.insert(Timings::new());
+        }
+    }
+
+    /// When a span is entered, start counting idle time for the parent span if
+    /// it exists and busy time for the entered span itself
+    fn on_enter(&self, id: &span::Id, ctx: Context<'_, S>) {
+        let span = ctx.span(id).expect("Span not found, this is a bug");
```
MichaReiser, line 117: Nit: How about extracting this into a helper function to avoid repeating the message?

crates/rome_cli/src/metrics.rs
```diff
@@ -0,0 +1,254 @@
...
+                duration,
+                "*".repeat(bar_length),
+                v.quantile_iterated_to() * 100.0,
+            );
+        }
+
+        // Print an empty line after each histogram
+        println!();
+    }
+}
```
MichaReiser, line 254: Is it possible to write a few test cases for the metrics layer? Mainly to make sure the calculations are correct.

MichaReiser (approved):

crates/rome_cli/src/metrics.rs
leops: Originally this was because I had stored the `METRICS` global storage as an actual `static` value instead of using `lazy_static` (Rust static initializers must be const-expressions, and the constructor for `std` mutexes is not const), and while this is no longer the case I expect the global `RwLock` to be mostly uncontended (it will only be locked for writing a few times on startup, then all the later locks will be for reading only) so using an implementation that has a fast path for this case doesn't hurt

crates/rome_cli/src/metrics.rs
leops: Deriving these traits would have a slightly different behavior as it would call through to their implementation on the inner `Metadata` object itself, while using `ptr::eq` and `ptr::hash` transparently casts the reference to `*const Metadata<'static>` and only compares / hashes the numerical value of the pointer without dereferencing it

crates/rome_cli/src/metrics.rs
leops: This is the number of "significant digits" retained by the histogram which controls the granularity at which the histogram internally buckets different values together, 3 is generally considered to be a good enough value (so 0.0345 is different for 0.0346, and 345000 is different from 346000, but 345000 and 345001 are not)

ematipico (approved):

PR #2240, opened 2022-03-16
fix(rslint_parser): Don't parse JSX as arrow functions

MichaReiser:
## Summary

Don't parse `<A extends>()` or `<A extends=B>`  as arrow functions, these must be starts of a JSX element.

Depends on attribute parsing for the tests to pass

## Test Plan

#2224

### DIFF
--- a/crates/rslint_parser/src/syntax/function.rs
+++ b/crates/rslint_parser/src/syntax/function.rs
@@ -667,6 +667,7 @@ fn is_parenthesized_arrow_function_expression_impl(
             // <A extends B>() => {};
             // <A=string>() => {};
             // <A, B>() => {};
+            // <A extends B<C>>() => {}
 
             // <a... JSX override
             else if JsSyntaxFeature::Jsx.is_supported(p) {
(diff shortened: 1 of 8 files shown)


crates/rslint_parser/src/syntax/function.rs
```diff
@@ -653,12 +652,47 @@ fn is_parenthesized_arrow_function_expression_impl(
...
+            // // These may look like a valid arrows but are JSX
+            // <A extends>() =</A>;
+            // <A extends="B">() =</A>;
+            // <A extends ok>() =</A>;
+
+            // test tsx tsx_type_arguments
+            // // These are valid type arguments
+            // <A extends B>() => {};
+            // <A=string>() => {};
+            // <A, B>() => {};
```
ematipico, line 669: Can we add 

```tsx
<A extends B<C>>() => {}
```

As a test?

MichaReiser: Added it, seems to be fine😅

ematipico (approved):

PR #2241, opened 2022-03-16
feat(rslint_parser): JSX Attribute name & value

MichaReiser:
## Summary

Support namespace attribute names, multiline string literal values, and lex Unicode escape sequences as regular strings in JSX attribute values. 

This PR further fixes:

* Create a diagnostic for unterminated string literals in JS `"a\n ...invalid"`
* Change the type of the `namespace` field of `JsxNamespaceName` to `JsxName`. 
* Parse `<a>` as elements with a `JsxName` rather than `JsxReferenceIdentifier` as these aren't referencing a local name. This isn't strictly spec compliant. This is a React specific and is the same as Rome classic is doing.

Part of #2153

## Test Plan

Added new parser tests


crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -238,9 +242,15 @@ fn parse_jsx_any_element_name(p: &mut Parser) -> ParsedSyntax {
             left = m.complete(p, JSX_MEMBER_NAME)
         }
 
-        Present(left)
-    } else {
         left
+    })
+}
+
+fn is_html_element(element_name: &str) -> bool {
```
ematipico, line 249: The function name and its implementation are a bit misleading. Based on its name, `<a-d>` is a valid HTML element, while it's not. I would change the name to actually reflect the implementation, e.g.: `is_allowed_tag` or `is_tag_starts_with_lowercase`, etc.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -275,29 +285,20 @@ fn parse_jsx_name(p: &mut Parser) -> ParsedSyntax {
...
-
-    Present(m.complete(p, JSX_REFERENCE_IDENTIFIER))
-}
-
 struct JsxAttributeList;
 // test jsx jsx_element_attributes
 // function f() {
 //     return <div string_literal="a" expression={1} novalue el=<a/>></div>;
 // }
+// <div dashed-name='test' use:validate="abcd" />;
```
ematipico, line 293: Can we add a test case like this:

```
<div use-dashed_underscore:validate="ahaha" />
```

crates/rslint_parser/src/syntax/jsx/mod.rs
MichaReiser: I went with `is_intrinsic_element` to use the same Terminology as TS.

ematipico (approved):

PR #2242, opened 2022-03-16
feat(rslint_parser): JSX Spread attribute

MichaReiser:
## Summary

Adds support for parsing `<a {...spread}>` attributes.

## Test Plan

Added new parser tests.


crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -296,17 +296,11 @@ struct JsxAttributeList;
 // <div invalid-unicode-escape="\u10000\u20000" />;
 impl ParseNodeList for JsxAttributeList {
     fn parse_element(&mut self, p: &mut Parser) -> ParsedSyntax {
-        let m = p.start();
-
-        let name = parse_jsx_name_or_namespace(p);
-        if name.is_absent() {
-            m.abandon(p);
-            return ParsedSyntax::Absent;
+        if matches!(p.cur(), T!['{'] | T![...]) {
```
ematipico, line 299: Out of curiosity, why do we check against `...` too? Shouldn't the spread node start with `{`?

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -329,9 +323,66 @@ impl ParseNodeList for JsxAttributeList {
...
+//     />;
+fn parse_jsx_spread_attribute(p: &mut Parser) -> ParsedSyntax {
+    if !matches!(p.cur(), T![...] | T!['{']) {
+        return Absent;
+    }
+
+    let m = p.start();
+
+    p.expect(T!['{']);
+    p.expect(T![...]);
```
ematipico, line 362: I think there's something wrong... at line 355 we assume that we can be at `...`, if the parser really is at that position, should line 361 (`p.expect(T!['{'])`) throw an error?

What I am saying is that if we are at `...` we should only call `p.expect(T![...])`

crates/rslint_parser/src/syntax/jsx/mod.rs
MichaReiser: It should, but we can still recover and try to parse it as a spread if the `{` is missing.

crates/rslint_parser/src/syntax/jsx/mod.rs
MichaReiser: Not sure if I understand what you mean. 

Line 355: `if !matches!(p.cur(), T![...] | T!['{'])` tests if the parser's either at `...` or `{`. If not, it returns. 

The lines 361 and 362 use `expect` to cover all these cases:
* `{...`: all good :)
* `{`: uh... the `...` is missing
* `...` ohoh, there was no `{`.

ematipico: > `...` ohoh, there was no `{`


That's what I am trying to understand... if we are at `...` it means that `{` is already passed, why do we check it again?. Unless something like this is valid:

```jsx
<Anchor {...{...valid}} />
```
If so, then `{...valid}` would be an object expression.

Or, most probably, there's something that I don't see.

MichaReiser: It's for something like this `<A ...test} />` We parse this as a spread element because that's most likely what the user meant.

ematipico: That's too fancy for my standard! LOL 😆

ematipico (approved):

PR #2229, opened 2022-03-14
Feature/jsx children elements

xunilrj:
see https://github.com/rome/tools/issues/2153

This PR implements recursive children parser.

We can now parse
```jsx
<a>
    <b>
        <d></d>
        <e></e>
    </b>
    <c></c>
</a>
```

```
<a>
    {something}
    {x => something}
</a>
```

```
<a>test</a>;
<a>   whitespace handling </a>;
<a> multi
line
node
</a>;
<test>\u3333</test> // no error for invalid unicode escape
```

# Expressions

The spec seems too restrictive in which expressions can be used. Different parsers end up parsing these expressions differently.

For example:

- babel seems to support everything but ```super``` and ```new.target```;
- ts supports everything but "import.meta", "super", "functions without name" and "new.target";

This PR explicitely error on:

- import.meta
- new.target
- class expression

Super already generates an error when used outside constructors.

```
error[SyntaxError]: This expression is not valid as a JSX expression.
   ┌─ C:\github\rome\tools\crates\rslint_parser\test_data\inline\ok\jsx_children_expression.jsx:14:4
   │
14 │   {import.meta}
   │    ^^^^^^^^^^^

error[SyntaxError]: This expression is not valid as a JSX expression.
   ┌─ C:\github\rome\tools\crates\rslint_parser\test_data\inline\ok\jsx_children_expression.jsx:21:4
   │
21 │   {class A{}}
   │    ^^^^^^^^^

error[SyntaxError]: `super` is only valid inside of a class constructor of a subclass.
   ┌─ C:\github\rome\tools\crates\rslint_parser\test_data\inline\ok\jsx_children_expression.jsx:38:4
   │
38 │   {super()}
   │    ^^^^^

error[SyntaxError]: This expression is not valid as a JSX expression.
   ┌─ C:\github\rome\tools\crates\rslint_parser\test_data\inline\ok\jsx_children_expression.jsx:48:4
   │
48 │   {new.target}
   │    ^^^^^^^^^^
```

PR #2243, opened 2022-03-16
feat(rslint_parser): Parse JSX Fragments

MichaReiser:
## Summary

Implements parsing of JSX Fragments `<>a</>` 

I decided to rename `JsxAnyElement` to `JsxAnyTag` so that we don't need a new expression node for fragments. Tag also seems to be the agreed term in React:

* Error message if you have multiple root nodes: `/App.js: Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>? (6:4)`
* [Documentation](https://beta.reactjs.org/learn/writing-markup-with-jsx#1-return-a-single-root-element) referring to Tag when explaining fragments.

I used this PR as an opportunity to restructure the folder structure for the JSX nodes so that `tag` and `attribute` now have their own folder. 

## Test Plan

Added new parser tests


crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -110,58 +110,88 @@ fn parse_jsx_expression(p: &mut CheckpointedParser<'_, '_>) -> ParsedSyntax {
...
+            return Absent;
+        }
+
+        p.bump_with_context(T![>], LexContext::JsxChild);
+
+        parse_children(p);
+
+        if !p.expect(T![<]) || !p.expect(T![/]) || !p.expect(T![>]) {
+            m.abandon(p);
+            return Absent;
```
ematipico, line 185: Is it safe `return Absent`? How do we know that `parse_children(p)` didn't advance the parser? We should have a comment explaining why it's safe

xtask/codegen/js.ungram
```diff
@@ -2178,13 +2179,22 @@ JsxClosingElement =
 	name: JsxAnyElementName
 	'>'
 
+// <>content</>
+JsxFragment =
+	opening_l_angle_token: '<'
+	opening_r_angle_token: '>'
+	children: JsxTextLiteral?
```
ematipico, line 2186: I don't think this is correct. Fragments can have everything inside, also other jsx tags:

```jsx
<>
	<Header />
	<Body />
	<Footer />
</>
```

This is valid syntax.

crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -110,58 +110,88 @@ fn parse_jsx_expression(p: &mut CheckpointedParser<'_, '_>) -> ParsedSyntax {
...
+// test jsx fragments
+// <>abcd</>;
+// <>   whitespace
+// </>;
+// <
+//   /*comment */
+//   >
+//   <
+//   /
+// >;
```
ematipico, line 122: Let's expand the tests with more children: other fragments, other jsx elements, expressions, etc.

ematipico (changes requested):

crates/rslint_parser/src/syntax/jsx/mod.rs
MichaReiser: I agree, but it's how the JSX Element parsing works at the moment ([discussion](https://github.com/rome/tools/pull/2224#discussion_r826656207)). This is tracked as part of #2153

xtask/codegen/js.ungram
MichaReiser: This is correct but parsing anything else other than `JsxText` isn't implemented yet and @xunilrj is working on it in #2229.

crates/rslint_parser/src/syntax/jsx/mod.rs
MichaReiser: More than happy to do so once other children elements are implemented. But that's all we support at the moment.

xtask/codegen/js.ungram
ematipico: Sorry, I wasn't aware of it

ematipico (approved):

xtask/codegen/js.ungram
MichaReiser: No worries, I should have mentioned it in the PR description.

PR #2244, opened 2022-03-16
feat(rome_formatter): add instanceof expr and in expr

ematipico:
## Summary

This PR adds `JsInstanceofExpression` and `JsInExpression` to be considered inside the formatting of binaryish expressions.

I took the opportunity to also refactor and generalize a bit the implementation of some parts of the code.

The formatted result of the new tests is a bit strange, also because it's highly impossible to have some code like this, but it's completely inline with the whole algorithm.

Also Prettier formats them at the same way.

## Test Plan

New tests

### DIFF
--- a/crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js
+++ b/crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js
@@ -106,3 +106,7 @@ foo && bar || baz;
 a instanceof Number instanceof String instanceof Boolean instanceof Number instanceof String instanceof Boolean instanceof Number instanceof String instanceof Boolean;
 
 a in LongClassName in LongClassName in LongClassName in LongClassName in LongClassName in LongClassName in LongClassName in LongClassName in LongClassName;
+
+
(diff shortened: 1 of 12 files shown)


crates/rome_formatter/tests/specs/js/module/expression/logical_expression.js
```diff
@@ -102,3 +102,7 @@ let foo = one && two || three && four || five && six;
 // Implicitly parenthesized && and || requires parens
 foo && bar || baz;
 
+
+a instanceof Number instanceof String instanceof Boolean instanceof Number instanceof String instanceof Boolean instanceof Number instanceof String instanceof Boolean;
```
leops, line 106: While useful for testing the line breaking behavior that code doesn't make much sense, how about adding a few more test cases mixing in and instanceof with existing binary and logical operations to test the parenthesis insertion code as well ?

ematipico: Good catch. There was some missing case that I wasn't handling! Check the new cases with the new implementation inside the last commit.

leops (approved):

PR #2247, opened 2022-03-16
chore: Fix snapshot after renaming `element` to `tag`

MichaReiser:
## Summary

#2243 added a snapshot which wasn't updated before merging because there weren't any conflicts.

## Test Plan

`cargo test`


MichaReiser:
Fixed with #2241

PR #2249, opened 2022-03-16
ci: includes paths of when running the workflow

ematipico:
## Summary

This PR adds `paths:` to the `parser_conformance` workflow.

At the moment, every PR hits this check, which is not ideal because some changes are not related to the parser. This would allow us to reduce the number and minutes of the actions we can trigger.

## Test Plan

It should work on this PR (or it should be merged, not sure)

### DIFF
--- a/.github/workflows/parser_conformance.yml
+++ b/.github/workflows/parser_conformance.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'crates/rslint_**/'
+      - 'crates/rome_js_syntax/'
+      - 'xtask/codegen'
 
 env:
   RUST_LOG: info


.github/workflows/parser_conformance.yml
```diff
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'crates/rslint_**/'
```
MichaReiser, line 8: Does it need two `*` as this isn't recursing into subdirectories but only matching a prefix?

MichaReiser (approved):
Awesome!

.github/workflows/parser_conformance.yml
ematipico: I checked examples online: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request

It uses `**` for file matching

xunilrj (approved):

PR #2252, opened 2022-03-17
refactor(rome_fs): implement a basic abstraction over the filesystem

leops:
## Summary

This PR renames the `rome_path` crate to `rome_fs` and turns it into a general purpose filesystem abstraction including:
- A generic `FileSystem` trait with two implementations: `OsFileSystem` calls through to the underlying OS while `MemoryFileSystem` works over a hashmap of virtual files
- The `PathInterner` trait as an abstraction over path interning with again two implementations: `IndexSetInterner` is based on the `UrlInterner` currently used in the LSP and is designed to eventually replace it, while `AtomicInterner` implements the fast (but specific) interning strategy of the formatting CLI

Additional changes include:
- The `rome_app::App` object now owns an instance of the filesystem
- The `run_cli` entrypoint no longer panics on error but returns a `Result<(), Termination>` instead
- Memory mapping of files in the CLI has been removed

Besides moving some existing dependencies of the CLI and LSP (`indexmap` and `rayon`) to the `rome_fs` crate, this PR introduces two additional dependencies:
- `parking_lot` is used internally in the `MemoryFileSystem` to implement "owned locks" over files (mutex guards containing a reference counted handle to the locked state instead of a borrow of the underlying mutex)
- `crossbeam` is used as a replacement for `std::sync::mpsc` channels since its implementation of `Sender` can be shared across threads (implements `Sync`) instead of having to be cloned like its standard library counterpart

## Test Plan

The CLI test has been rewritten and is no longer ignored, it now makes use of the in-memory file system to setup a virtual environment to run the CLI and verify its execution.


crates/rome_cli/src/commands/format.rs
```diff
@@ -149,7 +123,16 @@ pub(crate) fn format(mut session: CliSession) {
...
-        let source = fs::read_to_string(path).ok().unwrap_or_else(String::new);
+        let mut source = String::new();
+        session
+            .app
+            .fs
+            .open(&path)
+            .and_then(|mut file| file.read_to_string(&mut source))
+            // Because the diagnostics include IO errors, path could be a
+            // directory or a non-existent file so any error that may happen
+            // while reading the file is ignored
```
MichaReiser, line 134: I don't think I understand this comment. Why is it safe to ignore the error?

crates/rome_fs/src/fs/memory.rs
```diff
@@ -0,0 +1,103 @@
...
+        let content = str::from_utf8(&self.inner)
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))?;
+        // Append the content of the file to the buffer
+        buffer.push_str(content);
+        Ok(())
+    }
+
+    fn set_content(&mut self, content: &[u8]) -> io::Result<()> {
+        // Resize the memory buffer to fit the new content
+        self.inner.resize(content.len(), 0);
```
MichaReiser, line 81: I would expect that `.copy_from_slice` is optimized so that it sets the capacity itself. Or is this mainly to reduce memory usage to avoid overly large inner buffers?

crates/rome_fs/src/fs/memory.rs
```diff
@@ -0,0 +1,103 @@
...
+        // Traversal is implemented by iterating on all keys, and matching on
+        // those that are prefixed with the provided `base` path
+        for path in self.fs.files.keys() {
+            if path.strip_prefix(&base).is_ok() {
+                let file_id = ctx.interner().intern_path(path.into());
+                ctx.handle_file(path, file_id);
+            }
+        }
+    }
+}
```
MichaReiser, line 103: Could we add some unit tests for the memory file system?

MichaReiser (approved):

MichaReiser:
Have you done some profiling to see what the overhead is for the dynamic dispatches? 

I'm still not entirely familiar with Rust's dynamic dispatching. But would it be possible that the `FileSystem::open` returns `Self::File` rather than a `Box<&dyn File>` or is that not possible because it would then require that all type using the FileSystem must be generic over `FS`.

Have you looked at [rust-vfs](https://github.com/manuel-woelker/rust-vfs) and [rsfs](https://docs.rs/rsfs/latest/rsfs/)?

crates/rome_fs/src/fs/memory.rs
leops: `copy_from_slice` is a method implemented on the primitive `[u8]` slice type and not on `Vec<u8>` so it can't resize the underlying storage, and will in fact panic if the length of `self.inner` and `content` were different

crates/rome_cli/src/commands/format.rs
leops: There are a few separate things going here and the comment maybe doesn't explain it too well:
- The first is that if this code is reached this means an error diagnostic was emitted for this path, we can't really know what it was at this stage since its an opaque diagnostic but it could be that the file doesn't exist, is a directory, or the process doesn't have the permission to read it. There's a fairly high chance the same error will happen again when the file is loaded a second time, in which case we don't want to do anything with it since we're already in the process of printing the error diagnostic to the console anyway and we don't want to show the same error twice
- The second scenario is that the filesystem could be in an inconsistent state, for instance the file got deleted between the moment the diagnostic was emitted and the moment it gets printed. The probability of this happening is very low so for now I just opted to show the diagnostics in "degraded mode" with no code span information and not print any additional error. 

One thing I want to change in the future is how filesystem abstractions are composed, so that it would be possible to wrap an `OsFileSystem` inside a `MemoryFileSystem` instead of using either one of the other: the idea is that the file would only be loaded once from the OS then kept into memory, and all subsequent reads would be performed on that byte buffer instead which would ensure a coherent view of the filesystem for the duration of the CLI run as well as avoiding the file reloading that currently happens for diagnostics.

leops:
> Have you done some profiling to see what the overhead is for the dynamic dispatches?

I have done some profiling, and I'm not seeing any particular performance regression since the cost of the system calls being performed in the `OsFileSystem` is still way higher than the cost of dynamic dispatch. But maybe it would benefit from additional profiling on a machine / system that has a more efficient filesystem ...

> I'm still not entirely familiar with Rust's dynamic dispatching. But would it be possible that the `FileSystem::open` returns `Self::File` rather than a `Box<&dyn File>` or is that not possible because it would then require that all type using the FileSystem must be generic over `FS`.

Yes that's not really possible, in order for a trait to be "object safe" and usable in dynamic dispatch it must not have any associated type, associated constant, generic method or associated function not taking a "&self" parameter.

> Have you looked at [rust-vfs](https://github.com/manuel-woelker/rust-vfs) and [rsfs](https://docs.rs/rsfs/latest/rsfs/)?

I looked a bit into existing VFS implementations, but I ultimately decided to go with an implementation we own for several reasons:
- Related to the above, the `FileSystem` trait has to be object safe (this is not the case of `rsfs` for instance)
- The interface I implemented is a higher level abstraction over the underlying filesystem that's oriented for our own use case: for instance it does not directly expose a `read_dir` method but instead has a `traversal` scope that takes care of implementing an efficient multi-threaded recursive traversal
- Finally owning this doesn't incur too high of an engineering and maintenance cost: it's preferable to defer to libraries the implementation of complex multi-threaded logic like `crossbeam`, `rayon` or `parking_lot`, but this part is straightforward enough that can handle it ourselves

crates/rome_cli/src/commands/format.rs
ematipico: I think the comment should be reworded. Your explanation is way better compared to the comment. If in the future I would have to read it again, it would have hard time to understand it without the context given by this explanation of yours.

crates/rome_cli/src/lib.rs
```diff
@@ -14,29 +16,47 @@ pub struct CliSession {
...
-            app: App::new(),
+            app: App::from_env(),
             args: Arguments::from_env(),
         }
     }
 }
 
+/// Error message returned by the CLI when it aborts with an error
+#[derive(PartialEq, Eq)]
+pub struct Termination(pub(crate) String);
```
ematipico, line 27: nit: I think having enums for error messages is better, but I don't have a strong opinion. I suppose eventually they are all compiled to the same binary structure.

crates/rome_fs/src/fs.rs
```diff
@@ -0,0 +1,69 @@
...
+pub trait FileSystem: Sync + RefUnwindSafe {
+    /// Open a handle to the file at `path`
+    ///
+    /// Currently this locks down the file for both reading and writing
+    fn open(&self, path: &Path) -> io::Result<Box<dyn File>>;
+
+    /// Initiate a traversal of the filesystem
+    ///
+    /// This method creates a new "traversal scope" that can be used to
+    /// efficient batch many filesystem read operations
```
ematipico, line 25:
```suggestion
    /// efficiently batch many filesystem read operations
```

crates/rome_fs/src/fs.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn traversal<'scope>(&'scope self, func: BoxedTraversal<'_, 'scope>);
+}
+
+pub trait File {
+    /// Read the content of the file into `buffer`
+    fn read_to_string(&mut self, buffer: &mut String) -> io::Result<()>;
+
+    /// Overwrite the content of the file with the provided bytes
+    fn set_content(&mut self, content: &[u8]) -> io::Result<()>;
```
ematipico, line 34: Does it save the new content? Can we write the response inside the documentation? I think this particular is important

crates/rome_fs/src/fs.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type BoxedTraversal<'fs, 'scope> = Box<dyn FnOnce(&dyn TraversalScope<'scope>) + Send + 'fs>;
+
+pub trait TraversalScope<'scope> {
+    /// Spawn a new filesystem read task
+    ///
+    /// If the provided path exists and is a file, the [`handle_file`](TraversalContext::handle_file)
```
ematipico, line 42:
```suggestion
    /// If the provided path exists and is a file, then the [`handle_file`](TraversalContext::handle_file)
```

crates/rome_fs/src/fs.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// method of the provided [TraversalContext] will be called. If its a
```
ematipico, line 43:
```suggestion
    /// method of the provided [TraversalContext] will be called. If it's a
```

crates/rome_fs/src/fs.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// returns true for will be handled as well
+    fn spawn(&self, context: &'scope dyn TraversalContext, path: PathBuf);
+}
+
+pub trait TraversalContext: Sync {
+    fn interner(&self) -> &dyn PathInterner;
+
+    fn push_diagnostic(&self, file_id: FileId, code: &'static str, message: String);
+    fn can_handle(&self, path: &RomePath) -> bool;
+    fn handle_file(&self, path: &Path, file_id: FileId);
```
ematipico, line 55: Doc comments?

crates/rome_fs/src/fs/memory.rs
```diff
@@ -0,0 +1,103 @@
...
+use parking_lot::{lock_api::ArcMutexGuard, Mutex, RawMutex};
+
+use crate::{FileSystem, TraversalContext, TraversalScope};
+
+use super::{BoxedTraversal, File};
+
+/// Fully in-memory file system, stores the content of all known files in a hashmap
+#[derive(Default)]
+pub struct MemoryFileSystem {
+    files: HashMap<PathBuf, FileEntry>,
```
ematipico, line 19: Why do we use a `PathBuf` instead of a `FileId`? If I had to rephrase the question, can't we use something "smaller" to uniquely identify a file?

crates/rome_fs/src/interner.rs
```diff
@@ -0,0 +1,56 @@
+use std::{
```
ematipico, line 1: Can we document this module please? I have no idea what an "interner" is

ematipico (approved):

crates/rome_cli/src/lib.rs
leops: I do plan on moving towards a fully-fledged error type (maybe using anyhow or thiserror), but this will be part of a followup PR to rework the error handling of the CLI so that it only panics in case of a bug (and not for instance when the CLI is run with invalid arguments)

crates/rome_fs/src/fs/memory.rs
leops: That's the plan eventually yes, I was thinking the internal representation of `RomePath` could be simplified to only hold a `FileId`, and this is why I moved the implementations of `PathInterner` to the fs crate as well. I just held on to that for now since it wasn't really necessary for the CLI, and used `PathBuf` works well enough since the `MemoryFileSystem` is only currently being used for tests.

PR #2253, opened 2022-03-17
feat(rslint_parser): Parse TS-Decorators into Trivia

MichaReiser:
## Summary

This PR adds parser support for TypeScript's decorators by parsing the decorators into a new `skipped` token trivia. The reason for parsing it as trivia rather than just bumping the tokens is that a `skipped` trivia doesn't add an unexpected token to the parent node. Adding the unexpected node would make the parent node an `*UNKNOWN` node with the result that the formatter bails out of formatting the note completely. 

The implemented parsing doesn't perform any validation because the main goal is to be able to "skip" only decorators, kind of as an "intelligent" error recovery strategy. 

## Test Plan

There are a few tests that are regressing because they test for invalid decorator syntax and this PR doesn't add any diagnostics for invalid decorator syntax.

### DIFF
--- a/crates/rome_formatter/src/formatter.rs
+++ b/crates/rome_formatter/src/formatter.rs
@@ -338,13 +338,13 @@ impl Formatter {
             .and_then(|token| token.trailing_trivia().last())
             .map_or(false, |trivia| trivia.is_newline());
 
-        self.format_leading_trivia_pieces(
+        self.print_leading_trivia_pieces(
             token.leading_trivia().pieces(),
             trim_mode,
             has_trailing_newline,
         )
         .unwrap_or_else(|_| {
-            self.format_leading_with_skipped(token, trim_mode, has_trailing_newline)
+            self.print_leading_trivia_with_skipped_tokens(token, trim_mode, has_trailing_newline)
         })
     }
 
@@ -362,7 +362,7 @@ impl Formatter {
     /// ## Panics
     ///
     /// If called on a token that does not have skipped trivia
-    fn format_leading_with_skipped(
+    fn print_leading_trivia_with_skipped_tokens(
         &self,
         token: &SyntaxToken,
         trim_mode: TriviaPrintMode,
@@ -392,7 +392,7 @@ impl Formatter {
                     // Format the  collected leading trivia as the leading trivia of this "skipped token trivia"
                     skipped_trivia_range = Some(piece.text_range());
                     elements.push(
-                        self.format_leading_trivia_pieces(
+                        self.print_leading_trivia_pieces(
                             leading_trivia.drain(..),
                             trim_mode,
                             has_trailing_newline,
@@ -447,7 +447,7 @@ impl Formatter {
         elements.push(self.print_trailing_trivia_pieces(trailing_trivia.into_iter()));
 
         elements.push(
-            self.format_leading_trivia_pieces(leading_trivia.into_iter(), trim_mode, after_newline)
+            self.print_leading_trivia_pieces(leading_trivia.into_iter(), trim_mode, after_newline)
                 .expect("All skipped trivia pieces should have been filtered out"),
         );
 
@@ -461,7 +461,7 @@ impl Formatter {
     /// Returns [Err] if the leading trivia contains any skipped trivia. Returns the formatted
     /// leading trivia otherwise.
     ///
-    fn format_leading_trivia_pieces<I>(
+    fn print_leading_trivia_pieces<I>(
         &self,
         pieces: I,
         mut trim_mode: TriviaPrintMode,
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -598,19 +598,6 @@ impl<L: Language> Iterator for SyntaxTriviaPiecesIterator<L> {
         })
     }
 
-    fn last(self) -> Option<Self::Item>
-    where
-        Self: Sized,
-    {
-        let raw = self.iter.raw.clone();
-        self.iter.last().map(|(offset, trivia)| SyntaxTriviaPiece {
-            raw,
-            offset,
-            trivia,
-            _p: PhantomData,
-        })
-    }
-
     fn size_hint(&self) -> (usize, Option<usize>) {
         self.iter.size_hint()
     }
--- a/crates/rome_rowan/src/cursor.rs
+++ b/crates/rome_rowan/src/cursor.rs
@@ -576,15 +576,6 @@ impl Iterator for SyntaxTriviaPiecesIterator {
         Some(piece)
     }
 
-    fn last(self) -> Option<Self::Item>
-    where
-        Self: Sized,
-    {
-        let last = self.raw.last()?;
-
-        Some((self.raw.token.text_range().end() - last.length, *last))
-    }
-
     fn size_hint(&self) -> (usize, Option<usize>) {
         let len = self.end_index.saturating_sub(self.next_index);
         (len, Some(len))


MichaReiser:
!bench_parser

crates/rome_formatter/tests/specs/prettier/js/decorators/multiple.js.snap
```diff
@@ -52,12 +53,30 @@ error[SyntaxError]: expected a property, a shorthand property, a getter, a sette
 2 │   @readonly
   │   ^ Expected a property, a shorthand property, a getter, a setter, or a method here
 
+error[SyntaxError]: expected an expression but instead found '@'
```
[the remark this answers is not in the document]
ematipico: The usage of decorators inside JS files is really minimal. Also `mobx` decided to move away from them  (it's one of the most popular JS libs that use decorators, that I am aware of). So I wouldn't too worried about them for now.

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+                skipped_trailing.push(piece);
+            }
+        }
+
+        // Compute the offsets relative to the tokens text
+        let relative_skipped_range = skipped_range - token.text_range().start();
+        let text = &token.text()[relative_skipped_range];
+        let skipped_content = Token::new_dynamic(text.to_string(), skipped_range);
+
+        let skipped = FormatElement::Verbatim(Verbatim::new(
```
leops, line 430: The purpose of `FormatElement::Verbatim` was originally to track ranges that were left unformatted because the formatter didn't have a specialized implementation of `ToFormatElement` for that node: other verbatim ranges such as unknown nodes or suppressed nodes do not create a verbatim format element for instance. Now this case here is interesting skipped trivias are also a form of syntax elements that are not implemented, except they're not just unimplemented at the formatter level but also in the grammar and parser. In that sense I'm not entirely certain it makes sense to tag those range as missing some formatting since the root cause runs deeper than just the formatter

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -111,19 +111,41 @@ pub(super) fn parse_class_expression(p: &mut Parser) -> ParsedSyntax {
 // test_err ts typescript_abstract_classes_invalid_abstract_constructor
 // abstract class A { abstract constructor();};
 
+// test ts ts_decorate_computed_member
```
ematipico, line 114: Don't hate me, but there's another case we should cover: decorators for parameters:

```ts
class B {
	doGet(@Params params) {}
}
```

Here's an official example: https://docs.nestjs.com/controllers#request-object

crates/rslint_parser/src/syntax/class.rs
```diff
@@ -183,7 +201,9 @@ impl From<ClassKind> for JsSyntaxKind {
 
 // test ts ts_class_named_abstract_is_valid_in_ts
 // class abstract {}
-fn parse_class(p: &mut Parser, m: Marker, kind: ClassKind) -> CompletedMarker {
+#[inline]
```
ematipico, line 204: Any particular reason for inlining this function?

crates/rslint_parser/src/token_source.rs
```diff
@@ -245,6 +245,24 @@ impl<'l> TokenSource<'l> {
         }
     }
 
+    /// Skips the current token as skipped token trivia
+    pub fn skip_as_trivia(&mut self, context: LexContext) {
```
ematipico, line 249: I would call it `skip_token`, since it's a public API. The parser should not know anything about trivias (but token source does, which makes sense)

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+    /// If called on a token that does not have skipped trivia
+    fn format_leading_with_skipped(
+        &self,
+        token: &SyntaxToken,
+        trim_mode: TriviaPrintMode,
+        has_trailing_newline: bool,
+    ) -> FormatElement {
+        let mut pieces = token.leading_trivia().pieces().peekable();
+
+        // Collect the leading trivia up to the first skipped token trivia
```
ematipico, line 373:
```suggestion
        // Collect the leading trivia up to the first skipped token trivia (not included)
```

crates/rome_formatter/src/formatter.rs
MichaReiser: Oh, I didn't know that. Let me use `Token` then.

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+        let first_skipped = pieces.next().expect("Expected at least one skipped trivia. Don't call this method for tokens that don't have skipped trivia.");
+
+        let mut skipped_range = first_skipped.text_range();
+        let mut skipped_trailing = vec![];
+        let mut token_leading = vec![];
+        let mut after_newline = false;
+
+        for piece in pieces {
+            if piece.is_skipped() {
+                // pieces have been between to skipped trivia, thus, it's neither leading nor trailing trivia.
```
ematipico, line 402: between what? It seems something is missing...

ematipico (changes requested):
I have had hard time to understand the algorithm, mainly because the name of variables and comments sometimes are misleading and confusing. I left some suggestion on how to rephrase some internal concept, in order to avoid overlap and create confusion.

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+    /// * Skipped trailing trivia: Any trivia after last skipped token up to, but not including, the first line break
+    /// * Token leading trivia: The leading trivia of the token
+    ///
+    /// ## Returns
+    /// The format element for the tokens leading trivia.
+    ///
+    /// ## Panics
+    ///
+    /// If called on a token that does not have skipped trivia
+    fn format_leading_with_skipped(
```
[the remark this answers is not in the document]
MichaReiser: The leading is important because it only formats the leading trivia of a token, but not the trailing trivia of a token. 

The leading inside the function can be misleading as the documentation explains. There are multiple "leading" parts. I do like to keep the `leading` and `trailing` because that's what they conceptually are.

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+            }
+
+            // SAFETY: Safe because of the peek in the while condition.
+            skipped_leading_pieces.push(pieces.next().unwrap());
+        }
+
+        // Format the leading of the skipped tokens
+        let skipped_leading = self
+            .format_leading_trivia_pieces(
+                skipped_leading_pieces.into_iter(),
```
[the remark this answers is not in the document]
MichaReiser: This, unfortunately, doesn't work because the iterator must be `ExactSizeIterator + DoubleEndedIterator`.

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+        trim_mode: TriviaPrintMode,
     ) -> FormatElement {
-        let mut line_count = 0;
-        let mut elements = Vec::new();
-
         // Checks whether the previous token has any trailing newline
         let has_trailing_newline = token
             .prev_token()
-            .and_then(|token| token.trailing_trivia().pieces().next_back())
+            .and_then(|token| token.trailing_trivia().pieces().last())
```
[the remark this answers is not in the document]
MichaReiser: I added a last implementation

crates/rslint_parser/src/parser.rs
```diff
@@ -383,6 +389,20 @@ impl<'s> Parser<'s> {
         }
     }
 
+    /// Allows parsing an unsupported syntax as skipped token trivia.
+    pub fn parse_as_skipped_trivia<P>(&mut self, parse: P)
```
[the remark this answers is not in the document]
MichaReiser: I think the concept of trivia is relevant here. Because what it says is... OK, don't parse whatever comes here as normal nodes and tokens but instead parse it as a trivia. We can change it to `parse_as_skipped_trivia_tokens`.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: This already works as you can see in the decorator test file.

crates/rslint_parser/src/syntax/class.rs
MichaReiser: Mainly to motivate the parser to inline all `kind` match code.

crates/rslint_parser/src/token_source.rs
MichaReiser: I think using trivia here makes sense. Because it skips the current token as a trivia. It otherwise makes the impression as if we would just drop the token...

crates/rome_formatter/src/formatter.rs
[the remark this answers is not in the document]
leops: Oh yeah of course, I forgot how convoluted that algorithm is. Sorry about that ...

crates/rome_formatter/src/formatter.rs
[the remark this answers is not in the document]
ematipico: It might be me then

crates/rome_formatter/src/formatter.rs
```diff
@@ -330,27 +330,161 @@ impl Formatter {
...
+    fn format_leading_with_skipped(
+        &self,
+        token: &SyntaxToken,
+        trim_mode: TriviaPrintMode,
+        has_trailing_newline: bool,
+    ) -> FormatElement {
+        let mut pieces = token.leading_trivia().pieces().peekable();
+
+        // Collect the leading trivia up to the first skipped token trivia
+        let mut skipped_leading_pieces = vec![];
```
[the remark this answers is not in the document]
MichaReiser: Recommendation: The skipped means these are the leading trivia of the skipped token. `skipped_token_leading_trivia`?

crates/rome_formatter/src/formatter.rs
ematipico: It's written `to` and not `two` 😅

crates/rome_formatter/src/formatter.rs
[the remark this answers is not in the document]
MichaReiser: I restructured the code so that it only needs two collections: One for `trailing` and another for `leading`. These are now re-used between skipped token trivia and the leading trivia of the token. It also ensures that once over all pieces.

crates/rome_formatter/src/formatter.rs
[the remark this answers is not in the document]
ematipico: We can't even use `token` because there are no tokens involved... that's why I was suggesting `pieces_before_skipped_token`, something like that. Another alternative is `skipped_trivia_leading_trivia` (not a fan but it's coherent with the premises). Or I don't know... let's just leave it as is. Can't find better names 😆

crates/rslint_parser/src/parser.rs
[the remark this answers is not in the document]
ematipico: I guess it makes sense to leave it like this

crates/rslint_parser/src/syntax/class.rs
MichaReiser: Parameters were actually the main reason why I started exploring parsing into trivia. Mainly because they're just awkward :D 

https://github.com/rome/tools/pull/2253/files#diff-4af2ea35f1def1288547c1bfbee8e414ce026976e4f6daaf56f0be9c953c0909

crates/rome_formatter/src/formatter.rs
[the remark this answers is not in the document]
MichaReiser: I refactored the code so it's now gone, or better, called `leading_trivia` and `trailing_trivia` and the trivia either belongs to the skipped token trivia or the following token.

ematipico (approved):

crates/rslint_parser/src/syntax/class.rs
MichaReiser: This is a test in our test suite? I added it extra for you because I know that's what you prefer

PR #2257, opened 2022-03-18
feat(rome_console): create the rome_console and rome_markup crates

leops:
## Summary

This PR adds two new crates to the repository, `rome_console` and `rome_markup`

`rome_markup` implements a `markup` procedural macro inspired by the "template tag" of the same name in Rome JS, allowing formatted text to be expressed with a JSX-like syntax:

```rust
rome_console::markup! {
    <Info>"Formatted {count} files in {duration:?}"</Info>
}
```

`rome_console` implements the runtime-part of the markup (the actual structures being generated by the `macro` macro), as well as a `Console` trait used by the CLI to manage printing its output. This trait has two implementations: `WriteConsole` prints to any stream implementing `termcolor::WriteColor` (and has an `EnvConsole` type alias that prints to stdout / stderr), and `BufferConsole` simply stores all messages / diagnostics to an in-memory buffer (and is generally aimed at writing tests for the CLI)

## Test Plan

- Due to how the crates are structured, the test for `rome_markup` live in the `rome_console` crate (there's just a single test case for now, to verify it works)
- The test for the formatting CLI now verifies the command actually prints a line to the console, but does not verify the actual output since I'm not entirely certain how we would want to check it (`insta` snapshots ?)

### DIFF
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -112,9 +112,15 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
 
     let duration = start.elapsed();
     let count = formatted.load(Ordering::Relaxed);
-    session.app.console.message(rome_console::markup! {
-        <Info>"Formatted {count} files in {duration:?}"</Info>
-    });
+    if is_check {
+        session.app.console.message(rome_console::markup! {
+            <Info>"Checked {count} files in {duration:?}"</Info>
+        });
+    } else {
+        session.app.console.message(rome_console::markup! {
+            <Info>"Formatted {count} files in {duration:?}"</Info>
+        });
+    }
 
     let mut has_errors = false;
     let mut file_ids = HashSet::new();
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -13,8 +13,7 @@ use crossbeam::channel::{unbounded, Sender};
 use rome_core::App;
 use rome_diagnostics::{
     file::{FileId, Files, SimpleFile},
-    termcolor::{ColorChoice, StandardStream},
-    Diagnostic, Emitter,
+    Diagnostic,
 };
 use rome_formatter::{FormatOptions, IndentStyle};
 use rome_fs::{AtomicInterner, PathInterner, RomePath};
@@ -113,7 +112,9 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
 
     let duration = start.elapsed();
     let count = formatted.load(Ordering::Relaxed);
-    println!("Formatted {count} files in {duration:?}");
+    session.app.console.message(rome_console::markup! {
+        <Info>"Formatted {count} files in {duration:?}"</Info>
+    });
 
     let mut has_errors = false;
     let mut file_ids = HashSet::new();
@@ -163,15 +164,8 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
         files.storage.insert(file_id, SimpleFile::new(name, source));
     }
 
-    {
-        // Only lock stderr once for printing all the diagnostics
-        let out = StandardStream::stderr(ColorChoice::Always);
-        let mut out = out.lock();
-
-        let mut emit = Emitter::new(&files);
-        for diag in diagnostics {
-            emit.emit_with_writer(&diag, &mut out).unwrap();
-        }
+    for diag in diagnostics {
+        session.app.console.diagnostic(&files, &diag);
     }
 
     // Formatting emitted error diagnostics, exit with a non-zero code
@@ -211,9 +205,9 @@ impl Files for PathFiles {
 }
 
 /// Context object shared between directory traversal tasks
-struct FormatCommandOptions<'a> {
+struct FormatCommandOptions<'ctx, 'app> {
     /// Shared instance of [App]
-    app: &'a App,
+    app: &'ctx App<'app>,
     /// Options to use for formatting the discovered files
     options: FormatOptions,
     /// Boolean flag storing whether the command is being run in check mode
@@ -223,12 +217,12 @@ struct FormatCommandOptions<'a> {
     /// File paths interner used by the filesystem traversal
     interner: AtomicInterner,
     /// Shared atomic counter storing the number of formatted files
-    formatted: &'a AtomicUsize,
+    formatted: &'ctx AtomicUsize,
     /// Channel sending diagnostics to the display thread
     diagnostics: Sender<Diagnostic>,
 }
 
-impl<'a> FormatCommandOptions<'a> {
+impl<'ctx, 'app> FormatCommandOptions<'ctx, 'app> {
     /// Increment the formatted files counter
     fn add_formatted(&self) {
         self.formatted.fetch_add(1, Ordering::Relaxed);
@@ -240,7 +234,7 @@ impl<'a> FormatCommandOptions<'a> {
     }
 }
 
-impl<'a> TraversalContext for FormatCommandOptions<'a> {
+impl<'ctx, 'app> TraversalContext for FormatCommandOptions<'ctx, 'app> {
     fn interner(&self) -> &dyn PathInterner {
         &self.interner
     }
@@ -298,12 +292,12 @@ fn handle_file(ctx: &FormatCommandOptions, path: &Path, file_id: FileId) {
     }
 }
 
-struct FormatFileParams<'a> {
-    app: &'a App,
+struct FormatFileParams<'ctx, 'app> {
+    app: &'ctx App<'app>,
     options: FormatOptions,
     is_check: bool,
     ignore_errors: bool,
-    path: &'a Path,
+    path: &'ctx Path,
     file_id: FileId,
 }
 
--- a/crates/rome_console/src/lib.rs
+++ b/crates/rome_console/src/lib.rs
@@ -1,7 +1,7 @@
 use std::panic::RefUnwindSafe;
 
 use rome_diagnostics::{file::Files, Diagnostic, Emitter};
-use termcolor::{ColorChoice, NoColor, StandardStream, StandardStreamLock, WriteColor};
+use termcolor::{ColorChoice, NoColor, StandardStream, WriteColor};
 
 mod markup;
 
@@ -43,24 +43,14 @@ where
 }
 
 /// Type alias of [WriteConsole] printing to the standard output
-pub type EnvConsole = WriteConsole<StandardStreamLock<'static>, StandardStreamLock<'static>>;
+pub type EnvConsole = WriteConsole<StandardStream, StandardStream>;
 
 impl EnvConsole {
     /// Creates an instance of WriteConsole writing to the standard output
-    ///
-    /// This locks the stdout and stderr for the lifetime of the console instance,
-    /// so any attemps to write to the standard output (using println or panic)
-    /// may hang indefinitely if it happens from a different thread (the thread
-    /// owning the WriteConsole could be fine if the lock is reentrant)
     pub fn from_env() -> Self {
-        lazy_static::lazy_static! {
-            static ref STDOUT: StandardStream = StandardStream::stdout(ColorChoice::Always);
-            static ref STDERR: StandardStream = StandardStream::stderr(ColorChoice::Always);
-        }
-
         Self {
-            out: STDOUT.lock(),
-            err: STDERR.lock(),
+            out: StandardStream::stdout(ColorChoice::Always),
+            err: StandardStream::stderr(ColorChoice::Always),
         }
     }
 }
--- a/crates/rome_console/src/lib.rs
+++ b/crates/rome_console/src/lib.rs
@@ -1,13 +1,16 @@
 use std::panic::RefUnwindSafe;
 
+use markup::MarkupPrinter;
 use rome_diagnostics::{file::Files, Diagnostic, Emitter};
-use rome_diagnostics::termcolor::{ColorChoice, NoColor, StandardStream, StandardStreamLock, WriteColor};
+use termcolor::{ColorChoice, NoColor, StandardStream, StandardStreamLock, WriteColor};
 
 mod markup;
 
 pub use self::markup::{MarkupElement, MarkupNode};
 pub use rome_markup::markup;
 
+/// Generic abstraction over printing markup and diagnostics to an output,
+/// which can be a terminal, a file, a memory buffer ...
 pub trait Console: Sync + RefUnwindSafe {
     /// Prints a message (formatted using [markup]) to the console
     fn message(&mut self, args: MarkupNode);
--- a/crates/rome_console/src/lib.rs
+++ b/crates/rome_console/src/lib.rs
@@ -1,0 +1,96 @@
+use std::panic::RefUnwindSafe;
+
+use rome_diagnostics::{file::Files, Diagnostic, Emitter};
+use rome_diagnostics::termcolor::{ColorChoice, NoColor, StandardStream, StandardStreamLock, WriteColor};
+
+mod markup;
+
+pub use self::markup::{MarkupElement, MarkupNode};
+pub use rome_markup::markup;
+
+pub trait Console: Sync + RefUnwindSafe {
+    /// Prints a message (formatted using [markup]) to the console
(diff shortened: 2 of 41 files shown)


leops:
This is sort of what's happening already, in that `duration` is an `std::time::Duration` struct whose implementation of `std::fmt::Debug` looks at the actual value of the struct to device what time unit (ns / µs / ms / ...) it should use for display. But we could absolutely have a newtype `Duration` struct on top of that to further customize how we display timings to the user.

For now I opted to build the printing of values-to-string on top of `std::fmt`, which works but has the disadvantage that the values being printed cannot introduce any new markup if they want to, so the exact mechanism by which newtype wrappers get printed to `MarkupNode` could potentially change in the future

ematipico (changes requested):
I am having hard time to understand the usage of the new markup. Would it be possible to add a `README.md` that explains which crate should be used to actually imported and used to create the markup?

Also, I think there's some issue that could cause some misunderstanding. For example `rome_console` uses a generic term, which make sense, although we use `termcolor` inside it, which means that the abstraction of the console is not as abstract as I would expect. 

I think it makes sense for the time being, although I would add some`TODO` or comment that explains why it's like this and what to we have to do to make it generic and specialized.  For example, it makes sense to have `Console` as generic trait but I would expect to have the struct named `Terminal` in case of **this** implementation.

crates/rome_console/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+use rslint_errors::{
+    file::Files,
+    termcolor::{ColorChoice, NoColor, StandardStream, StandardStreamLock, WriteColor},
+    Diagnostic, Emitter,
+};
+
+mod markup;
+
+pub use self::markup::{MarkupElement, MarkupNode};
+pub use rome_markup::markup;
```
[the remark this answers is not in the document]
leops: The reason there are two different crates is because `rome_markup` is a "proc-macro" crate, which is essentially a plugin for the Rust compiler so the code it contains is executed at build time and does not exists anymore at runtime, while the `rome_console` create is a normal library exporting the types used in the code generated by the `markup` macro. This is the reason the macro is re-exported by the console crate and the tests for the macro live inside `rome_console`, since the macro can't be used without the supporting library.

leops:
> Also, I think there's some issue that could cause some misunderstanding. For example `rome_console` uses a generic term, which make sense, although we use `termcolor` inside it, which means that the abstraction of the console is not as abstract as I would expect.

The name `termcolor` is somewhat misleading as the `WriteColor` trait is not only implemented over terminals but also for pretty much anything that implements `std::io::Write` using the `NoColor` wrapper, so that also covers byte buffers, files or TCP streams for instance (and is the reason that implementation of `Console` is called `WriteConsole`)

> For the markup; is this something like https://github.com/dioxusLabs/rink?

This current implementation is much more limited, and is mostly focused on printed formatted text at the moment without any layout logic.

ematipico:
> > Also, I think there's some issue that could cause some misunderstanding. For example `rome_console` uses a generic term, which make sense, although we use `termcolor` inside it, which means that the abstraction of the console is not as abstract as I would expect.
> 
> The name `termcolor` is somewhat misleading as the `WriteColor` trait is not only implemented over terminals but also for pretty much anything that implements `std::io::Write` using the `NoColor` wrapper, so that also covers byte buffers, files or TCP streams for instance (and is the reason that implementation of `Console` is called `WriteConsole`)

It doesn't sound right. Do you plan to rename `termcolor` in something else?

ematipico (approved):

leops:
> > > Also, I think there's some issue that could cause some misunderstanding. For example `rome_console` uses a generic term, which make sense, although we use `termcolor` inside it, which means that the abstraction of the console is not as abstract as I would expect.
> > 
> > 
> > The name `termcolor` is somewhat misleading as the `WriteColor` trait is not only implemented over terminals but also for pretty much anything that implements `std::io::Write` using the `NoColor` wrapper, so that also covers byte buffers, files or TCP streams for instance (and is the reason that implementation of `Console` is called `WriteConsole`)
> 
> It doesn't sound right. Do you plan to rename `termcolor` in something else?

I could change the cargo configuration to alias `termcolor` to another name, but ultimately since its the name of an external crate it might create more confusion than it's worth

crates/rome_console/src/lib.rs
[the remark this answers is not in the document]
MichaReiser: Would it make sense to rename the current `rome_markup` to make it clear it contains to proc macro? 

I think it would also be nice to see how one can construct markup without having to use the proc-macro. Ideally, the `Markup` would be an independent format that isn't tight to a console so that it can be used elsewhere as well.

crates/rome_markup/README.md
```diff
@@ -0,0 +1,12 @@
+# `rome_markup`
+
+The crate contains procedural macros to build `rome_console` markup object with a JSX-like syntax
+
+The macro cannot be used alone as it generates code that requires supporting types declared in the
+`rome_console` crate, so it's re-exported from there and should be used as `rome_console::markup`
```
MichaReiser, line 6: I think it would help a lot to add a few examples here of what's possible. That would also help reviewing this PR.

crates/rome_console/src/markup.rs
```diff
@@ -0,0 +1,174 @@
+use core::fmt;
+use std::io;
+
+use termcolor::{Color, ColorSpec, WriteColor};
+
+/// Manages the state of an object implementing [WriteColor], tracking the
+/// successive styles being applied to the printer, and restoring the previous
+/// style when the current one goes out of scope
+pub(crate) struct MarkupPrinter<W: WriteColor> {
```
MichaReiser, line 9: I think it would help to split the printer and the markup into two separate modules because the markup should be independent of the printer. There could be many different printers for the same markup and the `ColorPrinter` is just one version of it.

crates/rome_console/src/markup.rs
```diff
@@ -0,0 +1,174 @@
...
+/// an element containing zero or more other nodes
+///
+/// Text nodes are formatted lazily by storing an [fmt::Arguments] struct, this
+/// means [MarkupNode] shares the same restriction as the values returned by
+/// [format_args] and can't be stored in a `let` binding for instance
+#[derive(Copy, Clone, Debug)]
+pub enum MarkupNode<'fmt> {
+    Text(fmt::Arguments<'fmt>),
+    Element {
+        kind: MarkupElement,
```
MichaReiser, line 126: [ProseMirror](https://prosemirror.net/docs/guide/#doc) uses an interesting format to store the formatting of an attribute. The main benefit is that it reduces nesting (fewer heap allocations) and it's also easier to later inspect the tree and make changes to it. Also see [this prototype](https://github.com/rome/rusty-rome/blob/main/rome-diagnostics/src/markup3/mod.rs)

We may still need something like an `Element` to implement Links which will be needed when linking to documentation.

crates/rome_console/src/markup.rs
```diff
@@ -0,0 +1,174 @@
...
+                        None => break Ok(()),
+                    }
+                };
+
+                kind.pop_style(fmt)?;
+
+                result
+            }
+        }
+    }
```
MichaReiser, line 173: I personally would avoid the dependency of the element on the printer so that it can support multiple printers without the need of dynamic dispatch (much like the formatter IR is independent of the printer)

crates/rome_core/src/lib.rs
```diff
@@ -14,10 +16,11 @@ struct Features {
     unknown: UnknownFileHandler,
 }
 
-pub struct App {
-    pub fs: Box<dyn FileSystem>,
+pub struct App<'app> {
```
MichaReiser, line 19: Nit: Is it worth adding the lifetime here or should we use `Box'ed `fs`/`console` type to avoid the somewhat complicated lifetimes that start to spread through all of the codebase?

MichaReiser (commented):
This is nice work! Especially the proc macro with all that error handling! 

In my view, it would be worth to explicitly separating `Markup` from the `MarkupPrinter` to allow different printer implementations (printing to HTML? VS Code?) 

I think it could also make sense to have a look at [this prototype exploration](https://www.notion.so/rometools/Markup-fd1362ab03b04acbafe3d48051f86d42) of Markup formats, particularly at option 3 that is inspired by ProseMirror's representation.

crates/rome_console/src/markup.rs
leops: The `MarkupPrinter` object is an implementation detail of the markup and is private to the crate, it's generic over the type of the underlying output being printed to and simply wraps it with state management utility functions similarly to the `std::fmt::Formatter` type used in the `Debug` and `Display` traits

crates/rome_console/src/markup.rs
leops: I've setup the `MarkupNode` enum (and accompanying `markup` macro) in a similar way to the `std::fmt::Arguments` struct (and accompanying `format_args` macro) so that it doesn't heap-allocate anything, the entire markup lives on the stack for the duration of the call to `Console::message`. In that sense reducing the amount of nesting probably wouldn't change much in terms of allocation performance, but I agree it would make the later traversal more efficient and shouldn't be too hard to implement in the `markup` macro.

Overall, I think the intent would be for the `MarkupNode` enum to be "public by necessity" (since it needs to be constructed from other crates), but should not be constructed by anything other than the `markup` macro (again similarly to format arguments, but we can't use the same "visibility breaking" trick the compiler has access to) so that we could change the internal representation of markup if the needs arise without having to refactor huge amounts of code.

crates/rome_core/src/lib.rs
leops: While the normal execution mode of the CLI can work with owned trait objects in `Box` (and in fact does exactly that), unfortunately `Box<dyn Trait>` makes it impossible to recover the original type which is important for tests, where we only want to provide the CLI with a mutable borrow of the underlying trait object so that it can be inspected after the `run_cli` function has completed. For the most part it shouldn't complicate the codebase too much as we can simply pass borrows of the app object around as `&App` and let the compiler guess the lifetime, the only exception to this is when references to the app are stored in structures as the lifetime need to be specified explicitly (but since the app is often threaded between functions by reference instead of moved that reference would need a lifetime too anyway)

crates/rome_console/src/markup.rs
leops: To be clear there is no dynamic dispatch here (the signature is using `impl WriteColor` instead of `dyn WriteColor`) so this function gets specialized for each printer type its used with (in the case of `NoColor<impl Write>` it doesn't do anything special and just prints the text for instance)

leops:
I reworked the way markup is internally represented taking inspiration of how it's done in ProseMirror. Markup is now a flat slice of nodes each containing a piece of text as well as the entire stack of elements being applied to it, and this completely eliminates the need for the `MarkupPrinter` type so I removed it

crates/rome_cli/src/commands/format.rs
```diff
@@ -113,7 +112,15 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
 
     let duration = start.elapsed();
     let count = formatted.load(Ordering::Relaxed);
-    println!("Formatted {count} files in {duration:?}");
+    if is_check {
+        session.app.console.message(rome_console::markup! {
+            <Info>"Checked {count} files in {duration:?}"</Info>
```
MichaReiser, line 117: Thx

crates/rome_console/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+/// Type alias of [WriteConsole] printing to the standard output
+pub type EnvConsole = WriteConsole<StandardStreamLock<'static>, StandardStreamLock<'static>>;
+
+impl EnvConsole {
+    /// Creates an instance of WriteConsole writing to the standard output
+    ///
+    /// This locks the stdout and stderr for the lifetime of the console instance,
+    /// so any attemps to write to the standard output (using println or panic)
+    /// may hang indefinitely if it happens from a different thread (the thread
+    /// owning the WriteConsole could be fine if the lock is reentrant)
```
MichaReiser, line 54: Isn't this somewhat unexpected? This requires that the console scope must be as short-lived as possible. 

I think it would help to make this decision at least visible in the API by either using a different method name

MichaReiser (approved):
Looks good to me, thanks for reworking the dependencies. 

Could you please add some examples to the README or somewhere else on how to use the Markup and what is supported? 

How would you envision the markup being used by a generic Diagnostic crate? And how would elements like hyperlinks work?

crates/rome_console/src/lib.rs
leops: This is probably the bit I'm the most unsure of (but fortunately is also incredibly easy to remove): in practice the console scope is far from short, in fact it spans at least the entire run of the CLI. It's based on the idea that nothing should be using println anyway, so it's okay to lock the standard streams for the duration of the CLI as an optimization instead of locking and unlocking the stream for every printed line because nothing should contend for the lock. But while that's true for release builds, println is still useful as a quick and easy way to debug things deep into the codebase, plus using the standard output while the console is active doesn't raise an error but silently deadlocks indefinitely so that's not very intuitive to debug either.

leops:
> How would you envision the markup being used by a generic Diagnostic crate? And how would elements like hyperlinks work?

For diagnostics I think on the basic level the current builder methods that take strings (`primary`, `secondary`, `label`, ...) would be modified to take a `Markup` element instead. How that markup would be handled internally is another problem: if the diagnostic is built up and immediately sent to the console (`console.diagnostic(Diagnostic::error())`) then the `Diagnostic` struct could simply hold onto the `Markup` it's been provided with, however in pretty much all the existing usage of diagnostics they get stored into an intermediate vector and these would need to convert the `Markup` into an owned form to be allowed by the borrow checker.

Hyperlinks would work by adding a new `Hyperlink { href: &str }` variant to the `MarkupElement` enum (and adding a syntax for props to the `markup` macro), what I'm not entirely sure about is how to detect platforms that don't support those and what to print as a fallback: currently all the existing markup elements only carry formatting information and are imply ignored on platforms that don't support them

crates/rome_console/src/lib.rs
ematipico: If we assume that we want to lock the console for the whole time, shouldn't we at least provide an alternative to `println`? As you said, a developer would like to use that method for debugging purpose. 

> doesn't raise an error but silently deadlocks indefinitely so that's not very intuitive to debug either.

Shouldn't we put this in the `README.md`? This is somewhat important and developers should know what's going on if something like this happens.

ematipico:
I am a bit concerned about locking the whole console. Let's think that "console" here is a generic term. It can be any output stream: terminal, VS Code output, web browser developer tools, etc.

I know that, for now, we only support the CLI, but what about other streams that might be "shared" with other programs? Is it going to be fine? 

We should carefully understand the implication of this architecture and provide a troubleshooting section in cases something strange happens. Given these premises, I honestly think that, for now, we should expose the `markup` feature as **unstable** or *opt-in* via feature system.

MichaReiser:
> > I know that, for now, we only support the CLI, but what about other streams that might be "shared" with other programs? Is it going to be fine?
> 
> Since the Console API is mutable any other kind of stream we might want to create will simply be owned by the console object and will not require any locking logic (or that lock would have to be put in place by the code using the console), the locking logic only exists when printing to stdout or stderr due to how those are implemented in the Rust standard library: the lock exists purely in userspace to ensure multiple simultaneous calls to println can't overlap each other and does not actually lock the file descriptor with the OS. Taking a lock on the streams for the whole run of the CLI is simpler than (implicitly) locking the stream for every message that gets printed, but if it's causing too much trouble I can just remove that code

I think it makes sense. It might be useful to point it out in the `README` advising devs to use the console for printing any debug message. We can re-visit the decision in the future and either remove it or add a concept like "Transaction" or "scope" to the CLI.

ematipico:
> but if it's causing too much trouble I can just remove that code

No need! We can leave it like this! The thing, I haven't started using the crate yet so I don't fully know how it is going to play with the programs that will consume it, so I suppose some skepticism is normal! As you said, changing it is trivial, so we can always change it later.

leops:
Oh well I already removed the lock since I wasn't really sure about it anyway, I don't think the standard library having to acquire the lock on the path to every console write has a significant performance impact considering the code is going to be doing IO right after

ematipico (approved):
Let's 🚢  it and start playing with it! I pumped by this new feature! Great work @leops !!

PR #2258, opened 2022-03-18
refactor(xtask): Factored out benchmarking into separate crate

NicholasLYang:
…marker crate. Will make it easier to benchmark on a worker

## Summary

In order to add automatic benchmarking, I decided to factor out the benchmarking logic in `xtask/bench` into a separate crate. That way, we can use `rome_benchmarker` for both xtask and for a benchmarking service.

Next step after this is to have a service that runs benchmarks on workers or a server. My idea is to spin up a bunch of Cloudflare Workers to benchmark in parallel. That'll allow us to get results pretty quickly and run benchmarks consistently.


MichaReiser:
How do the requirements of a worker/service differ that they can't call into the CLI? 

I would also keep this crate in the `xtask` folder and not call it `rome_benchmarker. I could see us adding a `benchmark` crate in the future that allows you to benchmark some JS code (similar to `cargo bench`) and it would be a pitty if that name's already taken. Also: `xtask`: Internal toolings, `crates`: Public crates that are useful to any user.

ematipico (changes requested):
Could you please add to the description **why** you decided to apply this refactor? This benchmark crate you created is strictly for our own internal tool and it can be misleading because we are not benchmarking... JS/TS files.

I fail to understand why this change was needed in order to achieve the benchmarking with cloudflare workers.

ematipico:
@NicholasLYang is the PR still valid?

NicholasLYang:
I'm going to take another look at benchmarking today in preparation for a post comparing us, prettier, and dprint. I'll see if I can convert this PR into something that can work with that

PR #2264, opened 2022-03-21
refactor(rome_rowan): Reduce Trivia Memory footprint

MichaReiser:
## Summary

This PR reduces the memory footprint for storing trivia by:

* Caching and re-using trivia across tokens
* Reducing the size of `GreenTrivia` from `16` to `8` bytes which, in turn, reduces the size of `GreenTokenData` from 36 bytes to 20 bytes. That means, we save 16 bytes for every allocated token. 
* Reducing the size for a "many" trivia from: 32 (8 byte pointer + tag, 16 byte vec + trivia) to 16 bytes (8 byte pointer, 8 byte len + trivia)

Downside:

* This now requires a heap allocation for trivia containing a single piece which is, for example, very common for whitespace. This comes with a slight memory increase because of the heap metadata but that is amortized by caching the trivia across multiple tokens. 

Related discussion #1809

## What has changed since the initial implementation?

The optimization for storing a single trivia inline with the token has become less effective since we decided to have two different trivia for 'newlines' and 'whitespace' (space, tabs). Splitting the trivia had the consequence that now any indented token at the beginning of a line requires a heap allocated `Vec` to store the leading line break + the indent spaces. 

## Test Plan

Performance is neutral. 

`cargo bench_parser`

```
group                                    green2                                 main
-----                                    ------                                 ----
parser/checker.ts                        1.01     84.4±1.66ms    30.8 MB/sec    1.00     83.6±1.29ms    31.1 MB/sec
parser/compiler.js                       1.00     50.8±1.52ms    20.6 MB/sec    1.01     51.1±1.16ms    20.5 MB/sec
parser/d3.min.js                         1.00     32.5±0.54ms     8.1 MB/sec    1.00     32.4±0.47ms     8.1 MB/sec
parser/dojo.js                           1.00      2.6±0.04ms    26.0 MB/sec    1.01      2.7±0.06ms    25.7 MB/sec
parser/ios.d.ts                          1.00     78.1±1.45ms    23.9 MB/sec    1.01     78.9±1.36ms    23.6 MB/sec
parser/jquery.min.js                     1.01      9.3±0.13ms     8.9 MB/sec    1.00      9.3±0.13ms     8.9 MB/sec
parser/math.js                           1.00     62.0±1.37ms    10.4 MB/sec    1.12     69.6±3.44ms     9.3 MB/sec
parser/parser.ts                         1.00  1761.2±36.09µs    26.2 MB/sec    1.06  1870.9±156.93µs    24.7 MB/sec
parser/pixi.min.js                       1.00     39.5±0.66ms    11.1 MB/sec    1.00     39.5±0.74ms    11.1 MB/sec
parser/react-dom.production.min.js       1.00     12.0±0.17ms     9.6 MB/sec    1.01     12.2±0.14ms     9.5 MB/sec
parser/react.production.min.js           1.00   578.1±10.38µs    10.6 MB/sec    1.00    579.5±8.05µs    10.6 MB/sec
parser/router.ts                         1.00  1480.2±33.52µs    41.0 MB/sec    1.01  1500.2±27.17µs    40.4 MB/sec
parser/tex-chtml-full.js                 1.01     84.1±1.74ms    10.8 MB/sec    1.00     83.5±1.34ms    10.9 MB/sec
parser/three.min.js                      1.01     44.6±0.70ms    13.2 MB/sec    1.00     44.4±0.83ms    13.2 MB/sec
parser/typescript.js                     1.00    344.9±5.11ms    27.5 MB/sec    1.03    357.0±8.25ms    26.6 MB/sec
parser/vue.global.prod.js                1.00     14.7±0.19ms     8.2 MB/sec    1.00     14.7±0.21ms     8.2 MB/sec
```

Memory consumption: 

* `pixi`: Minified file. Only contains whitespace trivia. From 7.74MB to 7.66MB (~1%). Memory reduction comes from the reduced token size. 

* `checker`: Huge source file, unmodified: Contains whitespace, new lines, and comments: From 16.86MB to 16.50 MB (~2%)

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -28,21 +28,73 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
     type Kind: SyntaxKind;
 }
 
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Debug)]
+pub enum TriviaPieceKind {
+    /// A newline character
+    Newline,
+    /// Any whitespace character
+    Whitespace,
+    /// Comment that does not contain any line breaks
+    SingleLineComment,
+    /// Comment that contains at least one line break
+    MultiLineComment,
+}
+
+impl TriviaPieceKind {
+    pub fn is_newline(&self) -> bool {
+        matches!(self, TriviaPieceKind::Newline)
+    }
+
+    pub fn is_whitespace(&self) -> bool {
+        matches!(self, TriviaPieceKind::Whitespace)
+    }
+
+    pub fn is_single_line_comment(&self) -> bool {
+        matches!(self, TriviaPieceKind::SingleLineComment)
+    }
+
+    pub fn is_multiline_comment(&self) -> bool {
+        matches!(self, TriviaPieceKind::MultiLineComment)
+    }
+}
+
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
-pub enum TriviaPiece {
-    Newline(TextSize),        // length
-    Whitespace(TextSize),     // length
-    Comments(TextSize, bool), // length, has_newline
+pub struct TriviaPiece {
+    pub(crate) kind: TriviaPieceKind,
+    pub(crate) length: TextSize,
 }
 
 impl TriviaPiece {
+    pub fn whitespace<L: Into<TextSize>>(len: L) -> TriviaPiece {
+        Self::new(TriviaPieceKind::Whitespace, len)
+    }
+
+    pub fn newline<L: Into<TextSize>>(len: L) -> TriviaPiece {
+        Self::new(TriviaPieceKind::Newline, len)
+    }
+
+    pub fn single_line_comment<L: Into<TextSize>>(len: L) -> TriviaPiece {
+        Self::new(TriviaPieceKind::SingleLineComment, len)
+    }
+
+    pub fn multi_line_comment<L: Into<TextSize>>(len: L) -> TriviaPiece {
+        Self::new(TriviaPieceKind::MultiLineComment, len)
+    }
+
+    pub fn new<L: Into<TextSize>>(kind: TriviaPieceKind, length: L) -> Self {
+        Self {
+            kind,
+            length: length.into(),
+        }
+    }
+
     #[inline]
     pub fn text_len(&self) -> TextSize {
-        match self {
-            TriviaPiece::Newline(n) => *n,
-            TriviaPiece::Whitespace(n) => *n,
-            TriviaPiece::Comments(n, _) => *n,
-        }
+        self.length
+    }
+
+    pub fn kind(&self) -> TriviaPieceKind {
+        self.kind
     }
 }
 
@@ -92,9 +144,9 @@ impl<L: Language> SyntaxTriviaPieceComments<L> {
     }
 
     pub fn has_newline(&self) -> bool {
-        match self.0.trivia {
-            TriviaPiece::Comments(_, has_newline) => has_newline,
-            _ => unreachable!("trivia is not a comment"),
+        match self.0.trivia.kind {
+            TriviaPieceKind::MultiLineComment => true,
+            _ => false,
         }
     }
 }
@@ -108,8 +160,8 @@ impl<L: Language> SyntaxTriviaPieceComments<L> {
 /// builder.token_with_trivia(
 ///     RawSyntaxKind(1),
 ///     "\n\t /**/let \t\t",
-///     &[TriviaPiece::Whitespace(3), TriviaPiece::Comments(4, false)],
-///     &[TriviaPiece::Whitespace(3)],
+///     &[TriviaPiece::whitespace(3), TriviaPiece::single_line_comment(4)],
+///     &[TriviaPiece::whitespace(3)],
 /// );
 /// });
 ///
@@ -136,8 +188,8 @@ impl<L: Language> SyntaxTriviaPiece<L> {
     ///     builder.token_with_trivia(
     ///         RawLanguageKind::LET_TOKEN,
     ///         "\n\t /**/let \t\t",
-    ///         &[TriviaPiece::Whitespace(TextSize::from(3)), TriviaPiece::Comments(TextSize::from(4), false)],
-    ///         &[TriviaPiece::Whitespace(TextSize::from(3))],
+    ///         &[TriviaPiece::whitespace(3), TriviaPiece::single_line_comment(4)],
+    ///         &[TriviaPiece::whitespace(3)],
     ///     );
     /// });
     /// let pieces: Vec<_> = node.first_leading_trivia().unwrap().pieces().collect();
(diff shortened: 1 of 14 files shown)


crates/rome_rowan/src/green/trivia.rs
```diff
@@ -0,0 +1,157 @@
...
+/// The identity of a trivia is defined by the kinds and lengths of its items but not by
+/// the texts of an individual piece. That means, that `\r` and `\n` can both be represented
+/// by the same trivia, a trivia with a single `LINEBREAK` piece with the length 1.
+/// This is safe because the text is stored on the token to which the trivia belongs and
+/// `a\n` and `a\r` never resolve to the same tokens. Thus, they only share the trivia but are
+/// otherwise two different tokens.
+#[derive(Eq, PartialEq, Hash, Clone)]
+#[repr(transparent)]
+pub(crate) struct GreenTrivia {
+    ptr: ThinArc<GreenTriviaHead, TriviaPiece>,
```
leops, line 57: Since `ThinArc` is pretty much just a wrapper around a `NonNull` pointer, I think `ptr` could be an `Option` without changing the size of `GreenTrivia`, and the `None` case would be the empty trivia so we wouldn't have to store that case explicitly in the trivia cache.
And if we wanted to try and be very clever about it, since the data pointed to in a `ThinArc` (an `ArcInner` struct) contains an `AtomicUsize` and thus has to be aligned to the size of a pointer, we could use a "tagged pointer" representation and fit the single trivia piece case inline inside the bits of the pointer.

MichaReiser: Uh, wow... trying to get my head around what you're suggesting. 

To the first suggestion. Are you recommending changing the type of `ptr` to `Option<ThinArc<...>>` or is it that you would define `GreenTrivia` as an enum that wraps the current `GreenTrivia` type?

Regarding tagged pointers. That sounds neat, but I'm probably the wrong person to tackle that. But would be fairly interested in reviewing a PR to change it.

crates/rome_rowan/src/api.rs
```diff
@@ -28,22 +28,79 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
...
 
 impl TriviaPiece {
-    #[inline]
-    pub fn text_len(&self) -> TextSize {
-        match self {
-            TriviaPiece::Newline(n) => *n,
-            TriviaPiece::Whitespace(n) => *n,
-            TriviaPiece::Comments(n, _) => *n,
+    /// Creates a new whitespace trivia piece with the given length
+    pub fn whitespace<L: Into<TextSize>>(len: L) -> TriviaPiece {
```
ematipico, line 69: nit: Can we change all places where we have `-> TriviaPiece` to be `-> Self`?

crates/rome_rowan/src/api.rs
```diff
@@ -28,22 +28,79 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
...
+
+    /// Creates a new comment trivia piece that does not contain any line breaks
+    pub fn single_line_comment<L: Into<TextSize>>(len: L) -> TriviaPiece {
+        Self::new(TriviaPieceKind::SingleLineComment, len)
+    }
+
+    /// Creates a new comment trivia piece that contains at least one line breaks
+    pub fn multi_line_comment<L: Into<TextSize>>(len: L) -> TriviaPiece {
+        Self::new(TriviaPieceKind::MultiLineComment, len)
+    }
```
ematipico, line 86: Are the doc comments correct for these two functions? 🤔

crates/rome_rowan/src/green/trivia.rs
```diff
@@ -0,0 +1,157 @@
...
+impl ops::Deref for GreenTrivia {
+    type Target = GreenTriviaData;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        unsafe {
+            let repr: &Repr = &self.ptr;
+            let repr: &ReprThin = &*(repr as *const Repr as *const ReprThin);
+            mem::transmute::<&ReprThin, &GreenTriviaData>(repr)
+        }
```
ematipico, line 83: can we add a `// SAFETY` comment please?

crates/rome_rowan/src/green/trivia.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+impl ops::Deref for GreenTrivia {
+    type Target = GreenTriviaData;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        unsafe {
+            let repr: &Repr = &self.ptr;
+            let repr: &ReprThin = &*(repr as *const Repr as *const ReprThin);
```
ematipico, line 81: I have never seen this `as *const`. Can you explain what are we doing here?

crates/rome_rowan/src/green/trivia.rs
xunilrj: I think @leops hit bull eyes here.

```
print-type-size type: `green::trivia::GreenTrivia`: 8 bytes, alignment: 8 bytes
print-type-size     field `.ptr`: 8 bytes
print-type-size type: `green::trivia::GreenTriviaData`: 8 bytes, alignment: 8 bytes
print-type-size     field `.data`: 8 bytes
```

If we try @leops suggestion: Nothing changes:

```
print-type-size type: `green::trivia::GreenTrivia`: 8 bytes, alignment: 8 bytes
print-type-size     field `.ptr`: 8 bytes
print-type-size type: `green::trivia::GreenTriviaData`: 8 bytes, alignment: 8 bytes
print-type-size     field `.data`: 8 bytes
```

This indeed open the space to do something like Cow (https://doc.rust-lang.org/src/alloc/borrow.rs.html#181-192)

```rust
pub(crate) enum GreenTrivia {
    None,
    Shared(ThinArc<GreenTriviaHead, TriviaPiece>),
}
```

and its sizes:

```
print-type-size type: `green::trivia::GreenTrivia`: 8 bytes, alignment: 8 bytes
print-type-size     variant `Shared`: 8 bytes
print-type-size         field `.0`: 8 bytes
print-type-size type: `green::trivia::GreenTriviaData`: 8 bytes, alignment: 8 bytes
print-type-size     field `.data`: 8 bytes
```

Maybe this is even open space to remove these pointers and unsafes in the future.

MichaReiser: You still need the unsafe to work with `ThinArc`. 

I'll give it a try. I need to see how the API should work because methods like `pieces()` that return a slice no longer work (because `None` has no slice. 

I guess, that would speak in favor of making it an enum and force the callers to deal with it

xunilrj: One option, in this case, is to a static empty slice and return it.

MichaReiser: Thx for that suggestion. 

I'll need to do some profiling to see if the matching has any effect on performance.

MichaReiser:
!bench_parser

crates/rome_rowan/src/api.rs
MichaReiser: Yes, why? What is misleading? These comments don't map to JS `/* */` and `//` comments directly. 

* `//` is always a single line comment because it isn't possible for a `//` to spawn across multiple files (it's a new comment on each line)
* `/*` can either be a `single` or `multiline` comment. This depends on whatever there's any line break character inside the comment.

crates/rome_rowan/src/green/trivia.rs
MichaReiser: Oh my, I copied this code from `GreenNode`

crates/rome_rowan/src/green/trivia.rs
MichaReiser: This cast's `repr` to a `const Repr` Pointer that then get's casted to a const `ReprThin` pointer.

xunilrj (approved):

PR #2268, opened 2022-03-21
fix(tests): use a platform-independent abstraction over time measurements in the CLI timings tests

leops:
## Summary

The metrics measurement layer in the CLI makes use of the `std::time::Instant` to measure timings, that internally uses a platform-specific representation of time points on a high-resolution steady clock. This works well for doing precise measurements of timings but not so much for testing, and the tests for the metrics layer are currently failing on M1 Macs due to a lack of precision in the underlying representation.

This PR introduces a `Timepoint` trait private to the `metrics` module providing a generic abstraction over time measurements. It's implemented for `Instant` as well as `TestTime`, a newtype struct wrapping an integer and performing straightforward arithmetic operations over it, in order to guarantee reproducible results for the tests working with timings on all platforms.

### DIFF
--- a/crates/rome_cli/src/metrics.rs
+++ b/crates/rome_cli/src/metrics.rs
@@ -2,6 +2,7 @@ use std::{
     borrow::Cow,
     collections::HashMap,
     hash::Hash,
+    ops::Sub,
     ptr,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


ematipico (approved):

yassere (approved):
Thanks! I can confirm this test passes now on my M1 Mac.

PR #2269, opened 2022-03-21
doc: move contributing sections in ad-hoc files

ematipico:
## Summary

This PR:
- creates a file that explain how to test the formatter
- it adds some help on how to identify issue while writing tests
- creates a file that explain how to implement the formatting on nodes

This an initial work, there's probably more work to do for implementing part. It would be ideal to have our docs published somewhere.

## Test Plan

None


crates/rome_formatter/docs/implement_the_formatter.md
```diff
@@ -0,0 +1,71 @@
...
+
+Usually the developer doesn't have to manually write it, `rome` has an automatic code generation that does 
+everything for us when the grammar is actually implemented. By default, all nodes will format verbatim,
+meaning that the formatter will print tokens and trivia as they are.
+
+Our formatter has its own [internal IR](https://en.wikipedia.org/wiki/Intermediate_representation), it creates its own abstraction from an AST.
+
+The developer won't be creating directly this IR, but they will use a series of utilities that will help
+to create this IR. The whole IR is represented by the `enum` `FormatElement`. Please refer to its internal
+documentation to understand the meaning of each variant.
```
leops, line 14: Should we have instructions on how to build the documentation with `cargo doc` somewhere in the main contributing file, to make sure even contributors that are not very familiar with Rust know about it ?

crates/rome_formatter/docs/write_tests.md
```diff
@@ -0,0 +1,70 @@
...
+4. the `cases` keyword is mandatory;
+5. then each object of the array will contain the matrix of options you'd want to test.
+   In this case the test suite will run a **second test case** with `line_width` to 120 and `ident_style` with  4 spaces
+6. when the test suite is run, you will have two outputs in your snapshot: the default one and the custom one
+
+## Identify issues
+
+There are four cases when a test is not correct:
+- you try to print/format the same token multiple times; the formatter will check at runtime when a test is run;
+- you some tokens haven't been printed; usually you will have this information inside the snapshot, under a section
```
leops, line 65: We could provide some hints on how to resolve these issues: for instance if removing a token from the input is actually the expected behavior, the correct way to implement that is using `formatter.format_replaced(token, empty_element())`

crates/rome_formatter/docs/implement_the_formatter.md
ematipico: I thought about it, but I have no idea where to store the generated documentation. Should we try GitHub pages for the time being? Free storage already set up for us. What do you think?

ematipico: Actually, it was better than a I thought. I added the changes for the docs

PR #2273, opened 2022-03-21
feature: Added formatting for self-closing JSX element

NicholasLYang:

### DIFF
--- a/crates/rome_formatter/src/jsx/tag/self_closing_element.rs
+++ b/crates/rome_formatter/src/jsx/tag/self_closing_element.rs
@@ -1,7 +1,25 @@
-use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
-use rome_js_syntax::{AstNode, JsxSelfClosingElement};
+use crate::formatter_traits::FormatTokenAndNode;
+use crate::group_elements;
+use crate::{
+    format_elements, soft_block_indent, soft_line_break_or_space, FormatElement, FormatResult,
+    Formatter, ToFormatElement,
+};
+use rome_js_syntax::JsxSelfClosingElement;
+
 impl ToFormatElement for JsxSelfClosingElement {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let mut attributes = Vec::new();
+        for attribute in self.attributes() {
+            attributes.push(attribute.format(formatter)?);
+            attributes.push(soft_line_break_or_space());
+        }
+
+        Ok(format_elements![
+            self.l_angle_token().format(formatter)?,
+            self.name().format(formatter)?,
+            group_elements(soft_block_indent(concat_elements(attributes))),
+            self.slash_token().format(formatter)?,
+            self.r_angle_token().format(formatter)?
+        ])
     }
 }
--- a/crates/rome_formatter/src/jsx/expressions/tag_expression.rs
+++ b/crates/rome_formatter/src/jsx/expressions/tag_expression.rs
@@ -1,7 +1,9 @@
+use crate::formatter_traits::FormatTokenAndNode;
 use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
-use rome_js_syntax::{AstNode, JsxTagExpression};
+use rome_js_syntax::JsxTagExpression;
+
 impl ToFormatElement for JsxTagExpression {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        self.tag().format(formatter)
     }
 }


crates/rome_formatter/tests/specs/jsx/self_closing.jsx.snap
```diff
@@ -0,0 +1,98 @@
...
+	onChange={(evn) => {
+		setCode(evn.target.value);
+	}}
+	style={{ fontSize: 12, height: "100vh", fontFamily:
+			"ui-monospace,SFMono-Regular,SF Mono,Consolas,Liberation Mono,Menlo,monospace",
+	}}
+/>;
+
+
+## Unimplemented nodes/tokens
```
ematipico, line 75: If the snapshot has this section, it means that we are not formatting some tokens/nodes. This means that a test should be **valid** if it **doesn't have this section**.

NicholasLYang:
Do we want a proper folder structure for JSX tests? I can have it mirror the JSX formatter directory structure

ematipico (changes requested):
The snapshot shows that there are a **lot of unimplemented** nodes (they still format verbatim).

ematipico:
You can check this section: https://github.com/rome/tools/blob/main/crates/rome_formatter/docs/implement_the_formatter.md to make sure to understand when a test is correct.

crates/rome_formatter/tests/specs/jsx/self_closing.jsx
```diff
@@ -0,0 +1,28 @@
...
+	value={code}
+	language="js"
+	placeholder="Enter JS here"
+	onChange={(evn) => {
+		setCode(evn.target.value);
+	}}
+	style={{ fontSize: 12, height: "100vh", fontFamily:
+			"ui-monospace,SFMono-Regular,SF Mono,Consolas,Liberation Mono,Menlo,monospace",
+	}}
+/>
```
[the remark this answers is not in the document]
NicholasLYang: Well the formatting does include how we place attributes, so it is important that we have tests with attributes. Should I remove the non trivial ones then?

crates/rome_formatter/tests/specs/jsx/self_closing.jsx.snap
ematipico: Exactly, we don't need to format everything in one go. Let's just focus on the small feature and gradually add things along the way. Simply formatting `<Foo />` is just fine

crates/rome_formatter/tests/specs/jsx/self_closing.jsx.snap
```diff
@@ -0,0 +1,98 @@
...
+
+"Foo " => 1..5
+"Foo           " => 11..25
```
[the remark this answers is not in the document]
NicholasLYang: Hmm. I'm confused here because it does correctly format `<Foo           />;` as `<Foo />`. Perhaps it's just printing the name part (which is purposefully not formatted) with trivia?

crates/rome_formatter/tests/specs/jsx/self_closing.jsx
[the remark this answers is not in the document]
ematipico: Sure, but the PR is exclusively about self closing tags, and having formatting for attributes can be done with a different PR without dependencies, that's why I created a task just for that.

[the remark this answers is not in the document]
NicholasLYang: Hmm okay. I was mostly testing [this code](https://github.com/rome/tools/pull/2273/files#diff-3bcac4358c094c6a3656853e9c5f90780652a05bb42bf21997975b54afcdbbe8R11) namely the soft line breaks around attributes.

crates/rome_formatter/tests/specs/jsx/self_closing.jsx.snap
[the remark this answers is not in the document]
ematipico: The reason why the second `Foo        ` have the spaces removed, is because we actually trim the tokens out from the verbatim tokens (the first leading and the first trailing), and by chance the first leading trivia is a white space. You could try to insert some new lines and could see the different.

Anyway, if you check the AST:

```
tag: JsxSelfClosingElement {
    l_angle_token: L_ANGLE@0..1 "<" [] [],
    name: JsxReferenceIdentifier {
        value_token: JSX_IDENT@1..13 "Foo" [] [Whitespace("         ")],
    },
    type_arguments: missing (optional),
    attributes: JsxAttributeList [],
    slash_token: SLASH@13..14 "/" [] [],
    r_angle_token: R_ANGLE@14..15 ">" [] [],
},
```

`name` is a `JsxReferenceIdentifier` which is still formatting verbatim

crates/rome_formatter/tests/specs/jsx/self_closing.jsx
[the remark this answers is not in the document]
ematipico: We can iterate again over the line breaks when we will implement the format of attributes.

You can leave the implementation as it is, the most important part is that the test doesn't have to have JSXAttributes.

crates/rome_formatter/src/jsx/tag/self_closing_element.rs
```diff
@@ -1,7 +1,26 @@
...
+use rome_js_syntax::JsxSelfClosingElement;
+
 impl ToFormatElement for JsxSelfClosingElement {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let mut attributes = Vec::new();
+        for attribute in self.attributes() {
+            attributes.push(attribute.format(formatter)?);
+            attributes.push(soft_line_break_or_space());
+        }
```
MichaReiser, line 15: You can use `join_elements` for this together with `format_nodes`
```suggestion
        let attributes = join_elements(soft_line_break_or_space(), formatter.format_nodes(self.attributes());
```

crates/rome_formatter/tests/specs/jsx/self_closing.jsx.snap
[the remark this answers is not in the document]
NicholasLYang: I don't handle any formatting of JsxAnyElementName. Should I?

crates/rome_formatter/src/jsx/tag/self_closing_element.rs
NicholasLYang: Hmm so I switched to this with a tweak:

```
let mut attributes = Vec::new();
for attribute in self.attributes() {
      attributes.push(attribute.format(formatter)?);
      attributes.push(soft_line_break_or_space());
}
```

But then for some reason the lines always became a hard line break. I don't really know why.

crates/rome_js_formatter/src/jsx/tag/self_closing_element.rs
```diff
@@ -14,9 +15,16 @@ impl ToFormatElement for JsxSelfClosingElement {
             attributes.push(soft_line_break_or_space());
         }
 
+        let type_arguments = self
+            .type_arguments()
+            .map(|arg| arg.format(formatter))
+            .transpose()?
+            .unwrap_or_else(empty_element);
```
ematipico, line 22: `type_arguments` is of type `TsTypeArguments`, which already implements the correct formatting. 

The current implementation won't be correct because it doesn't handle correctly the grouping of its nodes. We should just do `type_arguments.format(formatter)`.

NicholasLYang: `type_arguments` is of type `Option<TsTypeArguments>` which does not implement `ToFormatElement`. That's why I map `Option<TsTypeArguments>` to `Option<FormatResult<FormatElement>>`, then transpose to `FormatResult<Option<FormatElement>>`, then ? operator to `Option<FormatElement>` then unwrap_or_else to get `FormatElement`. We could make that a generic implementation, i.e.

```
impl<T: ToFormatElement> ToFormatElement for Option<T> {

}
```

NicholasLYang: Basically, I already do `type_arguments.format(formatter)`, I just have a bunch of stuff that lets me do it in an `Option`

ematipico: `format_or_empty` is the API you want

NicholasLYang: Is there a reason we have `format_or_empty` instead of just implementing `format` for `Option<T: ToFormatElement>`?

NicholasLYang: Hmm I don't quite understand. Judging by a quick grep, we mostly use `format_or_empty`, which could then be replaced by `format` by implementing `ToFormatElement` on `Option<T: ToFormatElement>`. We can still keep `format_or` but it'd make sense to just have `format_or_empty` be the default.

ematipico: We solved the conversation offline, it was a misunderstanding of two valid points :)

crates/rome_js_formatter/tests/specs/jsx/self_closing.jsx.snap
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Unimplemented nodes/tokens
+
+"Foo " => 1..5
+"Foo           " => 11..25
+"LineWidthInput " => 31..46
+"IndentStyleSelect" => 51..68
+"SourceTypeSelect" => 75..91
+"CodeEditor" => 99..109
```
ematipico, line 49: The presence of these information in the snapshot means that we haven't implemented these nodes: https://github.com/rome/tools/blob/main/xtask/codegen/js.ungram#L2202-L2203 (they still format verbatim)

As general rule, if we have these information, we consider the test as invalid: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/docs/write_tests.md#identify-issues

> some tokens haven't been printed; usually you will have this information inside the snapshot, under a section called "Unimplemeted tokens/nodes"; a test, in order to be valid, can't have that section;

ematipico (changes requested):

crates/rome_formatter/src/jsx/tag/self_closing_element.rs
ematipico: If possible, we should avoid creating new vectors. `join_elements` uses `concat_elements` under the hoods, which  is really optimized around the usage of new vectors.

crates/rome_js_formatter/tests/specs/jsx/self_closing.jsx.snap
NicholasLYang: I thought we had agreed that JsxName formatting was out of scope. But no worries, implemented it

ematipico (approved):

crates/rome_js_formatter/src/jsx/tag/self_closing_element.rs
```diff
@@ -1,9 +1,29 @@
...
-use rome_rowan::AstNode;
 
 impl ToFormatElement for JsxSelfClosingElement {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
-        Ok(formatter.format_verbatim(self.syntax()))
+        let mut attributes = Vec::new();
+        for attribute in self.attributes() {
+            attributes.push(attribute.format(formatter)?);
+            attributes.push(soft_line_break_or_space());
+        }
```
MichaReiser, line 15: You can use `join_elements(formatter.format_nodes(self.attributes())?, soft_line_break_or_space())` to get the same result.

NicholasLYang: I had the weirdest issue with that last time where it added line breaks when it shouldn't have. No clue why, but switched to it now.

PR #2283, opened 2022-03-22
refactor(rome_rowan): change the internal representation of green trivia

leops:
## Summary

This PR modifies how `GreenTrivia` are represented in memory, specifically allowing for single-piece trivias to be stored inline in a tagged pointer instead of allocating a trivia piece in the cache

### DIFF
--- a/crates/rome_rowan/src/api.rs
+++ b/crates/rome_rowan/src/api.rs
@@ -28,16 +28,23 @@ pub trait Language: Sized + Clone + Copy + fmt::Debug + Eq + Ord + std::hash::Ha
     type Kind: SyntaxKind;
 }
 
+// This enum must be `repr(u8)`, and has specially chosen discriminant values
+// with the least significant bit set to 1 to allow further optimizations in
+// the memory representation of trivia pieces
+#[repr(u8)]
+// Allow inefficient arithmetic operations to make what's going on more explicit,
+// these are all const-expressions that get executed at compilation anyway
+#[allow(clippy::identity_op)]
 #[derive(Clone, Copy, Eq, PartialEq, Hash, Debug)]
 pub enum TriviaPieceKind {
     /// A line break (`\n`, `\r`, `\r\n`, ...)
-    Newline,
+    Newline = (0 << 1) | 1,
     /// Any whitespace character
-    Whitespace,
+    Whitespace = (1 << 1) | 1,
     /// Comment that does not contain any line breaks
-    SingleLineComment,
+    SingleLineComment = (2 << 1) | 1,
     /// Comment that contains at least one line break
-    MultiLineComment,
+    MultiLineComment = (3 << 1) | 1,
 }
 
 impl TriviaPieceKind {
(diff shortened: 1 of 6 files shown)


leops:
!bench_parser

crates/rome_rowan/src/green/trivia.rs
```diff
@@ -93,17 +208,37 @@ impl GreenTrivia {
...
     /// Returns the pieces count
     pub fn len(&self) -> usize {
-        match &self.ptr {
-            None => 0,
-            Some(ptr) => ptr.len(),
+        // SAFETY: This reads the content of inner as bits, and use it to
+        // detect whether the content is empty, an inline trivia piece or a
+        // pointer
+        unsafe {
+            if self.inner.bits == 0 {
```
MichaReiser, line 215: Nit: How about adding methods like: `is_empty_trivia` / `is_single_piece_trivia()`, to avoid repeating the same test?

crates/rome_rowan/src/green/trivia.rs
```diff
@@ -68,15 +144,54 @@ impl GreenTrivia {
         I: IntoIterator<Item = TriviaPiece>,
         I::IntoIter: ExactSizeIterator,
     {
-        let data =
-            ThinArc::from_header_and_iter(GreenTriviaHead { _c: Count::new() }, pieces.into_iter());
+        let mut items = pieces.into_iter();
+        if items.len() == 1 {
```
MichaReiser, line 148: Nit: Should we add an assertion that items isn't empty or call into `empty` if items is empty?

xunilrj:
Wondering if we can take advantage of ```rustc_layout_scalar_valid_range_start``` in the future.
see https://zulip-archive.rust-lang.org/stream/219381-t-libs/topic/.23.5Brustc_layout_scalar_valid_range_start.5D.html#247288849

I doubt we can find a safe range on all platforms, though.

leops:
!bench_parser

MichaReiser:
> !bench_parser

In case you didn't know. You can easily run the benchmarks locally too, which can be quicker then running them with `!bench_parser`. 

1. Run `cargo bench_parser --save-baseline=main` on main once. 
2. Run `cargo bench_parser --save-baseline=pr` on your branch
3. Run `critcmp main pr` (install with `cargo install critcmp`)

Running the benchmarks can take a lot of time. That's why you can filter the benchmarks by using `cargo bench_parser --save-baseline=pr --filter=<name-filter>` I often pick a test to verify the improvements. For this case, I would pick a somewhat larger file that isn't minified. For example, `parser.ts` or `router.ts`

ematipico:
It seems that the action wasn't triggered at all

ematipico:
!bench_parser

crates/rome_rowan/src/api.rs
```diff
@@ -40,6 +43,29 @@ pub enum TriviaPieceKind {
     MultiLineComment,
 }
 
+impl From<u8> for TriviaPieceKind {
```
MichaReiser, line 46: Wdyt: Would it make sense to have an internal `TriviaPieceKind` type (`GreenTriviaPieceKind`?) that does the bit magic and is crate internal and then expose a transparent `TriviaPieceKind` that exposes a safe API (and, for example, hides that `from(u8)` can panic?

leops:
Archiving this since it doesn't result in any meaningful performance improvement

PR #2287, opened 2022-03-22
feat(rome_formatter): less allocation when concat elements

xunilrj:
When we concat elements, we create a list and flat items in it. Sometimes the first item is actually a list. Sometimes a big list, actually. In this case, we hijack this list instead of creating a new one and avoid moving all elements.

Another advantage is that we can use ```size_hint``` better.

On ```rsuite``` case, when formatting 600 files this hijacking happened more than 120K times and the biggest list had 150 items.
On our benchmark:

jquery: +38K times, +800 items;
threejs: +170K times; +3000 items;

Maybe we want a "lazy list" like rope? But a linked list of vecs.

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -250,7 +250,7 @@ where
 {
     let mut elements = elements.into_iter();
 
-    let mut size_hint = if let (_, Some(upper_bound)) = elements.size_hint() {
+    let size_hint = if let (_, Some(upper_bound)) = elements.size_hint() {
         upper_bound
     } else {
         0
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -248,20 +248,40 @@ pub fn concat_elements<I>(elements: I) -> FormatElement
 where
     I: IntoIterator<Item = FormatElement>,
 {
-    let elements = elements.into_iter();
+    let mut elements = elements.into_iter();
 
-    let mut concatenated: Vec<FormatElement> = if let (_, Some(upper_bound)) = elements.size_hint()
-    {
-        Vec::with_capacity(upper_bound)
+    let mut size_hint = if let (_, Some(upper_bound)) = elements.size_hint() {
+        upper_bound
     } else {
-        vec![]
+        0
     };
 
-    for element in elements {
-        match element {
-            FormatElement::List(list) => concatenated.extend(list.content),
-            FormatElement::Empty => (),
-            _ => concatenated.push(element),
+    // If the first non empty element is a vec, use it,
+    // otherwise create a new one with the current element
+    let mut concatenated = loop {
+        match elements.next() {
+            Some(FormatElement::List(list)) => {
+                let mut v = list.content;
+                v.reserve(size_hint);
+                break v;
+            }
+            Some(FormatElement::Empty) => continue,
+            Some(element) => {
+                let mut v = Vec::with_capacity(size_hint);
+                v.push(element);
+                break v;
+            }
+            None => return empty_element(),
+        }
+    };
+
+    // continue to the rest of the list
+    loop {
+        match elements.next() {
+            Some(FormatElement::List(list)) => concatenated.extend(list.content),
+            Some(FormatElement::Empty) => {}
+            Some(element) => concatenated.push(element),
+            None => break,
         }
     }
 
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -250,22 +250,19 @@ where
 {
     let mut elements = elements.into_iter();
 
-    let size_hint = if let (_, Some(upper_bound)) = elements.size_hint() {
-        upper_bound
-    } else {
-        0
-    };
+    let (lower_bound, upper_bound) = elements.size_hint();
+    let size_hint = upper_bound.unwrap_or(lower_bound);
 
     // If the first non empty element is a vec, use it,
     // otherwise create a new one with the current element
     let mut concatenated = loop {
         match elements.next() {
+            Some(FormatElement::Empty) => continue,
             Some(FormatElement::List(list)) => {
                 let mut v = list.content;
                 v.reserve(size_hint);
                 break v;
             }
-            Some(FormatElement::Empty) => continue,
             Some(element) => {
                 let mut v = Vec::with_capacity(size_hint);
                 v.push(element);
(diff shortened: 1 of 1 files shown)


xunilrj:
!bench_formatter

crates/rome_formatter/src/format_element.rs
```diff
@@ -248,20 +248,40 @@ pub fn concat_elements<I>(elements: I) -> FormatElement
...
+
+    // continue to the rest of the list
+    loop {
+        match elements.next() {
+            Some(FormatElement::List(list)) => concatenated.extend(list.content),
+            Some(FormatElement::Empty) => {}
+            Some(element) => concatenated.push(element),
+            None => break,
         }
     }
```
MichaReiser, line 286: Nit: This can stay a loop
```suggestion
    for element in elements {
        match element {
            FormatElement::List(list) => concatenated.extend(list.content),
            FormatElement::Empty => (),
            _ => concatenated.push(element),
        }
    }
```

crates/rome_formatter/src/format_element.rs
```diff
@@ -248,20 +248,40 @@ pub fn concat_elements<I>(elements: I) -> FormatElement
...
-            FormatElement::List(list) => concatenated.extend(list.content),
-            FormatElement::Empty => (),
-            _ => concatenated.push(element),
+    // If the first non empty element is a vec, use it,
+    // otherwise create a new one with the current element
+    let mut concatenated = loop {
+        match elements.next() {
+            Some(FormatElement::List(list)) => {
+                let mut v = list.content;
+                v.reserve(size_hint);
```
MichaReiser, line 265: Nit: In theory, it would now be `size_hint -1` but I don't think that matters :D

crates/rome_formatter/src/format_element.rs
```diff
@@ -248,20 +248,40 @@ pub fn concat_elements<I>(elements: I) -> FormatElement
...
     };
 
-    for element in elements {
-        match element {
-            FormatElement::List(list) => concatenated.extend(list.content),
-            FormatElement::Empty => (),
-            _ => concatenated.push(element),
+    // If the first non empty element is a vec, use it,
+    // otherwise create a new one with the current element
+    let mut concatenated = loop {
```
MichaReiser, line 261: Nit: We could avoid the loop here by using: `elements = elements.skip(|element| element.is_empty())`

You can then match on `elements.next()`

MichaReiser (approved):
Nice observation! This improves the perf significantly on my computer

```
group                                    concat                                 main
-----                                    ------                                 ----
formatter/checker.ts                     1.00    255.3±3.23ms    10.2 MB/sec  
formatter/compiler.js                    1.00    142.4±1.53ms     7.4 MB/sec    1.08    154.3±3.39ms     6.8 MB/sec
formatter/d3.min.js                      1.00    118.2±2.05ms     2.2 MB/sec    1.13   133.6±15.87ms  2009.4 KB/sec
formatter/dojo.js                        1.00      7.4±0.03ms     9.2 MB/sec    1.09      8.1±0.03ms     8.4 MB/sec
formatter/ios.d.ts                       1.00    182.1±1.40ms    10.2 MB/sec    1.12    204.5±2.39ms     9.1 MB/sec
formatter/jquery.min.js                  1.00     28.8±1.24ms     2.9 MB/sec    1.13     32.6±1.15ms     2.5 MB/sec
formatter/math.js                        1.00    236.6±3.94ms     2.7 MB/sec    1.06    250.8±5.17ms     2.6 MB/sec
formatter/parser.ts                      1.00      5.2±0.02ms     9.3 MB/sec    1.07      5.5±0.02ms     8.7 MB/sec
formatter/pixi.min.js                    1.00    130.1±1.73ms     3.4 MB/sec    1.07    139.5±2.85ms     3.1 MB/sec
formatter/react-dom.production.min.js    1.00     36.2±0.25ms     3.2 MB/sec    1.09     39.6±0.46ms     2.9 MB/sec
formatter/react.production.min.js        1.00   1795.1±8.81µs     3.4 MB/sec    1.11  1984.7±12.74µs     3.1 MB/sec
formatter/router.ts                      1.00      3.6±0.01ms    16.4 MB/sec    1.07      3.9±0.08ms    15.3 MB/sec
formatter/tex-chtml-full.js              1.00    300.9±2.60ms     3.0 MB/sec    1.09   327.0±17.51ms     2.8 MB/sec
formatter/three.min.js                   1.00    153.5±3.25ms     3.8 MB/sec    1.08    166.2±3.45ms     3.5 MB/sec
formatter/typescript.js                  1.00    935.4±5.82ms    10.2 MB/sec    1.08  1013.0±63.23ms     9.4 MB/sec
formatter/vue.global.prod.js             1.00     47.2±1.10ms     2.6 MB/sec    1.09     51.5±0.49ms     2.3 MB/sec
```

> Maybe we want a "lazy list" like rope? But a linked list of vecs.

What we do here is a memory optimisation to early throw away any `empty_element`s. I expected that simplifying `concat_elements` to simply creating a list (without testing for empty or anything like that) would improve perf on the cost of memory but perf regressed...

ematipico (approved):

crates/rome_formatter/src/format_element.rs
xunilrj: true; and easier to reason about.

crates/rome_formatter/src/format_element.rs
xunilrj: Yeah. The more the better!

crates/rome_formatter/src/format_element.rs
```diff
@@ -248,20 +248,40 @@ pub fn concat_elements<I>(elements: I) -> FormatElement
...
+    let mut elements = elements.into_iter();
 
-    let mut concatenated: Vec<FormatElement> = if let (_, Some(upper_bound)) = elements.size_hint()
-    {
-        Vec::with_capacity(upper_bound)
+    let size_hint = if let (_, Some(upper_bound)) = elements.size_hint() {
+        upper_bound
     } else {
-        vec![]
+        0
```
leops, line 256: This is somewhat unrelated but I wonder what's the impact of using `lower_bound` as a default here instead of 0

leops (approved):
In a sense the current implementation is already "rope-like", the most significant difference being the usage of a general tree structure and not a binary tree. In practice since the only operation used to build the tree is concatenation, and the resulting structure is only traversed once in the vast majority of use it would be interesting to explore flattening the tree into a list of events similarly to what's done in the parser

xunilrj:
> In a sense the current implementation is already "rope-like", the most significant difference being the usage of a general tree structure and not a binary tree. In practice since the only operation used to build the tree is concatenation, and the resulting structure is only traversed once in the vast majority of use it would be interesting to explore flattening the tree into a list of events similarly to what's done in the parser

hum... maybe we can use something like this? https://github.com/mamcx/tree-flat 
I was actually discussing this for the CST, with Micha.

crates/rome_formatter/src/format_element.rs
xunilrj: Let us see! I will push this and run the benchmarker.

xunilrj:
!bench_formatter

PR #2288, opened 2022-03-23
feat(rslint_parser): support for jsx spread child expression

xunilrj:
see https://github.com/rome/tools/issues/2153

This PR add support for 

```jsx
<div>
    {...a}
</div>
```


crates/rslint_parser/src/syntax/jsx/mod.rs
```diff
@@ -583,6 +585,31 @@ fn parse_jsx_expression_block(p: &mut Parser, kind: ExpressionBlock) -> ParsedSy
...
+                    p.error(err);
+                    m.change_to_unknown(p);
+                    m
+                }
+                _ => m,
+            }
+        });
+        expr.or_add_diagnostic(p, expected_expression);
+        p.expect(T!['}']);
+        return Present(m.complete(p, JSX_SPREAD_CHILD));
```
MichaReiser, line 610: I would recommend merging the code with the "regular" expression parsing to make sure things like `bump_with_context` are correctly handled

```suggestion
        let kind = match kind {
             ExpressionBlock::Attribute => JsSyntaxKind::JSX_EXPRESSION_ATTRIBUTE_VALUE,
             ExpressionBlock::Children => {
                  if p.eat(T![...]) {
                       JSX_SPREAD_CHILD
                  } else {
                      JSX_EXPRESSION_CHILD
                  }
             }
        };
```

crates/rslint_parser/test_data/inline/ok/jsx_children_spread.jsx
```diff
@@ -0,0 +1 @@
+<div>{...a}</div>
```
MichaReiser, line 1: Can you add a test that also has some `JSXText` after the spread?

crates/rome_js_parser/src/syntax/jsx/mod.rs
```diff
@@ -591,19 +593,36 @@ fn parse_jsx_expression_block(p: &mut Parser, kind: ExpressionBlock) -> ParsedSy
...
+            }
+            _ => None,
+        };
+
+        if let Some(err) = err {
             p.error(err);
             m.change_to_unknown(p);
             m
+        } else {
+            m
```
MichaReiser, line 624: Nit: You can move the `m` outside of the if block (and remove the `else` branch)

MichaReiser (approved):

PR #2290, opened 2022-03-23
feat(rome_formatter): suppress semicolon insertion after unknown nodes

leops:
## Summary

This PR changes all instances of `.format_or(token(";")` into a call to a new `format_with_semicolon` that only adds a semicolon that was missing in the input if the element that precedes it was not an unknown node.

To keep track of unknown nodes through the formatting process, `FormatElement::Verbatim` has been extended to cover not only ranges that were unformatted due an incomplete `ToFormatElement` implementation but also those that were printed verbatim due to an unknown node or a suppression comment. A new `VerbatimKind` property has been added to the `Verbatim` type to keep track of the origin of a verbatim range and maintain the existing behavior where only verbatim elements resulting from a call to `format_verbatim` are printed out in test snapshots.


crates/rome_formatter/tests/specs/prettier/js/ignore/ignore-2.js.snap
```diff
@@ -44,11 +44,11 @@ function HelloWorld() {
   );
 }
 
-a = <div {.../* prettier-ignore */b}/>;
-a = <div {...b/* prettier-ignore */}/>;
-a = <div {.../* prettier-ignore */{}}/>;
-a = <div {...{/* prettier-ignore */}}/>;
-a = <div {...{}/* prettier-ignore */}/>;
+a = <div {.../* prettier-ignore */b}/>
```
MichaReiser, line 47: Interesting. why do these parse as unknown CC: @xunilrj

crates/rome_formatter/src/utils/mod.rs
```diff
@@ -578,3 +578,49 @@ impl FormatPrecedence {
...
+/// preceeding element wasn't an unknown node
+pub(crate) fn format_with_semicolon(
+    formatter: &Formatter,
+    content: FormatElement,
+    semicolon: Option<SyntaxToken>,
+) -> FormatResult<FormatElement> {
+    let is_unknown = match last_element(&content) {
+        Some(FormatElement::Verbatim(elem)) => elem.is_unknown(),
+        _ => false,
+    };
```
MichaReiser, line 613: Hmm, what's preventing us from making the check on the element instead? 

Having something like `format_with_semicolon(formatted, element, With, semicolon)`. I would prefer inspecting the AST over having to inspect the Format IR.

ematipico (approved):

crates/rome_formatter/src/utils/mod.rs
ematipico: Inspecting the Format IR because it gives us less dependency from the AST, in general. Your suggestions gives us a closure, which in turns gives the responsibility to find the "last element" to the AST APIs itself... meaning that we also to rely to their performance and in finding the right element.

MichaReiser: Why does it need to be the last element? Normally the situation is that you have an expression that is then followed by a `;`. Thus, getting the expression is the same as calling a field accessor?

MichaReiser: The way I understand the current approach is that the semicolon should only be inserted if the node right in front of the semicolon isn't an unknown node. The challenge here is with properties where there can be multiple "last" nodes because the initializer is optional. But this shouldn't matter because we should always insert the `;` if the initializer is missing. 

I guess my main concern is that formatting behaviour that depends on inspecting the Format IR is harder to understand than formatting that is CST based. This complexity additionally increases every time we add a new IR, making it harder to approach the formatter.

crates/rome_formatter/src/js/module/export_named_from_clause.rs
```diff
@@ -36,18 +37,20 @@ impl ToFormatElement for JsExportNamedFromClause {
...
+        format_with_semicolon(
+            formatter,
+            format_elements![
+                type_token,
+                list,
+                space_token(),
+                from,
+                space_token(),
+                source,
+                space_token(),
```
MichaReiser, line 50: Is this space necessary? Doesn't it now add a space if `assertion` is empty?

crates/rome_formatter/tests/specs/prettier/js/ignore/ignore-2.js.snap
ematipico: It's more interesting when you see that it doesn't emit any errors: https://play.rome.tools/?lineWidth=80&indentStyle=tab&indentWidth=2&typescript=false&jsx=true#YSA9IDxkaXYgey4uLi8qIHByZXR0aWVyLWlnbm9yZSAqL2J9Lz4=

ematipico: I presume it's a bug in the implementation of  the algorithm of this PR. @leops  could you please check it before merging the PR?

ematipico: Nvm, it's like this because the formatter for JSX are verbatim. So it makes totally sense!

crates/rome_formatter/src/utils/mod.rs
leops: That's the approach I tried initially but it's not very ergonomic and hard to factor out into a function as in many case the node right before the semicolon could be one of multiple optional nodes (for instance in `JsPropertyClassMember` it could be either the name, the type annotation or the initializer clause)

crates/rome_formatter/src/js/module/export_named_from_clause.rs
leops: Good catch, that space isn't necessary

crates/rome_formatter/tests/specs/prettier/js/ignore/ignore-2.js.snap
leops: It's not exactly that, the new behavior only applies to unknown nodes and not to verbatim or suppressed ones. These nodes are indeed being parsed as unknown because the snapshots files doesn't use the JSX source type since unlike TypeScript there isn't really a way to recognize it will contain JSX based on the file name alone

MichaReiser (approved):

PR #2291, opened 2022-03-23
refactor(rome_formatter): use `prelude` terminology

ematipico:
## Summary

This PR uses the `prelude`  terminology to expose our traits. Now `formatter_prelude` only contains these traits.
The implementation has been moved inside `formatter.rs`. I could not find a better way to have these implementations. If you have a better idea, it's very welcome!

## Test Plan

Existing CI

### DIFF
--- a/crates/rome_formatter/src/formatter_traits.rs
+++ b/crates/rome_formatter/src/formatter_traits.rs
@@ -7,6 +7,9 @@ use crate::{
 use rome_js_syntax::{AstNode, SyntaxResult, SyntaxToken};
 
 /// Utility trait used to simplify the formatting of optional tokens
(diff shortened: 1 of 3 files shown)


leops (commented):
Overall the change makes sense, the only two things I would note are:
1. There isn't really a need to move any implementation around or rename the `formatter_traits` file, the prelude file can simply contain a list of `pub use` re-exports
2. Calling it `prelude` is more idiomatic than`formatter_prelude`, especially considering it might eventually be used in other auxiliary crates importing it using `use rome_formatter::formatter_prelude::*` seems more complicated than necessary

leops (approved):

PR #2293, opened 2022-03-23
test(rome_formatter): enable JSX parsing for js files

ematipico:
## Summary

This PR enables JSX parsing for `.js` files **only for the prettier testing suite**. 
Unfortunately they rely on the babel preset, which parses JSX inside `.js` files too.

This PR forces `SourceType::jsx()` when a file has the `js` extension.

## Test Plan

Current CI


leops (approved):
I think we might as well stop skipping the JSX tests altogether now that JSX parsing is getting stable (removing the `input.contains("jsx")` check at the start of `test_snapshot`, and adding `.jsx` / `.tsx` files to the glob expression that generates the tests cases)

ematipico:
> I think we might as well stop skipping the JSX tests altogether now that JSX parsing is getting stable (removing the `input.contains("jsx")` check at the start of `test_snapshot`, and adding `.jsx` / `.tsx` files to the glob expression that generates the tests cases)

It seems we are not ready yet for TSX. I am going to disable it. There's a syntax error of the formatted code which it requires a fix. I will create a different PR just for this.

PR #2294, opened 2022-03-23
refactor(rome_parser): Move lexer inside parser

MichaReiser:
## Summary

The lexer and parser heavily depend on each other which is why I believe it makes sense to move them into the same crate.

This also seems to improve performance by a bit (more inlining?) 

The downside of merging the crates is slightly longer compile times for parser / lexer changes because it must re-compile the lexer and parser at once. The build time for a single line change in the parser crate increases from:

* Debug: ~1.09s -> ~1.2s
* Release: ~1.61 -> 2s

## Test Plan

* `cargo test`
* `cargo codegen unicode`


MichaReiser:
!bench_parser

MichaReiser:
!bench_parser

ematipico (approved):

PR #2295, opened 2022-03-23
test(rome_formatter): enable TSX parsing for tsx files

ematipico:
## Summary

This PR does two things:

- enables `tsx` parsing inside the prettier test suite
- fixes a issue due to the trailing comma inside TS type parameters when parsed as `Source_type::tsx()`.

The issue was that before we were removing the trailing comma in case the group doesn't break. We are now forced to have it. Unfortunately we don't have a perfect heuristic because we can't access to the source type of the file, as a consequence I had to create a small workaround.

The drawback of the workaround is that the trailing comma is now kept even when we are in a TS context.

## Test Plan

Current prettier test suite should not fail

### DIFF
--- a/crates/rome_formatter/src/lib.rs
+++ b/crates/rome_formatter/src/lib.rs
@@ -596,9 +596,9 @@ mod test {
     // use this test check if your snippet prints as you wish, without using a snapshot
     fn quick_test() {
         let src = r#"
-3 + 3 === 7
+ const functionName1 = <T,>(arg) => false;
 "#;
-        let syntax = SourceType::ts();
+        let syntax = SourceType::tsx();
         let tree = parse(src, 0, syntax.clone());
         let result = format(FormatOptions::default(), &tree.syntax()).unwrap();
         check_reformat(CheckReformatParams {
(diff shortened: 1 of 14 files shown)


crates/rome_formatter/src/lib.rs
```diff
@@ -596,9 +596,9 @@ mod test {
     // use this test check if your snippet prints as you wish, without using a snapshot
     fn quick_test() {
         let src = r#"
-3 + 3 === 7
+ const functionName1 = <T,>(arg) => false;
```
MichaReiser, line 599: Nit: Revert?

crates/rome_js_parser/src/tests.rs
```diff
@@ -14,16 +14,10 @@ use std::path::{Path, PathBuf};
...
 fn parser_smoke_test() {
     let src = r#"
-function test() {}
-@test
-class Test {}
-@test.a?.c @test @test
-class Test2{}
-@test export class Test {}
-@test export default class Test {}
+const functionName1 = <T,>(arg) => false;
```
MichaReiser, line 17: Undo?

MichaReiser (approved):

crates/rome_formatter/src/lib.rs
ematipico: The test is kept ignored so contributors can change it without repercussions :)

PR #2296, opened 2022-03-24
feature(rome_cli): Rename binary to `rome`

MichaReiser:
## Summary

Rename the `rome_cli` binary to rome because `cli` isn't very telling.

I decided to go with `rome` because it's the main entry point
to rome for users.

## Test

```bash
cargo install --path ./crates/rome_cli
rome --help
Rome v0.0.0
Available commands:
- format
- help
```


ematipico (approved):

leops (commented):
The `release.yml` workflow needs to be changed too since it uploads the resulting file as a build artifact

MichaReiser:
> The `release.yml` workflow needs to be changed too since it uploads the resulting file as a build artifact

Is there a way for me to test the change?

MichaReiser:
Tested the release worfklow and it's passing [run](https://github.com/rome/tools/actions/runs/2033414137)

image

.github/workflows/release.yml
```diff
@@ -115,19 +115,19 @@ jobs:
...
-          cp target/${{ matrix.target }}/release/cli ./dist/rome_cli-${{ matrix.code-target }}
+          cp target/${{ matrix.target }}/release/rome ./dist/rome-${{ matrix.code-target }}
 
       # Upload the CLI binary as a build artifact
       - name: Upload CLI artifact
         uses: actions/upload-artifact@v2
         with:
           name: cli
-          path: ./dist/rome_cli-*
+          path: ./dist/rome-*
```
ematipico, line 130: Do we need to change this path?

MichaReiser: Yes, because I changed it on line 118/123. We could keep the name `rome_cil` in the distribution zip. But some people might then use `rome_cli format` and others `rome`.

PR #2299, opened 2022-03-24
feat(rome_cli): add a user-friendly panic hook to the CLI

leops:
## Summary

This PR replaces the default Rust panic hook with a custom printing logic for the CLI, directing users towards the GitHub issue tracker in case the CLI panics. The new format for panic errors is:

```text
> rome format --indent-size -1
The Rome CLI encountered an unexpected error

This is a bug in Rome, not an error in your code, and we would appreciate it if you could report it to https://github.com/rome/tools/ along with the following information to help us fixing the issue:

Source Location: crates\rome_cli\src\commands\format.rs:35:10
Thread Name: main
Message: failed to parse indent-size argument: Utf8ArgumentParsingFailed { value: "-1", cause: "invalid digit found in string" }
```

The new panic hook is implemented in the cli library, but is only installed in the cli binary and not the `run_cli` function to avoid overriding it in cargo tests

Currently this error message is going to show up for a number of errors that aren't actually bugs but the CLI using panics as an easy way to bail out of some failures (like parsing command line arguments), I'm working on a second PR besides this one to generalize the use of the `Termination` type for all errors and only panic in case something is seriously wrong

### DIFF
--- a/crates/rome_cli/src/panic.rs
+++ b/crates/rome_cli/src/panic.rs
@@ -16,10 +16,10 @@ fn panic_handler(info: &PanicInfo) {
     // panic at the same time
     let mut error = String::new();
 
-    writeln!(error, "The Rome CLI encountered an unexpected error").unwrap();
+    writeln!(error, "Rome encountered an unexpected error").unwrap();
     writeln!(error).unwrap();
 
-    writeln!(error, "This is a bug in Rome, not an error in your code, and we would appreciate it if you could report it to https://github.com/rome/tools/ along with the following information to help us fixing the issue:").unwrap();
+    writeln!(error, "This is a bug in Rome, not an error in your code, and we would appreciate it if you could report it to https://github.com/rome/tools/issues/ along with the following information to help us fixing the issue:").unwrap();
     writeln!(error).unwrap();
 
     if let Some(location) = info.location() {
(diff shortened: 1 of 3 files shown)


MichaReiser:
Love it! 

How about changing `The Rome CLI encountered an unexpected error` to `Rome encountered...`.

MichaReiser (approved):

crates/rome_cli/src/panic.rs
```diff
@@ -0,0 +1,41 @@
...
+fn panic_handler(info: &PanicInfo) {
+    // Buffer the error message to a string before printing it to stderr at once
+    // to prevent it from getting mixed with other errors if multiple threads
+    // panic at the same time
+    let mut error = String::new();
+
+    writeln!(error, "The Rome CLI encountered an unexpected error").unwrap();
+    writeln!(error).unwrap();
+
+    writeln!(error, "This is a bug in Rome, not an error in your code, and we would appreciate it if you could report it to https://github.com/rome/tools/ along with the following information to help us fixing the issue:").unwrap();
```
ematipico, line 22: We can redirect the user to the issues page: `https://github.com/rome/tools/issues/`

ematipico (approved):

PR #2300, opened 2022-03-24
fix(rome_formatter): check suppression before adding separator

ematipico:
## Summary

This PR fixes a subtle bug we have had for quite long time (since we introduced formatter traits).

With the traits, we introduced specialized checks around tokens and nodes, but these checks are skipped when calling directly `.to_format_element`.

This PR does two things:

- updates the codegen of the formatter to output the usage of `node.format(formatter)` on union nodes;
- fixes a bug where suppression comments were ignored inside the `TsTypeMemberList`;

## Test Plan

Added a new test to cover the fix; updated the prettier snapshots with the correct outcome.


xtask/codegen/src/formatter.rs
```diff
@@ -326,12 +326,12 @@ pub fn generate_formatter() {
                     .into_iter()
                     .map(|variant| {
                         let variant = Ident::new(&variant, Span::call_site());
-                        quote! { Self::#variant(node) => node.to_format_element(formatter), }
+                        quote! { Self::#variant(node) => node.format(formatter), }
```
leops, line 329: For union nodes using `to_format_element` was actually intentional, the assumption if that when this code is reached when the formatter of the parent node has called `union.format(formatter)` and thus the check for suppression comment has already been performed on this node so its not useful to run it a second time (this is also the case for node types that alias to a union by doing `JsAnyFunction::from(self).to_format_element(formatter)`)

crates/rome_formatter/tests/specs/js/module/suppression.js.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_formatter/tests/spec_test.rs
-assertion_line: 168
+assertion_line: 197
```
leops, line 3: Nit: This change could be discarded

xtask/codegen/src/formatter.rs
leops: In both case the call into the suppression check logic happens at the parent level (whatever calls into `function_expression.format(formatter)` for instance) and thus doesn't need to happen again since `has_formatter_suppressions` would return the same result (since that function looks at the comments attached to the first token, recursing into child nodes). This is true for any node type that only has a single (node) field, so `JsIdentifierExpression` could also call `to_format_element` directly for instance

leops (approved):

PR #2301, opened 2022-03-24
feat(rome_cli): refactor the Termination type into a proper error type

leops:
## Summary

This PR turns the `Termination` struct in the CLI into an enum using `thiserror` to actually define all the possible errors that can happen in a normal run of the CLI. All the errors resulting from incorrect inputs or configuration from the user are now handled properly instead of panicking, and are checked as part of the integration tests as well.

## Test Plan

New test cases have been added to exercise all the paths that can lead to the different error variants being emitted by the CLI

### DIFF
--- a/crates/rome_cli/src/termination.rs
+++ b/crates/rome_cli/src/termination.rs
@@ -1,0 +1,61 @@
+use std::{
+    env::current_exe,
+    ffi::OsString,
+    fmt::{self, Debug, Formatter},
+};
+use thiserror::Error;
+
+/// Error message returned by the CLI when it aborts with an error
+#[derive(Error)]
+pub enum Termination {
+    /// Returned by the CLI when it is called with a subcommand it doesn't know
+    #[error("unknown command '{command}'")]
+    UnknownCommand { command: String },
+
+    /// Return by the help command when it is called with a subcommand it doesn't know
+    #[error("cannot print help for unknown command '{command}'")]
+    UnknownCommandHelp { command: String },
+
+    /// Returned when the value of a command line argument could not be parsed
+    #[error("failed to parse argument '{argument}': {source}")]
+    ParseError {
+        argument: &'static str,
+        #[source]
+        source: pico_args::Error,
+    },
+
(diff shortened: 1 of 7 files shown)


crates/rome_cli/src/termination.rs
```diff
@@ -0,0 +1,61 @@
...
+    },
+
+    /// Returned when the CLI is passed a command line argument it doesn't know
+    #[error(
+        "unrecognized option {argument:?}. Type '{} format --help' for more information.",
+        command_name()
+    )]
+    UnexpectedArgument { argument: OsString },
+
+    /// Returned when the CLI when a required argument is not present in the command line
```
ematipico, line 34:
```suggestion
    /// Returned when a required argument is not present in the command line
```

crates/rome_cli/src/termination.rs
```diff
@@ -0,0 +1,61 @@
...
+    env::current_exe,
+    ffi::OsString,
+    fmt::{self, Debug, Formatter},
+};
+use thiserror::Error;
+
+/// Error message returned by the CLI when it aborts with an error
+#[derive(Error)]
+pub enum Termination {
+    /// Returned by the CLI when it is called with a subcommand it doesn't know
```
ematipico, line 11: We already mentioned that this error is returned by the CLI when we documented `Termination`
```suggestion
    /// Returned when it is called with a subcommand it doesn't know
```

crates/rome_cli/src/termination.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Returned when the value of a command line argument could not be parsed
+    #[error("failed to parse argument '{argument}': {source}")]
+    ParseError {
+        argument: &'static str,
+        #[source]
+        source: pico_args::Error,
+    },
+
+    /// Returned when the CLI is passed a command line argument it doesn't know
```
ematipico, line 27:
```suggestion
    /// Returned when the CLI  doesn't recognize a command line argument
```

ematipico (approved):

PR #2302, opened 2022-03-24
feature(rome_js_parser): JSX Error Recovery

MichaReiser:
## Summary

This PR adds diagnostics for invalid JSX syntax and improves the error recovery.

This PR further includes:

* Splitting fragment into `JsxFragment`, `JsxOpeningFragment` and `JsxClosingFragment`. This simplifies parsing and aligns the structure with `JsxElement`
* Change the optionality of `JsxExpressionChild`'s expression to optional ([spec](https://facebook.github.io/jsx/#prod-JSXChild)).
* Move JSX to primary expression according to the JSX spec.
* Only parse JSX in JSX (because of ambiguity)
* Disable parsing of Element type arguments in JSX to get improved error recovery in the case of `<Test <Nested>content</Nested></Test>`
* Don't parse `</` as `<`  followed by a Regular expression (only if there's no trivia in between) in JSX mode to have improved error recovery in case the `}` of an attribute or child is missing. 

## Tests

Adds a set of new parser test cases

Regressions: 

Babel incorrectly parses `<T>() => {}` as an arrow function. This isn't compatible with TypeScript's implementation that requires an explicit `,`: `<T, >() => {}`.


crates/rome_js_parser/src/syntax/expr.rs
```diff
@@ -1745,20 +1746,14 @@ pub(super) fn parse_unary_expr(p: &mut Parser, context: ExpressionContext) -> Pa
         return Present(expr);
     }
 
-    // if we are at "<"; or we have JSX or Typescript type assertions
-    if p.at(T![<]) {
-        let jsx = JsSyntaxFeature::Jsx.parse_exclusive_syntax(
+    // if we are at "<"; or we have Typescript type assertions
```
xunilrj, line 1749: This first phrase seems misplaced now. And there is type on the next line.

xunilrj (approved):

MichaReiser:
!bench_parser

crates/rome_js_parser/src/token_source.rs
```diff
@@ -175,6 +175,17 @@ impl<'l> TokenSource<'l> {
...
     }
 
+    /// Returns `true` if the next token has any preceding trivia (either trailing trivia of the current
+    /// token or leading trivia of the next token)
+    pub fn has_next_preceding_trivia(&mut self) -> bool {
+        let next_token_trivia = self
+            .lexer
+            .lookahead()
+            .next()
+            .and_then(|lookahead| TriviaPieceKind::try_from(lookahead.kind()).ok());
```
ematipico, line 185: I am failing to understand where we do check around "trailing trivia of the current token". Where does it happen?

crates/rome_js_parser/test_data/inline/err/jsx_closing_element_mismatch.rast
```diff
@@ -0,0 +1,213 @@
...
+              0: JSX_IDENT@49..55 "nested" [] []
+            3: R_ANGLE@55..56 ">" [] []
+      1: SEMICOLON@56..57 ";" [] []
+  3: EOF@57..58 "" [Newline("\n")] []
+--
+error[SyntaxError]: expected an identifier but instead found '>'
+  ┌─ jsx_closing_element_mismatch.jsx:1:9
+  │
+1 │ <test></>;
+  │         ^ Expected an identifier here
```
ematipico, line 181: The error of this case seems to be inconsistent when compared to the others, especially when we compare it to the other test case: `<>test</test>`. Is there a reason why we can't have a similar error message? (I would expect a message like "expected corresponding JSX closing tag for <test>")

crates/rome_js_parser/src/token_source.rs
MichaReiser: * It gets the next token from the lexer which can either be a trivia or a token. 
* `TriviaPieceKind::try_from` returns `Err` if the piece isn't a trivia
* That's why `is_some` is only true if the next lexed token is a trivia.

crates/rome_js_parser/test_data/inline/err/jsx_closing_element_mismatch.rast
MichaReiser: It's not that we can't, it's just that it's a bit more involved :) I can give it a try and have a special case in case the closing element turns out to be a closing fragment.

crates/rome_js_parser/src/token_source.rs
ematipico: Lol, dumb me. My head was still thinking to the AST façade 😂

Makes sense now, thank you!

ematipico (approved):

PR #2303, opened 2022-03-25
doc(website): add suppression explanation

ematipico:
## Summary

This PR adds documentation about suppression comments for the formatter.

For the time being, I preferred to have a section under the formatter because it's the only tool we have. Once we will release multiple tools that will benefit from different suppression (lint, bundler, etc.), we can revisit it and have it inside a dedicated section.

## Test Plan

Website

### DIFF
--- a/website/src/_includes/docs/formatter.md
+++ b/website/src/_includes/docs/formatter.md
@@ -58,9 +58,9 @@ OPTIONS:
 
 ### Suppression
 
-There are times when a developer wants to keep a specific formatting, and the formatter should not get in the way.
+There are times when a developer wants to keep a specific formatting.
 
-You can achieve this by adding a suppression comment right before the expression/statement.
+You can achieve this by adding a suppression comment right before the syntax node (expressions, statements, etc.).
 
 A suppression comment will have to look like this:
 
@@ -87,6 +87,13 @@ const   expr   =
     0,           0,           -1,            0,
 ];
 
+
+const   expr   =   [
+    (2*n)/(r-l), 0,            (r+l)/(r-l),  0,
+    0,           (2*n)/(t-b),  (t+b)/(t-b),  0,
+    0,           0,           -(f+n)/(f-n), -(2*f*n)/(f-n),
+    0,           0,           -1,            0,
+];
 ```
 
 After 
@@ -99,6 +106,25 @@ const expr =
     0,           (2*n)/(t-b),  (t+b)/(t-b),  0,
     0,           0,           -(f+n)/(f-n), -(2*f*n)/(f-n),
     0,           0,           -1,            0,
+];
+
+const expr = [
+    (2 * n) / (r - l),
+    0,
+    (r + l) / (r - l),
+    0,
+    0,
+    (2 * n) / (t - b),
+    (t + b) / (t - b),
+    0,
+    0,
+    0,
+    -(f + n) / (f - n),
+    -(2 * f * n) / (f - n),
+    0,
+    0,
+    -1,
+    0,
 ];
 ```
 
--- a/website/src/_includes/docs/formatter.md
+++ b/website/src/_includes/docs/formatter.md
@@ -55,3 +55,51 @@ OPTIONS:
     --indent-size <number>      If the indentation style is set to spaces, determine how many spaces should be used for indentation (default: 2)
 
 ```
+
+### Suppression
+
+There are times when a developer wants to keep a specific formatting, and the formatter should not get in the way.
+
+You can achieve this by adding a suppression comment right before the expression/statement.
+
+A suppression comment will have to look like this:
+
+```js
+// rome-ignore format: it's better this way
+```
+
+Where
+- `rome-ignore` is the start of each suppression comment in Rome;
+- `format:` is the category of the suppression;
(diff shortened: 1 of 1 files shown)


leops (approved):

website/src/_includes/docs/formatter.md
```diff
@@ -54,4 +54,52 @@ OPTIONS:
...
+```js
+const expr =
+// rome-ignore format: the array should not be formatted
+[
+    (2*n)/(r-l), 0,            (r+l)/(r-l),  0,
+    0,           (2*n)/(t-b),  (t+b)/(t-b),  0,
+    0,           0,           -(f+n)/(f-n), -(2*f*n)/(f-n),
+    0,           0,           -1,            0,
+];
+```
```
MichaReiser, line 103: I don't see the difference...

Does our markdown rendering support rendering diffs? Might be easier to spot the difference.

website/src/_includes/docs/formatter.md
```diff
@@ -54,4 +54,52 @@ OPTIONS:
     --indent-style <tabs|space> Determine whether the formatter should use tabs or spaces for indentation (default: tabs)
     --indent-size <number>      If the indentation style is set to spaces, determine how many spaces should be used for indentation (default: 2)
 
-```
\ No newline at end of file
+```
+
+### Suppression
+
+There are times when a developer wants to keep a specific formatting, and the formatter should not get in the way.
```
MichaReiser, line 61: ... and the formatter should not get in the way sounds negative. 

`rome-ignore format` disables the formatted for the next syntax (node?)

website/src/_includes/docs/formatter.md
```diff
@@ -54,4 +54,52 @@ OPTIONS:
...
+
+You can achieve this by adding a suppression comment right before the expression/statement.
```
MichaReiser, line 63: Suppressions aren't limitted to expressions or statements. You can add it anywhere in front of a node. However, it doesn't work in the middle of nodes (which may not be very intuitive).

MichaReiser:
How about taking some of the copy from the old website?

https://github.com/rome/tools/blob/main/website/src/_includes/docs/_old_linting.md#suppressions

website/src/_includes/docs/formatter.md
ematipico: The only difference around the spacing of `const`. I can change the example to show better where we format and where we don't!

website/src/_includes/docs/formatter.md
ematipico: I'd prefer a wording where we don't use technical terms. It could be confusing for people that don't know what a syntax node is.

website/src/_includes/docs/formatter.md
ematipico: Yeah, I wasn't sure about this bit.. initially I wrote "the line you don't want to format", but this isn't true either. I can't find a good way to express how the suppression works, unless we list ALL cases. What do you think?

website/src/_includes/docs/formatter.md
MichaReiser: Prettier uses a more obvious example. Does [that](https://prettier.io/docs/en/ignore.html#javascript) work with our formatter too?

ematipico:
> How about taking some of the copy from the old website?
> 
> [https://github.com/rome/tools/blob/main/website/src/_includes/docs/_old_linting.md#suppressions](https://github.com/rome/tools/blob/main/website/src/_includes/docs/_old_linting.md?rgh-link-date=2022-03-25T10%3A14%3A49Z#suppressions)

Is there a particular section you would like to copy? The old copy is more focused around diagnostics and linting. We could resurrect that copy once we will have diagnostics in place?

website/src/_includes/docs/formatter.md
```diff
@@ -54,4 +54,78 @@ OPTIONS:
...
+```
+
+Where
+- `rome-ignore` is the start of each suppression comment in Rome;
+- `format:` is the category of the suppression;
+- `it's better this way` is the reason why there's a suppression in the first place; 
+
+Here's an example of how a code will look like before and after the formatter does its job:
+
+Before
```
MichaReiser, line 78: Shouldn't the before be "before the suppression" is added and the after is with the suppression comment?

website/src/_includes/docs/formatter.md
```diff
@@ -54,4 +54,78 @@ OPTIONS:
...
\ No newline at end of file
+```
+
+### Suppression
+
+There are times when a developer wants to keep a specific formatting.
+
+You can achieve this by adding a suppression comment right before the syntax node (expressions, statements, etc.).
+
+A suppression comment will have to look like this:
```
MichaReiser, line 65: Nit:

Suppression comments have the following format:

```
// rome-ignore format: <explanation>
```

* `rome-ignore` is the start of a suppression comment
* `format`: suppresses the formatting
* `<explanation>`: An explanation why formatting is disabled.

website/src/_includes/docs/formatter.md
ematipico: Actually, now it doesn't make sense because the snippet contains the snippet with and without suppression. Maybe we should say "before running the formatter"

PR #2304, opened 2022-03-25
feature(rome_cli): distribute the Rome CLI to npm

leops:
## Summary

This PR enables the distribution of the CLI through npm (with the temporary package name `romecli`). In order to distribute the correct rebuilt binaries for the current platform, the `romecli` package has optional dependencies on a list of platform specific-packages, one for each supported OS + processor architecture target. These packages are generated with a script (`npm/romecli/scripts/generate-packages.mjs`) as part of the release CI workflow, that automatically creates a `package.json` manifest and copies the correct binary artifact in place for each package. Finally the root CLI package declares a `rome` binary pointing to a small node.js script that resolves the location of the correct binary package for the current platforms and spawns the actual CLI process, forwarding all command line arguments and stdio streams towards it.

As part of this I also bumped the version of the VSCode extension to `0.2.0`, this will make the next pre-releases use the `0.3.x` version, and the beta release will then use the `0.4.x` minor version.

## Test Plan

Run the GitHub Actions workflow until it works, I haven't set an npm token in the environment secrets yet so the publish step should fail and not actually push anything to the registry


.github/workflows/release.yml
```diff
@@ -211,6 +216,13 @@ jobs:
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
 
+      - name: Publish npm packages
+        # For now always publish the CLI to the "next" tag
+        # TODO: Actually call "npm publish" when this gets merged
+        run: for package in npm/romecli*; do echo "npm publish $package --tag next"; done
```
ematipico, line 222: Does this script work on windows?

npm/romecli/bin/rome
```diff
@@ -0,0 +1,18 @@
...
+
+const binName = isWindows ? "rome.exe" : "rome";
+const binPath = require.resolve(`romecli-${platform}-${arch}/${binName}`);
+
+const result = require("child_process").spawnSync(
+	binPath,
+	process.argv.slice(2),
+	{
+		shell: false,
+		stdio: "inherit",
```
MichaReiser, line 14: [`detached`](https://nodejs.org/api/child_process.html#optionsdetached) may be interesting. I can help you test the behaviour on linux.

npm/romecli/scripts/generate-packages.mjs
```diff
@@ -0,0 +1,75 @@
...
+				manifest.version,
+			])
+		)
+	);
+
+	const content = JSON.stringify(manifest);
+	await fs.writeFile(MANIFEST_PATH, content);
+}
+
+async function generatePackage(platform, arch) {
```
MichaReiser, line 30: Nit: It may be simpler to just call into the sync actions (and maybe even faster, at least that used to be the case for Metro on old node versions).

.github/workflows/release.yml
leops: I don't think it does, but this build step is only supposed to run on Linux

npm/romecli/bin/rome
leops: It would be interesting to test, because I'm not exactly sure how that would interact with inherited stdio (would the terminal wait for the detached child process to exit too before returning control to the shell, or would it return immediately when the parent process exits and mangle the output of the CLI with the command prompt is what I would like to know)

npm/romecli/scripts/generate-packages.mjs
leops: I've tested it locally with Node 14 and the whole script runs in pretty much the same time (140 milliseconds on average on Windows with a somewhat slow drive) whether its using the async or sync versions, so yes using sync everywhere might be simpler

npm/romecli/scripts/generate-packages.mjs
```diff
@@ -0,0 +1,73 @@
+import { resolve } from "path";
+import { fileURLToPath } from "url";
+import * as fs from "fs";
```
ematipico, line 3: I think using the `node:` prefix for importing node specific packages is preferred, at least we know for sure that they don't belong to external libraries.  

```suggestion
import { resolve } from "node:path";
import { fileURLToPath } from "node:url";
import * as fs from "node:fs";
```

npm/romecli/bin/rome
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+const { platform, arch } = process;
+const isWindows = platform == "win32";
+
+const binName = isWindows ? "rome.exe" : "rome";
+const binPath = require.resolve(`romecli-${platform}-${arch}/${binName}`);
+
+const result = require("child_process").spawnSync(
```
ematipico, line 9:
```suggestion
const result = require("node:child_process").spawnSync(
```

npm/romecli/bin/rome
leops: After reading the node.js documentation it appears to be neither, even if the child process is detached if it is created with inherited stdio it will exit anyway when the parent process exits. On Windows at least this means the wrapper script exits immediately and the child process is killed before it can do anything

.github/workflows/release.yml
```diff
@@ -211,6 +216,13 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
```
yassere, line 224: Currently, the (probably expired) npm token is in the `npm-publish` environment. If we want to add this token to the same environment as the `VSCE_PAT` we could rename the `marketplace` environment to something broader. Maybe just `publish`?

Would there be any advantage to keeping separate environments?

.github/workflows/release.yml
```diff
@@ -211,6 +216,13 @@ jobs:
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
 
+      - name: Publish npm packages
+        # For now always publish the CLI to the "next" tag
```
yassere, line 220: I'm not sure yet if the main cli package will be scoped or not, but it's likely that we'll want all of the platform-specific optional dependencies to be published under the `@rometools` scope. Would there be any reason for the scoped packages to be published with a "next" tag?

npm/romecli/scripts/generate-packages.mjs
```diff
@@ -0,0 +1,73 @@
...
+		cpu: [arch],
+	});
+
+	const manifestPath = resolve(packageRoot, "package.json");
+	console.log("Create manifest " + manifestPath);
+	fs.writeFileSync(manifestPath, manifest);
+
+	// Copy the CLI binary
+	const ext = platform === "win32" ? ".exe" : "";
+	const binarySource = resolve(REPO_ROOT, `rome-${platform}-${arch}${ext}`);
```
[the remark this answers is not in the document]
leops: For this script in particular a missing binary will raise an error when the file gets copied with a somewhat descriptive error (the OS missing file error), this will in turn cause the release workflow to fail and abort the publishing of the package so that's probably good enough.
For the wrapper script that spawns the CLI process though I should add a descriptive error like what's in the VSCode extension, but I was thinking we could even go one step beyond that and use a second script running in postinstall to cause the installation of the CLI to print a warning upfront if the current platform is unsupported (raising an error might be a bit much if people add the Rome CLI to their devDependencies since running an `npm install` in the workspace would fail on unsupported platforms even if they didn't intend to actually use it)

.github/workflows/release.yml
leops: Right now I didn't add any npm token to the publishing environment just to make sure I don't accidentally publish something while doing a test run of the workflow. Overall this workflow is intended to cover for the short-term publishing model where we push new versions of the toolchain, CLI and LSP all at once but ultimately I imagine these would be three separate components versioned independently, and so probably published independently as well.
In that regard since there would be 3 different publishing workflow we'd probably have a separate environment for each of these as well, but it doesn't really add anything other than making it just a tiny bit safer by making sure that even if there a security exploit in one of the workflows, at least only one of the tokens would get leaked instead of all three.

.github/workflows/release.yml
leops: There isn't really any reason for why everything is getting published under the `next` tag other than making the publish script command a one-liner, and nothing is using this tag since the main CLI package references the platform-specific binaries using an exact version number

ematipico:
@leops what's been missing from this PR to be ready for review?

leops:
> @leops what's been missing from this PR to be ready for review?

It's blocked on the "definitive" package names and scope to be used: whether we can use the `@rome` scope or have to use `@rometools`, and whether it'll be published under the existing `rome` package or if a new `rome-cli` frontend package needs to be created

.github/workflows/release.yml
yassere: Right, I saw that it was intentional that you haven't added an active npm token yet.

ematipico (approved):

leops:
The decision that was taken for the naming / versioning was:
- The entry CLI package is published as `rome` with version `0.3.0-next` and the `next` tag (and I think we'll bump it to `0.4.0-next` when the VSCode extension gets a stable release to keep them in sync for now ?)
- The platform specific binaries are published under `@rometools/cli-<platform>-<arch>` with the same version number and tag as the main package

ematipico:
> The decision that was taken for the naming / versioning was:
> 
>     * The entry CLI package is published as `rome` with version `0.3.0-next` and the `next` tag (and I think we'll bump it to `0.4.0-next` when the VSCode extension gets a stable release to keep them in sync for now ?)
> 
>     * The platform specific binaries are published under `@rometools/cli-<platform>-<arch>` with the same version number and tag as the main package

Yeah, I think we can keep them in sync for now, not a big deal

PR #2306, opened 2022-03-25
chore: bundle extension

ematipico:
## Summary

This PR adds:
- bundling the extension using `esbuild`; the code was inspired by [`rust-analyzer`](https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/package.json#L30);
- a `watch` command for development;
- a `format` command that uses our CLI to format the codebase;

This enables us the usage of modules inside the extension, which is something we need for the near feature (the code base will grow really soon)

## Test Plan

Manually checked that the `package` command works. Checked that the installed extension works.


editors/vscode/src/main.ts
```diff
@@ -1,102 +1,90 @@
...
-}
-
-export function deactivate(): Thenable<void> | undefined {
-	if (!client) {
-		return undefined;
-	}
-	return client.stop();
-}
\ No newline at end of file
+import { ExtensionContext, workspace, Uri, window } from 'vscode';
```
MichaReiser, line 1: Omg, so much to prettier compatible xD

leops (approved):
I was thinking we'd ultimately move to using our own bundler on the extension, but this works in the meantime, and it should improve startup performance

ematipico:
> I was thinking we'd ultimately move to using our own bundler on the extension, but this works in the meantime, and it should improve startup performance

Yes :) another chance to dogfood rome !

PR #2307, opened 2022-03-25
chore: Setup Rome Formatter in CI

MichaReiser:
## Summary

This PR uses the new [`rome/setup-rome`](https://github.com/rome/setup-rome) GitHub action to install rome and check the formatting of files in the `website` and `editors` directories.

The GitHub action asks users to explicitly specify a Rome version. This is hopefully temporary and the main motivation behind it is to make users aware that Rome formatted can introduce breaking changes even in minor or patch releases until Rome reaches version 1.0.0. Users can opt in to use the latest stable (`version: latest`) rome version or even the latest preview version (`version: preview`) but this may have the effect that a new Rome release breaks their CI.

The new GitHub Action has the following limitations:

* Only `version: preview` is working at the moment because there's no stable rusty-rome release yet
* Only supports single prerelease version: It assumes that the preview release with the newest publish date is the latest preview release. Meaning, the action picks the release with the most recent publish date if we publish pre-releases for 0.4 and 1.0. 

See [setup-rome](https://github.com/rome/setup-rome) for the code of the GitHub action (mainly `index.js`).

## Test Plan

Verified that Rome formatted runs as part of the PR workflow ([logs]([run](https://github.com/rome/tools/runs/5717088646?check_suite_focus=true)


ematipico (approved):

.github/workflows/pull_request.yml
```diff
@@ -35,10 +50,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: false
-      - name: Support longpaths
-        run: git config core.longpaths true
```
yassere, line 39: I don't know anything about why this was necessary before, but why isn't it necessary now? Was it just because of submodules?

MichaReiser: It is necessary for the TypeScript submodule but we only use the submodules for the coverage runs. So, not needed here.

PR #2309, opened 2022-03-28
feat(vscode): add `Show syntax tree` command for supported documents

ematipico:
## Summary

This PR adds a command to our VS Code extension called `Rome: Show Syntax Tree`. 

Screenshot 2022-03-28 at 13 58 01

Which will show the CST of the current document:

Screenshot 2022-03-28 at 13 58 16

This command was taken as inspiration from `rust-analyzer` which does the same for rust files. Although, this PR doesn't work on ranges for the time being. 

One cool thing that this command does, is to actually list to changes to the current document, which meas that every time a document is saved/changed, the view updates itself with the new CST!

## Test Plan

Manually checked that works. I plan to add integration tests (if possible) in a subsequent PR.


editors/vscode/src/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { commands, TextDocument, TextEditor } from "vscode";
+
+const SUPPORTED_LANGUAGES = new Set(["javascript", "typescript"]);
```
xunilrj, line 3: is vscode smart enought so this will work with "jsx", "tsx", "d.ts"?

crates/rome_lsp/src/requests/syntax_tree.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub const SYNTAX_TREE_REQUEST: &str = "rome/syntaxTree";
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct SyntaxTreePayload {
+    pub text_document: TextDocumentIdentifier,
+}
+
+pub fn syntax_tree(document: Document) -> Result<Option<String>> {
```
MichaReiser, line 16: Nit: Should we show the AST Facade instead? I find it more approachable than the flat node structures (which I don't like to call CST :D)

ematipico: We could, but at least the green node structure is able to give us the with of each node/token, which is something that 
we don't have with the red node structure. 🤔 

I don't mind showing both or just the red node structure.

MichaReiser: Each node (and token?) should have a `text_range()` method. The width of a node/token is `token_range().len()`.

ematipico: > Or do you mean because it's missing in the string representation?

Exactly. At the moment only tokens show their width, but we don't show their position/span: column:width

It's something we can add later but I can easily change and show the AST.

ematipico:
Switched to show the AST instead:
Screenshot 2022-03-28 at 15 57 49

crates/rome_lsp/src/requests/syntax_tree.rs
leops: I don't think we should change the textual representation of the AST as it will require updating every single parser test snapshot, what probably make the most sense is to serialize a meta-representation of the AST to JSON and use that in the extension, so that not only is the source range for each token / node sent to the VSCode client but that information can actually be parsed out and used for instance to highlight the node the cursor is currently at in the editor.

MichaReiser, line 16: What's the reason that this request returns a `Result<Option>>`. I don't see any `Err` or `Ok(None)` return paths. Is it possible to change it to `String`?

crates/rome_lsp/src/server.rs
```diff
@@ -71,6 +73,27 @@ impl LanguageServer for LSPServer {
         Ok(())
     }
 
+    async fn request_else(&self, method: &str, params: Option<Value>) -> LspResult<Option<Value>> {
```
MichaReiser, line 76: What's the meaning of `else?` Maybe add some documentation?

editors/vscode/src/main.ts
```diff
@@ -29,6 +34,13 @@ export async function activate(context: ExtensionContext) {
 
 	client = new LanguageClient("rome_lsp", "Rome", serverOptions, clientOptions);
 
+	const session = new Session(context, client);
+
+	// we are now in a rome project
+	setContextValue(IN_ROME_PROJECT, true);
```
MichaReiser, line 40: What's the benefit of setting `IN_ROME_PROJECT` to true? This seems to be the case for any project that loads our extension.

I also don't see any more reads of `inRomeProject`

crates/rome_lsp/src/requests/syntax_tree.rs
ematipico: The function is called inside a `spawn_blocking_task`, which requires the closure to return `Result`. I can remove the `Option` though

crates/rome_lsp/src/server.rs
ematipico: This the function given by the trait `LanguageServer`, which already gives us some documentation: 

```
This handler can be used to respond to all requests that are not handled by built in request handlers.
```

editors/vscode/src/main.ts
ematipico: This is a piece of code that I removed but that I am going to re introduce because it's important.

You will find `inRomeProject` inside the `package.json`, inside the following section:

```json
		"menus": {
			"commandPalette": [
				{
					"command": "rome.syntaxTree",
					"when": "inRomeProject"
				}
			]
		},
```

This is where we control the [command palette](https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette). The `when` field tells to VSCode **the circumstances of when showing a certain command**. I copied this strategy from `rust-analyzer`.

ematipico:
We now have a bit of syntax highlighting and collapsible sections! 

Screenshot 2022-03-30 at 09 33 26


I know that the highlighting is correct yet, but I would leave it out of scope and fix it later (I need to study a bit of the tmLanguage schema)

editors/vscode/src/main.ts
MichaReiser: That makes sense, but in what situation can our extension be loaded and `inRomeProject` be false?

ematipico: Ah, yes. For now we keep it always `true`. I will have to review this part and make more intelligent. Ideally the `isRomeProject` is `true` when:
- we discover our configuration file
- we attempt to parse a file that we actually support

But these are two information we don't have yet, so for debug purpose I wanted to keep it to `true` always.

MichaReiser: I personally would prefer removing it until it is needed (if ever).

MichaReiser (approved):

editors/vscode/src/main.ts
ematipico: If we remove it, the command will appear also on files that we don't support, for example `yml`, `py`, etc.

Should we show it regardless?

Issue #2310, opened 2022-03-28
☂️ <TITLE> 


xunilrj:
### Description

Not everything on this list is a bug. They are bullet points that we need to address. We can choose to keep it like it is today.

# Tasks

- [ ] Top level return
- [ ] export name from 'a'
- [ ] export { \n default as a } from 'a' 
- [ ] multiple "export default"
- [ ] const arguments/package and others invalid bindings
- [ ] export anonymous function on .d.ts
- [ ] Not all string are directives

## Top Level Return

 Some node.js scripts are using "top level return".

https://github.com/ant-design/ant-design/blob/master/scripts/check-version-md.js
https://stackoverflow.com/questions/28955047/why-does-a-module-level-return-statement-work-in-node-js

https://play.rome.tools/?lineWidth=80&indentStyle=tab&indentWidth=2&typescript=true&jsx=true#cmV0dXJuOw==

Examples:
https://github.com/ant-design/ant-design/blob/master/scripts/check-version-md.js#L60  
https://github.com/marmelab/react-admin/blob/master/cypress/start.js#L4  

## Export from

We don't support "export default from" and have a problem with "export { \n default ..." (mind the new line):

https://play.rome.tools/?lineWidth=80&indentStyle=tab&indentWidth=2&typescript=true&jsx=true#ZXhwb3J0IGRlZmF1bHQgZnJvbSAnYSc7CmV4cG9ydCB7IGRlZmF1bHQgYXMgQSB9IGZyb20gJ2EnOwpleHBvcnQgeyAKICAgIGRlZmF1bHQgYXMgQSB9IGZyb20gJ2EnOw==

```babel``` seems to like "export default from 'a';"; ```ts``` on the other hand does not.

https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAE2AMwIYFcA29lQgWzgHJUiBuIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.17.8&externalPlugins=%40babel%2Fplugin-proposal-export-default-from%407.16.7&assumptions=%7B%7D

[...]

PR #2313, opened 2022-03-28
fix(rome_js_parser: JSX `>=` parsing

MichaReiser:
## Summary

The parser failed to parse `<span>=</span>` because the lexer lexed `>=` as a single token rather than a `>` and a `=` token.

This PR changes the lexer to always lex `>=` (and `>>=` and `>>>=` as single tokens and only lex them as binary operators when called with `ReLex::BinaryOperator`.

Part of #2310

## Test Plan

Added a new test verifying that `>=` works correctly


ematipico (approved):

MichaReiser:
Just checked the regression. The expected errors are all errors related to unresolved symbols and not parsing errors.

PR #2314, opened 2022-03-28
fix(rome_js_parser): export default working with newlines

xunilrj:
This fixes the "export { \n default" bug at https://github.com/rome/tools/issues/2310

```js
export {
    default as a } from "b";
```

The problem is that we parse "export" initially without being an "export ... from...". Then we see a "from" token, rewind everything and parse everything again as an "export ... from...". They have completely different CST nodes.

Before "default" was faulting the "export" parsing, error recovery was kicking in and failing the "export" parse. In this case, the rewind would never occur. If this is not an "export", it cannot be an "export ... from..." is the logic used here.

Now we parse everything correctly and let the rewind kick in as expected.

### DIFF
--- a/crates/rome_js_parser/src/syntax/module.rs
+++ b/crates/rome_js_parser/src/syntax/module.rs
@@ -751,7 +751,8 @@ impl ParseSeparatedList for ExportNamedSpecifierList {
 }
 
 fn parse_any_export_named_specifier(p: &mut Parser) -> ParsedSyntax {
-    if !matches!(p.cur(), T![type] | T![as]) && !is_nth_at_reference_identifier(p, 0) {
+    if !matches!(p.cur(), T![type] | T![as] | T![default]) && !is_nth_at_reference_identifier(p, 0)
+    {
         return Absent;
     }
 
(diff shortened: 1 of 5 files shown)


crates/rome_js_parser/src/syntax/module.rs
```diff
@@ -780,7 +781,17 @@ fn parse_any_export_named_specifier(p: &mut Parser) -> ParsedSyntax {
             TextRange::new(p.cur_range().start(), p.cur_range().start()),
         ));
     } else {
-        parse_reference_identifier(p).or_add_diagnostic(p, expected_identifier);
+        // We need to parse "default" here and fail so the "export ... from..." rewind later works.
+        if p.cur() == T![default] {
```
MichaReiser, line 785: This fixes this specific problem but doesn't address `export { 'abcd' as B } from "./test";` which falls into the same category. 

I'm wondering if we could re-use some of the logic between the two variations Like a generic `parse_named_export` specifier that parses it as either and not only gives back the parsed syntax but also an indicator of what kind of named export this was. 

Another alternative could be to make the `rewind` path more intelligent. Instead of just testing if it is at a `from` token it could further test if it is at the end of a statement and if not, use the `STMT_RECOVERY` to find the next token before the next valid statement and if that's `from`, parse it out as an export... from. 

But obviously, this approach provides way better diagnostics and error recovery even in the case the user didn't write the `from` after all. So, I guess, it's probably worth keeping and all that needs to be added is to also cover string export names.

xunilrj: Yes. I prefer to tackle one problem at a time.

MichaReiser (approved):

PR #2315, opened 2022-03-28
ci: run checks only on relevant code

ematipico:
## Summary

This PR changes the workflow `pull_request.yml` to run it only on folders that really matter. For example, changes on the website should not trigger any format, lint, etc.

## Test Plan

This PR

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'crates/**'
+      - 'xtask/**'
+      - 'editors/**'
 
 env:
   RUST_LOG: info


.github/workflows/pull_request.yml
```diff
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'crates/**'
+      - 'xtask/**'
+      - 'editors/**'
```
MichaReiser, line 11: This will no longer work if I add the check to validate the formatting of the website. It would require to split the PR job into multiple sub-jobs. 

I would leave it as is. It doesn't add that match overhead. Unfortunately, GitHub actions don't seem to support paths on a steps level.

ematipico: I think it might make more sense to split the PR workflow. There's to need to run cargo tests, if there's a change only in JS files. And there's no need to run Rome format if there are only rust file updates.

My concern I want to address is the usage of resources.

MichaReiser: I understand the motivation but is the overhead significant enough to justify the complexity of split workflow files and the risk that we'll miss to run the jobs because they haven't been included in paths? For example, what if the `Cargo.lock` file was updated? We should run the tests in that caes.

PR #2319, opened 2022-03-29
docs: enhancements, add documentation for playground

ematipico:
## Summary

This PR
- moves some documentation of the parser inside its own `docs/` file, inside its crate;
- adds a new paragraph in the `CONTRIBUTING.md` where it points to the instructions of how to run the playground locally; 
- adds a small feature in the playground where the syntax highlighting changes based on the language;

## Test Plan

Check the rendered markdown

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,90 +161,15 @@ from the main repository (not a fork), use these comments to run specific workfl
 
  	When a rule's sole intention is to **mandate a single concept** - such as forcing the use of camel-casing - the rule should be named using the `use` prefix. For example, the rule to mandating the use of camel-cased variable names is named `useCamelCase`.
 
-### Write parsers and parsing rules
+### Parser
 
-To have a better understanding of our parsing infrastructure, please [read the in-depth section](/crates/rome_js_parser/docs/authoring_parse_rules.md)
-
-### Write tests for a parser
-
-If you want to create a new test for an existing parser, you will have to inline
-the code that you want to test in a comment that is created in a specific way.
-
-Let's say that you created a new parsing feature and you need new tests from scratch,
-just go to the source code where you parse this new feature if JavaScript, and add the following comment:
-
-```rust
-// test feature_name
-// let a = { new_feature : "" }
-// let b = { new_feature : "" }
-fn parse_new_feature(p: &mut Parser) -> ParsedSyntax {}
-```
-
-The first line, `// test feature_name` the important one. This will tell to the
-testing infrastructure to create a **positive test** (without parsing errors), called
-`feature_name.js` inside the `test_data/inline/ok` folder.
-
-The content of this file will be:
-
-```js
-let a = { new_feature : "" }
-let b = { new_feature : "" }
-```
-
-Basically, everything after the key comment will be the content of the new file.
-
-Now you need to run `cargo codegen test` and the task will actually generate this file for you.
-
-In case you want to create a **negative test** (*with* parsing errors), you will
-create a new comment like this:
-
-```diff
-// test feature_name
-// let a = { new_feature : "" }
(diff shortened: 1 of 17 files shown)


CONTRIBUTING.md
```diff
@@ -161,94 +161,23 @@ from the main repository (not a fork), use these comments to run specific workfl
...
 ### Versioning
 
 We follow the specs suggested by [the official documentation](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions):
 
 Odd minor versions are dedicated to pre-releases, e.g. `*.5.*` .
 Even minor versions are dedicated to official releases, e.g. `*.6.*`.
+
+### Playground
+
+- [run locally the playground](/crates/rome_playground/README.md)
```
MichaReiser, line 183: Nit: run the playground locally

crates/rome_formatter/docs/implement_the_formatter.md
```diff
@@ -86,7 +86,8 @@ This will automatically build and open a browser tab to the documentation.
    is that when you have a **typed node at hand**, prefer the `format*` traits instead of `.to_format_element`
 4. Use our [playground](https://play.rome.tools) to inspect the code that you want to format. You can inspect
 the AST given by a certain snippet. This will help you to understand which nodes need to be implemented/modified
-in order to correctly implement the formatter.
+in order to correctly implement the formatter. Alternatively, you can locally run the playground by following [its
```
MichaReiser, line 89:
```suggestion
in order to implement formatting.
```

Nit: Reduce the number of `implement`

MichaReiser, line 89: Nit
```suggestion
in order to correctly implement the formatter. Alternatively, you can locally run the playground by following the [playground instructions]
```

crates/rome_playground/src/App.tsx
```diff
@@ -113,8 +124,8 @@ function App() {
             <div className="w-1/2 p-5">
               <CodeEditor
                 value={code}
-                language="js"
-                placeholder="Enter JS here"
+                language={language}
```
MichaReiser, line 127: It's better to compute information that can be derived from other values than storing it in another state and trying to keep it up to date
```js
function getLanguage(isTypeScript, isJsx): "js" | "jsx" | "typescript" {
    if (isTypeScript) {
      return "typescript";
    } else if (isJsx) {
      return "jsx";
    } else {
      return "js"
    }
  }, [isJsx, isTypeScript]);
```

```suggestion
                language={getLanguage(isTypeScript, isJsx)}
```

MichaReiser (approved):

crates/rome_playground/src/App.tsx
ematipico: This is micro optimization :)

Also, with your suggestion the function `getLanguage` gets called even when `isTypeScript` or `isJsx` don't change, at every re-render.

MichaReiser: Yes, but computing this is cheap. Also, having state out of sync is a common source of problems and it's a common recommendation to recompute it rather than storing it (and if it's expensive, use `useMemo`).

ematipico: >  Also, having state out of sync

I really don't understand what this

MichaReiser: In case someone forgets to call `setState` after changing some language props. I know, you use `useEffect` but that means that the `language` variable is "stable` until the effect completed. Also be aware, that the hook only runs after render. So react first renders with the default "language" and only then runs the hook.

ematipico:
@NicholasLYang  I am going to go ahead and merge the PR. Let us know if there's something you would like to change.

PR #2321, opened 2022-03-29
chore: add banner to rome classic CLI

ematipico:
## Summary

This PR adds a small banner to the **Rome classic**. The banner is displayed every time the CLI is run, regardless of arguments and commands.

This is the look and feel:

Screenshot 2022-03-29 at 14 35 07

## Test Plan

None

### DIFF
--- a/internal/cli-flags/Parser.ts
+++ b/internal/cli-flags/Parser.ts
@@ -499,7 +499,7 @@ export default class Parser<T> {
 	public async init(): Promise<T> {
 		const flagsConsumer = this.getFlagsConsumer();
 		const {flags} = flagsConsumer;
-
+		writeBanner(this.reporter);
 		// Show help for --version
 		const version = this.opts.version;
 		if (version !== undefined) {
@@ -596,6 +596,7 @@ export default class Parser<T> {
 			) {
 				this.opts.onRunHiddenCommand(this.reporter);
 			}
+
 			await definedCommand.command.callback(definedCommand.flags);
 		}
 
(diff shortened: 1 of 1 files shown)


internal/cli-flags/Parser.ts
```diff
@@ -499,7 +499,7 @@ export default class Parser<T> {
 	public async init(): Promise<T> {
 		const flagsConsumer = this.getFlagsConsumer();
 		const {flags} = flagsConsumer;
-
+		writeBanner(this.reporter);
```
MichaReiser, line 502: Are there any commands that generate a JSON output where we may not want to print the banner?

internal/cli-flags/Parser.ts
```diff
@@ -1315,3 +1316,23 @@ export class ParserInterface<T> {
 		this.parser.addCommand(opts);
 	}
 }
+
+function writeBanner(reporter: Reporter) {
+	reporter.hr(markup`Please read carefully`);
```
MichaReiser, line 1321: This sounds intimating and doesn't provide additional information. That's why I would leave it off.

internal/cli-flags/Parser.ts
```diff
@@ -1315,3 +1316,23 @@ export class ParserInterface<T> {
...
+	reporter.indentSync(() => {
+		reporter.warn(markup`This is rome classic and the project is archived.`);
+		reporter.br();
+		reporter.warn(
+			markup`The project is now <underline><hyperlink target="https://rome.tools/blog/2021/09/21/rome-will-be-rewritten-in-rust">written in Rust</hyperlink></underline>.`,
+		);
+		reporter.br();
+		reporter.info(
+			markup`Please update the project to the <emphasis>next</emphasis> tag.`,
```
MichaReiser, line 1330: We should tell them the name of the package here. It may not be clear to the user how they can install the next`. 

```
Install `rome@next` if you want to use the new blazing fast rust rewrite of Rome. 
```

internal/cli-flags/Parser.ts
```diff
@@ -1315,3 +1316,23 @@ export class ParserInterface<T> {
...
+		);
+		reporter.br();
+		reporter.info(
+			markup`Please visit the official website for details on how to install the correct version: <underline><hyperlink target="https://rome.tools/#getting-started">https://rome.tools/#getting-started</hyperlink></underline>.`,
+		);
```
MichaReiser, line 1335:
```suggestion
			markup`Visit the getting started guide for in depth installation instructions: <underline><hyperlink target="https://rome.tools/#getting-started">https://rome.tools/#getting-started</hyperlink></underline>.`,
		);
```

MichaReiser (approved):

PR #2328, opened 2022-03-30
feat(rome_console): Display trait and diff printing

leops:
## Summary

This PR introduces a new `rome_console::fmt::Display` trait that is directly equivalent to `std::fmt::Display`: the trait declares a single `fn fmt(&self, &mut Formatter) -> Result` method where implementing types can decide on how they want to be printed by submitting additional printing commands to the `Formatter`. Said formatter implements a `write_markup` method that allows `Display`-able types to submit new markup pieces to the console, as well as two `write_fmt` and `write_str` methods equivalent to the ones found in `std::fmt::Formatter` and making the console formatter compatible with the `write!` and `writeln!` macros.

To make use of this new trait the PR also modifies the `markup!` procedural macro to allow any literal (of a type that implements `Display`) inside markup elements, where previously only string literals were allowed as they were treated as Rust format strings. Since literals are now treated as plain values, expressions of type implementing `Display` can instead be added to markup using curly braces groups:
```rust
markup! {
    <Info>"Parsed "{count}" files in "{duration}</Info>
}
```

Finally, this PR adds a new `rome_console::diff::Diff` struct making use of the new `Display` trait to calculate and print a formatted diff between two pieces of text to the console, with the following formatting:
```
@@ -3,10 +3,8 @@
 3 dolor                 3 dolor
 4 sit                   4 sit
 5 amet,                 5 amet,
 6 function              6 function name(args) {
 7 name(                 7 }
 8     args             
 9 ) {}                 
10 consectetur           8 consectetur
11 adipiscing            9 adipiscing
12 elit,                10 elit,
@@ -15,7 +13,5 @@
15 eiusmod              13 eiusmod
16                      14 
17 incididunt           15 incididunt
18 function             16 function name(args) {
19 name(                17 }
20     args             
21 ) {}                 
```

With colors:
diff_colors

### DIFF
--- a/crates/rome_console/src/diff.rs
+++ b/crates/rome_console/src/diff.rs
@@ -8,8 +8,69 @@ use crate::{
     markup,
 };
 
+/// Determines how a diff should be printed in the console
+pub enum DiffMode {
+    /// Render the diff in a single column by stacking the changes
+    ///
+    /// # Example
+    /// ```text
+    ///       | @@ -3,10 +3,8 @@
+    ///  2  2 |   dolor
+    ///  3  3 |   sit
+    ///  4  4 |   amet,
+    ///  5    | - function
+    ///  6    | - name(
+    ///  7    | -     args
+    ///  8    | - ) {}
+    ///     5 | + function name(args) {
+    ///     6 | + }
+    ///  9  7 |   consectetur
+    /// 10  8 |   adipiscing
+    /// 11  9 |   elit,
+    ///       | @@ -15,7 +13,5 @@
+    /// 14 12 |   eiusmod
+    /// 15 13 |   
+    /// 16 14 |   incididunt
+    /// 17    | - function
+    /// 18    | - name(
+    /// 19    | -     args
+    /// 20    | - ) {}
+    ///    15 | + function name(args) {
+    ///    16 | + }
+    /// ```
+    Unified,
+    /// Render the diff over two columns, with the old text on the
+    /// left and the new text on the right
+    ///
+    /// # Example
+    /// ```text
+    /// @@ -3,10 +3,8 @@
+    ///  3   dolor                 3   dolor
+    ///  4   sit                   4   sit
+    ///  5   amet,                 5   amet,
+    ///  6 - function              6 + function name(args) {
+    ///  7 - name(                 7 + }
+    ///  8 -     args
+    ///  9 - ) {}
+    /// 10   consectetur           8   consectetur
+    /// 11   adipiscing            9   adipiscing
+    /// 12   elit,                10   elit,
+    /// @@ -15,7 +13,5 @@
+    /// 15   eiusmod              13   eiusmod
+    /// 16                        14
+    /// 17   incididunt           15   incididunt
+    /// 18 - function             16 + function name(args) {
+    /// 19 - name(                17 + }
+    /// 20 -     args
+    /// 21 - ) {}
+    /// ```
+    Split,
+}
+
 /// Utility struct to print a diff between two texts in the console
 pub struct Diff<'a> {
+    /// Rendering mode for this diff
+    pub mode: DiffMode,
     /// The previous version of the text
     pub left: &'a str,
     /// The next version of the text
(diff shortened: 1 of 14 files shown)


crates/rome_console/src/diff.rs
```diff
@@ -0,0 +1,270 @@
...
+12 elit,                10 elit,
+@@ -15,7 +13,5 @@
+15 eiusmod              13 eiusmod
+16                      14 
+17 incididunt           15 incididunt
+18 function             16 function name(args) {
+19 name(                17 }
+20     args             
+21 ) {}                 
+";
```
MichaReiser, line 248: I'm having a hard time understanding this diff. Could that be because the colors are missing? CAn we add `-` and `+` signs in front of changed lines?

crates/rome_console/src/fmt.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl<'a> MarkupElements<'a> {
+    /// Iterates on all the element slices depth-first
+    fn for_each(&self, func: &mut impl FnMut(&'a [MarkupElement])) {
+        if let Self::Node(parent, elem) = self {
+            parent.for_each(func);
+            func(elem);
+        }
+    }
```
MichaReiser, line 20: Nit: Why not implement `iter()` or `Iterator` so that the caller can decide if they want to find a specific element or if they want to execute a function?

crates/rome_console/src/lib.rs
```diff
@@ -31,7 +33,9 @@ where
     E: WriteColor + Sync + RefUnwindSafe,
 {
     fn message(&mut self, args: Markup) {
-        args.print(&mut self.out).unwrap();
+        fmt::Formatter::new(&mut self.out)
+            .write_markup(args)
+            .unwrap();
```
MichaReiser, line 38: Should `message` return a `Result` so that we can avoid the `unwrap` calls here?

crates/rome_console/src/lib.rs
```diff
@@ -74,7 +78,8 @@ impl Console for BufferConsole {
 
         {
             let mut writer = NoColor::new(&mut message);
-            args.print(&mut writer).unwrap();
+            let mut fmt = fmt::Formatter::new(&mut writer);
+            fmt.write_markup(args).unwrap();
```
MichaReiser, line 82: Same here, return `Result`?

MichaReiser (approved):
This is so cool! 

I'm not so sure about the diff format. I have actually no idea what I'm looking at or what has changed in the Diff you added to the summary (or test). I think using something that's closer to what `rustfmt` uses is easier to understand and also works if color printing is disabled.

ematipico:
Is there some plan to make the diff to how it was? This is how the old diff looked like: 
Screenshot 2022-03-31 at 07 49 40

crates/rome_console/src/fmt.rs
```diff
@@ -0,0 +1,217 @@
...
+
+/// The [Formatter] is the `rome_console` equivalent to [std::fmt::Formatter]:
+/// its never constructed directly by consumers and can only be used through
+/// the mutable reference passed to implementations of the [Display] trait).
+/// It manages the state of the markup being printed, and implementations of
+/// [Display] can call into its methods to append content into the current
+/// printing session
+pub struct Formatter<'fmt> {
+    state: MarkupElements<'fmt>,
+    writer: &'fmt mut dyn WriteColor,
```
ematipico, line 31: Can we document these two little fellas?

crates/rome_console/src/fmt.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Return a new instance of the [Formatter] with `elements` appended to its element stack
+    fn with_elements<'b>(&'b mut self, elements: &'b [MarkupElement]) -> Formatter<'b> {
+        Formatter {
+            state: MarkupElements::Node(&self.state, elements),
+            writer: self.writer,
+        }
+    }
+}
+
+impl<'fmt> Formatter<'fmt> {
```
ematipico, line 52: Why this `impl` and the `impl` at line 34 are separate? Can they be unified?

crates/rome_console/src/fmt.rs
```diff
@@ -0,0 +1,217 @@
...
+    pub fn write_markup(&mut self, markup: Markup) -> io::Result<()> {
+        for node in markup.0 {
+            let mut fmt = self.with_elements(node.elements);
+            node.content.fmt(&mut fmt)?;
+        }
+
+        Ok(())
+    }
+
+    fn with_format(
```
ematipico, line 63: I know it's a private function, but can we document its usage please?

crates/rome_console/src/fmt.rs
```diff
@@ -0,0 +1,217 @@
...
+        Formatter {
+            state: MarkupElements::Node(&self.state, elements),
+            writer: self.writer,
+        }
+    }
+}
+
+impl<'fmt> Formatter<'fmt> {
+    /// Write a piece of markup into this formatter
+    pub fn write_markup(&mut self, markup: Markup) -> io::Result<()> {
```
ematipico, line 54: Do you think we can alias this `io::Result`? Something like `ConsoleResult`?

crates/rome_console/src/markup.rs
```diff
@@ -46,18 +45,24 @@ impl MarkupElement {
                 color.set_fg(Some(Color::Yellow));
             }
             MarkupElement::Info => {
-                color.set_fg(Some(Color::Blue));
+                // Blue is really difficult to see on the standard windows command line
```
ematipico, line 48: Is it PowerShell? (sorry, haven't used windows in looooong time)

crates/rome_console/src/diff.rs
leops: The colors do help a lot to understand what's going on (I added an image in the PR description to show what it looks like), but I agree it's also a good idea to add the +/- signs somewhere to help make it more accessible

crates/rome_console/tests/macro.rs
```diff
@@ -8,17 +8,17 @@ fn test_macro() {
...
     rome_markup::markup! {
-        <Info><Emphasis>"{category}"</Emphasis>" Commands"</Info>
+        <Info><Emphasis>{category}</Emphasis>" Commands"</Info>
     }
     {
         Markup(markup) => {
             let node_0 = &markup[0];
             assert_eq!(&node_0.elements, &[MarkupElement::Info, MarkupElement::Emphasis]);
-            assert_eq!(node_0.content.to_string(), category.to_string());
+            // assert_eq!(node_0.content.to_string(), category.to_string());
```
ematipico, line 17: should we delete?

crates/rome_console/tests/macro.rs
```diff
@@ -8,17 +8,17 @@ fn test_macro() {
...
 
             let node_1 = &markup[1];
             assert_eq!(&node_1.elements, &[MarkupElement::Info]);
-            assert_eq!(node_1.content.to_string(), " Commands".to_string());
+            // assert_eq!(node_1.content.to_string(), " Commands".to_string());
```
ematipico, line 21: delete?

ematipico (commented):
It looks good to me, although I think we should really start documenting well what we are implementing before it's too late (AKA we forgot what we've done).

The old markup had a great documentation, especially because it had visual images: https://github.com/rome/tools/tree/archived-js/internal/markup

I think we should do the same, it would really help developers to understand what markup to write and how it would look like. I would not consider it a blocker, but I would like to understand if that's the intention of the team going forward.

crates/rome_console/src/fmt.rs
leops: The fact that it's both a stack-allocated linked-list (where each node points to its parent) and is iterated starting from the root towards to deepest child node means implementing an `Iterator` would have to keep some heap allocated state around to be able to backtrack through the list, while by using recursion the call stack itself is doing that bookkeeping making the implementation simpler

crates/rome_console/src/lib.rs
leops: The decision to panic on error here was based on `println!` doing the same thing, with the idea that printing to the console will never fail under the most common circumstances, and if it did there would be a good chance printing the error would also fail anyway, so having to do error handling on every call to the console felt unnecessary

crates/rome_console/src/diff.rs
MichaReiser: Yes the colors help a lot! 

I only just now understood that it uses a two-column layout. I'm not sure how well this works on terminals with limited space (works with your example but the files aren't using a 80 characters text width).

leops:
> Is there some plan to make the diff how it was?
This is really a matter of personal preference but I find side-by-side diffs more readable, and since it's a bit more complicated to implement than showing the merged diff because of the two-column layout I figured I would implement side-by-side diffs first, then it's really straightforward to add a second implementation or a configuration option to print the merged diff instead later on

crates/rome_console/src/markup.rs
leops: This bit originally comes from [here](https://github.com/rome/tools/blob/main/crates/rome_diagnostics/src/codespan/term/config.rs#L185-L189), and I decided to use the same logic when printing markup since at least when using the Windows Terminals the default blue text color is using a very dark shade that's hard to read on the default dark background

crates/rome_console/tests/macro.rs
leops: This is a check I intend to bring back later on, but it requires a public API to make the `Markup` inspectable that's not in place yet

ematipico:
> > Is there some plan to make the diff how it was?
> 
> This is really a matter of personal preference but I find side-by-side diffs more readable, and since it's a bit more complicated to implement than showing the merged diff because of the two-column layout I figured I would implement side-by-side diffs first, then it's really straightforward to add a second implementation or a configuration option to print the merged diff instead later on

I am not sure is **only** a matter of personal preference. I don't mind having one, the other, or both. I think we should evaluate where it's going to be used.

Side-by-side is good for long diffs because it takes advantage of the width of the screen, but what about big changes with large diffs in indentation? How does it behave in cases where the two columns exceed the width of the screen?

Stacked diffs instead it's good in cases of large diffs where the code might exceed the width of the screen. Although they are really long and tedious. 


Now, I am skeptical of this new approach - and I am so sorry about it, really am - because I come from the old diffing (which I used a lot in many situations: snapshots, terminal, CI, web, etc.), so I had the chance to use it real world example and it worked really well. I wonder if we can more examples or tests? Or, if not possible to do it now, what's your plan going forward? How to use it, where to use it, etc.

MichaReiser:
> > > Is there some plan to make the diff how it was?
> > 
> > 
> > This is really a matter of personal preference but I find side-by-side diffs more readable, and since it's a bit more complicated to implement than showing the merged diff because of the two-column layout I figured I would implement side-by-side diffs first, then it's really straightforward to add a second implementation or a configuration option to print the merged diff instead later on
> 
> I am not sure is **only** a matter of personal preference. I don't mind having one, the other, or both. I think we should evaluate where it's going to be used.
> 
> Side-by-side is good for long diffs because it takes advantage of the width of the screen, but what about big changes with large diffs in indentation? How does it behave in cases where the two columns exceed the width of the screen?
> 
> Stacked diffs instead it's good in cases of large diffs where the code might exceed the width of the screen. Although they are really long and tedious.
> 
> Now, I am skeptical of this new approach - and I am so sorry about it, really am - because I come from the old diffing (which I used a lot in many situations: snapshots, terminal, CI, web, etc.), so I had the chance to use it real world example and it worked really well. I wonder if we can more examples or tests? Or, if not possible to do it now, what's your plan going forward? How to use it, where to use it, etc.

I don't consider this blocking. We can later decide to change the format or support both (either user configurable or Rome decides deppending on the use case). 

What may be worth taking over from rome classic is to label the two files to make it sure which version is which.

leops:
> > > Is there some plan to make the diff how it was?
> > 
> > 
> > This is really a matter of personal preference but I find side-by-side diffs more readable, and since it's a bit more complicated to implement than showing the merged diff because of the two-column layout I figured I would implement side-by-side diffs first, then it's really straightforward to add a second implementation or a configuration option to print the merged diff instead later on
> 
> I am not sure is **only** a matter of personal preference. I don't mind having one, the other, or both. I think we should evaluate where it's going to be used.
> 
> Side-by-side is good for long diffs because it takes advantage of the width of the screen, but what about big changes with large diffs in indentation? How does it behave in cases where the two columns exceed the width of the screen?
> 
> Stacked diffs instead it's good in cases of large diffs where the code might exceed the width of the screen. Although they are really long and tedious.
> 
> Now, I am skeptical of this new approach - and I am so sorry about it, really am - because I come from the old diffing (which I used a lot in many situations: snapshots, terminal, CI, web, etc.), so I had the chance to use it real world example and it worked really well. I wonder if we can more examples or tests? Or, if not possible to do it now, what's your plan going forward? How to use it, where to use it, etc.

As I mentioned adding stacked diff rendering is very easy, so I could do it right in this PR, but it would be a "manual switch" using a new `mode: DiffMode` property in the `Diff` struct with `DiffMode` being an enum the let the code creating the markup decide how the diff should be printed (I'll call the variants `Unified` and `Split` since that's how GitHub calls them).

Obviously the most problematic part of side-by-side printing is the layout, as it's prone to overflowing on small terminal windows so I want to add a method for the diff renderer to try and query the size of the terminal (in a cross-platform way) and adjust the rendering of diffs based on that metric (I think this would work by introducing a third `Auto` variant to the `DiffMode` enum that would try to print in split mode but fall back to unified if the size of the terminal could not be determined or if the content being printed is too large)

> What may be worth taking over from rome classic is to label the two files to make it sure which version is which.

This is something I explicitly decided not to implement in the `Diff` struct itself, since it's supposed to be a very generic utility for comparing two pieces of text. The intent if for users of this struct to insert additional markup themselves to print the file names if they need to:

```rust
console.message(markup! {
    "--- "{prev_file_name}"\n"
    "+++ "{new_file_name}"\n"
    {diff}
});
```

ematipico:
Thank you very much for addressing my concerns!

ematipico (approved):

PR #2332, opened 2022-03-31
fix(rome_playground): Unicode Support

MichaReiser:
They playground crashed because `btoa` only supports ASCII.
This PR ensures that non ASCII characters are propertly encoded before running the code throuh `btoa`.

## Test Plan

Screenshot from 2022-03-31 11-38-50

### DIFF
--- a/.github/workflows/deploy_playground.yml
+++ b/.github/workflows/deploy_playground.yml
@@ -22,7 +22,7 @@ jobs:
           node-version: '14'
       - run: npm install --prefix crates/rome_playground
       - uses: jetli/wasm-pack-action@v0.3.0
-      - run: wasm-pack build crates/rome_playground --target web
+      - run: wasm-pack build crates/rome_playground --target web --release
       - run: npm run build --prefix crates/rome_playground
       - name: Publish
         uses: cloudflare/wrangler-action@1.3.0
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_playground.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           node-version: '14'
       - run: npm install --prefix crates/rome_playground
       - uses: jetli/wasm-pack-action@v0.3.0
-      - run: wasm-pack build crates/rome_playground --target web
+      - run: wasm-pack build crates/rome_playground --target web --release
```
MichaReiser, line 25: This reduces the wasm size by 0.5 mb

crates/rome_playground/src/App.tsx
```diff
@@ -2,7 +2,7 @@ import CodeEditor from "@uiw/react-textarea-code-editor";
 import "react-tabs/style/react-tabs.css";
 import init, { run } from "../pkg/rome_playground";
 import { Tabs, Tab, TabList, TabPanel } from "react-tabs";
-import {useEffect, useMemo, useState} from "react";
+import { useEffect, useMemo, useState } from "react";
```
MichaReiser, line 5: Nit: We should probably move the playground into the website directory. It isn't a proper "rome" rust crate

.github/workflows/deploy_playground.yml
xunilrj: hum... if we have not done anything to reduce, can this help?
https://rustwasm.github.io/book/reference/code-size.html

xunilrj (approved):

crates/rome_playground/src/App.tsx
ematipico: It's the same thing that I suggested a while ago.

ematipico (approved):

PR #2333, opened 2022-03-31
feat(npm): add a postinstall script to the cli package

leops:
## Summary

Add a `postinstall` script to the `rome` npm package to verify the installation of the native binary, and print a warning to the user immediately if something went wrong.

I decided to only print a warning when something is wrong instead of exiting with an error so that it won't cause an `npm install` on a whole workspace where `rome` is a dependency to fail, in case the user didn't intend to use the Rome CLI anyway.

### DIFF
--- a/npm/rome/scripts/postinstall.js
+++ b/npm/rome/scripts/postinstall.js
@@ -1,0 +1,44 @@
+const { platform, arch } = process;
+
+const PLATFORMS = {
+	win32: {
+		x64: "@rometools/cli-win32-x64/rome.exe",
+		arm64: "@rometools/cli-win32-arm64/rome.exe",
+	},
+	darwin: {
+		x64: "@rometools/cli-darwin-x64/rome",
+		arm64: "@rometools/cli-darwin-arm64/rome",
+	},
+	linux: {
+		x64: "@rometools/cli-linux-x64/rome",
+		arm64: "@rometools/cli-linux-arm64/rome",
+	},
+};
+
+const binName = PLATFORMS?.[platform]?.[arch];
+if (binName) {
+	let binPath;
+	try {
+		binPath = require.resolve(binName);
+	} catch (err) {
+		console.warn(
+			`The Rome CLI postinstall script failed to resolve the binary file "${binName}". ` + "Running Rome from the npm package will probably not work correctly.",
+		);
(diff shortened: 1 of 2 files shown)


npm/rome/scripts/postinstall.js
```diff
@@ -0,0 +1,44 @@
...
+};
+
+const binName = PLATFORMS?.[platform]?.[arch];
+if (binName) {
+	let binPath;
+	try {
+		binPath = require.resolve(binName);
+	} catch (err) {
+		console.warn(
+			`The Rome CLI postinstall script failed to resolve the binary file "${binName}". ` + "Running Rome from the npm package will probably not work correctly.",
```
MichaReiser, line 25: Interesting formatting...

npm/rome/scripts/postinstall.js
```diff
@@ -0,0 +1,44 @@
+const { platform, arch } = process;
+
+const PLATFORMS = {
```
MichaReiser, line 3: Do we still need the `bin/rome` file. This file looks very much the same?

leops: The logic in both files is mostly the same, the difference being that the postinstall script only prints a warning while the wrapper script in `bin/rome` prints an error and exit with a non-zero status code if an error happens

npm/rome/scripts/postinstall.js
leops: That's a bug in how binary expressions are formatted, they don't create a new group if there is less than three items but that doesn't account for the size of these sub-expressions

npm/rome/scripts/postinstall.js
```diff
@@ -0,0 +1,44 @@
...
+		binPath = require.resolve(binName);
+	} catch (err) {
+		console.warn(
+			`The Rome CLI postinstall script failed to resolve the binary file "${binName}". ` + "Running Rome from the npm package will probably not work correctly.",
+		);
+	}
+
+	if (binPath) {
+		try {
+			require("fs").chmodSync(binPath, 0o755);
```
ematipico, line 31: Let's use `node:fs`

ematipico (approved):
Looks good to me, just a minor comment.

PR #2334, opened 2022-03-31
chore: rename `rome_formatter` to `rome_js_formatter`

ematipico:
## Summary

This the first step of making our formatter generic and reusable for other languages/data structures.

As first step, this PR names the crate `rome_formatter` to `rome_js_formatter`.

## Test Plan

Existing CI. No tests should break.


ematipico:
> So the intent here is to move everything to the `rome_js_formatter` crate at first, then move the common code back into `rome_formatter` once we start implementing formatting for additional languages ?

Yes exactly, there are different reasons to do so:
- at the moment the formatter is bound to `SyntaxNode`, etc, which are an alias of `SyntaxNode<JsLanguage>`. The moment we will create a new language (for simplicity, while still using `rome_rowan`, for example `JsonLanguage`), this won't work with the current implementation;
- the formatter should try to be language agnostic; 
- one day (way in the future) we can potentially release `rome_formatter` as standalone crate and anybody should be able to implement it using their language;


For now I think we should focus on the first part.

PR #2335, opened 2022-03-31
chore: Move playground to `website` directory

MichaReiser:
## Summary

The playground isn't a crate that exposes any functionality of Rome. Its sole purpose is to expose Rome's API. There, I decided to move it into the website directory.


website/playground/package.json
```diff
@@ -2,10 +2,11 @@
   "name": "playground",
   "version": "0.0.0",
   "scripts": {
-    "dev": "vite",
+    "start": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "format": "cargo run --bin rome format ./src"
+		"build:wasm": "wasm-pack build --target web --release",
```
NicholasLYang, line 8: super tiny nit, but in my experience it's a convention that if you have a `build:wasm` task and `build` task then `build` should also do what `build:wasm` does. So you could have `vite` be `build:js` then `start` be `build:wasm & build:js`

ematipico: I second that!

website/playground/package.json
```diff
@@ -2,10 +2,12 @@
...
   "version": "0.0.0",
   "scripts": {
-    "dev": "vite",
-    "build": "tsc && vite build",
+    "start": "vite",
     "preview": "vite preview",
-    "format": "cargo run --bin rome format ./src"
+		"build": "npm run build:js && npm run build:wasm",
+    "build:js": "tsc && vite build",
+		"build:wasm": "wasm-pack build --target web --release",
```
ematipico, line 9: There seems to be some issue with the formatting

ematipico (approved):

PR #2337, opened 2022-03-31
feat(rome_cli): integrate the new diff printing with CI mode

leops:
## Summary

This integrates the new `Diff` printer with the CLI to show the difference between the current content of a file and the expected output from the formatter in CI mode. It's displaying the diff in unified mode by default since I haven't implemented an automatic layout detection logic yet.

I also included some safety checks in the diff renderer and the CLI to guard around printing files with very long lines or very large files in general, and added a new `rome-cli` Cargo alias to allow running the CLI in release mode with `cargo rome-cli`


crates/rome_cli/src/commands/format.rs
```diff
@@ -170,8 +182,40 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
+                file_name,
+                old,
+                new,
+            } => {
+                // Skip printing the diff for files over 1Mb (probably a minified file)
+                let max_len = old.len().max(new.len());
+                if max_len >= 1_000_000 {
+                    session.app.console.message(markup! {
+                        {file_name}": "
+                        <Error>"error[CI]"</Error>": File content differs from formatting output\n"
```
ematipico, line 200: What do you think we create a `console.br()` function that adds a line break? Rome classic had something like this: `reporter.br()`.

ematipico (approved):
Looks good to me!

Quick question: what's your plan around the usage of the markup? At the moment we have `console.message`, which is really generic, and there aren't guidelines about how to use it in cases of errors, info, warnings, etc. Everything boils down to how the markup is used.

Rome classic, with its reporter, had various methods:. `.info()`, `.warning()`, `.success()`, `.command()`, `.error()`, etc. There were small wrappers around the markup (passed as parameter) and they used to set color, and prefix (an unicode icon.. for example the warning had a little triangle), and the stream too (stderr or stdout).

Is there some plan to mimic what we had?

leops:
> Quick question: what's your plan around the usage of the markup? At the moment we have `console.message`, which is really generic, and there aren't guidelines about how to use it in cases of errors, info, warnings, etc. Everything boils down to how the markup is used.

Currently the `Console` trait has two methods, one to print markup and one to print diagnostics, but ultimately I would want the diagnostics to implement the `Display` trait and be printed with markup as well, so the console trait itself would only have a single `print` method that takes a piece of markup along with some metadata (I think at least some kind of "log level" to select which stream to print into, but this could include additional information such as the file name and line number the message was emitted from similar to what the `log` and `tracing` crates do). All the utility methods would then be implemented on top of that generic `print`, in a `ConsoleExt` trait implemented for all the types implementing `Console`

PR #2341, opened 2022-04-01
feat(rome_console): add a Codespan utility struct to display annotated code fragments

leops:
## Summary

This PR adds a simplified copy of the code in `crates/rome_diagnostics/src/codespan` to `rome_console`, exposing a `Codespan` struct intended to display annotated code as markup in the console. This is the first step in a refactor of `Diagnostic` to make the struct implement `rome_console::fmt::Display` on top of common markup and utilities, essentially turning the rendering logic for diagnostics into the following markup:

```rust
markup! {
    {level}"["{code}"]: "{message}"\n"
    {codespan}
    "Suggested fix:\n"
    {diff}
}
```

## Example

```rust
let source = SourceFile::new(SOURCE);

let mut codespan = Codespan::new(
    &source,
    Severity::Error,
    Some(Locus::FileLocation {
        name: String::from("file_name"),
        location: Location {
            line_number: 2,
            column_number: 12,
        },
    }),
);

codespan.add_label(Label {
    style: LabelStyle::Primary,
    range: 40..55,
    message: markup! {
        <Emphasis>"Important"</Emphasis>" message"
    },
})?;

codespan.add_label(Label {
    style: LabelStyle::Secondary,
    range: 71..99,
    message: markup! {
        "Multiline message"
    },
})?;

codespan.add_label(Label {
    style: LabelStyle::Secondary,
    range: 100..109,
    message: markup! {
        "Secondary message"
    },
})?;
```

codespan_colors

### DIFF
--- a/crates/rome_console/src/codespan/mod.rs
+++ b/crates/rome_console/src/codespan/mod.rs
@@ -1,0 +1,592 @@
+///! This module if a fork of https://github.com/brendanzab/codespan,
+/// adapted to use the `rome_console` markup for formatting
+use std::collections::BTreeMap;
+use std::io;
+use std::ops::Range;
+
+use crate::fmt::{Display, Formatter};
+use crate::Markup;
+
+use self::render::{MultiLabel, Renderer, SingleLabel};
+
+mod render;
+
+const START_CONTEXT_LINES: usize = 3;
+const END_CONTEXT_LINES: usize = 1;
+
+/// A label describing an underlined region of code associated with a diagnostic.
+#[derive(Clone)]
+pub struct Label<'diagnostic> {
+    /// The style of the label.
+    pub style: LabelStyle,
+    /// The range in bytes we are going to include in the final snippet.
+    pub range: Range<usize>,
+    /// An optional message to provide some additional information for the
+    /// underlined code. These should not include line breaks.
+    pub message: Markup<'diagnostic>,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd)]
+pub enum LabelStyle {
+    /// Labels that describe the primary cause of a diagnostic.
+    Primary,
+    /// Labels that provide additional context for a diagnostic.
+    Secondary,
+}
+
+/// A severity level for diagnostic messages.
+///
+/// These are ordered in the following way:
+#[derive(Copy, Clone, PartialEq, Hash, Debug)]
+pub enum Severity {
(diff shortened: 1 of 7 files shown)


crates/rome_console/src/codespan/mod.rs
```diff
@@ -0,0 +1,592 @@
+///! This module if a fork of https://github.com/brendanzab/codespan,
+/// adapted to use the `rome_console` markup for formatting
```
ematipico, line 2:
```suggestion
//! This module if a fork of https://github.com/brendanzab/codespan ,
//! adapted to use the `rome_console` markup for formatting
```

ematipico: Could we add this project to the credits section? https://rome.tools/credits/#forks

crates/rome_console/src/codespan/mod.rs
```diff
@@ -0,0 +1,592 @@
...
+    /// An unexpected bug.
+    Bug,
+    /// An error.
+    Error,
+    /// A warning.
+    Warning,
+    /// A note.
+    Note,
+    /// A help message.
+    Help,
```
ematipico, line 51: What's the the difference between these two? How do we know when to use one or the other? Should we add it to the documentation?

crates/rome_console/src/codespan/mod.rs
```diff
@@ -0,0 +1,592 @@
...
+    pub line_number: usize,
+    /// The user-facing column number.
+    pub column_number: usize,
+}
+
+pub struct Codespan<'diagnostic> {
+    severity: Severity,
+    locus: Option<Locus>,
+    source_file: &'diagnostic SourceFile<'diagnostic>,
+    labeled_file: Option<LabeledFile<'diagnostic>>,
```
ematipico, line 81: What's the difference between these two?

ematipico (approved):

xunilrj (approved):

crates/rome_console/src/codespan/mod.rs
leops: The basic difference seem to be that `Note` is printed in green and is converted to the `Hint` severity in the LSP, while `Help` is printed in blue and is converted to the `Information` severity in the LSP. Overall they don't seem to be used much yet, a quick grep over the entire codebase returned no usage of the help severity while the note severity is used for only two diagnostics, and even then it's in footer notes and not in actual codespan labels.

PR #2342, opened 2022-04-01
refactor(rome_js_formatter): move language agnostic code into `rome_formatter`

ematipico:
## Summary

This PR moves all the language agnostic code of the formatter from `rome_js_formatter` to `rome_formatter`. The only "language" dependency inside the formatter now is `rome_rowan` as we need `TextRange`.

Some of the functions like `join_elements_soft_line` are still inside `rome_js_formatter` because they are tight to `AstNode`, which comes from `rome_js_syntax`.

I didn't move `Formatter` and `ToFormatElement` because they are more difficult to move. I think we will need to refactor them somehow. `Formatter` has an heavy dependency on `SyntaxNode` and `SyntaxToken`. While `ToFormatElement` we might not be able to move it because we can't define trait inside `rome_formatter` and then doing an implementation of 
`impl ToFormatElement for SyntaxNode` (inside `cst.rs`)  because it will trigger a compiler error.

## Test Plan

Current CI should not break


crates/rome_js_formatter/src/join_elements.rs
```diff
@@ -0,0 +1,88 @@
...
+pub fn join_elements_hard_line<I, N>(elements: I) -> FormatElement
+where
+    I: IntoIterator<Item = (N, FormatElement)>,
+    N: AstNode,
+{
+    join_elements_with(elements, hard_line_break)
+}
+
+#[inline]
+pub fn join_elements_with<I, N>(elements: I, separator: fn() -> FormatElement) -> FormatElement
```
leops, line 39: These helpers could be moved to `rome_formatter` by using an iterator over explicit `SyntaxNode` instead of a generic `N: AstNode`, but keeping the generic implementation of each function in `rome_js_formatter` as a simple wrapper over the one in `rome_formatter` so we don't have to rewrite each call site

crates/rome_js_formatter/src/lib.rs
```diff
@@ -398,19 +207,20 @@ pub fn format_range(
...
     };
 
     let input_range = TextRange::new(start_source, end_source);
     let output_range = TextRange::new(start_dest, end_dest);
-    let code = &formatted.code[output_range];
-
+    let sourcemap = Vec::from(formatted.sourcemap());
+    let verbatim = Vec::from(formatted.verbatim());
+    let code = &formatted.into_code()[output_range];
     Ok(Formatted::new(
```
leops, line 219: Most of the range formatting logic (outside of the actual call to `format_node`) could be moved to `rome_formatter`, to avoid having to poke holes into the abstraction by making some of these function public

crates/rome_js_formatter/src/join_elements.rs
ematipico: When you mention `SyntaxNode`, you mean the one we have in `rome_rowan`?

ematipico: Uhm, it doesn't seem that would work, because of this: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/format_element.rs#L416

`SyntaxNode<L>` doesn't have the `.syntax()` method. (it belongs to the `AstNode` trait)

ematipico: Even if we fixed the problem highlighted, this has a waterfall effect on all the usages:
```
   --> crates/rome_js_formatter/src/js/any/class.rs:45:17
    |
45  |                 join_elements_hard_line(
    |                 ^^^^^^^^^^^^^^^^^^^^^^^ expected struct `SyntaxNode`, found enum `JsAnyClassMember`
````

leops: The `syntax()` method is the only method of `AstNode` that's being used in these functions and it returns a `SyntaxNode` so the idea was to remove the call to syntax and pass the iterator of `SyntaxNode` directly to the version of the function that exists in `rome_formatter`, and have the implementation of these functions in `rome_js_formatter` simply call the generic one by wrapping the iterator with `.map(|node| node.syntax())`

crates/rome_js_formatter/src/lib.rs
ematipico: Unfortunately not at the moment. `FormatResult` has a dependency on `FormatError`, which implements a `From<SyntaxError>`. `SyntaxError` now leave under the crate `rome_js_syntax`.

I would like to leave this refactor for later. The majority of this PR moves code around (and exposes some stuff to be public) and doesn't change logic of the existing code. I would like to apply these kind of changes later, so the PR is not too big.

leops (approved):

PR #2346, opened 2022-04-04
feat(rome_diagnostics): print diagnostics as markup

leops:
## Summary

This PR refactors how diagnostics are being printed to leverage `rome_console` markup as well as the `Codespan` and `Diff` formatting utilities. The changes include:

- Introducing `MarkupBuf`, an owned counterpart to `Markup` (this is now used in tests to verify diffs or codespans are emitted with the correct formatting)
- Creating a `Write` trait in `rome_console::fmt` to add a level of indirection between the formatter and the termcolor crate (allowing markup to be written into types that do not implement `WriteColor` like `MarkupBuf`)
- Modifying to how `Codespan` structs are created to better fit how data is stored inside diagnostics (in particular the text of codespan notes are now owned pieces of markup to facilitate the management of allocations)
- Reducing the `Console` trait to a single method `print(LogLevel, Markup)` and introducing a `ConsoleExt` extension trait implementing convenience methods (at the moment only `error` and `log`) on top of the generic printing
- Removing `codespan` from `rome_diagnostics` since that code has been moved to `rome_console`, retaining the `Emitter` API for backwards compability (but we could decide to deprecate / remove it)
- Removing `WriteConsole` and making `EnvConsole` a fully fledged struct, to allow it to lock the stdout / stderr stream while a message is being printed

## Test Plan

There's a large number of diagnostics captured in the snapshot tests for the parser and formatter, and only one of them has been modified after these changes (the file location wasn't using the position of the primary label but this is now fixed)


crates/rome_cli/src/commands/format.rs
```diff
@@ -195,7 +197,7 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
                 // Skip printing the diff for files over 1Mb (probably a minified file)
                 let max_len = old.len().max(new.len());
                 if max_len >= 1_000_000 {
```
xunilrj, line 199: When printing to the console, ok (maybe a force flag)... but when the stdout is redirected to a file, this may be useful.

leops: This is something I could have included in the PR description, but the other reason I added the `rome_console::fmt::Write` trait was to be able to add more methods to it in the future, for instance for checking if the stream being printed to is a TTY or querying the width of the screen. So this condition could make of these additional methods to chose exactly when to print the diff or not

crates/rome_console/src/codespan/render.rs
```diff
@@ -976,17 +950,11 @@ impl<'render, 'fmt> Renderer<'render, 'fmt> {
...
-                        elements: &[element],
-                        content: &ch,
-                    },
-                    MarkupNode {
-                        elements: &[element],
-                        content: &ch,
-                    },
-                ]))?;
+                self.writer.write_markup(markup! {
+                    {WithSeverity(label_style, severity, &markup!{
```
ematipico, line 954: This was weird at the beginning. Does that mean that nested markup need the `&` every time? If so, can we update the `README.md` with an example?

crates/rome_console/src/lib.rs
```diff
@@ -1,54 +1,51 @@
...
-            .emit_with_writer(diag, &mut self.err)
-            .unwrap();
+    fn log(&mut self, args: Markup) {
+        self.print(LogLevel::Log, args);
     }
 }
 
-/// Type alias of [WriteConsole] printing to the standard output
-pub type EnvConsole = WriteConsole<StandardStream, StandardStream>;
+/// Implementation of [Console] printing messages to the standard output
```
ematipico, line 44:
```suggestion
/// Implementation of [Console] printing messages to the standard output and standard error
```

crates/rome_console/src/lib.rs
```diff
@@ -1,54 +1,51 @@
...
-    O: WriteColor + Sync + RefUnwindSafe,
-    E: WriteColor + Sync + RefUnwindSafe,
-{
-    fn message(&mut self, args: Markup) {
-        fmt::Formatter::new(&mut self.out)
-            .write_markup(args)
-            .unwrap();
-        writeln!(self.out).unwrap();
+impl<T: Console + ?Sized> ConsoleExt for T {
+    fn error(&mut self, args: Markup) {
```
ematipico, line 35: Can we add some comment saying that error messages are streamed to standard error?

ematipico (commented):
In this PR we introduced a `LogLevel` feature, I have a couple of questions:
- are `LogLevel` related somehow?
- how do you plan to introduce the log level feature? Where exactly these logs will be filtered based on the log level passed (via CLI, configuration, etc.)?

crates/rome_console/src/codespan/render.rs
leops: This is specific to the `WithSeverity` utility struct since it takes an `&dyn Display` field in order to support any value that can be printed and not just markup, but that forces the value to be a reference. In general though nested markup doesn't need to be an explicit reference as the markup macro will insert all the required `&` tokens automatically

crates/rome_console/src/codespan/mod.rs
```diff
@@ -65,6 +68,21 @@ pub enum Locus {
     },
 }
 
+impl Display for Locus {
+    fn fmt(&self, fmt: &mut Formatter) -> io::Result<()> {
+        match self {
+            Locus::File { name } => write!(fmt, "{name}"),
+            Locus::FileLocation { name, location } => write!(
+                fmt,
+                "{name}:{line_number}:{column_number}",
```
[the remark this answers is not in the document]
leops: I didn't want to change the formatting of Diagnostics in this PR to make the change more readable (since it would imply changing all the error snapshots of the parser and formatter), but the intention is to make it more practical for us to change how things get printed later on for instance to bring the diagnostics in line with what they looked like in Rome JS

ematipico (approved):
That's a great improvement! Our diagnostics architecture are slowly getting a good shape! :)

PR #2349, opened 2022-04-05
chore(rome_cli): move ending message after diagnostics

ematipico:
## Summary

This PR moves the ending message for the `format` command after having the diagnostics printed.

Myself and Leo noticed that having the message "Formatted files ..." before the diagnostics wasn't helpful, especially if the number of diagnostics/warning was high.

This PR doesn't handle the "errors where emitted while formatting" message.

## Test Plan

Existing CI. I checked manually that the message shows up after printing diagnostics.

Screenshot 2022-04-05 at 08 39 12

### DIFF
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -116,15 +116,6 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
 
     let duration = start.elapsed();
     let count = formatted.load(Ordering::Relaxed);
-    if is_check {
-        session.app.console.message(rome_console::markup! {
-            <Info>"Checked "{count}" files in "{duration}</Info>
-        });
-    } else {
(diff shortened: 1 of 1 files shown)


leops (approved):
The final "errors were emitted" line is the exit status of the entire CLI so we can't really move it, it has to be printed last (in fact it's even being printed after the main function has returned). Maybe we could try to make it more clear that it is the overall "return code" of the program ?

ematipico:
> The final "errors were emitted" line is the exit status of the entire CLI so we can't really move it, it has to be printed last (in fact it's even being printed after the main function has returned). Maybe we could try to make it more clear that it is the overall "return code" of the program ?

Prettier CLI does a good thing, that we should document do to: https://prettier.io/docs/en/cli.html#exit-codes

It shows the error code returned and in their doc they explain the meaning of exist codes (if we intend to do so, I think we should add this explanation in the `--help` section of the CLI as well). They print the exit code too.

PR #2353, opened 2022-04-05
feat(rome_playground): add support for source type

### LINKED ISSUE #2338: 📎 Toggle SourceType in Rome Playground
### Description

For quickly testing out syntax errors.

Example from AST Explorer:

ematipico:
## Summary

Closes #2338 

Choosing "Script" also resets TypeScript and JSX, making them disabled.

Screenshot 2022-04-05 at 13 51 28

## Test Plan

Checked locally that works

Screenshot 2022-04-05 at 13 59 17

### DIFF
--- a/crates/rome_playground/src/App.tsx
+++ b/crates/rome_playground/src/App.tsx
@@ -8,8 +8,9 @@ import prettier from "prettier";
 import parserBabel from "prettier/esm/parser-babel";
 import IndentStyleSelect from "./IndentStyleSelect";
 import LineWidthInput from "./LineWidthInput";
-import { IndentStyle } from "./types";
+import {IndentStyle, SourceType} from "./types";
 import SourceTypeSelect from "./SourceTypeSelect";
+import {Source} from "postcss";
 
 enum LoadingState { Loading, Success, Error }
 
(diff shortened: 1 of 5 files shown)


crates/rome_playground/src/App.tsx
```diff
@@ -8,8 +8,9 @@ import prettier from "prettier";
 import parserBabel from "prettier/esm/parser-babel";
 import IndentStyleSelect from "./IndentStyleSelect";
 import LineWidthInput from "./LineWidthInput";
-import { IndentStyle } from "./types";
+import {IndentStyle, SourceType} from "./types";
 import SourceTypeSelect from "./SourceTypeSelect";
+import {Source} from "postcss";
```
leops, line 13: This looks like an autocomplete mistake ?

crates/rome_playground/src/lib.rs
```diff
@@ -78,15 +78,20 @@ pub fn run(
...
 
-    let source_type = match (is_typescript, is_jsx) {
-        (true, true) => SourceType::tsx(),
-        (true, false) => SourceType::ts(),
-        (false, true) => SourceType::jsx(),
-        (false, false) => SourceType::js_module(),
+    let source_type = if source_type == "script" {
+        SourceType::js_script()
+    } else {
+        match (is_typescript, is_jsx) {
```
leops, line 89: I think `is_typescript` could be used to call either `SourceType::js_module()` or `SourceType::ts()` directly, then the resulting source type could be further edited by calling `with_variant` based on the value of `is_jsx` and `with_module_kind` based on the value of `source_type`. Using a builder pattern would avoid having too many arms on this match expression if we want to add support for `.d.ts` syntax or changing the ECMAScript version in the playground

PR #2360, opened 2022-04-06
feat(rome_js_formatter): implement JsxText cleaning

### LINKED ISSUE #2359: 📎 format of JsxText should clean jsx text
### Description

Opportunity identified by this conversation here: https://discord.com/channels/678763474494423051/841927329265090590/959760493092933643

Formatting `JsxText` should change any two sequential ascii whitespace characters into a single space.

Examples of ascii whitespace characters:
```rust
'\n' // newline

'\t' // tab

' ' // space
```

Please assign this to @darrow-olykos , I have a solution and will open a PR

darrow-olykos:
closes #2359

## Summary

JsxText should be normalized / cleaned. This means that, for `JsxText` nodes, the formatter should replace any two or more ascii_whitespace characters (tab, space, or newline) in a row with a single space

#2359

## Test Plan

local unit tests:
- in `jsx/auxiliary/text.rs`, these just call `clean_jsx_text` function.

snapshot tests (after creating the text.jsx file in `tests/...`)
- `touch crates/rome_js_formatter/tests/spec_tests.rs && cargo test -p rome_js_formatter formatter`

I used `cargo insta review` to review my newly generated snapshot test, and hit `a` to accept it.


crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,7 +1,57 @@
...
+    use crate::jsx::auxiliary::text::clean_jsx_text;
+
+    #[test]
+    fn clean_jsx_text_works() {
+        assert_eq!("Foo", clean_jsx_text("Foo"));
+        assert_eq!(" Foo", clean_jsx_text(" Foo"));
+        assert_eq!(" Foo", clean_jsx_text("\nFoo"));
+        assert_eq!(" Foo", clean_jsx_text("\tFoo"));
+        assert_eq!(" Foo", clean_jsx_text("\n \t Foo"));
+        assert_eq!(" Foo", clean_jsx_text("\n \t \n \t\nFoo"));
```
ematipico, line 45: Could we add a test like `"     Foo             bar               lorem"`? Spaces between words should be reduced to just one.

ematipico: Can we add a test against an empty string and a string with a space?

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,7 +1,57 @@
...
+        let new_text = clean_jsx_text(token.text());
+
+        Ok(formatter.format_replaced(
+            &token,
+            Token::new_dynamic(new_text, token.text_trimmed_range()).into(),
+        ))
+    }
+}
+
+fn clean_jsx_text(text: &str) -> String {
```
ematipico, line 17: Few suggestions. It seems we always manipulate the string regardless, but sometimes we don't need to do it.

- We could return a `Cow<str>`. A similar strategy is used here: https://github.com/rome/tools/blob/main/crates/rome_formatter/src/format_element.rs#L1199-L1223
- If the text has `len` 0, we should just bail and avoid any operation and return `Cow::borrowed`
- If the text doesn't have any `ascii_whitespace`, we can avoid any sort of transformation and return `Cow::borrowed`
- If the initial string and the new string are the same, we could just return `Cow::borrowed` (the original string)
- If the new string differs from the original, we can return `Cow::Owned`

ematipico (changes requested):
Thank you! The tests look good and it seems the implementation does what we need. I added  few comments where we can improve the logic of the algorithm.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,7 +1,57 @@
...
+            words.insert(0, ' ');
+        }
+    }
+    if let Some(c) = chars.last() {
+        if c.is_ascii_whitespace() {
+            words.push(' ');
+        }
+    }
+    words
+}
```
xunilrj, line 32: I would try to avoid:
- text.split_ascii_whitespace().collect(): this is iterating the whole string and allocating a vec;
- words_vec.join(" "): whis is iterating the whole vec and allocating a new String;
- words.insert(0, ' '): This is a O(n) operation.

darrow-olykos:
Great feedback, making these changes, thank you!

darrow-olykos:
I made some changes. I decided to go with a more pure `Iterator` solution, because I don't think it is safe to do string manipulation in rust using byte indexes and slices.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,9 +1,74 @@
...
+
+fn clean_jsx_text(text: &str) -> Cow<str> {
+    if text.len() == 0 {
+        Cow::Borrowed(text)
+    } else {
+        let terminators = [' ', '\n', '\t'];
+        let mut result = String::new();
+        if text.starts_with(terminators) {
+            result.push(' ');
+        }
```
ematipico, line 29: I think this is an edge case but, I think if a string starts only with a newline, e.g. `"\nFoo"`, we should not add a white space.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,9 +1,74 @@
...
+            Token::new_dynamic(new_text.to_string(), token.text_trimmed_range()).into(),
+        ))
+    }
+}
+
+fn clean_jsx_text(text: &str) -> Cow<str> {
+    if text.len() == 0 {
+        Cow::Borrowed(text)
+    } else {
+        let terminators = [' ', '\n', '\t'];
```
ematipico, line 25: We will need to take in consideration `\r\n`. If you want we can tackle this case in a separate PR, as you prefer

ematipico (approved):
Overall, as a start, it looks good to me. There are some edge case that we should tackle as well but we can address them in a separate PR if you want.

We should address linting errors and test errors

xunilrj:
@darrow-olykos I did a quick test where we use ```Cow``` and do only one pass on the string. What do you think?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a40270c609fb73e71973165bad7907e8

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
MichaReiser: I think the classic implementation even worked differently in that it trimmed all whitespace except if it is the first line.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
MichaReiser: ...and `\r`.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,9 +1,74 @@
...
+        let mut result = String::new();
```
MichaReiser, line 26: I wonder if it would be possible to avoid the allocation of a new string for the case that the two strings are identical by:

* Implementing the cleaning as an Iterator that returns one character at a time.
* Iterating over both iterators at the same time and compare character by character
  * If there's no difference (should be case for the majority), return the original `&str` slice
  * If there's a difference, allocate a new string and concat the characters from the original slice up (but not including) the current character, with the current character, and the remainder of the "cleaning" iterator.

xunilrj: > I wonder if it would be possible to avoid the allocation of a new string for the case that the two strings are identical...

I think my suggestion achieves this.

My initial assumption is: 99.99% of the times we are already formatted. Probably because the formatter already did its job. So we we first check if everything is fine. If it is, return ```Cow::Borrowed```.

MichaReiser: Sorry, I didn't see your suggestion. Yeah, I think this is the same with the difference whatever to use an iterator for cleaning or not.

darrow-olykos:
Hey all, sorry but I need to step away from this PR to focus on some interviews (non-Rust), I am not sure if I'll be freed-up any time soon. Feel free to steal the last leg of this if you'd like.

darrow-olykos:
Will be taking another look this Monday night

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
karl: Yeah, replacing new lines that are next to tags with white spaces changes the meaning of the code so isn’t a safe transformation here.

If you are putting all the text on a single line you can remove the new line:
https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAPAvAAgBQB4CGAfADpSbnkz4BGFFAzgA75hz2nlMtykdRwB3ADYBLfuzJ0q1XAHoiAShAAaEBEYwR0eslD4ATvogCACgYQ6U+IQPwBPHaur6WAazgwAyvgC2cADJicMgAZtb0cE4uYO5ezGBiAObIMPoArpEgcD7UcAAmefn++FCJafiJcABiEPo++DCapcgg+GkwECogABYwPkIA6t0i8FysnhYjIgBuI3YtYPSOIGIR+jAmLon1oeGZAFb0qJ5JQnAAimkQ8LtCEarM+mst1DRwQl2M+mIwAyJ5MG6yAAHAAGB5GCIDFyMFpfNhwfTTYKqACOV3gm3Ulla9AAtPx8vkuvo4OiRKTNhUdkgwndMhEfCIUukGacLhjgrS9qppH8AUCkAAmXkuESiUoAYQgPhpWXoAFYumkIgAVGiWOn3EDTDIASSghVgnjA3w0AEFDZ4YHYzrcIgBfB1AA

Or if you are splitting the text onto multiple lines then you can keep it:
https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAPAvAAgBQB4CGAfADpSbnkz4BGFFAzgA75hz2nlMtykdRwB3ADYBLfuzJ0qtIRBj1MEAGaZ4qGJnoRVAC3waACgCc4MGCLhHMAkUKGZ+cACartAo-kaYRG-FBcwOpZwShAmmADWcHBegXDeUD4i+Pb8wmJwAHS4APREAJQgADQgEIzm0PTIoPhGRhACBrUIVSgpAvgAnlUl1B5gUTAAyvgAtnAAMhnISin0cL39g0PMYGIA5sgwRgCuCyBwo9TOTs4Tfus7+OtwAGJho-rmUJttOzAQxSA6MKNCAOo6HxsVZwIYtJIANx8nWQ4HoPRAYnmRhgxmujxmc32ACt6KghhshHAAIo7ORwLFCeYlZhGFFw6g0OBCL6MIxiGD-EROQLIAAcAAZafV5v8PIw4ey2JZIZSSgBHcnwYxlVogfD0AC0jlOTi+JiVIhM6PWmKQs2p+3moxEW121qJpOVlIt2JK0m5vJ0yAATB6PLYNgBhCCjc0HegAVi+O3mABUaK1LTSQJC9gBJfwIYZgDnlACC-iGME6xKp8wAvpWgA

darrow-olykos: @karl and @ematipico thanks a bunch! I didn't realize HTML and JSX specs treat that leading whitespace differently.
Just curious, would either of you know how one might derive that detail from JSX spec if I didn't have the prettier formatter and you helpful people as a reference?

Does it come down to the definition of a `JsxStringCharacter`, `HTMLCharacterReference`, and how an "implementation" of JSX may convert a `JSXString` into an `ECMAScript String value` https://facebook.github.io/jsx/#sec-jsx-JSXString-SV ?

darrow-olykos: I'll move forward w/ this PR, I'm also just building my mental-maps-of-the-world as I go.

darrow-olykos:
@xunilrj @MichaReiser 

regarding @karl's statements
> If you are putting all the text on a single line you can remove the new line
and
> Or if you are splitting the text onto multiple lines then you can keep it

Does have methods available to check this from within `JSXText`'s `ToFormatElement` implementation?
````rust
// impl ToFormatElement for JsxText ...

if pos == 0 && !fit_on_single_line() {
    pos += current.len_utf8(); // keep the first char
}
// else dont increment pos, don't keep it
```

darrow-olykos:
> @darrow-olykos I did a quick test where we use `Cow` and do only one pass on the string. What do you think? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a40270c609fb73e71973165bad7907e8

@xunilrj Nice! I see that you use one loop to iterate characters as far as possible until either the end of (valid) text, or until we detect first issue (two whitespaces in a row), and then you allow the fn to return a `Cow::Borrowed` in case we actually already have valid text, and otherwise we allocate new string and push the valid section of the text so far, and then you use second loop to pick up where first loop left off, and start pushing only valid chars into the new string we've allocated.

As for whether it works, I think I'll need to determine the answer to my Q here: https://github.com/rome/tools/pull/2360#issuecomment-1101920449

I notice you have a comment that says _"needs fixing"_, however I don't quite see what needs fixing. Is it related to my Q linked above?

MichaReiser:
> @xunilrj @MichaReiser
> 
> regarding @karl's statements:
> 
> > If you are putting all the text on a single line you can remove the new line
> 
> and
> 
> > Or if you are splitting the text onto multiple lines then you can keep it
> 
> Does Rome formatter have methods available to check this from within `JSXText`'s `ToFormatElement` implementation?
> 
> ```rust
> // impl ToFormatElement for JsxText ...
> 
> if pos == 0 && !fit_on_single_line() {
>     pos += current.len_utf8(); // keep the first char
> }
> // else dont increment pos, don't keep it
> ```
> 
> Or, should I just always remove whitespace in position 0 and expect to rely on a parent node using a helper method like `group_elements`?
> 
> group_element docs:
> 
> ```rust
> /// Creates a logical `Group` around the content that should either consistently be printed on a single line
> /// or broken across multiple lines.
> ///
> /// The printer will try to print the content of the `Group` on a single line, ignoring all soft line breaks and
> /// emitting spaces for soft line breaks or spaces. The printer tracks back if it isn't successful either
> /// because it encountered a hard line break, or because printing the `Group` on a single line exceeds
> /// the configured line width, and thus it must print all its content on multiple lines,
> /// emitting line breaks for all line break kinds.
> ///
> ```

Rome currently exposes now way of knowing the print width when formatting. This is only done when printing. However, `fill_elements` might be useful here: It tries to fit all elements on a sine (joined by spaces) but breaks as soon as the first element exceeds the line width. 

You may also want to take a look at how Prettier implements this https://github.com/prettier/prettier/blob/d13feed42b6478710bebbcd3225ab6f203a914c1/src/language-js/print/jsx.js#L62-L257 (rather involved)

ematipico:
@darrow-olykos are there any updates on the progress of this PR?

If you think you don't have enough time, please don't worry to say so. We can over a polish what you're done!

darrow-olykos:
@ematipico Looking for input on this thread in the Rome discord: https://discord.com/channels/678763474494423051/969333047419887657/969360029109092444

This will help me wrap up this PR!

darrow-olykos:
Here is a functional implementation of something that is hopefully more performant than using some of the built-in trimming functions (would need to benchmark):

Note, this does not appropriately handle `/r` nor `/r/n`

```rust
use std::borrow::Cow;

fn clean_jsx_text(text: &str) -> Cow<str> {
    if text.is_empty() {
        // if text is empty, we can return a Cow::Borrowed to reuse existing allocation
        return std::borrow::Cow::Borrowed(text);
    }

    // use of char_indices is significant here, because it tracks byte position (not char position),
    //   which may skip a couple integers in cases where the char is greater than 1 byte
    let mut chars = text.char_indices();

    // this while loop behaves like `start_pos = text.find(is_not_whitespace);`
    //   with the added benefit of saving our place in the chars iterator
    let mut start_pos: Option<usize> = None;
    while let Some((pos, c)) = chars.next() {
        if !c.is_whitespace() {
            start_pos = Some(pos);
            break;
        }
    }
    match start_pos {
        // if non-whitespace is not found, then whole text is whitespace
        None => return std::borrow::Cow::Borrowed(""),
        Some(start_pos) => {
            // if non-whitespace found, then compliant text starts at that byte position
            let mut compliant_end_pos = start_pos;
            let mut prev_char_is_whitespace = false;
            while let Some((pos, c)) = chars.next() {
                // track how much of beginning of text is already compliant
                let current_char_is_whitespace = c.is_whitespace();
                if prev_char_is_whitespace && current_char_is_whitespace {
                    // if two whitespaces are found in a row in the middle of the text, then we encountered non-compliant whitespace
                    break;
                } else {
                    // if one or less whitespace in a row in middle of the text, then this is compliant, so save byte position and continue
                    compliant_end_pos = pos;
                    if !current_char_is_whitespace {
                        compliant_end_pos += c.len_utf8();
                    }
                    prev_char_is_whitespace = current_char_is_whitespace;
                }
            }

            if start_pos == 0 && !prev_char_is_whitespace && chars.next() == None {
                // if text has no whitespace in the beginning and if we've consumed compliant chars until the end of the string,
                // then the text is already compliant and we can return without allocating
                return std::borrow::Cow::Borrowed(text);
            }

            // allocate new string with however much text has been determined to be compliant so far
            let mut new_text = String::with_capacity(text.len());
            new_text.push_str(&text[start_pos..compliant_end_pos]);

            let mut end_char: Option<char> = None;
            while let Some((_pos, c)) = chars.next_back() {
                if !c.is_whitespace() {
                    end_char = Some(c);
                    break;
                }
            }

            prev_char_is_whitespace = false;
            while let Some((_pos, c)) = chars.next() {
                // iterate remaining characters (if any remain) and push compliant characters one by one
                let current_char_is_whitespace = c.is_whitespace();
                if prev_char_is_whitespace && current_char_is_whitespace {
                    // skip non-compliant chars
                    continue;
                }

                if current_char_is_whitespace {
                    new_text.push(' ');
                } else {
                    new_text.push(c);
                }

                prev_char_is_whitespace = current_char_is_whitespace;
            }

            if let Some(end_char) = end_char {
                new_text.push(end_char);
            }
            
            std::borrow::Cow::Owned(new_text)
        }
    }
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn clean_jsx_text_works() {
        assert_eq!("", clean_jsx_text(""));
        assert_eq!("", clean_jsx_text(" "));
        assert_eq!("", clean_jsx_text("  "));
        assert_eq!("Foo", clean_jsx_text("Foo"));
        assert_eq!("Foo", clean_jsx_text(" Foo"));
        assert_eq!("Foo", clean_jsx_text("\nFoo"));
        assert_eq!("Foo", clean_jsx_text("\tFoo"));
        assert_eq!("Foo", clean_jsx_text("\n \t Foo"));
        assert_eq!("Foo", clean_jsx_text("\n \t \n \t\nFoo"));
        assert_eq!("Foo bar lorem", clean_jsx_text(" Foo bar lorem"));
        assert_eq!("Foo", clean_jsx_text("Foo "));
        assert_eq!("Foo", clean_jsx_text("Foo\n"));
        assert_eq!("Foo", clean_jsx_text("Foo\t"));
        assert_eq!("Foo", clean_jsx_text("Foo\t \n "));
        assert_eq!("Foo", clean_jsx_text("Foo\n \t \n \t\n"));
        assert_eq!("Foo Bar", clean_jsx_text("Foo\n \t\t\n \tBar"));
        assert_eq!(
            "Foo Bar",
            clean_jsx_text("\n \t\t\n \tFoo\n \t\t\n \tBar\n \t\t\n \t")
        );
        assert_eq!(
            "💣 Foo 💣 Bar 💣",
            clean_jsx_text("\n \t💣\t\n \tFoo\n \t💣\t\n \tBar\n \t💣\t\n \t")
        );
        // assert_eq!("Foo", clean_jsx_text("\rFoo"));
        // assert_eq!("Foo", clean_jsx_text("Foo\r"));
        // assert_eq!("Foo Bar", clean_jsx_text("Foo\rBar"));
        // assert_eq!("Foo Bar", clean_jsx_text("\rFoo\rBar\r"));
        // assert_eq!("Foo", clean_jsx_text("\r\nFoo"));
        // assert_eq!("Foo", clean_jsx_text("Foo\r\n"));
        // assert_eq!("Foo Bar", clean_jsx_text("Foo\r\nBar"));
        // assert_eq!("Foo Bar", clean_jsx_text("\r\nFoo\r\nBar\r\n"));
    }
}

```

darrow-olykos:
Slightly cleaner:

```rust
use std::{borrow::Cow, str::CharIndices};

type CharIndex = (usize, char);

fn clean_jsx_text(text: &str) -> Cow<str> {
    if text.is_empty() {
        // if text is empty, we can return a Cow::Borrowed to reuse existing allocation
        return std::borrow::Cow::Borrowed(text);
    }

    let mut chars = text.char_indices();
    match get_first_compliant_char_index(&mut chars) {
        None => return std::borrow::Cow::Borrowed(""),
        Some((start_pos, _)) => {
            match get_last_compliant_char_index(&mut chars) {
                None => panic!("Something is wrong, if there is a first_compliant char, then there should be a last_compliant char"),
                Some((compliant_end_pos, last_compliant_char)) => {
                    if start_pos == 0 && !last_compliant_char.is_whitespace() && chars.next() == None { // this seems sus.
                        // if text has no whitespace in the beginning and if we've consumed compliant chars until the end of the string,
                        // then the text is already compliant and we can return without allocating
                        return std::borrow::Cow::Borrowed(text);
                    }
                    
                    // allocate new string with however much text has been determined to be compliant so far from the left
                    let mut new_text = String::with_capacity(text.len());
                    new_text.push_str(&text[start_pos..compliant_end_pos]);

                    // consume all whitespace from the right-side of text, may consume right most compliant char
                    let end_char_tuple = chars.rfind(|(_pos, c)| !c.is_whitespace());

                    build_compliant_text(&mut chars, &mut new_text);

                    // add right most compliant char to end if consumed from rfind
                    if let Some((_pos, c)) = end_char_tuple {
                        new_text.push(c);
                    }

                    std::borrow::Cow::Owned(new_text)
                }
            }
        }
    }
}

fn get_first_compliant_char_index(chars: &mut CharIndices) -> Option<CharIndex> {
    chars.find(|(_pos, c)| !c.is_whitespace())
}

fn get_last_compliant_char_index(chars: &mut CharIndices) -> Option<CharIndex> {
    let mut is_prev_char_whitespace = false;
    let mut last_compliant_char_index: Option<CharIndex> = None;
    while let Some((pos, c)) = chars.next() {
        let cur_char_is_whitespace = c.is_whitespace();
        if is_prev_char_whitespace && cur_char_is_whitespace {
            break;
        }
        else {
            if !cur_char_is_whitespace {
                last_compliant_char_index = Some((pos + c.len_utf8(), c));
            } else {
                last_compliant_char_index = Some((pos, c));
            }
            is_prev_char_whitespace = cur_char_is_whitespace;
        }
    }
    return last_compliant_char_index;
}

fn build_compliant_text<'a>(chars: &mut CharIndices, new_text: &'a mut String) {
    let mut prev_char_is_whitespace = false;
    while let Some((_pos, c)) = chars.next() {
        // iterate remaining characters (if any remain) and push compliant characters one by one
        let current_char_is_whitespace = c.is_whitespace();
        if prev_char_is_whitespace && current_char_is_whitespace {
            // skip non-compliant chars
            continue;
        }

        if current_char_is_whitespace {
            new_text.push(' ');
        } else {
            new_text.push(c);
        }

        prev_char_is_whitespace = current_char_is_whitespace;
    }
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn clean_jsx_text_works() {
        assert_eq!("", clean_jsx_text(""));
        assert_eq!("", clean_jsx_text(" "));
        assert_eq!("", clean_jsx_text("  "));
        assert_eq!("Foo", clean_jsx_text("Foo"));
        assert_eq!("Foo", clean_jsx_text(" Foo"));
        assert_eq!("Foo", clean_jsx_text("\nFoo"));
        assert_eq!("Foo", clean_jsx_text("\tFoo"));
        assert_eq!("Foo", clean_jsx_text("\n \t Foo"));
        assert_eq!("Foo", clean_jsx_text("\n \t \n \t\nFoo"));
        assert_eq!("Foo bar lorem", clean_jsx_text(" Foo bar lorem"));
        assert_eq!("Foo", clean_jsx_text("Foo "));
        assert_eq!("Foo", clean_jsx_text("Foo\n"));
        assert_eq!("Foo", clean_jsx_text("Foo\t"));
        assert_eq!("Foo", clean_jsx_text("Foo\t \n "));
        assert_eq!("Foo", clean_jsx_text("Foo\n \t \n \t\n"));
        assert_eq!("Foo Bar", clean_jsx_text("Foo\n \t\t\n \tBar"));
        assert_eq!(
            "Foo Bar",
            clean_jsx_text("\n \t\t\n \tFoo\n \t\t\n \tBar\n \t\t\n \t")
        );
        assert_eq!(
            "💣 Foo 💣 Bar 💣",
            clean_jsx_text("\n \t💣\t\n \tFoo\n \t💣\t\n \tBar\n \t💣\t\n \t")
        );
        assert_eq!("Foo", clean_jsx_text("\rFoo"));
        assert_eq!("Foo", clean_jsx_text("Foo\r"));
        // assert_eq!("Foo Bar", clean_jsx_text("Foo\rBar"));
        // assert_eq!("Foo Bar", clean_jsx_text("\rFoo\rBar\r"));
        assert_eq!("Foo", clean_jsx_text("\r\nFoo"));
        assert_eq!("Foo", clean_jsx_text("Foo\r\n"));
        //assert_eq!("Foo Bar", clean_jsx_text("Foo\r\nBar"));
        // assert_eq!("Foo Bar", clean_jsx_text("\r\nFoo\r\nBar\r\n"));*/
    }
}
```

PR #2361, opened 2022-04-06
feat(rome_cli): CLI Improvements

leops:
## Summary

I batched together a few QoL improvements to the CLI:
- Show the help for a subcommand when it's called with no other arguments (`rome format`), but still exit with an error
- Expose the line width formatting option through the `--line-width` CLI arguments
- Add a `--dry-run` mode to the CLI to print the diff to the console instead of writing the formatting result to the file (similar to how CI mode works, except the diff isn't reported as an error)
- Reworked how messages are printed to the console: they are now being handled by a dedicated thread in parallel to the formatting process instead of being buffered and flushed once the traversal is complete
- Add a `LineWidth` wrapper type for the `line_width` formatter setting to validate the value is within an acceptable range (`1..=320`)

## Test Plan

I've added a few additional tests for the CLI verifying the new arguments are properly handled and validated

### DIFF
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -1,12 +1,13 @@
 use std::{
     collections::HashMap,
+    ffi::OsString,
     fmt::Display,
     io,
     ops::Range,
     panic::catch_unwind,
     path::{Path, PathBuf},
     sync::atomic::{AtomicUsize, Ordering},
-    time::Instant,
+    time::{Duration, Instant},
 };
 
 use crossbeam::channel::{unbounded, Receiver, Sender};
@@ -70,15 +71,15 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
         options.line_width = line_width;
     }
 
-    let is_check = session.args.contains("--ci");
-    let is_dry_run = session.args.contains("--dry-run");
+    let is_write = session.args.contains("--write");
+    let is_ci = session.args.contains("--ci");
     let ignore_errors = session.args.contains("--skip-errors");
 
-    let mode = match (is_check, is_dry_run) {
-        (true, true) => return Err(Termination::IncompatibleArguments("--ci", "--dry-run")),
-        (true, false) => FormatMode::Check,
-        (false, true) => FormatMode::Print,
-        (false, false) => FormatMode::Write,
+    let mode = match (is_write, is_ci) {
+        (true, true) => return Err(Termination::IncompatibleArguments("--write", "--ci")),
+        (true, false) => FormatMode::Write,
+        (false, true) => FormatMode::Check,
+        (false, false) => FormatMode::Print,
     };
 
     // Check that at least one input file / directory was specified in the command line
@@ -115,34 +116,24 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
     let console = &mut *session.app.console;
 
     let (has_errors, duration) = join(
-        || console_thread(mode, console, recv_files, recv_msgs),
+        || print_messages_to_console(mode, console, recv_files, recv_msgs),
         || {
-            let start = Instant::now();
-
             // The traversal context is scoped to ensure all the channels it
             // contains are properly closed once the traversal finishes
-            let ctx = FormatCommandOptions {
+            traverse_inputs(
                 fs,
-                features,
-                options,
-                mode,
-                ignore_errors,
-                interner,
-                formatted: &formatted,
-                messages: send_msgs,
-            };
-
-            let ctx = &ctx;
-            session
-                .app
-                .fs
-                .traversal(Box::new(move |scope: &dyn TraversalScope| {
-                    for input in inputs {
-                        scope.spawn(ctx, PathBuf::from(input));
-                    }
-                }));
-
-            start.elapsed()
+                inputs,
+                &FormatCommandOptions {
+                    fs,
+                    features,
+                    options,
+                    mode,
+                    ignore_errors,
+                    interner,
+                    formatted: &formatted,
+                    messages: send_msgs,
+                },
+            )
         },
     );
 
@@ -174,9 +165,27 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
     }
 }
 
+/// Initiate the filesystem traversal tasks with the provided input paths and
+/// run it to completion, returning the duration of the process
+fn traverse_inputs(
+    fs: &dyn FileSystem,
+    inputs: Vec<OsString>,
+    ctx: &FormatCommandOptions,
+) -> Duration {
+    let start = Instant::now();
+
+    fs.traversal(Box::new(move |scope: &dyn TraversalScope| {
+        for input in inputs {
+            scope.spawn(ctx, PathBuf::from(input));
+        }
+    }));
+
+    start.elapsed()
+}
+
 /// This thread receives [Message]s from the workers through the `recv_msgs`
 /// and `recv_files` channels and prints them to the console
-fn console_thread(
+fn print_messages_to_console(
     mode: FormatMode,
     console: &mut dyn Console,
     recv_files: Receiver<(usize, PathBuf)>,
@@ -321,7 +330,7 @@ impl Files for PathFiles {
 enum FormatMode {
     /// Write the result to the original file
     ///
-    /// This is the default behavior when no CLI argument is specified
+    /// This is the behavior of the CLI when the `--write` argument is specified
     Write,
     /// Compare the result to the original content of the file and return an
     /// error if they differ
@@ -330,7 +339,7 @@ enum FormatMode {
     Check,
     /// Print a diff of the formatting result with original content of the file
     ///
-    /// This is the behavior of the CLI when the `--dry-run` argument is specified
+    /// This is the default behavior when no CLI argument is specified
     Print,
 }
 
@@ -372,7 +381,7 @@ impl<'ctx, 'app> TraversalContext for FormatCommandOptions<'ctx, 'app> {
     }
 
     fn push_diagnostic(&self, file_id: FileId, code: &'static str, message: String) {
-        self.push_message(Error {
+        self.push_message(FormatError {
             severity: Severity::Error,
             file_id,
             code,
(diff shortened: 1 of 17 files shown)


crates/rome_cli/src/termination.rs
```diff
@@ -38,6 +38,14 @@ pub enum Termination {
     )]
     MissingArgument { argument: &'static str },
 
+    /// Returned when a subcommand is called without any arguments
+    #[error("empty arguments")]
+    EmptyArguments,
```
ematipico, line 43: Why do we call it "subcommand"? Isn't `rome format` a command? (`format` in this case)

crates/rome_cli/tests/main.rs
```diff
@@ -111,6 +111,27 @@ fn test_indent_size_parse_errors() {
     }
 }
 
+#[test]
+fn test_line_width_parse_errors() {
```
ematipico, line 115: Can we add a test against a high number of `line-width`?

crates/rome_cli/tests/main.rs
```diff
@@ -111,6 +111,27 @@ fn test_indent_size_parse_errors() {
...
+            OsString::from("--line-width"),
+            OsString::from("-1"),
+            OsString::from("file.js"),
+        ]),
+    });
+
+    match result {
+        Err(Termination::ParseError { argument, .. }) => assert_eq!(argument, "--line-width"),
+        _ => panic!("run_cli returned {result:?} for an invalid argument value, expected an error"),
+    }
```
ematipico, line 132: The assertion is a bit obscure, because I don't know what is the value of`{result:?}`.  Wouldn't it be better to hardcode all the string here?

crates/rome_cli/tests/main.rs
```diff
@@ -176,3 +197,68 @@ fn test_formatting_error() {
         _ => panic!("run_cli returned {result:?} for a failed CI check, expected an error"),
     }
 }
+
+#[test]
+fn test_empty_arguments() {
```
ematipico, line 202: Same for all these tests. `{result:?}` should be replaced with an hardcoded value, so we know:
- what's the exact result given a certain input
- if labels change, we can catch the regression (and see if there are typos, for example)

crates/rome_cli/src/commands/help.rs
```diff
@@ -22,11 +22,12 @@ OPTIONS:
     --skip-errors               Skip over files containing syntax errors instead of returning an error
     --indent-style <tabs|space> Determine whether the formatter should use tabs or spaces for indentation (default: tabs)
     --indent-size <number>      If the indentation style is set to spaces, determine how many spaces should be used for indentation (default: 2)
+    --line-width <number>       Determine how many characters the formatter is allowed to print in a single line (default: 80)
```
ematipico, line 25: `--dry-run` is missing from the help description

crates/rome_cli/src/commands/format.rs
```diff
@@ -56,9 +58,29 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
+        .args
+        .opt_value_from_str("--line-width")
+        .map_err(|source| Termination::ParseError {
+            argument: "--line-width",
+            source,
+        })?;
+
+    if let Some(line_width) = line_width {
+        options.line_width = line_width;
+    }
```
ematipico, line 71: Somewhere (formatter, or CLI) we should have some boundaries around the actual value of `--line-width` (for example, the maximum value in VSCode is 320). Do we have this check (and test) somewhere?

crates/rome_cli/src/commands/format.rs
```diff
@@ -84,113 +106,166 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
+                .app
+                .fs
+                .traversal(Box::new(move |scope: &dyn TraversalScope| {
+                    for input in inputs {
+                        scope.spawn(ctx, PathBuf::from(input));
+                    }
+                }));
+
+            start.elapsed()
+        },
```
ematipico, line 146: At this point, wouldn't it make more sense to move this logic into a function?

crates/rome_cli/src/commands/format.rs
```diff
@@ -84,113 +106,166 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
-                has_errors |= diag.is_error();
-                file_ids.insert(diag.file_id);
-            }
-            Error::Diff { .. } => {
-                has_errors = true;
-            }
+    match mode {
+        FormatMode::Write => {
+            console.log(rome_console::markup! {
+                <Info>"Formatted "{count}" files in "{duration}</Info>
```
ematipico, line 154: One thing I noticed is that this message might be misleading. While I was testing the CLI on a huge code base (ours), the message said that only **few** files were formatted but the **elapsed time** was **big** compared to the number of files actually formatted (this because the program inspected loads of files). This could lead that our binary was slow overall in formatting only few files. I think we should add the **total number of files inspected** against the ones we actually formatted.

crates/rome_cli/src/commands/format.rs
```diff
@@ -84,113 +106,166 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
+            });
         }
+        FormatMode::Check => {
+            console.log(rome_console::markup! {
+                <Info>"Checked "{count}" files in "{duration}</Info>
```
ematipico, line 159: Can't we save `"{count}" files in "{duration}` into a variable?

crates/rome_cli/src/commands/format.rs
```diff
@@ -84,113 +106,166 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
+}
 
-    let mut files = PathFiles::default();
-    while let Ok((file_id, path)) = recv_files.recv() {
-        if !file_ids.contains(&file_id) {
-            continue;
-        }
+/// This thread receives [Message]s from the workers through the `recv_msgs`
+/// and `recv_files` channels and prints them to the console
+fn console_thread(
```
ematipico, line 179: The name of this function seems waaay too generic, while it's actually really specialized. Can't we call it something like `print_format_messages_to_console`?

crates/rome_cli/src/commands/format.rs
```diff
@@ -257,22 +316,42 @@ impl Files for PathFiles {
     }
 }
 
+/// Determines how the result of formatting should be handled
+#[derive(Clone, Copy)]
+enum FormatMode {
+    /// Write the result to the original file
+    ///
+    /// This is the default behavior when no CLI argument is specified
+    Write,
```
ematipico, line 325: I don't think we should make it the default behaviour. Writing a file is a disruptive operation and irreversible in cases the file is not under version control. Rome classic didn't use write as default option, and I think we should do the same here.

Prettier CLI is similar, by default it prints and with `--write` argument it actually write the files. 

This would be a breaking change, yes, but we are in `0.` version, so I don't mind breaking stuff

crates/rome_cli/src/commands/format.rs
```diff
@@ -398,46 +497,62 @@ fn format_file(params: FormatFileParams) -> Result<Vec<Diagnostic>, Error> {
...
-        Self::Diagnostic(diagnostic)
+impl From<Error> for Message {
+    fn from(err: Error) -> Self {
+        Self::Error(err)
     }
 }
 
-/// Extension trait for turning [Display]-able error types into [Diagnostic]
+/// Generic error type returned by the formatting process
+struct Error {
```
ematipico, line 541: `FormatError`?

crates/rome_cli/src/termination.rs
leops: The name "subcommand" comes from `pico_args`: since we're parsing the entire command line the actual command is `rome`, and `format` is a subcommand of that

crates/rome_cli/tests/main.rs
leops: I'm not exactly sure about what you're suggesting here, when this branch is executed it means `result` does not have the expected value so whatever it actually contains is printed in debug mode and the test fails. Being exhaustive here would mean covering every possible error case (and updating every single test when we add new errors) when that's really not supposed to happen anyway, so it seem a bit impractical.

crates/rome_cli/src/commands/format.rs
leops: Since `line_width` is a `u16` it has a hard limit of `65535`, obviously that way too high (and setting it to a very large value would might cause terrible performance in the printer as it would have to backtrack across very large ranges of elements) so I think the formatter could have a `LineWidth(u16)` newtype struct implementing `TryFrom<u16>` and performing this validation

crates/rome_cli/src/commands/format.rs
leops: It could be stored into a second shared piece of markup, but for such a short piece of text I decided to prioritize readability by having the whole message spelled out for each case

crates/rome_cli/src/commands/format.rs
ematipico: Now that we are live, documentation changes should be deployed right before/after the release of a new version. Before we were using `latest` as branch to deploy to production, but I don't think it's a good compromise. For the time being.

I can suggest to create an umbrella issue where we collect the documentation changes we have to write before releasing.

crates/rome_cli/tests/main.rs
ematipico: Oooh I see now! For some reason I thought the test was about that message 😅

crates/rome_cli/src/commands/format.rs
leops: The `format_file` function counts all the files it formats, whether the result is actually written to disk or not, so I think the discrepancy here comes from files with syntax errors being skipped (then the time is quite large because we need to emit diagnostics for all these errors). So what we could do is keep track of the number of files that have errors, and print this counter as a secondary information if it's non-zero

crates/rome_formatter/src/lib.rs
```diff
@@ -55,13 +56,79 @@ impl Display for IndentStyle {
...
+
+/// Error type returned when converting a u16 to a [LineWidth] fails
+#[derive(Clone, Copy, Debug)]
+pub struct LineWidthFromIntError(pub u16);
+
+impl TryFrom<u16> for LineWidth {
+    type Error = LineWidthFromIntError;
+
+    fn try_from(value: u16) -> Result<Self, Self::Error> {
+        if value > 0 && value <= 320 {
```
ematipico, line 111: It would be great if we could have this `320` in a `const` and exported. I would be useful in the LSP

ematipico (approved):

Issue #2350, opened 2022-04-05
☂️ CSS support


ematipico:
### Description

We are going to use this issue to track the various work around CSS. We don't plan, for the time being, to support super languages, although we might think about preparing the infrastructure to be able to support them.

- [ ] create a `css.ungram` file where we will slowly implement new grammars
- [ ] create a `rome_css_syntax` crate, where we will have all things related to the CSS syntax, generated code too
- [ ] expand the codegen of the grammar to include CSS too
- [ ] evaluate how to implement the parser, lexer, etc. (it will require a separate discussion)
- [ ] create a `rome_css_formatter` crate
- [ ] make the `rome_formatter` generic
- [ ] prepare the Rome architecture to include CSS (file handlers, capabilities, etc.)
- [ ] add support for CSS inside the VSCode extension

nstepien:
I was planning on enabling the Rome formatter with its next release to replace prettier in one of our repos, but I realized that CSS formatting, specifically css-in-js, isn't supported yet.

Very much looking forward to Rome having CSS formatting support!

ematipico:
CSS-in-JS is still JavaScript, right? If so, then I believe that that the JS formatter should do that

nstepien:
>CSS-in-JS is still JavaScript, right?

It depends on the css-in-js lib, in our case we use linaria so it looks like this:
```tsx
const myClass = css`
  color: red;
`;
```
which is plain CSS (without the topmost selector) in a template string.
I believe #2572 is related.

With some other libs like `vanilla-extract` it may be
```tsx
const myClass = style({
  color: 'red'
});
```
which is indeed plain js.

ematipico:
> > CSS-in-JS is still JavaScript, right?
> 
> It depends on the css-in-js lib, in our case we use linaria so it looks like this:
> 
> ```tsx
> const myClass = css`
>   color: red;
> `;
> ```
> 
> which is plain CSS (without the topmost selector) in a template string. I believe #2572 is related.

I did a quick search, I think prettier call these cases "embedded" languages: 
https://github.com/prettier/prettier/blob/d13feed42b6478710bebbcd3225ab6f203a914c1/src/language-js/embed/css.js

And yes, they use a different parser/formatter inside these isntances. Thank you for pointing it out!

BasixKOR:
[...]

PR #2363, opened 2022-04-06
feat(rome_css_syntax): codegen for CSS grammar

ematipico:
## Summary

This PR is part of #2350 

It starts to spin the ball around multi language support, by generating the code for the CSS grammar. The scope of this PR is exclusively having a code generation that works and emits code that compiles.

There are few things won't be tackled in this PR:
- repeated code between `rome_js_syntax` and `rome_css_syntax`; for `AstNode` is repeated at the moment and it's intentional for now;
- how valid is the `css.ungram`; it is still incomplete and it doesn't hold the grammar, the code was shown to Nicholas and Micha few weeks back to show them the differences from the JS grammar and if we can achieve that;

What's been done:
- created a new `rome_css_syntax` crate that contains the generated code and some other code related to the syntax. Some code is copied and repeated;
- the code generation now uses a `LanguageKind` to generate different code accordingly to the value of the variant;
- handled some edge cases that we now have with CSS. We have the `$=` which causes issues inside Rust (`$` is mistaken as a pattern of a macro), so I had to play with `proc_macro2` to generate the correct code;
- created an empty crate called `rome_syntax` where we will have our shared syntax, such as `AstNode`, `AstSeparatedList`, etc. 

## Test Plan

`cargo check` should compile


crates/rome_css_syntax/src/syntax_node.rs
```diff
@@ -0,0 +1,29 @@
+//! This module defines the Concrete Syntax Tree used by RSLint.
```
leops, line 1:
```suggestion
//! This module defines the Concrete Syntax Tree used by Rome.
```

xtask/codegen/src/css_kinds_src.rs
```diff
@@ -0,0 +1,275 @@
...
+        (")", "R_PAREN"),
+        ("{", "L_CURLY"),
+        ("}", "R_CURLY"),
+        ("[", "L_BRACK"),
+        ("]", "R_BRACK"),
+        ("<", "L_ANGLE"),
+        (">", "R_ANGLE"),
+        ("~", "TILDE"),
+        ("#", "HASH"),
+        // These are *not* question AND dot tokens, they are one
```
leops, line 17: This comment seems out of place for the CSS syntax

crates/rome_css_syntax/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    SyntaxText, TextLen, TextRange, TextSize, TokenAtOffset, TriviaPieceKind, WalkEvent,
+};
+pub use syntax_node::*;
+
+use crate::CssSyntaxKind::*;
+use rome_rowan::RawSyntaxKind;
+
+impl From<u16> for CssSyntaxKind {
+    fn from(d: u16) -> CssSyntaxKind {
+        assert!(d <= (CssSyntaxKind::__LAST as u16));
```
xunilrj, line 20: Equals here make sense? Given that ```__LAST``` is just a marker and not a node.

crates/rome_syntax/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(test)]
```
xunilrj, line 1: Is this here by mistake? Or just for the crate to have something?

xtask/codegen/src/kinds_src.rs
```diff
@@ -598,7 +599,7 @@ pub struct AstEnumSrc {
 }
 
 impl Field {
-    pub fn method_name(&self) -> proc_macro2::Ident {
+    pub fn method_name(&self, language_kind: LanguageKind) -> proc_macro2::Ident {
         match self {
             Field::Token { name, .. } => {
                 let name = match name.as_str() {
```
xunilrj, line 605: hum... should this be 

```rust
match (name.as_str(), language_kind) {
   ...
   ("&=", _) => "bitwise_and_assign",
   ("|=", Js) => "bitwise_or_assign",
   ("|=", Css) => "exactly_or_hyphen",
   ...
}
```

xunilrj (approved):

crates/rome_syntax/src/lib.rs
ematipico: This is created by cargo automatically when you run `cargo new --lib`. I didn't touch this crate at all, only created via cargo

crates/rome_css_syntax/src/lib.rs
MichaReiser: Using equals ensures it's symmetric with the `impl From<CssSyntaxKind> for u16`

crates/rome_css_syntax/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+                | CssSyntaxKind::WHITESPACE
+                | CssSyntaxKind::COMMENT
+                | CssSyntaxKind::MULTILINE_COMMENT
+        )
+    }
+
+    /// Returns `true` for any contextual (await) or non-contextual keyword
+    #[inline]
+    pub const fn is_keyword(self) -> bool {
+        true
```
MichaReiser, line 45: This doesn't look right

crates/rome_css_syntax/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    #[inline]
+    pub const fn is_contextual_keyword(self) -> bool {
+        (self as u16) >= (ALICEBLUE_KW as u16) && (self as u16) <= (CSS_SELECTOR as u16)
+    }
+
+    /// Returns true for all non-contextual keywords (includes future reserved keywords)
+    #[inline]
+    pub const fn is_non_contextual_keyword(self) -> bool {
+        self.is_keyword() && !self.is_contextual_keyword()
+    }
```
MichaReiser, line 58: Does CSS has contextual keywords? If not. delete these methods

crates/rome_css_syntax/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl rome_rowan::SyntaxKind for CssSyntaxKind {
+    fn is_unknown(&self) -> bool {
+        matches!(self, CSS_UNKNOWN)
+    }
+
+    fn to_unknown(&self) -> Self {
+        todo!()
```
MichaReiser, line 67: Implement. The syntax factory relies on this method

crates/rome_css_syntax/src/util.rs
```diff
@@ -0,0 +1,68 @@
...
+        })
+    }
+
+    /// Try to cast this node to a certain AST node
+    fn try_to<T: AstNode>(&self) -> Option<T> {
+        T::cast(self.to_node().to_owned())
+    }
+
+    /// Whether the node contains any comments.
+    fn contains_comments(&self) -> bool {
```
MichaReiser, line 35: Can you remove any methods not needed by now?

xtask/codegen/src/generate_nodes.rs
```diff
@@ -1,11 +1,17 @@
...
+use crate::{to_lower_snake_case, to_upper_snake_case, LanguageKind};
+use quote::{format_ident, quote, TokenStreamExt};
 use xtask::Result;
 
-pub fn generate_nodes(ast: &AstSrc) -> Result<String> {
+pub fn generate_nodes(ast: &AstSrc, language_kind: LanguageKind) -> Result<String> {
+    let syntax_kind = match language_kind {
+        LanguageKind::Js => quote! { JsSyntaxKind },
+        LanguageKind::Css => quote! { CssSyntaxKind },
+    };
```
MichaReiser, line 14: Nit: Implement a `syntax_kind()` method on `LanguageKind` and then call this method instead of copy pasting the match in all code paths

crates/rome_css_syntax/src/lib.rs
ematipico: This was a copy paste from `rome_js_syntax`. I think it would make senses to keep `debug_assert`, or remove it. Although the same change should be applied to the other crate too. I would leave it out of the scope for now, for this PR.

leops: This should stay as an `assert!` since it's guarding unsafe code, if something goes wrong with this conversion a panic will be a lot easier to debug than some random undefined behavior down the line when the code matches on an invalid enum

crates/rome_css_syntax/src/lib.rs
ematipico: Yeah it does: `not`, `and`, `from`, `to`, etc.

xtask/codegen/src/css_kinds_src.rs
```diff
@@ -0,0 +1,275 @@
...
+        ("++", "PLUS2"),
+        ("*", "STAR"),
+        ("**", "STAR2"),
+        ("/", "SLASH"),
+        ("^", "CARET"),
+        ("%", "PERCENT"),
+        (".", "DOT"),
+        (":", "COLON"),
+        ("=", "EQ"),
+        ("==", "EQ2"),
```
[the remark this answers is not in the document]
ematipico: Looking into it! (yes, CSS allows `^=`)

PR #2366, opened 2022-04-06
fix(playground): checkbox bug

NicholasLYang:
## Summary
Fixes bug in playground where changing the typescript checkbox deactivates the jsx one.

### DIFF
--- a/website/playground/src/SourceTypeSelect.tsx
+++ b/website/playground/src/SourceTypeSelect.tsx
@@ -51,8 +51,6 @@ export default function SourceTypeSelect(
 							type="checkbox"
 							checked={isTypeScript}
 							onChange={(e) => {
-								setIsTypeScript(false);
-								setIsJsx(false);
 								setIsTypeScript(e.target.checked);
 							}}
 							className="focus:ring-indigo-500 h-4 w-4 text-indigo-600 border-gray-300 rounded disabled:opacity-30"


ematipico:
That is intended. Script snippet don't allow JSX nor Typescript

MichaReiser:
> That is intended. Script snippets don't allow JSX nor Typescript

Why so?

MichaReiser:
> > > That is intended. Script snippets don't allow JSX nor Typescript
> > 
> > 
> > Why so?
> 
> A script can't have JSX because it requires a compilation step. TypeScript can't be loaded as a script because it requires a compilation step. Is there something I am missing?

The only difference between `SourceType::Script` and `Module` is that 
* `Module` is in strict mode by default
* `Module` allows ES `import`/`export` statements. 

The `Script` and `Module` wording matches the wording of the ECMAScript specification ([Script](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#prod-Script), [Module](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#sec-modules)

NicholasLYang:
If we didn't want script snippets to have TypeScript or JSX, perhaps the better option would be to hide the checkboxes when we're on script mode? Or deactivate them. But I don't see anything wrong with someone trying script + jsx + typescript

ematipico (approved):

website/playground/src/SourceTypeSelect.tsx
```diff
@@ -51,8 +51,6 @@ export default function SourceTypeSelect(
 							type="checkbox"
 							checked={isTypeScript}
 							onChange={(e) => {
-								setIsTypeScript(false);
-								setIsJsx(false);
```
ematipico, line 55: Yeah this was an oversight of mine. I wanted to apply these two calls when source type changes. Sorry about that

PR #2372, opened 2022-04-07
refactor: Move AST types to `rome_rowan`

MichaReiser:
## Summary

This PR parametrizes the `Ast` traits with a `Language` type and moves them to the `rome_rowan` crate so that they can be shared between `rome_js_syntax` and `rome_css_syntax` (and potential future crates). 

Other changes: 

* Renames the JS/CSS `Syntax`* aliases to `JsSyntax*` and `CssSyntax`. For example: `type SyntaxNode = rome_rowan::SyntaxNode<JsLanguage>` is now called `JsSyntaxNode`. I believe this helps understanding the difference between the different nodes and aligns the naming with the kind types (that are called `JsSyntaxKind` and `CssSyntaxKind`
* Moved most `util` methods to the `SyntaxNode` struct. 
* Renamed `cast` to `try_cast` and move the `SyntaxNodeExt::to` to method to `AstNode` with the name `cast`. 

## Test Plan

`cargo test`


crates/rome_rowan/src/ast/mod.rs
```diff
@@ -0,0 +1,636 @@
...
+pub trait AstNode<L: Language> {
+    /// Returns `true` if a node with the given kind can be cased to this AST node.
+    fn can_cast(kind: L::Kind) -> bool;
+
+    /// Tries to cast the passed syntax node to this AST node.
+    ///
+    /// # Returns
+    ///
+    /// [None] if the passed node is of a different kind. [Some] otherwise.
+    fn try_cast(syntax: SyntaxNode<L>) -> Option<Self>
```
MichaReiser, line 29: Not sure about the naming of `cast` and `try_cast`. Most languages implement the `cast` operator in a way that it throws if casting isn't possible at runtime and is why I renamed the function returning an option to `try_cast`. However, I don't like that it may push users towards using `cast` which isn't entirely safe.

ematipico (commented):
This a great change! I will need to for the formatter too :)

ematipico (approved):

crates/rome_js_syntax/src/syntax_node.rs
```diff
@@ -15,15 +15,13 @@ impl Language for JsLanguage {
...
-pub type SyntaxSlots = rome_rowan::SyntaxSlots<JsLanguage>;
-pub type SyntaxSlot = rome_rowan::SyntaxSlot<JsLanguage>;
+pub type JsSyntaxNode = rome_rowan::SyntaxNode<JsLanguage>;
+pub type JsSyntaxToken = rome_rowan::SyntaxToken<JsLanguage>;
+pub type JsSyntaxElement = rome_rowan::SyntaxElement<JsLanguage>;
+pub type JsSyntaxNodeChildren = rome_rowan::SyntaxNodeChildren<JsLanguage>;
+pub type JsSyntaxElementChildren = rome_rowan::SyntaxElementChildren<JsLanguage>;
+pub type JsSyntaxList = rome_rowan::SyntaxList<JsLanguage>;
+pub type SyntaxResult<T> = rome_rowan::SyntaxResult<T>;
+pub type SyntaxError = rome_rowan::SyntaxError;
```
ematipico, line 25: I think `SyntaxError` should be be exposed (it doesn't have any generic type). Hence we can import  from `rome_rowan`. This will make the changes of the formatter easier

.gitattributes
```diff
@@ -1,3 +1,4 @@
 * text=auto eol=lf
 crates/rome_js_parser/src/lexer/tables.rs linguist-generated=true text=auto eol=lf
 crates/rome_js_syntax/src/generated/* linguist-generated=true text=auto eol=lf
+crates/rome_css_syntax/src/generated/* linguist-generated=true text=auto eol=lf
```
leops, line 4: We could probably make that rule generic for every folder called "generated"

crates/rome_rowan/src/ast/mod.rs
leops: Maybe renaming `cast` to `unwrap_cast` (like `Result::unwrap_err`) would better communicate it's a panicking function ?

PR #2374, opened 2022-04-08
perf(build): improve compile times

leops:
## Summary

Based on the build timings information newly exposed by Cargo in the 1.60 toolchain, I modified the way the project is built to remove dependencies from the critical path (or at least shift them off). The changes include:

- Removed `thiserror` and `tracing::instrument` from critical crates so they don't depend on `syn` to be built
- Removed `BigInt` parsing from `rome_js_syntax`, at the moment we don't need to inspect the numeric values of those, if it becomes needed later on it would be better to perform that in a utility crate downstream
- Moved `rome_fs`, `rome_core` and `rome_js_parser` to the dev dependencies of the formatter so they don't block the build

The critical path for building the CLI is now `autocfg` -> `memoffset` -> `rome_rowan` -> `rome_js_syntax` -> `rome_js_formatter`, and for the LSP `proc-macro2` -> `quote` -> `syn` -> `serde-derive` -> `serde` -> `serde_json` -> `lsp-types` -> `rome_lsp`


crates/rome_formatter/src/printer.rs
```diff
@@ -102,9 +102,8 @@ impl<'a> Printer<'a> {
     }
 
     /// Prints the passed in element as well as all its content
-    #[tracing::instrument(level = "debug", skip_all)]
     pub fn print(self, element: &'a FormatElement) -> Formatted {
-        self.print_with_indent(element, 0)
+        tracing::debug_span!("Printer::print").in_scope(move || self.print_with_indent(element, 0))
```
ematipico, line 106: If I understood correctly, the tracing macro makes use of `syn`, is it?

ematipico (approved):
I leave the final approve to @MichaReiser regarding the bigint parsing

crates/rome_formatter/src/printer.rs
leops: `syn` is only being used for the `tracing::instrument` procedural macro since it needs to parse and rewrite the body of the function it's attached to in order to insert the instrumentation code. The other macros in `tracing` are fine, they're just regular `macro_rules` and do not require any external dependency to work

MichaReiser (approved):

xunilrj:
Would be nice to post here some numbers, they can be on your machine. Just to give us an idea of the improvement...

leops:
> Would be nice to post here some numbers, they can be on your machine. Just to give us an idea of the improvement...

Here's the build timings for the latest main:
perf-before

And on this branch after rebasing:
perf-after

ematipico:
20 seconds saved, not bad

PR #2376, opened 2022-04-08
refactor(rome_js_formatter): Support arbitrarily deep binary expressions

MichaReiser:
## Summary

`format_binarish` used to use recursion to format the binary expressions. While this led to somewhat simpler code it had the downside that it stack overflowed for binary expressions with deep left subtrees. 

This PR rewrites the implementation to use an `Iterator` instead that visits the left-hand sides of binary-like-expressions in post order (first visiting the deepest left-hand side, then traversing upwards). 

This PR further fixes a unicode issue in the diff printing logic and renames `operator` to `operator_token` for a consistent naming.

## Test Plan

`cargo test`

Running the formatter on all test repositories (t262, TypeScript...) no longer fails with a stack overflow. There still seems to be some issue that rome enters an infinite loop. I need to look into this separately.

### DIFF
--- a/crates/rome_console/src/diff.rs
+++ b/crates/rome_console/src/diff.rs
@@ -30,7 +30,7 @@ pub enum DiffMode {
     /// 11  9 |   elit,
     ///       | @@ -15,7 +13,5 @@
     /// 14 12 |   eiusmod
-    /// 15 13 |   
+    /// 15 13 |
     /// 16 14 |   incididunt
     /// 17    | - function
     /// 18    | - name(
@@ -154,8 +154,11 @@ impl<'a> Display for Diff<'a> {
                         fmt.write_str(" | ")?;
 
                         let line = change.value().trim_end();
-                        let line_length = line.len().min(MAX_LINE_LENGTH);
-                        let line = &line[..line_length];
+                        let line = line
+                            .char_indices()
+                            .nth(MAX_LINE_LENGTH)
+                            .map(|(byte_index, _)| &line[..byte_index])
+                            .unwrap_or_else(|| &line);
 
                         match change.tag() {
                             ChangeTag::Delete => {
--- a/crates/rome_console/src/diff.rs
+++ b/crates/rome_console/src/diff.rs
@@ -158,7 +158,7 @@ impl<'a> Display for Diff<'a> {
                             .char_indices()
                             .nth(MAX_LINE_LENGTH)
                             .map(|(byte_index, _)| &line[..byte_index])
-                            .unwrap_or_else(|| &line);
+                            .unwrap_or_else(|| line);
 
                         match change.tag() {
                             ChangeTag::Delete => {
--- a/crates/rome_js_formatter/src/utils/binary_like_expression.rs
+++ b/crates/rome_js_formatter/src/utils/binary_like_expression.rs
@@ -5,12 +5,12 @@ use crate::{
     space_token, token, FormatElement, FormatResult, Formatter, ToFormatElement,
 };
 
-use rome_js_syntax::JsSyntaxKind::*;
 use rome_js_syntax::{
-    AstNode, JsAnyExpression, JsAnyInProperty, JsBinaryExpression, JsBinaryOperator,
-    JsInExpression, JsInstanceofExpression, JsLogicalExpression, JsLogicalOperator, JsPrivateName,
-    JsSyntaxKind, SyntaxNode, SyntaxNodeExt, SyntaxResult, SyntaxToken,
+    JsAnyExpression, JsAnyInProperty, JsBinaryExpression, JsBinaryOperator, JsInExpression,
+    JsInstanceofExpression, JsLanguage, JsLogicalExpression, JsLogicalOperator, JsPrivateName,
+    JsSyntaxKind, JsSyntaxKind::*, JsSyntaxNode, JsSyntaxToken,
 };
+use rome_rowan::{AstNode, SyntaxResult};
 use std::cmp::Ordering;
 use std::fmt::Debug;
 use std::iter::FusedIterator;
@@ -173,7 +173,7 @@ enum BinaryLikeOperator {
 /// In order to make this distinction more obvious, we wrap `foo && bar` in parenthesis.
 fn format_with_or_without_parenthesis(
     parent_operator: BinaryLikeOperator,
-    node: &SyntaxNode,
+    node: &JsSyntaxNode,
     formatted_node: FormatElement,
 ) -> FormatResult<(FormatElement, bool)> {
     let compare_to = match JsAnyExpression::cast(node.clone()) {
@@ -245,7 +245,7 @@ fn can_hard_group(flatten_nodes: &[FlattenItem]) -> bool {
             .all(|node| !node.has_comments() && !node.is_group())
 }
 
-fn is_inside_parenthesis(current_node: &SyntaxNode) -> bool {
+fn is_inside_parenthesis(current_node: &JsSyntaxNode) -> bool {
     let parent_kind = current_node.parent().map(|parent| parent.kind());
 
     matches!(
@@ -265,7 +265,7 @@ fn is_inside_parenthesis(current_node: &SyntaxNode) -> bool {
 ///
 /// There are some cases where the indentation is done by the parent, so if the parent is already doing
 /// the indentation, then there's no need to do a second indentation.
-fn should_not_indent_if_parent_indents(current_node: &SyntaxNode) -> bool {
+fn should_not_indent_if_parent_indents(current_node: &JsSyntaxNode) -> bool {
     let parent_kind = current_node.parent().map(|parent| parent.kind());
 
     matches!(
@@ -278,7 +278,7 @@ fn should_not_indent_if_parent_indents(current_node: &SyntaxNode) -> bool {
 /// these cases the decide to actually break on a new line and indent it.
 ///
 /// This function checks what the parents adheres to this behaviour
-fn should_indent_if_parent_inlines(current_node: &SyntaxNode) -> bool {
+fn should_indent_if_parent_inlines(current_node: &JsSyntaxNode) -> bool {
     let parent = current_node.parent();
     let grand_parent = parent.as_ref().and_then(|p| p.parent());
 
@@ -302,7 +302,7 @@ impl FlattenItems {
     fn format_binary_expression_right_hand_side(
         &mut self,
         expression: JsAnyBinaryLikeExpression,
-        parent_operator: Option<SyntaxToken>,
+        parent_operator: Option<JsSyntaxToken>,
         formatter: &Formatter,
     ) -> FormatResult<()> {
         let should_flatten = expression.can_flatten()?;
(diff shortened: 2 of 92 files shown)


MichaReiser:
!bench_formatter

crates/rome_console/src/diff.rs
```diff
@@ -154,8 +154,11 @@ impl<'a> Display for Diff<'a> {
                         fmt.write_str(" | ")?;
 
                         let line = change.value().trim_end();
-                        let line_length = line.len().min(MAX_LINE_LENGTH);
-                        let line = &line[..line_length];
+                        let line = line
+                            .char_indices()
+                            .nth(MAX_LINE_LENGTH)
+                            .map(|(byte_index, _)| &line[..byte_index])
+                            .unwrap_or_else(|| line);
```
MichaReiser, line 161: @leops I believe this still does about the same as you intended, right? It should take the first 250 characters. The only difference now is that it takes the first 250 unicode characters and avoids slicing in the middle of a unicode character.

leops (approved):
The change in diff printing looks good, I've been meaning to add more tests around the handling of Unicode there anyway. On the `PostorderIterator` I think it looks quite similar to other existing iterators in `rome_rowan`, would it make sense to move this kind of utility there (in a more generic form) ?

MichaReiser:
> The change in diff printing looks good, I've been meaning to add more tests around the handling of Unicode there anyway. On the `PostorderIterator` I think it looks quite similar to other existing iterators in `rome_rowan`, would it make sense to move this kind of utility there (in a more generic form) ?

The `PostorderIterator` is inspired by the `PreorderIterator` and I first planned to write it in a genric way. However, the iterator here is special for two reasons:

* It only traverses into the left-hand sides of binary expressions. It never visits the right-hand side. 
* It only operates on binary expressions. 

Making it more generic would require a lot of additional casting (casting the `SyntaxNode` back to a binary expression) and some way to tell the iterator to only visit Binary like expressions (similar to `skip_subtree`), and skip over the right hand side

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            match self.next.take()? {
+                VisitEvent::Enter(binary) => {
+                    let left_expression = binary
+                        .left()
+                        .ok()
+                        .and_then(|left| left.as_expression().cloned())
+                        .and_then(|left| JsAnyBinaryLikeExpression::cast(left.syntax().clone()));
```
ematipico, line 619: I haven't thought about implementing `AstNode` to our internal enums! Sometimes I am too narrow minded! 🤯

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+
+                    if let Some(expression) = left_expression {
+                        self.next = Some(VisitEvent::Enter(expression));
+                    } else {
+                        // If left is missing or it isn't a binary expression, then format it as part of the parent binary like expression
```
ematipico, line 624:
```suggestion
                        // If left is missing or it isn't a binary like expression, then format it as part of the parent binary like expression
```

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+            start: start.syntax().clone(),
+            next: Some(VisitEvent::Enter(start)),
+        }
+    }
+}
+
+impl Iterator for PostorderIterator {
+    type Item = JsAnyBinaryLikeExpression;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
ematipico, line 611: Can explain with a comment what are doing when we call `next`?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+#[derive(Debug)]
+enum VisitEvent {
+    Enter(JsAnyBinaryLikeExpression),
+    Exit(JsAnyBinaryLikeExpression),
+}
+
+/// Iterator that visits the left hand sides of all binary like expressions in post-order.
+struct PostorderIterator {
+    next: Option<VisitEvent>,
+    start: JsSyntaxNode,
```
ematipico, line 596: Can we document these two fields?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+                VisitEvent::Exit(node) => {
+                    if node.syntax() != &self.start {
+                        self.next = node.syntax().parent().map(|parent| {
+                            let expression = JsAnyBinaryLikeExpression::cast(parent).unwrap();
+                            VisitEvent::Exit(expression)
+                        });
+                    }
+
+                    return Some(node);
+                }
```
ematipico, line 637: Can we document what we are doing here? Also, can you add a `// SAFETY` comment around `unwrap`?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+            JsAnyBinaryLikeExpression::JsLogicalExpression(logical) => logical.syntax(),
+            JsAnyBinaryLikeExpression::JsBinaryExpression(binary) => binary.syntax(),
+            JsAnyBinaryLikeExpression::JsInstanceofExpression(instanceof) => instanceof.syntax(),
+            JsAnyBinaryLikeExpression::JsInExpression(in_expression) => in_expression.syntax(),
+        }
+    }
+}
+
+impl JsAnyBinaryLikeExpression {
+    fn can_flatten(&self) -> SyntaxResult<bool> {
```
ematipico, line 752: Can we briefly document when a binary like expression can be flattened?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+}
+
+impl ToFormatElement for JsAnyBinaryLikeLeftExpression {
+    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        match self {
+            JsAnyBinaryLikeLeftExpression::JsAnyExpression(expression) => {
+                expression.to_format_element(formatter)
+            }
+            JsAnyBinaryLikeLeftExpression::JsPrivateName(private_name) => {
+                private_name.to_format_element(formatter)
```
ematipico, line 829: As both `expression` and `private_name` are typed (not `SyntaxNode`), we have to use `format(formatter)`. `expression` though is a `JsAny` node, so it's fine using `to_format_element`, although `private_name` it is not, so we have to use `format` so we don't bypass the check around comment suppression.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+    Instanceof,
+    In,
+}
+
+/// This function is in charge of formatting a node inside a binaryish expression with parenthesis or not
+///
+/// At the moment this logic is applied only to logical expressions.
+///
+/// A logical expressions should be decorated with parenthesis only if its previous operation has a lower
+/// precedence.
```
ematipico, line 162: I believe this comment is outdated. We now apply parenthesis to all type of binary like expressions. Could you please remove it?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+    /// The left hand-side expression and the current operator cannot be flattened.
+    /// Format the left hand side by its own and potentially wrap it in parentheses before formatting
+    /// the right-hand side of the current expression.
+    fn format_new_binary_like_group(
+        &mut self,
+        binary_like_expression: JsAnyBinaryLikeExpression,
+        parent_operator: Option<JsSyntaxToken>,
+        formatter: &Formatter,
+    ) -> FormatResult<()> {
+        if let Some(last) = self.items.last_mut() {
```
ematipico, line 350: As far as I understood `last_mut` is a O(n) operation. Wouldn't it better to keep track of the items pushed and then use this index to surgically retrieve the last item?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -0,0 +1,858 @@
...
+        let (trailing_separator, parent_operator_has_comments) = if let Some(parent_operator) =
+            parent_operator.as_ref()
+        {
+            let previous_operator_has_trailing_comments = parent_operator.has_trailing_comments();
+            (
+                if previous_operator_has_trailing_comments {
+                    TrailingTerminator::HardLineBreak
+                } else {
+                    TrailingTerminator::None
+                },
```
ematipico, line 390: nit: a little hard to read, I would move it inside a variable

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: This is the same as for any other union

https://github.com/rome/tools/blob/437a579e745e4b178a0b3325d412b8e4aa64e3e7/crates/rome_js_formatter/src/js/any/array_assignment_pattern_element.rs#L6-L14

And we call into it using the formatter:

```
let left = parent.left()?;

let formatted = left.format(formatter)?;
```

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: I can extend the documentation on the iterator but next is otherwise documented by the `Iterator` trait.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
ematipico: > This is the same as for any other union

That's what I said. But `JsPrivateName` is not an union, we need to call `format`

crates/rome_js_formatter/src/utils/binary_like_expression.rs
ematipico: Sorry for the misunderstanding, what I meant is **how we are retrieving** `next`, the logic of the algorithm you implemented.

MichaReiser: The struct can describe the overall idea but I rather not document the exact algorithm because it will likely soon be out of sync with the implementation.

ematipico: As this is a complex algorithm, and I don't fully understand what we're doing, I hoped to have some comment as explanation. 

Agree to disagree.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: `array_assignment_pattern_element` also calls `to_format_element` for non unions. This is safe because this is only the `to_format_element` dispatch to the right implementation. The `format_binary_like` correctly calls into `.format` for the left hand side above which will handle trivia and such. 

https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/utils/binary_like_expression.rs#L125-L128

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: `last_mut` is `O(1)`. It retrieves a reference to the last element. It then mutates the fields in place. An alternative would have been to use `pop` and then push a new element but that wouldn't change anything.

PR #2378, opened 2022-04-08
feature(rome_js_parser): better handle of unicode for string literals, templates and regex

xunilrj:
We are doing something dangerous, that was the cause of https://github.com/rome/tools/pull/2344.

When parsing the contents of string literals, regexes (https://english.stackexchange.com/questions/508834/what-is-the-plural-form-of-regex) and string templates; we normally do

```rust
while let Some(byte) = self.next_byte() {
    ...
}
```

This is dangerous because sometimes we check with an ```if```, if we are in a linebreak. Although the Unicode committee try to avoid ambiguity when creating Unicode chars, we fall into traps because we also use ```from_utf8_unchecked``` out of char boundaries.

An example is the ‍ Zero Width Joiner Unicode char (https://unicode-table.com/en/200D/) that caused the #2344 bug. Its bytes are ```E2 80 8D```, and  ```from_utf8_unchecked``` will return an ````\r```` when we call it on its  second byte.

See this here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=66e9d16029decb91cdfcb7a41f2757bf

Look at the 15th byte;

To avoid this issue this PR correctly iterate the code using utf8 chars when they are possible. I tried to keep ```next_byte``` where it is safe because it is faster.

### DIFF
--- a/crates/rome_js_parser/src/lexer/mod.rs
+++ b/crates/rome_js_parser/src/lexer/mod.rs
@@ -807,6 +807,12 @@ impl<'src> Lexer<'src> {
                 }
                 'u' => self.read_unicode_escape(true).is_ok(),
                 'x' => self.validate_hex_escape(),
+                '\r' => {
+                    if self.next_byte() == Some(b'\n') {
+                        self.advance(1);
+                    }
+                    true
+                }
                 _ => {
                     self.advance_char_unchecked();
                     true
@@ -875,11 +881,8 @@ impl<'src> Lexer<'src> {
                         valid = false;
                     }
                 }
-                chr if chr == quote => {
-                    self.advance_char_unchecked();
-                    return valid;
-                }
-                chr if is_linebreak(chr) && !jsx_attribute => {
+
+                '\r' | '\n' if !jsx_attribute => {
                     let unterminated =
                         Diagnostic::error(self.file_id, "", "unterminated string literal")
                             .primary(start..self.position, "")
@@ -887,6 +890,12 @@ impl<'src> Lexer<'src> {
                     self.diagnostics.push(unterminated);
                     return false;
                 }
+
+                chr if chr == quote => {
+                    self.advance_char_unchecked();
+                    return valid;
+                }
+
                 _ => {
                     self.advance_char_unchecked();
                 }
@@ -1496,6 +1505,7 @@ impl<'src> Lexer<'src> {
         let start = self.position;
         let mut in_class = false;
 
+        let mut last_len = 1;
         while let Some(c) = self.next_char() {
             match c {
                 '[' => in_class = true,
@@ -1592,12 +1602,14 @@ impl<'src> Lexer<'src> {
                     );
 
                     // Undo the read of the new line trivia
-                    self.position -= c.len_utf8();
+                    self.position -= last_len;
 
                     return JsSyntaxKind::JS_REGEX_LITERAL;
                 }
                 _ => {}
             }
+
+            last_len = c.len_utf8();
         }
 
         self.diagnostics.push(
--- a/crates/rome_js_parser/src/lexer/mod.rs
+++ b/crates/rome_js_parser/src/lexer/mod.rs
@@ -392,22 +392,20 @@ impl<'src> Lexer<'src> {
         debug_assert!(!self.is_eof());
 
         // SAFETY: `lex_token` only calls this method if it isn't passed the EOF
-        let byte = unsafe { self.current_unchecked() };
+        let chr = unsafe { self.current_char_unchecked() };
 
-        match byte {
+        match chr {
             // `<`: empty jsx text, directly followed by another element or closing element
-            b'<' => self.eat_byte(T![<]),
+            '<' => self.eat_byte(T![<]),
             // `{`: empty jsx text, directly followed by an expression
-            b'{' => self.eat_byte(T!['{']),
+            '{' => self.eat_byte(T!['{']),
             _ => {
-                while let Some(byte) = self.current_byte() {
+                while let Some(chr) = self.current_char() {
                     // but not one of: { or < or > or }
-                    match byte {
+                    match chr {
                         // Start of a new element, the closing tag, or an expression
-                        b'<' | b'{' => {
-                            break;
-                        }
-                        b'>' => {
+                        '<' | '{' => break,
+                        '>' => {
                             self.diagnostics.push(
                                 Diagnostic::error(
                                     self.file_id,
@@ -418,7 +416,7 @@ impl<'src> Lexer<'src> {
                             );
                             self.next_byte();
                         }
-                        b'}' => {
+                        '}' => {
                             self.diagnostics.push(
                                 Diagnostic::error(
                                     self.file_id,
@@ -430,7 +428,7 @@ impl<'src> Lexer<'src> {
                             self.next_byte();
                         }
                         _ => {
-                            self.next_byte();
+                            self.advance_char_unchecked();
                         }
                     }
                 }
@@ -444,12 +442,11 @@ impl<'src> Lexer<'src> {
         debug_assert!(!self.is_eof());
 
         // Safety: Guaranteed because we aren't at the end of the file
-        let byte = unsafe { self.current_unchecked() };
+        let chr = unsafe { self.current_char_unchecked() };
 
-        match byte {
-            b'\'' | b'"' => {
+        match chr {
+            '\'' | '"' => {
                 self.read_str_literal(true);
-
                 JSX_STRING_LITERAL
             }
             _ => self.lex_token(),
@@ -506,7 +503,9 @@ impl<'src> Lexer<'src> {
     }
 
     /// Get the UTF8 char which starts at the current byte
-    /// Safety: Must be called at the begining of a UTF8 char.
+    ///
+    /// ## Safety
+    /// Must be called at the begining of a UTF8 char.
     fn current_char_unchecked(&self) -> char {
         // This is unreachable for all intents and purposes, but this is just a precautionary measure
         debug_assert!(!self.is_eof());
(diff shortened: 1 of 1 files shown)


xunilrj:
!bench_parser

Boshen:
I'm not sure if we should just rewrite the lexer using https://doc.rust-lang.org/stable/std/str/struct.Chars.html, all these boundary checking can be tucked away, and it shouldn't affect performance. 

I'm bringing this up because I studied this https://github.com/mozilla-spidermonkey/jsparagus/blob/master/crates/parser/src/lexer.rs#L27 , its tokenization code is so much cleaner, especially for unicode handling.

xunilrj:
> I'm not sure if we should just rewrite the lexer ...

Yes. I am also wondering that. I actually have a "poc" to create a discussion on how to abstract encoding for the lexer. In this case, we would also support UTF-16 and others...

ematipico:
On that note, I think we should make this generic and not isolated to a single parser. There are many places where we will need to handle strings with different encoding: different parsers, formatter, analysers, documentation generation (hence comments), etc.

MichaReiser:
>  I'm not sure if we should just rewrite the lexer ...

Rewriting the lexer isn't a priority right now. It's working well enough and there are many other technical challenges that need to be tackled first. 

> On that note, I think we should make this generic and not isolated to a single parser. There are many places where we will need to handle strings with different encoding: different parsers, formatter, analysers, documentation generation (hence comments), etc.

The Lexer is very performance sensitive which is why a generic solution might not be feasible. I would even argue that any non UTF8 encoding should be converted to UTF8 before passing it to the lexer because non UTF8 is an edge case and we shouldn't sacrifice performance to handle the few non UTF8 files. 

Also, the `rust-cssparser` is working directly on bytes which is working fine. [source](https://github.com/servo/rust-cssparser/blob/master/src/tokenizer.rs). We probably just need to be more careful when we use `advance`, `next_byte` vs `next_char`.

xunilrj:
!bench_parser

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -462,41 +463,54 @@ impl<'src> Lexer<'src> {
...
+                b'\r' | b'\n' => {
+                    self.advance(1);
+                }
+                chr => {
+                    if chr >= 0x80 {
+                        let chr = self.current_char_unchecked();
+                        if is_linebreak(chr) {
+                            self.advance(chr.len_utf8());
+                        }
+                    }
```
MichaReiser, line 480: Can we abstract away this check? Why is it needed here but not when matching on `b'r'` directly?

xunilrj:
!bench_parser

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -430,7 +428,11 @@ impl<'src> Lexer<'src> {
                             self.next_byte();
                         }
                         _ => {
-                            self.next_byte();
+                            if chr < 0x80 {
+                                self.next_byte();
+                            } else {
+                                self.advance_char_unchecked();
+                            }
```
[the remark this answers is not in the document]
xunilrj: Done. Using ```is_ascii``` is more clear.

> Would it be easier to just always call advance_char?

Yes. But this check is faster because 99% of the time we don't have UTF8 chars.

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -462,54 +463,82 @@ impl<'src> Lexer<'src> {
...
+    ///
+    /// ## Safety
+    /// Must be called at a valid UT8 char boundary
+    fn consume_whitespaces(&mut self) {
+        self.assert_at_char_boundary();
+
+        while let Some(chr) = self.current_byte() {
+            match lookup_byte(chr) {
+                Dispatch::WHS => {
+                    if let b'\r' | b'\n' = chr {
```
MichaReiser, line 500: Shouldn't this also test for the unicode line break characters (`is_linebreak`?)

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -462,54 +463,82 @@ impl<'src> Lexer<'src> {
...
+                    let chr = self.current_char_unchecked();
 
-            let dispatcher = lookup_byte(current);
-            if dispatcher == self::bytes::Dispatch::WHS
-                || (UNICODE_WHITESPACE_STARTS.contains(&current) && UNICODE_SPACES.contains(&chr))
-            {
-                self.advance(chr.len_utf8());
-            } else {
-                break;
+                    if UNICODE_SPACES.contains(&chr) {
```
MichaReiser, line 509: How does this ensure that the `chr` is a whitespace?

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -540,6 +569,12 @@ impl<'src> Lexer<'src> {
         }
     }
 
+    /// Asserts that the lexer is at a UTF8 char boundary
+    #[inline]
+    fn assert_at_char_boundary(&self) {
```
MichaReiser, line 574: Would it be safe to also add this chack to `advance`?

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -850,45 +882,48 @@ impl<'src> Lexer<'src> {
...
+                chr if chr == quote => {
+                    self.advance(1);
+                    return valid;
+                }
+                chr => {
+                    if chr.is_ascii() {
+                        self.advance(1);
+                    } else {
+                        self.advance_char_unchecked();
+                    }
```
MichaReiser, line 919: Nit: How about moving this into a helper? `advance_byte_or_char`

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -1455,19 +1493,25 @@ impl<'src> Lexer<'src> {
...
+                    if let b'\r' | b'\n' = chr {
                         return COMMENT;
+                    } else if chr.is_ascii() {
+                        self.advance(1);
+                    } else {
+                        let chr = self.current_char_unchecked();
+                        if is_linebreak(chr) {
+                            return COMMENT;
+                        } else {
+                            self.advance(chr.len_utf8());
```
MichaReiser, line 1507: When to use `advance(chr.len_utf8())` vs `advance_next_char_unchecked`?

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -1933,39 +2001,45 @@ impl<'src> Lexer<'src> {
...
-                }
-                Some(b'$') if self.peek_byte() == Some(b'{') => {
-                    token = Some(JsSyntaxKind::TEMPLATE_CHUNK);
-                    break;
-                }
-                Some(_) => {
-                    let _ = self.next_byte();
+                b'$' => {
+                    if let Some('{') = self.peek_char() {
+                        if self.position == start {
```
MichaReiser, line 2026: Isn't `peek_byte` enough here because `{` is ascii?

MichaReiser (commented):
Thanks for adding the asserts and the overall improvement of the lexer code. 

What are the changes to coverage submodules?

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: No, because of the ```lookup_byte```. If it were a Unicode it would return an ```Dispatch::UNI```.

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: Because UNICODE_SPACES contains all whitespaces we support.

What the old code does is check the first byteagainst ```UNICODE_WHITESPACE_STARTS``` and after the whole char at ```UNICODE_SPACES```, if we had a match at the first case.

I can skip the first byte check because it makes sense as optimization only for non-Unicode chars. But this case I moved to the above match arm.

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: Ideally yes. But we still have a small problem with identifiers and keywords that I want to deal with later.

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: ```advance(chr.len_utf8())```: you already have the ```char```.
```advance_next_char_unchecked```: you don't have the ```char```.

The second has the cost of reading the buffer.

An option to simplify all these cases could be:

```
fn advance(step: impl LexerAdvance) {
    self.pos += step.step();
}

trait LexerAdvance {
    fn step(&self, buffer: &[u8]) -> usize;
}

impl LexerAdvance for usize {
    fn step(&self, buffer: &[u8]) -> usize { self }
}

impl LexerAdvance for u8 {
    fn step(&self, buffer: &[u8]) -> usize {
        if self.is_ascii() {
            1
        } else {
            let chr = unsafe { std::str::from_utf8_unchecked(buffer) }.chars().next().unwrap();
            chr.len_utf8()
        }
    }
}

impl LexerAdvance for u8 {
    fn step(&self, buffer: &[u8]) -> usize {
        self.len_utf8()
    }
}
```

all we need to do now is call ```self.advance(1)``` or ```self.advance(chr)``` for both ```u8``` and ```char```.
And the compiler will get the best case, inline and optimize everything away.

We need to be careful with what the compiler will call on ```self.advance(1)```. Maybe an enum to disambiguate will be the better option here:

```
self.advance(Step::OneByte)
```

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: Yes.

crates/rome_js_parser/src/lexer/mod.rs
MichaReiser: I see, and the line breaks aren't part of the `UNICODE_SPACES`, thus, the method breaks for line breaks. Neat

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: Done.

crates/rome_js_parser/src/lexer/mod.rs
xunilrj: Done.

ematipico (approved):

PR #2379, opened 2022-04-09
perf: replace all assert_eq with debug_assert_eq

Boshen:
Just a quick test to see how much perf we can get from this.


Boshen:
!bench_parser

Boshen:
Not significant enough :-(

PR #2380, opened 2022-04-09
feature(rome_js_formatter): Add support for printing single quotes

cpojer:
## Summary
This PR adds a `--quote-style` option to the Rome Formatter so that users can choose between single and double quotes. Arguably, template literals should also be an available option, but that can be discussed later.

## Test Plan

* `cargo test`
* `cargo run -p rome_cli -- format --quote-style=single <file>` works.

### DIFF
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -23,7 +23,7 @@ use rome_diagnostics::{
 };
 use rome_fs::{AtomicInterner, FileSystem, PathInterner, RomePath};
 use rome_fs::{TraversalContext, TraversalScope};
-use rome_js_formatter::{FormatOptions, IndentStyle, QuoteStyle};
+use rome_js_formatter::{FormatOptions, IndentStyle};
 use rome_js_parser::{parse, SourceType};
 
 use crate::{CliSession, Termination};
@@ -66,14 +66,8 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
             source,
         })?;
 
-    match quote_style {
-        Some(QuoteStyle::Double) => {
-            options.quote_style = QuoteStyle::Double;
-        }
-        Some(QuoteStyle::Single) => {
-            options.quote_style = QuoteStyle::Single;
-        }
-        None => {}
+    if let Some(quote_style) = quote_style {
+        options.quote_style = quote_style;
     }
 
     let line_width = session
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -23,7 +23,7 @@ use rome_diagnostics::{
 };
 use rome_fs::{AtomicInterner, FileSystem, PathInterner, RomePath};
 use rome_fs::{TraversalContext, TraversalScope};
-use rome_js_formatter::{FormatOptions, IndentStyle};
+use rome_js_formatter::{FormatOptions, IndentStyle, QuoteStyle};
 use rome_js_parser::{parse, SourceType};
 
 use crate::{CliSession, Termination};
@@ -40,7 +40,7 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
             source,
         })?;
 
-    let style = session
+    let indent_style = session
         .args
         .opt_value_from_str("--indent-style")
         .map_err(|source| Termination::ParseError {
@@ -48,12 +48,30 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
             source,
         })?;
 
-    match style {
+    match indent_style {
         Some(IndentStyle::Tab) => {
             options.indent_style = IndentStyle::Tab;
         }
         Some(IndentStyle::Space(default_size)) => {
             options.indent_style = IndentStyle::Space(size.unwrap_or(default_size));
+        }
+        None => {}
+    }
+
+    let quote_style = session
+        .args
+        .opt_value_from_str("--quote-style")
+        .map_err(|source| Termination::ParseError {
+            argument: "--quote-style",
+            source,
+        })?;
+
+    match quote_style {
+        Some(QuoteStyle::Double) => {
+            options.quote_style = QuoteStyle::Double;
+        }
+        Some(QuoteStyle::Single) => {
+            options.quote_style = QuoteStyle::Single;
         }
         None => {}
     }
--- a/crates/rome_js_formatter/src/utils/mod.rs
+++ b/crates/rome_js_formatter/src/utils/mod.rs
@@ -7,7 +7,7 @@ mod simple;
 use crate::formatter_traits::{FormatOptionalTokenAndNode, FormatTokenAndNode};
 use crate::{
     empty_element, empty_line, format_elements, hard_group_elements, space_token, token,
-    FormatElement, FormatResult, Formatter, Token,
+    FormatElement, FormatResult, Formatter, QuoteStyle, Token,
 };
 pub(crate) use binary_like_expression::{format_binary_like_expression, JsAnyBinaryLikeExpression};
 pub(crate) use call_expression::format_call_expression;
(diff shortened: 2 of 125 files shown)


crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -614,16 +614,31 @@ pub(crate) fn format_string_literal_token(
...
+            if quoted.starts_with('"') && !quoted.contains('\'') {
+                let s = &quoted[1..quoted.len() - 1];
+                let s = format!("'{}'", s);
+                match normalize_newlines(&s, ['\r']) {
+                    Cow::Borrowed(_) => s,
+                    Cow::Owned(s) => s,
+                }
+            } else {
+                normalize_newlines(quoted, ['\r']).into_owned()
+            }
```
cpojer, line 640: How do I make this look nice?

crates/rome_js_formatter/tests/specs/js/module/string/string.js.snap
```diff
@@ -0,0 +1,167 @@
...
+import hey from 'hey';
+import hey from 'hey';
+import 'x' assert { type: 'json' };
+import 'foo' assert { 'type': 'json' };
+import foo from 'foo.json' assert { type: 'json' };
+import foo from 'foo.json' assert { type: 'json' };
+import foo2 from 'foo.json' assert {
+	'type': 'json',
+	type: "html",
+	"type": "js",
```
cpojer, line 125: Any idea why only the first two strings are reprinted with single-quotes in asserts?

crates/rome_js_formatter/tests/specs/js/module/array/array_nested.js.snap
```diff
@@ -1,8 +1,6 @@
 ---
 source: crates/rome_js_formatter/tests/spec_test.rs
-assertion_line: 157
 expression: array_nested.js
-
```
cpojer, line 5: I don't know what I did to trigger this removal across all files.

Boshen: You have the latest insta, which removed this line, no worries. https://github.com/mitsuhiko/insta/commit/5ab9f4093f6d62d8889918a8a84c99b386f1c816

crates/rome_js_formatter/src/utils/mod.rs
kassens: 👋 was just taking a look at Rome and surprise do see you around.

start with this, then use the them in the code to avoid the duplication.
```
let (primary_quote_char, secondary_quote_char) = match quote_style {
  Style::Double => ('"', '\''),
  Style::Single => ('\'', '"'),
};
```

cpojer: Nice to see you here ❤️ Thanks, nice destructuring syntax. I cleaned it up.

website/playground/src/lib.rs
```diff
@@ -111,6 +113,7 @@ pub fn run(
     let options = FormatOptions {
         indent_style,
         line_width: line_width.try_into().unwrap_or_default(),
+        quote_style: QuoteStyle::from_str(&quote_style).unwrap_or_default(),
```
ematipico, line 116: Here we can use use `quote_style.try_into()` like we do on line 115

crates/rome_cli/src/commands/format.rs
```diff
@@ -58,6 +58,24 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
         None => {}
     }
 
+    let quote_style = session
+        .args
+        .opt_value_from_str("--quote-style")
```
ematipico, line 63: The help message of the command should be updated with the new option

crates/rome_js_formatter/tests/specs/js/module/string/string.js.snap
ematipico: Because this code is not valid. https://play.rome.tools/?lineWidth=80&indentStyle=tab&indentWidth=2&typescript=false&jsx=false&sourceType=module#aQBtAHAAbwByAHQAIABmAG8AbwAyACAAZgByAG8AbQAgACcAZgBvAG8ALgBqAHMAbwBuACcAIABhAHMAcwBlAHIAdAAgAHsACgAJACcAdAB5AHAAZQAnADoAIAAnAGoAcwBvAG4AJwAsAAoACQB0AHkAcABlADoAIAAiAGgAdABtAGwAIgAsAAoACQAiAHQAeQBwAGUAIgA6ACAAIgBqAHMAIgAsAAoACQB9AA==

Check the errors tab. Broken code won't be formatted

ematipico (changes requested):
That's a great contribution! I would be great to add new test cases to make sure we propertly test the "single quote" case. Please follow this guide to how to add new tests: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/docs/write_tests.md

crates/rome_cli/src/commands/format.rs
cpojer: Just added the comment. Would be awesome to auto-generate the help text somehow so this will always be kept in sync.

website/playground/src/lib.rs
cpojer: I didn't implement TryInto for QuoteStyle though, mainly because indentStyle doesn't have it either and I thought this is probably fine. It's only implemented for the u16 → LineWidth conversion.

Let me know if you'd like me to add it, though.

crates/rome_js_formatter/tests/specs/js/module/string/string.js.snap
cpojer: Ohhh I see. Well, glad to know my code works but also time for me to learn the ES assert syntax ;)

cpojer:
@ematipico thanks for the review! I updated the help text.

Re tests, I added one test (which you commented on) where I copy-pasted some other examples together. Please let me know which other tests you'd like me to add.

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -44,6 +44,16 @@ pub(crate) fn to_format_options(
         );
     }
 
+    let custom_ident_style =
```
MichaReiser, line 47: `custom_quote_style`?

MichaReiser:
> @ematipico thanks for the review! I updated the help text.
> 
> Re tests, I added one test (which you commented on) where I copy-pasted some other examples together. Please let me know which other tests you'd like me to add.
> 
> If you have instructions on how to try out the cli and VSCode extension in development for testing I can test those flows too!

Thanks @cpojer for this amazing contribution! 

You can run the CLI with `cargo run -p rome_cli -- <args>` I defer to @ematipico for how to test the VS code extension.

crates/rome_cli/src/commands/format.rs
```diff
@@ -58,6 +58,24 @@ pub(crate) fn format(mut session: CliSession) -> Result<(), Termination> {
...
+        .map_err(|source| Termination::ParseError {
+            argument: "--quote-style",
+            source,
+        })?;
+
+    match quote_style {
```
leops, line 69: The `indent_style` had to use a match expression in order to inject the number of spaces into the option, but for this one `if let Some(quote_style) = quote_style { options.quote_style = quote_style; }` should be enough

website/playground/src/lib.rs
leops: Since string-to-type conversion is impemented with the `FromStr` trait instead of the generic `TryFrom`, it can be called as a method using `parse` instead of `try_into`: `quote_style.parse().unwrap_or_default()`

ematipico (commented):
At the moment we don't have tests for VSCode, this is something we still need to set up. For now I think it's fine. 

We need to add this option to the [`package.json`](https://github.com/rome/tools/blob/main/editors/vscode/package.json#L84-L102). The name of the option has to be same that has been added to `rome_lsp/src/config.rs`, in camel case, and it should be an enum like `indentStyle`

crates/rome_lsp/src/handlers/formatting.rs
cpojer: Oops, fixing.

website/playground/src/lib.rs
cpojer: Ah, yay `parse` works!

crates/rome_js_formatter/tests/specs/js/module/string/string.js
```diff
@@ -0,0 +1,45 @@
+import hey from "hey"
+import hey from "hey";
+import "x" assert { type: "json" }
+import "foo" assert { "type": "json" };
+import foo from "foo.json" assert { type: "json" };
+import foo from "foo.json" assert {
+
+    type:
+        "json" };
+import foo2 from "foo.json" assert { "type": "json", type: "html", "type": "js" };
```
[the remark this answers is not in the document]
cpojer: I actually think this is a good test to have though because it will ensure that we don't break this invariant of not formatting strings in broken code.

crates/rome_lsp/src/handlers/formatting.rs
```diff
@@ -44,6 +44,16 @@ pub(crate) fn to_format_options(
...
+        QuoteStyle::from_str(workspace_settings.quote_style.as_str()).unwrap_or(QuoteStyle::Double);
```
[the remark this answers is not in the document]
cpojer: Ah good point. I changed that for ident style above as well.

cpojer:
Thanks for all the thorough feedback. I must admit, I haven't written a single line of Rust in more than two years and as you can clearly see I just started writing code on a Saturday night to solve my own problems so I can start adopting the formatter. Really appreciate all the help here, next up is an option for `bracketSpacing`, after which I can finally start using the Rome formatter :)

ematipico (approved):
I think we're good with this option. We will take care of the documentation of the website

cpojer:
Added the option to the playground, see screenshot:
Screen Shot 2022-04-11 at 23 23 48

website/playground/src/App.tsx
```diff
@@ -201,16 +220,25 @@ function App() {
...
+											quoteStyle,
 										})}
+										key={
+											code +
+											lineWidth +
+											indentStyle +
+											indentWidth +
+											language +
+											quoteStyle
+										}
```
cpojer, line 233: `formatWithPrettier` is called on every render, but the `<CodeEditor>` component does not re-render when options change. You can see this by changing one of the dropdowns and you'll see that the prettier window does not change until the code is changed again. Here I'm just concatenating all the variables to make it look like a poor-mans's `useEffect`.

MichaReiser: Nit: Can we exclude `code` because the editor is already updating when the `code` change. This would avoid one full unmount & mount (may be noticeable for large files).

MichaReiser (approved):

ematipico (approved):

PR #2381, opened 2022-04-11
perf(rome_rowan): `SyntaxTriviaPiece.text()`

MichaReiser:
The existing `SyntaxTriviaPiece.text()` method called into `SyntaxTrivia.text_range()` which is expensive because it must iterate over all pieces to get the absolute offset.

The result was that `fourslash/reallyLargeFile` took minutes on my computer to format.

This PR changes the `piece.text` implementation to instead retrieve the text directly from the token because the piece is already aware of its absolute offset in the file, making the operation `O(1)`. 

I don't expect this change to have any perf improvements for the majority of files. It's only relevant for files with a massive number of trivia pieces. 

There are still some places where the implementation calls into `text_range` to get the offsets which always requires iterating over all trivia pieces to get the total length. This happens when calling the `pieces` iterator. We may want to consider caching the total length on the `GreenTrivia`.

This PR further fixes an issue where `trivia.offset()` was not correctly computed for trailing trivia (always initialized to the start of the token)

## Tests

Formatting all TS files in the coverage directory now takes less than 2 seconds.

### DIFF
--- a/crates/rome_rowan/src/green/token.rs
+++ b/crates/rome_rowan/src/green/token.rs
@@ -124,24 +124,6 @@ impl GreenTokenData {
         &text[start..end]
     }
 
-    #[inline]
-    pub fn text_leading_trivia(&self) -> &str {
-        let leading_len = self.leading_trivia().text_len();
-
-        let end: usize = leading_len.into();
-        let text = unsafe { std::str::from_utf8_unchecked(self.data.slice()) };
-        &text[0..end]
-    }
-
-    #[inline]
-    pub fn text_trailing_trivia(&self) -> &str {
-        let (_, trailing_len, total_len) = self.leading_trailing_total_len();
-
-        let start: usize = (total_len - trailing_len).into();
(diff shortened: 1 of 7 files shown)


crates/rome_rowan/src/green/token.rs
```diff
@@ -253,27 +251,6 @@ mod tests {
         assert_eq!(len, GreenTrivia::single_line_comment(len).text_len());
     }
 
-    #[test]
-    fn many_text_len_dont_panic() {
-        let trivia = GreenTrivia::new(vec![
-            TriviaPiece::whitespace(u32::MAX),
-            TriviaPiece::single_line_comment(1),
-        ]);
-        assert_eq!(TextSize::from(u32::MAX), trivia.text_len());
```
MichaReiser, line 262: saturating add is about twice as expensive as a normal add. Considering that calling `text()` on any root node with a length more than 4GB panics, it's probably not worth paying the cost here.

ematipico (approved):

leops (approved):

Issue #1945, opened 2022-01-14
📎 Fuzz Testing for Parser


Boshen:
### Description

After #1909 is done, we should probably look into [Fuzz Testing](https://rust-fuzz.github.io/book/introduction.html) to make the Parser panic resilient.

### Setup
https://github.com/Boshen/tools/commit/c0f33d66b8e2b2f114310e945b64400a061d4526
```
cargo install cargo-fuzz
cd crates/rome_js_parser
cargo fuzz init && cd fuzz
rm fuzz_targets/fuzz_target_1.rs
cargo fuzz add parser
```

Add to `fuzz_targets/parser.rs`:
```rust
#![no_main]
use libfuzzer_sys::fuzz_target;
fuzz_target!(|data: &[u8]| {
    if let Ok(s) = std::str::from_utf8(data) {
        if s.chars().all(|s| !s.is_control()) {
            rome_js_parser::parse_module(s, 0);
        }
    }
});

```

Run `cargo +nightly fuzz run parser -- -only_ascii=1`

### Panics Found
* 2022.01.14: `(:`
* 2022.01.24: `0E+`
* 2022.04.09:
  * `rome_js_parser/src/parser.rs:213`
    * `/=0*_:m/=/*_:|` 
    * `([=[(p[=[(p%]>([=[(p[=[(` 
  * `rome_js_parser/src/parser/rewrite_parser.rs:70`
    * `(--(((--$?:<(`  
    * `  (--   @ (--(-((QQQQQ $`
    * `= 8?0  @ (--(-(  A  S=` 
* 2022.04.11:
    * `rome_js_parser/src/lexer/mod.rs:512:9`
      * `+\\u{A` `+\\u{C`
* 2022.04.14 b0b0b36
  * rome_js_parser/src/syntax/assignment.rs:71:61
    * `(([zAgRvz=[=(e{V{`
    * `({[>[[      =\\c{`

Boshen:
@MichaReiser found 2 crashing points for you to play with 😁 

* 2022.04.09:
  * `/=0*_:m/=/*_:|` crashing point `rome_js_parser/src/parser.rs:213`
  * `([=[(p[=[(p%]>([=[(p[=[(` crashing point `rome_js_parser/src/parser.rs:213`
  * `(--(((--$?:<(`  crashing point `rome_js_parser/src/parser/rewrite_parser.rs:70`
[...]

PR #2382, opened 2022-04-11
fix(rome_js_parser): Regex assignment

MichaReiser:
An assignment to a regex (or any other token that uses re-lexing) may result in a panic if it is the last expression in the file. The source of the problem is that the code wrapping an assignment in an unknown assignment target rewinded the parser and then consumed all tokens up to the same position. However, this results in re-lexing all tokens in "default" mode.

This PR fixes the issue by simply changing the kind of the completed node to `UNKOWN_ASSIGNMENT` rather than using some custom recovery code. The result is more in line with our other error recovery logic that tries to retain valid sub-trees whenever possible.

This fixes the first issue from #1945


MichaReiser:
I need to look into this regression.

ematipico (approved):

Issue #1765, opened 2021-11-10
📎 Detect overlaping sub-uninos


MichaReiser:
### Description

#1745 introduced codegen support for unions over sub-unions. For example, we can now write `JsExprOrSpread = JsExpr | JsSpread` in our grammar and the codegen then knows that `JsExprOrSpread` can be any expression or a spread element. 

The goal is to add a check to the codegen to fail if any two sub-unions of a union type are overlapping because it leads to disambiguity. For example, let's assume we have the following definition:

```
JsThisIsAnIssue = JsExprOrFunctionDeclaration | JsExprOrSpread
JsExprOrFunctionDeclaration = JsExpr | JsFunctionDeclaration
JsExprOrSpread = JsExpr | JsSpread
```

The sub-unions `JsExprOrFunctionDeclaration` and `JsExprOrSpread ` are overlapping because both contain `JsExpr`. 

The problem is that it's now unclear if an expression should be wrapped as a `JsExprOrFunctionDeclaration` or `JsExprOrSpread`. 

You can trigger the code gen by running `cargo xtask syntax` in the project root.

ematipico:
We should explain why in this example there's ambiguity. :) 

I mean, I read `JsExprOrFunctionDeclaration` and `JsExprOrSpread` and I don't know where is the overlapping :) I can guess it by their name... because both nodes will contain `Expr`?

daud9:
Need to learn about this coding pls

loginn:
Hey I'd be down to try to work on this :)

PR #2385, opened 2022-04-11
feat(xtask): Check for sub unions overlap

loginn:
## Summary

Adding a BFS search implementation to check if unions trees contain circular dependencies or reuse types.
The goal is to prevent potential union circular dependencies as well as unions having the same type in multiple branches.

Based on https://github.com/rome/tools/issues/1765

## Test Plan

Here are the test cases I used : 

**These should succeed : **
```
UnionA =
	JsIdentifierAssignment
	| JsStaticMemberAssignment
	| JsComputedMemberAssignment

UnionB =
	JsIdentifierAssignment
	| JsUnknownAssignment
```
```
UnionA =
	JsIdentifierAssignment
	| JsUnknownAssignment

UnionB =
	JsIdentifierAssignment
	| JsUnknownAssignment
```

**These should fail : **

```
UnionA =
	UnionA
	| JsUnknownAssignment
```
```
UnionA =
	UnionB
	| JsUnknownAssignment
UnionB =
	JsxText
	| UnionA
```
```
UnionA =
	UnionB
	| JsUnknownAssignment

UnionB =
	JsxText
	| JsUnknownAssignment
```
```
UnionA =
	UnionB
	| JsUnknownAssignment

UnionB =
	JsxText
	| JsAnyObjectAssignmentPatternMember
``` 
(`JsAnyObjectAssignmentPatternMember` contains `JsUnknownAssignment` )

```
UnionA =
	UnionB
	| JsxSelfClosingElement

UnionB =
	UnionD
	| UnionC

UnionC =
	JsxText
	| JsxExpressionChild

UnionD =
	JsxSelfClosingElement
	| JsAnyObjectAssignmentPatternMember
```
```
UnionA =
	JsxText
	| JsxText
```

### DIFF
--- a/xtask/codegen/js.ungram
+++ b/xtask/codegen/js.ungram
@@ -926,7 +926,6 @@ JsAnyArrayAssignmentPatternElement =
 	| JsAnyAssignmentPattern
 	| JsArrayAssignmentPatternRestElement
 	| JsArrayHole
-	| JsUnknownAssignment
 
 // [a, b, ...rest] = [];
 //        ^^^^^^^
--- a/xtask/codegen/src/ast.rs
+++ b/xtask/codegen/src/ast.rs
@@ -76,7 +76,7 @@ fn check_unions(unions: &[AstEnumSrc]) {
     // Setup a map to find the unions quickly
     let union_map: HashMap<_, _> = unions
         .iter()
-        .map(|en| -> (&str, &AstEnumSrc) { (en.name.as_str(), en) })
+        .map(|en| { (&en.name, en) })
         .collect();
 
     // Iterate over all unions
@@ -85,11 +85,11 @@ fn check_unions(unions: &[AstEnumSrc]) {
             "\n******** START ERROR STACK ********\nChecking {}, variants : {:?}",
             union.name, union.variants
         );
-        let mut union_set: HashSet<&str> = HashSet::from([union.name.as_str()]);
-        let mut union_queue: VecDeque<&str> = VecDeque::new();
+        let mut union_set: HashSet<&String> = HashSet::from([&union.name]);
+        let mut union_queue: VecDeque<&String> = VecDeque::new();
 
         // Init queue for BFS
-        union_queue.extend::<Vec<&str>>(union.variants.iter().map(|x| x.as_str()).collect());
+        union_queue.extend(&union.variants);
 
         // Loop over the queue getting the first variant
         while let Some(variant) = union_queue.pop_front() {
@@ -104,9 +104,7 @@ fn check_unions(unions: &[AstEnumSrc]) {
                 // Try to insert the current variant into the set
                 if union_set.insert(&current_union.name) {
                     // Add all variants into the BFS queue
-                    union_queue.extend::<Vec<&str>>(
-                        current_union.variants.iter().map(|x| x.as_str()).collect(),
-                    );
+                    union_queue.extend(&current_union.variants);
                 } else {
                     // We either have a circular dependency or 2 variants referencing the same type
                     println!("{}", stack_string);
--- a/xtask/codegen/src/ast.rs
+++ b/xtask/codegen/src/ast.rs
@@ -72,34 +72,30 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
     Ok(())
 }
 
-fn check_unions(unions: &Vec<AstEnumSrc>) -> bool {
+fn check_unions(unions: &[AstEnumSrc]) {
     // Setup a map to find the indices of unions quickly
-    let idx_map: HashMap<&String, usize> = unions
+    let idx_map: HashMap<_, _> = unions
             .into_iter()
             .enumerate()
-            .map(|(idx, en)| -> (&String, usize) { (&en.name, idx) })
+            .map(|(idx, en)| -> (&str, usize) { (&en.name.as_str(), idx) })
             .collect();
 
     // Iterate over all unions
     for union in unions {
-        let mut union_set: HashSet<&String> = HashSet::from([&union.name]);
-        let mut union_queue: VecDeque<&String> = VecDeque::new();
-        println!("******** START ********\nChecking {}, variants : {:?}", union.name, union.variants);
+        let mut stack_string = format!("\n******** START ERROR STACK ********\nChecking {}, variants : {:?}", union.name, union.variants);
+        let mut union_set: HashSet<&str> = HashSet::from([union.name.as_str()]);
+        let mut union_queue: VecDeque<&str> = VecDeque::new();
 
         // Init queue for BFS
-        for v in &union.variants {
-            union_queue.push_back(v);
-        }
+        union_queue.extend::<Vec<&str>>(union.variants.iter().map(|x| -> &str { &x.as_str() }).collect());
 
-        while !union_queue.is_empty() {
-            // Get the first variant in the queue
-            // We can unwrap because the queue is never empty at this stage
-            let variant = union_queue.pop_front().unwrap();
+        // Loop over the queue getting the first variant
+        while let Some(variant) = union_queue.pop_front() {
             if idx_map.contains_key(variant) {
                 // The variant is a compound variant
                 // Get the struct from the map
                 let current_union = &unions[idx_map[variant]];
-                println!("SUB-ENUM CHECK : {}, variants : {:?}", current_union.name, current_union.variants);
+                stack_string.push_str(&format!("\nSUB-ENUM CHECK : {}, variants : {:?}", current_union.name, current_union.variants));
                 // Try to insert the current variant into the set
                 if union_set.insert(&current_union.name) {
                     // Add all variants into the BFS queue
@@ -108,31 +104,27 @@ fn check_unions(unions: &Vec<AstEnumSrc>) -> bool {
                     }
                 } else {
                     // We either have a circular dependency or 2 variants referencing the same type
-                    println!("Variant '{variant}' used twice or circular dependency");
-                    return false;
+                    println!("{}", stack_string);
+                    panic!("Variant '{variant}' used twice or circular dependency");
                 }
             } else {
                 // The variant isn't another enum
-                println!("BASE-VAR CHECK : {}", variant);
+                stack_string.push_str(&format!("\nBASE-VAR CHECK : {}", variant));
                 if !union_set.insert(variant) {
                     // The variant already used
-                    println!("Variant '{variant}' used twice");
-                    return false;
+                    println!("{}", stack_string);
+                    panic!("Variant '{variant}' used twice");
                 }
             }
         }
     }
-
-    return true
 }
 
 pub(crate) fn load_js_ast() -> AstSrc {
     let grammar_src = include_str!("../js.ungram");
     let grammar: Grammar = grammar_src.parse().unwrap();
     let ast : AstSrc = make_ast(&grammar);
-    if !check_unions(&ast.unions) {
-        panic!("Invalid Unions") // TODO Graceful error, not sure how it should be handled ?
-    }
+    check_unions(&ast.unions);
     return ast;
 }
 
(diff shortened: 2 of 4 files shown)


loginn:
I left the PR in draft because I still have some debug prints in there and I'm not sure about the error handling when finding an overlapping union.

It also seems like there is an overlap in the current `js.ungram` at line 929

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
     Ok(())
 }
 
+fn check_unions(unions: &Vec<AstEnumSrc>) -> bool {
```
leops, line 75: It's generally better to represent references to vectors as `&[AstEnumSrc]` (and references to strings as `&str`) as these representations are slightly more efficient (internally use a "wide pointer" instead of a "pointer-to-pointer") as well as being somewhat more readable

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+    // Setup a map to find the indices of unions quickly
+    let idx_map: HashMap<&String, usize> = unions
```
leops, line 77: In general the type inference in Rust is good enough that these type annotations can be written only partially (here `HashMap<_, _>`) or omitted entirely (for instance the `-> (&String, usize)` in the map closure)

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+            .collect();
+
+    // Iterate over all unions
+    for union in unions {
+        let mut union_set: HashSet<&String> = HashSet::from([&union.name]);
+        let mut union_queue: VecDeque<&String> = VecDeque::new();
+        println!("******** START ********\nChecking {}, variants : {:?}", union.name, union.variants);
+
+        // Init queue for BFS
+        for v in &union.variants {
```
leops, line 90: This loop could be replaced with `union_queue.extend(&union.variants)`

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+            union_queue.push_back(v);
+        }
+
+        while !union_queue.is_empty() {
+            // Get the first variant in the queue
+            // We can unwrap because the queue is never empty at this stage
+            let variant = union_queue.pop_front().unwrap();
```
leops, line 97: Here the loop could be written as `while let Some(variant) = union_queue.pop_front()` to avoid having to unwrap

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+                if !union_set.insert(variant) {
+                    // The variant already used
+                    println!("Variant '{variant}' used twice");
+                    return false;
+                }
+            }
+        }
+    }
+
+    return true
```
leops, line 126: The explicit return is not needed, you can just write `true` here

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+}
+
 pub(crate) fn load_js_ast() -> AstSrc {
     let grammar_src = include_str!("../js.ungram");
     let grammar: Grammar = grammar_src.parse().unwrap();
-    make_ast(&grammar)
+    let ast : AstSrc = make_ast(&grammar);
+    if !check_unions(&ast.unions) {
+        panic!("Invalid Unions") // TODO Graceful error, not sure how it should be handled ?
```
leops, line 134: I think panicking is the correct way to error here, this is just a utility script after all. In fact is could even panic straight away in the `check_unions` function instead of the `println!` calls that precede the code paths returning false

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,68 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+fn check_unions(unions: &Vec<AstEnumSrc>) -> bool {
+    // Setup a map to find the indices of unions quickly
+    let idx_map: HashMap<&String, usize> = unions
+            .into_iter()
+            .enumerate()
+            .map(|(idx, en)| -> (&String, usize) { (&en.name, idx) })
+            .collect();
+
+    // Iterate over all unions
+    for union in unions {
```
leops, line 84: I never realized that `union` is allowed as an identifier, since `struct` or `enum` are not for instance. I guess `union` being a keyword is only contextual ...

ematipico: Out of curiosity, when `union` is a keyword? I have never seen `union` used as a keyword

leops: Rust has union types for compatibility with C (`union AnyType { integer: u64, float: f64, ptr: *const c_void }`) but they're really only used for FFI since its completely unsafe to either read or write from an union.
Being rarely used plus the fact there's a `union` method on `HashSet` that I think predates the introduction of union types to the language probably explains why `union` is not a full keyword, it didn't really justify a breaking change.

xtask/codegen/src/ast.rs
loginn: Just to confirm, I should use `&str` everywhere instead of `&String` ?

leops: That's preferred at least for places where the type is actually required to be written down (like here in the function signature), but don't feel obligated to go out of your way and force the type to be `&str` everywhere the Rust compiler would have otherwise inferred `&String` if it would make the code unnecessarily harder to read

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,71 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+                stack_string.push_str(&format!(
+                    "\nSUB-ENUM CHECK : {}, variants : {:?}",
+                    current_union.name, current_union.variants
+                ));
+                // Try to insert the current variant into the set
+                if union_set.insert(&current_union.name) {
+                    // Add all variants into the BFS queue
+                    for child in &current_union.variants {
+                        union_queue.push_back(child);
+                    }
```
MichaReiser, line 115: Nit. Can we use `extend` here?

MichaReiser (approved):

xtask/codegen/src/ast.rs
loginn: I made an update but I'm not sure it actually makes the code easier to read or more efficient

xtask/codegen/src/ast.rs
```diff
@@ -71,10 +72,65 @@ pub(crate) fn generate_syntax(ast: AstSrc, mode: &Mode, language_kind: LanguageK
...
+                stack_string.push_str(&format!(
+                    "\nSUB-ENUM CHECK : {}, variants : {:?}",
+                    current_union.name, current_union.variants
+                ));
+                // Try to insert the current variant into the set
+                if union_set.insert(&current_union.name) {
+                    // Add all variants into the BFS queue
+                    union_queue.extend::<Vec<&str>>(
+                        current_union.variants.iter().map(|x| x.as_str()).collect(),
+                    );
```
MichaReiser, line 109: Ah sorry. I wasn't aware that you need to convert from `String` to`&str`. 

You can avoid collecting by using 

```
union_queue.extend(current_union.variants.iter().map(String::as_str))
```

Or go back to your initial versioon.

xtask/codegen/src/ast.rs
leops: Calling extend is preferred to inserting in a loop because it lets the container allocate the required capacity at once instead of having to repeatedly check if reallocation is needed (it's also not necessary to call collect there since extend accepts any iterator as a parameter, this allows skipping allocating an intermediate buffer)

xtask/codegen/src/ast.rs
loginn: I originally wrote it with `Strings` but I switched to `&str` because of another comment from @leops. I'm not sure I understood it fully though.
The code felt more readable to me when using `String` but they suggested `&str` for performance reasons. However I was not sure if it only concerned the function signature or if it was also all usage of `String` inside the function

leops: Yes that was only about the function signature because the type is required to be written down there, when that's not the case it's better to not write any type annotation and just let the type inference of Rust select an appropriate type

MichaReiser:
I just checked out your PR and ran `cargo codegen grammar` and there seems to be one issue with the grammar: 

```
******** START ERROR STACK ********
Checking JsAnyArrayAssignmentPatternElement, variants : ["JsAssignmentWithDefault", "JsAnyAssignmentPattern", "JsArrayAssignmentPatternRestElement", "JsArrayHole", "JsUnknownAssignment"]
BASE-VAR CHECK : JsAssignmentWithDefault
SUB-ENUM CHECK : JsAnyAssignmentPattern, variants : ["JsAnyAssignment", "JsArrayAssignmentPattern", "JsObjectAssignmentPattern"]
BASE-VAR CHECK : JsArrayAssignmentPatternRestElement
BASE-VAR CHECK : JsArrayHole
BASE-VAR CHECK : JsUnknownAssignment
SUB-ENUM CHECK : JsAnyAssignment, variants : ["JsIdentifierAssignment", "JsStaticMemberAssignment", "JsComputedMemberAssignment", "JsParenthesizedAssignment", "TsNonNullAssertionAssignment", "TsAsAssignment", "TsTypeAssertionAssignment", "JsUnknownAssignment"]
BASE-VAR CHECK : JsArrayAssignmentPattern
BASE-VAR CHECK : JsObjectAssignmentPattern
BASE-VAR CHECK : JsIdentifierAssignment
BASE-VAR CHECK : JsStaticMemberAssignment
BASE-VAR CHECK : JsComputedMemberAssignment
BASE-VAR CHECK : JsParenthesizedAssignment
BASE-VAR CHECK : TsNonNullAssertionAssignment
BASE-VAR CHECK : TsAsAssignment
BASE-VAR CHECK : TsTypeAssertionAssignment
BASE-VAR CHECK : JsUnknownAssignment
thread 'main' panicked at 'Variant 'JsUnknownAssignment' used twice', xtask/codegen/src/ast.rs:116:21
```

The fix is to remove `JsUnknownAssignment` from 

```
JsAnyArrayAssignmentPatternElement =
	JsAssignmentWithDefault
	| JsAnyAssignmentPattern
	| JsArrayAssignmentPatternRestElement
	| JsArrayHole
        | JsUnknownAssignment
```

Because it's already part of `JsAnyAssignmentPattern`

We can merge this PR once this is fixed.

xtask/codegen/js.ungram
```diff
@@ -926,7 +926,6 @@ JsAnyArrayAssignmentPatternElement =
 	| JsAnyAssignmentPattern
 	| JsArrayAssignmentPatternRestElement
 	| JsArrayHole
-	| JsUnknownAssignment
```
MichaReiser, line 929: You'll need to run `cargo codegen grammar` and then commit the generated files as well.

loginn: Ooh sorry :)

MichaReiser (approved):
Perfect! Thank you for this contribution that already proved valuable.

xtask/codegen/js.ungram
MichaReiser: Ok, there's one more file that needs updating. Sorry that I didn't mention that before. You need to run `cargo codegen formatter` too and commit the changes.

loginn: Done ! Thanks for your help :)

PR #2387, opened 2022-04-11
chore: Switch to pnpm.

cpojer:
## Summary
pnpm is fast and extremely solid. I've been using it for all my projects recently, and I see adoption growing where previously Yarn/npm were used. I propose switching to it, but I realize it might be bikeshedding so feel free to close if it doesn't align with your plans. Happy to spend more time cleaning dependency things up in follow-up PRs.

*Note: I did not switch `vscode/editor` because `vsce` does not seem to currently support `pnpm` yet. See https://github.com/microsoft/vscode-vsce/issues/421.*

## Test Plan
* Tried out the website → works.
* Tried out the playground → works.
* Tried out the `website/api` folder, works besides not having a database set up.
* `playground/workers-site` seems like a mostly empty template, so probably fine.


ematipico (changes requested):
Thank you for the contribution!

There are few changes that are missing in this PR:
- update the `"scripts"` field of the `playground/package.json` to use `pnpm` instead of `npm`
- update the `"scripts"` field of the `website/package.json` to use `pnpm` instead of `npm`. Here we also use `npm-run-all`, I think we can use this https://pnpm.io/cli/run#--parallel as replacement
- update the `"engines"` field of  `playground/package.json` to use the latest version of `pnpm`. Ideally `6.*`, as 7 is still in pre release

ematipico:
I believe the PR is not up to date with the latest version of the code, that's why there are failing tests.

cpojer:
Sorry for the delay here.

Updaes:
* Rebased.
* Updated with all your recommendations. The engines one definitely makes sense, the other ones previously worked as well but you are right we should only use pnpm in the end so I changed it all.

ematipico (approved):

PR #2388, opened 2022-04-11
chore: Document binary like expression formatting

MichaReiser:
Follow up on #2376


ematipico (approved):
Thank you very much for looking after this. I believe this is one of most complex logic we have built from scratch and I wanted to share as much knowledge as we can about these algorithms. There are developers like to see such comments :)

Mounkoung:
https://github.com/rome/tools/pull/2388#issue-1200196394

PR #2389, opened 2022-04-11
docs(website): update formatter options

### LINKED ISSUE #2362: ☂️ Documentation changes 0.5.0
### Description

- [ ] changes to `rome format` arguments and behaviour, now the default behaviour of the command is printing in the console
- [ ] update changelog with new features and breaking changes (format CLI now doesn't write by default, but it prints. Using `--write` will write to file)
- [ ] update website documentation for `--quote-style` option

ematipico:
## Summary

Closes #2362

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Rome changelog
 
+## 0.5.0
+
+- BREAKING CHANGES: the `format` command doesn't write on disk by default. Now the command prints on terminal.
+
+    **Migration**: add the `--write` argument when calling `rome format`
+    
+    ```shell
+    rome format --write
+    ```
+
+- Added a new option called `--quote-style` to the formatter. This option is also available on VSCode.
+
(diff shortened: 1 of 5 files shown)


cpojer (approved):
Just curious, how did you settle on `--write` over defaulting to writing and having a `--lint` or similar command?

ematipico:
Writing is a disruptive operation, regardless of what we write, and writing a file (especially in non-VCS projects) is always dangerous. We decided that the user should have intent into overwriting files.

This is, for example, the same behaviour that prettier or clippy use. Once we will have VSC support for projects, `--write` can be alias for `--allow-dirty`, which it seems to be the standard option of allowing writing files when the current branch has uncommitted changes.

Rome classic had `--allow-dirty` for some commands, like `rome init`. 

But at the moment we don't have knowledge of a project, so using [what `prettier` does](https://prettier.io/docs/en/cli.html#--write) sounds reasonable.

PR #2391, opened 2022-04-12
fix(rome_js_formatter): Insert Space after `type` for import equals declaration

MichaReiser:
Insert a space after the `type` keyword for import equals declarations.

### DIFF
--- a/crates/rome_js_formatter/tests/specs/prettier/typescript/import-require/type-imports.ts.snap
+++ b/crates/rome_js_formatter/tests/specs/prettier/typescript/import-require/type-imports.ts.snap
@@ -1,8 +1,6 @@
 ---
 source: crates/rome_js_formatter/tests/prettier_tests.rs
-assertion_line: 119
 expression: type-imports.ts
-
 ---
 # Input
 ```js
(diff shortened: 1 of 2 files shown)


crates/rome_js_formatter/tests/specs/prettier/typescript/import-require/type-imports.ts.snap
```diff
@@ -29,16 +27,16 @@ b = require("b");
...
-import typeA = require("A");
+import type A = require("A");
 
-import typea = require("a");
+import type a = require("a");
 
-export import typeB = require("B");
+export import type B = require("B");
 
-export import typeb = require("b");
```
ematipico, line 41: @MichaReiser  Does this mean that this is valid syntax??

MichaReiser:
```
export import type B = require("B");
```

This is valid, yes :D

leops (approved):

crates/rome_js_formatter/tests/specs/prettier/typescript/import-require/type-imports.ts.snap
ematipico: No no, I mean, is this `export import typeb = require("b");` valid?

Because this is really strange. Our testing suite makes sure that emitted code is actually valid (it parses again) and it seems that it's not doing so... so, there are two possible explanation:
- the formatter testing suite is not catching this case (it doesn't reparse) [likely]
- the parse doesn't emit any syntax errors [unlikely]

ematipico: (yeah valid syntax lol, an unfortunate coincidence lol)

MichaReiser: Yeah, it's valid because `typeb` is a valid identifier...

PR #2399, opened 2022-04-13
Chore: update submodules

MichaReiser:
Update the test262, babel, and TypeScript submodules.

The TypeScript regressions are related to the newly added Syntax in TS 4.7. I created #2400 to track the new syntax features.

### DIFF
--- a/.github/workflows/parser_conformance.yml
+++ b/.github/workflows/parser_conformance.yml
@@ -5,9 +5,10 @@ on:
     branches:
       - main
     paths:
-      - 'crates/rome_js**/'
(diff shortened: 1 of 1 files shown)


.github/workflows/parser_conformance.yml
```diff
@@ -7,7 +7,8 @@ on:
     paths:
       - 'crates/rome_js**/'
       - 'crates/rome_rowan'
```
ematipico, line 9: Should we add `/*` to `rome_rowan` too?

ematipico (approved):

PR #2432, opened 2022-04-13
feature(playground): Mobile playground

NicholasLYang:
## Summary

We didn't have an explicit mobile view of the playground. That made it pretty hard to use because the tile layout couldn't fit into the tiny viewport of a phone.

This PR creates an explicit mobile view that puts the settings and input tiles under a tab. That way users can navigate the playground pretty easily. I also refactored the code a little, pulling out the PlaygroundSettings component and adding a utils module.

## Test Plan

Screen Shot 2022-04-16 at 3 56 41 PM
Screen Shot 2022-04-16 at 3 56 54 PM
Screen Shot 2022-04-16 at 3 57 00 PM

### DIFF
--- a/website/playground/src/utils.ts
+++ b/website/playground/src/utils.ts
@@ -1,0 +1,122 @@
+// Define general type for useWindowSize hook, which includes width and height
+import { useEffect, useState } from "react";
+import prettier from "prettier";
+// @ts-ignore
+import parserBabel from "prettier/esm/parser-babel";
+import { IndentStyle, QuoteStyle } from "./types";
+
+interface Size { width: number | undefined, height: number | undefined }
+
+// Hook
+export function useWindowSize(): Size {
+	// Initialize state with undefined width/height so server and client renders match
+	// Learn more here: https://joshwcomeau.com/react/the-perils-of-rehydration/
+	const [windowSize, setWindowSize] = useState<Size>({
+		width: undefined,
+		height: undefined,
+	});
+	useEffect(
+		() => {
+			// Handler to call on window resize
+			function handleResize() {
+				// Set window width/height to state
+				setWindowSize({ width: window.innerWidth, height: window.innerHeight });
+			}
+			// Add event listener
+			window.addEventListener("resize", handleResize);
+			// Call handler right away so state gets updated with initial window size
+			handleResize();
+			// Remove event listener on cleanup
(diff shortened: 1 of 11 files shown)


MichaReiser:
Can you share a picture of your changes. It's hard to imagine how the playground looks after your changes (ideally on Mobile and Desktop)

ematipico:
It would be great to rebase this branch after https://github.com/rome/tools/pull/2433 so we can have a preview

ematipico:
Not sure why, but the deployed playground doesn't load. From the web console:

```
Loading module from “https://play.rome.tools/assets/index.00b713e5.js” was blocked because of a disallowed MIME type (“text/html”).
```

NicholasLYang:
Yeah...I don't know why but our base url parameter doesn't propagate anymore. No clue why

ematipico:
Could you please rebase your PR with the new code? I am running your branch locally and it doesn't have the new style of the options, which are now aligned horizontally.

Screenshot 2022-04-14 at 17 49 03
s

NicholasLYang:
I purposefully reverted that because it wasn't rendering properly. I can fix it in this PR I suppose.
Screen Shot 2022-04-14 at 2 20 35 PM

ematipico:
It works on Firefox. Weird. What's better: fixing it or revert it?

NicholasLYang:
Fixed it :D

MichaReiser:
Please add a screenshot for the mobile view and on desktop to your test plan.

NicholasLYang:
Done. You can also check it out at the link above

website/playground/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+		setQuoteStyle,
+		indentWidth,
+		setIndentWidth,
+		isTypeScript,
+		setIsTypeScript,
+		isJsx,
+		setIsJsx,
+		sourceType,
+		setSourceType,
+	};
```
MichaReiser, line 101: Nit: You could safe yourself some typing and simplify the "has the playground state changed" logic by instead only using  a single state to store all variables and offer a single `setPlaygroundState(newState: Partial<State>)` function instead. 

OR, use `useReducer` and have different actions to dispatch the different updates.

MichaReiser (approved):

PR #2433, opened 2022-04-13
ci(playground): Trying, yet again, to fix deploy_playground workflow trigger

NicholasLYang:
## Summary
Finally, finally getting the workflow to work. I removed the paths requirement entirely because it allowed me to test the workflow on this PR itself, and because I'm not really sure what was preventing the workflow from being run with other PRs.

### DIFF
--- a/.github/workflows/deploy_playground.yml
+++ b/.github/workflows/deploy_playground.yml
@@ -31,6 +31,7 @@ jobs:
           run_install: |
             args: [--prefix website/playground]
 
+      # We use the short SHA for shorter, simpler links
       - name: Set GitHub SHA
         id: sha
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_playground.yml
```diff
@@ -34,21 +30,26 @@ jobs:
           version: 6
           run_install: |
             args: [--prefix website/playground]
-      - run: pnpm build --prefix website/playground -- --base=/$GITHUB_SHA/
+
+      - name: Set GitHub SHA
```
MichaReiser, line 34: What is this for? And why prefer it over `$GITHUB_SHA`. Might be worth a comment for whoever touches this file in the future.

MichaReiser (approved):

PR #2434, opened 2022-04-14
feat(playground): align options horizontally

jer3m01:
## Summary

The options at the top of the playground take up a lot of space and are pretty empty. This PR aligns the options horizontally instead of vertically to give more space to the code.
image
Into:
image

## Test Plan

Tested on local playground.

PS: Awesome work team! Congrats 👏 
Really need to get back to contributing!


MichaReiser:
Awesome! Thanks for this contribution. That makes much better use of today's wide screen display.

PR #2435, opened 2022-04-14
fix(rome_js_parser): Rewind at EOF

MichaReiser:
The Buffered Lexer tested if the Lexer's current position is the same as the position of the checkpoint to avoid rewinding in that case (mainly to keep the lookahead).

However, only comparing the positions (that point to the end index of the last token) doesn't work if the snapshot was taken right before the EOF token because the EOF token has a width of 0 (and thus, doesn't move the position).

A possible fix would be to compare the `start_position` instead but this isn't safe either in case the token has been re-lexed and now has a different kind.

This PR removes the "optimisation" all together and the benchmarks show no significant change in performance.

```
group                                 fix                                    main
-----                                 ---                                    ----
parser/checker.ts                     1.01     61.6±0.67ms    42.2 MB/sec    1.00     61.3±1.35ms    42.4 MB/sec
parser/compiler.js                    1.00     40.2±0.78ms    26.1 MB/sec    1.01     40.7±0.32ms    25.8 MB/sec
parser/d3.min.js                      1.00     23.3±0.13ms    11.2 MB/sec    1.02     23.9±0.10ms    11.0 MB/sec
parser/dojo.js                        1.02      2.2±0.01ms    30.9 MB/sec    1.00      2.2±0.00ms    31.6 MB/sec
parser/ios.d.ts                       1.00     51.4±0.37ms    36.3 MB/sec    1.08     55.4±0.55ms    33.7 MB/sec
parser/jquery.min.js                  1.00      6.5±0.11ms    12.8 MB/sec    1.03      6.7±0.01ms    12.4 MB/sec
parser/math.js                        1.01     44.6±0.27ms    14.5 MB/sec    1.00     44.2±0.26ms    14.7 MB/sec
parser/parser.ts                      1.00  1519.4±36.28µs    31.8 MB/sec    1.02  1553.7±22.59µs    31.1 MB/sec
parser/pixi.min.js                    1.00     27.3±0.17ms    16.1 MB/sec    1.04     28.3±0.21ms    15.5 MB/sec
parser/react-dom.production.min.js    1.00      8.7±0.03ms    13.2 MB/sec    1.04      9.1±0.02ms    12.7 MB/sec
parser/react.production.min.js        1.00    450.7±2.66µs    13.6 MB/sec    1.02    457.6±2.10µs    13.4 MB/sec
parser/router.ts                      1.00   1212.2±4.64µs    49.3 MB/sec    1.01   1223.5±8.01µs    48.9 MB/sec
parser/tex-chtml-full.js              1.00     59.0±0.56ms    15.5 MB/sec    1.00     58.9±0.55ms    15.5 MB/sec
parser/three.min.js                   1.00     31.0±0.08ms    19.0 MB/sec    1.02     31.5±0.12ms    18.6 MB/sec
parser/typescript.js                  1.00    268.7±6.36ms    35.4 MB/sec    1.05   283.5±21.07ms    33.5 MB/sec
parser/vue.global.prod.js             1.01     11.1±0.02ms    10.9 MB/sec    1.00     11.0±0.03ms    11.0 MB/sec
```

## Tests

Verified that this PR fixes the two new reported fuzzing issues #1945

```
(([zAgRvz=[=(e{V{
({[>[[ =\\c{
```


MichaReiser:
!bench_parser

leops (approved):
Could the fuzzing test case be added to the test suite of the parser to check for regressions like quickcheck does ?

ematipico (approved):

MichaReiser:
> Could the fuzzing test case be added to the test suite of the parser to check for regressions like quickcheck does ?

I added one of the test cases since both are triggering the same exception.

PR #2438, opened 2022-04-14
fix(rome_js_parser): fix lexer panicking on unterminated unicode escape at EOF

Boshen:
## Summary

Found via fuzzing https://github.com/rome/tools/issues/1945

It panics on unterminated unicode scape at EOF. e.g. `+\\u{A`

## Test Plan
Added a unit test.

### DIFF
--- a/crates/rome_js_parser/src/lexer/mod.rs
+++ b/crates/rome_js_parser/src/lexer/mod.rs
@@ -508,7 +508,7 @@ impl<'src> Lexer<'src> {
     /// Get the UTF8 char which starts at the current byte
     /// Safety: Must be called at the begining of a UTF8 char.
     fn current_char_unchecked(&self) -> char {
-        // This is unreachable for all intents and purposes, but this is just a precautionary measure
+        // Precautionary measure for making sure the unsafe code below does not read over memory boundary
         debug_assert!(!self.is_eof());
 
         // Safety: We know this is safe because we require the input to the lexer to be valid utf8 and we always call this when we are at a char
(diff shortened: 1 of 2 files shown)


crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -628,7 +628,14 @@ impl<'src> Lexer<'src> {
         let start = self.position + 1;
         self.read_hexnumber();
 
-        if self.current_byte() != Some(b'}') {
+        let current_byte = self.current_byte();
+
+        // Abort on EOF
+        if current_byte.is_none() {
+            return Err(());
```
MichaReiser, line 635: Isn't it necessary to emit a diagnostic in this case?

crates/rome_js_parser/src/lexer/tests.rs
```diff
@@ -1028,6 +1028,15 @@ fn fuzz_fail_6() {
     }
 }
 
+#[test]
+fn fuzz_fail_7() {
```
xunilrj, line 1032: Given that we have a nice semantic for this test, should we name it ```err_on_unterminated_unicode```?

crates/rome_js_parser/src/lexer/mod.rs
Boshen: This function is `fn read_codepoint_escape(&mut self) -> Result<char, ()>`. Diagnostic is emitted at the call site with a `Result::is_ok()` check, hence it's not necessary here.

crates/rome_js_parser/src/lexer/tests.rs
Boshen: renamed.

Boshen:
Both reviews addressed, ready to merge.

xunilrj (approved):

PR #2452, opened 2022-04-14
fix(playground): base url not propagated to vite

NicholasLYang:
## Summary

Turns out we don't propagate the base url parameter properly with nested NPM scripts

### DIFF
--- a/.github/workflows/deploy_playground.yml
+++ b/.github/workflows/deploy_playground.yml
@@ -36,7 +36,11 @@ jobs:
         id: sha
         run: echo "::set-output name=sha::$(git rev-parse --short HEAD)"
 
-      - run: pnpm build --prefix website/playground -- --base=/${{ steps.sha.outputs.sha }}/
+      - run: pnpm --prefix website/playground build:wasm
+
+      # We don't run the general build script because we need to pass in the base flag
+      # directly to the vite command.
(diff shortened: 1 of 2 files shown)


ematipico:
Is it now possible to contain the playground deploy only when the actual playground changes?

NicholasLYang:
I can try adding that back, although we do want to deploy when the code changes too, because the whole point is that you can play around with people's changes to the code with the link to the commit. So really, the only changes where we wouldn't want to deploy would be small changes to non-code or to the website.

ematipico:
Oh yeah good point, I haven't though about it! It makes totally sense then!

.github/workflows/deploy_playground.yml
```diff
@@ -36,7 +36,11 @@ jobs:
         id: sha
         run: echo "::set-output name=sha::$(git rev-parse --short HEAD)"
 
-      - run: pnpm build --prefix website/playground -- --base=/${{ steps.sha.outputs.sha }}/
+      - run: pnpm --prefix website/playground build:wasm
+
+      # We don't run the general build script because we need to pass in the base flag
+      # directly to the vite command.
+      - run: pnpm --prefix website/playground build:js --base=/${{ steps.sha.outputs.sha }}/
```
MichaReiser, line 43: https://stackoverflow.com/a/51594722

Seems like another option but is probably more obscure than your fix.

MichaReiser (approved):

PR #2453, opened 2022-04-14
fix(rome_js_formatter): group and break sequence expressions

### LINKED ISSUE #2413: Long sequence expressions

ematipico:
## Summary

Fixes #2413

The algorithm is borrowed from the binary like expression with a simple recursion where we flat the chain of sequence expressions. 

We pass the comma token to the next node which will be in charge of formatting it.

## Test Plan

Added new tests that cover short and long expressions.

There's a prettier test that still doesn't look like the original because now the grouping affects subsequent breaking. That is issue is already tracked in another issue, so it's not part of this PR goal.


crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,25 +1,98 @@
...
-        ])
+        format_sequence_expression(self, formatter)
+    }
+}
+
+#[derive(Debug)]
+enum FlattenItem {
+    SequenceExpression(JsSequenceExpression, FormatElement),
+    OtherNode(JsSyntaxNode, FormatElement),
+}
```
MichaReiser, line 19: Do we need to go over `FlattenItem` here or can we just create the correct `FormatElement` instead? Because you aren't using the information if the node is a nested sequence expression or not inside of the `into_format_element` function. 

The reason why it would be better to construct the `FormatElement` right away is because this avoids the allocation of another vector for `FormatItems`.

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,25 +1,98 @@
...
+        formatter,
+        None,
+    )?;
+
+    Ok(group_elements(flatten_item_list.into_format_element()))
+}
+
+fn flat_sequence_expression(
+    list: &mut FlattenItemList,
+    node: JsSyntaxNode,
```
MichaReiser, line 71: I would change this to only accept a `JsSequenceExpression` and handle the non sequence expression when trying to format the left hand side.

crates/rome_js_formatter/tests/specs/js/module/expression/sequence_expression.js
```diff
@@ -1 +1,6 @@
 a,b
+
+
+abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsfddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd;
+
+abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf, abcdefghijklkmnopqrstufwkdsf;
```
MichaReiser, line 6: This doesn't match prettiers formatting

```
abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf,
	abcdefghijklkmnopqrstufwkdsf;
```

crates/rome_js_formatter/tests/specs/prettier/js/sequence-break/break.js.snap
```diff
@@ -25,22 +25,35 @@ for (aLongIdentifierName = 0, aLongIdentifierName = 0, aLongIdentifierName = 0,
 # Output
 ```js
 const f = (argument1, argument2, argument3) => (
-  doSomethingWithArgument(argument1), doSomethingWithArgument(argument2), argument1
+  doSomethingWithArgument(argument1),
+    doSomethingWithArgument(argument2),
+    argument1
```
MichaReiser, line 30: This doesn't match prettier's formatting

```
const f = (argument1, argument2, argument3) => (
	doSomethingWithArgument(argument1),
	doSomethingWithArgument(argument2),
	argument1
);
```

MichaReiser (changes requested):
Please add the test cases from the issue and paste the expected output from prettier into the snapshot to ensure Rome produces the exact same formatting as prettier.

crates/rome_js_formatter/tests/specs/js/module/expression/sequence_expression.js
ematipico: This is not the snapshot, so I think you're something else.

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
ematipico: Could you provide an explanation please?

MichaReiser: You do 

```
fn flat_sequence_expression(
    list: &mut FlattenItemList,
    node: JsSequenceExpression,
)
```

You can then remove the inner match, but you'll have to change 

```
let left = sequence_expression.left()?;
let right = sequence_expression.right()?;
let comma = sequence_expression.comma_token()?;
flat_sequence_expression(list, left.syntax().clone(), formatter, Some(comma))?;
let formatted =
    format_elements![space_token(), right.format(formatter)?, formatted_comma];
list.push(FlattenItem::SequenceExpression(
    sequence_expression,
    formatted,
))
```

to 

```
let left = sequence_expression.left()?;
let right = sequence_expression.right()?;
let comma = sequence_expression.comma_token()?;

if let Some(JsSequenceExpression(left)) = JsSequenceExpression::cast(left) {
   flat_sequence_expression(list, left, formatter, Some(comma))?;  
} else {
  let formatted = format_elements![left.to_format_element(formatter)?, formatted_comma];
  list.push(FlattenItem::OtherNode(left.syntax(), formatted))
}

flat_sequence_expression(list, left.syntax().clone(), formatter, Some(comma))?;
let formatted =
    format_elements![space_token(), right.format(formatter)?, formatted_comma];
list.push(FlattenItem::SequenceExpression(
    sequence_expression,
    formatted,
))
```

ematipico: I didn't want to know how, I wanted to know why. To me they seems two ways to achieve the same thing

MichaReiser: It encodes the constraint that `flat_sequence_expression` only gets called on `JsSequenceExpression`. I find it a bit  confusing that the method can be called with any expression even though it's called `flat_sequence_expression`. 
This helps the reader to understand the code because it encodes more information

It also avoids one additional function call if the left hand side isn't a `JsSequenceExpression`, making it easier to reason about the recursion (and potentially replacing it wth an iterator). 

It also simplifies the method's code because you no longer need that match (that is kind of expensive if we already know it's a `JsSequenceExpression).

ematipico:
Closed in favour of https://github.com/rome/tools/pull/2541

The algorithm changed a lot so it was best to start from scratch.  I addressed also @MichaReiser 's feedback

Issue #2351, opened 2022-04-05
JSON Parser


ematipico:
### Description

We are going to use this issue to track the various work around JSON. We don't plan, for the time being, to support super languages, like JSONC, which is the format used by TypeScript.

- [x] create a `json.ungram` file where we will slowly implement new grammars
- [x] create a `rome_json_syntax` crate, where we will have all things related to the JSON syntax, generated code too
- [x] expand the codegen of the grammar to include JSON too
- [ ] evaluate how to implement the parser, lexer, etc. (it will require a separate discussion)

loginn:
Hey so I wrote a Json parser in rust not too long ago so I could maybe look into some of that if you'd like ?

PR #2455, opened 2022-04-15
Feature/json ungram

loginn:
## Summary
This PR is part of https://github.com/rome/tools/issues/2351

The goal is to support JSON. This PR creates a json.ungram file as well as makes some changes to load the ast and generate the code in a new crate called `rome_json_syntax`)

## Test Plan

Not really sure how to test. The grammar should cover all of JSON and codegen should run without any issue. Looking for advice on how to test everything


xtask/codegen/json.ungram
```diff
@@ -0,0 +1,69 @@
...
+
+
+SyntaxElement = SyntaxElement
+
+JsonUnknown = SyntaxElement*
+
+// {} or []
+JsonRoot =
+  JsonObjectStatement
+  | JsonArrayStatement
```
MichaReiser, line 29: Thanks for working on this. I would probably take inspiration from the JSON spec on how to name things. https://www.json.org/json-en.html That's why I would use the wording `Value` rather than `Statement` because `Json` doesn't really know a statement concept. 

The spec is mainly built around values. A value can be anything from `object`, `array`, to a `string`. That's why we should define a `JsonAnyValue` union that includes all the different value types + a `JsonUnknownValue` for better error recovery. 

The root (or a `JsonDocument`) is then any value plus the EOF token `JsonDocument = value: JsonAnyValue EOF

loginn: But values in JSON can be of a type that is not a valid start for a json file like `"string"`. A valid json has to start with a `JsonArray` or a `JsonObject`
I think this is a better approach : 
```
JsonDocument =
  value: JsonObject
  | JsonArray
  | JsonUnknown
  eof: 'EOF'
```

MichaReiser: A JSON file can start with any value as mentioned by the spec. I normally play around with existing parsers to validate my understanding of the spec. One just needs to keep in mind that these parsers are not always 100% spec compliant (not even babel or TypeScript). Here an example of a document only containing a string:  https://astexplorer.net/#/gist/ed6d345fe75ef1bf0a78dc805b54e112/4c58db547cd2688271119cf0872636ee3743e42f

loginn: Oh I hadnt realised that any value was a valid json. Thanks !

MichaReiser:
Thank you @loginn for your contribution! Your work has been picked up in #2466 and is now merged!

MichaReiser:
No worries. Your work helped a lot to get #2466 bootstrapped and I'm sorry that two people ended working on this.

PR #2456, opened 2022-04-15
perf(rome_js_formatter): Reduce `FormatElement` size

MichaReiser:
This PR reduces the size of a `FormatElement` from 56 to 32 bytes by:

 * Using a `Box<str>` to store a verbatim string as it doesn't need to be mutable (safes 8 bytes for the `capacity`)
 * Only storing the start position for a `DynamicToken` because the length isn't used by the printer
 *


crates/rome_formatter/src/format_element.rs
```diff
@@ -823,46 +823,28 @@ where
...
-        // Count the lines separating the two statements,
-        // starting with the trailing trivia of the previous node
-        let mut line_count = prev_node
-            .last_trailing_trivia()
-            .and_then(|prev_token| {
-                // Newline pieces can only come last in trailing trivias, skip to it directly
-                prev_token.pieces().next_back()?.as_newline()
-            })
-            .is_some() as usize;
-
```
MichaReiser, line 843: @leops Am I missing a case where the trailing trivia contains new lines (other than for the EOF token which isn't relevant here)? 

According to our definition, just checking the leading should be sufficient

> *    A token owns all of its trailing trivia up to, but not including, the next newline character.
> *  Looking backward in the text, a token owns all of the leading trivia up to and including the first newline character.

leops: I'm about 80% sure this code doesn't do anything and I added it "just in case" when I experimented with introducing the logic for preserving empty lines

crates/rome_formatter/src/format_element.rs
```diff
@@ -823,46 +823,28 @@ where
...
-        // Then add the newlines in the leading trivia of the next node
+    fn get_lines_before<L: Language>(next_node: &SyntaxNode<L>) -> usize {
```
ematipico, line 826: While we are at it, could you please make this function public? We will need it for other cases where `format_list` doesn't fit certain cases.

crates/rome_formatter/src/format_element.rs
```diff
@@ -823,46 +823,28 @@ where
...
-                    // will handle newlines between the comment and the node
-                    break;
-                }
-            }
+            leading_trivia
+                .pieces()
+                .take_while(|piece|
+                // Stop at the first comment piece, the comment printer
+                // will handle newlines between the comment and the node
+                !piece.is_comments())
```
ematipico, line 834: nit: could you enclose comment and code in a block `{` `}` please? It might help better the readability of the comment

crates/rome_formatter/src/lib.rs
```diff
@@ -259,12 +259,16 @@ impl Formatted {
...
-        &self.verbatim_source
+    /// The text in the formatted code that has been formatted as verbatim.
+    pub fn verbatim(&self) -> impl Iterator<Item = (TextRange, &str)> {
+        self.verbatim_ranges
+            .iter()
+            .map(|range| (*range, &self.code[*range]))
     }
 
-    pub fn into_verbatim(self) -> Vec<(String, TextRange)> {
-        self.verbatim_source
```
ematipico, line 267: Any particular reason why we removed this function? It was there mainly to be in line with `into_sourcemap` and `into_code`. If it's been removed because not used, then we should do the same with `into_sourcemap` too.

crates/rome_formatter/src/format_element.rs
MichaReiser: Let's leave that to the PRs that need it. We could even move this to rowan.

crates/rome_formatter/src/lib.rs
MichaReiser: I removed it because it wasn't used and the struct no longer stores the `Vec` that it could just take and return. Having the iterator specific method makes it clear that taking the verbatim sources requires a vector allocation.

crates/rome_formatter/src/format_element.rs
MichaReiser: Much better, thanks!

ematipico (approved):

Issue #2408, opened 2022-04-13
Object literal & type line break after `{`


MichaReiser:
Rome formatter tries to fit all object or object literal type members on a single line. Prettier does the same, but formats each member on a single line if the object literal or object literal type has a line break between the opening `{` and the first member. 

## Input
```ts
// Print each member on a new line if there's a line break between `{` and the first member
({ 
a: "test",
b: "test" });


type Test = { 
a: string };

// Fit on single line if line break is after first member
type Test2 = { a: string
};

({ a: "test",
b: "test" });
```

## Prettier

```ts
// Print each member on a new line if there's a line break between `{` and the first member
({
	a: "test",
	b: "test",
});

type Test = {
	a: string;
};

// Fit on single line if line break is after first member
type Test2 = { a: string };
```

## Rome
```ts
// Print each member on a new line if there's a line break between `{` and the first member
({ a: "test", b: "test" });

type Test = { a: string };

// Fit on single line if line break is after first member
type Test2 = { a: string };

({ a: "test", b: "test" });
```

## Expected

Rome to match Prettiers formatting. Format object literal and object-literal type members each on their own line if the first member has a leading line break.

Issue #2421, opened 2022-04-13
Call expression arguments formatting


MichaReiser:
Prettier has sophisticated formatting for call arguments that prevents unnecessary line breaks. For example, it formats the following snipped without breaking the passed arrow function over multiple lines (except the body). 

```
// Grouped if less than 3 arguments
testMethod(() => {
  expected(() =>
    moduleMocker.generateFromMetadata(moduleMocker.getMetadata(/a/))
  ).not.toThrow();
}, "string");

// Prettier:
// Grouped if less than 3 arguments
testMethod(() => {
  expected(() => moduleMocker.generateFromMetadata(moduleMocker.getMetadata(/a/))).not.toThrow();
}, "string");
```

Relevant test cases: 
* [`arrow/call.js`](https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsarrow-callarrow_calljs)
* [`break-call/react.js`](https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsbreak-callsreactjs)
* [`last-argument/expensions/empty-lines.js`](https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jslast-argument-expansionempty-linesjs)
* [`preserve-line/argument-list.js`](https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jspreserve-lineargument-listjs)
* [`js/functional-composition/functional_compose.js`](https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsfunctional-compositionfunctional_composejs)

## Expected
Output matches Prettier's output (or closely)

ematipico:
We need to study prettier' source and see what they are doing. They use [`Alternatives`](https://github.com/rome/tools/pull/2534) for their solution. So probably we should wait until we have a better IR that matches prettier's.

ematipico:
I am going to a look at this now

PR #2458, opened 2022-04-17
feat(rome_js_formatter): Hard line breaks in object expressions.

cpojer:
## Summary
Fixes https://github.com/rome/tools/issues/2408.

This is the initial draft for respecting a newline in an object expression (or import specifier) in between an opening curly brace `{` and the first item in the member list. This should make Rome match prettier in this situation. Again, I have no idea what I'm doing :) 

**Open questions to get this over the line:**
* For some reason `formatter::js_module::specs::js::module::arrow::arrow_nested` fails and I'm unsure why. Please help!
* See some inline questions.

## Test Plan

* Added tests.

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/object_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/object_expression.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-
+use crate::utils::has_leading_newline;
 use crate::FormatNodeFields;
 use rome_js_syntax::JsObjectExpression;
 use rome_js_syntax::JsObjectExpressionFields;
@@ -15,12 +15,18 @@ impl FormatNodeFields<JsObjectExpression> for FormatNodeRule<JsObjectExpression>
             r_curly_token,
         } = node.as_fields();
 
+        let has_newline = has_leading_newline(members.syntax());
         let members_content = formatted![formatter, [members.format()]]?;
 
         if members.is_empty() {
             formatter
                 .delimited(&l_curly_token?, members_content, &r_curly_token?)
                 .soft_block_indent()
+                .finish()
+        } else if has_newline {
+            formatter
+                .delimited(&l_curly_token?, members_content, &r_curly_token?)
+                .block_indent()
                 .finish()
         } else {
             formatter
--- a/crates/rome_js_formatter/src/utils/mod.rs
+++ b/crates/rome_js_formatter/src/utils/mod.rs
@@ -105,6 +105,18 @@ pub(crate) fn has_formatter_trivia(node: &JsSyntaxNode) -> bool {
     false
 }
 
+/// Returns true if this node contains newlines in trivias.
+pub(crate) fn has_leading_newline(node: &JsSyntaxNode) -> bool {
+    if let Some(leading_trivia) = node.first_leading_trivia() {
+        for piece in leading_trivia.pieces() {
+            if piece.is_newline() {
+                return true;
+            }
+        }
+    }
+    false
+}
+
 /// Format an element with a single line head and a body that might
 /// be either a block or a single statement
 ///
--- a/crates/rome_js_formatter/tests/specs/js/module/object/getter_setter.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/object/getter_setter.js.snap
@@ -18,6 +18,10 @@ Indent style: Tab
 Line width: 80
 Quote style: Double Quotes
 -----
-let a = { get foo() {} };
-let b = { set foo(a) {} };
+let a = {
+	get foo() {},
+};
+let b = {
+	set foo(a) {},
+};
 
--- a/crates/rome_js_formatter/src/js/module/export_named_from_clause.rs
+++ b/crates/rome_js_formatter/src/js/module/export_named_from_clause.rs
@@ -1,10 +1,11 @@
 use crate::prelude::*;
-
 use crate::utils::format_with_semicolon;
+use crate::utils::has_leading_newline;
 
 use crate::FormatNodeFields;
 use rome_js_syntax::JsExportNamedFromClause;
 use rome_js_syntax::JsExportNamedFromClauseFields;
+use rome_rowan::AstNode;
 
 impl FormatNodeFields<JsExportNamedFromClause> for FormatNodeRule<JsExportNamedFromClause> {
     fn format_fields(
@@ -22,14 +23,25 @@ impl FormatNodeFields<JsExportNamedFromClause> for FormatNodeRule<JsExportNamedF
             semicolon_token,
         } = node.as_fields();
 
-        let list = formatter
-            .delimited(
-                &l_curly_token?,
-                formatted![formatter, [specifiers.format()]]?,
-                &r_curly_token?,
-            )
-            .soft_block_spaces()
-            .finish()?;
+        let list = if has_leading_newline(specifiers.syntax()) {
+            formatter
+                .delimited(
+                    &l_curly_token?,
+                    formatted![formatter, [specifiers.format()]]?,
+                    &r_curly_token?,
+                )
+                .block_indent()
+                .finish()?
+        } else {
+            formatter
+                .delimited(
+                    &l_curly_token?,
+                    formatted![formatter, [specifiers.format()]]?,
+                    &r_curly_token?,
+                )
+                .soft_block_spaces()
+                .finish()?
+        };
 
         format_with_semicolon(
             formatter,
--- a/crates/rome_js_formatter/src/ts/declarations/enum_declaration.rs
+++ b/crates/rome_js_formatter/src/ts/declarations/enum_declaration.rs
@@ -1,6 +1,8 @@
 use crate::prelude::*;
+use crate::utils::has_leading_newline;
 use crate::FormatNodeFields;
 use rome_js_syntax::{TsEnumDeclaration, TsEnumDeclarationFields};
+use rome_rowan::AstNode;
 
 impl FormatNodeFields<TsEnumDeclaration> for FormatNodeRule<TsEnumDeclaration> {
     fn format_fields(
@@ -16,11 +18,16 @@ impl FormatNodeFields<TsEnumDeclaration> for FormatNodeRule<TsEnumDeclaration> {
             r_curly_token,
         } = node.as_fields();
 
+        let has_newline = has_leading_newline(members.syntax());
         let list = formatter
             .delimited(
                 &l_curly_token?,
                 join_elements(
-                    soft_line_break_or_space(),
+                    if has_newline {
+                        hard_line_break()
+                    } else {
+                        soft_line_break_or_space()
+                    },
                     formatter.format_separated(&members, || token(","))?,
                 ),
                 &r_curly_token?,
--- a/crates/rome_js_formatter/src/ts/types/object_type.rs
+++ b/crates/rome_js_formatter/src/ts/types/object_type.rs
@@ -1,4 +1,5 @@
 use crate::prelude::*;
+use crate::utils::has_leading_newline;
 use crate::FormatNodeFields;
 use rome_js_syntax::{TsObjectType, TsObjectTypeFields};
 
@@ -13,13 +14,24 @@ impl FormatNodeFields<TsObjectType> for FormatNodeRule<TsObjectType> {
             r_curly_token,
         } = node.as_fields();
 
-        formatter
-            .delimited(
-                &l_curly_token?,
-                formatted![formatter, [members.format()]]?,
-                &r_curly_token?,
-            )
-            .soft_block_spaces()
-            .finish()
+        if has_leading_newline(members.syntax()) {
+            formatter
+                .delimited(
+                    &l_curly_token?,
+                    formatted![formatter, [members.format()]]?,
+                    &r_curly_token?,
+                )
+                .block_indent()
+                .finish()
+        } else {
+            formatter
+                .delimited(
+                    &l_curly_token?,
+                    formatted![formatter, [members.format()]]?,
+                    &r_curly_token?,
+                )
+                .soft_block_spaces()
+                .finish()
+        }
     }
 }
--- a/crates/rome_js_formatter/tests/specs/js/module/expression/member-chain/computed.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/expression/member-chain/computed.js.snap
@@ -19,5 +19,10 @@ Quote style: Double Quotes
 -----
 nock(/test/)
 	.matchHeader("Accept", "application/json")[httpMethodNock(method)]("/foo")
-	.reply(200, { foo: "bar" });
+	.reply(
+		200,
+		{
+			foo: "bar",
+		},
+	);
 
--- a/crates/rome_js_formatter/tests/specs/js/module/expression/member-chain/inline-merge.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/expression/member-chain/inline-merge.js.snap
@@ -28,7 +28,11 @@ _.flatMap(this.visibilityHandlers, (fn) => fn())
 	.concat(this.record.resolved_legacy_visrules)
 	.filter(Boolean);
 
-Object.keys(availableLocales({ test: true })).forEach(
+Object.keys(
+	availableLocales({
+		test: true,
+	}),
+).forEach(
 	(locale) => {
 		// ...
 	},
--- a/crates/rome_js_formatter/tests/specs/js/module/import/import_specifiers.js
+++ b/crates/rome_js_formatter/tests/specs/js/module/import/import_specifiers.js
@@ -1,5 +1,8 @@
 import { hey } from "hey"
 import { hey } from "hey";
+import {
+    apple,
+banana } from "fruits";
 import {test} from "foo.json" assert { for: "for" }
 import { // some funky comment
     loooooooooooooooooooong as moreeeeeeloooooooooooooooooooong,
--- a/crates/rome_js_formatter/tests/specs/js/module/import/import_specifiers.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/import/import_specifiers.js.snap
@@ -5,6 +5,9 @@ expression: import_specifiers.js
 # Input
 import { hey } from "hey"
 import { hey } from "hey";
+import {
+    apple,
+banana } from "fruits";
 import {test} from "foo.json" assert { for: "for" }
 import { // some funky comment
     loooooooooooooooooooong as moreeeeeeloooooooooooooooooooong,
@@ -30,6 +33,7 @@ Quote style: Double Quotes
 -----
 import { hey } from "hey";
 import { hey } from "hey";
+import { apple, banana } from "fruits";
 import { test } from "foo.json" assert { for: "for" };
 import {
 	// some funky comment
--- a/crates/rome_js_formatter/tests/specs/js/module/object/object.js
+++ b/crates/rome_js_formatter/tests/specs/js/module/object/object.js
@@ -19,3 +19,14 @@ let a = {
 
 	...spread,
 }
+
+const x = {apple: "banana"};
+
+const y = {
+	apple: "banana",
+};
+
+({a, b, c} = {a: 'apple', b: 'banana', c: 'coconut'});
+
+({
+	a, b, c} = {a: 'apple', b: 'banana', c: 'coconut'});
--- a/crates/rome_js_formatter/tests/specs/js/module/object/object.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/object/object.js.snap
@@ -25,6 +25,17 @@ let a = {
 	...spread,
 }
 
+const x = {apple: "banana"};
+
+const y = {
+	apple: "banana",
+};
+
+({a, b, c} = {a: 'apple', b: 'banana', c: 'coconut'});
+
+({
+	a, b, c} = {a: 'apple', b: 'banana', c: 'coconut'});
+
 =============================
 # Outputs
 ## Output 1
@@ -52,4 +63,14 @@ let a = {
 
 	...spread,
 };
+
+const x = { apple: "banana" };
+
+const y = {
+	apple: "banana",
+};
+
+({ a, b, c } = { a: "apple", b: "banana", c: "coconut" });
+
+({ a, b, c } = { a: "apple", b: "banana", c: "coconut" });
 
--- a/crates/rome_js_formatter/tests/specs/js/module/object/property_key.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/object/property_key.js.snap
@@ -22,6 +22,10 @@ Quote style: Double Quotes
 -----
 const foo = {
 	"foo-bar": true,
-	"bar": { "lorem_ispsum": { "lorem-ipsum": true } },
+	"bar": {
+		"lorem_ispsum": {
+			"lorem-ipsum": true,
+		},
+	},
 };
 
--- a/crates/rome_js_formatter/tests/specs/prettier/js/assignment/call-with-template.js.snap
+++ b/crates/rome_js_formatter/tests/specs/prettier/js/assignment/call-with-template.js.snap
@@ -1,8 +1,6 @@
 ---
 source: crates/rome_js_formatter/tests/prettier_tests.rs
-assertion_line: 57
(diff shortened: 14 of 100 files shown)


crates/rome_js_formatter/src/formatter.rs
```diff
@@ -203,6 +203,36 @@ impl Formatter {
         )
     }
 
+    /// Formats a group delimited by an opening and closing token, placing the
+    /// content in an [indent] group with [hard_line_break] tokens at the
+    /// start and end
+    pub(crate) fn format_delimited_block_hard_line_break(
```
cpojer, line 209: This is mostly a copy of `format_delimited_soft_block_spaces` and honestly I don't really understand the name of that function, so I'm not sure if the name of this function here makes sense. Please let me know if there is a better one.

crates/rome_js_formatter/tests/specs/js/module/object/getter_setter.js.snap
```diff
@@ -18,6 +18,10 @@ Indent style: Tab
...
 Quote style: Double Quotes
 -----
-let a = { get foo() {} };
-let b = { set foo(a) {} };
+let a = {
+	get foo() {},
+};
+let b = {
+	set foo(a) {},
+};
```
cpojer, line 26: This matches the input in `getter_setter.js`.

crates/rome_js_formatter/src/js/expressions/object_expression.rs
```diff
@@ -11,11 +11,16 @@ impl ToFormatElement for JsObjectExpression {
...
         if members.is_empty() {
             formatter.format_delimited_soft_block_indent(&l_curly_token?, members, &r_curly_token?)
+        } else if has_newlines == true {
+            formatter.format_delimited_block_hard_line_break(
+                &l_curly_token?,
+                members,
+                &r_curly_token?,
+            )
         } else {
             formatter.format_delimited_soft_block_spaces(&l_curly_token?, members, &r_curly_token?)
```
cpojer, line 25: We have about 10 callsites for `format_delimited_soft_block_spaces`. So far I'm only adding this feature to two of them because I'm looking for guidance on the optimal approach:

1. Use the above approach and add the newline check and explicit call to `format_delimited_block_hard_line_break` to every callsite.
2. Rename `format_delimited_soft_block_spaces` to `???` and pass a `has_newlines` boolean to the call, then decide on using soft or hard breaks within that single function.

Option 1 is more explicit, option 2 is less code. Which fits better with Rome?

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -203,6 +203,36 @@ impl Formatter {
...
+                    empty_element()
+                } else {
+                    format_elements![
+                        indent(format_elements![
+                            hard_line_break(),
+                            trailing_trivia,
+                            content,
+                            leading_trivia,
+                        ]),
+                        hard_line_break(),
```
MichaReiser, line 228: You can use `block_indent` here: 
```suggestion
                        block_indent(format_elements![
                            trailing_trivia,
                            content,
                            leading_trivia,
                        ]),
```

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -98,6 +98,18 @@ pub(crate) fn has_formatter_trivia(node: &JsSyntaxNode) -> bool {
     false
 }
 
+/// Returns true if this node contains newlines in trivias.
+pub(crate) fn has_newline(node: &JsSyntaxNode) -> bool {
```
MichaReiser, line 102: `has_newlines` might be misleading here that it tests if the node contains *any* new line. 

That's why I would rename this function to `has_leading_new_line()` or `has_new_line_before()`.

MichaReiser (commented):
Thanks Christoph. 

Regarding the failing `:arrow_nested` test. 

The test is failing because the formatter output isn't stable. Meaning, that re-formatting the formatted output yields a different result. Why or how it can be avoided I don't know right now. @leops and @ematipico have more experience with this kind of issues.

crates/rome_js_formatter/src/formatter.rs
ematipico: Apart from the initial check, this function is identical to `format_delimited_block_indent`: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/formatter.rs#L144-L157

Also, the `is_empty` check is already done inside `Formatter::format_delimited`, so it's not needed. And check `content.is_empty` is a bit... weird, why would a developer pass an empty `FormatElement`?

crates/rome_js_formatter/src/js/expressions/object_expression.rs
```diff
@@ -11,11 +11,16 @@ impl ToFormatElement for JsObjectExpression {
             members,
             r_curly_token,
         } = self.as_fields();
-
+        let has_newlines = has_newline(&members.syntax());
         let members = members.format(formatter)?;
-
         if members.is_empty() {
             formatter.format_delimited_soft_block_indent(&l_curly_token?, members, &r_curly_token?)
+        } else if has_newlines == true {
```
ematipico, line 18: I think this `has_newline` function is not needed, we can achieve the same result with `FormatElement` API.

In this particular case what you can do is:

```rs
let members = members.format(formatter)?;
let has_line_breaks = members.has_hard_line_breaks();
```

Instead, if you need a wider check (soft and hard line breaks), then I would suggest to create a new method similar to `has_hard_line_breaks`.

The good thing about inspecting our IR instead of checking the trivia, is that even though a node might have lines breaks in the source code, we don't know for sure how it would be formatted until is actually printed. This means it's not reliable. On the other hand, inspect IR allows us to know 100% how this code will be printed and react accordingly.

At the moment we use this method only in one place, but I think we should start using it more because we have cases where we need to change the IR based on if we have line breaks in the children/siblings: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/utils/call_expression.rs#L468-L480

crates/rome_js_formatter/src/js/expressions/object_expression.rs
ematipico: The functions `format_delimited_*` were an attempt to group common patterns that we were starting to see across the board. If you think that this is going to be a common pattern, then it makes sense to make it generic, instead if this is just a single case, then it should stay isolated.

The `has_newlines` argument won't be needed, as explained in my comment: https://github.com/rome/tools/pull/2458/files#r851766933

crates/rome_js_formatter/src/js/expressions/object_expression.rs
MichaReiser: I'm starting to develop a different perspective on this. I'm coming to the conclusion that inspecting our IR is an anti-pattern and I'm currently working on an API re-work that will make this intentionally difficult, to push people to instead make decisions based on the `SyntaxNode`. 

The main reason for this is that `has_hard_line_breaks` is inherently slow, it needs to traverse over all elements, even into nested lists. It also makes it harder to reason about our formatter. I formatted something like X, but then has a complete different outcome (split trivia is an example, which again, is slow and has up to 2 vec allocations). 

However, regarding whatever we should inspect the IR or not, inspecting the IR won't yield the correct result in this case because the formatting should depend on whatever the source code contained a line break between the `{` and the first member (first token).

ematipico: > whatever the source code contained a line break between the { and the first member (first token).

But how do you assure that line break won't disappear when the first member is formatted?

MichaReiser: The code force inserts a line break in that case

```
  formatter.format_delimited_block_hard_line_break(
        &l_curly_token?,
        members,
        &r_curly_token?,
    )
```

crates/rome_js_formatter/src/formatter.rs
cpojer: Updated to use `format_delimited_block_indent`, I wasn't aware of how it worked! For context, I previously mostly just copy-pasted `format_delimited_soft_block_indent`.

crates/rome_js_formatter/src/utils/mod.rs
cpojer: Done!

cpojer:
Thanks for the feedback so far. Remaining questions:
* The unstable printing test issue.
* Are you happy with the `has_leading_newline` + `format_delimited_block_indent` calls or do you somehow want to pull responsibility into `format_delimited_block_indent` so it can be handled generically? The decision on that will determine the final state of this PR.

leops:
After looking into it, the stability issue seem to be caused by the `has_hard_line_breaks` check in the formatting of call expression (at [`crates\rome_js_formatter\src\utils\call_expression.rs:L480`](https://github.com/rome/tools/tree/main/crates/rome_js_formatter/src/utils/call_expression.rs#LL480)): replacing the soft line breaks with hard ones in the formatting of the object expression causes the outer call chain to emit an additional group that cause the entire statement to print differently.
This check was originally intended to approximately replicate the logic of the printer but the approximation becomes evident in the presence of soft line breaks, so I think we should get rid of this logic in favor of something closer to what Prettier does (and doesn't need to inspect `FormatElement`) since we're aiming for better compatibility. But that's outside of the scope of this PR ...

MichaReiser:
> After looking into it, the stability issue seem to be caused by the `has_hard_line_breaks` check in the formatting of call expression (at [`crates\rome_js_formatter\src\utils\call_expression.rs:L480`](https://github.com/rome/tools/tree/main/crates/rome_js_formatter/src/utils/call_expression.rs#LL480)): replacing the soft line breaks with hard ones in the formatting of the object expression causes the outer call chain to emit an additional group that cause the entire statement to print differently. This check was originally intended to approximately replicate the logic of the printer but the approximation becomes evident in the presence of soft line breaks, so I think we should get rid of this logic in favor of something closer to what Prettier does (and doesn't need to inspect `FormatElement`) since we're aiming for better compatibility. But that's outside of the scope of this PR ...

Is anyone looking into this? If not, can we track this as an issue?

leops:
> Is anyone looking into this? If not, can we track this as an issue?

I don't think anyone is working on this right now, I can create an independent issue for it. Since it seems to be blocking for this PR I'm not really sure how we want to handle that, do would we hold on to merging this until the issue is fixed or merge this change now and fix the stability issue later ?

cpojer:
We cannot merge this PR unless we disable the test. It would be great to fix this issue, and then I can rebase my work on that.

@MichaReiser Still looking for guidance on my second question about whether the implementation in this PR is acceptable or whether things should be moved around.

MichaReiser:
> We cannot merge this PR unless we disable the test. It would be great to fix this issue, and then I can rebase my work on that.
> 
> @MichaReiser Still looking for guidance on my second question about whether the implementation in this PR is acceptable or whether things should be moved around.

`format_delimited_block_indent` is used for example for the `{}` of classes where prettier doesn't preserve the line breaks. There are more call sites where this is the case as well. That's why I think your current implementation that explicitly handles it for objects is the right choice.

cpojer:
Ok this PR is ready for review. It updates a lot of snapshots to look like the input code (I initially got confused and was comparing the snapshot changes to the previous snapshot version, which obviously has changed!). I also changed how TypeScript interface declarations are printed since in Prettier they will always break, even with just one prop (thanks @MichaReiser).

The only blocker before merging besides review comments is the failing arrows test. I think it would also be great to fix the test infra so that it keeps running instead of crashing on the first failing test like this one. It's quite confusing otherwise when it doesn't make it through the whole test run.

crates/rome_js_formatter/src/js/assignments/object_assignment_pattern.rs
```diff
@@ -11,10 +14,18 @@ impl FormatNode for JsObjectAssignmentPattern {
...
+                &r_curly_token?,
+            )
+        } else {
+            formatter.format_delimited_soft_block_spaces(
+                &l_curly_token?,
+                properties.format(formatter)?,
+                &r_curly_token?,
+            )
+        }
     }
```
MichaReiser, line 30: Prettier doesn't seem to preserve line breaks in object assignment expressions.

[Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAKYACAOlD3f5454CG+OAvhgLwbwDOMAlANwgA0IEADjAJbR6yUCQBOoiAHcACmIRCUJADaSSATyGcARqJJgA1nBgBlEgFs4AGT5Q4yAGbL6cbboNHj3PTYDmyGKIAri4gcGZacAAmkVGWJFA+gSQ+cABiEKJmJDD8CcggJIEwEBwgABYwZkoA6mV8DF5gcMby9XwAbvVq+WD0miA2zqIw0ro+WQ5OIQBW9AAexr5KcACKgRDwk0rOnF6iQ-laJBFKpdyiNjDVfJEwZcgAHAAMuxLO1brc+edwQ+12nAAjut4KMeAoCvQALS2KJRUqiODAviI0bJCZIRzbELOMx8fxBHFLVYguyYqacGDHa63e5IABMlN0fCUvgAwhAzBjQvQAKylQLOAAqxwUWJ2IHawQAklAYrBjGALrwAIJy4wwNTLLbOCgUIA)

crates/rome_js_formatter/src/js/bindings/object_binding_pattern.rs
```diff
@@ -11,10 +14,18 @@ impl FormatNode for JsObjectBindingPattern {
             r_curly_token,
         } = self.as_fields();
 
-        formatter.format_delimited_soft_block_spaces(
-            &l_curly_token?,
-            properties.format(formatter)?,
-            &r_curly_token?,
-        )
+        if has_leading_newline(properties.syntax()) {
```
MichaReiser, line 17: Nor in binding patterns [Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEwA6VOZ4FGH6YCGpeAvpgLybwDOMA3CADQgQAOMAltCbJQ5AE5iIAdwAK4hMJTlUU8gE9hXAEZjyYANYYAyuQC2cADL8ocZADNlTONt0HjPPdYDmyGGICuziBwplpwACbhERbkUF7+5F5wAGIQYqbkMAJxyCDk-jAQnCAAFjCmqADqJfzMHmBwRgq1-AButWq5YEyaINZOYjAyul4Z9o5BAFZMAB5G3ugAiv4Q8OOoTlweYgO5WuRhqMU8YtYwlfzhMCXIABwADFuSTpW6PLkncAOttlwAjit4MNeIo8kwALQ2CIRYpiOAA-hw4aJMZIBwbIJOUz8XwBTELODLVa2NETLgwA4XK43JAAJnJun4qG8AGEIKZUcEmABWYr+JwAFQOinRmxArUCAEkoFFYEYwKc+ABBGVGGBqdDrJzUahAA)

crates/rome_js_formatter/src/js/module/export_named_clause.rs
```diff
@@ -21,13 +23,18 @@ impl FormatNode for JsExportNamedClause {
...
 
+        let has_newline = has_leading_newline(specifiers.syntax());
         let specifiers = specifiers.format(formatter)?;
 
-        let list = formatter.format_delimited_soft_block_spaces(
-            &l_curly_token?,
-            specifiers,
-            &r_curly_token?,
-        )?;
+        let list = if has_newline {
```
MichaReiser, line 29: Same for export named and export named from. Prettier doesn't seem to preserve the line breaks [Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACYPAHSgEMSBfPAM2wgFtiR4BnGIkAbhABoQIMMAJbQWyUKWx0A7gAVJCMSlIAbaaQCeYvgCNspMAGs4MAMql6cADJCocZNVUs4u-UZOmMB2wHNkMbABXF1R6HTgAEwjIq1IoH0DSHzgAMRx6UhhheOQQUkCYCF4QAAsYehUAdRKhVi8wOFNFWqEAN1qNXLAWbRBbZ1xZfR8MhycQgCsWNFNfFTgARUCIeDGVZz4vbAHcnVJwlWKMbFsYSqEImBLkAA4ABk26Z0r9DFzjuAHW+z4AR2X4EMBEo8iwALR2SKRYrYOD-ISwoZJUZIRzrELOehCfxBDFzRYA+yo8Z8GD7c6Xa5IABMpP0QhUvgAwgwUagWABWYqBZwAFX2SjRGxArWCAEkoNFYKYwCdBABBSWmGAaeZrZwUChAA)

crates/rome_js_formatter/src/js/module/import_assertion.rs
```diff
@@ -14,14 +16,19 @@ impl FormatNode for JsImportAssertion {
...
+        let has_newline = has_leading_newline(assertions.syntax());
         let assert_token = assert_token.format(formatter)?;
         let assertions = assertions.format(formatter)?;
 
-        let result = formatter.format_delimited_soft_block_spaces(
-            &l_curly_token?,
-            assertions,
-            &r_curly_token?,
-        )?;
+        let result = if has_newline {
```
MichaReiser, line 23: Neither is prettier preserving line breaks for import assertions. [Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAgA98AzHCdfAHRBhvwEMBnJuPfYKqffGAT0xwk1WnCZ0QXAL4gANCAiYYqaE2SgGOcgHcACpoRqUDADbaGfNfIBGOBmADWcGAGUG6OABlUUOMhKmrDZ2js4umPY+AObIMDgArnDycOjWcAAm6RmeDFBR8QxRcABiuOgMMMp5yCAM8TAQciAAFjDoJgDqzajwTBFgcC6GPagAbj18NWAsTT6seLp2UeX+gUkgAFZMhC7RJnAAivEQ8KsmQSAROPM11gxpJk2YOD4wHajpMM3IABwADPJnhBWB07Jgas8xGxRn55ABHY7wRaKIy1JgAWl8GQyTRwcARqDxi0KKyQAXO61Y6FQsQSlL2h0RfjJa3kMHu70+3yQACY2XZUCZogBhCikkBiACsTXirAAKvcjOSLqNEgBJKBZWAuMAvJQAQU1Ln4+zOrCkUiAA)

crates/rome_js_formatter/src/js/module/named_import_specifiers.rs
```diff
@@ -11,8 +14,17 @@ impl FormatNode for JsNamedImportSpecifiers {
             r_curly_token,
         } = self.as_fields();
 
+        let has_newline = has_leading_newline(specifiers.syntax());
         let specifiers = specifiers.format(formatter)?;
 
-        formatter.format_delimited_soft_block_spaces(&l_curly_token?, specifiers, &r_curly_token?)
+        if has_newline {
```
MichaReiser, line 20: Or import named [Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAmAB0p98SyBDAGnwCMSBffAMxwnXJAZBGpAiYYqaAGdkoSjnYB3AApSE4lJQA2MygE9x-OjkpgA1nBgBlSujgAZVFDjIWa0XF36jJ05gO2A5shg4AK4uIHDodHAAJpFRVpRQPoGUPnAAYrjolDDCCcgglIEwEHwgABYw6KoA6qWo8KJeYHCmSnWoAG51mnlgojpoUM54cvo+mQ5OIQBWogAepr6qcACKgRDwE6rO-F44Q3l0lBGqJZg4tjBVqJEwpcgAHAAMO+zOVfqYeWdwQ+32-ABHNbwEaCZT5UQAWjsUSiJRwcCBqARI2S4yQji2IWc6FQ-iC2MWK2B9gxk34MCOVxudyQACYKfpUKpfABhDjo0KiACsJUCzgAKkdlJjtiB2sEAJJQGKwUxgc5CACCMtMME0S02zkYjCAA)

MichaReiser (changes requested):
Thanks, @cpojer for updating your PR and sorry for the merge conflicts. 

The code looks good to me but this PR now goes further than what prettier does. Prettier only preserves line breaks for object literals and object type literals, but not for imports/exports/assertions/etc...

I would recommend keeping compatibility with Prettier for now.

cpojer:
Oh that's unfortunate. Thank you so much for double checking. Subjectively, what I had is better for consistency but objectively it's better to align with prettier. I updated the PR and reverted the changes.

MichaReiser:
> Oh that's unfortunate. Thank you so much for double checking. Subjectively, what I had is better for consistency but objectively it's better to align with prettier. I updated the PR and reverted the changes.

Thanks. Would you mind having a look at the lint and test failures. We should then be good to merge

cpojer:
Done (I think!).

ematipico:
Tests still failing 😔

It seems there are some snapshots that break when reformatting... And few needs to be updated again.


This happens because if a change affects prettier test suite AND our internal suite, we need to run and update the snapshots twice.

Not sure why this happens but for now that's how it is (sorry)

cpojer:
@ematipico how do I update the snapshots? When I run `cargo test` the only failing test is `formatter::js_module::specs::js::module::arrow::arrow_nested` like we discussed.

ematipico:
`cargo insta review `

Instead if you keep getting a diff error, it means that what it is formatted the first time doesn't match the output of the second pass of the formatter (on the emitted code of the first time)

cpojer:
@ematipico Rebased and resolved conflicts, now the arrow test is failing with a different issue it looks like.

crates/rome_js_formatter/tests/specs/js/module/object/getter_setter.js.snap
ematipico: That file is the source code, you should check the file that finishes with `.prettier-snap`. 

[Check their playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEBDTBeTYTAcw0wDMIIAKASkIF9MGBuEAGhAgAcYBLaAGdkobACcxEAO4AFcQmEpsqKdgCewzgCMx2MAGsMAZWwBbOABk+UOMnLLBcbboPHue68WQwxAVycgcKZacAAmoWEW2FDEvtikAGIQYqbYMPwxyCDYvjAQHCAAFjCmqADqhXzwgu5gcEYKVXwAblVqWWCCmiDWjmIwMrrEqXYOAQBWggAeRp7oAIq+EPCjqI6c7mJ9WVrYIagF3GLWMGV8oTCFyAAcAAwbko5lutxZR3B9zbacAI5L8IMeIpsoIALQ2MJhApiOB-PgwwbxEZIexrAKOUx8bx+dFzOCLZa2FFjTgwPZnC5XJAAJlJuj4qE8AGEIKZkYFBABWAq+RwAFT2ih8-k4zX8AEkoBFYEYwMdeABBKVGGBqdCrRwMBhAA)

ematipico (changes requested):
It seems that the algorithm is not working as intended.

Now the object expressions **always** break, regardless if they have new lines or not.

I **strongly** suggest to create a new test case in our test suite, ideally the code that Micha attached to the issue. This will help me to identify the new file and see if it matches the intended behaviour

cpojer:
As explained on Discord, this is very confusing but the code here is correct except for the example where it breaks the outer parameters list in a function call into newlines which prettier doesn't do.

The input code for all of these examples has those newlines in place, so the snapshots are being updated from not printing those newlines to printing them. When you look at the snapshot changes, it looks like things are broken, but when you compare the source with the new snapshot versions you'll see that it is correct. I got confused about the exact same thing before I wrapped my head around that.

ematipico (approved):

ematipico:
The reason why this PR keeps failing is because we don't yet format call arguments as prettier does.

Prettier uses #2529 and an [extensive heuristic](https://github.com/prettier/prettier/blob/a9de2a128cc8eea84ddd90efdc210378a894ab6b/src/language-js/print/call-arguments.js) to how to print them. The work in tracked in #2421 , so I guess we should put this PR on hold until we have everything we need in place.

cpojer:
@MichaReiser I redid this PR with the new formatting infra, please check out whether this matches expectations!

cpojer:
Not sure if I did it right (`env REPORT_PRETTIER=1 cargo test -p rome_js_formatter --test prettier_tests`), but here is the diff compared to `main`:

**File Based Average Prettier Similarity**: 69.38% → 70.89%
**Line Based Average Prettier Similarity**: 64.39% → 65.39%

MichaReiser (approved):
This is such an awesome contribution. Thank you @cpojer .

PR #2459, opened 2022-04-17
chore(ci): add conditions to workflow runs

jer3m01:
## Summary

Some PRs make changes to only specific parts of the repo and do not require all checks to run on unmodified parts.
This PR adds conditions to workflow runs to save time and resources.

Example: #2457 only modified `.md` files but the ci still reformatted all files and tried to deploy playground (and failed without reasons).

Reference used for this PR: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths.

Also renamed jobs with clearer titles to avoid confusion (cherry pick if not wanted).

## Test Plan
Unable to test myself but *should* work.


jer3m01:
> ## Test Plan
> Unable to test myself but *should* work.

Well github seems to be picking up the fork's workflow configs. And no jobs were ran so it works! *I think*

.github/workflows/pull_request.yml
```diff
@@ -5,14 +5,20 @@ on:
   pull_request:
     branches:
       - main
+    paths: # Only run when changes are made to rust or js code
+      - '**.rs'
+      - '**.js'
+      - '**.jsx'
```
ematipico, line 11: Our formatter is not yet stable enough to cover most of JSX syntax... Should we wait?

jer3m01: It currently does it anyway, for all commits. This just restricts to commits with these file extensions, it doesn't force rome to format them.

MichaReiser: This will also run the action when changes have been made to the website only and doesn't cover the case if someone changes the `Cargo.toml`. 

Is it possible to use a negative pattern instead and exclude the `website` folder?

jer3m01: Currently the `pull_request.yml` file handles all pull requests and runs both rs formatting & testing and js formatting.
Excluding the website folder wouldn't work as `website/` still requires js formatting

Also the job should not run for every `!website/` as it would still run for `.md` changes which we don't want.

I could split the workflow into 2 separate configs: 
1. For `**.(rs|toml)` & `crates/**.(js|jsx|ts|tsx)` (these files are used in tests and snapshots, modifying them requires rerunning tests) pull requests: 
    Runs rust formatting, linting and testing

2. For `**.(js|jsx|ts|tsx)` pull requests: 
    Runs js formatting

.github/workflows/pull_request.yml
```diff
@@ -5,14 +5,19 @@ on:
   pull_request:
     branches:
       - main
+    paths: # Only run when changes are made to rust code or root Cargo
+      - 'crates/**'
+      - 'xtask/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
```
MichaReiser, line 12: Can we include `rust-toolchain.toml` and `rustfmt.toml`?

MichaReiser (approved):

MichaReiser:
Thank you. Let's give this a try!

PR #2461, opened 2022-04-18
Update GitHub Actions version

Gumichocopengin8:
## Summary

Update `actions/xxxxx@vx` to `actions/xxxxxx@v3 or 6` since Node12 will be out of life after Aril 30, 2022 [[Reference](https://nodejs.org/en/about/releases/)].  
`actions/xxxx@v3` or `actions/xxxx@v6`  use Node16 whose support lasts until April 30, 2024.

## Test Plan

Check if GitHub Actions workflows work well

### DIFF
--- a/.github/workflows/bench_formatter.yml
+++ b/.github/workflows/bench_formatter.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - name: Get PR SHA
         id: sha
-        uses: actions/github-script@v4
+        uses: actions/github-script@v6
         with:
           result-encoding: string
           script: |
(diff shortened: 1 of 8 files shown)


Gumichocopengin8:
Hello @ematipico,
Could you manually run the GitHub Actions to test this change?

Gumichocopengin8:
I couldn't see the GitHub Actions running at first but it was just invisible.
Could somebody run this flow again? I think the error, `AWS_S3_BUCKET is not set. Quitting.` comes from somewhere else. https://github.com/jakejarvis/s3-sync-action/issues/53

NicholasLYang:
Appears to be a permissions issue. I'll see if we can fix that. But agreed, it shouldn't be a blocker for this PR

Gumichocopengin8:
@NicholasLYang Thanks

MichaReiser (approved):
Thanks!

PR #2463, opened 2022-04-19
refactor(rome_rowan): Use associated types for `Ast*` traits

MichaReiser:
The `AstNode`, `AstNodeList`, and `AstSeparatedList` currently use generics to
parametrize the `Language` and `AstNode` (in case of the lists).

This PR changes the `Ast*` traits to use associated types instead. The semantic
difference of this change is that a single struct can no longer implement the traits for different languages.

Or, as explained in the Rust book:

> With associated types, we don’t need to annotate types because we can’t implement a trait on a type multiple times. In Listing 19-12 with the definition that uses associated types, we can only choose what the type of Item will be once, because there can only be one impl Iterator for Counter. We don’t have to specify that we want an iterator of u32 values everywhere that we call next on Counter. [source](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html)

For example, the following will no longer be possible:

```rust
impl AstNode<JsLanguage> for JsAnyExpression {... }
impl AstNode<CssLanguage> for JsAnyExpression { ... }
```

I don't see a use case where we want to implement `AstNode` for different languages for a specific node.
The only use case that I'm aware of that gets close are super languages, but I think this should instead use a different `Language` and be union types instead.

 Using associated types allows us to create generic trait implementations for `AstNodeList` that otherwise aren't possible because a single node could implement the same trait multiple times:

 ```rust
 impl<T> MyTrait for T
   where T: AstNodeList {
 }
 ```

### DIFF
--- a/crates/rome_analyze/src/analysis_server.rs
+++ b/crates/rome_analyze/src/analysis_server.rs
@@ -46,13 +46,16 @@ impl AnalysisServer {
         suppressions::compute(tree)
     }
 
-    pub fn query_nodes<T: AstNode<JsLanguage>>(&self, file_id: FileId) -> impl Iterator<Item = T> {
+    pub fn query_nodes<T: AstNode<Language = JsLanguage>>(
+        &self,
+        file_id: FileId,
+    ) -> impl Iterator<Item = T> {
         trace!("Query nodes: {:?}", std::any::type_name::<T>());
         let tree = self.parse(file_id);
         tree.descendants().filter_map(|n| T::cast(n))
     }
 
-    pub fn find_node_at_range<T: AstNode<JsLanguage>>(
+    pub fn find_node_at_range<T: AstNode<Language = JsLanguage>>(
         &self,
         file_id: FileId,
         range: TextRange,
--- a/crates/rome_rowan/src/ast/mod.rs
+++ b/crates/rome_rowan/src/ast/mod.rs
@@ -11,33 +11,35 @@ use std::marker::PhantomData;
 use text_size::TextRange;
 
 use crate::syntax::{SyntaxSlot, SyntaxSlots};
-use crate::{Language, SyntaxList, SyntaxNode, SyntaxToken};
+use crate::{ast, Language, SyntaxList, SyntaxNode, SyntaxToken};
 
 /// The main trait to go from untyped `SyntaxNode`  to a typed ast. The
 /// conversion itself has zero runtime cost: ast and syntax nodes have exactly
 /// the same representation: a pointer to the tree root and a pointer to the
 /// node itself.
-pub trait AstNode<L: Language> {
+pub trait AstNode {
+    type Language: Language;
+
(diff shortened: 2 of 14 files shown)


crates/rome_analyze/src/analysis_server.rs
```diff
@@ -46,13 +46,16 @@ impl AnalysisServer {
         suppressions::compute(tree)
     }
 
-    pub fn query_nodes<T: AstNode<JsLanguage>>(&self, file_id: FileId) -> impl Iterator<Item = T> {
+    pub fn query_nodes<T: AstNode<Language = JsLanguage>>(
```
leops, line 49: I think for most of these the language was specified because it was required, but in practice the `Language = JsLanguage` could be removed and the code would still work, with the advantage that it would now be language agnostic

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -11,33 +11,35 @@ use std::marker::PhantomData;
...
 /// conversion itself has zero runtime cost: ast and syntax nodes have exactly
 /// the same representation: a pointer to the tree root and a pointer to the
 /// node itself.
-pub trait AstNode<L: Language> {
+pub trait AstNode {
+    type Language: Language;
+
     /// Returns `true` if a node with the given kind can be cased to this AST node.
-    fn can_cast(kind: L::Kind) -> bool;
+    fn can_cast(kind: <<Self as ast::AstNode>::Language as Language>::Kind) -> bool;
```
leops, line 24: Can't this be shortened to at least `<Self::Language as Language>::Kind` ?

crates/rome_analyze/src/analysis_server.rs
MichaReiser: The problem is that `query_nodes` calls into `parse` which is only implemented for JS

```rust

    pub fn query_nodes<T: AstNode>(&self, file_id: FileId) -> impl Iterator<Item = T> {
        trace!("Query nodes: {:?}", std::any::type_name::<T>());
        let tree = self.parse(file_id);
        tree.descendants().filter_map(|n| T::cast(n))
    }
```

```
error[E0308]: mismatched types
  --> crates/rome_analyze/src/analysis_server.rs:52:51
   |
52 |         tree.descendants().filter_map(|n| T::cast(n))
   |                                                   ^ expected associated type, found struct `JsLanguage`
   |
   = note: expected struct `SyntaxNode<<T as AstNode>::Language>`
              found struct `SyntaxNode<JsLanguage>`
   = help: consider constraining the associated type `<T as AstNode>::Language` to `JsLanguage`
   = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
```

leops: Ah right we don't have a generic parse function. Well I think we'll want to refactor the analyzer code soon anyway both for linting and multi-language support so we can leave it like that for now, and get back to making it generic when the required functionality comes online

xunilrj (approved):

PR #2464, opened 2022-04-19
feat(rome_rowan): add `has_comments` API

ematipico:
## Summary

This PR adds a new API to `rome_rowan` called `has_comments`.

This API does a shallow check if the current node contains trivia, it doesn't check the descendants.

I then updated the places where we meant to do a shallow check inside the formatter.

## Test Plan

Added a new test

### DIFF
--- a/crates/rome_rowan/src/syntax/node.rs
+++ b/crates/rome_rowan/src/syntax/node.rs
@@ -374,10 +374,16 @@ impl<L: Language> SyntaxNode<L> {
         SyntaxList::new(self)
     }
 
-    /// Whether the node contains any comments.
+    /// Whether the node contains any comments. This function checks
+    /// **all the descendants** of the current node.
     pub fn contains_comments(&self) -> bool {
         self.descendants_tokens()
             .any(|tok| tok.has_trailing_comments() || tok.has_leading_comments())
+    }
+
+    /// It checks if the current node has trailing or leading trivia
+    pub fn has_comments(&self) -> bool {
+        self.has_trailing_comments() || self.has_leading_comments()
     }
 
     /// Whether the node contains trailing comments.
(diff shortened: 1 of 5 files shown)


ematipico:
!bench_formatter

MichaReiser (approved):

leops (commented):
I think having two functions called `has_comments` and `contains_comments` is ambiguous, explaining the precise difference between them in the doc comment is alright but it's also very easy to miss. I feel like they should have more explicit names like `has_comments_direct` and `has_comments_descendants` to make sure anyone using them knows exactly what they're doing at a glance (and to be honest I'm not even sure those are explicit enough)

ematipico:
> I think having two functions called `has_comments` and `contains_comments` is ambiguous, explaining the precise difference between them in the doc comment is alright but it's also very easy to miss. I feel like they should have more explicit names like `has_comments_direct` and `has_comments_descendants` to make sure anyone using them knows exactly what they're doing at a glance (and to be honest I'm not even sure those are explicit enough)

Good point. To be fair I wondered if we actually need the deep check... I mean, are there cases where is it needed? Food for thoughts

crates/rome_rowan/src/syntax/node.rs
```diff
@@ -374,12 +374,18 @@ impl<L: Language> SyntaxNode<L> {
...
-    pub fn contains_comments(&self) -> bool {
+    /// Whether the node contains any comments. This function checks
+    /// **all the descendants** of the current node.
+    pub fn has_comments_descendants(&self) -> bool {
         self.descendants_tokens()
             .any(|tok| tok.has_trailing_comments() || tok.has_leading_comments())
     }
 
+    /// It checks if the current node has trailing or leading trivia
+    pub fn has_comments_direct(&self) -> bool {
```
MichaReiser, line 385: Nit: `has_leading_or_trailing_comments()`. Expresses most precisely what this method does.

MichaReiser (approved):

PR #2466, opened 2022-04-19
Feat(json): finish json ungram

IWANABETHATGUY:
## Summary
part of https://github.com/rome/tools/pull/2466#issue-1208593134


IWANABETHATGUY:
I basic finish those check point
 
 create a json.ungram file where we will slowly implement new grammars
 create a rome_json_syntax crate, where we will have all things related to the JSON syntax, generated code too
 expand the codegen of the grammar to include JSON too

IWANABETHATGUY:
I have ran `cargo check` locally.

IWANABETHATGUY:
I found there has been a PR about json ungrammar already, but I guess my implementation would respect json spec

IWANABETHATGUY:
@MichaReiser Any idea about this, It seems that i need a maintainer to trigger the CI

ematipico:
> @MichaReiser Any idea about this? It seems that i need a maintainer to trigger the CI

GitHub actions are now set up in a way where first time contributor can't run the CI. This was done by GitHub to prevent people to use CIs as a tool for mining cryptocurrency

IWANABETHATGUY:
> 
Yeah, Thanks for explaining that

IWANABETHATGUY:
@ematipico sorry for interupting, need to trigger again, i fixed the lint error

crates/rome_json_syntax/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl JsonSyntaxKind {
+    pub fn is_trivia(self) -> bool {
+        matches!(
+            self,
+            JsonSyntaxKind::NEWLINE | JsonSyntaxKind::WHITESPACE | JsonSyntaxKind::COMMENT
+        )
+    }
+
+    /// Returns `true` for any contextual (await) or non-contextual keyword
```
MichaReiser, line 34: This comment doesn't apply to JSON that doesn't have contextual keywords. I would recommend removing it.

crates/rome_json_syntax/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl rome_rowan::SyntaxKind for JsonSyntaxKind {
+    fn is_unknown(&self) -> bool {
+        matches!(self, JsonSyntaxKind::JSON_UNKNOWN)
+    }
+
+    fn to_unknown(&self) -> Self {
+        match self {
+            JsonSyntaxKind::JSON_UNKNOWN => JsonSyntaxKind::JSON_UNKNOWN,
+            _ => unreachable!(),
```
MichaReiser, line 49: This method gets called if the Parser creates a node structure that doesn't match the structure defined in the grammar. This could for example happen if the parser performs some error recovery and parses out a number in an objects literal's key position. 

It should be sufficient to use a single `UNKNOWN` node for json. That's why I would propose to always return `JSON_UNKNOWN` in this method.

crates/rome_json_syntax/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl TryFrom<JsonSyntaxKind> for TriviaPieceKind {
+    type Error = ();
+
+    fn try_from(value: JsonSyntaxKind) -> Result<Self, Self::Error> {
+        if value.is_trivia() {
+            match value {
+                JsonSyntaxKind::NEWLINE => Ok(TriviaPieceKind::Newline),
+                JsonSyntaxKind::WHITESPACE => Ok(TriviaPieceKind::Whitespace),
+                JsonSyntaxKind::COMMENT => Ok(TriviaPieceKind::SingleLineComment),
```
MichaReiser, line 72: You can remove the comment because (plain) JSON doesn't have comments.

xtask/codegen/json.ungram
```diff
@@ -0,0 +1,62 @@
...
+// and their children will need to yield nodes and tokens as well.
+// For this reason, SyntaxElement = SyntaxElement
+
+SyntaxElement = SyntaxElement
+
+JsonUnknown = SyntaxElement*
+
+JsonRoot = JsonValue
+
+JsonValue =  JsonString | JsonBoolean | JsonNull | JsonNumber | JsonArray | JsonObject
```
MichaReiser, line 44: Can we add `JsonUnknown` to this union. This will give us better error recovery in case there's some value that the parser isn't able to parse.

xtask/codegen/json.ungram
```diff
@@ -0,0 +1,62 @@
...
+
+JsonArrayElementList = (JsonValue (',' JsonValue)* )
+
+JsonBoolean = 'true' | 'false'
+
+JsonNull = 'null'
+
+JsonString = 'json_string_literal'
+
+JsonNumber = 'json_number_literal'
```
MichaReiser, line 62: Nit: Add a newline at the end of the file

xtask/codegen/src/json_kinds_src.rs
```diff
@@ -0,0 +1,33 @@
...
+        (")", "R_PAREN"),
+        ("{", "L_CURLY"),
+        ("}", "R_CURLY"),
+        ("[", "L_BRACK"),
+        ("]", "R_BRACK"),
+    ],
+    keywords: &["null", "true", "false"],
+    literals: &["JSON_STRING_LITERAL", "JSON_NUMBER_LITERAL"],
+    // keep comment token for json 5 extension
+    tokens: &["ERROR_TOKEN", "NEWLINE", "WHITESPACE", "COMMENT"],
```
MichaReiser, line 17: `Comment` won't be needed until we decide to implement any JSON dialect.

MichaReiser (approved):
Nice work! I've only a few small improvement suggestions before merging.

MichaReiser:
Awesome! Thank you for this contribution.

yisibl:
Nice work

Issue #2426, opened 2022-04-13
Normalize Big-Int literals


MichaReiser:
Prettier normalizes hex/octal/binary escape sequences whereas Rome doesn't [Playground](https://play.rome.tools/?lineWidth=80&indentStyle=space&quoteStyle=double&indentWidth=2&typescript=true&jsx=false&sourceType=module#MQAwADAAbgAKADkAMgAyADMAMwA3ADIAMAAzADYAOAA1ADQANwA3ADUAOAAwADcAbgAKADAAbwAxADYANAAzADIAbgAKAC8ALwAgADAATwAgAC0APgAgADAAbwAKADAATwAxADYANAAzADIAbgAKAC8ALwAgAEYARgAgAC0APgAgAGYAZgAKADAAeABGAEYARgAxADIAMwBuAAoALwAvACAAMABYACAALQA+ACAAMAB4AAoAMABYAEYARgBGADEAMgAzAG4ACgAwAGIAMQAwADEAMAAxADEAMQAwADEAbgAKAC8ALwAgADAAQgAgAC0APgAgADAAYgAKADAAQgAxADAAMQAwADEAMQAxADAAMQBuAAoAMgAwADAAXwAwADAAMABuAAoALwAvACAAQQAgAC0APgAgAGEACgAwAHgAMAAwADAAMABfADAAMAAwAEEAbgAKADAAYgAwADEAMQAxAF8AMQAxADEAMQBuAA==). 

[Report](https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsbig-intliteraljs)

## Input 

```ts
100n
9223372036854775807n
0o16432n
// 0O -> 0o
0O16432n
// FF -> ff
0xFFF123n
// 0X -> 0x
0XFFF123n
0b101011101n
// 0B -> 0b
0B101011101n
200_000n
// A -> a
0x0000_000An
0b0111_1111n
```

## Prettier

```ts
100n;
9223372036854775807n;
0o16432n;
// 0O -> 0o
0o16432n;
// FF -> ff
0xfff123n;
// 0X -> 0x
0xfff123n;
0b101011101n;
// 0B -> 0b
0b101011101n;
200_000n;
// A -> a
0x0000_000an;
0b0111_1111n;
```

## Rome

```ts
100n;
9223372036854775807n;
0o16432n;
// 0O -> 0o
0O16432n;
// FF -> ff
0xFFF123n;
// 0X -> 0x
0XFFF123n;
[...]

PR #2467, opened 2022-04-19
normalize big int literals

xunilrj:
see https://github.com/rome/tools/issues/2426

# to_ascii_lowercase_cow

For this task, I don't need a general Unicode to_lowercase, and Rust as of today only give me access to lowering the case of a Unicode through ```char::to_lowercase``` which returns an iterator (because of some crazy languages!).

This made my method slower than calling ```str::to_lowercase```.

If I simplify the assumption to support only ASCII, I can make my method faster and avoid allocations.

```
to_ascii_lowercase
  Instructions:                 439
  L1 Accesses:                  589
  L2 Accesses:                    1
  RAM Accesses:                  13
  Estimated Cycles:            1049

to_ascii_lowercase_cow
  Instructions:                 482
  L1 Accesses:                  612
  L2 Accesses:                    1
  RAM Accesses:                  12
  Estimated Cycles:            1037
```

### DIFF
--- a/crates/rome_js_formatter/Cargo.toml
+++ b/crates/rome_js_formatter/Cargo.toml
@@ -25,3 +25,10 @@ parking_lot = "0.12.0"
 similar = "2.1.0"
 similar-asserts = "1.2.0"
 ctor = "0.1.21"
+quickcheck = "1.0.3"
+quickcheck_macros = "1.0.0"
+iai = "0.1.1"
+
+[[bench]]
+name = "iai"
+harness = false
--- a/crates/rome_js_formatter/src/js/expressions/big_int_literal_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/big_int_literal_expression.rs
@@ -1,5 +1,10 @@
+use std::borrow::Cow;
+
+use crate::formatter_traits::FormatTokenAndNode;
+use crate::utils::string_utils::ToAsciiLowercaseCow;
 use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
 
+use rome_formatter::Token;
 use rome_js_syntax::JsBigIntLiteralExpression;
 use rome_js_syntax::JsBigIntLiteralExpressionFields;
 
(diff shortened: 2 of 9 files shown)


xunilrj:
!bench_formatter

MichaReiser:
> !bench_formatter

This won't show any meaningful result because none of the libraries use (or at least not heavily) bigints.

crates/rome_js_formatter/Cargo.toml
```diff
@@ -25,3 +25,10 @@ parking_lot = "0.12.0"
 similar = "2.1.0"
 similar-asserts = "1.2.0"
 ctor = "0.1.21"
+quickcheck = "1.0.3"
+quickcheck_macros = "1.0.0"
+iai = "0.1.1"
```
MichaReiser, line 30: What's `iai` used for?

crates/rome_js_formatter/src/js/expressions/big_int_literal_expression.rs
```diff
@@ -8,7 +6,10 @@ use rome_js_syntax::JsBigIntLiteralExpressionFields;
 impl ToFormatElement for JsBigIntLiteralExpression {
     fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
         let JsBigIntLiteralExpressionFields { value_token } = self.as_fields();
-
-        value_token.format(formatter)
+        let value_token = value_token?;
+        Ok(crate::utils::format_big_int_literal_token(
+            value_token,
+            formatter,
+        ))
```
MichaReiser, line 13: Nit: I would move the `format_big_int_literal_token` into this `big_int_literal_expression` file. This logic is specific to big ints and can't be shared with any other tokens (and big int literals can only appear in `BigIntLiteralExpression`s

MichaReiser (approved):

crates/rome_js_formatter/Cargo.toml
xunilrj: I am comparing my ```to_ascii_lowercase_cow```with the std ```to_ascii_lowercase```. Result is on the description.

crates/rome_js_formatter/src/js/expressions/big_int_literal_expression.rs
xunilrj: I also prefer. I was following how string does it. What do you think @ematipico?

ematipico: I agree too, I doubt that we will be using this  function in other places

PR #2468, opened 2022-04-20
fix(rome_js_parser): consume `...` in import expression

Boshen:
## Summary

Import expression forbids `...`, but the parser didn't consume it, so it created an invalid AST. 

This was noticed when I tried to format
```
import(x, ...y)
```

and it ended up as:
```
import(x,
...y)
```

## Test Plan
Ran cargo `UPDATE_EXPECT=1 cargo test parser` and inspected the AST

## Rome Playground
https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#aQBtAHAAbwByAHQAKAB4ACwAIAAuAC4ALgB5ACkA

p.s. I didn't understand what recoverable meant when I wrote this part of code three months ago.

### DIFF
--- a/crates/rome_js_parser/src/syntax/expr.rs
+++ b/crates/rome_js_parser/src/syntax/expr.rs
@@ -1235,15 +1235,17 @@ fn parse_primary_expression(p: &mut Parser, context: ExpressionContext) -> Parse
                         error_range_start = p.cur_range().start();
                     }
 
-                    if p.at(T![...]) {
+                    let expr = if p.at(T![...]) {
                         let err = p
                             .err_builder("`...` is not allowed in `import()`")
                             .primary(p.cur_range(), "");
                         p.error(err);
+                        parse_spread_element(p, context)
                     } else {
                         parse_assignment_expression_or_higher(p, ExpressionContext::default())
-                            .or_add_diagnostic(p, js_parse_error::expected_expression_assignment);
-                    }
+                    };
+
+                    expr.or_add_diagnostic(p, js_parse_error::expected_expression_assignment);
 
                     if p.at(T![,]) {
                         p.bump_any();
--- a/crates/rome_js_parser/src/syntax/expr.rs
+++ b/crates/rome_js_parser/src/syntax/expr.rs
@@ -1235,17 +1235,20 @@ fn parse_primary_expression(p: &mut Parser, context: ExpressionContext) -> Parse
                         error_range_start = p.cur_range().start();
                     }
 
-                    let expr = if p.at(T![...]) {
-                        let err = p
-                            .err_builder("`...` is not allowed in `import()`")
-                            .primary(p.cur_range(), "");
-                        p.error(err);
+                    if p.at(T![...]) {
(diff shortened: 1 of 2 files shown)


crates/rome_js_parser/src/syntax/expr.rs
```diff
@@ -1235,15 +1235,17 @@ fn parse_primary_expression(p: &mut Parser, context: ExpressionContext) -> Parse
                         error_range_start = p.cur_range().start();
                     }
 
-                    if p.at(T![...]) {
+                    let expr = if p.at(T![...]) {
                         let err = p
                             .err_builder("`...` is not allowed in `import()`")
                             .primary(p.cur_range(), "");
                         p.error(err);
+                        parse_spread_element(p, context)
```
MichaReiser, line 1243: I'm torn if the parser should convert this expression into a `JsUnknownExpression` but I'm leaning towards that it should. 

The main reason is that this is a syntactical error and, therefore, shouldn't create a valid tree. It will prevent analyzers from deriving incorrect information. 

The counterargument is that it prevents the formatter from formatting the code even though it could potentially format it. I think this is fine and the "solution" for this would be to add recovery logic to the formatting of arguments so that the formatter tries to format all "valid" arguments even if one of them is invalid.

crates/rome_js_parser/test_data/inline/err/import_invalid_args.rast
```diff
@@ -106,7 +108,7 @@ JsModule {
         1: JS_CALL_ARGUMENTS@15..27
           0: L_PAREN@15..16 "(" [] []
           1: JS_CALL_ARGUMENT_LIST@16..26
-            0: JS_SPREAD@16..26
+            0: JS_UNKNOWN_EXPRESSION@16..26
```
Boshen, line 111: Changed to JS_UNKNOWN_EXPRESSION

crates/rome_js_parser/src/syntax/expr.rs
ematipico: Both are valid arguments. I think it's just a matter of weighting pros and cons. My take is that we should give priority to the formatter for the time being, we don't have anything yet around analyzers while the formatter is already out there. 

Analyzers might have more freedom and power to inspect unknown nodes.

ematipico (commented):
It looks good to me but I defer the final approval to @MichaReiser

crates/rome_js_parser/test_data/inline/err/import_invalid_args.rast
```diff
@@ -160,7 +169,7 @@ error[SyntaxError]: `...` is not allowed in `import()`
   ┌─ import_invalid_args.js:2:8
   │
 2 │ import(...["foo"])
-  │        ^^^
+  │        ^^^^^^^^^^
```
MichaReiser, line 172: Nice, this improves the code span too :)

MichaReiser (approved):

PR #2473, opened 2022-04-21
chore(xtask): choose language when generating grammar

IWANABETHATGUY:
## Summary
introduce ability to optional codegen for specific language kind

## Test Plan
1.  run `cargo codegen grammar` should code gen all ungram just like before
2. run `cargo codegen Json css` should warn you `Json` language is not support, but code gen `css` ungram works just
fine
image
image

3. run `cargo codegen json css` should codegen `json` and `css` ungram

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1740,9 +1740,9 @@ dependencies = [
 
 [[package]]
 name = "termcolor"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2dfed899f0eb03f32ee8c6a0aabdb8a7949659e3466561fc0adf54e26d88c5f4"
+checksum = "bab24d30b911b2376f3a13cc2cd443142f0c81dda04c118693e35b3835757755"
 dependencies = [
  "winapi-util",
 ]
@@ -2301,12 +2301,12 @@ name = "xtask_codegen"
 version = "0.0.0"
 dependencies = [
  "anyhow",
- "colored",
  "filetime",
  "git2",
  "pico-args",
  "proc-macro2",
  "quote",
+ "termcolor",
  "ungrammar",
  "ureq",
  "walkdir",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2306,7 +2306,6 @@ dependencies = [
  "pico-args",
  "proc-macro2",
  "quote",
- "termcolor",
  "ungrammar",
  "ureq",
  "walkdir",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2301,6 +2301,7 @@ name = "xtask_codegen"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "colored",
  "filetime",
  "git2",
  "pico-args",
--- a/xtask/codegen/src/ast.rs
+++ b/xtask/codegen/src/ast.rs
@@ -2,6 +2,7 @@
 //! This is derived from rust-analyzer/xtask/codegen
 
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::str::FromStr;
 use std::vec;
 
 use super::{
@@ -40,7 +41,7 @@ pub fn generate_ast(mode: Mode, args: Arguments) -> Result<()> {
     } else {
         arg_list
             .iter()
-            .filter_map(|kind| LanguageKind::from_string(kind.to_str().unwrap()))
+            .filter_map(|kind| LanguageKind::from_str(kind.to_str().unwrap()).ok())
             .collect::<Vec<_>>()
     };
     for kind in codegen_language_kinds {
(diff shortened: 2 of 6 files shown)


IWANABETHATGUY:
for example:
if you want to generate code for `json` only, you could run 
```bash
cargo codegen grammar json
```
image
This would be very useful when you modify `ungram` and maybe codegen more than once

IWANABETHATGUY:
One Question:
Is it alright to instroduce a new crate for printing colorful logging?

IWANABETHATGUY:
By defualt, running
```bash
cargo codegen grammar
```
would codegen for all `ungram` file just like before

Cargo.lock
```diff
@@ -2301,6 +2301,7 @@ name = "xtask_codegen"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "colored",
```
MichaReiser, line 2304: Other crates are using `termcolor`. Another option would be to build on top of `rome_console` and use it's markup for printing.

IWANABETHATGUY: I would use termcolor crate.

xtask/codegen/src/lib.rs
```diff
@@ -52,6 +52,14 @@ pub enum LanguageKind {
...
 
 impl LanguageKind {
+    pub(crate) fn from_string(kind: &str) -> Option<Self> {
+        match kind {
+            "js" => Some(LanguageKind::Js),
+            "css" => Some(LanguageKind::Css),
+            "json" => Some(LanguageKind::Json),
+            _ => None,
+        }
+    }
```
ematipico, line 62: Is there a specific reason why you used a function instead of using `impl FromStr for LanguageKind`?

IWANABETHATGUY: Well, I just don't know such a trait exist, I usually use `From` trait

xtask/codegen/src/lib.rs
```diff
@@ -51,15 +52,20 @@ pub enum LanguageKind {
...
-            "css" => Some(LanguageKind::Css),
-            "json" => Some(LanguageKind::Json),
-            _ => None,
+            "js" => Ok(LanguageKind::Js),
+            "css" => Ok(LanguageKind::Css),
+            "json" => Ok(LanguageKind::Json),
+            _ => Err(()),
         }
     }
+}
```
ematipico, line 66: Great, this will give us more room for handling things:

- we can add `kind.to_lowercase()` so we can cover cases where we write `cargo codegen grammar CSS`
- the error  can be `type Err = String`, so we can actually return an error and tell to the developer **why** it failed. Like we can say "Language not supported, please use: js, css or json"


And we can test by passing some bogus argument `cargo codegen grammar foobar`

IWANABETHATGUY:
I have resolved all issues, any idea on this ? @MichaReiser

xtask/codegen/src/ast.rs
```diff
@@ -12,31 +13,59 @@ use crate::css_kinds_src::CSS_KINDS_SRC;
...
+                |kind| match LanguageKind::from_str(kind.to_str().unwrap()) {
+                    Ok(kind) => Some(kind),
+                    Err(err) => {
+                        println_string_with_fg_color(err, Color::Red).unwrap();
+                        None
+                    }
+                },
+            )
+            .collect::<Vec<_>>()
+    };
```
MichaReiser, line 52: Nit: In my view, `generate_ast` should keep the signature it had before and the handling of CLI arguments should instead be part of `src/main`. 

I would recommend moving `ALL_LANGUAGE_KIND` next to `LanguageKind` and add documentation to `LanguageKind` that devs need to add it to that array.

xtask/codegen/src/lib.rs
```diff
@@ -155,3 +175,13 @@ pub fn to_lower_snake_case(s: &str) -> String {
...
+
+pub fn println_string_with_fg_color(content: String, color: Color) -> std::io::Result<()> {
+    let mut stdout = StandardStream::stdout(ColorChoice::Auto);
+    stdout
+        .set_color(ColorSpec::new().set_fg(Some(color)))
+        .unwrap();
+    println!("{}", content);
+    stdout.set_color(ColorSpec::new().set_fg(None)).unwrap();
+    Ok(())
+}
```
MichaReiser, line 187: This is only related to the CLI, can we move it to `src/main`.

MichaReiser (commented):
Thanks for your contribution.

The colors are a nice addition but also seem to complicate things a bit and increase the compile time of the crate. Something you may want to consider.

ematipico:
Would you mind fill the "Test plan" part? We would like to know how are you testing your new changes, if the new changes handle error cases, etc.

If you want, you can also attach screenshots

IWANABETHATGUY:
> 

I have been updated the test plan

xtask/codegen/src/lib.rs
IWANABETHATGUY: I move this function to a seprate file `termcolorful`,  remove the crate before and using some hardcode `term` color instead,  you chould review it here, xtask/codegen/src/termcolorful.rs

ematipico (approved):

xunilrj (approved):

PR #2476, opened 2022-04-21
refactor(rome_js_formatter): Move node formatting into its own trait

MichaReiser:
Today, the formatting that must happen for each JS node happens inside `Formatter.format_syntax_node` and the `formatter_traits`.

This PR moves the common logic applied by all nodes and tokens into the `ToFormatElement` implementation and renames it to `Format`.

* Rename `ToFormatElement` to `Format` and `to_format_element` to `format`
* Implement `Format` for `JsSyntaxToken`. CSS, JSON etc. can implement their own generic token formatting logic
* Introduce a new `FormatNode` trait (language specific) that implements the formatting of a node and calls into `format_fields` that is implemented for each field.
* Make the `FormatterTraits` node and language independent. Rename to `FormatTraits`.

`Format`, `FormatterTraits`, `FormatResult`, and `FormatError` are now all language independent. However, some still depend on the `Formatter`.

The next step is to move many of the `Formatter`'s helper function out of the Formatter so that it becomes language independent.

## How to review

* The conceptual changes are in `rome_js_formatter/lib` and `format_traits` 
* Review codegen 

## Test Plan

`cargo test`

Deleted a union, a list, and a node formatter implementation and regenerated the formatter files. Verified that the generated files have no compile errors.


crates/rome_js_formatter/src/js/any/array_assignment_pattern_element.rs
```diff
@@ -1,14 +1,14 @@
 //! Generated file, do not edit by hand, see `xtask/codegen`
 
-use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
+use crate::{Format, FormatElement, FormatResult, Formatter};
 use rome_js_syntax::JsAnyArrayAssignmentPatternElement;
-impl ToFormatElement for JsAnyArrayAssignmentPatternElement {
-    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+impl Format for JsAnyArrayAssignmentPatternElement {
```
MichaReiser, line 5: Unions don't need to implement `FormatNode` because the "node" logic will be handled by the inner node.

crates/rome_js_formatter/src/js/lists/array_assignment_pattern_element_list.rs
```diff
@@ -1,9 +1,9 @@
 use crate::utils::array::format_array_node;
-use crate::{FormatElement, FormatResult, Formatter, ToFormatElement};
+use crate::{Format, FormatElement, FormatResult, Formatter};
 use rome_js_syntax::JsArrayAssignmentPatternElementList;
 
-impl ToFormatElement for JsArrayAssignmentPatternElementList {
-    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+impl Format for JsArrayAssignmentPatternElementList {
```
MichaReiser, line 5: I believe it's fine to not implement `FormatNode` for lists because suppressions either apply to the whole wrapper node or a specific element:

```
// rome-ingore
[ a, b, c] // ignores whole array expression
[
  // rome-ignore
  a, // Ignores first node
  b,
  c
]
```

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,99 @@ pub use rome_formatter::{
...
+    fn format(&self, f: &Formatter) -> FormatResult<FormatElement> {
+        Format::format(&**self, f)
+    }
+}
+
+impl<T: ?Sized + Format> Format for RefMut<'_, T> {
+    fn format(&self, f: &Formatter) -> FormatResult<FormatElement> {
+        Format::format(&*(self.deref()), f)
+    }
+}
```
MichaReiser, line 67: I plan to introduce a `format_args[node1, node2]` macro that accepts anything implementing `Format`. This implementations will be necessary for such a trait to work

crates/rome_js_formatter/src/format_traits.rs
```diff
@@ -236,49 +205,7 @@ impl IntoFormatResult for FormatResult<FormatElement> {
...
-            }
-        }
-
-        with(format_elements![
-            formatter.print_leading_trivia(self, TriviaPrintMode::Full),
-            Token::from(self),
-            formatter.print_trailing_trivia(self),
-        ])
-        .into_format_result()
-    }
```
MichaReiser, line 278: Moved to `Format` and `FormatNode`

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
+    T: Format + ?Sized,
+{
+    fn format(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        Format::format(&**self, formatter)
+    }
+}
+
+impl<T> Format for &mut T
+where
+    T: ?Sized + Format,
```
ematipico, line 51: nit: can we keep the same order of traits? `Format + ?Sized`?

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
+        }
+    }
+}
+
+/// Formatting of a JS AST Node. Must be implemented for each JS Ast node.
+///
+/// Implements the JavaScrip
+///
+/// The code-gen generates a [Format] implementation for each `FormatNode`.
+pub trait FormatNode: AstNode<Language = JsLanguage> {
```
ematipico, line 99: As this trait is specific for `JsLanguage`, should we call this trait `FormatJsNode`?

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
+        match self {
+            Ok(value) => value.format(formatter),
+            Err(err) => Err(err.into()),
+        }
+    }
+}
+
+/// Formatting of a JS AST Node. Must be implemented for each JS Ast node.
+///
+/// Implements the JavaScrip
```
ematipico, line 96:
```suggestion
/// Implements the JavaScript
```

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
+///
+/// The code-gen generates a [Format] implementation for each `FormatNode`.
+pub trait FormatNode: AstNode<Language = JsLanguage> {
+    /// Performs the formatting that applies to all nodes.
```
ematipico, line 100: I think can add more description to the API, to help to understand its usage/workflow. We can add that: 
- the default implementation checks suppressions first; 
- then that `format_fields` is called;
- re-implemting `format_node` implies that the developer needs to call `format_fields` in order to actually get a format element, because `format_node` doesn't the actual formatting (well, the default implemetation doesn't)

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
-use rome_rowan::TokenAtOffset;
+use rome_js_syntax::{JsLanguage, JsSyntaxNode, JsSyntaxToken};
+use rome_rowan::{AstNode, TextSize};
 use rome_rowan::{SyntaxError, TextRange};
+use rome_rowan::{SyntaxResult, TokenAtOffset};
 
 /// This trait should be implemented on each node/value that should have a formatted representation
-pub trait ToFormatElement {
-    fn to_format_element(&self, formatter: &Formatter) -> FormatResult<FormatElement>;
+pub trait Format {
```
ematipico, line 35: What's preventing you from having this trait inside the `rome_formatter` crate? It's generic enough to have it there, right?

ematipico: Can we have a doc example of how we should implement and use it? We them for `format_traits` so I think it's worth having the same here.

crates/rome_js_formatter/src/format.rs
```diff
@@ -0,0 +1,1416 @@
+use crate::{Format, FormatElement, FormatNode, FormatResult, Formatter};
```
ematipico, line 1: If this file is autogenerated, we shold add a comment at the top saying so, like all the others generated files that should not be changed manually.

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
+impl<T> Format for &mut T
+where
+    T: ?Sized + Format,
+{
+    fn format(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        Format::format(&**self, formatter)
+    }
+}
+
+impl<T: ?Sized + Format> Format for Ref<'_, T> {
```
ematipico, line 58: Which case does this impl cover? Precesily, I would like to see some code where we needed this `impl`.

crates/rome_js_formatter/src/lib.rs
```diff
@@ -22,14 +26,108 @@ pub use rome_formatter::{
...
+    fn format(&self, f: &Formatter) -> FormatResult<FormatElement> {
+        Format::format(&**self, f)
+    }
+}
+
+impl<T: ?Sized + Format> Format for RefMut<'_, T> {
```
ematipico, line 64: Which case does this impl cover? Precesily, I would like to see some code where we needed this `impl`.

ematipico (commented):
One step forward to make things more generic! I added some feedback and few questions where I have some doubts!

crates/rome_js_formatter/src/lib.rs
MichaReiser: I would stick to `FormatNode` so that the trait has the same name across all languages. This simplifies

* Documentation: Only one trait to mention (users will automatically import the right one that belongs to their crate)
* Code gen: No language specialisation.

crates/rome_js_formatter/src/lib.rs
MichaReiser: Mainly the `Formatter` that isn't language agonist yet.

crates/rome_js_formatter/src/lib.rs
MichaReiser: I can delete it for now but will be needed once we have a `format_args!` so that you can pass any `Ref`. These implementations are copy pasted from `Debug` + `Display`

crates/rome_js_formatter/src/lib.rs
MichaReiser: I'll add some more documentation but some of what you described is evident when looking at the implementation.

crates/rome_js_formatter/src/lib.rs
MichaReiser: I added an example on how to implement it. How to use it is more difficult because our formatter currently only supports formatting nodes (the root must be a node).

ematipico: Got it! Thank you!

crates/rome_js_formatter/src/lib.rs
ematipico: It is and you're right. I was a bit concerned/unsettled about having two functions instead of one and how they are related to each other. I presume we will have to follow the same patter for `FormatNode` of other languages? Or are we free to do whatever we see fit?

MichaReiser: Having a `FormatNode` trait for every language will help to keep the code-gen simple. But a language (for example JSON?) can decide to only have a `format_node` method.

ematipico (approved):
Looks good to me! Thank you for addressing my questions! This is great!

leops (approved):
I still wonder if it would have been possible this trait hierarchy using only generics somehow instead of relying on codegen, but until specialization gets stabilized (and perhaps even after) I think this is probably the only way this can be expressed with the Rust type system

MichaReiser:
> I still wonder if it would have been possible this trait hierarchy using only generics somehow instead of relying on codegen, but until specialization gets stabilized (and perhaps even after) I think this is probably the only way this can be expressed with the Rust type system

I would love that if you find a way. I even tried to go fancy (at least as far as I'm capable of) and played around with `From` and `Into` to remove the `&T` impl from `Format` but nothing helped. I in the end decided that the code gen is in place and using it to work around Rust's limitation seems reasonable enough.

PR #2477, opened 2022-04-21
fix(rome_rowan): fix syntaxtokentext display infinite recursion

xunilrj:
Fix the infinite recursion of the playground.

image

### DIFF
--- a/crates/rome_rowan/src/syntax/token.rs
+++ b/crates/rome_rowan/src/syntax/token.rs
@@ -48,6 +48,22 @@ impl<L: Language> SyntaxToken<L> {
     }
 
     /// Returns the text of a token, including all trivia as an owned value.
+    ///  
+    /// ```
+    /// use rome_rowan::*;
+    /// use rome_rowan::raw_language::{RawLanguage, RawLanguageKind, RawSyntaxTreeBuilder};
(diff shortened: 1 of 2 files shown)


xunilrj:
> Do you think we are able to add a test that cover the fix? If so that would be great, so we cover possible regressions. If not, it's fine!

Done.

crates/rome_rowan/src/syntax/token.rs
```diff
@@ -48,6 +48,22 @@ impl<L: Language> SyntaxToken<L> {
...
+    ///         RawLanguageKind::LET_TOKEN,
+    ///         "\n\t let \t\t",
+    ///         &[TriviaPiece::whitespace(3)],
+    ///         &[TriviaPiece::whitespace(3)],
+    ///     );
+    /// }).first_token().unwrap();
+    /// assert_eq!("\n\t let \t\t", token.token_text());
+    /// assert_eq!(format!("{}", "\n\t let \t\t"), format!("{}", token.token_text()));
+    /// assert_eq!(format!("{:?}", "\n\t let \t\t"), format!("{:?}", token.token_text()));
+    /// ```
```
MichaReiser, line 66: Thanks @xunilrj and sorry for this!

ematipico (approved):
Thank you Daniel! Let's 🚢 it!

PR #2478, opened 2022-04-21
refactor(rome_js_formatter): Remove `format_root` from `Formatter`

MichaReiser:
This PR removes the `format_root` method from the `Formatter` so that it can be used to format any value implementing `Format`.

* Rename existing `format_node` method to `format_sub_tree`
* Move `FormatError` and `FormatResult` to `rome_formatter`
* Rename `Formatted` to `Printed`
* `format_node` to reduce `Formatted` which is the formatted IR that hasn't been printed. Can be printed by calling `print`
* `Formatted` replaces `to_format_element`
* Introduce new `format` method to format any type implementing `Format`

### DIFF
--- a/crates/rome_js_formatter/src/js/any/array_assignment_pattern_element.rs
+++ b/crates/rome_js_formatter/src/js/any/array_assignment_pattern_element.rs
@@ -1,6 +1,7 @@
 //! Generated file, do not edit by hand, see `xtask/codegen`
 
-use crate::{Format, FormatElement, FormatResult, Formatter};
+use crate::{Format, Formatter};
+use rome_formatter::{FormatElement, FormatResult};
 use rome_js_syntax::JsAnyArrayAssignmentPatternElement;
 impl Format for JsAnyArrayAssignmentPatternElement {
     fn format(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -511,8 +511,9 @@ fn format_file(params: FormatFileParams) -> FormatResult {
             });
         }
 
-        let result = rome_js_formatter::format(params.options, &root.syntax())
-            .with_file_id_and_code(params.file_id, "Format")?;
+        let result = rome_js_formatter::format_node(params.options, &root.syntax())
+            .with_file_id_and_code(params.file_id, "Format")?
+            .print();
 
         let output = result.as_code().as_bytes();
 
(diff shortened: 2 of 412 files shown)


crates/rome_js_formatter/src/js/any/array_assignment_pattern_element.rs
```diff
@@ -1,6 +1,7 @@
 //! Generated file, do not edit by hand, see `xtask/codegen`
 
-use crate::{Format, FormatElement, FormatResult, Formatter};
+use crate::{Format, Formatter};
+use rome_formatter::{FormatElement, FormatResult};
```
ematipico, line 4: I believe all these changes are can be avoided if we correctly re-export their usage from `rome_js_formatter::lib`. I know it because I did it while moving stuff from `rome_js_formatter` to `rome_formatter` and avoiding to apply to many changes.

Essentially we can keep 

```rs
use crate::{Format, FormatElement, FormatResult, Formatter};
```

for all the these autogenerated files and the rest of the nodes

ematipico (changes requested):
I think the quantity of changes can be drastically reduced by keeping the majority of `use create::` and by re-exporting `FormatResult` and `FormatError` inside `rome_js_formatter::lib`

crates/rome_js_formatter/src/js/any/array_assignment_pattern_element.rs
MichaReiser: This has actually been an intentional change. These re-exports get out of sync and the result is that there are now many places where we import some types from `rome_formatter` and `rome_js_formatter`. In my view, all types would ideally be imported from `rome_formatter` except they belong to `rome_js_formatter`.

ematipico: Which is fair, then do you plan to do the same with the rest? (`FormatElement`, `indent`, `group_elements`, traits, etc.)

MichaReiser: Don't know yet and these are more of a pain because my IDE doesn't support an automatic refactor. I want to wait a little longer until I have decided whatever I want to rework the formatter IR in which case it might be necessarily to have a `v1::FormatElement` and `v2::FormatElement` during the migration period and I rather not do this rename twice

ematipico: Thank you for clarifying!

ematipico (approved):

leops (approved):

PR #2481, opened 2022-04-22
refactor(rome_js_formatter): Track token offsets instead of tokens

MichaReiser:
The `Formatter` tracks all formatted tokens in debug builds to catch if the same token gets formatted twice or not formatted at all.

The current implementation tracks the `JsSyntaxToken`s. The downside of doing this is that a language-agnostic implementation needs to track `SyntaxToken<L>` which:

* has the consequence that `Formatter` must be generic over `L`
* which in turn has the consequence that `Format` must be generic over `L`

This PR changes the node tracking to instead only track the absolute offsets of the formatted nodes rather than holding on to the node itself. This removes the need for a generic `Language` parameter and provides the same functionality.

A next step is to move out all `format_` functions from the `Formatter` and put them into standalone functions / extension traits. 

## Test

* Commented out the `track_replaced` call in `format_replaced` and verified that some tests now panic with a message that token X wasn't formatted.
* Added one an additional `track_replaced` call to `format_replaced` and verified that tests are now failing with a message saying that the token X was formatted twice.

### DIFF
--- a/crates/rome_js_formatter/src/printed_tokens.rs
+++ b/crates/rome_js_formatter/src/printed_tokens.rs
@@ -1,0 +1,85 @@
+use rome_rowan::{Language, SyntaxNode, SyntaxToken, TextSize};
+use std::collections::BTreeMap;
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+enum TokenTrackMode {
+    Replaced,
+    Formatted,
+    Verbatim,
+}
+
+/// Tracks the ranges of the formatted (including replaced or tokens formatted as verbatim) tokens.
+///
+/// This implementation uses the fact that no two tokens can have an overlapping range to avoid the need for an interval tree.
+/// Thus, testing if a token has already been formatted only requires testing if a token starting at the same offset has been formatted.
+#[derive(Debug, Clone, Default)]
+pub(crate) struct PrintedTokens {
+    /// Key: Start of range, value: whatever this is formatted or replaced
+    offsets: BTreeMap<TextSize, TokenTrackMode>,
+}
+
+impl PrintedTokens {
+    fn track_token<L: Language>(&mut self, token: &SyntaxToken<L>, mode: TokenTrackMode) {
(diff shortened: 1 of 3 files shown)


crates/rome_js_formatter/src/printed_tokens.rs
```diff
@@ -0,0 +1,85 @@
...
+    offsets: BTreeMap<TextSize, TokenTrackMode>,
+}
+
+impl PrintedTokens {
+    fn track_token<L: Language>(&mut self, token: &SyntaxToken<L>, mode: TokenTrackMode) {
+        let range = token.text_trimmed_range();
+
+        if let Some(previous_mode) = self.offsets.insert(range.start(), mode) {
+            panic!("You tried to print the token '{token:?}' twice, and this is not valid. Printed now as {mode:?}, previously printed as {previous_mode:?}.");
+        }
```
ematipico, line 27: Can we have a cases where two tokens are formatted with the same `mode`?

ematipico (changes requested):
Before we had the message "You tried to print the token '{:?}' twice, and this is not valid" which was somehow descriptive and helped what went wrong. 

Now some message are different, which makes sense. Although they don't provide much context sometimes. Would you mind expand [this section](https://github.com/rome/tools/blob/main/crates/rome_js_formatter/docs/write_tests.md#identify-issues) , explaining how a developer is supposed to do once they find these new messages?

MichaReiser:
> Before we had the message "You tried to print the token '{:?}' twice, and this is not valid" which was somehow descriptive and helped what went wrong.
> 
> Now some message are different, which makes sense. Although they don't provide much context sometimes. Would you mind expand [this section](https://github.com/rome/tools/blob/main/crates/rome_js_formatter/docs/write_tests.md#identify-issues) , explaining how a developer is supposed to do once they find these new messages?

I removed the `mode` and extended the panic message

ematipico (approved):
Way better! Thank you!

leops (approved):

PR #2485, opened 2022-04-22
chore(rome_lsp):  migration to tower_lsp crate

IWANABETHATGUY:
## Summary
1. migrate lsp server crate `lspower` to `tower_lsp`
2. add launch.json which is convenient for debugging, unlike before we need to compile and reinstall `.vsix` every modify,
we could now use `Launch Client` 
image

## Test Plan
1. run `cargo build --package rome_lsp`, build our `rome_lsp` server   
2. cd `editors/code`, and ` code .` open this directory as workspace root
3. switch to this pannel, and click `Launch Cilent` 
image
4. open any js,  ts, file, use `Format document` command and choose `rome`, your file should be formatted correctly
5. Using `Rome: show Syntax tree` command, you  should see syntax tree on another split window
image

### DIFF
--- a/editors/vscode/src/main.ts
+++ b/editors/vscode/src/main.ts
@@ -1,5 +1,5 @@
-import { ExtensionContext, Uri, window, workspace } from 'vscode';
-import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from 'vscode-languageclient/node';
+import { ExtensionContext, Uri, window, workspace } from "vscode";
+import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from "vscode-languageclient/node";
 import { setContextValue } from "./utils";
 import { Session } from "./session";
 import { syntaxTree } from "./commands/syntaxTree";
@@ -10,7 +10,9 @@ let client: LanguageClient;
 const IN_ROME_PROJECT = "inRomeProject";
 
 export async function activate(context: ExtensionContext) {
-	const command = process.env.DEBUG_SERVER_PATH || (await getServerPath(context));
+	const command = process.env.DEBUG_SERVER_PATH || (
+		await getServerPath(context)
+	);
 	if (!command) {
 		await window.showErrorMessage(
 			"The Rome extensions doesn't ship with prebuilt binaries for your platform yet. " +
@@ -28,11 +30,7 @@ export async function activate(context: ExtensionContext) {
 	// only override serverOptions.options when developing extension,
 	// this is convenient for debugging
 	if (process.env.DEBUG_SERVER_PATH) {
-		serverOptions.options = {
-			env: {
-				...process.env
-			},
-		};
+		serverOptions.options = { env: { ...process.env } };
 	}
 
 	const clientOptions: LanguageClientOptions = {
@@ -54,7 +52,7 @@ export async function activate(context: ExtensionContext) {
 	client.start();
 }
 
-type Architecture = 'x64' | 'arm64';
+type Architecture = "x64" | "arm64";
 
 type PlatformTriplets = {
 	[P in NodeJS.Platform]?: {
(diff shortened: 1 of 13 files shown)


ematipico:
@IWANABETHATGUY Thank you for you contribution. Have you actually tested that the VSCode extension still works after the migration?

IWANABETHATGUY:
> @IWANABETHATGUY Thank you for you contribution. Have you actually tested that the VSCode extension still works after the migration?

Yeah, i have tested locally. IIRC, @yassere would be familar to rome_lsp, would you like to help for testing locally?

ematipico:
Could you fill out the test plan part of the template and explain the steps you did to test it?

My concern is that the current deployed VSCode extension might not pick up the new path.

IWANABETHATGUY:
> Could you fill out the test plan part of the template and explain the steps you did to test it?
> 
> My concern is that the current deployed VSCode extension might not pick up the new path.

Hi, I have been updated the test plan.

editors/vscode/src/main.ts
```diff
@@ -25,6 +27,12 @@ export async function activate(context: ExtensionContext) {
 		transport: TransportKind.stdio,
 	};
 
+	// only override serverOptions.options when developing extension,
+	// this is convenient for debugging
+	if (process.env.DEBUG_SERVER_PATH) {
+		serverOptions.options = { env: { ...process.env } };
+	}
+
```
ematipico, line 35: What does it do this new snippet? Why is it convenient? Maybe we should explain why it is so also new developers will know

ematipico (approved):
Looks good to me. Just one minor comment

editors/vscode/src/main.ts
IWANABETHATGUY: 1. Before, every time we modify the client package, we need to rebuild vscode extension and install, and for now, we could using `Launching Client` or press `F5` to open a seprate debug window and doing some check, finally we could bundle the vscode and do some final check. 
2. Passing such varialbe via `Launch.json`, you need not to add extra environment variable or change the `setting.json` 
`rome.lspBin`,

IWANABETHATGUY:
I found that i don't commit the `.vscode/launch.json` because the `.gitignore`,  which should be 
```json
// A launch configuration that compiles the extension and then opens it inside a new window
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "extensionHost",
      "request": "launch",
      "name": "Launch Client",
      "runtimeExecutable": "${execPath}",
      "args": [
        "--extensionDevelopmentPath=${workspaceRoot}"
      ],
      "outFiles": [
        "${workspaceRoot}/out/**/*.js"
      ],
      "preLaunchTask": {
        "type": "npm",
        "script": "compile"
      },
      "env": {
        "DEBUG_SERVER_PATH": "${workspaceRoot}/../../target/debug/rome_lsp"
      }
    },
  ],
}
```
You could also reference `rust-analyzer` https://github.com/rust-lang/rust-analyzer/blob/master/.vscode/launch.json

IWANABETHATGUY:
Would you mind if i remove the `.vscode` from `.gitignore` ?

editors/vscode/src/main.ts
ematipico: What do you think if we add a more descriptive comment that explains what you just said to me here? If a new contributor will see that comment, they will wonder **why it is convenient**. We should give an explanation

ematipico:
> Would you mind if i remove the `.vscode` from `.gitignore` ?

Maybe we can do that in another PR? It seems that it's not related to `tower_lsp` refactor. It might be better to not mix things up, so we can track two separate tasks in two different PRs and have two separate commits.

editors/vscode/src/main.ts
IWANABETHATGUY: I have added an extra comment to explain why this variable is convenient.

Issue #2488, opened 2022-04-25
☂️ Symbol/Scope Resolution


xunilrj:
This issue is tracking all the implementation to achieve two tasks:

1 - Allow "go to definition" on the editor;
2 - Allow to warn against "unused Variables" when linting;

See this discussion about the semantic model architecture: https://github.com/rome/tools/discussions/2614
See this discussion about the architecture on how the semantic model is going to be used inside linters: https://github.com/rome/tools/discussions/2603

#### Steps

This is my suggestion for implementing the two tasks above. It is actually a progressive implementation:

1 - We do not offer a semantic façade at all. "unusedVariables" lint would just consume the parsing tree, and iterate it to find unused variables;
2 - We cache this iteration somehow to make the lint have a decent performance;
3 - We build the "service/context" architecture that will be passed into the lint;
4 - We store the "service/context" inside ```NodeData```;
5 - We create the untyped Semantic Tree;
6 - We change the "unusedVariable" lint rule to use the semantic tree.
7 - We implement the typed Semantic Tree. Probably just the first ```SemanticJsVariableDeclaration```.
8 - We change the "unusedVariable" lint rule to query for the type semantic tree.

#### Tasks 

- [ ] Parse Microsoft/ts .symbols file (see https://github.com/rome/tools/pull/2588)
    - [x] basic coverage test running
    - [ ] achieve +90% of passess
- [x] Generate Symbols Events
    - [x] DeclarationFound
    - [x] ReferenceFound
- [x] Implement simple Scope resolution
    - [x] ScopeStarted
    - [x] ScopeEnded
- [x] Support Hoisting
    - [x] ReferenceSolved
- [ ] LSP
    - [ ] F12 - go to definition
- [x] Lint
    - [x] lint for "unused variable"

ematipico:
Will this be part of the parsing phase or another another phase after the parsing phase?

xunilrj:
Another pass.
The final architecture, we are still discussing.

So for these tasks, it will actually be another function that will be called manually when testing.

xunilrj:
To reach 90% we need to deal with a lot of edge cases that are not relevant for the moment.

xunilrj:
"Goto definition" will only be relevant when we run and the only LSP (non vscode).

PR #2489, opened 2022-04-25
feature(rome_js_parser): Symbols Test Suite and SymbolIterator

xunilrj:
This PR will not be merged, because I will break it down into multiple ones. See
- https://github.com/rome/tools/pull/2588

see https://github.com/rome/tools/issues/2488

This PR is the foundation for the symbol/scope resolution. It now:
- parses Microsoft/ts .symbols file;
- iterate the CST and solve references (local scope resolution);
- implement a framework to test scope resolution.

#### Microsoft ts .symbols file

We are running only files without any errors.
We are more than 90% compliant and some divergences I am not entirely sure I agree with ```ts```.

```
symbols/msts: Ran 5946 tests in 19.05s
│      5946 │   5391 │    550 │      5 │    90.67 │
```

All ```parse_*``` functions are simplified versions of ```nom```, just to avoid including another crate.

#### Iterator

This PR also creates the ```SymbolIterator``` that lazily returns a symbol and if it is a declaration or a reference. We still do not solve references (they point to nowhere).

I am not sure "symbol" is the best word here. Will revisit this later.

#### .symbols files

Now we also have our .symbols file. That is why this PR has +500 new files. They are just a dumb dump of the symbols that we are currently finding.

#### Scope Resolution assertion

There is no easy way to test for scope resolution. Working with assertions on ranges to check if a reference points to the correct symbol would be not only a disaster but would spoil the day of anyone working on this. For this reason I came up with the following solution:

```js
function f(a/*#A1*/) {
    let b/*#B*/ = 1;
    let c/*#C1*/ = b/*@B*/ + 1;

    if (c/*@C1*/ == 1) {
         let c/*#C2*/ = 2;
        console.log(b/*@B*/, c/*@C2*/);
    }
}
```

We decorate symbols and references with comments following this pattern:
- ```/*#NAME*/``` - this assert that the corresponding token is introducing a new symbol. There is no correlation between the symbol text and the assertion labe.; I can this "declaration assertion";
- ```/*@NAME*/``` - this asserts that the corresponding token is a reference and it points to a symbol declaration that has an attached declaration assertion. I can this "reference assertion";
- ```/*@?*/``` - this is a specific reference assertion that points to nothing; I am calling this "unreferenced symbol" to avoid the confusing (but philosophical interesting) "unreferenced reference".

A huge chunk of code was reserved to generate nice errors when testing the scope resolution. For example:

image

#### LSP

This PR also implements F12 (goto definition) for vscode.

romef12

#### Hoisting

This PR also implements preliminary support for hoisting.

image


ematipico (commented):
I am not sure what I am looking at to be honest. I presume we have now a new test coverage?  Why is it not running?

xunilrj:
>  Why is it not running?

I have not put it to run as default.
It is now.

Not sure why the comment is not updated, though.

xtask/coverage/src/symbols/msts.rs
```diff
@@ -0,0 +1,252 @@
...
+                }
+
+                println!();
+            }
+        }
+
+        if expected.symbols.len() != actual.len() {
+            TestRunOutcome::IncorrectlyErrored {
+                files: t,
+                errors: vec![],
```
MichaReiser, line 91: It would be helpful to print a diff if the test errors Same only  one 103.

xtask/coverage/src/symbols/msts.rs
```diff
@@ -0,0 +1,252 @@
...
+    path: String,
+    decls: Vec<Decl>,
+}
+
+struct SymbolsFile {
+    code_file: PathBuf,
+    symbols: Vec<Symbol>,
+}
+
+fn parse_decl(input: &str) -> Option<(&str, Decl)> {
```
MichaReiser, line 162: Nit: Could help to rename this to `parse_symbol_file_declaration` or `parse_expected_decl` to make it clear that this isn't about parsing a JS declaration.

crates/rome_js_syntax/src/generated/nodes.rs
```diff
@@ -9134,24 +9134,116 @@ pub enum JsAnyArrayAssignmentPatternElement {
     JsArrayHole(JsArrayHole),
     JsAssignmentWithDefault(JsAssignmentWithDefault),
 }
+impl JsAnyArrayAssignmentPatternElement {
+    pub fn as_js_any_assignment_pattern(&self) -> Option<&JsAnyAssignmentPattern> {
```
MichaReiser, line 9138: Nit: Could be worth splitting the addition of the `as_*` methods into its own PR to reduce the size of this PR

crates/rome_js_parser/src/symbols.rs
```diff
@@ -0,0 +1,190 @@
...
+            range: node.text_range(),
+        }),
+        JsSyntaxKind::JS_IDENTIFIER_ASSIGNMENT
+        | JsSyntaxKind::JS_SUPER_EXPRESSION
+        | JsSyntaxKind::JS_THIS_EXPRESSION
+        | JsSyntaxKind::JS_MODULE_SOURCE => Some(Symbol::Reference {
+            name: node.text_trimmed().to_string(),
+            range: node.text_range(),
+            declared_at: None,
+        }),
```
MichaReiser, line 53: Nit: Create a `Symbol::reference(node)` helper function to avoid repeating the `node.text_trimmed` and `node.text_range`. calls.

crates/rome_js_parser/src/symbols.rs
```diff
@@ -0,0 +1,190 @@
...
+
+pub struct SymbolIterator(VecDeque<JsSyntaxNode>);
+
+fn extract_symbol(node: JsSyntaxNode) -> Option<Symbol> {
+    match node.kind() {
+        JsSyntaxKind::JS_IDENTIFIER_BINDING
+        | JsSyntaxKind::TS_IDENTIFIER_BINDING
+        | JsSyntaxKind::JS_LITERAL_EXPORT_NAME => Some(Symbol::Declaration {
+            name: node.text_trimmed().to_string(),
+            range: node.text_range(),
```
MichaReiser, line 44: Should this be `text_trimmed_range`? If so, how about using `SyntaxNodeText`? Or directly storing the node in the `declaration`/`reference` rather than a range and `String` (also allows to jump back to the node)

crates/rome_js_parser/src/symbols.rs
```diff
@@ -0,0 +1,190 @@
...
+        }),
+        JsSyntaxKind::JS_IDENTIFIER_ASSIGNMENT
+        | JsSyntaxKind::JS_SUPER_EXPRESSION
+        | JsSyntaxKind::JS_THIS_EXPRESSION
+        | JsSyntaxKind::JS_MODULE_SOURCE => Some(Symbol::Reference {
```
MichaReiser, line 49: What's the definition of a reference? `JS_MODULE_SOURCE` is a string literal that isn't referencing any identifier.

crates/rome_js_parser/src/symbols.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+
+pub struct SymbolIterator(VecDeque<JsSyntaxNode>);
+
+fn extract_symbol(node: JsSyntaxNode) -> Option<Symbol> {
+    match node.kind() {
+        JsSyntaxKind::JS_IDENTIFIER_BINDING
+        | JsSyntaxKind::TS_IDENTIFIER_BINDING
+        | JsSyntaxKind::JS_LITERAL_EXPORT_NAME => Some(Symbol::Declaration {
```
MichaReiser, line 42: `LiteralExportName` (and `LiteralImportName`) are the identifier names in the imported module, but not of the current module. I don't think this should be part of the current modules scope.

crates/rome_js_parser/src/symbols.rs
```diff
@@ -0,0 +1,190 @@
...
+            let parent_kind = node.parent()?.kind();
+            let parent_ok = matches!(
+                parent_kind,
+                JsSyntaxKind::JS_STATIC_MEMBER_ASSIGNMENT
+                    | JsSyntaxKind::JS_STATIC_MEMBER_EXPRESSION
+                    | JsSyntaxKind::TS_EXPORT_AS_NAMESPACE_CLAUSE
+                    | JsSyntaxKind::TS_QUALIFIED_MODULE_NAME
+                    | JsSyntaxKind::TS_QUALIFIED_NAME
+            );
+            parent_ok.then(|| Symbol::Reference {
```
MichaReiser, line 106: Please add additional documentation to the definition of `Reference`. How is reference defined? How to distinguish between a reference and a property/member?  

A name isn't a reference in that sense (that's why it isn't a `JsIdentifierReference`) but a property name. I'm not sure if members should be part of the symbol table.

crates/rome_js_parser/src/symbols.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+        _ => None,
+    }
+}
+
+impl Iterator for SymbolIterator {
+    type Item = Symbol;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(node) = self.0.pop_front() {
```
MichaReiser, line 172: Isn't the `VecDequeue` the same as using `node.descendants`?

MichaReiser:
Great to see that we're getting close to have symbol resolution. 

This PR needs some clarification on how it defines a `Reference` and what it considers to be a reference.

Issue #2475, opened 2022-04-21
📎 Improve the formatting of call chains


leops:
### Description

Currently the call expression formatter makes a decision on whether or not to introduce a new group for a "call chain" by partially emulating the way the printer works and inspecting the emitted IR to check for hard line breaks. This logic breaks down if the content of the expression contain hard or soft line breaks depending on the input source code (like in the change proposed in #2458), since the formatting of the entire expression is no longer stable between runs.

Since we're aiming for a better compatibility with Prettier, the fix for this would be to use the logic Prettier is using to decide on how to print call chains as a direct reference for our implementation instead of the current ad-hoc solution.

ematipico:
I believe this is what Prettier does to understand when a call chain should break: https://github.com/prettier/prettier/blob/main/src/language-js/print/member-chain.js#L381-L405

PR #2490, opened 2022-04-25
fix(rome_formatter): better grouping of call chain

ematipico:
## Summary

Part of https://github.com/rome/tools/issues/2475 

This PR doesn't close the bug around the call chain, as there are other limitations of our printer and architecture that don't allow us to achieve 100% compatibility with prettier. This will be tackled in a different PR to avoid to add too much logic to this PR.

The main focus of this PR is to add a better heuristic of **when** a call chain should break on multiple lines. Prettier has an extensive heuristic where it checks:
1. if there's a group that will break
2.  if the call chain contains call expressions (at least two) that contains "complex" arguments
3. if the call chain has comments
4. 1. if the last group breaks and it has a function/arrow function with arguments

The first logic was already in our assets and it works as it is.

The "complex" arguments have been implemented using a new enum called `SimpleArgument` argument, which checks if each argument is not complex. Generally speaking, an argument is "simple" when is literal. When it's not, then sometimes we go further and check inside these expressions. The match `depth` of these checks is 2, after which we consider the argument as complex. Refer to Prettier's implementation: https://github.com/prettier/prettier/blob/a9de2a128cc8eea84ddd90efdc210378a894ab6b/src/language-js/utils/index.js#L802-L886

 The check on comments was not perfect, because our algorithm is slightly different compared to Prettier, especially how we handle the "head" group and the "tail" group. The head usually is detached from the rest of the group, but it should be included when we check for comments. Only *trailing* comments for the head group.

There's also one more logic that has been added, which is the "merging" logic. This is another heuristic added by prettier where we have some cases where the first element of the "tail" group has to be part of the "head".

This an example:

```js
Object.
	.keys(foo)
	.map(bar)
	.forEach(lorem)
	.filter(ipsum)
```

In this case, the `.keys(foo)` groups has to be part of the head, resulting in:

```js
Object.keys(foo)
	.map(bar)
	.forEach(lorem)
	.filter(ipsum)
```

The criteria are:
- the name starts with a capital letter, e.g `Array`, `Object`, `Promise`, etc.
- the length of the name is lower then the `tab_width`, e.g. `d3`,
- the name starts with `_` or `$`, e.g. `_forEach`, `_.forEach`, etc.

## Test Plan

I added new tests to cover some of the cases explained above. I also checked some of the screenshot of the prettier test suite.

As explained before, the fix is not finished and we don't cover yet long call expressions, meaning that:

```js
lorem.ipsum().looooooooooooooooooooooooooong().looooooooooooooooooooooooooong().looooooooooooooooooooooooooong();
```

Doesn't break yet


crates/rome_js_formatter/tests/specs/js/module/arrow/arrow_nested.js.snap
```diff
@@ -1,6 +1,8 @@
 ---
 source: crates/rome_js_formatter/tests/spec_test.rs
+assertion_line: 242
```
leops, line 3: The diffs that only change the snapshot header could be discarded (unless this has changed in the latest version of insta ?)

crates/rome_js_formatter/tests/specs/prettier/js/method-chain/fluent-configuration.js.snap
```diff
@@ -30,12 +30,10 @@ domain
 # Output
 ```js
 domain.concept("Page").val("title", "string").vals("widgets", "Widget");
-domain
-  .concept("Widget")
-  .val("title", "string")
-  .val("color", "Color")
-  .val("foo", "Foo")
-  .val("bar", "Bar");
+domain.concept("Widget").val("title", "string").val("color", "Color").val(
```
leops, line 33: This one doesn't look right, but I think it would need the `conditionalGroup` element to be solved

crates/rome_js_formatter/src/utils/call_expression.rs
```diff
@@ -419,16 +576,47 @@ impl From<FlattenItem> for FormatElement {
...
+            formatter,
+            in_expression_statement,
+            groups: Vec::new(),
+            current_group: Vec::new(),
+            cutoff: 2,
+        }
+    }
+
+    /// This function checks if the current grouping should be merged with the first group.
+    pub fn should_merge(&mut self, head_group: &HeadGroup) -> SyntaxResult<bool> {
```
leops, line 612: Why is this method taking `&mut self` ?

crates/rome_js_formatter/src/utils/call_expression.rs
```diff
@@ -496,15 +699,185 @@ impl Groups {
...
+
+    /// Checks if the groups contain comments.
+    pub fn has_comments(&self) -> bool {
+        let has_leading_comments = if self.groups.len() > 1 {
+            // SAFETY: access guarded by the previous check
+            self.groups
+                .iter()
+                .flat_map(|item| item.iter())
+                .enumerate()
+                .filter_map(|(index, item)| if index == 0 { None } else { Some(item) })
```
leops, line 728: This could be written as `.skip(1)`

crates/rome_js_formatter/src/utils/call_expression.rs
ematipico: Old code, good catch!

crates/rome_js_formatter/src/utils/simple.rs
```diff
@@ -224,3 +226,291 @@ fn is_simple_parameter(node: JsFormalParameter) -> SyntaxResult<bool> {
...
+/// [JsThisExpression]: [rome_js_syntax::JsThisExpression]
+/// [JsIdentifierExpression]: [rome_js_syntax::JsIdentifierExpression]
+/// [JsSuperExpression]: [rome_js_syntax::JsSuperExpression]
+/// [is_simple_static_member_expression]: [Simple::is_simple_static_member_expression]
+/// [is_simple_array_expression]: [Simple::is_simple_array_expression]
+/// [JsUnaryExpression]: [rome_js_syntax::JsUnaryExpression]
+/// [TsNonNullAssertionExpression]: [rome_js_syntax::TsNonNullAssertionExpression]
+/// [Prettier]: https://github.com/prettier/prettier/blob/a9de2a128cc8eea84ddd90efdc210378a894ab6b/src/language-js/utils/index.js#L802-L886
+#[derive(Debug)]
+pub(crate) enum SimpleArgument {
```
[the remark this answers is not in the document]
ematipico: Because the majority of the internal functions are specific for the call chain logic, and should not be mixed with the other "simple" function. 

For example the internal function `is_simple_unary_expression` should not be used in other occasions, but only when we call `SimpleArgument::is_simple`. I was conflicted if having this code here or inside `call_expression.rs`.

That's why I preferred to have and intermediate enum, where the details of the logic are closed.

Maybe it makes more sense having outside of the `simple` module?  What do you think?

crates/rome_js_formatter/tests/specs/js/module/arrow/arrow_nested.js.snap
ematipico: Yeah I was using the old version at the beginning. Then I updated to the latest but it didn't revert the change... I will do it manually

crates/rome_js_formatter/src/utils/simple.rs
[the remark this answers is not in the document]
leops: The `simple` module is already a loosely-tied collection of functions that check for an arbitrary "complexity threshold" below which a certain node may bypass the full formatting path, an a sense these are all internal functions that should not be used outside of a specific context anyway. I think the existing ones have to be in this shared module because some of them are called from multiple place in the formatting logic, but if the new methods are only used in the formatting of call expressions it might be better to put them directly alongside of this logic (maybe in a submodule of `call_expression` since this logic is becoming quite large)

leops (approved):

PR #2491, opened 2022-04-25
refactor(rome_analyze): syntax mutation and factories API

leops:
## Summary

This PR is an exploratory work on the design of a CST-based diagnostics and code actions infrastructure, using new mutations and factory methods backed by a fully immutable syntax tree. An outline of the changes:

- Removed the existing mutation system from Rowan SyntaxNodes (red nodes): these nodes are now fully immutable which simplifies their internal representation as the library no longer needs to maintain a linked list of live nodes and can use the `Rc` reference counter pointer type from the standard library instead
- Simplified the low-level syntax mutation methods to just two methods: `detach` return a clone of `self` detached from its parent node, and `replace_slots` returns a new version of `self` with a specified range of its children replaced by the content of an iterator
- Added new high-level mutation methods on `AstNodes` created by the codegen step using the grammar file. This API mirrors the fields accessor API and provide a type-checked method for replacing each fields of a node eg. for `JsArrowFunctionExpression::body(&self) -> JsAnyFunctionBody` the corresponding mutation method is `JsAnyFunctionBody::with_body(&self, new_body: JsAnyFunctionBody) -> Self`
- Added new mid-level mutation functions `replace_{node,token}` and `replace_{node,token}_retain_trivia` implemented on all `AstNodes` to help with replacing nodes and tokens at arbitrary depth within a syntax tree
- Added new high-level mutation methods on all types implementing `AstNodeList` and `AstSeparatedList` through extension traits: for now a single function `fn splice(&self, range: Range, replace_with: Iterator) -> Self` is provided, but it should be easy to derive implementations of `insert`, `remove` or `replace` on top of the splicing operation
- Added new low-level methods for creating detached nodes and tokens (`SyntaxNode::new_detached` and `SyntaxToken::new_detached`), and a high-level factory API under `rome_<language>_factory::make` to create typed AST nodes: these factory functions are code-generated from the grammar file and generate either a builder object or a simple function depending on whether the associated node has optional fields or not
- Moved the implementations of `SyntaxFactory` to the corresponding `rome_<language>_factory` crate
- Refactored the `rome_analyze` to make use of these new APIs to implement code fixes and assists. The analyzer has been refactored to be entirely stateless (until we decide if and how we want to implement incremental analysis), instead each run of the analyzer on a root node will traverse the entire syntax tree once and evaluate a set of rules for each node. Each rule declares "interest" in a certain node type and is notified each time the traversal encounters a node matching the query, it can then decide to emit a "signal" to the consumer of the analyzer that can be a diagnostic, a code fix or both.

## Test Plan

For the low-level changes to Rowan besides the existing test suites for the parser or formatter making heavy use of the API to create and traverse trees, I added an additional check running on exit of the prettier test suite (in addition to the existing logic for printing a diff report) to verify that no green or red node was leaked using the `countme` library. I also changed the exit callback to be scheduled using the `atexit` function from libc instead of using the `dtor` attribute since that didn't seem to work correctly on all platforms.

For the analyzer the `rome_analyze` now has a suite of snapshot tests verifying the correct diagnostics and / or code fixes are being emitted for each implemented rule.

## Unresolved questions

- Should Rowan expose an annotation feature (similar to Roslyn), and how to store these annotations ? It seems like these could be useful to help writing code fixes: annotate a certain node before performing syntax operations to help find it in the mutated version of the tree, mark nodes that should be passed through the formatter before being printed back, ask the editor to move the cursor to a certain token after applying a code action ...
- Rules can be categorized into "lint", "action" and "syntax"', with the last one aimed at checks we'd want to perform as part of a secondary checking step once the parser is done emitting the initial syntax. At the moment no rule is implemented using this category as it is unclear what would happen to the nodes with error (replace with unknown nodes ? attach `SyntaxError` annotation ?)


crates/rome_analyze/src/lib.rs
```diff
@@ -1,14 +1,44 @@
...
-pub use assists::AssistContext;
-pub use categories::ActionCategory;
-pub use signals::{Action, Analysis, DiagnosticExt, Signal, TextAction};
-pub use syntax_edit::{Indel, SyntaxEdit};
+
+use crate::analysis_server::RuleRegistry;
+pub use crate::categories::ActionCategory;
+pub use crate::signals::{AnalyzerCodeFix, AnalyzerDiagnostic, AnalyzerSignal};
+
+/// Allows filtering the list of rules and source code range to emit diagnostics or code fixes for
```
ematipico, line 14: for... ?

crates/rome_analyze/tests/specs/noDelete.js.snap
```diff
@@ -0,0 +1,76 @@
...
+```
+
+# Diagnostics
+```
+  ┌─ noDelete.js:2:1
+  │
+2 │ delete a.b;
+  │ ^^^^^^^^^^ This is an unexpected use of the delete operator.
+
+```
```
ematipico, line 24: Do you think it's possible to actually move together  code fix and diagnostic? I might get difficult to track an error with the relative code fix

crates/rome_analyze/tests/spec_tests.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    let source_type = input_file.try_into().unwrap();
+    let parsed = parse(&input_code, 0, source_type);
+    let root = parsed.tree();
+
+    let rule_name = input_file.file_stem().and_then(OsStr::to_str).unwrap();
+    let filter = AnalysisFilter {
+        rules: Some(slice::from_ref(&rule_name)),
+        ..AnalysisFilter::default()
+    };
```
ematipico, line 30: Can we add a comment or, even better, a md file where we explain how tests of rules work? It seems that the name of the file must match the rule in order to be run with that particular rule.

Although, this is a limitation where we can't have multiple files/cases to apply a single rule. Maybe we can apply the same logic to folders too. If a folder has the name of the rule, all files will be analyzed with that single rule.

ematipico (commented):
First round of my code review. I haven't managed to go through all of it. There are many things that are unclear to me, which is probably a good excuse to add some comment to the functionality of the analyzers.

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
...
-    ///
-    /// [`Analyzer`]: crate::Analyzer
-    pub fn diagnostics(&self, file_id: FileId) -> impl Iterator<Item = AnalyzeDiagnostic> {
-        self.analyze(file_id).into_diagnostics()
+/// Representation of a single rule in the registry as a generic function pointer
+type RegistryRule =
+    for<'a> fn(&'a JsAnyRoot, &'a JsSyntaxNode) -> Option<Box<dyn AnalyzerSignal + 'a>>;
+
+/// Generic implementation of RegistryRule for any rule type R
+fn run<'a, R: Rule + 'static>(
```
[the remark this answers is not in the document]
leops: The type of `R` needs to be static so the vtables needed by instances of `RuleSignal::new_boxed` are also statically allocated (this is a technical requirement to allow `Box<dyn AnalyzerSignal>` to be possible)

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
...
-                }
-            }
-        }
-        signals.into()
+    fn code_fix(
+        _root: &JsAnyRoot,
+        _node: &Self::Query,
+        _result: &Self::Result,
+    ) -> Option<RuleCodeFix> {
+        None
```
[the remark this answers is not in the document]
leops: Yes this allows an "assist" rule to only implement `code_fix` and have `diagnostic` default to None, or potentially for "analyzer" rules to only implement `diagnostic` but not providing an associated `code_fix`

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
...
+pub struct RuleDiagnostic {
+    pub severity: Severity,
+    pub range: TextRange,
+    pub message: MarkupBuf,
+}
+
+/// Code fix object returned by a single analysis rule
+pub struct RuleCodeFix {
+    pub root: JsAnyRoot,
+}
```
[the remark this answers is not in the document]
leops: I tried to keep the functionality of the analyzer as generic as possible, but unfortunately had to keep using JS-specific types in a few places since some multi-language abstractions aren't in place yet. Specifically for the case of `JsAnyRoot` these could be replaced with a `Root` associated type on the `Language` trait (the signature of the `analyze` function would become `fn<L>(&L::Root, ...) where L: Language` for instance)

crates/rome_analyze/src/analyzers/no_delete.rs
```diff
@@ -0,0 +1,155 @@
...
+
+use crate::categories::ActionCategory;
+
+use crate::analysis_server::{Rule, RuleCodeFix, RuleDiagnostic};
+
+pub(crate) enum NoDelete {}
+
+impl Rule for NoDelete {
+    const NAME: &'static str = "noDelete";
+    const ACTION_CATEGORIES: &'static [ActionCategory] = &[];
```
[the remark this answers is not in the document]
leops: It's being used in the `flipBinExp` code assist in `crates\rome_analyze\src\assists\flip_bin_exp.rs`

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
-use std::{collections::HashMap, sync::Arc};
-
-use rome_js_parser::parse_script;
-use rome_js_syntax::{JsLanguage, JsSyntaxNode, TextRange};
-use rome_rowan::AstNode;
-use tracing::trace;
+use rome_console::MarkupBuf;
```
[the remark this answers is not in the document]
leops: Since this file underwent some heavy refactoring I think I could safely rename it to something like `registry.rs`, the "server" concept may come back once we have a plan for incremental / stateful analysis but this isn't really laid out at the moment

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
...
 
-    pub fn find_node_at_range<T: AstNode<Language = JsLanguage>>(
+impl_registry_builders!(
+    // Analyzers
+    NoDelete,
+    NoDoubleEquals,
+    UseWhile,
+    // Assists
+    FlipBinExp
+);
```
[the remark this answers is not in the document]
leops: I agree with the idea, but it will probably have to wait until the `no_vars` and `use_single_case_statement` rules are ported since they are disabled at the moment

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
...
+pub(crate) trait Rule {
+    const NAME: &'static str;
+    const ACTION_CATEGORIES: &'static [ActionCategory];
+
+    type Query: AstNode + 'static;
+    type Result: 'static;
+
+    fn run(node: &Self::Query) -> Option<Self::Result>;
+
+    fn diagnostic(_node: &Self::Query, _result: &Self::Result) -> Option<RuleDiagnostic> {
```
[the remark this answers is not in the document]
leops: The flow of control here is a bit complicated since it switches from a push-based to a pull-based model in the middle: the consumer of the API calls `analyze` on a root node and this will eagerly execute `run` on all the nodes matching the query, but the `run` method returning `Some` will result in a generic "signal" object (internally a type-erased wrapper around the `Result` type of the rule) being raised in a consumer-provided callback function. At this point the consumer may decide to call `diagnostic`, `code_fix` or both methods depending on the use case (for instance the test suite calls both methods, but the Language Server Protocol has two different queries for diagnostics and code actions so one callback will only ask for a diagnostic while the other will only request a code fix)

crates/rome_analyze/src/analyzers/no_delete.rs
```diff
@@ -0,0 +1,155 @@
...
+            .to_owned(),
+        })
+    }
+
+    fn code_fix(
+        root: &JsAnyRoot,
+        node: &Self::Query,
+        result: &Self::Result,
+    ) -> Option<RuleCodeFix> {
+        let root = root.clone().replace_node_retain_trivia(
```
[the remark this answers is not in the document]
leops: I think most if not all code fixes will need an owned version of the root to modify it, so yes it would make sense to move the call to clone to the common `RuleSignal::code_fix` function

crates/rome_analyze/src/analyzers/no_delete.rs
```diff
@@ -0,0 +1,155 @@
...
+                {
+                    return Err(());
+                }
+
+                Ok(Self::JsComputedMemberExpression(expr))
+            }
+            _ => Err(()),
+        }
+    }
+}
```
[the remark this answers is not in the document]
leops: `MemberExpression` is designed as an intermediate type to allow `Js*MemberExpression` to be safely converted into `Js*MemberAssignment`, this implementation of `TryFrom` is what does the actual checking by verifying the provided expression is a valid member expression with no syntax error, so that later calls to `member_expression.into()` can return a member assignment without panicking

crates/rome_analyze/src/analyzers/no_delete.rs
```diff
@@ -0,0 +1,155 @@
...
+    type Result = MemberExpression;
+
+    fn run(node: &Self::Query) -> Option<Self::Result> {
+        let op = node.operator().ok()?;
+        if op != JsUnaryOperator::Delete {
+            return None;
+        }
+
+        let argument = node.argument().ok()?;
+        MemberExpression::try_from(argument).ok()
```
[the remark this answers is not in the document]
leops: It's only being used in the code fix to rewrite `delete a.b` as `a.b = undefined`: in the input `a.b` is a member expression but in the input it's a member assignment, and `MemberExpression` is the wrapper doing the conversion. The validation is done early on inside of `run` by calling `TryFrom` so the diagnostic will never be raised if the right-hand side of the delete expression could not be rewritten as an assignment (for instance `delete a?.b`)

crates/rome_analyze/src/analysis_server.rs
```diff
@@ -1,146 +1,123 @@
...
-        analyzer_actions.chain(assist_actions)
+/// Trait implemented by all analysis rules: declares interest to a certain AstNode type,
+/// and a callback function to be executed on all nodes matching the query to possibly
+/// raise an analysis event
+pub(crate) trait Rule {
+    const NAME: &'static str;
+    const ACTION_CATEGORIES: &'static [ActionCategory];
+
+    type Query: AstNode + 'static;
+    type Result: 'static;
```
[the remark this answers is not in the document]
leops: I think it might make more sense to call it `State`, although that's not entirely exact either but the entire point of this type is to be kept around between a call to `run` and subsequent calls to `diagnostic` or `code_fix`

crates/rome_analyze/src/analyzers/use_while.rs
```diff
@@ -1,20 +1,81 @@
...
+pub(crate) enum UseWhile {}
+
+impl Rule for UseWhile {
+    const NAME: &'static str = "useWhile";
+    const ACTION_CATEGORIES: &'static [crate::ActionCategory] = &[];
+
+    type Query = JsForStatement;
+    type Result = ();
+
+    fn run(n: &Self::Query) -> Option<Self::Result> {
```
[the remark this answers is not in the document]
leops: The `Result` type is sort of optional, it has to be explicitly defined but rules that don't use it can set it to `()`. The point of this type is to allow some data to be transferred between `run` (that raise a signal to the consumer of the analyzer) and `diagnostic` or `code_fix` that emit actual data requested by the consumer, but some rules like this one don't need to transfer anything as they can easily derive all the information needed to generate a diagnostic or a code fix from the `Query` node

crates/rome_analyze/src/assists/flip_bin_exp.rs
```diff
@@ -1,44 +1,81 @@
...
+        let prev_op = new_node.operator_token().ok()?;
+        let new_op = make::token(*op);
+        let new_node = new_node.replace_token_retain_trivia(prev_op, new_op)?;
 
-    let op_range = node.operator_token().ok()?.text_trimmed_range();
-    if !op_range.contains_range(ctx.cursor_range) {
-        return None;
+        let prev_right = new_node.right().ok()?;
+        let new_right = node.left().ok()?;
+        let new_node = new_node.replace_node_retain_trivia(prev_right, new_right)?;
```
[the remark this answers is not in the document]
leops: Yes in general any field that returns a `SyntaxError` will have raised a `Diagnostic` in the parser already so most lint rules will want to silently ignore nodes with missing children

crates/rome_analyze/tests/specs/flipBinExp.js.snap
```diff
@@ -0,0 +1,202 @@
...
+
+```
+
+```
+      | @@ -10,7 +10,7 @@
+ 9  9 |   }
+10 10 |   if (a === b) {
+11 11 |   }
+12    | - if (a != b) {
+   12 | + if (b != a) {
```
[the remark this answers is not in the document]
leops: This is a code action that can be triggered manually by the user, in VSCode for instance it will show up in a floating light bulb menu next to the current line when the cursor is on a binary expression. This rule is tagged with the `Refactor` category so it will never be executed when doing an auto-fix run of the CLI for instance, only when the user explicitly asks for it to make a condition more readable for instance

crates/rome_analyze/src/analyzers/no_delete.rs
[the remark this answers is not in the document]
ematipico: OK, I start to understand how things work, a bit. Here's a follow up question. `MemberExpression` is used as intermediate to do grammar check due to how the grammar works. 

Do you think it makes sense to generate a function like `has_valid_grammar()` that is codegen from the grammar? I mean, in a sense we already have all the information (the syntax factory uses the same information for its slots).  

The reason I am asking this is because I presume we will need to make these checks in mostly all rules (for now, we don't know how a linter on broken grammar will plays out) and sometimes we will need to check the same node more then once.

crates/rome_analyze/tests/specs/noDoubleEquals.js.snap
```diff
@@ -0,0 +1,56 @@
...
+const isNonNull = a != null;
+
+```
+
+# Diagnostics
+```
+  ┌─ noDoubleEquals.js:1:18
+  │
+1 │ const isZero = a == 0;
+  │                  ^^ Do not use the == operator
```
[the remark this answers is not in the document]
leops: I absolutely agree that showing very clear information in the diagnostics is crucial to a high-quality linter, I just haven't implemented it yet so currently only the codespan is being printed in the snapshots, without the common diagnostic header or the inline fix for instance

crates/rome_analyze/tests/specs/noDelete.js.snap
leops: It would absolutely be possible to merge the diagnostic and code fix in a single message (this is what I did in the playground integration for instance), I just thought it would make the logic a bit more complex to accommodate for "assist" rules that only emit a code fix without a diagnostic

crates/rome_js_factory/src/make.rs
```diff
@@ -0,0 +1,33 @@
+use rome_js_syntax::{JsSyntaxKind, JsSyntaxToken, TriviaPieceKind};
```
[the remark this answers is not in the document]
leops: This is by design, a TriviaPiece by itself only stores a kind and a length and its text is actually stored in the (non trimmed) content of its parent token

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -257,17 +257,39 @@ struct DiffReport {
...
+            extern "C" fn print_report() {
+                REPORTER.print();
+            }
+
+            unsafe {
+                atexit(print_report);
+            }
+        });
+
+        &REPORTER
```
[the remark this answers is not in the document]
leops: This is a replacement for the previous version of the diff printing that used the `dtor` attribute, it seems this didn't work correctly on all platforms so I replaced it with a `print_report` callback being explicitly scheduled to run when the process exits (using the `atexit` function from libc), with the scheduling being executed lazily the first time a test case uses the reporter (using an atomic `Once` to only schedule the callback a single time)

crates/rome_analyze/src/analyzers/no_delete.rs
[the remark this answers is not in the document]
leops: It would certainly make sense to have a generic way of checking that a certain node doesn't have any syntax error, and we may want to skip these nodes entirely during the traversal when they are know to have such an error (although some rules may still want to process invalid nodes for instance if we move some of the validation that currently exists in the parser to the analyzer). But on a practical level all the AstNodes have accessor methods typed to return a `SyntaxResult` anyway so I think most of the code would still be using `.ok()?` (or `.unwrap()`) and perform these checks even if the node is known to be valid

crates/rome_analyze/src/lib.rs
leops: That was the actual end of the sentence, but the wording wasn't very clear so I rewrote that comment entirely

ematipico:
> Figure out the high-level mutation API for lists (needs a use case, none of the currently implemented rules ever modifies a syntax list)

Here's a possible use case: https://github.com/rome/tools/blob/archived-js/internal/compiler/lint/rules/js/useSingleVarDeclarator.test.md#0

leops:
!bench_parser

leops:
!bench_formatter

leops:
!bench_formatter

crates/rome_analyze/src/assists/flip_bin_exp.rs
```diff
@@ -1,44 +1,81 @@
...
+    fn run(node: &Self::Query) -> Option<Self::State> {
+        let JsBinaryExpressionFields {
+            left,
+            operator_token: _,
+            right,
+        } = node.as_fields();
+
+        // Ensure the node doesn't have any syntax error
+        left.ok()?;
+        right.ok()?;
```
MichaReiser, line 31: Less related to your work but seems like our intention behind using `Result` to signal syntax errors isn't really protecting this analyzer from making changes when the node has any syntax errors because it requires explicitly verifying that the node is OK. It would even be simpler if the node itself exposed a `node.has_diagnostics()` method.

crates/rome_analyze/src/registry.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl_registry_builders!(
+    // Analyzers
+    NoDelete,
+    NoDoubleEquals,
+    UseSingleVarDeclarator,
+    UseWhile,
+    // Assists
+    FlipBinExp
+);
```
MichaReiser, line 44: How does this macro work if we have multiple languages with rules defined in different crates?

crates/rome_css_factory/src/generated/node_factory.rs
```diff
@@ -0,0 +1,805 @@
...
+
+#![allow(clippy::redundant_closure)]
+#![allow(clippy::too_many_arguments)]
+use rome_css_syntax::{
+    CssSyntaxElement as SyntaxElement, CssSyntaxNode as SyntaxNode, CssSyntaxToken as SyntaxToken,
+    *,
+};
+use rome_rowan::AstNode;
+pub fn css_any_function(css_simple_function: CssSimpleFunction) -> CssAnyFunction {
+    CssAnyFunction::unwrap_cast(SyntaxNode::new_detached(
```
MichaReiser, line 11: What are your thoughts on caching nodes created with the node factory or through the `with_` methods?

MichaReiser (commented):
Thanks, @leops for working on this core piece, and challenging part, of our infrastructure. 

I left a few questions and will need to take some more time to review the PR in detail. 

One common issue with manipulating binary expression is the one of operator precedence. For example, for `4 * 5 + 3` the `*` operator has higher precedence than `+` and is represented as

```yml
Binary:
  - left: Binary
    - left: 4
    - op: *
    - right: 5
  -  op: +
  - right: 3
```

If a refactor swaps the operators of the two binary expressions it produces the following tree


```yml
Binary:
  - left: Binary
    - left: 4
    - op: +
    - right: 5
  -  op: *
  - right: 3
```

and prints to `4 + 5 * 3`. This isn't strictly wrong but a refactor author might expect that the tree structure reflects the precedence (that `4+5` binds higher than `*3`). Reparsing the printed source also produces a different tree:

```yml
Binary:
  - left: 4
  -  op: +
  - right: Binary
    - left: 5
    - op: *
    - right 3
```

That's why tools like babel automatically insert parenthesis to ensure that the printed code reflects the precedence of the tree [playground](https://astexplorer.net/#/gist/fa156f45b5e9053ea1815c04056b40cd/e4fdcd521323c88f3c848edd34117c6c38102958). 

What are your thoughts on how we can help authors avoid these kinds of problems?

crates/rome_rowan/src/cursor.rs
```diff
@@ -86,224 +86,117 @@ mod node;
...
-                    }
-                    GreenElement::Token { ptr } => {
-                        let _ = GreenToken::from_raw(*ptr);
-                    }
-                }
-                break;
-            }
+        match self.green {
+            GreenElement::Node { ptr } => unsafe {
+                GreenNode::from_raw(ptr);
```
MichaReiser, line 137: Would you mind adding some documentation on why this is necessary here? Is it so that the `Drop` of `GreenNode` get's executed?

crates/rome_rowan/src/cursor/node.rs
```diff
@@ -411,33 +381,30 @@ impl SyntaxNode {
...
-        for (i, child) in self.children_with_tokens().enumerate() {
-            if to_delete.contains(&i) {
-                child.detach();
-            }
-        }
-        let mut index = to_delete.start;
-        for child in to_insert {
-            self.attach_child(index, child);
-            index += 1;
+    #[must_use]
```
MichaReiser, line 384: Would you mind adding some documentation here and to `splice_slots` why the result must be used?

crates/rome_analyze/src/assists/flip_bin_exp.rs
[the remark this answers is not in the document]
leops: I also thought reusing trivias should probably be the default, I just didn't have any good ideas for the name of the "plain replace" method at the time so thanks for the suggestion

crates/rome_analyze/src/registry.rs
leops: Eventually the `RuleRegistry` would need to hold a (phantom) type parameter to the language it's analyzing so I think this macro would then change to generate a plain function `fn make_registry(filter) -> RuleRegistry<$language>` instead of an associated function, and either the root `analyze` function would take that as a parameter or be implemented as a method on it (with the current `analyze` method renamed to `analyze_node`)

MichaReiser: That makes sense. I've been burned a lot in the past so I want to bring this question up. How does this work with Rust's orphan rule? The `RuleRegistry` struct is part of the `rome_analyze` crate but the macro call to `impl_registry_builders` would be part of the `rome_language_analze` create right. My understanding is that Rust then permits you from implementing the `RuleRegistry` because it's a type from another crate and `JsLanguage` is so too (except if you plan to implement the analyzer in `rome_js_syntax`).

crates/rome_css_factory/src/generated/node_factory.rs
leops: We discussed some ideas for caching strategies in the last Linter WG meeting, and I plan on testing some of these eventually not only to evaluate how caching would improve (or degrade) performance, but also what would be the cognitive overhead of having to pass a handle to the cache everywhere vs. the performance overhead of having it be a shared global state (like the memory allocator)

crates/rome_analyze/src/registry.rs
leops: This is the reason why I want to turn `make_registry` into a plain function, as the orphan rule would disallow it to be implemented as a method in a different crate from `rome_analyze`

MichaReiser: Oh, I missed that part. Thanks for pointing that out again.

crates/rome_rowan/src/cursor.rs
leops: This is a part of the code I had implemented incorrectly, so it resulted in green nodes being leaked during mutations, but in the latest version I pushed this is now fixed (along with adding comments explaining what the unsafe code is doing and why)

crates/rome_analyze/src/assists/flip_bin_exp.rs
leops: Having nodes tagged when they have any missing children would be one of the intended use for syntax annotations, but in the meantime manually checking for errors isn't too bad: in this case for instance not doing it wouldn't change anything, in case of a missing child the rule would emit a signal but no diagnostic or code action and nothing would happen, this is just an optimization to avoid emitting a useless signal in the first place. It's a little more important however to avoid emitting a signal for rules have both diagnostics and code actions, as we generally don't want to emit the diagnostic if we wouldn't be able to emit a code fix for it.

MichaReiser: I guess my main concern is that it results in a lot of boilerplate code to test every individual field as you can see in the other rules you implemented. It's also something devs need to explicitly do and may not be immediately clear what needs to be done if a new field has been added. 

I think it would be helpful if we can built this into our infrastructure in some way or another so that a refactor can decide what should happen if a changed node has or contains a syntax error. I'm not saying that this must be solved in this PR but I think it should be something that we follow up on to think about how lints/refactors should work with nodes that have syntax errors and how to guide devs so that there rules do the right thing by default.

crates/rome_analyze/src/assists/flip_bin_exp.rs
```diff
@@ -1,44 +1,79 @@
...
+        JsBinaryOperator::LessThan => Some(T![>]),
+        JsBinaryOperator::GreaterThan => Some(T![<]),
+        JsBinaryOperator::LessThanOrEqual => Some(T![>=]),
+        JsBinaryOperator::GreaterThanOrEqual => Some(T![<=]),
+        JsBinaryOperator::Equality => Some(T![==]),
+        JsBinaryOperator::StrictEquality => Some(T![===]),
+        JsBinaryOperator::Inequality => Some(T![!=]),
+        JsBinaryOperator::StrictInequality => Some(T![!==]),
+        JsBinaryOperator::Plus => Some(T![+]),
+        JsBinaryOperator::Minus => None,
```
MichaReiser, line 67: Nit: Shouldn't these operators be symmetric?

```
+ => -
- => +
* => /
/ => *

File: crates/rome_analyze/src/categories.rs [hunk 27]
```diff
@@ -1,6 +1,45 @@
...
+}
+
+bitflags! {
+    pub struct RuleCategories: u8 {
+        const SYNTAX = 1 << RuleCategory::Syntax as u8;
+        const LINT = 1 << RuleCategory::Lint as u8;
+        const ACTION = 1 << RuleCategory::Action as u8;
+    }
+
+    pub struct ActionCategory: u8 {
```
MichaReiser, line 24: Nit: Should this be plural too considering that the bitflag for rule categories uses plural

crates/rome_analyze/src/categories.rs
```diff
@@ -1,6 +1,45 @@
...
+        const SAFE_FIX = 1 << 0;
+        const SUGGESTION = 1 << 1;
+        const REFACTOR = 1 << 2;
+    }
+}
+
+impl Default for RuleCategories {
```
MichaReiser, line 31: Does it make sense to default to all? Doesn't make that run rules at too many stages?

crates/rome_analyze/src/registry.rs
```diff
@@ -0,0 +1,134 @@
...
+
+                $( if filter.categories.contains($rule::CATEGORY.into()) && filter.rules.map_or(true, |rules| rules.contains(&$rule::NAME)) {
+                    rules.push(run::<$rule>);
+                } )*
+
+                Self { rules }
+            }
+        }
+    };
+}
```
MichaReiser, line 34: Checking whatever a rule matches the filter seems to be cheap. What are your thoughts on having a single static registry where the filtered registry is an adaptor that filters the rules on the fly?

crates/rome_analyze/src/registry.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// rules when running the analyzer
+    const CATEGORY: RuleCategory;
+
+    /// The type of AstNode this rule is interested in
+    type Query: AstNode + 'static;
+    /// A generic type that will be kept in memory between a call to `run` and
+    /// subsequent executions of `diagnostic` or `action`, allows the rule to
+    /// hold some temporary state between the moment a signal is raised and
+    /// when a diagnostic or action needs to be built
+    type State: 'static;
```
MichaReiser, line 96: What' does the `'static` lifetime restrict for the `Query` and `State` type? It's unclear to me what the lifetime means in this context.

crates/rome_analyze/src/registry.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// The default implementation returns None
+    fn diagnostic(_node: &Self::Query, _state: &Self::State) -> Option<RuleDiagnostic> {
+        None
+    }
+
+    /// Called by the consumer of the analyzer to try to generate a code action
+    /// from a signal raised by `run`
+    ///
+    /// The default implementation returns None
+    fn action(_root: JsAnyRoot, _node: &Self::Query, _state: &Self::State) -> Option<RuleAction> {
```
MichaReiser, line 116: What's the reason that the action passes in the root rather than the node's parent?

How to you plan to make this language agnostic?

crates/rome_css_factory/src/generated/node_factory.rs
```diff
@@ -0,0 +1,805 @@
...
+                Some(SyntaxElement::Token(self.and_token)),
+                self.condition_token
+                    .map(|token| SyntaxElement::Token(token)),
+                Some(SyntaxElement::Node(self.ty.into_syntax())),
+            ],
+        ))
+    }
+}
+pub fn css_at_media_query_feature(
+    l_paren_token: SyntaxToken,
```
MichaReiser, line 144: What are your thoughts on default initializing all "trivia"/punctuation tokens where it's known what string value they contain like `(`, `if`?

crates/rome_rowan/src/cursor.rs
```diff
@@ -37,274 +32,170 @@
 //
 // `NodeData` which doesn't have a parent (is a root) owns the corresponding
 // green node or token, and is responsible for freeing it.
-//
```
MichaReiser, line 40: Would it be worth replacing this documentation with something that explains the new representation?

crates/rome_rowan/src/cursor.rs
```diff
@@ -37,274 +32,170 @@
...
+/// holds a strong reference to the root of the green tree) or a "child node"
+/// (holds a strong reference to its parent red node and a weak reference to its
+/// counterpart green node)
+#[derive(Debug)]
+enum NodeKind {
+    Root {
+        green: GreenElement,
+    },
+    Child {
+        green: WeakGreenElement,
```
MichaReiser, line 88: What's the reason that the node holds a weak reference to its green node. My understanding is that the node itself isn't of much use anymore if the underlying green node gets deallocated (all operations point into the void).

crates/rome_rowan/src/cursor.rs
```diff
@@ -37,274 +32,170 @@
...
+        parent: Rc<NodeData>,
+    },
+}
+
+/// Child SyntaxNodes use "unsafe" weak pointers to refer to their green node,
+/// unlike the safe [std::sync::Weak] these are just a raw pointer with no
```
MichaReiser, line 94: Nit
```suggestion
/// Child SyntaxNodes use "unsafe" weak pointers to refer to their green node.
/// Unlike the safe [std::sync::Weak] these are just a raw pointer with no
```

crates/rome_rowan/src/cursor.rs
```diff
@@ -37,274 +32,170 @@
...
+/// additional semantics meaning the corresponding [ThinArc](crate::arc::ThinArc)
+/// doesn't have to keep a counter of outstanding weak references or defer the
```
MichaReiser, line 96: There's something off with this sentence. I think it's something along *...with no additional semantics meaning. The corresponding ThinArc....*

crates/rome_rowan/src/cursor.rs
```diff
@@ -37,274 +32,170 @@
...
+/// release of the underlying memory until the last `Weak` is dropped. On the
```
MichaReiser, line 97: Why is it that it doesn't have to keep the counter?

crates/rome_rowan/src/syntax/node.rs
```diff
@@ -22,6 +22,26 @@ impl<L: Language> SyntaxNode<L> {
         SyntaxNode::from(cursor::SyntaxNode::new_root(green))
     }
 
+    pub fn new_detached<I>(kind: L::Kind, slots: I) -> SyntaxNode<L>
```
MichaReiser, line 25: Nit: Some documentation of when this is useful?

crates/rome_rowan/src/syntax/token.rs
```diff
@@ -12,6 +16,34 @@ pub struct SyntaxToken<L: Language> {
 }
 
 impl<L: Language> SyntaxToken<L> {
+    /// Create a new detached token
```
MichaReiser, line 19: What's the meaning of detached?

crates/rome_analyze/src/assists/flip_bin_exp.rs
leops: The `flipBinExp` refactor only swaps the right and left hand side of a binary expression but tries to leave the actual operation intact. For logical operations this means replacing `a == b` with `b == a` and `a < b` with `b > a`, and for arithmetic operations it's possible to replace `a + b` with `b + a` however `a - b` doesn't have the same meaning as `b - a` which is why `invert_op` returns None for that case

crates/rome_analyze/src/categories.rs
leops: The `RuleCategories` type is used in the rule filter when running the analyzer, so having it default to `all` means it doesn't filter out any rule by default

crates/rome_analyze/src/registry.rs
leops: Filtering out rules isn't that cheap at the moment because the list of allowed rule names is an array of strings but it could rather easily be changed to a more efficient data structure, then performing filtering on the fly would be a lot more viable and that seems necessary anyway to support suppressing a certain rule by name for a section of a file

crates/rome_analyze/src/registry.rs
leops: This lifetime constraint was needed because both the `Query` and `State` types later get stores in a type erased `Box<dyn AnalyzerSignal>` that didn't have any lifetime associated to it (so it potentially needed to outlive the static lifetime) but this is no longer the case so these can be removed

crates/rome_analyze/src/registry.rs
leops: This if for flexibility, the code action is provided with the entire document and can return a new version of it modified however it wants (for instance this would allow the code fix for "await expression in non-async context" to be "add the async keyword to the outer function"). In order for this to be language agnostic the `Language` trait would need to have a `Root` associated type (bounded with `AstNode<Language = Self>` if Rust accepts the recursion)

crates/rome_css_factory/src/generated/node_factory.rs
leops: That seems like a good idea, I guess I would just need to turn the factories for these nodes into builders where the token being None means it should be Default-initialized

crates/rome_rowan/src/cursor.rs
leops: This is an optimization that was already present in Rowan but I've turned it into an invariant enforced by the type system. Basically a child `SyntaxNode` (red node) holds a strong reference to its parent, that parent (or nth ancestor up to the root of the tree) holds a strong reference to its corresponding green node, and that green node is holding strong references to its children going back down the tree to the corresponding green node for our original red node. So this means if a child red node exists, there is a guaranteed cycle of strong references to its green counterpart anyway so the SyntaxNode itself doesn't need to bother with managing a reference counted handle and can use a raw pointer instead since its statically guaranteed to never outlive it

crates/rome_rowan/src/cursor.rs
leops: I think I actually meant "*... with no additional semantics, meaning the corresponding ThinArc ...*", I'll see if I can rewrite that to make it clearer

crates/rome_rowan/src/cursor.rs
MichaReiser: Got it. I think it would be helpful to paste some of what you explained here into the documentation explaining why a weak reference is still safe.

MichaReiser (approved):
This is a big step forward. I haven't reviewed every line and there are a few things that we may want to iterate on but don't see reason why this should be blocking the PR from merging. 

It may be helpful for the future to maybe split the PR in three parts:

* Refactor of rowan's mutation API
* Code gen
* Linting

But I know that this is a pain with GitHub's lack of stacked PRs.

PR #2493, opened 2022-04-25
feature(cli): Filter files using .gitignore

### LINKED ISSUE #2492: Add .gitignore support
Currently we have hard coded folders that we ignore. We should change that to follow the .gitignore

NicholasLYang:
## Summary
Closes #2492 
I'm trying out using the `ignore` crate to filter files. 

Some notes/questions:
- I'm keeping `DEFAULT_IGNORE` for now because we still want to ignore `.git`. Are there any other folders that we'd want to ignore by default?
- Right now I have the filtering logic in `rome_fs`. That might be too general, in which case we should make it specific to `rome_cli`. When I tried that implementation, it was a little hard to make `TraversalContext` store gitignore patterns without adding mutability across threads. But I can try again/get advice on how to do that.
- One option is to try replacing our entire walking logic with [ignore](https://crates.io/crates/ignore) or [jwalk](https://crates.io/crates/jwalk). 

## Test Plan

We have a basic example in `examples` with `test.js` being .gitignored. We should probably come up with more tests for the CLI in general.


ematipico:
The feature seems to be different than I expected, compared to how we used it in Rome classic.

I think we should revisit it later one, when Rome will have support for a configuration file. The reasons why I think we should wait are the following:

- we don't know if the folder of the project is using `git` (or a VCS system). This info should come [from the configuration](https://github.com/rome/tools/blob/main/website/src/_includes/docs/_old_project-config.md#root), and from there we can check if we are inside a VCS (git in this case) and act;
- using `.gitignore` as default could lead to issues where we can't process files because they are not tracked. Meaning sometimes files are not tracked on purpose, but they can be analyzed. For example lock files. Or built files. If the user want to tell Rome to ignore file, they should use Rome's configuration file;
- in the old repository we decided, on purpose, to [ignore test files](https://github.com/rome/tools/blob/archived-js/.gitignore#L9-L10) BUT we wanted to be able to process them so we could test the CLI. It was **really helpful**, and I think we should keep it that way;

leops:
I think eventually we could have our own format of `.romeignore` file with a syntax closer to `.gitattributes` to allow some features to be selectively disabled on files matching a pattern (for instance disable formatting in `node_modules` but still include these files in semantic analysis)

NicholasLYang:
I agree that a `romeignore` feature would be nice. However, I do think we should have a feature to use `gitignore` for now. It encapsulates a very common use case and there's a simple escape hatch. Like the original solution of having hard coded paths to ignore, it doesn't solve the entire problem, but it gets us most of the way there.

> we don't know if the folder of the project is using git (or a VCS system). This info should come [from the configuration](https://github.com/rome/tools/blob/main/website/src/_includes/docs/_old_project-config.md#root), and from there we can check if we are inside a VCS (git in this case) and act;

Is there any harm in obeying a .gitignore if we're not in a git repository? That seems like a very rare situation.

> using .gitignore as default could lead to issues where we can't process files because they are not tracked. Meaning sometimes files are not tracked on purpose, but they can be analyzed. For example lock files. Or built files. If the user want to tell Rome to ignore file, they should use Rome's configuration file;

I understand that this can happen, but it does seem like an exceedingly rare situation, especially for the formatter. Also I'm not a huge fan of tools forcing the user to do work that the tool can do itself. Instead I like tools such as [fd](https://github.com/sharkdp/fd) and [ripgrep](https://github.com/BurntSushi/ripgrep) who include sane defaults like ignoring `gitignore` by default. Having to use `romeignore` would require most users to duplicate their `gitignore` config, which can cause confusion when a file is added to the `gitignore` but not to `romeignore`. 

Also, if users accidentally run our formatter on a large directory that they purposefully ignored, they'll perceive our formatter as significantly slower than it actually is. We want to encourage people to stay on the fast path as much as possible.

ematipico:
> Is there any harm in obeying a .gitignore if we're not in a git repository? That seems like a very rare situation.

Two cases:
1. we run Rome where there's the global `.gitignore`;
2. by leaning towards using `.gitignore`, we are basically saying to all users that **don't** use git, that we don't care about them (FYI Facebook uses mercurial);

We can't assume that git is the standard, because there are places where it's not, unfortunately.

>  understand that this can happen, but it does seem like an exceedingly rare situation

It happens more often than you might think (it happened many times to me, and having a formatter there was sane). Automation scripts, scripts for refactoring, codemods, etc. A user might want to get these files formatted but not included in the repository.


> Also, if users accidentally run our formatter on a large directory that they purposefully ignored, they'll perceive our formatter as significantly slower than it actually is. We want to encourage people to stay on the fast path as much as possible.

Isn't it the same for each tool that doesn't support `.gitignore`?

leops:
To add to the discussion I do think it's valuable to have the CLI default to skipping VCS ignored files (I think we could easily support more than git there) when reformatting a directory, since as these files aren't checked in any modification to them may result in data loss for the user, if the user doesn't expect the files to be modified they may not even realize what happened since they wont show up in any changeset. But it could still be possible to bypass that default behavior by explicitly pointing the CLI to a single file or passing in a flag to disable ignore files

NicholasLYang:
> we run Rome where there's the global .gitignore;
In this situation, we're running a formatter on a folder that contains a global gitignore, but that folder itself is not version controlled by git, and there are javascript files that need to be formatted? This doesn't seem like a very common situation, unless I'm misunderstanding.

> by leaning towards using .gitignore, we are basically saying to all users that don't use git, that we don't care about them (FYI Facebook uses mercurial);

Isn't this the cost of building software iteratively, no? One could argue that by building a tool that doesn't parse Flow, we're saying to all users who don't use TypeScript that we don't care about them. Facebook does use Flow after all. I agree that in the future we should accommodate more than just git, but the question is whether accommodating git for now is detrimental.

> Isn't it the same for each tool that doesn't support .gitignore?

That is why a lot of modern tools support gitignores. It's extremely tedious to, say, run `find` and filter out files that are gitignored. Hence, `fd` supports filtering via gitignore, and indeed has it activated by default.

NicholasLYang:
> It happens more often than you might think (it happened many times to me, and having a formatter there was sane). Automation scripts, scripts for refactoring, codemods, etc. A user might want to get these files formatted but not included in the repository.

In that case, we have `rome format --use-gitignore false`

ematipico:
> > It happens more often than you might think (it happened many times to me, and having a formatter there was sane). Automation scripts, scripts for refactoring, codemods, etc. A user might want to get these files formatted but not included in the repository.
> 
> In that case, we have `rome format --use-gitignore false`
 
Not sure this will play well. The file won't be formatted in the IDE, which is our top priority.

NicholasLYang:
We can have `--use-gitignore` turned off by default if you'd prefer. Or have it turned off for the editor

ematipico:
> > by leaning towards using .gitignore, we are basically saying to all users that don't use git, that we don't care about them (FYI Facebook uses mercurial);
> 
> Isn't this the cost of building software iteratively, no? One could argue that by building a tool that doesn't parse Flow, we're saying to all users who don't use TypeScript that we don't care about them. Facebook does use Flow after all. I agree that in the future we should accommodate more than just git, but the question is whether accommodating git for now is detrimental.

But we haven't talked about it, when did we decide it?

NicholasLYang:
Ah yeah my bad. I opened an issue with the assumption this wouldn't be a controversial feature. We can discuss this further, although I do believe that the feature is justified according to my comments above. If that isn't the case, we can open a discussion and get other's thoughts.

yassere:
Since this doesn't have consensus, let's close this PR for now. We can revisit it later.

PR #2494, opened 2022-04-25
chore(cli): Updating crate version to match Rome version

NicholasLYang:
## Summary

Part of #2362

As long as there's no objections, I think we should version the `rome_cli` crate according to our overall Rome version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1266,7 +1266,7 @@ dependencies = [
 
 [[package]]
 name = "rome_cli"
-version = "0.0.0"
+version = "0.5.0"
 dependencies = [
  "crossbeam",
  "hdrhistogram",
--- a/crates/rome_cli/Cargo.toml
+++ b/crates/rome_cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rome_cli"
-version = "0.0.0"
+version = "0.5.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html


ematipico:
I am a bit conflicted here. While this is a good change, the reason why the crates have `0.0.0` as version is because they are not published on `crates.io`. So I am not sure if it makes sense or not 🤔

NicholasLYang:
I'd be in favor of us releasing the crates with a big disclaimer that they are not meant for public API usage. The main reason I'd like a version is to make sure that when people are benchmarking, they're benchmarking the right version of Rome

ematipico:
@yassere what do you think? You were the one that introduced this convention.

yassere:
I'm also a bit conflicted. And ideally, I wouldn't want this to be kept manually in sync.

@leops You worked on the release automation and workflows. Is there a better way to include our current version so that it shows up when people use the released cli?

leops:
Well for the time being we could decide to tie the displayed version number of the CLI to the version of the npm package, since I changed the way that's being built in #2495 it would be possible to also store the value of `env.version` in an environment variable as part of the release workflow and use that instead of `CARGO_PKG_VERSION`

yassere:
@leops That sounds like a good idea to me.

@NicholasLYang What do you think? If he adds that to his PR, would you close this one in favor of that approach? And then later we can address the broader question of whether or not to publish anything to crates.io

NicholasLYang:
Sure that's fine. As long as there's a way to verify the version in the CLI

NicholasLYang:
@leops, is this completed? If so, I can close this PR. Right now when I run `npx rome@next` I get:

```
Rome v
Available commands:
- format
- help

```

leops:
Yes, it's done but I messed up the value of the `ROME_VERSION` variable in the release workflow so its set to an empty string in the latest build

Issue #2405, opened 2022-04-13
Only add quotes around object properties where needed


MichaReiser:
## Description

It isn't necessary to put object properties in quotes (string literal name) if they form a valid identifier. 

### Input

```ts
type X = {
  "a": string;
  b: number;
};

interface X {
  "member": string;
}

class Y {
  "other" = 4;
  "method"() {}
  get "getter"() {}
  set "setterr"(value) {}
  "constructor"() {}
}

let value = { "a": "test", "quotes-required": "test" };

({ "a": test } = value);

let { "a": test } = value;
```

### Prettier

```ts
type X = {
	a: string;
	b: number;
};

interface X {
	member: string;
}

class Y {
	"other" = 4;
	method() {}
	get getter() {}
	set setterr(value) {}
	constructor() {}
}

let value = { a: "test", "quotes-required": "test" };

({ a: test } = value);

let { a: test } = value;
```

### Rome

```ts
type X = { "a": string, b: number };

interface X { "member": string }

class Y {
	"other" = 4;
	"method"() {}
	get "getter"() {}
	set "setterr"(value) {}
	"constructor"() {}
}

let value = { "a": "test", "quotes-required": "test" };

({ "a": test } = value);

let { "a": test } = value;
```

## Expected

Rome formatter to match Prettier's output

## Note

Don't remove quotes from numeric properties in TypeScript because that can result in type checker errors (prettier doesn't remove the quotes as well).

[...]

Issue #2482, opened 2022-04-22
📎 Implement option `preserve_quotes`


ematipico:
### Description

See discussion for context: https://github.com/rome/tools/discussions/2465#discussioncomment-2592318

The team agrees that we would give the possibility to preserve the quotes inside the properties an object, if requested.

Given the following input:

```js
const a = { "a": 2 }
```

The default behaviour will format the snippet like this:

```js
const a = { a: 2 }
```

While using the option `--preserve-quotes` will format the snippet like this:

```js
const a = { "a": 2 }
```

Please refer to Prettier's documentation about the option: https://prettier.io/docs/en/options.html#quote-props

ematipico:
This is where we should start having a look: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/js/objects/property_object_member.rs

Then, the actual literal is here: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/js/objects/literal_member_name.rs

We should mind the fact taht `JsLiteralMemberName` is also used inside TypeScript, which means that we have read what Prettier does in case of TypeScript (it says that we should change things because `"a"` and `a` have two different meanings) and then apply the option accordingly.

MichaReiser:
This work depends on #2405

bhbs:
I would like to work on this!

MichaReiser:
> I would like to work on this!

Awesome. I assigned you the issue. Let me know if you need any help.

PR #2502, opened 2022-04-27
fix(rome_js_formatter): elide quotes inside members

ematipico:
## Summary

Partially implements https://github.com/rome/tools/issues/2405

The feature has been turned off for `Type` members, they will be turn on in a separate PR where I will tackle the number literals (which affect JS too).

I also update `insta` to the latest.

## Test Plan

Added new tests to cover cases where it's safe to elide quotes VS cases where it's not safe.


leops (commented):
It was mentioned in the options discussion that we would eventually add a "preserve quotes" option for the formatter, is this something we'd want to introduce in this PR or add back in later (since preserving quotes used to be the default behavior of the formatter before this change) ?

ematipico:
> It was mentioned in the options discussion that we would eventually add a "preserve quotes" option for the formatter, is this something we'd want to introduce in this PR or add back in later (since preserving quotes used to be the default behavior of the formatter before this change) ?

The plan is to add the option after this PR lands: https://github.com/rome/tools/issues/2482

Essentially eliding quotes will be the **default behaviour** (which mimic "as-needed" from Prettier) and we will have a boolean option to mimic the "preserve" from Prettier: https://github.com/rome/tools/discussions/2465#discussioncomment-2614130

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -635,3 +635,88 @@ pub(crate) fn format_string_literal_token(
...
+                !text[1..text.len() - 1]
+                    .chars()
+                    .any(|c| !c.is_alphanumeric())
+            }
+        {
+            let new_text = &text[1..text.len() - 1];
+            Ok(formatter.format_replaced(
+                &name,
+                Token::from_syntax_token_cow_slice(
+                    Cow::Owned(new_text.to_string()),
```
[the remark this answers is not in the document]
ematipico: Internally there's a check, making sure that the string was preserved. I have to allocate a new string, unfortunately:

```
---- specs::prettier::js::object_prop_break_in::long_value stdout ----
thread 'specs::prettier::js::object_prop_break_in::long_value' panicked at 'assertion failed: `(left == right)`
  left: `"ABC"`,
 right: `"\"AB"`: The borrowed string doesn't match the specified token substring. Does the borrowed string belong to this token and range?', /Users/emanuele/www/rometools/rome/crates/rome_formatter/src/format_element.rs:1190:17
```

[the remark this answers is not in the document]
ematipico: Maybe we can cheat it by passing a new range...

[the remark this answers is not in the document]
leops: Looking at the code for `Token::from_syntax_token_cow_slice` I think offsetting the start of the token by 1 should work, it will only change the way sourcemap markers are emitted, and even then the marker probably just be more accurate since the output identifier will start at the correct position inside the original string literal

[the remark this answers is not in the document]
ematipico: That worked :)

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -635,3 +636,98 @@ pub(crate) fn format_string_literal_token(
...
+
+pub(crate) enum PropertyNameCheckMode {
+    Alphabetic,
+    Alphanumeric,
+}
+
+impl PropertyNameCheckMode {
+    fn text_can_be_replaced(&self, text_to_check: &str) -> bool {
+        match self {
+            PropertyNameCheckMode::Alphabetic => !text_to_check.chars().any(|c| !c.is_alphabetic()),
```
IWANABETHATGUY, line 674: `text_to_check.chars().all(char::is_alphabetic)` would be more  efficient and point free

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -635,3 +636,98 @@ pub(crate) fn format_string_literal_token(
...
+            PropertyNameCheckMode::Alphanumeric => {
+                !text_to_check.chars().any(|c| !c.is_alphanumeric())
```
IWANABETHATGUY, line 676: same, `text_to_check.chars().all(char::is_alphanumeric)` would be more efficient and point free

crates/rome_js_formatter/tests/specs/prettier/typescript/compiler/ClassDeclaration22.ts.snap
```diff
@@ -17,7 +17,7 @@ class C {
 ```js
 class C {
   "foo"();
-  "bar"() {}
```
IWANABETHATGUY, line 20: https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#YwBsAGEAcwBzACAAQwAgAHsACgAgACAAIgBmAG8AbwAiACgAKQA7AAoAIAAgACIAYgBhAHIAIgAoACkAIAB7AH0ACgB9AA==
It seems that prettier don't remove the quotes

IWANABETHATGUY (changes requested):

IWANABETHATGUY:
Also `clippy` need to pass

crates/rome_js_formatter/tests/specs/prettier/typescript/compiler/ClassDeclaration22.ts.snap
ematipico: You have to enable typescript. That is not a valid JS syntax (there's a syntax error)

crates/rome_js_formatter/src/utils/mod.rs
leops: In fact it's possible to return the function pointers to `char::is_alphabetic` or `char::is_alphanumeric` from the match expression and use that as the argument to `.all()`

ematipico:
> We should handle the case where the key is an empty string: [Playground](https://play.rome.tools/dd62e8f/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#bABlAHQAIABhACAAPQAgAHsAIAAiACIAOgAgADEAMAAsACAAIgBjAF8AZAAiADoAIAAzADAAIAB9AA==)

Nice catch Nicholas!!! Will do that

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,109 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+}
+
+/// Function used by the formatter, where we pass a complaint member and it returns a [FormatElement[
+/// where the text has its quotes removed.
+pub(crate) fn format_property_name<Member: Into<PropertyName>>(
+    member_name: Member,
+    formatter: &Formatter,
+    checker: PropertyNameCheckMode,
+) -> FormatResult<FormatElement> {
+    fn replace_node(
```
NicholasLYang, line 710: Is there a reason why this is nested inside another function?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,109 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+const QUOTES_TO_OMIT: [char; 2] = ['\"', '\''];
+
+pub(crate) enum PropertyNameCheckMode {
+    Alphabetic,
+    Alphanumeric,
+}
+
+impl PropertyNameCheckMode {
+    /// We can change the text only if there alphanumeric or alphabetic characters, depending on the mode
+    fn text_can_be_replaced(&self, text_to_check: &str) -> bool {
```
NicholasLYang, line 695: Should we not allow underscore as well? Prettier appears to allow underscore when unquoting: [Playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#bABlAHQAIABmAG8AbwAgAD0AIAB7ACAAIgBiAGEAcgBfAGIAYQB6ACIAOgAgADIAMAAgAH0A)

NicholasLYang (dismissed):

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: Isn't prettier unquoting all strings that form valid identifiers according to the spec? 

I'm wondering if we should, therefore, 

* rename `PropertyNameCheckMode` to `MemberContext` which is either `Type` or `Member`
* Or extend the `PropertyName` enum to have additional viarants for the `Type` cases. 

Nit: Rename method to `is_valid_identifier_name`?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,109 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+) -> FormatResult<FormatElement> {
+    fn replace_node(
+        name: JsSyntaxToken,
+        formatter: &Formatter,
+        checker: PropertyNameCheckMode,
+    ) -> FormatResult<FormatElement> {
+        let text = name.text_trimmed();
+
+        // there are cases where we might have an empty string, which means that if the text length
+        // is higher than 2, it means that the text has something in it
```
MichaReiser, line 718:
```suggestion
        // is higher than 2, it means that the identifier name isn't empty
```

It might be worth to move this test into the `text_can_be_replaced`. It then becomes a simple `ident.is_empty()`.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,109 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+        if text.starts_with(QUOTES_TO_OMIT) && text.ends_with(QUOTES_TO_OMIT) && text.len() > 2 {
+            let quote_less_text = &text[1..text.len() - 1];
+            if checker.text_can_be_replaced(quote_less_text) {
+                Ok(formatter.format_replaced(
+                    &name,
+                    Token::from_syntax_token_cow_slice(
+                        Cow::Borrowed(quote_less_text),
+                        &name,
+                        // slide the offset of one as we removed a character from the beginning
+                        name.text_trimmed_range().start().add(TextSize::from(1)),
```
MichaReiser, line 728: You're adding ` to the start of the range to skip over the start quote. Isn't it necessary to subtract 1 from the end as well?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,109 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+                replace_node(name.value()?, formatter, checker)?
+            }
+        },
+        PropertyName::Class(class) => match class {
+            JsAnyClassMemberName::JsComputedMemberName(node) => node.format(formatter)?,
+            JsAnyClassMemberName::JsLiteralMemberName(node) => {
+                replace_node(node.value()?, formatter, checker)?
+            }
+            JsAnyClassMemberName::JsPrivateClassMemberName(node) => node.format(formatter)?,
+        },
```
MichaReiser, line 753: Would it be simpler to restructure the `PropertyName` enum to have the following variants:

```
ComputedMemberName
LiteralMemberName
```

crates/rome_js_formatter/src/utils/mod.rs
ematipico: No particular reason, would you prefer it outside? I didn't want to make visible because it's exclusive  to this logic, but used multiple times.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Probably yes

ematipico: It seems not: 

```
---- specs::prettier::js::object_prop_break_in::long_value stdout ----
thread 'specs::prettier::js::object_prop_break_in::long_value' panicked at 'assertion failed: `(left == right)`
  left: `"ABC"`,
 right: `"\"AB"`: The borrowed string doesn't match the specified token substring. Does the borrowed string belong to this token and range?', /Users/emanuele/www/rometools/rome/crates/rome_formatter/src/format_element.rs:1190:17

````

File: crates/rome_js_formatter/src/utils/mod.rs (hunk 6 above)
Comment (ematipico) [reply crates/rome_js_formatter/src/utils/mod.rs:695]: > Isn't prettier unquoting all strings that form valid identifiers according to the spec?

I don't know

> I'm wondering if we should, therefore,
> 
>     * rename `PropertyNameCheckMode` to `MemberContext` which is either `Type` or `Member`
> 
>     * Or extend the `PropertyName` enum to have additional viarants for the `Type` cases.
> 

`MemberContext` is a better suggestion, thank you!

> Nit: Rename method to `is_valid_identifier_name`?

I don't know, maybe? But this logic is applied to properties only, I would like to keep it that way for now.

Comment (ematipico) [reply crates/rome_js_formatter/src/utils/mod.rs:695]: > Should we not allow underscore as well?

True, I had that in my initial implementation then forgot to put it. I tried to implement a KISS solution, it should cover all the cases.

Comment (MichaReiser) [reply crates/rome_js_formatter/src/utils/mod.rs:695]: > > Isn't prettier unquoting all strings that form valid identifiers according to the spec?

> I don't know

Can you please verify with the prettier implementation.

File: crates/rome_js_formatter/src/utils/mod.rs (hunk 8 above)
Comment (MichaReiser) [reply crates/rome_js_formatter/src/utils/mod.rs:736]: Ah, this is only the start offset and not the full range. Yes, then this is correct.

File: crates/rome_js_formatter/src/utils/mod.rs [hunk 10]
```diff
@@ -649,3 +650,117 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+
+/// Data structure used to merge into one the following nodes:
+///
+/// - [JsAnyObjectMemberName]
+/// - [JsAnyClassMemberName]
+/// - [JsLiteralMemberName]
+///
+/// Once merged, the enum is used to get specific members (the literal ones) and elide
+/// the quotes from them, when the algorithm sees fit
+pub(crate) enum PropertyName {
```
MichaReiser, line 662: Nit: `MemberName`?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,117 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+            MemberContext::Member => text_to_check
+                .split('_')
+                .all(|sub_text| sub_text.chars().all(char::is_alphanumeric)),
+        }
+    }
+}
+
+/// Function used by the formatter, where we pass a complaint member and it returns a [FormatElement[
+/// where the text has its quotes removed.
+pub(crate) fn format_property_name<Member: Into<PropertyName>>(
```
MichaReiser, line 709: Nit: `format_member_name`

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,117 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+    fn replace_node(
+        name: JsSyntaxToken,
+        formatter: &Formatter,
+        checker: MemberContext,
+    ) -> FormatResult<FormatElement> {
+        let text = name.text_trimmed();
+
+        if text.is_empty() {
+            return Ok(format_string_literal_token(name, formatter));
+        }
```
MichaReiser, line 723: This check is now redundant because of the check on line 727.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: > Can you please verify with the prettier implementation.

If you meant identifiers that contain `_`, then yes, prettier removes the quotes.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: We still needed to cover cases where we have empty strings:

```js
const a = { "": something }
```

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: This is the JS spec defining what a valid identifier is: https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-IdentifierName

Now try the different production with prettier to see when it removes the quotes. For example, prettier removes the quotes for `$a`. Does rome do so too?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,116 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+    checker: MemberContext,
+) -> FormatResult<FormatElement> {
+    fn replace_node(
+        name: JsSyntaxToken,
+        formatter: &Formatter,
+        checker: MemberContext,
+    ) -> FormatResult<FormatElement> {
+        let text = name.text_trimmed();
+
+        if text.starts_with(QUOTES_TO_OMIT) && text.ends_with(QUOTES_TO_OMIT) {
```
MichaReiser, line 726: Let's keep the check for `text.len() >= 2` to not remove quotes from `"` (not sure if that gets turned into an error token or not)

ematipico: It seems redundant. `text.len() > 2` and `text_to_check.is_empty()` do the same thing

crates/rome_js_formatter/src/utils/mod.rs
ematipico: It should be done, I think

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -649,3 +650,118 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+        }
+        // we split by the characters that can be considered valid identifiers
+        let mut patterns = text_to_check.split(['_', '$']);
+        match self {
+            MemberContext::Type => {
+                patterns.all(|sub_text| sub_text.chars().all(char::is_alphabetic))
+            }
+            MemberContext::Member => {
+                patterns.all(|sub_text| sub_text.chars().all(char::is_alphanumeric))
+            }
```
MichaReiser, line 709: Nit: can't this be written as

```suggestion
        match self {
            MemberContext::Type => {
                text.chars().all(|c| matches!(c, '_', '$') || c.is_alphabetic)
            }
            MemberContext::Member => {
                text.chars().all(|c| matches!(c, '_', '$') || c.is_alphanumeric)
            }
```

MichaReiser (approved):

MichaReiser:
Please address 

> Don't remove quotes from numeric properties in TypeScript because that can result in type checker errors (prettier doesn't remove the quotes as well).

In it's own PR

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Not sure, an identifier can be "_&£_foo", and this case we can't elide the quotes.

Would your suggestion cover  case?

MichaReiser: The problem here is that you can't remove the quotes because of £ and not because of the `_`. For example, prettier removes the quotes for `let a = {"_f_$$_oo": 1 }`

ematipico:
@NicholasLYang I am going to merge this one. If you have some concerns feel free to leave a comment and I will address them later on

Issue #2448, opened 2022-04-14
🐛 loop/if with empty statement body


jer3m01:
### Environment information

```bash
WSL (W11/Ubuntu)
Yarn 3.2.0
```

### What happened?

1. Create project using yarn 3.2.0
2. Install Rome 
3. Run `yarn format` (`rome format .`)
4. Rerun `yarn`
```
project$ yarn format
Formatted x files in xms

project$ yarn
project/.yarn/releases/yarn-3.2.0.cjs:37184
                                                !e && !this.buffer.length && !this[sA] && this.emit("drain");
                                                ^

SyntaxError: Unexpected token '!'
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1031:15)
    at Module._compile (node:internal/modules/cjs/loader:1065:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:12)
    at node:internal/main/run_main_module:17:47

```

### Expected result

Rome should correctly format cjs without breaking it or not touch cjs at all.

### Code of Conduct

- [X] I agree to follow Rome's Code of Conduct

MichaReiser:
[Playground](https://play.rome.tools/?lineWidth=80&indentStyle=space&quoteStyle=double&indentWidth=2&typescript=true&jsx=true&sourceType=module#ZABvADsAdwBoAGkAbABlACgAdABoAGkAcwBbAEMANgBdACgAdABoAGkAcwBbAGYARABdACgAKQApACkAOwAhAGUAJgAmACEAdABoAGkAcwAuAGIAdQBmAGYAZQByAC4AbABlAG4AZwB0AGgAJgAmACEAdABoAGkAcwBbAHMAQQBdACYAJgB0AGgAaQBzAC4AZQBtAGkAdAAoACIAZAByAGEAaQBuACIAKQA=)

This is probably a common issue in all places where an `EmptyStatement` is the body of a loop/if condition. The empty statement must be printed if it's the direct child of an if/while to print valid syntax.

leops:
Additionally `.yarn` should be added to the list of directories ignored by default in the formatting CLI

ematipico:
[...]

PR #2503, opened 2022-04-27
chore(rome_fs): add `svn`,  `hg` and `yarn` to ignored folders

ematipico:
## Summary

This PR adds `.svn`, `.hg` and `.yarn`

## Test Plan

No plan, it should work as is.


leops:
While we're at it, there was an issue where the formatter was reaching into the `.yarn` directory and messing up some scripts in there (#2448), while this is not a proper fix for the formatter we should probably also add `.yarn` to the list

ematipico:
It really seems we need a ignore file. I will create a placeholder task for that and add it to the comment as TODO

ematipico:
Added

leops (approved):

PR #2504, opened 2022-04-27
fix(rome_js_formatter): print empty statements as empty blocks

### LINKED ISSUE #2448: 🐛 loop/if with empty statement body

ematipico:
## Summary

Fixes #2448

## Test Plan

Added new cases, one for the `if` and one for `do while`

### DIFF
--- a/crates/rome_js_formatter/src/js/statements/empty_statement.rs
+++ b/crates/rome_js_formatter/src/js/statements/empty_statement.rs
@@ -1,13 +1,22 @@
-use crate::{empty_element, FormatElement, FormatNode, Formatter};
+use crate::{empty_element, Format, FormatElement, FormatNode, Formatter};
 use rome_formatter::FormatResult;
-
-use rome_js_syntax::JsEmptyStatement;
 use rome_js_syntax::JsEmptyStatementFields;
+use rome_js_syntax::{JsEmptyStatement, JsSyntaxKind};
+use rome_rowan::AstNode;
 
 impl FormatNode for JsEmptyStatement {
     fn format_fields(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
         let JsEmptyStatementFields { semicolon_token } = self.as_fields();
-
-        Ok(formatter.format_replaced(&semicolon_token?, empty_element()))
+        let parent_kind = self.syntax().parent().map(|p| p.kind());
+        if matches!(
+            parent_kind,
+            Some(JsSyntaxKind::JS_DO_WHILE_STATEMENT)
+                | Some(JsSyntaxKind::JS_IF_STATEMENT)
+                | Some(JsSyntaxKind::JS_ELSE_CLAUSE)
+        ) {
+            semicolon_token.format(formatter)
+        } else {
+            Ok(formatter.format_replaced(&semicolon_token?, empty_element()))
+        }
     }
 }
--- a/crates/rome_js_formatter/src/js/statements/empty_statement.rs
+++ b/crates/rome_js_formatter/src/js/statements/empty_statement.rs
@@ -1,5 +1,5 @@
-use crate::{empty_element, Format, FormatElement, FormatNode, Formatter};
-use rome_formatter::FormatResult;
+use crate::{empty_element, FormatElement, FormatNode, Formatter};
+use rome_formatter::{format_elements, hard_line_break, token, FormatResult};
 use rome_js_syntax::JsEmptyStatementFields;
 use rome_js_syntax::{JsEmptyStatement, JsSyntaxKind};
 use rome_rowan::AstNode;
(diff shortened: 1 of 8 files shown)


crates/rome_js_formatter/src/js/statements/empty_statement.rs
```diff
@@ -1,13 +1,22 @@
...
 use rome_js_syntax::JsEmptyStatementFields;
+use rome_js_syntax::{JsEmptyStatement, JsSyntaxKind};
+use rome_rowan::AstNode;
 
 impl FormatNode for JsEmptyStatement {
     fn format_fields(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
         let JsEmptyStatementFields { semicolon_token } = self.as_fields();
-
-        Ok(formatter.format_replaced(&semicolon_token?, empty_element()))
+        let parent_kind = self.syntax().parent().map(|p| p.kind());
```
NicholasLYang, line 10: Nit: this pattern shows up a couple times. Should we have a function that gets the parent kind directly?

NicholasLYang (approved):

ematipico:
> When it comes to unifying the printing behavior of empty statements, I would suggest we go the opposite way and extend the behavior of `if` statements (always insert a block) to all other affected statements (`for` loops, `for in` loops, `for of` loops, `while` loops, `do while` loops and `with` blocks). The reasoning behind this is that Rome JS had a `useBlockStatements` lint rule that emitted a fixable warning for these nodes if they were using a single statement as their body, but for this kind of "style fix" we might as well rewrite it directly in the formatter.

Does that mean that we want to intentionally diverge from prettier?

leops:
> Does that mean that we want to intentionally diverge from prettier?

Well at least I think we should discuss whether we want to have code-style related lint rules in the analyzer or to enforce these decisions in the formatter directly. Ultimately the decision is probably about how disruptive the change is (how often are these single-statement syntax actually used) vs. how confusing it is for users to see fixable diagnostics in their code for things they'd expect the formatter to take care of.

ematipico:
> > Does that mean that we want to intentionally diverge from prettier?
> 
> Well at least I think we should discuss whether we want to have code-style related lint rules in the analyzer or to enforce these decisions in the formatter directly. Ultimately the decision is probably about how disruptive the change is (how often are these single-statement syntax actually used) vs. how confusing it is for users to see fixable diagnostics in their code for things they'd expect the formatter to take care of.

I completely forgot about that lint rule. Yes, in Rome classic there was this idea where "coding style rules" should not exists because we own the formatter too. 

This is the rule FYI: https://github.com/rome/tools/blob/archived-js/internal/compiler/lint/rules/js/useBlockStatements.test.md

At this point I think we should just diverge from prettier. I agree with you @leops

yassere:
> At this point I think we should just diverge from prettier. I agree with you @leops

I don't think we should. This is a style decision, and we should have a very high bar for where we diverge from Prettier purely for style reasons (rather than implementation complexity).

We don't necessarily need to add every lint rule that existed in Rome classic. But if we do end up with some "code style rules", I think that's ok because we've intentionally relinquished some of our control over the formatting output by targeting Prettier equivalence.

ematipico:
@yassere Could let us know why though? Why you think we should not put blocks and why we don't need that lint rule anymore.

I think it makes sense to have blocks. Having empty statements can lead to errors, same for expressions without blocks.

NicholasLYang:
Personally I view diverging from Prettier as a large enough reason. If we get far enough in linting that we create this rule again, we can consider diverging then. But until that time, we should do what's best for the formatter, and not what might be best for the linter. IMO, as long as we're not seriously inhibiting future tools, we should do what's best for the tools that currently exist. Each of our tools needs to gain adoption and the promise of a linter feature does not outweigh breaking prettier compatibility, at least for me.

yassere (commented):
@ematipico I think it makes sense to have blocks too. Maybe that's something we can have as a safe autofix for the linter. But for the _formatter_, I think we need to respect our commitment to match Prettier's styling decisions as much as possible, unless we have a very good reason not to.

Given our previous messaging, intentionally deviating from Prettier's style is not a decision we should make lightly. Each case should be discussed in its own issue so that we can make a strong case to justify that divergence and document it.

PR #2508, opened 2022-04-27
fix(rome_js_parser): jsx vs type assertions

xunilrj:
This PR fixes a problem with jsx versus type assertions at primary expressions.

Currently, something like the code below was failing:

```
let a = <number>b;
```

We would treat ```b``` as ```JSXText``` and later complaint about a missing closing tag.
Now we do our best to parse JSX and type assertions separately and warn appropriately.

This changes an error on an HTML comment test, which does not matter because we still do not support these comments.
If anyone thinks that the current message is worst, I can try something to improve it.

### DIFF
--- a/crates/rome_js_parser/src/syntax/jsx/mod.rs
+++ b/crates/rome_js_parser/src/syntax/jsx/mod.rs
@@ -57,8 +57,12 @@ pub(crate) fn parse_jsx_tag_expression(p: &mut Parser) -> ParsedSyntax {
 
     let m = p.start();
 
-    parse_any_jsx_tag(p, true).unwrap();
-    Present(m.complete(p, JSX_TAG_EXPRESSION))
+    if parse_any_jsx_tag(p, true).is_present() {
+        Present(m.complete(p, JSX_TAG_EXPRESSION))
+    } else {
+        m.abandon(p);
+        Absent
+    }
 }
 
 // <a ...> or <a ... />
@@ -103,8 +107,12 @@ fn parse_any_jsx_tag(p: &mut Parser, in_expression: bool) -> ParsedSyntax {
 
             parse_jsx_children(p);
 
-            expect_closing_element(p, in_expression, name, opening_range);
-            Present(element.complete(p, JSX_ELEMENT))
+            if expect_closing_element(p, in_expression, name, opening_range).is_none() {
+                element.abandon(p);
+                Absent
+            } else {
+                Present(element.complete(p, JSX_ELEMENT))
+            }
         }
         None => Absent,
     }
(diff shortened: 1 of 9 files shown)


xunilrj:
About the regressions:

#### compiler/newExpressionWithCast.ts
We were incorrectly accepting this file. It does not have an associated ```.errors.txt``` to signal parsing errors, but it does contain parsing errors. So I think we are correct now. Specifically this part:

```ts
// parse error
var test2 = new <any>Test2();
```

#### conformance/parser/ecmascript5/Expressions/parserTypeAssertionInObjectCreationExpression1.ts

Exact same thing:

```
new <T>Foo()
```

xunilrj:
!bench_parser

ematipico (commented):
Looks good to me!

ematipico (approved):

crates/rome_js_parser/src/syntax/jsx/mod.rs
```diff
@@ -103,8 +107,12 @@ fn parse_any_jsx_tag(p: &mut Parser, in_expression: bool) -> ParsedSyntax {
...
             parse_jsx_children(p);
 
-            expect_closing_element(p, in_expression, name, opening_range);
-            Present(element.complete(p, JSX_ELEMENT))
+            if expect_closing_element(p, in_expression, name, opening_range).is_none() {
+                element.abandon(p);
+                Absent
+            } else {
+                Present(element.complete(p, JSX_ELEMENT))
+            }
```
MichaReiser, line 115: This is dangerous. It will progress the parser but return `Absent`. The processed tokens now get lost if the caller isn't using a checkpoint. Yes, you added a checkpoint inside of `parse_expression` but this function is also called when parsing the children of an element, in which case the parser silently skips over content.

crates/rome_js_parser/src/syntax/jsx/mod.rs
```diff
@@ -213,10 +221,18 @@ fn expect_closing_element(
...
+
+    if !p.expect(T![<]) {
+        m.abandon(p);
+        return None;
+    }
+
+    if !p.expect(T![/]) {
+        m.abandon(p);
+        return None;
+    }
```
MichaReiser, line 235: Same here. The parser has now eaten over at least one `<` and then returns `Absent`... which violates the constraint that parse methods should never progress the parser if they return `Absent`.

crates/rome_js_parser/test_data/inline/ok/type_assertion_primary_expression.ts
```diff
@@ -0,0 +1 @@
+let a = <number>undefined;
```
MichaReiser, line 1: This case works for me even without this fix in place. I'm going to revert this change. The slightly less good error message is tolerable in my view. We can add explicit handling of `let a = new<number> Test()` if we think this is a common pattern but we need to change the error message to mention that type assertion are disallowed in case JSX is enabled.

PR #2509, opened 2022-04-27
feature: AST Serialization

NicholasLYang:
## Summary

Navigating the AST/CST in the playground is a little difficult right now. We can't collapse nested nodes and there's no syntax highlighting. By serializing to JSON, we can use [React JSON View](https://github.com/mac-s-g/react-json-view). This will also let us doing other forms of testing like parser regression testing.

## Test Plan

Not sure how we'd test this. We could do regression testing to ensure that the serialization doesn't change.

Without `clean_up_json`:
Screen Shot 2022-05-05 at 1 08 14 PM

With `clean_up_json`:
Screen Shot 2022-05-05 at 1 10 03 PM

[Playground without `clean_up_json`](https://play.rome.tools/948d96e)


crates/rome_rowan/src/green/node.rs
```diff
@@ -9,12 +9,14 @@ use std::{
     ops, ptr, slice,
 };
 
+#[cfg(target_pointer_width = "64")]
```
xunilrj, line 12: Why do we need this?
How does this fit when targeting wasm32?

crates/rome_rowan/src/serde_impls.rs
```diff
@@ -71,13 +71,19 @@ impl<L: Language> Serialize for Children<&'_ SyntaxNode<L>> {
...
-    builder.token(crate::RawSyntaxKind(0), "\n\tlet ");
-    builder.finish_node();
-    let root = builder.finish();
+#[cfg(test)]
+mod test {
+    use crate::raw_language::{RawLanguage, RawLanguageKind, RawLanguageSyntaxFactory};
 
-    assert!(serde_json::to_string(&root).is_ok());
+    #[test]
+    pub fn serialization() {
```
xunilrj, line 79: Can we:
1 - parse;
2 - serialize;
3 - deserialize;
4 - rebuild the code;
5 - assert_eq them?

website/playground/src/utils.ts
```diff
@@ -5,7 +5,44 @@ import prettier from "prettier";
...
 import { IndentStyle, PlaygroundState, QuoteStyle, SourceType } from "./types";
 
-interface Size { width: number | undefined, height: number | undefined }
+interface Size { width: number | undefined; height: number | undefined }
+
+function isObject(n: any): boolean {
+	return typeof n === "object" && n !== null;
+}
+
+/** Simple helper function to clean up the AST object.
```
xunilrj, line 14: There is nothing we can do on the rust side to generate a better json?

crates/rome_rowan/src/green/node.rs
NicholasLYang: Oh this is just because we kept on getting a warning when building on Wasm that this import is unnecessary. It's because of [this line](https://github.com/rome/tools/blob/9851e7de647ac37b405626ece602f5292acec7ee/crates/rome_rowan/src/green/node.rs#L57)

crates/rome_rowan/src/serde_impls.rs
NicholasLYang: I didn't implement deserialization yet. That might be kind of hard with Rowan's structure. In my experience deserializing interned/arena-d structures is not fun.

website/playground/src/utils.ts
NicholasLYang: I just wrote an implementation that takes the AST, converts it to a generic `serde_json::Value`, then traverses it on the Rust side. I'm not convinced it'll be significantly faster, mostly because we're now doing two conversions: AST -> serde_json::Value -> String. But it is nicer in Rust haha

NicholasLYang:
I could leave the regular text option and include a toggle?

MichaReiser:
I do like the addition but have some concerns about the implementation. Mainly that it adds `serde` as a dependency of the parser and syntax and that it adds a lot of additional generated code. 

Overall, it seems that we have multiple use cases where we want to use the information from the grammar to generate some output:

* `Debug` implementation: Print each field by first printing the field name and then the content
* `Missing` check (https://github.com/rome/tools/pull/2176#pullrequestreview-892744070) 
* `JSON` serialization

That's why I think it may be time to instead have a reflection like API that allows quering additional information about a node. Given a kind, what are its fields, for each field, what's the optionality, the name, ...

The reflection information could be part of its own crate (or the same for now?) and would allow you to implement the Serde serialization in the playground only.

NicholasLYang:
Serde is listed as an optional dependency if I'm not mistaken. If the `serde` feature is not used, it shouldn't be included. I'm also confused as to how we could implement serde serialization in a different crate. How would that work with the orphan rule?

ematipico:
> Serde is listed as an optional dependency if I'm not mistaken. If the `serde` feature is not used, it shouldn't be included. I'm also confused as to how we could implement serde serialization in a different crate. How would that work with the orphan rule?

Serde has a workaround for it: https://serde.rs/remote-derive.html

website/playground/src/lib.rs
```diff
@@ -71,6 +87,31 @@ impl WriteColor for ErrorOutput {
     }
 }
 
+fn clean_up_json(json: serde_json::Value) -> serde_json::Value {
```
MichaReiser, line 90: Some documentation of what this function does would help to understand the logic.

website/playground/src/lib.rs
```diff
@@ -71,6 +87,31 @@ impl WriteColor for ErrorOutput {
...
+    match json {
+        serde_json::Value::Array(mut entries) => {
+            if entries.len() == 1 {
+                clean_up_json(entries.pop().unwrap())
+            } else {
```
MichaReiser, line 95: Does that mean that a list with only a single child node gets unwrapped?

website/playground/src/lib.rs
MichaReiser: This feels to me like the JSON serializer is kind of what you want but also not completely so it needs to be patched up. 

That's why I think it might be more useful to expose the metadata of nodes so that external crates can implement the serialization they need, using the shape described in the node metadata. However, that might get tricky with separated lists.

xtask/codegen/src/generate_nodes.rs
```diff
@@ -149,6 +149,28 @@ pub fn generate_nodes(ast: &AstSrc, language_kind: LanguageKind) -> Result<Strin
                 })
                 .unzip();
 
+            let (serialize_impl, slots_derive) = if language_kind == LanguageKind::Js {
```
MichaReiser, line 152: I would restrain from having language-specific code in our codegen for as long as possible. We should either enable this for all or no language.

crates/rome_js_syntax/src/generated/nodes.rs
```diff
@@ -44,6 +48,17 @@ impl ImportMeta {
...
+impl Serialize for ImportMeta {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.as_fields().serialize(serializer)
+    }
+}
+#[cfg_attr(feature = "serde", derive(Serialize))]
+#[cfg_attr(feature = "serde", serde(crate = "serde_crate"))]
```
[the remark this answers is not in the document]
NicholasLYang: Because we renamed the `serde` package to `serde_crate` so we can use `serde` as a feature name. Slight annoying cargo issue

website/playground/pnpm-lock.yaml
```diff
@@ -3179,12 +3356,23 @@ packages:
     resolution: {integrity: sha512-bKr1DkiNa2krS7qxNtdrtHAmzuYGFQLiQ13TsorsdT6ULTkPLKuu5+GsFpDlg6JFjUTwX2DyhMPG2be8uPrqsQ==}
     dev: false
 
+  /webidl-conversions/3.0.1:
```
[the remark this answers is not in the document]
NicholasLYang: Before:
```
vite v2.7.13 building for production...
✓ 839 modules transformed.
dist/assets/favicon.5e112f83.svg               1.10 KiB
dist/assets/rome_playground_bg.19165dd8.wasm   2228.59 KiB
dist/index.html                                0.61 KiB
dist/assets/vendor.f761b4ed.css                3.99 KiB / gzip: 0.59 KiB
dist/assets/index.81f4defd.js                  17.15 KiB / gzip: 4.65 KiB
dist/assets/index.f701b5bf.css                 13.22 KiB / gzip: 3.65 KiB
dist/assets/vendor.3bd07a61.js                 1877.24 KiB / gzip: 593.44 KiB
```
After
```
vite v2.7.13 building for production...
✓ 842 modules transformed.
dist/assets/favicon.5e112f83.svg               1.10 KiB
dist/assets/rome_playground_bg.19165dd8.wasm   2228.59 KiB
dist/index.html                                0.61 KiB
dist/assets/vendor.f761b4ed.css                3.99 KiB / gzip: 0.59 KiB
dist/assets/index.f701b5bf.css                 13.22 KiB / gzip: 3.65 KiB
dist/assets/index.9bf19a9f.js                  17.03 KiB / gzip: 4.67 KiB
dist/assets/vendor.e51ebfcc.js                 1995.12 KiB / gzip: 627.46 KiB
```

website/playground/src/DesktopPlayground.tsx
```diff
@@ -75,10 +76,10 @@ export default function DesktopPlayground(
 							/>
 						</TabPanel>
 						<TabPanel>
-							<pre className="h-screen overflow-y-scroll">{cst}</pre>
+							<ReactJson src={JSON.parse(cst)} />
 						</TabPanel>
 						<TabPanel>
-							<pre className="h-screen overflow-y-scroll">{ast}</pre>
+							<AstView ast={ast} />
```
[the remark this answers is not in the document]
NicholasLYang: Yeah, fixed that. Actually, they don't get rendered as far as I can tell. If we throw an error in AstView, it only shows up when you click on the AST tab.

website/playground/src/lib.rs
NicholasLYang: Well, we can't really change how Result or Vec types are serialized. That's an issue of the orphan rule. Not to mention, we may want the `Ok` wrapper for other cases.

website/playground/src/lib.rs
NicholasLYang: Yep!

xtask/codegen/src/generate_nodes.rs
NicholasLYang: Okay, I can implement it for all languages

website/playground/src/DesktopPlayground.tsx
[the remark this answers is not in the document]
MichaReiser: That's correct, moving `JSON.parse` into the tab implementation should make sure that the string is only parsed when the tab is visible.

website/playground/src/lib.rs
MichaReiser: This is more an argument why an implementation using a nodes Metadata and manually implementing a serializer would be "cleaner". 

Would you mind adding some documentation to this method to explain what happens

website/playground/pnpm-lock.yaml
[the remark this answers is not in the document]
ematipico: I suppose the biggest change will be the wasm file?

MichaReiser:
Please add some screenshots to the test plan demonstrating specific use cases (e.g. removal of OK/Err, removing of list wrapper if it's a single element list)

NicholasLYang:
Without `clean_up_json`:

```
{
  "JsFunctionDeclaration": {
    "async_token": null,
    "body": {
      "Ok": {
        "directives": [],
        "l_curly_token": {
          "Ok": {
            "kind": "L_CURLY",
            "text": "{",
            "text_range": [
              16,
              17
            ]
          }
        },
        "r_curly_token": {
          "Ok": {
            "kind": "R_CURLY",
            "text": "\n}",
            "text_range": [
              29,
              31
            ]
          }
        },
        "statements": [
          {
            "JsReturnStatement": {
              "argument": {
                "JsIdentifierExpression": {
                  "name": {
                    "Ok": {
                      "value_token": {
                        "Ok": {
                          "kind": "IDENT",
                          "text": "a",
                          "text_range": [
                            27,
                            28
                          ]
                        }
                      }
                    }
                  }
                }
              },
              "return_token": {
                "Ok": {
                  "kind": "RETURN_KW",
                  "text": "\n  return ",
                  "text_range": [
                    17,
                    27
                  ]
                }
              },
              "semicolon_token": {
                "kind": "SEMICOLON",
                "text": ";",
                "text_range": [
                  28,
                  29
                ]
              }
            }
          }
        ]
      }
    },
    "function_token": {
      "Ok": {
        "kind": "FUNCTION_KW",
        "text": "function ",
        "text_range": [
          0,
          9
        ]
      }
    },
    "id": {
      "Ok": {
        "JsIdentifierBinding": {
          "name_token": {
            "Ok": {
              "kind": "IDENT",
              "text": "foo",
              "text_range": [
                9,
                12
              ]
            }
          }
        }
      }
    },
    "parameters": {
      "Ok": {
        "items": [
          {
            "Ok": {
              "JsAnyFormalParameter": {
                "JsFormalParameter": {
                  "binding": {
                    "Ok": {
                      "JsAnyBinding": {
                        "JsIdentifierBinding": {
                          "name_token": {
                            "Ok": {
                              "kind": "IDENT",
                              "text": "a",
                              "text_range": [
                                13,
                                14
                              ]
                            }
                          }
                        }
                      }
                    }
                  },
                  "initializer": null,
                  "question_mark_token": null,
                  "type_annotation": null
                }
              }
            }
          }
        ],
        "l_paren_token": {
          "Ok": {
            "kind": "L_PAREN",
            "text": "(",
            "text_range": [
              12,
              13
            ]
          }
        },
        "r_paren_token": {
          "Ok": {
            "kind": "R_PAREN",
            "text": ") ",
            "text_range": [
              14,
              16
            ]
          }
        }
      }
    },
    "return_type_annotation": null,
    "star_token": null,
    "type_parameters": null
  }
}
```

With `clean_up_json`:
```
{
  "JsAnyStatement": {
    "JsFunctionDeclaration": {
      "async_token": null,
      "body": {
        "directives": [],
        "l_curly_token": {
          "kind": "L_CURLY",
          "text": "{",
          "text_range": [
            16,
            17
          ]
        },
        "r_curly_token": {
          "kind": "R_CURLY",
          "text": "\n}",
          "text_range": [
            29,
            31
          ]
        },
        "statements": {
          "JsReturnStatement": {
            "argument": {
              "JsIdentifierExpression": {
                "name": {
                  "value_token": {
                    "kind": "IDENT",
                    "text": "a",
                    "text_range": [
                      27,
                      28
                    ]
                  }
                }
              }
            },
            "return_token": {
              "kind": "RETURN_KW",
              "text": "\n  return ",
              "text_range": [
                17,
                27
              ]
            },
            "semicolon_token": {
              "kind": "SEMICOLON",
              "text": ";",
              "text_range": [
                28,
                29
              ]
            }
          }
        }
      },
      "function_token": {
        "kind": "FUNCTION_KW",
        "text": "function ",
        "text_range": [
          0,
          9
        ]
      },
      "id": {
        "JsIdentifierBinding": {
          "name_token": {
            "kind": "IDENT",
            "text": "foo",
            "text_range": [
              9,
              12
            ]
          }
        }
      },
      "parameters": {
        "items": {
          "JsAnyFormalParameter": {
            "JsFormalParameter": {
              "binding": {
                "JsAnyBinding": {
                  "JsIdentifierBinding": {
                    "name_token": {
                      "kind": "IDENT",
                      "text": "a",
                      "text_range": [
                        13,
                        14
                      ]
                    }
                  }
                }
              },
              "initializer": null,
              "question_mark_token": null,
              "type_annotation": null
            }
          }
        },
        "l_paren_token": {
          "kind": "L_PAREN",
          "text": "(",
          "text_range": [
            12,
            13
          ]
        },
        "r_paren_token": {
          "kind": "R_PAREN",
          "text": ") ",
          "text_range": [
            14,
            16
          ]
        }
      },
      "return_type_annotation": null,
      "star_token": null,
      "type_parameters": null
    }
  }
}
```

ematipico:
If possible, I would like to make the code sections in overflow and scrollable. Mostly because if I start scrolling using the main scroll bar of the window, and I dive into the AST, I lose sight of the source code. And now that we have more information on the right, this happens even with small snippets  like `const f =  () => {}`.

MichaReiser (approved):
I personally would prefer it if we always render lists as array, regardless of how many items they contain to make it clear to users that their IR must support multiple elements and that it isn't just one element.

PR #2512, opened 2022-04-28
feat(rome_js_formatter): import sorting

ematipico:
## Summary

This PR adds a experimental logic to the formatter to sort the import statements inside a module.

The logic is based on some assumptions:

- bare imports, meaning things are like `import "polyfill"` are not safe, meaning they have side effects;
- the rest of the imports are safe, meaning they are side effects free;

The feature is behind a feature flag.

## Test Plan

I created few tests, where we handle different statements, line breaks and cases of premature ending of the statements.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -753,9 +753,9 @@ dependencies = [
 
 [[package]]
 name = "insta"
-version = "1.11.0"
+version = "1.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


MichaReiser:
We'll need to figure out if we indeed want the sorting to be part of the formatter or if it's rather something that should happen inside of the linter. The main issue with doing it inside of the formatter is that it isn't a 100% safe operation, it can change the semantic of the program and this may not be something that the formatter should do. 

However, we can provide this as a refactor and allow users to (maybe even by default) opt in to run the sorting on safe.

Issue #2447, opened 2022-04-14
Spacing around block comments


MichaReiser:
Rome always inserts a whitespace character before and after a block comment. Prettier doesn't do so:

* Leading whitespace: if the comment is directly following a `(` token
* Trailing whitespace: if the comment is immediately followed by a `(` or `)` token.

[...]

PR #2515, opened 2022-04-28
fix(rome_js_formatter): don't add spaces to comments

ematipico:
## Summary

Fixes https://github.com/rome/tools/issues/2447

No big changes in this one, I had to make the iterator a `peekable` as it needs to check if there's a trailing trivia and keep the whitespace in that case.

## Test Plan

Added two test cases that cover the cases we want to fix


crates/rome_js_formatter/src/formatter.rs
```diff
@@ -533,31 +533,59 @@ impl Formatter {
...
+        let mut pieces = pieces.peekable();
+
+        while let Some(piece) = pieces.next() {
             if let Some(comment) = piece.as_comments() {
                 let is_single_line = comment.text().trim_start().starts_with("//");
 
+                let next_piece = pieces.peek();
+                let leading_trivia = if let Some(previous_piece) = previous_piece {
+                    if previous_piece.is_whitespace() {
+                        space_token()
```
ematipico, line 547: I am thinking that maybe we can do another optimisation here, where we create a new token out of the trivia, by keeping the same text.

It would an optimization similar to what Micha did recently. This would work only for comments, as they are the only ones that contain actual strings.

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -533,31 +533,59 @@ impl Formatter {
...
+                    } else {
+                        empty_element()
+                    }
+                } else {
+                    empty_element()
+                };
+
+                let trailing_trivia = if let Some(next_piece) = next_piece {
```
leops, line 555: Nit: this could be written as `if next_piece.map_or(false, SyntaxTriviaPiece::is_whitespace) { space_token } else { empty_element }`

yassere:
!bench_formatter

crates/rome_js_formatter/src/lib.rs
```diff
@@ -485,7 +485,7 @@ mod test {
     #[ignore]
     // use this test check if your snippet prints as you wish, without using a snapshot
     fn quick_test() {
-        let src = r#"xyz.a(b!).a(b!).a(b!)
+        let src = r#"strOrNum = unionWithRestParameter3(); // error no call signature
```
yassere, line 488: Is this just from local testing?

ematipico: Yes. Ignored on purpose

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -533,30 +533,51 @@ impl Formatter {
     {
         let mut elements = Vec::new();
 
-        for piece in pieces {
+        let mut has_seen_whitespace = false;
+
+        let mut pieces = pieces.peekable();
+
+        while let Some(piece) = pieces.next() {
```
NicholasLYang, line 540: If I'm reading this correctly, we should be checking for the pieces that are not comments to see if they are whitespace and changing `has_seen_whitespace` to true, no? Because currently we do not print a space when there's a line of code followed by a space then a single line comment: [Playground](https://play.rome.tools/1d07d6a/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#bABlAHQAIABhACAAPQAgADEAMAAgACAAIAAgACAALwAvACAAZgBvAG8ACgA=)

NicholasLYang (changes requested):

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -533,30 +533,51 @@ impl Formatter {
...
                 let is_single_line = comment.text().trim_start().starts_with("//");
 
+                let next_piece = pieces.peek();
+                let leading_trivia = if has_seen_whitespace {
+                    space_token()
+                } else {
+                    empty_element()
+                };
+
+                let trailing_trivia = if next_piece.map_or(false, SyntaxTriviaPiece::is_whitespace)
```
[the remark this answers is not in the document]
ematipico: I hoped this could have been consistent but it seems that the spacing of the comments depends on where they are placed.. meaning in which nodes they are.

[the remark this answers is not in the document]
MichaReiser: Yes, the formatting is specific to what the previous/next token is. See the related issue, I outlined the rules.

MichaReiser:
Prettier categories comments as:

* Leading: In front of a node `class /* in front of identifier Test */ Test {}
* Trailing: After a node, but not followed by a whitespace and only if it is the first comment: `class Test {} /* Trailing for class node */` 
* Dangling: Between two tokens: `function test(/*no arguments*/) {}` 

And Prettier has an extensive rules on when it diverges from these simple rules 

https://github.com/prettier/prettier/blob/1dffa0f9dac571bf0e03969c11fd4548efd44d7a/src/language-js/comments.js#L1-L1036

crates/rome_js_formatter/src/formatter.rs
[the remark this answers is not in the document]
MichaReiser: I'm currently looking into how we should format comments. I do like this implementation but believe that we should instead test if the next/previous token is a punctuation token and, if so, not insert a space. I haven't yet figured out how to do this best in here with skipped token trivia but I'm honestly inclined to just bail out formatting for skipped token trivia

MichaReiser:
I'm closing this PR because #2701 implements the spacing before/after inline comments based on the previous/next token.

PR #2516, opened 2022-04-28
fix(ci): fix the release workflows for the CLI and LSP

leops:
## Summary

This corrects two outstanding issues in the release workflows following #2495:

- The version number injected into the CLI binary was read from an environment variable set in a later step so the version name ended up being empty, it's now being read from the correct source
- The builds of the VSCode extension were failing as the `dist` directory used as a target location for the compiled bundle did not exist, there's now an explicit `mkdir` to create it beforehand
- All builds of the CLI npm package were being tagged as pre-release since it's using a version tagged with `-next`, the logic for checking this has been disabled so that only manual dispatch of the workflow are tagged as pre-release

## Test Plan

Manually dispatch the workflows and see if it works

### DIFF
--- a/.github/workflows/release_cli.yml
+++ b/.github/workflows/release_cli.yml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Check prerelease status
         id: prerelease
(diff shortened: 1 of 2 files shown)


ematipico (approved):

leops:
I'm reverting the version number of the extension in order to change it back again once this PR is merged to trigger the deployment, this will also trigger a deployment when this PR gets merged but I'll just reject it

ematipico (approved):

xunilrj (approved):

PR #2519, opened 2022-04-29
fix(rome_js_formatter): number with trailing whitespace

### LINKED ISSUE #2501: 🐛 Fail to format `Oq = wq(1 .toString);`
### Environment information

```bash
Latest version of Rome
```

### What happened?

Given `Oq = wq(1 .toString);`

The formatter doesn't preserve the space. [Playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=true&sourceType=module#TwBxACAAPQAgAHcAcQAoADEAIAAuAHQAbwBTAHQAcgBpAG4AZwApADsACgA=)

### Expected result

It should match Prettier's formatting and not result in a broken code.

### Code of Conduct

mrkldshv:
## Summary

Fixes #2501.

This PR adds conditional to `JsStaticMemberExpression` that checks if its value is a number literal expression that has trailing whitespace.

## Test Plan

Added new test that covers the case.

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
@@ -28,8 +28,7 @@ impl FormatNode for JsStaticMemberExpression {
         let formatted_object = object?.format(formatter)?;
 
         if is_object_number_literal && (has_object_trailing_trivia || has_operator_leading_trivia) {
-            let (object_leading, object_content, object_trailing) =
-                formatted_object.clone().split_trivia();
+            let (object_leading, object_content, object_trailing) = formatted_object.split_trivia();
 
             Ok(group_elements(format_elements![
                 object_leading,
--- a/crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
@@ -1,4 +1,7 @@
-use crate::{format_elements, group_elements, Format, FormatElement, FormatNode, Formatter};
+use rome_js_syntax::JsSyntaxKind;
+use rome_rowan::AstNode;
+
+use crate::{format_elements, group_elements, token, Format, FormatElement, FormatNode, Formatter};
 use rome_formatter::FormatResult;
 
 use rome_js_syntax::JsStaticMemberExpression;
@@ -12,10 +15,35 @@ impl FormatNode for JsStaticMemberExpression {
             member,
         } = self.as_fields();
 
-        Ok(group_elements(format_elements![
-            object.format(formatter)?,
-            operator_token.format(formatter)?,
-            member.format(formatter)?,
-        ]))
+        let object_syntax = object.clone()?.syntax().clone();
+
+        let is_object_number_literal =
+            object_syntax.kind() == JsSyntaxKind::JS_NUMBER_LITERAL_EXPRESSION;
+
+        let has_object_trailing_whitespace = object_syntax
+            .text()
+            .to_string()
+            .chars()
+            .last()
+            .unwrap()
+            .is_whitespace();
+
+        let formatted_object = object?.format(formatter)?;
+
+        if is_object_number_literal && has_object_trailing_whitespace {
+            Ok(group_elements(format_elements![
+                token("("),
+                formatted_object,
+                token(")"),
+                operator_token.format(formatter)?,
+                member.format(formatter)?,
+            ]))
+        } else {
+            Ok(group_elements(format_elements![
+                formatted_object,
+                operator_token.format(formatter)?,
+                member.format(formatter)?,
+            ]))
+        }
     }
 }
--- a/crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
@@ -20,21 +20,23 @@ impl FormatNode for JsStaticMemberExpression {
         let is_object_number_literal =
             object_syntax.kind() == JsSyntaxKind::JS_NUMBER_LITERAL_EXPRESSION;
 
-        let has_object_trailing_whitespace = object_syntax
-            .text()
-            .to_string()
-            .chars()
-            .last()
-            .unwrap()
-            .is_whitespace();
+        let has_object_trailing_trivia =
+            object_syntax.last_trailing_trivia().unwrap().pieces().len() > 0;
+        let has_operator_leading_trivia =
+            operator_token.clone()?.leading_trivia().pieces().len() > 0;
 
         let formatted_object = object?.format(formatter)?;
 
-        if is_object_number_literal && has_object_trailing_whitespace {
+        if is_object_number_literal && (has_object_trailing_trivia || has_operator_leading_trivia) {
+            let (object_leading, object_content, object_trailing) =
+                formatted_object.clone().split_trivia();
+
             Ok(group_elements(format_elements![
+                object_leading,
                 token("("),
-                formatted_object,
+                object_content,
                 token(")"),
+                object_trailing,
                 operator_token.format(formatter)?,
                 member.format(formatter)?,
             ]))
--- a/crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
+++ b/crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
@@ -1,1 +1,6 @@
 123 .toString
+123
+/****/.toString
+123/**/.toString
+123
+.toString
--- a/crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
+++ b/crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
@@ -1,0 +1,1 @@
+123 .toString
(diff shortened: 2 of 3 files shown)


crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
```diff
@@ -0,0 +1 @@
+123 .toString
```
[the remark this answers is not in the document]
ematipico: The issue here is that the formatter emitted `123.toString();` which is not a valid JS program, in fact it shows a syntax error.

crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
```diff
@@ -12,10 +15,35 @@ impl FormatNode for JsStaticMemberExpression {
...
+        let is_object_number_literal =
+            object_syntax.kind() == JsSyntaxKind::JS_NUMBER_LITERAL_EXPRESSION;
+
+        let has_object_trailing_whitespace = object_syntax
+            .text()
+            .to_string()
+            .chars()
+            .last()
+            .unwrap()
+            .is_whitespace();
```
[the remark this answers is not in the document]
ematipico: Yes, there's a better way.

1. Look at the [playground](https://play.rome.tools/1d07d6a/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=true&jsx=false&sourceType=module#MQAyADMAIAAuAHQAbwBTAHQAcgBpAG4AZwA=) and check the "AST" tab, you will see that the `value_token` has trivia token attached (`Whitespace`)
2. We have APIs to retrieve those trivias, so you don't need to check the actual text
3. Use the API `object_syntax.trailing_trivia().pieces()` which will give you the iterator of trivias

Now, in this case, we don't need to check for white spaces because we need to cover the following case:
```js
123/****/.toString();

123
.toString();
```
Which means that you just need to check if `pieces().len() > 0`. If so, then we add parenthesis.

crates/rome_js_formatter/src/js/expressions/static_member_expression.rs
```diff
@@ -12,10 +15,35 @@ impl FormatNode for JsStaticMemberExpression {
...
+
+        let formatted_object = object?.format(formatter)?;
+
+        if is_object_number_literal && has_object_trailing_whitespace {
+            Ok(group_elements(format_elements![
+                token("("),
+                formatted_object,
+                token(")"),
```
ematipico, line 37: I would suggest to use `hard_group_elements` here, so possible comments are kicked outside of the parenthesis.

```suggestion
				hard_group_elements(format_elements![
        	        token("("),
    	            formatted_object,
	                token(")"),				
				]),
```

crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
ematipico, line 1: Could you add more cases? 

```js
123
/****/.toString;
123/**/.toString;
123
.toString;
```

ematipico (changes requested):
Thank you for picking up this! There are some edge cases that we have to cover (thank you JS), I added them to the comments.

ematipico:
> @ematipico Thank you for your review and suggestions!
> 
>     * As for getting trailing trivia from object (referring to this [comment](https://github.com/rome/tools/pull/2519#discussion_r861636117)): `object_syntax` is `SyntaxNode` and it has `last_trailing_trivia` on it. Do we need to use this API or we should get first token of `object_syntax` and use `trailing_trivia` on it?

Yeah `last_trailing_trivia` is totally fine :) I missed that API

crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js
mrkldshv: Sure, will add them. Although, when I use `hard_group_elements` it doesn't kick comments outside of parenthesis. 
For example:
```js
123/**/.toString;
```
gets formatted to:
```js
(123 /**/ ).toString;
```

ematipico: Oh yeah, so in this case you have to use a specific API for this edge case. No need for hard group then. 

```rust
let (leading, content, trailing) = formatted_object.split_trivia();
```

This API will split the formatted element, by giving you leading triva, content and trailing trivia.

Then you can decide the order of the formatting:
```rs
Ok(format_elements![
	leading,
	trailing,
	token("("),
	content,
	token(")"),
])
```

crates/rome_js_formatter/tests/specs/js/module/number/number-with-space.js.snap
```diff
@@ -0,0 +1,26 @@
...
+## Output 1
+-----
+Indent style: Tab
+Line width: 80
+Quote style: Double Quotes
+-----
+(123).toString;
+(123)
+/****/ .toString;
+(123) /**/ .toString;
```
ematipico, line 24: @rome/staff 

Could you please leave your feedback here? Prettier is not consistent about how the trailing/leading comments are placed between the number and the operator, [check the link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBGATAZgPQCpfYB0MEAyjAE4CWUA5gDpSOMaaN4HFmU20gA0ICAAcYVaAGdkoAIYUKEAO4AFOQikoZAG0UyAnlMEAjCjLABrODFIyAtnAAyNOMgBm2iXGOmLV0sLNeZEoAVy8QOFsjOAATGNiHGToQmVo4ADEIClsZGDE6ZBAZEJIBEAALGFstAHVyqngJALA4UnUGqgA3Br1CsAlDEBpPChhlU1octw9wgCsJAA9SXi04AEUQiHhprU9BAIoRwpg9YTgJMGpRMuFqWBqqGJhy5AAOAAZ9hU8a02FC27nOAUTouQQAR028HGIg0RQkAFooHBYrEyhQ4JCqBjxqkpkh3Ltwp5bFRghQwoIJCt1lCXASZoIYDIjA8ni8kOgmaYqFpeABhCC2fERCQAVjKIU8ABUWRpQuFOmEAJJQeKwUiXKiiACCavIelWO08AF8TUA).

Could you please choose your preference?
1. move comments on the very left, before the opening `(`
2. leave the comments between the number and the operator

MichaReiser: It seems to me that prettier is consistent:

* If there's no new line, leave it as is
* If there's a newline before the comment, move it to the front. 

And this logic isn't specific to number formatting, it also happens for `a /* */ .toString()`. 

I would probably tackle that separately if it's something we want to change.

ematipico: > And this logic isn't specific to number formatting, it also happens for a /* */ .toString().

I just used this on prettier playground and the behaviour doesn't match. Did mean something else by any chance?

MichaReiser: [Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDA9AKk+gdDCAZRgCcBLKAcwB0pbbVasd8jSLKQAaECABxhloAZ2ShUJEhADuABQkJRKVABtpqAJ6ieAIxKowAazgxCqALZwAMhTjIAZquFxd+oycJ8DHZKQCuLiBw5jpwACZh4VaoVH6olHAAYhAk5qgwglTIaH4E3CAAFjDmKgDqBWTwwl5gcISKlWQAbpUa2WDC2iAUziQwsvqUaQ5OgQBWwgAehBwqcACKfhDwIyrOPF4kvdkwGnxwwmDkAvl85LClZGEwBcgAHAAMG1LOpfp82WcHcCRNdjwARyW8AG-CUaGEAFooHBwuF8iQ4ECyIiBvFhkhHGtAs5zGRfCQAjxhLMFsC7JjRjwYKgdJdrrckAAman6MgqDgAYQg5gxQWEAFZ8n5nAAVWlKfyBJoBACSUEisEIRzIAgAggriBo5qtnABfPVAA)

Input:

```
a/**/.toString


a
/**/.toString
```

Output:

```
a /**/.toString;

/**/ a.toString;
```

Or is this about inserting the parentheses? It then probably is a problem with all expressions that must be parenthesized.

ematipico: It's wasn't the same code you pasted 😅 [here](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDA9AKgASfdgOhggGUYAnASygHMAKASgB0oWMc9sWjSLr7mrIahZZc+HmSq0QAGhAQADjErQAzslCpy5CAHcACtoQaUqADZ7UATw3yARuVRgA1nBglUAWzgAZanDIAGYWanAOTq7uJIrO-MgUAK7hIHBe9nAAJplZvqi0iag0cABiEOReqDAqMmaJxHIgABYwXuYA6k2U8GqxYHAkJt2UAG7d1sjganYg1GHkMAZONJXBoSkAVmoAHiT85nAAiokQ8GvmYfKx5POTMNaKcGpgVMqNitIw7ZSZME3IAA4AAxXXRhdpORSTD5PODkEaBeQARxO8CWSlMaDUAFooHAsllGuQ4CjKMSlkVVkgQhcUmEvJQEuRkvI1PsjqjAtT1vIYKh7N9fv8kAAmXlOSjmfgAYQgXipqTUAFZGokwgAVfmmJIpEbJACSUBysBIL0oygAgkayNYDucwgBfB1AA)

ematipico (approved):

Issue #2400, opened 2022-04-13
📎 TypeScript 4.7


MichaReiser:
### Description

Add parsing and formatter support for the Syntax introduced by TypeScript 4.7

* [x] Instantiation Expressions: [PR](https://github.com/microsoft/TypeScript/pull/47607), [Announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#instantiation-expressions) #3147
* [x] `extends` constraints on `infer` type variables: [Announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#extends-constraints-on-infer-type-variables) #4018
* [ ] Type parameter variance: [Announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#optional-variance-annotations-for-type-parameters) @nissy-dev  
* [ ] ~`typeof` on `private` fields: [Announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#typeof-on-private-fields)~ Being [reverted](https://github.com/microsoft/TypeScript/pull/48959)

nissy-dev:
I'll take the support for "extends constraints on infer type variables" 🚀

Louis-Michalak:
I'll try to take the support for variance annotations 😃

Louis-Michalak:
Is anyone currently supporting the type parameter variance?

MichaReiser:
> Is anyone currently supporting the type parameter variance?

No one is working on this at the moment. Do you want to tackle it?

Louis-Michalak:
> No one is working on this at the moment. Do you want to tackle it?

Yes I'll take it

LouisStairwage:
@MichaReiser by the way why has this PR been closed https://github.com/rome/tools/pull/2556

Louis-Michalak:
@MichaReiser by the way why has this PR been closed https://github.com/rome/tools/pull/2556

Conaclos:
Hi [Louis-Michalak](https://github.com/Louis-Michalak)! Did you make some progress?

nissy-dev:
This issue is assigned to @Louis-Michalak , but the PR is not submitted for a long time. 
Therefore, I will take the "Type parameter variance"🚀

PR #2524, opened 2022-05-02
Feat(rome_js_parser): typeof private property name

IWANABETHATGUY:
## Summary
part of https://github.com/rome/tools/issues/2400
typeof on private fields: [Announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#typeof-on-private-fields)

## Test Plan
I copy the test case from official pull request https://github.com/microsoft/TypeScript/pull/47696/files#diff-c7d8064feb0c4b5ba2cb75b7987eefc92a880fded5f86a9564f5d30ffef48bff, and pass it.

### DIFF
--- a/xtask/codegen/js.ungram
+++ b/xtask/codegen/js.ungram
@@ -2129,7 +2129,7 @@ TsAnyName =
 TsQualifiedName =
 	left: TsAnyName
 	'.'
-	right: JsName
+	right: JsAnyName
 
 TsTypeArguments =
 	'<'
--- a/crates/rome_js_parser/src/syntax/js_parse_error.rs
+++ b/crates/rome_js_parser/src/syntax/js_parse_error.rs
@@ -23,6 +23,10 @@ pub(crate) fn expected_class_member_name(p: &Parser, range: TextRange) -> Diagno
     .to_diagnostic(p)
 }
 
+pub(crate) fn expected_private_field_name(p: &Parser, range: TextRange) -> Diagnostic {
+    expected_node("private field name", range).to_diagnostic(p)
+}
+
 pub(crate) fn expected_arrow_body(p: &Parser, range: TextRange) -> Diagnostic {
     expected_any(&["function body", "expression"], range).to_diagnostic(p)
 }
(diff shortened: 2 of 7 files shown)


IWANABETHATGUY:
!bench_parser

IWANABETHATGUY:
@MichaReiser could you help with reviewing my pull request about the new typescript 4.7 syntax?

xtask/codegen/js.ungram
```diff
@@ -2129,7 +2129,7 @@ TsAnyName =
 TsQualifiedName =
 	left: TsAnyName
 	'.'
-	right: JsName
+	right: JsAnyName
```
MichaReiser, line 2132: I'm having concerns about adding `JsAnyName` to the right-hand side. This will not only allow private names in `typeof` expressions but anywhere else where `TsQualifiedName` is used (and none of these allow private names). 

Can you go through all types that have a field with the type `TsQualifiedName` and check if TS now allows the usage of private names for these use cases? For example, `TsImportTypeQualifier`, `TsNameWithTypeArguments`, `TsReferenceType`, `TsAnyModuleReference`

Does TS support `typeof this.#a.b`?

IWANABETHATGUY: > p

Ts support `typeof this.#a.b` , https://www.typescriptlang.org/play?#code/FAYwNghgzlAEAqsDexZtgYgrAvMiAXLAOTEC+q6ALgKZRUAUAlMpetQJ4AONA9gGawqACwCWUAHRYJEANxtYFMkA, but I don't have enough knowledge about Ts specification, so I can't write code snippet  include the ast node you mentioned above.

MichaReiser: We'll need to understand the exact change of typescript to model the AST correctly. 

Are you able to find the related PR in the TypeScript repository?

IWANABETHATGUY: > 
https://github.com/microsoft/TypeScript/pull/47696, this one

MichaReiser: Reading the code of this PR should help you understand which rules allow private names and which do not (`allowPrivateIdentifiers` parameter). 

The next step is then to go to TypeScript's playground and try to come up with examples for the cases where private names are/aren't allowed.

jakebailey:
We're [reverting](https://github.com/microsoft/TypeScript/pull/48959) the linked change in TS, so you probably shouldn't merge this (we may not even allow this syntax).

MichaReiser:
> We're [reverting](https://github.com/microsoft/TypeScript/pull/48959) the linked change in TS, so you probably shouldn't merge this (we may not even allow this syntax).

Thanks @jakebailey for letting us know!

PR #2526, opened 2022-05-02
refactor: Split out `JsFormatter`

MichaReiser:
This PR extracts many of the methods defined on `Formatter` into standalone utiltiy functions
or the `JsFormatter` trait. This is a first step to make the `Formatter` trait language agnostic. 

It isn't possible yet to move the `Formatter` and `Format` traits to `rome_formatter` because
it then becomes impossible to implement `Format` for any `AstNode` inside of `rome_js_formatter`. 
This is likely going to be a larger change and I outlined some ideas in https://github.com/rome/tools/discussions/2525

### DIFF
--- a/crates/rome_js_formatter/src/formatter.rs
+++ b/crates/rome_js_formatter/src/formatter.rs
@@ -563,21 +563,6 @@ impl Formatter {
         concat_elements(elements)
     }
 
-    /// "Formats" a node according to its original formatting in the source text. Being able to format
-    /// a node "as is" is useful if a node contains syntax errors. Formatting a node with syntax errors
-    /// has the risk that Rome misinterprets the structure of the code and formatting it could
-    /// "mess up" the developers, yet incomplete, work or accidentally introduce new syntax errors.
-    ///
-    /// You may be inclined to call `node.text` directly. However, using `text` doesn't track the nodes
-    /// nor its children source mapping information, resulting in incorrect source maps for this subtree.
-    ///
-    /// These nodes and tokens get tracked as [FormatElement::Verbatim], useful to understand
-    /// if these nodes still need to have their own implementation.
-    pub fn format_verbatim(&self, node: &JsSyntaxNode) -> FormatElement {
-        let verbatim = self.format_verbatim_node_or_token(node);
-        FormatElement::Verbatim(Verbatim::new_verbatim(verbatim, node.text_range().len()))
-    }
-
     /// Formats unknown nodes. The difference between this method  and `format_verbatim` is that this method
     /// doesn't track nodes/tokens as [FormatElement::Verbatim]. They are just printed as they are.
     pub fn format_unknown(&self, node: &JsSyntaxNode) -> FormatElement {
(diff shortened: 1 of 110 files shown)


crates/rome_js_formatter/src/formatter.rs
```diff
@@ -63,110 +65,517 @@ impl Formatter {
...
+
+enum DelimitedContent {
+    BlockIndent(FormatElement),
+    SoftBlockIndent(FormatElement),
+    SoftBlockSpaces(FormatElement),
+}
+
+/// JS specific formatter extensions
+pub(crate) trait JsFormatter {
+    fn formatter(&self) -> &Formatter;
```
leops, line 532: Nit: since the intent is this trait will be implemented directly on `Formatter`, this could be `as_formatter` instead

crates/rome_js_formatter/src/js/assignments/array_assignment_pattern.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{Format, FormatElement, FormatNode, Formatter};
+use crate::{Format, FormatElement, FormatNode, Formatter, JsFormatter};
```
leops, line 1: We probably should add these traits to the prelude of `rome_js_formatter` and glob-import that in all the node formatters instead of having to rewrite all these imports on every refactor

MichaReiser: Yes, preludes is something I want to tackle once I figured out a stable module structure because changing the imports manually is rather painful ;)

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -63,110 +65,517 @@ impl Formatter {
...
-    /// Calling this method is required to correctly handle the comments attached
-    /// to the opening and closing tokens and insert them inside the group block
-    fn format_delimited(
-        &self,
-        open_token: &JsSyntaxToken,
-        content: impl FnOnce(FormatElement, FormatElement) -> FormatElement,
-        close_token: &JsSyntaxToken,
-    ) -> FormatResult<FormatElement> {
+    /// Tracks the given token as formatted
+    pub fn track_token<L: Language>(&self, token: &SyntaxToken<L>) {
```
ematipico, line 69: Looking at the warning, maybe the whole function can stay behind `#[cfg(debug_assertions)]`

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -63,110 +65,517 @@ impl Formatter {
...
-            format_elements![open_token_trailing_trivia, soft_line_break_or_space()]
-        } else {
-            empty_element()
-        };
-        let close_token_leading_trivia = if !close_token_leading_trivia.is_empty() {
-            format_elements![soft_line_break_or_space(), close_token_leading_trivia]
-        } else {
-            empty_element()
-        };
+    pub fn assert_formatted_all_tokens<L: Language>(&self, root: &SyntaxNode<L>) {
```
ematipico, line 77: Same here, it seems

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -63,110 +65,517 @@ impl Formatter {
...
+                let printed_tokens = self.printed_tokens.borrow();
+                printed_tokens.assert_all_tracked(root);
+            }
+        }
+    }
+
+    /// Formats all elements and returns the formatted result
+    ///
+    /// Returns [Err] if any child couldn't be formatted.
+    pub fn format_all<T: Format>(
```
ematipico, line 89: I think "all" is misleading. It formats an iterator (hence a list). I prefer the old name

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -279,412 +664,108 @@ impl Formatter {
...
-    }
 }
 
-/// Determines if the whitespace separating comment trivias
-/// from their associated tokens should be printed or trimmed
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub(super) enum TriviaPrintMode {
-    Full,
-    Trim,
+impl JsFormatter for Formatter {
```
ematipico, line 697: Can we add a small description to this new trait?

crates/rome_js_formatter/src/formatter.rs
MichaReiser: What warning? The downside is that it means that all call sites must also be wrapped in `#[cfg_if]` making it super verbose. The compiler will realise that this function is empty and remove all call sites.

crates/rome_js_formatter/src/formatter.rs
MichaReiser: `format_nodes` is no longer fitting because it can format non-nodes too. The idea was that it's similar to `iterator.all()` that returns true if all elements matches. This method formats all nodes of the iterator or returns early if any of them fail to format.

crates/rome_js_formatter/src/formatter.rs
MichaReiser: There's a small description on the trait itself (but not on the impl)

```
/// JS specific formatter extensions
pub(crate) trait JsFormatter {
```

crates/rome_js_formatter/src/formatter.rs
ematipico: This warning

Screenshot 2022-05-03 at 13 45 37

crates/rome_js_formatter/src/formatter.rs
ematipico: Completely missed it, thank you

ematipico (approved):

PR #2527, opened 2022-05-02
feature(formatter): Clean JSX Text

### LINKED ISSUE #2359: 📎 format of JsxText should clean jsx text

NicholasLYang:
## Summary

Continuing the work on #2360. Closes #2359. Here's the summary from the old PR:

JsxText should be normalized / cleaned. This means that, for JsxText nodes, the formatter should replace any two or more ascii_whitespace characters (tab, space, or newline) in a row with a single space

## Test Plan

local unit tests:
- in `jsx/auxiliary/text.rs`, these just call `clean_jsx_text` function.

snapshot tests (after creating the text.jsx file in `tests/...`)
- `touch crates/rome_js_formatter/tests/spec_tests.rs && cargo test -p rome_js_formatter formatter`

I used `cargo insta review` to review my newly generated snapshot test, and hit `a` to accept it.


NicholasLYang:
The branch I inherited had some changes to formatting for JsxElement. I'm open to moving these to a different PR if the scope is too large

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,10 +1,278 @@
...
+struct TextCleaner<'a> {
+    pub text: &'a str,
+    pub leading_whitespace_type: Option<WhitespaceType>,
+    /// Whitespace ranges are the ranges of text that contain whitespace. We keep track of them
+    /// so that on our second pass, we strip them out.
+    ///
+    ///  "A  Brighter \n\t Summer  \n\n Day"
+    ///    ^^        ^^^^^^      ^^^^^^^
+    ///
+    pub whitespace_ranges: Vec<Range<usize>>,
```
MichaReiser, line 31: This will always allocate a new vector for every jsx text. Is there a way we can avoid allocating that vector all together? 

My idea would be to implement an iterator that returns the cleaned characters. The formatter can then iterate over the cleaned and actual characters and if any differ, create a new string instead (similar to what @xunilrj mentioned on the original PR)

This would be similar to how we detect leading and trailing whitespace inside of the lexer. It requires tracking some state information like if there has been a newline between the last non-whitespace character and the next non-whitespace character.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -1,10 +1,278 @@
...
+
+        let mut current_whitespace_range_idx = 0;
+
+        // Invariant: idx is **never** larger than the end of the current whitespace range
+        for (idx, c) in char_indices {
+            let current_whitespace_range = self.whitespace_ranges.get(current_whitespace_range_idx);
+            if let Some(range) = current_whitespace_range {
+                // If the index is the end of the current whitespace range,
+                // then we increment the whitespace range index and
+                // push on an empty string.
```
MichaReiser, line 118:
```suggestion
                // push on a whitespace character
```

ematipico:
> The branch I inherited had some changes to formatting for JsxElement. I'm open to moving these to a different PR if the scope is too large

I think we can remove changes to `JSXElement` and just keep the clean text work here. Clean text is quite a troublesome work.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
xunilrj: Yes. I am running some benchmarks here to compare possible solutions. Will post my findings below.

NicholasLYang: I'm open to other solutions, but I would say that we should avoid bike shedding this issue too much. We've already had a fair amount of debate, and we can always re-open this issue if the implementation is too slow.

With the iterator, that seems like it'd be a little difficult with a string slice, as the iterator would need to have the same lifetime as the slice. I could see a solution where we do a pass to see if there is any redundant whitespace and if we encounter some, we immediately bail and switch into new string mode.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
NicholasLYang: I'd say "a space character" actually because it's not any whitespace character.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
NicholasLYang: Also I'll note that unless we come across a whitespace range, we [won't actually allocate](https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees). And unless the whitespace range happens to be a single space (which we can exclude if we wish), we will have to replace the string, which means we will have to allocate a new string. Therefore we will either allocate a new string *and* allocate a vector, or we will allocate neither.

xunilrj:
I am still a strong believer in a simpler implementation for this. We can solve this with a single iteration: advance as much as possible if the string is ok, then allocate if needed.

I think the numbers below demonstrate this.

They are comparisons between this PR implementation and one implementation I managed to cook here to assess the performance (see code at the bottom); and then against what would be a minimum threshold, something that does not solve the problem, but it is necessarily faster than any possible real implementation.

#### Current implementation

```rust
fn clean_jsx_ok() {
    let a = rome_js_formatter::jsx::auxiliary::text::clean_jsx_text(iai::black_box("Hey! Spaces!"));
}
```

```
clean_jsx_ok
  Instructions:                1389
  L1 Accesses:                 1867
  L2 Accesses:                    5
  RAM Accesses:                  48
  Estimated Cycles:            3572
```

```rust
fn clean_jsx_nok() {
    let a = rome_js_formatter::jsx::auxiliary::text::clean_jsx_text(iai::black_box(
        " \r\n\t Hey! \t\n\r\n     Spaces   !   ",
    ));
}
```

```
clean_jsx_nok
  Instructions:                2241
  L1 Accesses:                 2989
  L2 Accesses:                    5
  RAM Accesses:                  56
  Estimated Cycles:            4974
```

#### My implementation

The strings used are the same.

```
clean_jsx_ok
  Instructions:                 270
  L1 Accesses:                  292
  L2 Accesses:                    2
  RAM Accesses:                  18
  Estimated Cycles:             932

clean_jsx_nok
  Instructions:                 855
  L1 Accesses:                 1023
  L2 Accesses:                    2
  RAM Accesses:                  23
  Estimated Cycles:            1838
```

#### Minimum threshold

The minimum fake implementation to compare us against could be:

```rust
pub fn fake_clean_jsx_text_no_alloc(text: &str) -> (usize, Cow<str>) {
    let mut count = 0;
    for chr in text.chars() {
        if chr.is_whitespace() {
            count += 1;
        }
    }
    (count, Cow::Borrowed(text))
}
```

```count``` is here just to avoid the optimizer throwing everything away.
That gives us:

```
fake_clean_jsx_text_no_alloc
  Instructions:                 208
  L1 Accesses:                  235
  L2 Accesses:                    3
  RAM Accesses:                   9
  Estimated Cycles:             565

fake_clean_jsx_nok_no_alloc
  Instructions:                 447
  L1 Accesses:                  494
  L2 Accesses:                    3
  RAM Accesses:                   9
  Estimated Cycles:             824
```

My implementation is very close when we do not allocate.

#### Code

```rust
fn clean_jsx_text(text: &str) -> std::borrow::Cow<str> {
    let mut chars = text.chars();
    let mut pos = 0;

    let mut last_char = chars.next();

    // we allow one single ' ' at the beginning
    // all others whitespaces are not ok
    let (compliant, mut is_space_not_accepted) = match &last_char {
        Some(' ') => (true, true),
        Some(chr) if chr.is_whitespace() => (false, true),
        Some(_) => (true, false),
        None => return std::borrow::Cow::Borrowed(text),
    };

    if compliant {
        while let Some(current) = chars.next() {
            last_char = Some(current);

            let is_current_whitespace = current.is_whitespace();
            if is_space_not_accepted && is_current_whitespace {
                break;
            } else {
                pos += current.len_utf8();
                is_space_not_accepted = is_current_whitespace;
            }
        }
    }

    // text seems to be compliant...
    if pos >= text.len() - 1 {
        // ... but if we end up in a ..
        match last_char {
            // ...in a' ', accept it.
            Some(' ') => {}
            // ... in a '\t', normalize it to space.
            Some('\t') => {
                let mut new_text = String::with_capacity(text.len());
                new_text.push_str(&text[0..pos]);
                new_text.push(' ');
                return std::borrow::Cow::Owned(new_text);
            }
            // ... in any other whitespace, remove it.
            Some(chr) if chr.is_whitespace() => {
                let mut new_text = String::with_capacity(text.len());
                new_text.push_str(&text[0..pos]);
                return std::borrow::Cow::Owned(new_text);
            }
            _ => {}
        };

        return std::borrow::Cow::Borrowed(text);
    }

    // everything after "pos" needs to be checked and normalized
    let mut new_text = String::with_capacity(text.len());
    new_text.push_str(&text[0..pos]);

    // if we are here, it is because we saw two spaces,
    // which mean last_char is whitespace
    debug_assert!(matches!(last_char, Some(x) if x.is_whitespace()));

    let last_char = last_char.unwrap();
    let mut last_char_is_whitespace = last_char.is_whitespace();

    let mut add_space_before_next_letter = false;

    if new_text.len() == 0 && last_char == '\t' {
        last_char_is_whitespace = false;
        add_space_before_next_letter = true;
    }

    // we need to be aware of newlines now, because
    // when they are present they change what we do
    // with spaces at the end
    let mut has_newline = false;

    // we postpone pushing the space until we see a non whitespace
    // and we never push a space to an empty string.

    while let Some(current) = chars.next() {
        let is_current_whitespace = current.is_whitespace();
        if is_current_whitespace {
            has_newline |= current == '\n';
            add_space_before_next_letter = last_char_is_whitespace;
            continue;
        }

        if add_space_before_next_letter {
            new_text.push(' ');
        }
        new_text.push(current);

        has_newline = false;
        add_space_before_next_letter = false;
        last_char_is_whitespace = false;
    }

    // if whitespaces at the end have a newline, we trim them all
    // if not, we accept ' ' or tab at the end
    // normalizing them to ' '.
    if !has_newline && add_space_before_next_letter {
        new_text.push(' ');
    }

    return std::borrow::Cow::Owned(new_text);
}
````

NicholasLYang:
I'm open to closing this PR if you want to open one with your own solution! I'd reiterate that I think we're spending a lot more time on this than is necessarily warranted. Have we demonstrated that this code is a significant bottleneck?

NicholasLYang:
If I'm reading this code correctly though, it doesn't appear to handle the case where we have a single tab or newline in between words: [Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAJgSwG4D4BmEEAOjAEYCGATqgPRZ4DcxULGOBRLlN9HIAGhAQADjEzQAzslDUqEAO4AFagmkoKAGwUUAntKFkqFMAGs4MAMoUAtnAAymKHGT4tkuIeNmLlkSacAc2QYKgBXTxA4GzI4dHQ4+wooQLCKQLgAMQgqGwoYcRTkEAowmAhBEAALGBtNAHUqzHhJfzA4SzVmnGbdYrBJAxAnDyoYJWNAvNd3SIArSQAPSyDNOABFMIh4Gc0PIX8qUeLKWM1KkSonGHrMdBgq5AAOAAYD+Q964xFiy7hR7AuIQARy28AmonUJUkAFpnHE4pUqHBQZhkRN0tMkG49pEPDZMCFwnjVhswS5sbMhDAKGRbvdHkgAEzU4yYTRBADCEBsWKikgArJUwh4ACq09Q4-YgbARACSUASsEsYCuYgAgorLDBdGtdh4AL4GoA).

MichaReiser:
> I'm open to closing this PR if you want to open one with your own solution! I'd reiterate that I think we're spending a lot more time on this than is necessarily warranted. Have we demonstrated that this code is a significant bottleneck?

We don't have any JSX benchmarks making it difficult. The main reason why I believe it's worth spending a little more time on this is because it introduces a new pattern on how we clean text. 


> With the iterator, that seems like it'd be a little difficult with a string slice, as the iterator would need to have the same lifetime as the slice. I could see a solution where we do a pass to see if there is any redundant whitespace and if we encounter some, we immediately bail and switch into new string mode.

That shouldn't be an issue because the slice should outlive the iterator. The way I'm imagining this is that 
* the iterator returns the cleaned text char by char as it should be. 
* The `format_jsx_text` function iterates over the slice and cleaned-iterator at the same time (should be OK, both read-only) character by character and checks if they're identical
   * if they are, continue to the next character
   * if at the end, return the token as is (matches the "cleaned" version)
   * if not, allocate a new `String`, copy the sofar checked characters (excluding the current character) from the slice into the string, write the current character, and then write the remainder of the iterator.

MichaReiser (approved):
I think the implementation is good but probably a bit longer than it needs to be but don't mind moving forward with it. 

It's probably worth thinking about how to break JSX text across multiple lines when it exceeds line width. I assume we'll need to break the text by every word and use `fill` to print as many words as possible on a single line. 

You can find an iterator based implementation [here](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=d5ab5dca26b08ce2b15465cd62416523)

PR #2528, opened 2022-05-03
fix(playground): Update pnpm requirement to 7.0.0.

cpojer:
## Summary
pnpm 7 is out, let's use it.


MichaReiser:
Looks good. Can you check if this requires any change to our github workflows. I remember that we had to explicitly specify the pnpm version. 

See [documentation](https://github.com/pnpm/action-setup#version)

cpojer:
Good call! Updated the PR.

MichaReiser (approved):
Thx!

PR #2530, opened 2022-05-03
 feat(rome_formatter): multi group and conditional

### LINKED ISSUE #2529: Conditional groups / Alternatives
Alternatives are a way to express different IRs, and the printer decides which ones best fits the situation.

Requirements:
- at least two alternatives must be passed;
- each alternative should be print flat and bail if it can't fit; if all alternatives don't fit in one line, print the last one in flat mode;
- alternatives should be placed from the one line version to the most multi line version, e.g.:

	  ```
	  [
		  (something.in.one.line),
		  (something.in.\n.one.line),
		  (something\n.in\n.one\n.line)
	  ]
	  ```

- the logic of fitting an alternative should be different from the normal logic. At the moment current flatting logic of the printer stops when it finds a hard line break. If so, it rolls back to it's previous state and it print everything in multi line mode. For alternatives, should go in a different way and flat them progressively until we find an hardline break. This is only a concept and it should be verified during the implementation if that's what we need
- during the flatting mode, the printer should choose the first alternative as element to print

ematipico:
## Summary

This PR is going to the first part of many.

This PR closes #2529 

Conditional groups are different from utilities like `if_breaks`. Conditional groups hold different states where the printer tries to print all of them in flat mode, from the first to the last one. If none fits, the last one is chosen and printed in multi line mode.

Conditional group will solve various issue, such as member chains and call expression arguments.

The functionality hasn't been used yet because it will break various things inside the member chain call, causing lot of things not match prettier, especially because the member chain lacks many other features and logic before making everything right.

The plan is also to use just `Group` in the long run. It makes sense having `hard_group_elements`, `group_elements`, etc. but they all generate the same data structure `Group` with different `mode`.

## Test Plan

Created new tests inside the printer

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -4,7 +4,7 @@ use rome_rowan::{
     Language, SyntaxNode, SyntaxToken, SyntaxTokenText, SyntaxTriviaPieceComments, TextLen,
 };
 use std::borrow::Cow;
-use std::fmt::{self, Debug, Formatter};
+use std::fmt::{self, Debug, Display, Formatter};
 use std::ops::Deref;
 
 type Content = Box<FormatElement>;
@@ -613,6 +613,73 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
     format_elements![leading, Group::new(content), trailing]
 }
 
+/// A conditional group instructs the printer to try to print all the elements passed into flat mode.
+/// If none of the elements can't be printed in flat mode, then the last group is printed in multiline
+/// mode.
+/// This mean that also the last element is printed in flat mode.
+///
+/// This is different from conditional content, where some content is printed if already inside a group.
+///
+/// ## Examples
+///
+/// The first element can be printed one single line
+///
+/// ```
+/// use rome_formatter::{conditional_group, Formatted, format_elements, space_token, token, soft_line_break_or_space, FormatOptions, soft_block_indent, group_elements};
+///
+/// let elements = conditional_group(vec![
+///     format_elements![token("summer"), token(","), space_token(), token("spring")],
+///     format_elements![
+///         group_elements(
+///             format_elements![
+///                 token("summer"), token(","), space_token(), token("spring")
+///             ]
+///         )
+///     ]
+/// ]);
+///
+/// assert_eq!("summer, spring", Formatted::new(elements, FormatOptions::default()).print().as_code());
+/// ```
+/// The first element can be printed one single line, so the last one is used
+///
+/// ```
+/// use rome_formatter::{conditional_group, Formatted, space_token, LineWidth, format_elements, token, soft_line_break_or_space, FormatOptions, soft_block_indent, group_elements};
+///
+/// let elements = conditional_group(vec![
+///     format_elements![token("summer"), token(","), space_token(), token("spring")],
+///     format_elements![
+///         group_elements(
+///             format_elements![
+///                 token("summer"), token(","), soft_line_break_or_space(), token("spring")
+///             ]
+///         )
+///     ]
+/// ]);
+///
+/// let options = FormatOptions {
+///   line_width: LineWidth::try_from(10).unwrap(),
+///   ..FormatOptions::default()
+/// };
+///
+/// assert_eq!("summer,\nspring", Formatted::new(elements, options).print().as_code());
+/// ```
+///
+#[inline]
+pub fn conditional_group<Tries>(elements: Tries) -> FormatElement
+where
+    Tries: IntoIterator<Item = FormatElement>,
+{
+    let elements: Vec<_> = elements
+        .into_iter()
+        .map(|element| {
+            let (leading, content, trailing) = element.split_trivia();
+            format_elements![leading, Group::new(content), trailing]
+        })
+        .collect();
+
+    Group::new_conditional(elements).into()
+}
+
 /// Creates a group that forces all elements inside it to be printed on a
 /// single line. This behavior can in turn be escaped by introducing an inner
 /// `Group` element that will resume the normal breaking behavior of the printer.
@@ -634,11 +701,11 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
 ///   FormatOptions, empty_line, if_group_breaks, if_group_fits_on_single_line
 /// };
 ///
-/// let elements = group_elements(hard_group_elements(format_elements![
+/// let elements = hard_group_elements(format_elements![
 ///   if_group_breaks(token("not printed")),
 ///   empty_line(),
 ///   if_group_fits_on_single_line(token("printed")),
-/// ]));
+/// ]);
 ///
 /// assert_eq!("\nprinted", Formatted::new(elements, FormatOptions::default()).print().as_code());
 /// ```
@@ -650,10 +717,8 @@ pub fn hard_group_elements<T: Into<FormatElement>>(content: T) -> FormatElement 
         content
     } else {
         let (leading, content, trailing) = content.split_trivia();
-        format_elements![
-            leading,
-            FormatElement::HardGroup(Group::new(format_elements![content, trailing])),
-        ]
+        let group = Group::new(format_elements![content, trailing]).with_never_break();
+        format_elements![leading, group]
     }
 }
 
@@ -716,7 +781,7 @@ pub fn if_group_breaks<T: Into<FormatElement>>(content: T) -> FormatElement {
     if content.is_empty() {
         content
     } else {
-        FormatElement::from(ConditionalGroupContent::new(
+        FormatElement::from(ConditionalContentInGroup::new(
             content,
             GroupPrintMode::Multiline,
         ))
@@ -782,7 +847,7 @@ where
     if flat_content.is_empty() {
         flat_content
     } else {
-        FormatElement::from(ConditionalGroupContent::new(
+        FormatElement::from(ConditionalContentInGroup::new(
             flat_content,
             GroupPrintMode::Flat,
         ))
@@ -880,12 +945,9 @@ pub enum FormatElement {
     /// See [crate::group_elements] for documentation and examples.
     Group(Group),
 
-    /// See [crate::hard_group_elements] for documentation and examples.
-    HardGroup(Group),
-
     /// Allows to specify content that gets printed depending on whatever the enclosing group
     /// is printed on a single line or multiple lines. See [crate::if_group_breaks] for examples.
-    ConditionalGroupContent(ConditionalGroupContent),
+    ConditionalContentInGroup(ConditionalContentInGroup),
 
     /// Concatenates multiple elements together. See [concat_elements] and [join_elements] for examples.
     List(List),
@@ -963,14 +1025,16 @@ impl Debug for FormatElement {
             FormatElement::Line(content) => content.fmt(fmt),
             FormatElement::Indent(content) => content.fmt(fmt),
             FormatElement::Group(content) => {
-                write!(fmt, "Group")?;
-                content.fmt(fmt)
+                if let Some(expanded) = &content.expanded {
+                    write!(fmt, "Group [Conditional]")?;
+                    expanded.fmt(fmt)
+                } else {
+                    write!(fmt, "Group [{}] ", content.mode)?;
+                    content.fmt(fmt)
+                }
             }
-            FormatElement::HardGroup(content) => {
-                write!(fmt, "HardGroup")?;
-                content.fmt(fmt)
-            }
-            FormatElement::ConditionalGroupContent(content) => content.fmt(fmt),
+
+            FormatElement::ConditionalContentInGroup(content) => content.fmt(fmt),
             FormatElement::List(content) => {
                 write!(fmt, "List ")?;
                 content.fmt(fmt)
@@ -1068,6 +1132,22 @@ impl Deref for List {
     }
 }
 
+/// When creating a group, we instruct it with a mode, which tells
+/// the printer how it should be printed.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) enum GroupMode {
+    /// An hard group, it should never break
+    NeverBreak,
+    /// The group can break
+    MaybeBreak,
+}
+
+impl Display for GroupMode {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        write!(fmt, "{self:?}")
+    }
+}
+
 /// Group is a special token that controls how the child tokens are printed.
 ///
 /// The printer first tries to print all tokens in the group onto a single line (ignoring soft line wraps)
@@ -1075,6 +1155,17 @@ impl Deref for List {
 #[derive(Clone, PartialEq, Eq)]
 pub struct Group {
     pub(crate) content: Content,
+    // pub(crate) content: Vec<FormatElement>,
+    pub(crate) mode: GroupMode,
+
+    /// A conditional group is a rare state where we define a series of states of the same element.
+    ///
+    /// The states from the most flat to the least flat. Essentially, the first one is the state that
+    /// should fit the whole line, while the last one should be the one that fits multiple lines.
+    ///
+    /// When the printer will decided which element to use, it would try to fits them one  by one.
+    /// If none of them fits, it will use the last one, printed using [GroupPrintMode::Multiline] moode
+    pub(crate) expanded: Option<Vec<FormatElement>>,
 }
 
 impl Debug for Group {
@@ -1084,9 +1175,27 @@ impl Debug for Group {
 }
 
 impl Group {
-    pub fn new(content: FormatElement) -> Self {
+    pub fn new(element: FormatElement) -> Self {
         Self {
-            content: Box::new(content),
+            content: Box::new(element),
+            mode: GroupMode::MaybeBreak,
+            expanded: None,
+        }
+    }
+
+    pub fn new_conditional(content: Vec<FormatElement>) -> Self {
+        debug_assert!(content.len() >= 2);
+        Self {
+            content: Box::new(content.get(0).unwrap().clone()),
+            mode: GroupMode::MaybeBreak,
+            expanded: Some(content),
+        }
+    }
+
+    pub fn with_never_break(self) -> Self {
+        Self {
+            mode: GroupMode::NeverBreak,
+            ..self
         }
     }
 }
@@ -1098,7 +1207,7 @@ pub enum GroupPrintMode {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub struct ConditionalGroupContent {
+pub struct ConditionalContentInGroup {
     pub(crate) content: Content,
 
     /// In what mode the content should be printed.
@@ -1107,7 +1216,7 @@ pub struct ConditionalGroupContent {
     pub(crate) mode: GroupPrintMode,
 }
 
-impl ConditionalGroupContent {
+impl ConditionalContentInGroup {
     pub fn new(content: FormatElement, mode: GroupPrintMode) -> Self {
         Self {
             content: Box::new(content),
@@ -1320,10 +1429,12 @@ impl FormatElement {
             FormatElement::Space => false,
             FormatElement::Line(line) => matches!(line.mode, LineMode::Hard | LineMode::Empty),
             FormatElement::Indent(indent) => indent.content.has_hard_line_breaks(),
-            FormatElement::Group(group) | FormatElement::HardGroup(group) => {
-                group.content.has_hard_line_breaks()
-            }
-            FormatElement::ConditionalGroupContent(group) => group.content.has_hard_line_breaks(),
+            FormatElement::Group(Group { mode, content, .. }) => match mode {
+                GroupMode::NeverBreak => content.has_hard_line_breaks(),
+                // TODO: review when there's an expanded state
+                GroupMode::MaybeBreak => content.has_hard_line_breaks(),
+            },
+            FormatElement::ConditionalContentInGroup(group) => group.content.has_hard_line_breaks(),
             FormatElement::List(list) | FormatElement::Fill(list) => {
                 list.content.iter().any(FormatElement::has_hard_line_breaks)
             }
@@ -1376,16 +1487,20 @@ impl FormatElement {
                     (FormatElement::List(list), empty_element(), empty_element())
                 }
             }
-            FormatElement::HardGroup(group) => {
-                let (leading, content, trailing) = group.content.split_trivia();
-                // re-create the grouping around the content only
-                (leading, hard_group_elements(content), trailing)
-            }
 
-            FormatElement::Group(group) => {
-                let (leading, content, trailing) = group.content.split_trivia();
-                // re-create the grouping around the content only
-                (leading, group_elements(content), trailing)
+            FormatElement::Group(Group {
+                content, expanded, ..
+            }) => {
+                if expanded.is_none() {
+                    let (leading, content, trailing) = content.split_trivia();
+                    // re-create the grouping around the content only
+                    (leading, group_elements(content), trailing)
+                } else {
+                    // TODO: implement the pick of trivias on conditional groups
+                    let (_leading, content, _trailing) = content.split_trivia();
+
+                    (empty_element(), group_elements(content), empty_element())
+                }
             }
             // Non-list elements are returned directly
             _ => (empty_element(), self, empty_element()),
@@ -1404,9 +1519,7 @@ impl FormatElement {
             FormatElement::Empty | FormatElement::Line(_) | FormatElement::Comment(_) => None,
 
             FormatElement::Indent(indent) => indent.content.last_element(),
-            FormatElement::Group(group) | FormatElement::HardGroup(group) => {
-                group.content.last_element()
-            }
+            FormatElement::Group(group) => group.content.last_element(),
 
             _ => Some(self),
         }
@@ -1431,9 +1544,9 @@ impl From<List> for FormatElement {
     }
 }
 
-impl From<ConditionalGroupContent> for FormatElement {
-    fn from(token: ConditionalGroupContent) -> Self {
-        FormatElement::ConditionalGroupContent(token)
+impl From<ConditionalContentInGroup> for FormatElement {
+    fn from(token: ConditionalContentInGroup) -> Self {
+        FormatElement::ConditionalContentInGroup(token)
     }
 }
 
@@ -1459,7 +1572,7 @@ impl From<Option<FormatElement>> for FormatElement {
 mod tests {
 
     use crate::format_element::{
-        empty_element, join_elements, normalize_newlines, ConditionalGroupContent, List,
+        empty_element, join_elements, normalize_newlines, ConditionalContentInGroup, List,
         VerbatimKind, LINE_TERMINATORS,
     };
     use crate::{concat_elements, space_token, token, FormatElement, TextRange, Token, Verbatim};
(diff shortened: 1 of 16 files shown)


crates/rome_formatter/src/format_element.rs
```diff
@@ -613,6 +613,73 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
...
+/// let options = FormatOptions {
+///   line_width: LineWidth::try_from(10).unwrap(),
+///   ..FormatOptions::default()
+/// };
+///
+/// assert_eq!("summer,\nspring", Formatted::new(elements, options).print().as_code());
+/// ```
+///
+#[inline]
+pub fn conditional_group<Tries>(elements: Tries) -> FormatElement
```
IWANABETHATGUY, line 668: Would it be better to use other generic types than `Tries`, because  Trie is a data structure in most case

crates/rome_formatter/src/format_element.rs
```diff
@@ -1068,13 +1132,40 @@ impl Deref for List {
...
+    pub(crate) mode: GroupMode,
+
+    /// A conditional group is a rare state where we define a series of states of the same element.
+    ///
+    /// The states from the most flat to the least flat. Essentially, the first one is the state that
+    /// should fit the whole line, while the last one should be the one that fits multiple lines.
+    ///
+    /// When the printer will decided which element to use, it would try to fits them one  by one.
+    /// If none of them fits, it will use the last one, printed using [GroupPrintMode::Multiline] moode
+    pub(crate) expanded: Option<Vec<FormatElement>>,
```
MichaReiser, line 1168: Hmm, this will be difficult to represent in a flat IR structure because the group has now two nesting. But I guess it could also use a vec for that case. 

How does `expanded` work in case of `split_trivia` or even with something like `has_line_break`?

Nit: You can use a `Box<[&FormatElement]>` if the `Vec` doesn't need to be mutable. You can then test if the `slice` is empty to see if there are any expanded alternatives.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1068,13 +1132,40 @@ impl Deref for List {
...
+    NeverBreak,
+    /// The group can break
+    MaybeBreak,
+}
+
+impl Display for GroupMode {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        write!(fmt, "{self:?}")
+    }
+}
```
MichaReiser, line 1149: What do we need a `Display` implementation for. This seems to be the only `FormatElement` implementing it.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1068,13 +1132,40 @@ impl Deref for List {
...
+
 /// Group is a special token that controls how the child tokens are printed.
 ///
 /// The printer first tries to print all tokens in the group onto a single line (ignoring soft line wraps)
 /// but breaks the array cross multiple lines if it would exceed the specified `line_width`, if a child token is a hard line break or if a string contains a line break.
 #[derive(Clone, PartialEq, Eq)]
 pub struct Group {
     pub(crate) content: Content,
+    // pub(crate) content: Vec<FormatElement>,
```
MichaReiser, line 1158: Delete?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1068,13 +1132,40 @@ impl Deref for List {
     }
 }
 
+/// When creating a group, we instruct it with a mode, which tells
+/// the printer how it should be printed.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) enum GroupMode {
+    /// An hard group, it should never break
+    NeverBreak,
```
MichaReiser, line 1140: What's the difference to `FormatElement::HardGroup`?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1068,13 +1132,40 @@ impl Deref for List {
...
 /// but breaks the array cross multiple lines if it would exceed the specified `line_width`, if a child token is a hard line break or if a string contains a line break.
 #[derive(Clone, PartialEq, Eq)]
 pub struct Group {
     pub(crate) content: Content,
+    // pub(crate) content: Vec<FormatElement>,
+    pub(crate) mode: GroupMode,
+
+    /// A conditional group is a rare state where we define a series of states of the same element.
+    ///
+    /// The states from the most flat to the least flat. Essentially, the first one is the state that
```
MichaReiser, line 1163: Are there any invariants that apply to the `Group`? For example, is it possible that a Group has `mode: NeverBreak` and `expanded` content?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1084,9 +1175,27 @@ impl Debug for Group {
...
         Self {
-            content: Box::new(content),
+            content: Box::new(element),
+            mode: GroupMode::MaybeBreak,
+            expanded: None,
+        }
+    }
+
+    pub fn new_conditional(content: Vec<FormatElement>) -> Self {
+        debug_assert!(content.len() >= 2);
```
MichaReiser, line 1187: Nit: You could do `new_conditional(default: FormatElement, alternative: Vec<FormatElement>)` to ensure you always have at least one option. This also avoids the cloning on line `1189`.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1084,9 +1175,27 @@ impl Debug for Group {
...
+        Self {
+            content: Box::new(content.get(0).unwrap().clone()),
+            mode: GroupMode::MaybeBreak,
+            expanded: Some(content),
```
MichaReiser, line 1191: Isn't it now the case that expanded still holds the first variant (that is also stored in content)? Does it make sense, that a `Group` with `expanded` also has the `content` field?

crates/rome_formatter/src/format_element.rs
MichaReiser: Or use something like `Variants`, `Alternatives`.

MichaReiser: Should we rename the method to `group_with_alternatives` because it renders a group with different content. It's not that the group itself is conditional.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1320,10 +1429,12 @@ impl FormatElement {
             FormatElement::Space => false,
             FormatElement::Line(line) => matches!(line.mode, LineMode::Hard | LineMode::Empty),
             FormatElement::Indent(indent) => indent.content.has_hard_line_breaks(),
-            FormatElement::Group(group) | FormatElement::HardGroup(group) => {
-                group.content.has_hard_line_breaks()
-            }
-            FormatElement::ConditionalGroupContent(group) => group.content.has_hard_line_breaks(),
+            FormatElement::Group(Group { mode, content, .. }) => match mode {
+                GroupMode::NeverBreak => content.has_hard_line_breaks(),
+                // TODO: review when there's an expanded state
```
MichaReiser, line 1434: TODO

crates/rome_formatter/src/format_element.rs
```diff
@@ -1376,16 +1487,20 @@ impl FormatElement {
...
-                (leading, group_elements(content), trailing)
+            FormatElement::Group(Group {
+                content, expanded, ..
+            }) => {
+                if expanded.is_none() {
+                    let (leading, content, trailing) = content.split_trivia();
+                    // re-create the grouping around the content only
+                    (leading, group_elements(content), trailing)
+                } else {
+                    // TODO: implement the pick of trivias on conditional groups
```
MichaReiser, line 1499: TODO. 

It's unclear to me what the expected behaviour should be here. You can't split the trivia from all variants, but not splitting it from a variant also seems incorrect.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1404,9 +1519,7 @@ impl FormatElement {
             FormatElement::Empty | FormatElement::Line(_) | FormatElement::Comment(_) => None,
 
             FormatElement::Indent(indent) => indent.content.last_element(),
-            FormatElement::Group(group) | FormatElement::HardGroup(group) => {
-                group.content.last_element()
-            }
+            FormatElement::Group(group) => group.content.last_element(),
```
MichaReiser, line 1522: Same here. It's unclear to me what should happen if a `Group` has variants because the result now differs depending on which variant is rendered.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1585,12 +1698,12 @@ mod tests {
...
         assert_eq!(24, std::mem::size_of::<Token>());
-        assert_eq!(16, std::mem::size_of::<ConditionalGroupContent>());
+        assert_eq!(16, std::mem::size_of::<ConditionalContentInGroup>());
         assert_eq!(24, std::mem::size_of::<List>());
         // Increasing the size of FormatElement has serious consequences on runtime performance and memory footprint.
         // Is there a more efficient way to encode the data to avoid increasing its size? Can the information
         // be recomputed at a later point in time?
         // You reduced the size of a format element? Excellent work!
-        assert_eq!(32, std::mem::size_of::<FormatElement>());
+        assert_eq!(48, std::mem::size_of::<FormatElement>());
```
MichaReiser, line 1707: This will result in a significant performance penalty. We should revisit our design to make sure we can avoid the size increase.

crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
```diff
@@ -89,6 +89,13 @@ impl FlattenItem {
         }
     }
 
+    pub(crate) fn should_insert_empty_line_after(&self) -> bool {
+        // TODO: prettier does a text based search to figure out the first newline?
```
MichaReiser, line 93: Can you link an issue for this todo

crates/rome_js_formatter/src/utils/member_chain/groups.rs
```diff
@@ -124,29 +125,37 @@ impl<'f> Groups<'f> {
...
     /// and the other one that goes on multiple lines.
     ///
     /// It's up to the printer to decide which one to use.
-    pub fn into_format_elements(self) -> (FormatElement, FormatElement) {
-        let formatted_groups = self.into_formatted_groups();
-        (
-            concat_elements(formatted_groups.clone()),
-            join_elements(soft_line_break(), formatted_groups),
-        )
+    #[allow(dead_code)]
```
MichaReiser, line 153: Delete?

crates/rome_js_formatter/src/utils/member_chain/groups.rs
```diff
@@ -316,26 +325,44 @@ impl<'f> Groups<'f> {
 #[derive(Debug)]
 pub(crate) struct HeadGroup {
     items: Vec<FlattenItem>,
+    was_expanded: bool,
```
MichaReiser, line 328: Documentation? Rename to `is_expanded`?

crates/rome_js_formatter/src/utils/member_chain/groups.rs
```diff
@@ -316,26 +325,44 @@ impl<'f> Groups<'f> {
...
+        self.was_expanded = true;
     }
 
-    fn has_comments(&self) -> bool {
+    pub(crate) fn has_comments(&self) -> bool {
         self.items.iter().any(|item| item.has_trailing_comments())
     }
+
+    // TODO: figure out what's this logic about
+    #[allow(dead_code)]
```
MichaReiser, line 357: Delete/TODO/Document?

crates/rome_js_formatter/src/utils/member_chain/mod.rs
```diff
@@ -295,20 +292,18 @@ fn format_groups(
...
-        let head_formatted = head_group.into_format_element();
-        let (one_line, _) = groups.into_format_elements();
+        let head_formatted = head_group.as_format_element();
+        let body = groups.one_line_element();
 
         // TODO: this is not the definitive solution, as there are few restrictions due to how the printer works:
         // - groups that contains other groups with hard lines break all the groups
         // - conditionally print one single line is subject to how the printer works (by default, multiline)
-        Ok(format_elements![head_formatted, one_line])
+        // let body = groups.into_format_elements(&head_group);
```
MichaReiser, line 305: Delete?

crates/rome_js_formatter/tests/specs/prettier/js/method-chain/inline_merge.js.snap
```diff
@@ -37,9 +37,9 @@ Object.keys(availableLocales({ test: true }))
 
 this.layoutPartsToHide =
   this.utils.hashset(
-    _.flatMap(this.visibilityHandlers, (fn) => fn())
-      .concat(this.record.resolved_legacy_visrules)
-      .filter(Boolean),
+    _.flatMap(this.visibilityHandlers, (fn) => fn()).concat(
```
MichaReiser, line 40: Can we add any new tests to our spec that verify that the formatting now works as expected?

crates/rome_formatter/src/format_element.rs
```diff
@@ -613,6 +613,73 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
     format_elements![leading, Group::new(content), trailing]
 }
 
+/// A conditional group instructs the printer to try to print all the elements passed into flat mode.
+/// If none of the elements can't be printed in flat mode, then the last group is printed in multiline
+/// mode.
+/// This mean that also the last element is printed in flat mode.
```
MichaReiser, line 619: So that means that the last element acts as the fallback.

crates/rome_formatter/src/format_element.rs
```diff
@@ -613,6 +613,73 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
...
+///             format_elements![
+///                 token("summer"), token(","), space_token(), token("spring")
+///             ]
+///         )
+///     ]
+/// ]);
+///
+/// assert_eq!("summer, spring", Formatted::new(elements, FormatOptions::default()).print().as_code());
+/// ```
+/// The first element can be printed one single line, so the last one is used
```
MichaReiser, line 643:
```suggestion
/// The first element exceeds the line width, so the last one is used
```

crates/rome_formatter/src/format_element.rs
```diff
@@ -613,6 +613,73 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
...
+///         )
+///     ]
+/// ]);
+///
+/// let options = FormatOptions {
+///   line_width: LineWidth::try_from(10).unwrap(),
+///   ..FormatOptions::default()
+/// };
+///
+/// assert_eq!("summer,\nspring", Formatted::new(elements, options).print().as_code());
```
MichaReiser, line 664: How does it work with more than two variants? Is this for cases where the formatter removes certain tokens if it doesn't fit on a single line? Because I don't understand how it's possible that the first content doesn't fit on a single line, but then a second content does but uses the exact same tokens.

crates/rome_formatter/src/format_element.rs
ematipico: > Nit: You can use a Box<[&FormatElement]> if the Vec doesn't need to be mutable. You can then test if the slice is empty to see if there are any expanded alternatives

This is a temporary solution, as I didn't want to break too many things. My plan is the following:
- having one single `Group` struct 
- replace `Box<FormatElement>` with `Vec<FormatElmeent>` (or `Box<[&FormatElement]>` as you suggested), and remove the `expanded` state.

As this changes have great repercussions, I wanted to first apply the first step and make sure it does what I want.

crates/rome_formatter/src/format_element.rs
ematipico: **It is** the hard group, the plan is to use just `FormatElement::Group`

crates/rome_js_formatter/src/utils/member_chain/mod.rs
ematipico: It's part of the `TODO`

crates/rome_formatter/src/format_element.rs
ematipico: @MichaReiser  Would you prefer "alternatives" or use the prettier internal terminology?

MichaReiser:
How does the output differ between these two IRs?

```
conditional_groups(vec![opt1, opt2])

group(
  if_group_fits_on_single_line(opt1),
  if_group_breaks(group(
    if_group_fits_on_single_line(opt2),
    if_group_breaks(opt2)
  )),
)
```

crates/rome_formatter/src/format_element.rs
ematipico: Theoretically yes, if someones uses the primitives directly, I will add some debug assertion to prevent the misuse of these case.

crates/rome_formatter/src/format_element.rs
ematipico: No it doesn't, but it was needed for the introduction of the concept. This will be change afterwards

crates/rome_formatter/src/format_element.rs
ematipico: This will contain the same logic of `List`

crates/rome_formatter/src/format_element.rs
ematipico: What do you need here?

crates/rome_formatter/src/format_element.rs
ematipico: Yes, this will change with the next iteration

crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
ematipico: There's no issue here, I just need to understand what they are doing and why: https://github.com/prettier/prettier/blob/a9de2a128cc8eea84ddd90efdc210378a894ab6b/src/language-js/print/member-chain.js#L71-L90

crates/rome_js_formatter/src/utils/member_chain/groups.rs
ematipico: No, it's here on purpose

crates/rome_js_formatter/src/utils/member_chain/groups.rs
ematipico: It's left on purpose and there's no issue. It's always part of the infrastructure work. What do you need me to put here?

crates/rome_formatter/src/format_element.rs
ematipico: Yes

crates/rome_formatter/src/format_element.rs
ematipico: Oh, it's for cases like these: https://prettier.io/playground/#N4Igxg9gdgLgprEAuEB3AllAJhVA6ABwgGcYBZOY4gQwHM4AKYAHSgAI3JY4APGJNungBbPF3h8ANKw4wIBdGAFC4ouQrDSoAXzbVibAJIBBMDHTQKVOnACUAbhCSQ889GLJQ1AE7fcABR8EDxRqABtUagBPD2cAI29qMABrOBgAZWphOAAZTDhkADNw4jh4xJS09IIkzFpkGG8AVzKQVTi4LCxOnOooWiabADEIb2FqGHN+5BBqJrknEAALGGEwgHUllWIasDh04KF0ADchKJmwKkXMUu8Yf0TacaKS1oArYh50urC4AEUmhB4C8wqVnDVvLcZjAogRKGBvOgCDBFgREbB1ugsDAlsgABwABnBflK60SBBmaMocG8xwKzgAjoD4A95CFZsQALRQOCdTqLbxwJnoQUPOjPJDFUGtUrCdANZoyn7-ZkFSWvZwwahxTHY3FIABMmsS6DCdQAwhBhBK2sQAKyLJqlAAq2pCjRazmOLUM2AQGQRSJgxmw6RhvxBpW02iAA

It's not about the tokens, but where the grouping occurs between the tokens

crates/rome_js_formatter/tests/specs/prettier/js/method-chain/inline_merge.js.snap
ematipico: This doesn't work as expected because there are still many things missing (and it was like this already when I implemented many of them last week: https://github.com/rome/tools/pull/2490), and conditional groups is just one of them.

I scattered various `TODO` of the things needed to implement.

crates/rome_formatter/src/format_element.rs
ematipico: But the `alternative` could be empty, which is not what we want. It would be a misuse of the API.

crates/rome_formatter/src/format_element.rs
ematipico: It's unclear to me too, I will have to figure it out.

crates/rome_formatter/src/format_element.rs
ematipico: Same for me, that's why this addition allows to not break existing functionality and make me think how the next refactor should look like.

crates/rome_js_formatter/src/utils/member_chain/groups.rs
ematipico: The name seems clear to me, and it's set to `true` inside the `expand_group` function. Is there anything else you need?

crates/rome_formatter/src/format_element.rs
ematipico: I will reword it, but the suggestion is not the only case

ematipico:
> How does the output differ between these two IRs?
> 
> ```
> conditional_groups(vec![opt1, opt2])
> 
> group(
>   if_group_fits_on_single_line(opt1),
>   if_group_breaks(group(
>     if_group_fits_on_single_line(opt2),
>     if_group_breaks(opt2)
>   )),
> )
> ```

It seems you are confused as I was. That's why I renamed the variant to `ConditionalContentInGroup`.

`if_group_breaks` and  acts `if_group_fits_on_single_line` *when it's already inside a group**, which means that if it's not inside a group, the logic doesn't work anymore... and creating a group sometimes it's what we want because having a group also trigger the break logic. Instead `conditional_group` acts regardless if its inside a group or not, which is a different (and new) functionality.

crates/rome_formatter/src/format_element.rs
MichaReiser: I'm still confused about this... and I looked at the example and don't get it, sorry 

The part that confuses me is that it seems to pick the second (I think)? variant for the given example. But the printed code doesn't fit on a single line, the inner groups do break. 

So my impression is that these conditional groups work entirely differently from normal groups in that they don't print the parent on multiple lines if any of the children don't fit. 

That might even be a reason to not use the `Group` IR to represent this concept and instead name it like `Variants`.  Could you please explain in more detail how these groups work and compare them to regular groups.

crates/rome_formatter/src/format_element.rs
MichaReiser: I don't know what prettier uses. I don't think we have to stick to prettier's terminology to closely (we can add a comment in the documentation and mention how the same is called in prettier).

crates/rome_formatter/src/format_element.rs
MichaReiser: The downside of moving this inside is that it increases the size of `FormatElement` (get's aligned to bytes, that leaves 7  bits unused)

crates/rome_formatter/src/format_element.rs
MichaReiser: Would it be possible to model the enum/struct in a way that these invariants are impossible to violate? For example, by making `Group` an enum with a `Hard` and `Conditional`, and `Maybe` invariant?

crates/rome_formatter/src/format_element.rs
MichaReiser: What's the motivation for having a single `Group` struct?

crates/rome_formatter/src/format_element.rs
MichaReiser: You can still add the assertion that `alternative` isn't empty if we indeed need two. A yet other solution to support more than two alternatives could be `Alternative(A, Alternative(B, C))`

crates/rome_formatter/src/format_element.rs
MichaReiser: It's unclear if this TODO needs to be addressed as part of this PR or if you plan to follow up in another PR. It also doesn't provide enough information for me to know what I would need to do in case it doesn't get addressed in this PR.

crates/rome_formatter/src/format_element.rs
MichaReiser: I'm not sure if there's a way to implement this correctly. The problem is that the trivia will be printed twice if `Alternative` is used and the second variant get's printed because the trivia only got removed from the first variant. 

A solution could be to always remove the trivia from all variants and add an assertion that the leading and trailing trivia of all variants must be equal. This variant could already be added when constructing the alternatives.

crates/rome_formatter/src/format_element.rs
MichaReiser: Can you please outline how the next iteration looks like. I want to make sure we aren't increasing the `FormatElement` size except if strictly necessary and if it avoids a lot of work (which I'm not sure it does right now). Because this increase will add roughly a 10% slowdown to the formatter.

crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
MichaReiser: Yes, but what's the issue in which you want to follow up on this TODO? Is it in this PR?

crates/rome_js_formatter/src/utils/member_chain/groups.rs
MichaReiser: And that is?

crates/rome_js_formatter/src/utils/member_chain/groups.rs
MichaReiser: It's unclear to me what the meaning of `expanded` means. What's the implication of it.

crates/rome_formatter/src/format_element.rs
ematipico: Prettier's terminology is "conditionaGroup" :) 

But if "alternatives" help, I can change it

crates/rome_formatter/src/format_element.rs
ematipico: So that the printing logic can be stored and configured in one single primitive: `Group`

crates/rome_formatter/src/format_element.rs
ematipico: I will apply your suggestion in the next iteration, because this code will likely change in the that iteration. Is that OK?

crates/rome_formatter/src/format_element.rs
ematipico: How can I make you understand that this TODO **won't** be part of this PR?

MichaReiser: See other comments:
* Link to an issue
* document it in the PR
* explain how this should work in the future

crates/rome_js_formatter/src/utils/member_chain/groups.rs
MichaReiser: The main issue is that it's unclear for reviewers what you plan to follow up in another PR and what are things you might have missed to address. 

I suggest you to either document all follow up work in the PR summary, track the pending work in an existing or new issue and add the issue number here, or expand the documentation. If this is something left for a later PR, then the documentation should be detailed enough for someone else to understand in case you don't have the time to follow up.

crates/rome_formatter/src/format_element.rs
ematipico: > I'm still confused about this... and I looked at the example and don't get it, sorry
> 
> The part that confuses me is that it seems to pick the second (I think)? variant for the given example. But the printed code doesn't fit on a single line, the inner groups do break.
> 
> So my impression is that these conditional groups work entirely differently from normal groups in that they don't print the parent on multiple lines if any of the children don't fit.
> 
> That might even be a reason to not use the `Group` IR to represent this concept and instead name it like `Variants`. Could you please explain in more detail how these groups work and compare them to regular groups.

Because we don't have all the infrastructure yet. We need to implement groups that always break. If you check the example, you'll se a `{ shouldBreak: true }, which instructs the printer to always break that group, regardless.

This is not part of this PR.

crates/rome_formatter/src/format_element.rs
ematipico: I will implement your suggestion in the next iteration, as it will change regardless

Issue #2533, opened 2022-05-03
Merge grouping in one single primitive


ematipico:

MichaReiser:
> We want to merge all these logic in one single Group primitive and change the logic of the printing inside it.

Can you outline the motivation behind this besides that it is a single primitive. Achievhing this might not be possible without increasing the size of `FormatElement` and any motivation must outweigh that regression.

MichaReiser:
I think we can close this. `HardGroup` is gone and `BestFitting` isn't a group.

ematipico:
Thank you Micha!

PR #2534, opened 2022-05-03
feat(rome_js_formatter): alternatives

### LINKED ISSUE #2529: Conditional groups / Alternatives

ematipico:
## Summary

This PR closes #2529

This feature is not used yet, it will be used to fix futures bugs (which may contain possible fixes to the current implementation)

## Test Plan

I created two new tests

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -613,6 +613,68 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
     format_elements![leading, Group::new(content), trailing]
 }
 
+/// A conditional group instructs the printer to try to print all the elements passed into flat mode.
+/// If none of the elements can't be printed in flat mode, then the last group is printed in multiline
+/// mode.
+/// This mean that also the last element is printed in flat mode.
+///
+/// This is different from conditional content, where some content is printed if already inside a group.
+///
+/// ## Examples
+///
+/// The first element can be printed one single line
+///
+/// ```
+/// use rome_formatter::{alternatives, Formatted, format_elements, space_token, token, soft_line_break_or_space, FormatOptions, soft_block_indent, group_elements};
+///
+/// let elements = alternatives(vec![
+///     format_elements![token("summer"), token(","), space_token(), token("spring")],
+///     format_elements![
+///         group_elements(
+///             format_elements![
+///                 token("summer"), token(","), space_token(), token("spring")
+///             ]
+///         )
+///     ]
+/// ]);
+///
+/// assert_eq!("summer, spring", Formatted::new(elements, FormatOptions::default()).print().as_code());
+/// ```
+/// The first element can't be printed on a single line (there are line breaks, exceeds line
+/// width, there are line suffix, etc.), so the last one is used
+///
+/// ```
+/// use rome_formatter::{alternatives, Formatted, space_token, LineWidth, format_elements, token, soft_line_break_or_space, FormatOptions, soft_block_indent, group_elements};
+///
+/// let elements = alternatives(vec![
+///     format_elements![token("summer"), token(","), space_token(), token("spring")],
+///     format_elements![
+///         group_elements(
+///             format_elements![
+///                 token("summer"), token(","), soft_line_break_or_space(), token("spring")
+///             ]
+///         )
+///     ]
+/// ]);
+///
+/// let options = FormatOptions {
+///   line_width: LineWidth::try_from(10).unwrap(),
+///   ..FormatOptions::default()
+/// };
+///
+/// assert_eq!("summer,\nspring", Formatted::new(elements, options).print().as_code());
+/// ```
+///
+#[inline]
+pub fn alternatives<Tries>(elements: Tries) -> FormatElement
+where
+    Tries: IntoIterator<Item = FormatElement>,
+{
+    let elements: Vec<_> = elements.into_iter().collect();
+
+    FormatElement::Alternatives(elements.into_iter().collect())
+}
+
 /// Creates a group that forces all elements inside it to be printed on a
 /// single line. This behavior can in turn be escaped by introducing an inner
 /// `Group` element that will resume the normal breaking behavior of the printer.
@@ -883,6 +945,9 @@ pub enum FormatElement {
     /// See [crate::hard_group_elements] for documentation and examples.
     HardGroup(Group),
 
+    /// See [crate::alternatives] for documentation and examples.
+    Alternatives(Vec<FormatElement>),
+
     /// Allows to specify content that gets printed depending on whatever the enclosing group
     /// is printed on a single line or multiple lines. See [crate::if_group_breaks] for examples.
     ConditionalGroupContent(ConditionalGroupContent),
@@ -988,6 +1053,9 @@ impl Debug for FormatElement {
                 .debug_tuple("Verbatim")
                 .field(&verbatim.element)
                 .finish(),
+            FormatElement::Alternatives(alternatives) => {
+                fmt.debug_tuple("Alternatives").field(alternatives).finish()
+            }
         }
     }
 }
@@ -1331,6 +1399,8 @@ impl FormatElement {
             FormatElement::LineSuffix(_) => true,
             FormatElement::Comment(content) => content.has_hard_line_breaks(),
             FormatElement::Verbatim(verbatim) => verbatim.element.has_hard_line_breaks(),
+            // it's not possible to be retrieve to hard line breaks from alternatives, we return false
+            FormatElement::Alternatives(_) => false,
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/rome_formatter/src/format_element.rs
```diff
@@ -1312,8 +1378,14 @@ impl FormatElement {
...
-    /// Returns true if this [FormatElement] recursively contains any hard line break
+    /// Returns true if this [FormatElement] will break definitely break,
     /// ([hard_line_break], [empty_line], [Token] containing the '\n' character)
+    ///
+    /// Some primitives can't know ahead of time if they contain primitive that breaks.
+    /// Because of that, they will assume that don't have line breaks.
+    ///
+    /// An example if the [FormatElement::Alternatives] where it holds different versions
+    /// of an element, and only the printer will know which one to use.
     pub fn has_hard_line_breaks(&self) -> bool {
```
leops, line 1389: I think `has_hard_line_breaks` is only used to print member chains, is it still needed in a fully Prettier-compatible implementation of that logic (and is so is this the way Prettier handles it) or could it be removed ?

crates/rome_formatter/src/printer.rs
```diff
@@ -214,6 +227,19 @@ impl<'a> Printer<'a> {
                 }
             }
 
+            FormatElement::Alternatives(alternatives) => {
+                for (index, alternative) in alternatives.iter().enumerate() {
+                    if index + 1 == alternatives.len() {
```
leops, line 232: This bit here doesn't look correct, from my understanding of [the logic in Prettier](https://github.com/prettier/prettier/blob/893f43a599dcaba249f012fecbc28a3dac008f81/src/document/doc-printer.js#L364) it seems it will also try to print the last group in flat mode before trying again in breaking mode if that doesn't fit

crates/rome_formatter/src/printer.rs
```diff
@@ -359,6 +385,24 @@ impl<'a> Printer<'a> {
                 args.with_hard_group(false),
             )),
 
+            FormatElement::Alternatives(alternatives) => {
+                for (index, alternative) in alternatives.iter().enumerate() {
```
leops, line 389: I think in flat mode [Prettier](https://github.com/prettier/prettier/blob/893f43a599dcaba249f012fecbc28a3dac008f81/src/document/doc-printer.js#L327) doesn't try to print all the expanded states successively, only the first one (`doc.contents` is set to the first expanded state [here](https://github.com/prettier/prettier/blob/9dcfd16e6a7a92b00ed5076786ccb2985992f34e/src/document/doc-builders.js#L134))

crates/rome_formatter/src/format_element.rs
```diff
@@ -613,6 +613,66 @@ pub fn group_elements<T: Into<FormatElement>>(content: T) -> FormatElement {
...
+/// let options = FormatOptions {
+///   line_width: LineWidth::try_from(10).unwrap(),
+///   ..FormatOptions::default()
+/// };
+///
+/// assert_eq!("summer,\nspring", Formatted::new(elements, options).print().as_code());
+/// ```
+///
+#[inline]
+pub fn alternatives<Tries>(elements: Tries) -> FormatElement
```
MichaReiser, line 669: The documentation is now outdated since we renamed the IR element. I would change `group` to `variant` but let's wait a second before doing that until we're all happy with the terminology.

What are the main differences of `alternatives` to group. You mentioned something that it also acts as a barrier that it doesn't break the parent if it contains any line breaks. Are there any other differences?

MichaReiser: Name ideas (I'll spend more time tomorrow). `TryFit`

crates/rome_formatter/src/format_element.rs
ematipico: We should keep it for now, it seems it's been used in [other places](https://github.com/prettier/prettier/blob/a9de2a128cc8eea84ddd90efdc210378a894ab6b/src/language-js/print/call-arguments.js#L107-L108). We could remove it later if we think we don't need it there.

I haven't look at that logic yet, though

crates/rome_formatter/src/printer.rs
ematipico: Although `Alternatives` is not a group yet, so we don't have access to that information.

Also, prettier doesn't always pick the first element. If the group *always breaks*, [it tries to fit the last one](https://github.com/prettier/prettier/blob/893f43a599dcaba249f012fecbc28a3dac008f81/src/document/doc-printer.js#L196-L204).

As requested by Micha, he wanted to first understand this feature and maybe later join it with the `Group` primitive. So, as for now, the `try_print_flat` for `Format::Alternatives` can't work unfortunately.

MichaReiser: My goal was to build out the most primitive approach but we should still implement the same behaviour as prettier. That's why I'm trying to understand what the exact behaviour is supposed to be. It would help if we can summarise that somewhere. What does the printing to in flat mode, what does it do in the normal mode.

crates/rome_formatter/src/format_element.rs
ematipico: In theory, `alternatives` **should** act as a group because it needs to [have access to some information like the breaking mode](https://github.com/prettier/prettier/blob/893f43a599dcaba249f012fecbc28a3dac008f81/src/document/doc-printer.js#L196-L204), although I am not sure if this will be true for us...

crates/rome_formatter/src/printer.rs
ematipico: I believe you were referring to the other function?

crates/rome_formatter/src/printer.rs
ematipico: This will be done in the next iteration https://github.com/rome/tools/issues/2533

crates/rome_formatter/src/format_element.rs
```diff
@@ -1331,6 +1404,8 @@ impl FormatElement {
             FormatElement::LineSuffix(_) => true,
             FormatElement::Comment(content) => content.has_hard_line_breaks(),
             FormatElement::Verbatim(verbatim) => verbatim.element.has_hard_line_breaks(),
+            // it's not possible to be retrieve to hard line breaks from alternatives, we return false
+            FormatElement::Alternatives(_) => false,
         }
     }
```
MichaReiser, line 1410: What about `last_element` and `split_trivia`?

MichaReiser (commented):
I'm still failing to understand the exact behaviour of the `Alternatives` IR element. 

Can you please precisely formulate how `Alternatives` works. 

* How is it different/similar to groups?
* How is it different from using a combination of `group(if_group_breaks(content1), if_group_fits_single_line(content2))`?
* How does it behave in flat/normal print mode

crates/rome_formatter/src/format_element.rs
ematipico: `last_element`: no need, it matches the description of the function
`split_trivia`: no need, it matches the behaviour of the function

ematipico:
> I'm still failing to understand the exact behaviour of the `Alternatives` IR element.
> 
> Can you please precisely formulate how `Alternatives` works.
> 
>     * How is it different/similar to groups?

`Group` is just one representation of one state of the IR. We can use `ConditionalGroupContent`, but as the names says, that is used to show/hide content if an element **is already inside a group**.

`Alternatives` allows you to represent different versions of an IR. Essentially, there are cases (like member chains) where we can't use groups for conditional content, and other cases where we want to have more than two representations of the same content.

> 
>     * How is it different from using a combination of `group(if_group_breaks(content1), if_group_fits_single_line(content2))`?

These IRs requires having groups (or being inside a group). There are cases where this can't happen (member chains). These IR weren't the tool I needed because they require a group 

> 
>     * How does it behave in flat/normal print mode

The idea behind `Alternatives` is that in the first place you put the alternative that takes the most space horizontally, essentially the ones that fits one line, or less lines possible (vertically), from another point of view, the one that doesn't break at all.

The last one should be the alternative that occupies the least space, horizontally. The alternative that breaks most.

The printer will try to print each alternative in **flat mode**. All of them, even the last one. If all fail, the last one is printed in normal mode.

The tricky part that is not yet fully implemented, is that also `Alternatives` should have a break mode (like groups), because during the `try_flat_element` phase, the grouping mode is important because it decides which element the print: the first or last alternative.

MichaReiser:
I ported this PR to the new printer that uses the `fits_on_line` function to determine if some content fits #2591. The difference is that this new PR correctly returns `true` for a variant that contains line breaks.

PR #2535, opened 2022-05-03
feat(rome_js_parser): quick cast methods for nodes

xunilrj:
This PRs creates quick ```as...``` conversions methods for easy navigation on the tree. Example:

```rust
impl JsAnyExpression {
    pub fn as_import_meta(&self) -> Option<&ImportMeta> {
        match &self {
            JsAnyExpression::ImportMeta(item) => Some(item),
            _ => None,
        }
    }
    pub fn as_js_any_literal_expression(&self) -> Option<&JsAnyLiteralExpression> {
        match &self {
            JsAnyExpression::JsAnyLiteralExpression(item) => Some(item),
            _ => None,
        }
    }
    ....
}
```` 

This will be used in https://github.com/rome/tools/pull/2489

### DIFF
--- a/crates/rome_css_syntax/src/generated/nodes.rs
+++ b/crates/rome_css_syntax/src/generated/nodes.rs
@@ -1405,20 +1405,88 @@ pub enum CssAnyAtMediaQueryFeatureType {
     CssAtMediaQueryFeaturePlain(CssAtMediaQueryFeaturePlain),
     CssAtMediaQueryFeatureRange(CssAtMediaQueryFeatureRange),
 }
+impl CssAnyAtMediaQueryFeatureType {
+    pub fn as_css_at_media_query_feature_boolean(&self) -> Option<&CssAtMediaQueryFeatureBoolean> {
+        match &self {
+            CssAnyAtMediaQueryFeatureType::CssAtMediaQueryFeatureBoolean(item) => Some(item),
+            _ => None,
+        }
+    }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by ematipico** [comment]
Out of curiosity, how they will be used and for what?

**Review by MichaReiser — APPROVED** [review]

**Comment by xunilrj** [comment]
I am still discussing if this makes sense or not when searching for symbols, but just as an example: normally a string literal expression would never be a symbol, nor a reference to a symbol. If you click F12 on it, it would go anywhere.

But inside computed member expressions, maybe it does make sense to treat it like a reference to a symbol. That is what the Typescript compiler does.

```js
let a = { b: 1 };
console.log(a["b"]);
```

In this case, I found it easier to navigate downward from ```JS_COMPUTED_MEMBER_EXPRESSION``` using our typed API, than upwards from ```JS_STRING_LITERAL_EXPRESSION", which is an untyped navigation.

```rust
JsSyntaxKind::JS_COMPUTED_MEMBER_EXPRESSION => {
            let value_token = unsafe { JsComputedMemberExpression::new_unchecked(node) }
                .as_fields()
                .member
                .ok()?
                .as_js_any_literal_expression()?
                .as_js_string_literal_expression()?
                .as_fields()
                .value_token
                .ok()?;

            Some(Symbol::Reference {
                name: value_token.text_trimmed().to_string(),
                range: value_token.text_range(),
                declared_at: None,
            })
        }

```

PR #2536, opened 2022-05-04
feat(rome_js_formatter): elide quotes in object/class members

### LINKED ISSUE #2405: Only add quotes around object properties where needed

ematipico:
## Summary

This PR closes #2405

I cherry-picked the commits of the previous PR that introduced these changes in the first place, then applied the logic for checking the source type.

## Test Plan

I created new test cases.

Unfortunately it's not easy to compare it with Prettier  output, that's why I suggest to [compare their playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACAGtgXm2AB0psLsSQBbOGgIzgCdqlsBnGZgSygHMA3GUrYA9GOxoIAV04wAhmmwAbHgGscMABY8OIytTqMWAWmNNWybFBknmAGgMVqAFjY27lp+UMgA7jw6APoyUAAmLByQzHAetvbOVCAADACMAJwA7PFeLD6i1ADMULn2BX5FGGXeSdQYGTUsdSDBACTBAGadHTQQzFB8-MFQEDDBAI4yY3D61gm1vi6tHd29-YMCwWizk9PwHMEAAk2sUAC+wqXXfPDMnQpgOACaxC0WzdZcvAJXohJSWTyJSqDRaXT6JbJD7Mcz0SynCrLdzzPKOFqBEJhSLMaL9OKoxJQ6jpbKIlolcnEkBVKmFEANU4tdpdHrBPoDIYjMZ7GZzdgLfLM1ZsjmbYY7Q5TPnHJkXMhkFRwGBvakwjzfIZI8SSaRyLggtSabA6PTveFmdWExb0lECtHa6iY7ShCJRGIE+3lFqknLWoXUyn+9GB6rBx0Mxrh4Ws9acrajcbSg50vwstbsjZcyW8g6y6MXQjYABuChUMjgghADhAEAwMB40A4yFACmYzAg-gACm2EM2UGX-EpmzWGMxHpoYABlBR0AAyfDgyAeKg4cFH47Ak6nGEeQ2Q3ArNYt4Ui4TnCgEMgU-DgADF+jQFDAGwJkCAFDIYBBqyBtDAaBUAB1XQDl3J4pz7IIeGLII0HfMAOBHEA+DXZgYC7cd+CfZcyzXGsACsOAADynIYlQARX2JckBXfCGTbND33QLBol4etfwwH4YCAnhwh0ZAAA4UhrLiIDXIDxwwd8uNmFhiyXGtkzgTC637D8OFMKA4DgM9f1iKYeFiTCbxw2i8PXEA1xoHgD2YI8rPIuAqJmXDV0sxQGF4-jtGQAAmGtuAUHg1AEABhCAaDMkBZgAVl-GQ1wAFQUBh+0PSziwrABJN1YCnMB2JgABBCIp3QJU3LXc5ziAA) with our snapshots.


crates/rome_js_formatter/tests/specs/prettier/typescript/custom/computedProperties/string.ts.snap
```diff
@@ -26,16 +26,16 @@ type B = {
...
-type T = { "string": "T" };
+type T = { string: "T" };
 
 interface A {
-  "string": "A";
+  string: "A";
 }
 
-type B = { "string": "B" };
+type B = { string: "B" };
```
ematipico, line 38: This matches prettier's output

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -698,18 +698,14 @@ impl MemberContext {
...
-                    .all(|c| matches!(c, '_' | '$') || c.is_alphanumeric())
+        text_to_check.chars().enumerate().all(|(index, c)| {
+            // Members that starts with numbers in their name are not valid syntax,
+            // hence we can't elide the quotes
+            if index == 0 && c.is_numeric() {
+                return false;
             }
-        }
+            matches!(c, '_' | '$') || c.is_alphanumeric()
+        })
```
IWANABETHATGUY, line 708: in most case property-name would be `alphanumeric`, so I guess it would be more efficient change   the code to ` c.is_alphanumeric() ||  matches!(c, '_' | '$')  `

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -698,18 +698,14 @@ impl MemberContext {
...
-        match self {
-            // TODO: #2405 implement number literals on JS and TS
-            // patterns.all(|sub_text| sub_text.chars().all(char::is_alphabetic))
-            MemberContext::Type => false,
-            MemberContext::Member => {
-                // patterns.all(|sub_text| sub_text.chars().all(char::is_alphanumeric))
-                text_to_check
-                    .chars()
-                    .all(|c| matches!(c, '_' | '$') || c.is_alphanumeric())
+        text_to_check.chars().enumerate().all(|(index, c)| {
```
MichaReiser, line 701: Do we still need the `MemberContext` now that the rule seems to be the same for types and regular members.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -698,18 +698,14 @@ impl MemberContext {
...
+            // Members that starts with numbers in their name are not valid syntax,
+            // hence we can't elide the quotes
+            if index == 0 && c.is_numeric() {
```
MichaReiser, line 704: I just checked this again and we forgot that we're talking about members, not identifiers. 

A all digits string member can be turned into a number literal that.

```
{
  "123": true,
  // same as
  123: true
}
```

However, Prettier doesn't unquote all number members in TypeScript.

So the rule must support:

* string forming a valid identifier -> print as identifier
* string forming a valid number -> print as number literal (only for JS)

MichaReiser (commented):
>  It turns out that prettier applies the same logic to both JS and TS.

This isn't the case for all number

ematipico:
> This isn't the case for all number

Could you please share some cases that I might have missed? I couldn't find one

crates/rome_js_formatter/src/utils/mod.rs
ematipico: That is true, but prettier doesn't seem to the quotes from numeric like members (that can be valid members):  [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAExMB0QCMATAMz5KYBmmAvpgLxUDcmIANCBAA4wCW0AZ2SgAhgCcxEAO4AFcQiEoRqKSICeQjgCMxIsAGsMAZREBbOABleUOMkrKBcbboPGue6wHNkMMQFcnEDhTLTgAEzDwixEoTz8RTzgAMQgxUxEYPljkEBE-GAh2EAALGFNUAHVi3ngBdzA4IwUa3gA3GrUcsAFNEGtHMRgZXU90uwdAgCsBAA8jL3QART8IeHHURw53MQGcmDUuOAEwMV4eIq5T2AreMJhi5AAOAAYtyUcK3S4cy6O4MVatg4AEcVvBhtxFLkBABaGzhcJFMRwUG8ZHDBJjJD2DaBRymXg+fx4hZwZarWzYiYcGAiLQ3O4PJBEGm6XioLwAYQgpixQQEAFYin5HAAVOmKXwBDitAIASSgkVgRhOZxgAEFFUZ9uh1o4aDQgA)

MichaReiser:
> > This isn't the case for all number
> 
> Could you please share some cases that I might have missed, please? I couldn't find one

Sorry, I didn't intend to submit that comment. I first started this comment before I then wrote the more detailed comment in `utils/mod.rs`

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -698,18 +699,27 @@ impl MemberContext {
...
+        text_to_check.chars().enumerate().all(|(index, c)| {
+            if index == 0 && c.is_numeric() {
+                // In TypeScript, numbers like members have different meaning from numbers.
+                // Hence, if we see a number, we bail straightaway
+                if self == &MemberContext::Type {
+                    return false;
+                } else {
+                    has_seen_number = true;
+                }
             }
```
MichaReiser, line 713: My understanding is that the difference isn't based on whether this is a type or not but IF typescript is enabled or not. 

In my view, the algorithm must be something along the following lines:

Create two booleans: `validIdentifier` and `validNumberLiteral` and initialize both to true

Then iterate over all characters:
* If you encounter any non-numeric value, set `validNumberLiteral` to `false`
* If you encounter a number and it's the first character, set `validIdentifier` to `false`
* If you encounter any non-alphanumeric character that isn't `_` or `$` set `validIdentifier` to false

If `source_language != TypeScript` && `validNumberLiteral` || `validIdentifier`, remove the quotes. Otherwise don't. 

You should then also be able to remove the `MemberContext`

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: You need to change the parser to `babel`.

MichaReiser (changes requested):

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: To add two examples:

## JavaScript file

**Input**

```js
({
  "1": test
})
```

**Output**

```js
({
  1: test,
});
```

## Typescript file

**Input**

```ts
({
  "1": test
})
```

**Output**

```ts
({
  "1": test,
});
```

ematipico:
PR on hold as at the moment the formatter doesn't have the understanding if it's parsing a TS file or a JS file. 

In order to give this information to the formatter, we have to extract it from the `SourceType`. We will need to pass this information first, and then we might be able to achieve this.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: This should be flagged now, I added a new test that should cover the case

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -394,3 +394,69 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+
+impl From<JsLiteralMemberName> for MemberName {
+    fn from(literal: JsLiteralMemberName) -> Self {
+        Self::Literal(literal)
+    }
+}
+
+/// Function used by the formatter, where we pass a complaint member and it returns a [FormatElement[
+/// where the text has its quotes removed.
+pub(crate) fn format_member_name<Member: Into<MemberName>>(
```
MichaReiser, line 432: Nit: Can we implement `Format` on `MemberName` instead. This will fit nicer into the new architecture. 

`write!(f, [MemberName::from(member)])?`

And I would recommend renaming this struct to `FormatMemberName` to make sure it's an object that can format a member name.

crates/rome_js_formatter/src/utils/string_utils.rs
```diff
@@ -210,42 +170,153 @@ impl<'token> LiteralStringNormaliser<'token> {
...
+    /// already written
+    AlreadyPrinted(char),
+}
+
+/// Struct of convenience used to manipulate the string. It saves some state in order to apply
+/// the normalise process.
+pub(crate) struct LiteralStringNormaliser<'token> {
+    token: &'token FormatLiteralStringToken<'token>,
+    ///
+    ///
```
MichaReiser, line 201: Delete or document?

MichaReiser (commented):
I'll need to review the rest tomorrow. This is some sophisticated stuff that my brain isn't capable of handling at this time of a day :D

crates/rome_js_formatter/src/js/assignments/object_assignment_pattern_property.rs
```diff
@@ -21,7 +21,7 @@ impl FormatNodeFields<JsObjectAssignmentPatternProperty>
         formatted![
             formatter,
             [
-                member.format(),
+                FormatMemberName::from(member?).format(formatter),
```
MichaReiser, line 24: Nit: You can remove the `.format` here. 
```suggestion
                FormatMemberName::from(member?),
```

crates/rome_js_formatter/src/js/classes/constructor_class_member.rs
```diff
@@ -21,7 +21,7 @@ impl FormatNodeFields<JsConstructorClassMember> for FormatNodeRule<JsConstructor
             [
                 modifiers.format(),
                 space_token(),
-                name.format(),
+                FormatMemberName::from(name?).format(formatter),
```
MichaReiser, line 24:
```suggestion
                FormatMemberName::from(name?),
```

crates/rome_js_formatter/src/js/classes/getter_class_member.rs
```diff
@@ -26,7 +26,7 @@ impl FormatNodeFields<JsGetterClassMember> for FormatNodeRule<JsGetterClassMembe
                 space_token(),
                 get_token.format(),
                 space_token(),
-                name.format(),
+                FormatMemberName::from(name?).format(formatter)?,
```
MichaReiser, line 29:
```suggestion
                FormatMemberName::from(name?),
```

crates/rome_js_formatter/src/js/classes/method_class_member.rs
```diff
@@ -30,7 +30,7 @@ impl FormatNodeFields<JsMethodClassMember> for FormatNodeRule<JsMethodClassMembe
                     .format()
                     .with_or_empty(|token| formatted![formatter, [token, space_token()]]),
                 star_token.format(),
-                name.format(),
+                FormatMemberName::from(name?).format(formatter),
```
MichaReiser, line 33:
```suggestion
                FormatMemberName::from(name?),
```

crates/rome_js_formatter/src/js/classes/setter_class_member.rs
```diff
@@ -26,7 +26,7 @@ impl FormatNodeFields<JsSetterClassMember> for FormatNodeRule<JsSetterClassMembe
                 space_token(),
                 set_token.format(),
                 space_token(),
-                name.format(),
+                FormatMemberName::from(name?).format(formatter),
```
MichaReiser, line 29:
```suggestion
                FormatMemberName::from(name?),
```

crates/rome_js_formatter/src/ts/auxiliary/property_signature_type_member.rs
```diff
@@ -25,7 +25,7 @@ impl FormatNodeFields<TsPropertySignatureTypeMember>
             [
                 readonly_token.format(),
                 space_token(),
-                name.format(),
+                FormatMemberName::from(name?).format(formatter),
```
MichaReiser, line 28:
```suggestion
                FormatMemberName::from(name?),
```

crates/rome_js_formatter/src/ts/classes/method_signature_class_member.rs
```diff
@@ -31,7 +31,7 @@ impl FormatNodeFields<TsMethodSignatureClassMember>
                         .format()
                         .with_or_empty(|token| formatted![formatter, [token, space_token()]]),
                     space_token(),
-                    name.format(),
+                    FormatMemberName::from(name?).format(formatter),
```
MichaReiser, line 34:
```suggestion
                    FormatMemberName::from(name?),
```

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -394,3 +394,70 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+/// - [JsAnyObjectMemberName]
+/// - [JsAnyClassMemberName]
+/// - [JsLiteralMemberName]
+///
+/// Once merged, the enum is used to get specific members (the literal ones) and elide
+/// the quotes from them, when the algorithm sees fit
+pub(crate) enum FormatMemberName {
+    Object(JsAnyObjectMemberName),
+    Class(JsAnyClassMemberName),
+    Literal(JsLiteralMemberName),
```
MichaReiser, line 409: The formatting seems to mainly depend on the kind of the member name and less on whatever the parent is an object, class or literal. How about 

```rust
pub(crate) FormatMemberName {
  LiteralMemberName(JsLiteralMemberName),
  ComputedMemberName(JsComputedMemberName),
  PrivateClassMemberName(JsPrivateClassMemberName),
}
```

This adds a bit more complexity to `from` but simplifies the formatting (removes a lot of duplication)

crates/rome_js_formatter/src/utils/string_utils.rs
```diff
@@ -210,42 +170,153 @@ impl<'token> LiteralStringNormaliser<'token> {
...
+    token: &'token FormatLiteralStringToken<'token>,
+    /// The quote that was set inside the configuration
+    chosen_quote: QuoteStyle,
+}
+
+#[derive(Eq, PartialEq)]
+pub(crate) enum FileSource {
+    Ts,
+    Js,
+}
```
MichaReiser, line 209: Nit: `SourceFileKind`? and maybe write out `TypeScript` and `JavaScript` (or add documentation but I would prefer more descriptive names in favour of documentation).

crates/rome_js_formatter/src/utils/string_utils.rs
```diff
@@ -210,42 +170,153 @@ impl<'token> LiteralStringNormaliser<'token> {
...
         }
-        self.raw_content_has_quotes =
-            Some(preferred_quotes_count > 0 || alternate_quotes_count > 0);
     }
+}
 
-    fn clean_directive(&mut self) -> Cow<str> {
-        let content = self.normalize_string();
+impl<'token> LiteralStringNormaliser<'token> {
+    pub fn new(token: &'token FormatLiteralStringToken<'_>, chosen_quote: QuoteStyle) -> Self {
```
MichaReiser, line 222: I think you can make all these types private (`FileSource`, `LiteralStringNormalizer`, `StringInformation`) as well as all the functions. Or are they called from outside this module?

crates/rome_js_formatter/src/utils/string_utils.rs
```diff
@@ -210,42 +170,153 @@ impl<'token> LiteralStringNormaliser<'token> {
...
+            Self::Js
         }
-        self.raw_content_has_quotes =
-            Some(preferred_quotes_count > 0 || alternate_quotes_count > 0);
     }
+}
 
-    fn clean_directive(&mut self) -> Cow<str> {
-        let content = self.normalize_string();
+impl<'token> LiteralStringNormaliser<'token> {
```
MichaReiser, line 221: What's the motivation of splitting this out into its own struct. Seems like most methods are only operating on the `FormatStringLiteralToken`

MichaReiser (commented):
Looks good. One thing that I'm wondering is if it would be better to have small helper functions for the string cleaning but then have different

* `FormatLiteralMemberName` 
* `FormatDirectiveLiteral` 
* ... 

Functions that call into this helpers. It could help to remove some branching and may help to split this code into some smaller files. 

I'm considering this because it feels like it's becoming this one string manipulation that is able to handle ALL cases which makes it harder to reason about how the code works (for a specific case)

crates/rome_js_formatter/src/utils/string_utils.rs
ematipico: I allows me to have a better testing suite, because I don't need to have a fully fledged AST. Like this I can just create a token out of a tree builder and then pass the options I need for the cleaning.

MichaReiser: You can use `parse` to get tokens or maybe even @leops mutation API?

ematipico: Still, the testing is **easier this way**, because even though I could use the parser, I would need to assert it against the final IR, which is not the point of the testing. Testing against IR is more error prone because it relies on how the parents are formatted. I am happy with current setup.

ematipico:
> Looks good. One thing that I'm wondering is if it would be better to have small helper functions for the string cleaning but then have different
> 
>     * `FormatLiteralMemberName`
> 
>     * `FormatDirectiveLiteral`
> 
>     * ...
> 
> 
> Functions that call into this helpers. It could help to remove some branching and may help to split this code into some smaller files.
> 
> I'm considering this because it feels like it's becoming this one string manipulation that is able to handle ALL cases which makes it harder to reason about how the code works (for a specific case)

Makes sense. Allow me to address these concerns in the next PR. The works on these PR are not finished yet. The cases are not over unfortunately. At this point it might make more sense to also have a small module for each case (directive, member, literal string)

MichaReiser:
> > Looks good. One thing that I'm wondering is if it would be better to have small helper functions for the string cleaning but then have different
> > ```
> > * `FormatLiteralMemberName`
> > 
> > * `FormatDirectiveLiteral`
> > 
> > * ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Functions that call into this helpers. It could help to remove some branching and may help to split this code into some smaller files.
> > I'm considering this because it feels like it's becoming this one string manipulation that is able to handle ALL cases which makes it harder to reason about how the code works (for a specific case)
> 
> Makes sense. Allow me to address these concerns in the next PR. The works on these PR are not finished yet. The cases are not over unfortunately (we still have the escaping of other chars) . At this point it might make more sense to also have a small module for each case (directive, member, literal string)

I'm not opposed to it. Is this something you're planning to work on soon?

ematipico:
> I'm not opposed to it. Is this something you're planning to work on soon?

Yes, it's in my bucket list :)

MichaReiser (approved):

PR #2537, opened 2022-05-04
refactor(formatter): Implement JS formatting on `FormatRule` trait

MichaReiser:


crates/rome_js_formatter/src/lib.rs
```diff
@@ -100,6 +101,157 @@ where
...
+where
+    R: FormatRule<T>,
+{
+    fn format(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        self.rule.format(&self.item, formatter)
+    }
+}
+
+/// Formats the referenced `item` with the specified rule.
+pub struct FormatRefWithRule<'a, T, R> {
```
MichaReiser, line 134: I haven't figured out a way to handle both `FormatWithRule` and `FormatRefWithRule` in a single struct.

MichaReiser:
I'll follow up with an alternative design

Issue #2449, opened 2022-04-14
Parenthesize expressions


MichaReiser:
Prettier uses an exhaustive heuristic of when it should insert parentheses or not. Rome should implement a similar heuristic. 

https://github.com/prettier/prettier/blob/6d02cd9650be29d1b11ab6ff1442eac4f5415c06/src/language-js/needs-parens.js#L20-L793

Relevant test cases with differences:

* js/assignment/sequence.js
* js/async/await-parse.js
* js/async/conditional-expression.js
* js/async/inline-await.js
* js/async/nested.js
* js/babel-plugins/private-fields-in-in.js
* js/binary-expressions/equality.js
* js/binary-expressions/exp.js
* js/binary-expressions/math.js
* js/binary_math/parens.js
* js/class-extends/extends.js
* js/classes/binary.js
* js/classes/call.js
* js/classes/member.js
* js/classes/new.js
* js/classes/ternary.js
* js/conditional/no-confusing-arrow.js
* js/export-default/function_tostring.js
* js/for-of/async-identifier.js
* js/function/function_expression.js
* js/in/arrow-function.js
* js/logical_expressions/issue-7024.js
* js/logical_expressions/logical_expression_operators.js
* js/new-expression/new_expression.js
* js/nullish-coalescing/nullish_coalesing_operator.js
* js/objects/expand.js
* js/optional-chaining/chaining.js
* js/optional-chaining/eval.js
* js/reserved-word/interfaces.js
* js/spread/spread.js
* js/template/parenthesis.js
* js/unary-expression/urnary_expression.js
* js/yield/arrow.js
* js/yield/conditional.js
* js/yield/jsx.js

ematipico:
FYI, our formatting architecture is different from prettier's:

- By default they always remove parenthesis and they have a function that tells the IR to not remove them.
- We format by node, they use a different strategy. Which means that check of parenthesis in our case must be done on node level, no need to recreate prettier's function.

MichaReiser:
Closed by #3057

PR #2538, opened 2022-05-05
fix(rome_js_formatter): use parenthesis

ematipico:
## Summary

This is part of https://github.com/rome/tools/issues/2449

It's a long list so it won't be fixed in one single PR. Given this file: https://github.com/prettier/prettier/blob/a9de2a128cc8eea84ddd90efdc210378a894ab6b/src/language-js/needs-parens.js

> One thing to note about the above file, is that it contains logic of syntax that is not official yet. Also, it contains some weird cases that are syntactically wrong, probably because they needed to cover some new syntax case. Given that, the progression on this task takes longer then expected

I fixed cases for:
- classes
- parenthesized expressions (removal of unnecessary parenthesis)
- await expressions
- some edge case with sequence expressions

I also did some work for the conditional expressions, although it can't be completed because at the moment we don't correctly format them. 

Once we do, can add parenthesis to [this case](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=true&jsx=false&sourceType=module#dAB5AHAAZQAgAEEAIAA9ACAAQgAgAGUAeAB0AGUAbgBkAHMAIABTAHQAcgBpAG4AZwAgAD8AIAAgAEMAIABlAHgAdABlAG4AZABzACAAbgB1AG0AYgBlAHIAIAA/ACAAbgBlAHYAZQByACAAOgAgAG4AZQB2AGUAcgAgADoAIABuAGUAdgBlAHIA)

## Test Plan

I progressively added new test cases every time I covered a new case.


crates/rome_js_formatter/tests/specs/js/module/class/class.js.snap
```diff
@@ -124,6 +130,10 @@ x = class foo extends Boar {};
 x =
 	class aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa extends bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb {};
 
+class A extends (new A()) {}
+class C extends Something! {}
```
ematipico, line 134: For some reason this doesn't get the parenthesis, but when I run the test individually, I do see the parenthesis. Any help to understand what's going on would be helpful!

crates/rome_js_formatter/tests/specs/js/module/expression/unary_expression.js.snap
```diff
@@ -45,6 +52,12 @@ x =
 x =
 	!aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
 
++(++f);
+-(--f);
+
+-(await fn());
+typeof await fn();
```
ematipico, line 59: Missed this case! Going to push a new commit

crates/rome_js_formatter/tests/specs/js/module/class/class.js.snap
MichaReiser: Is this a TS file? `Something!` is TS syntax.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -581,6 +581,18 @@ impl FormatPrecedence {
...
+            JsSyntaxKind::JS_SEQUENCE_EXPRESSION => {
+                let parent_kind = node.parent().map(|p| p.kind());
+                if matches!(
+                    parent_kind,
+                    Some(JsSyntaxKind::JS_FOR_STATEMENT | JsSyntaxKind::JS_RETURN_STATEMENT)
+                ) {
+                    FormatPrecedence::Low
+                } else {
+                    FormatPrecedence::High
+                }
```
ematipico, line 593: This piece logic covers: 

```js
for ((i = 0), (len = arr.length); i < len; i++) {
  console.log(arr[i])
}
```

crates/rome_js_formatter/src/js/classes/extends_clause.rs
```diff
@@ -14,10 +13,26 @@ impl FormatNode for JsExtendsClause {
...
+                | JsSyntaxKind::JS_OBJECT_EXPRESSION
+                | JsSyntaxKind::TS_NON_NULL_ASSERTION_EXPRESSION
+        );
+        let super_class = super_class.format_with(formatter, |super_class| {
+            if needs_parens {
+                format_elements![token("("), super_class, token(")")]
+            } else {
+                super_class
+            }
+        })?;
```
MichaReiser, line 30: Adding parentheses to a sub-expression seems to be a common pattern. Could we introduce a `format_sub_expression(expression: JsAnyExpression, formatter; &Formatter) -> FormatResult<FormatElement>` method instead that internally uses a match similar to prettier's to determine if parentheses for this expression are needed. 

This would reduce the boilerplate code that now had to be added to every place where we format a sub expression.

ematipico: I did try this approach at first, but I didn't work for some particular cases. 

- prettier's approach orbits around their architecture of formatting things, which is different from ours; by default their approach is elide parenthesis as much as they can, and then they use that monster switch to refrain algorithm from doing so; (I didn't study it in detail, but I saw some issue where prettier was removing parenthesis where shouldn't not, and they added some cases there... we don't have this problem here)
- their AST is different from ours, and sometimes it didn't work inside a big match; I found myself moving code from the generic function and put it inside the specific node;
- adding parenthesis requires great context, which would lead to the same issue I introduced in the sequence expressions, where I added a context but it wouldn't work for formatting range;

crates/rome_js_formatter/src/js/expressions/unary_expression.rs
```diff
@@ -27,10 +29,18 @@ impl FormatNode for JsUnaryExpression {
         );
 
         if is_keyword_operator {
+            let needs_parenthesis = JsAwaitExpression::can_cast(argument.syntax().kind());
```
NicholasLYang, line 32: nit: we can do a match on argument directly: 
```suggestion
            let needs_parenthesis = matches!(argument, JsAnyExpression::JsAwaitExpression(_));
```

crates/rome_js_formatter/src/js/expressions/unary_expression.rs
```diff
@@ -53,9 +63,11 @@ impl FormatNode for JsUnaryExpression {
                         | (JsUnaryOperator::Minus, JsPreUpdateOperator::Decrement)
                 )
             }
+            JsAnyExpression::JsAwaitExpression(_) => true,
             _ => false,
         };
 
+        dbg!(&is_ambiguous_expression);
```
NicholasLYang, line 70: Any reason to keep this?

NicholasLYang (commented):
Unrealistic situation, but any reason [this doesn't match prettier?](https://play.rome.tools/9bc842f/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#YQB3AGEAaQB0ACAAKABhAHcAYQBpAHQAIABmAG8AbwApAA==)

MichaReiser:
I think we need a more general approach to handling parentheses rather than doing it on a per-node level as discussed on [discord](https://discord.com/channels/678763474494423051/895012159011053599/977251904834900078). Do you want to explore if it would be feasible to handle parentheses inside the `FormatNodeRule`?

MichaReiser:
> > I think we need a more general approach to handling parentheses rather than doing it on a per-node level as discussed on [discord](https://discord.com/channels/678763474494423051/895012159011053599/977251904834900078). Do you want to explore if it would be feasible to handle parentheses inside the `FormatNodeRule`?
> 
> If you feel strong about this, then I am happy to delegate the implementation to someone else. As I explained in [#2538 (comment)](https://github.com/rome/tools/pull/2538#discussion_r866572587), I already explored this approach and it didn't work as intended.

What I'm looking for in an implementation is that it is possible to keep the removal of parentheses in the parenthesized expression formatting rule and the addition of parentheses in the case where they are needed in sync.

PR #2539, opened 2022-05-05
fix(playground): update lock file

ematipico:
## Summary

This PR fixes the current broken build. We upgraded `pnpm` but we forgot to update the lock file.

## Test Plan

Merging it, it should fix the issue

### DIFF
--- a/website/playground/pnpm-lock.yaml
+++ b/website/playground/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: 5.3
+lockfileVersion: 5.4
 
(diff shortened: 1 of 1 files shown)


cpojer:
Sorry about that. The CI actions didn't run when I sent the PR, so I had no idea and also no way of testing a fix.

MichaReiser:
Thanks @ematipico for fixing the CI job!

ematipico:
I didn't unfortunately. Not sure what's the issue, but we should revert it.

PR #2541, opened 2022-05-05
fix(rome_js_formatter): correctly format sequence expressions

### LINKED ISSUE #2413: Long sequence expressions

ematipico:
## Summary

This PR fixes #2413 

The formatter must give to the sequence expression formatting a context, because based on that the type of indentation changes.

Because of that, we now use a function called `format_sequence_expression` that is called in particular cases: when it's a body of an arrow function and in a return statement. Obviously also when it's inside a normal sequence expression, where the context is irrelevant.

## Test Plan

Added three cases that aim to fix them

### Metrics

PR
**File Based Average Prettier Similarity**: 70.87%  
**Line Based Average Prettier Similarity**: 64.93% 

`main`

**File Based Average Prettier Similarity**: 70.84%  
**Line Based Average Prettier Similarity**: 64.85%


crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -18,7 +20,30 @@ impl FormatNode for JsParenthesizedExpression {
             r_paren_token,
         } = self.as_fields();
 
+        let expression = expression?;
         let parenthesis_can_be_omitted = parenthesis_can_be_omitted(self)?;
+        let is_in_function = self.syntax().parent().map_or(false, |p| {
```
MichaReiser, line 25: I find the `is_in_function` name misleading because it is only true if the direct parent is an arrow function but isn't in the case where the direct parent is a function expression or function statement. 

If I look at [Prettier's source](https://github.com/prettier/prettier/blob/d13feed42b6478710bebbcd3225ab6f203a914c1/src/language-js/print/function.js#L461-L473) then it handles sequence expression differently if used with a `throw` statement but it manages to do so without making any changes to how the `SequenceExpression` gets printed. It's where it adds the parentheses and the grouping. 

And [this](https://github.com/prettier/prettier/blob/d13feed42b6478710bebbcd3225ab6f203a914c1/src/language-js/print/function.js#L263-L267) is the source for printing sequence expressions inside of arrow functions (only, not sure how they handle the case where it's wrapped in a parenthesized expression). But your implementation right now isn't handling the opposite case where there's a sequence expression in an arrow body that isn't parenthesized. 

That's why I think it should be possible to remove the Context from `format_sequence_expression` altogether and handle the "contextual" formatting all in the parent (that already has the necessary context).

crates/rome_js_formatter/src/utils/format_sequence_expression.rs
```diff
@@ -0,0 +1,85 @@
...
+use rome_js_syntax::{JsSequenceExpression, JsSequenceExpressionFields};
+use rome_rowan::AstNode;
+
+#[derive(Eq, PartialEq)]
+pub(crate) enum SequenceContext {
+    InFunction,
+    None,
+}
+
+pub(crate) fn format_sequence_expression(
```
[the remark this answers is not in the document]
ematipico: It is, but it's used in other places other than the simple sequence expression. That's the motive why I moved it elsewhere. Do you still think it's worth keeping it there?

crates/rome_js_formatter/src/utils/format_sequence_expression.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let mut formatted = vec![format_elements![
+        left.format(formatter)?,
+        comma_token.format(formatter)?
+    ]];
+
+    let mut previous_right = right;
+
+    // Traverse upwards again and concatenate the sequence expression until we find the first non-sequence expression
+    while let Some(parent) = current.syntax().parent() {
```
leops, line 44: This could be more directly written as `while let Some(_) = current.syntax().parent().and_then(JsSequenceExpression::cast)`

crates/rome_js_formatter/src/utils/format_sequence_expression.rs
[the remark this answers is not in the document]
MichaReiser: In my view it is conceptually closest to where we implement the formatting of sequence expressions. The fact that it is called from different places doesn't change that. If you think about the parser, the parsing rules are grouped by the node types and not from where the parsing rules are called from. 

Putting files into `utils` should be our last resort because that means. This is something that we don't really know where it conceptually belongs so we put it here. The downside of this is that it becomes difficult to find files because I first start searching for the logic that formats sequence expressions in the `js/expression` folder but I won't be able to find it.

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -18,7 +20,30 @@ impl FormatNode for JsParenthesizedExpression {
...
+                p.kind(),
+                JsSyntaxKind::JS_ARROW_FUNCTION_EXPRESSION | JsSyntaxKind::JS_RETURN_STATEMENT
+            )
+        });
+
+        let expression = if is_in_function {
+            if let Some(sequence_expression) =
+                JsSequenceExpression::cast(expression.syntax().clone())
+            {
+                format_sequence_expression(
```
leops, line 36: The `SequenceContext` flag will not work properly when doing range formatting: if the user has selected only the sequence expression then it's implementation of `ToFormatElement` will be called directly, and in turn call `format_sequence_expression` with a context of `None` regardless of what the parent node actually is

ematipico: What would you suggest then?

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
ematipico: I studied that source code, and came to the conclusion that their  approach would not always work for us. 

It's true that we can't use a context because otherwise range formatting would not work. On the other hand we have to signal to the sequence expression formatting because of the indentation of each sub expression. Not sure what we should solve that than.

 > But your implementation right now isn't handling the opposite case where there's a sequence expression in an arrow body that isn't parenthesized.
 
 Could you share some example so I can add it to the test suite?

ematipico: > But your implementation right now isn't handling the opposite case where there's a sequence expression in an arrow body that isn't parenthesized.

If you mean 
```js 

const f = () => aLongIdentifierName, aLongIdentifierName, aLongIdentifierName, aLongIdentifierName

```

It's not a valid program

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
ematipico: I removed the context and made the check of the parents inside the function itself

crates/rome_js_formatter/tests/specs/prettier/js/objects/expression.js.snap
```diff
@@ -55,7 +56,9 @@ const a2 = {
 
 const a3 = {
   someKey: (
-    longLongLongLongLongLongLongLongLongLongLongLongLongLongName, longLongLongLongLongLongLongLongLongLongLongLongLongLongName, longLongLongLongLongLongLongLongLongLongLongLongLongLongName
+    longLongLongLongLongLongLongLongLongLongLongLongLongLongName,
+      longLongLongLongLongLongLongLongLongLongLongLongLongLongName,
+      longLongLongLongLongLongLongLongLongLongLongLongLongLongName
   ),
```
yassere, line 62: This doesn't match Prettier's formatting. Although it's maybe out of scope for this PR since it's more of an issue with parenthesized expressions than sequence expressions.

crates/rome_js_formatter/tests/specs/js/module/expression/sequence_expression.js.snap
```diff
@@ -1,10 +1,24 @@
...
 
+const f = () => (
+    ____________first, ____________second, ____________third, ____________third, ____________third, ____________third, ____________third
+
+)
+
+(
+    ____________first, ____________second, ____________third, ____________third, ____________third, ____________third, ____________third
+
+)
```
yassere, line 17: If this is intended to be an arrow function followed by a parenthesized sequence expression, they need to be divided by a semicolon.

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,56 +1,107 @@
...
+        // Return statement already does the indentation for us
+        // Arrow function body can't have a sequence expression unless it's parenthesized, otherwise
+        // would be a syntax error
+        if matches!(parent.kind(), JsSyntaxKind::JS_RETURN_STATEMENT) {
+            true
+        } else if matches!(parent.kind(), JsSyntaxKind::JS_PARENTHESIZED_EXPRESSION) {
+            // In case we are inside a sequence expression, we have to go up a level and see the great parent.
+            // Arrow function body and return statements applying indentation for us, so we signal the
+            // sequence expression to not add other indentation levels
+            let great_parent = parent.parent().map(|gp| gp.kind());
```
yassere, line 31: Nit: Calling this `grand_parent` would be more conventional than `great_parent`

crates/rome_js_formatter/tests/specs/prettier/js/objects/expression.js.snap
ematipico: Actually, you're right. Excluded the parenthesis, the sequence expression should not have indentations.

ematipico:
PR rebased.

crates/rome_js_formatter/tests/specs/prettier/js/sequence-break/break.js.snap
```diff
@@ -27,25 +28,37 @@ const f = (
...
 );
 for (
-  aLongIdentifierName =
-    0, aLongIdentifierName =
-    0, aLongIdentifierName = 0, aLongIdentifierName = 0;
+  aLongIdentifierName = 0,
+    aLongIdentifierName = 0,
+    aLongIdentifierName = 0,
+    aLongIdentifierName = 0
+  ;
```
leops, line 61: This semicolon should not be getting pushed to the next line, I'm guessing the indent block should not have a trailing line break if the sequence is not parenthesized ?

ematipico: True, I suppose I have to manually add the indentation

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,60 +1,117 @@
...
     ) -> FormatResult<FormatElement> {
-        let mut current = node.clone();
+        format_sequence_expression(node, formatter)
+    }
+}
+
+pub(crate) fn format_sequence_expression(
+    node: &JsSequenceExpression,
+    formatter: &Formatter<JsFormatOptions>,
+) -> FormatResult<FormatElement> {
```
MichaReiser, line 19: Nit: What's the motivation of moving this into its own function rather than keeping it inside `format_fields`? Calling `format_sequence_expression` directly risks that you'll skip over suppression comments.

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,60 +1,129 @@
...
+        current = parent_sequence;
+    }
+
+    if has_already_indentation {
+        formatted.push(formatted![
+            formatter,
+            [soft_line_break_or_space(), previous_right.format()?,]
+        ]?);
+    } else {
+        formatted.push(formatted![
```
MichaReiser, line 116: Nit: We can now use `ConcatBuilder` to imperatively build up a list

```
let mut formatted = ConcatBuilder::default();

formatted.entry(formatted![....]);

...

Ok(formatted.finish())
```

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,60 +1,129 @@
...
+                        ]
+                    ]?),
+                ]
+            ]?)
+        }
+        previous_right = right;
+        current = parent_sequence;
+    }
+
+    if has_already_indentation {
```
MichaReiser, line 110: I'm not sure if I read this wrong but Prettier always seems to add the additional `indent` 

```
function test() {
	return (
		____________first,
		____________second,
		____________third,
		____________third,
		____________third,
		____________third,
		____________third
	);
}
```

Source:

```js
function test() {
return ____________first, ____________second, ____________third, ____________third, ____________third, ____________third, ____________third
}
```

ematipico: Probably a mistake. Prettier doesn't add the additional indent, I checked their playground with tabs

crates/rome_js_formatter/tests/specs/js/module/expression/sequence_expression.js.snap
```diff
@@ -15,3 +36,47 @@ Quote style: Double Quotes
...
+	____________second,
+	____________third,
+	____________third,
+	____________third,
+	____________third,
+	____________third
+);
+
+(
+	____________first,
```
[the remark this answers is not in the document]
ematipico: Yes, that's how [prettier does](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAdKACA+j3ecBmAlgE4DOMANNvrWXJFACbW20wAWpLNbun3Vn35cSPYSMG9hAsehCUQEAA4wi0MslABDEiQgB3AAq6EmlNoA2B7QE9NigEYltYANZwYAZW0BbOAAyRFBwyARW9E4u7p5eyq7BAObIMCQArnCKcL6OcExMeQHaUIlp2olwAGIQJL7aMGolyCDaaTAQCiAcML6WAOpc8GTxYHBeZkRqAG6Tts1gZA4gwfQkMEYuiXVhEZkgAFZkAB5eSZZwAIppEPA7lpEg8eRwJM2O2rmWncokwTB9RCYnGQAA4AAyKH4Qeh9FzKZo-OCrKahRQAR2u8A2KnMLTIAFoQnk8p0SHAMaQ4BtytskOF7nt6L4iCl0oyzpdMaE6btFDAPgCgRxkAAmPkuIiWJIAYQgvlpICRAFZOml6AAVD7mVIZRRTDIASWYCG8YF+qgAgswvDBbOc7vQAL6OoA)

ematipico:
Thanks @MichaReiser , indeed that soft line break was redundant. I removed it.

MichaReiser (approved):
Would you mind adding the metrics to the test plan? Looks good.

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
MichaReiser: I get it now. If it's a return, then we don't add the `indent` to each element because that already happens in the parent return.

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,7 +1,7 @@
 use crate::prelude::*;
-
-use crate::FormatNodeFields;
-use rome_js_syntax::{JsSequenceExpression, JsSequenceExpressionFields};
+use crate::{FormatElement, FormatNodeFields, FormatNodeRule, Formatter, JsFormatOptions};
```
MichaReiser, line 2: Nit: `FormatElement`, `Formatter`, `FormatNodeRule`, and `JsFormatOptions` should be part of the `prelude`

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -1,7 +1,7 @@
...
+use rome_formatter::{group_elements, ConcatBuilder, FormatResult};
```
MichaReiser, line 3: `group_elements` and `FormatResult` should already be part of the prelude.

ematipico:
> Would you mind adding the metrics to the test plan? Looks good.

Added!

Issue #2275, opened 2022-03-22
format of `JsxElement`


ematipico:

darrow-olykos:
RustPhilly will be Mob Programming on this issue on this on Monday, March 28th starting at 7:00pm EST! (RSVP to our Mob Programming events here: https://www.meetup.com/RustPhilly/)

We already have some working notes detailing how `rome_formatter` rules get executed and have a basic "happy path" implemented for `<a></a>` in our feature branch (The simplest AST to format for a JSX expression is one that needs no formatting). So if you're new to contributing to Rome, then we can help you help us help Rome (lol)!

If anyone thinks of example test cases that we can target, for example:
```rust
// before
<Foo>
    {myvar}
    </Foo>
    
// after
<Foo>{myvar}</Foo>
```
Then please feel free to recommend test cases on this thread

Link to our discord server is in the meetup linked above.

darrow-olykos:
Learning notes: https://github.com/RustPhilly/documents/blob/main/knowledge/ROME.md

darrow-olykos:
```rust
 // input
 <Foo>
     text
    </Foo>

// expected output
List [
    // TODO: define what this should look like
]
```

darrow-olykos:
2.
For this, we want to find formatter helpers or utils that more or less achieve this behavior:
  - [ ] Let's make sure a `Group(Hard)` exists around both the opening and closing Jsx Element tags, so that they remain on their own lines.
  - [ ] Let's leave the JsxChild content (which in our simple case is just JsxText) alone until we determine how we're normalizing whitespace and returns in that text
  - [ ] Let's group jsx opening tag, child content, and jsx closing tag in a `Group` to see if this is enough to get the formatted output to break onto their own lines

estaub:
The case above: 

>
> // before
> ```jsx
> <Foo>
>     text
>     </Foo>
> ```
> // after
> ```jsx
> <Foo>text</Foo>
> ```
> 

is incorrect, in that it entirely removes whitespace.  White space in html may be shrunk or expanded, but not _entirely_ added or removed.  In this example, imagine that `<Foo>` translates to `<b>`, and that there is adjacent text in the unknown surrounding context, like

```jsx
xxx<Foo>           text       </Foo>yyy
```

Space removal would change the output from "xxx **text** yyy" to "xxx**text**yyy".

estaub:
Other input:

- [a draft JSX spec from Facebook](https://facebook.github.io/jsx/)
[...]

PR #2542, opened 2022-05-05
feature(formatter): JSX Element

NicholasLYang:
## Summary

Closes https://github.com/rome/tools/issues/2275

This PR is a first pass on prettier's implementation of JSX element formatting. Prettier's algorithm is quite tricky in a few ways; it adds parentheses to JSX tag expressions in certain situations; it uses fill printing for printing jsx children when the children contain non-trivial text; it breaks on certain niche conditions.

## Test Plan

Added some tests to [element.jsx](https://github.com/rome/tools/blob/feature/jsx-element/crates/rome_js_formatter/tests/specs/jsx/element.jsx)


NicholasLYang:
Finally done haha. Took a bit to fix all the issues. Unfortunately because I now track the separator in fill, we have to create a `Fill` struct that's boxed to avoid making `FormatElement` too big. If there's an alternative, I'd be open to ideas. We can't just create a list with the separators inside, because of how fill printing is implemented.

crates/rome_formatter/src/format_element.rs
```diff
@@ -365,7 +365,7 @@ where
 /// let b = from_utf8(&[b'b'; 30]).unwrap();
 /// let c = from_utf8(&[b'c'; 30]).unwrap();
 /// let d = from_utf8(&[b'd'; 30]).unwrap();
-/// let expr = fill_elements([token(a), token(b), token(c), token(d)]);
+/// let expr = fill_elements(space_token(), [token(a), token(b), token(c), token(d)]);
```
MichaReiser, line 368: Please add another test case with a separator that prints differently in flat/expanded mode (e.g. `soft_line_break_or_space`) and verify that the printer correctly handles the cases:

* The element, separator, and next element fit on the same line
* The element and separator, but not the next element fit on the same line
* Neither the element nor the separator ,nor the next element fit on the same line.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1616,6 +1639,14 @@ impl FormatElement {
         matches!(self, FormatElement::Empty)
     }
 
+    pub fn is_empty_string(&self) -> bool {
```
MichaReiser, line 1642: Please document this new method

crates/rome_formatter/src/format_element.rs
```diff
@@ -1616,6 +1639,14 @@ impl FormatElement {
...
+        match self {
+            FormatElement::Empty => true,
```
MichaReiser, line 1644: Is this necessary because `token("")` returns `empty_element`?

I otherwise don't think the method should return `true` for `Empty` because empty means that the element has no content but it doesn't mean that it is a token. Can we move the `is_empty` check to the call site?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1631,8 +1662,10 @@ impl FormatElement {
             FormatElement::Indent(indent) => indent.content.will_break(),
             FormatElement::Group(group) => group.content.will_break(),
             FormatElement::ConditionalGroupContent(group) => group.content.will_break(),
-            FormatElement::List(list) | FormatElement::Fill(list) => {
-                list.content.iter().any(FormatElement::will_break)
+            FormatElement::List(list) => list.content.iter().any(FormatElement::will_break),
+            FormatElement::Fill(fill) => {
+                fill.list.content.iter().any(FormatElement::will_break)
+                    || fill.separator.will_break()
```
MichaReiser, line 1668: You may be able to make use of the `Fill` semantics here. 

Checking the separator shouldn't be necessary, at least not in a correctly constructed `FillElement`. `will_break` returns `true` if the element is a `HardLineBreak` or `ExpandParent`. Now, using `fill` with a hardline break or `expand parent` separator is kind of useless and should then not be used. 

But, this is certainly the safer implementation.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -1110,11 +1110,14 @@ two lines`,
...
-                soft_block_indent(format_elements![fill_elements(vec![
-                    format_elements![token("1"), token(",")],
-                    format_elements![token("2"), token(",")],
-                    format_elements![token("3"), if_group_breaks(token(","))]
-                ])]),
+                soft_block_indent(format_elements![fill_elements(
+                    token(" "),
+                    vec![
+                        format_elements![token("1"), token(",")],
+                        format_elements![token("2"), token(",")],
```
MichaReiser, line 1117: It could have made sense to split the `Fill` addition into its own PR as it is substantial on its own and there are a few different design decisions to be made.

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -562,22 +563,31 @@ pub(crate) trait JsFormatter {
         Ok(result.into_iter())
     }
 
+    /// It formats a list of nodes that are not separated. Essentially a wrapper for [format_list] that
+    /// joins the result together with [join_elements_hard_line].
```
MichaReiser, line 567: What's the meaning of *not separated*. It will join the elements with a hard line break?

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -562,22 +563,31 @@ pub(crate) trait JsFormatter {
...
     /// It formats a list of nodes that are not separated. It's an ad-hoc function to
     /// format lists that implement [rome_js_syntax::AstNodeList].
     ///
-    /// The elements of the list are joined together using [join_elements_hard_line], which will
-    /// end up separated by hard lines or empty lines.
+    /// Returns a list of syntax nodes and format elements.
     ///
     /// If the formatter fails to format an element, said element gets printed verbatim.
-    fn format_list<List, Node>(&self, list: &List) -> FormatElement
+    fn format_list<List, Node>(&self, list: &List) -> Vec<(SyntaxNode<JsLanguage>, FormatElement)>
```
MichaReiser, line 582: I think we should rename this method because its main purpose is about formatting a list with a fallback to verbatim printing

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -595,7 +605,8 @@ pub(crate) trait JsFormatter {
 
             (module_item.syntax().clone(), elem)
         });
-        join_elements_hard_line(formatted_list)
+
+        formatted_list.collect()
```
MichaReiser, line 609: Have you profiled the performance penalty of this change? The downside is that this now always requires an additional `Vec` allocation that wasn't necessary before. 

It could make sense to extract this into a `FormatListWithRecovery` struct (see `FormatDelimited`) that allows setting a mode for the line separator.

crates/rome_js_formatter/src/js/classes/static_initialization_block_class_member.rs
```diff
@@ -22,11 +22,12 @@ impl FormatNodeFields<JsStaticInitializationBlockClassMember>
         let separated = formatter
             .delimited(
                 &l_curly_token?,
-                formatter.format_list(&statements),
+                formatter.format_list_with_hard_line(&statements),
```
MichaReiser, line 25: I believe this method is used by the code gen by default. Make sure to update the code gen

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -70,7 +73,7 @@ impl FormatNodeFields<JsParenthesizedExpression> for FormatNodeRule<JsParenthesi
         // ");
         // ```
         // this is what we want
-        else if JsStringLiteralExpression::can_cast(expression.syntax().kind()) {
+        else if JsStringLiteralExpression::can_cast(kind) {
```
MichaReiser, line 76: Nit, matching on the expression directly might be faster

```
match expression {
  JsAnyExpression::JsStringLiteralExpression(...) => ,
  JsAnyExpression::JsxTagExpression(..) => 

```

crates/rome_js_formatter/src/js/lists/array_element_list.rs
```diff
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use rome_formatter::GroupId;
+use rome_formatter::{FormatResult, GroupId};
```
MichaReiser, line 2: `FormatResult` should be part of the `prelude`. If not, please add it.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
```diff
@@ -172,16 +174,24 @@ fn get_trimmed_text(
 ) -> Cow<str> {
     match (leading_whitespace_type, trailing_whitespace_type) {
         (Some(WhitespaceType::HasNewline), Some(WhitespaceType::HasNewline)) => {
-            Cow::Borrowed(text.trim())
+            Cow::Borrowed(text.trim_matches(&WHITESPACE[..]))
```
MichaReiser, line 177: Shouldn't `&WHITESPACE` work because `..` selects the whole slice?

crates/rome_js_formatter/src/jsx/tag/closing_element.rs
```diff
@@ -1,14 +1,28 @@
...
+            r_angle_token,
+        } = node.as_fields();
+
+        Ok(group_elements(formatted![
+            formatter,
+            [
+                l_angle_token.format(),
+                slash_token.format(),
+                name.format(),
+                line_suffix_boundary(),
```
MichaReiser, line 23: Not sure about this line suffix boundary. Won't this result in `<a>test</a // comment>`?

crates/rome_js_formatter/src/jsx/tag/opening_element.rs
```diff
@@ -1,14 +1,47 @@
...
+        } = node.as_fields();
+
+        if attributes.is_empty() && type_arguments.is_none() {
+            formatted![
+                formatter,
+                [
+                    l_angle_token.format(),
+                    name.format(),
+                    type_arguments.format(),
+                    line_suffix_boundary(),
```
MichaReiser, line 25: Is it necessary to have two line suffix boundaries here or should instead the formatting differ if there are any comments inside of a JSX opening element (that e.g. breaks it across multiple lines? My impression is that they are now mainly used because you discovered them and it can help to fix stability issue but they should really only be used in cases where it otherwise risks to create invalid syntax.

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+    JsFormatOptions,
+};
+use rome_formatter::{format_elements, space_token, FormatElement};
+use rome_js_syntax::kind::JsSyntaxKind;
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyLiteralExpression, JsLanguage, JsxAnyChild, JsxChildList,
+};
+use rome_rowan::SyntaxNode;
+
+pub fn jsx_space(formatter: &Formatter<JsFormatOptions>) -> FormatElement {
```
MichaReiser, line 14: Can we please add some documentation to these methods.

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub static WHITESPACE: [char; 4] = [' ', '\n', '\t', '\r'];
+
+pub fn is_meaningful_jsx_text(text: &str) -> bool {
+    let mut has_newline = false;
+    let mut has_non_whitespace = false;
+    for c in text.chars() {
+        // If there is a non-whitespace character
+        if !WHITESPACE.contains(&c) {
+            has_non_whitespace = true;
```
MichaReiser, line 63: You should be able to early return here because it is `!has_newline`

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+        if c == '\n' {
```
MichaReiser, line 65: Nit: `else if`

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -424,3 +425,20 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
             | JsAnyExpression::JsCallExpression(_)
     )
 }
+
+type PredicateFn = Box<dyn Fn(SyntaxNode<JsLanguage>) -> bool>;
```
MichaReiser, line 429: Why is it necessary that the predicate fn are boxed?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -424,3 +425,20 @@ pub(crate) fn is_call_like_expression(expression: &JsAnyExpression) -> bool {
...
+
+pub fn match_ancestors(
+    node: &SyntaxNode<JsLanguage>,
+    predicates: Vec<Option<PredicateFn>>,
```
MichaReiser, line 433: I'm not sure if I understand this predicate function. Is it a list for the `nth` ancestor? @xunilrj looked into this too. 

Please document this method

I think it would be better to pass a slice instead of a vector `&[FnOnce(&SyntaxNode<JsLanguage>)]` to avoid an allocation if the number of predicates is constant (You can then pass `&[Some(...)]`)

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Prettier defines it to be that way.
+pub fn is_jsx_inside_arrow_function_inside_call_inside_expression_child(
+    node: &SyntaxNode<JsLanguage>,
+) -> bool {
+    // ```jsx
+    //  let bar = <div>
+    //    {foo(() => <div> the quick brown fox jumps over the lazy dog </div>)}
+    //  </div>;
+    // ```
+    match_ancestors(
```
MichaReiser, line 155: The name is misleading because it isn't really matching. But it might be nice to make this a macro, I just don't know how to make it clear that it is matching the anchestor chain

```
match_anchestors!(node, [
  ..,  // To ignore
  node => node.syntax() == JsSyntaxKind::JS_...,
  node => 
]);
```

MichaReiser (changes requested):
Please add a summary and a test plan to your PR. What syntax is this PR covering? What is it not covering? How does it change our metric, etc.

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/utils/mod.rs
leops: These functions have to be boxed to be stored in a `Vec` since each entry into the array is a closure with its own unique type, however if these closure do not capture any context they could be simplified to `fn()` function pointers instead of boxed `Fn()` traits

crates/rome_js_formatter/tests/specs/prettier/js/ternaries/nested-in-condition.js.snap
```diff
@@ -68,22 +69,14 @@ const value = (
...
-    <Sub />
-  </Element>
-) : (
-  <Element2>
-    <Sub />
-    <Sub />
-    <Sub />
-  </Element2>
-);
+) ? <Element>
```
leops, line 72: Shouldn't this be parenthesized ?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1175,10 +1181,11 @@ impl Debug for FormatElement {
                 write!(fmt, "List ")?;
                 content.fmt(fmt)
             }
-            FormatElement::Fill(content) => {
-                write!(fmt, "Fill ")?;
-                content.fmt(fmt)
-            }
+            FormatElement::Fill(fill) => fmt
```
leops, line 1184: Nit: derive `Debug` on `Fill` instead

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+    false
+}
+
+pub enum WrapState {
+    NoWrap,
+    WrapOnBreak,
+    AlwaysWrap,
+}
+
+/// Takes in a syntax node because we need to handle both JsxElement and JsxSelfClosingElement
```
ematipico, line 108: This is not the kind of documentation we want to write 😅 
I think this was a normal comment? 

Maybe the function should be name `should_wrap_jsx_element_in_parens`? As it should handle only JSX elements

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+pub fn should_wrap_element_in_parens(node: &SyntaxNode<JsLanguage>) -> WrapState {
+    let mut ancestors = node.ancestors();
+    // We skip 1 for the current node
```
ematipico, line 111: Could you add the "why" we skip the first one? The comment is not useful like this.
Can't we use `skip` instead? It seems more appropriate

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+    ancestors.next();
+
+    ancestors
+        .next()
```
ematipico, line 115: Why are we skipping also the second node too? Should we have a comment explaining it?

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    false
+}
+
+pub enum WrapState {
+    NoWrap,
+    WrapOnBreak,
+    AlwaysWrap,
+}
```
ematipico, line 106: How does this state related to `should_wrap_element_in_parens`? In my head "should wrap" seems more a boolean (two options) but instead we have three options. It would be great to have some doc to explain each variant.

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+                | JsSyntaxKind::JS_COMPUTED_MEMBER_EXPRESSION => WrapState::AlwaysWrap,
+                _ => WrapState::WrapOnBreak,
+            }
+        })
+        .unwrap_or(WrapState::NoWrap)
+}
+
+/// This is a very special situation where we're returning a JsxElement
+/// from an arrow function that's passed as an argument to a function,
+/// which is itself inside a JSX expression child.
```
ematipico, line 143: Maybe a small snippet might help the developer to understand what's described here. (I have difficulties to visualize the case).

We could move the snippet that is the function inside here?

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+                | JsxAnyChild::JsxFragment(_)
+        ) {
+            return true;
+        }
+    }
+
+    false
+}
+
+pub fn contains_multiple_expressions(children: &JsxChildList) -> bool {
```
ematipico, line 88: `contains_at_least_two_expressions`? Because one expression is considered "multiple"

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+            child,
+            JsxAnyChild::JsxElement(_)
+                | JsxAnyChild::JsxSelfClosingElement(_)
+                | JsxAnyChild::JsxFragment(_)
+        ) {
+            return true;
+        }
+    }
+
+    false
```
ematipico, line 85: You could use `any` here.

```rs
children.iter().any(|child| {
	...
})
```

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+            if let Ok(token) = string_literal_expr.value_token() {
+                return token.text() == " ";
+            }
+        }
+    }
+
+    false
+}
+
+pub fn contains_text(children: &JsxChildList) -> bool {
```
ematipico, line 41: `contains_meaningful_jsx_text`?

You could use `Iter::any`

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+        QuoteStyle::Single => "{{\' \'}}",
+    };
+
+    format_elements![
+        if_group_breaks(format_elements![token(jsx_space), soft_line_break()]),
+        if_group_fits_on_single_line(space_token())
+    ]
+}
+
+pub fn is_jsx_whitespace_expression(child: JsxAnyChild) -> bool {
```
ematipico, line 26: It might help to have some documentation here. What's a "whitespace expression"?

crates/rome_formatter/src/format_element.rs
```diff
@@ -1278,6 +1285,22 @@ impl Deref for List {
     }
 }
 
+#[derive(Clone, PartialEq, Eq)]
+pub struct Fill {
```
ematipico, line 1289: Could you document this new struct, please?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -325,9 +325,9 @@ impl<'a> Printer<'a> {
         &mut self,
         queue: &mut ElementCallQueue<'a>,
         content: &'a List,
+        separator: &'a FormatElement,
         args: PrintElementArgs,
```
ematipico, line 329: I wonder if this should be a `struct` now? We have too many arguments and having a struct would allow us to document each argument.

ematipico (changes requested):
Could you please create a separate PR with the `Fill` changes? They are difficult to review, together with JSX elements changes.

crates/rome_formatter/src/printer/mod.rs
MichaReiser: In my view, this is still fine. `self` is implicit and all arguments have different types.

crates/rome_formatter/src/format_element.rs
NicholasLYang: Hmm actually that raises a good point. What should a separator that prints differently do in a fill? Because if we're breaking using the fill, it wouldn't make sense to also break with the separator, no?

crates/rome_js_formatter/src/utils/jsx_utils.rs
NicholasLYang: I'm confused. When are we skipping the second node?

crates/rome_js_formatter/src/formatter.rs
NicholasLYang: I don't actually know. I copied this from the original `format_list` doc comment.

crates/rome_js_formatter/src/formatter.rs
NicholasLYang: I thought about this and yeah, that could be an issue, although I suspect rustc might be smart enough to realize that we're collecting a vector that we immediately iterate.

crates/rome_js_formatter/src/jsx/auxiliary/text.rs
NicholasLYang: It doesn't work unfortunately:

```
error[E0277]: the trait bound `CharArrayRefSearcher<'_, '_, 4_usize>: DoubleEndedSearcher<'_>` is not satisfied
    --> crates/rome_js_formatter/src/jsx/auxiliary/text.rs:187:45
     |
187  |             Cow::Owned(format!(" {} ", text.trim_matches(&WHITESPACE)))
     |                                             ^^^^^^^^^^^^ the trait `DoubleEndedSearcher<'_>` is not implemented for `CharArrayRefSearcher<'_, '_, 4_usize>`
     |
note: required by a bound in `core::str::<impl str>::trim_matches`
    --> /Users/nicholas/.rustup/toolchains/1.61.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/str/mod.rs:2052:34
     |
2052 |         P: Pattern<'a, Searcher: DoubleEndedSearcher<'a>>,
     |                                  ^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `core::str::<impl str>::trim_matches`
```

crates/rome_js_formatter/src/jsx/tag/closing_element.rs
NicholasLYang: Won't it force a line break after the comment? So we'd get:
```
<a>
  test
</a \\ comment
>
```

crates/rome_js_formatter/src/jsx/tag/opening_element.rs
NicholasLYang: Well if we're following prettier, the line suffix boundaries [make sense](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&isTypeScript=false&isJsx=true#bABlAHQAIABhACAAPQAgADwAIAAvAC8AIAB0AGUAcwB0AGkAbgBnAAoAZABpAHYAIAAvAC8AIAB0AGUAcwB0AGkAbgBnACAAMgAKAD4AIAA8AC8AZABpAHYAPgA=). I'm open to other ways, but I'm not really familiar with how we could format differently that wouldn't just emulate a line suffix boundary.

crates/rome_js_formatter/src/utils/jsx_utils.rs
ematipico: Sorry, my mistake!

crates/rome_js_formatter/src/formatter.rs
MichaReiser: I'm not so sure about this if you look at the latest benchmark results. It would be best to bench this separately to narrow it down if the regressions are because of this change or not.

crates/rome_js_formatter/src/jsx/lists/child_list.rs
```diff
@@ -11,6 +13,21 @@ impl FormatRule<JsxChildList> for FormatJsxChildList {
         node: &JsxChildList,
         formatter: &Formatter<JsFormatOptions>,
     ) -> FormatResult<FormatElement> {
-        verbatim_node(node.syntax()).format(formatter)
+        let children = formatter
+            .format_list(node)
```
MichaReiser, line 17: Do you want to get the error recovery here. You can otherwise just use `format_all`

```
formatter.format_all(node.iter().formatted())
```

crates/rome_js_formatter/src/utils/jsx_utils.rs
NicholasLYang: Ehh the `Iter::any` version is not that much less verbose
```
    children.iter().any(|child| {
        if let JsxAnyChild::JsxText(jsx_text) = child {
            if let Ok(token) = jsx_text.value_token() {
                if is_meaningful_jsx_text(token.text()) {
                    return true;
                }
            }
        }

        false
    })
```

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,193 @@
...
+                    return true;
+                }
+            }
+        }
+    }
+
+    false
+}
+
+pub static WHITESPACE: [char; 4] = [' ', '\n', '\t', '\r'];
```
ematipico, line 68: Did you mean to add a `const` too? Also, we don't need the `pub`, maybe `pub(crate)`? Also, it seems that this array doesn't contain only white spaces. Maybe we should change its name. Something like `NON_MEANINGFUL_CHARS`?

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,193 @@
...
+    !has_newline
+}
+
+pub enum WrapState {
+    NoWrap,
+    WrapOnBreak,
+    AlwaysWrap,
+}
+
+/// Checks if a JSX Element should be wrapped in parentheses. Returns a WrapState which
```
ematipico, line 102:
```suggestion
/// Checks if a JSX Element should be wrapped in parentheses. Returns a [WrapState] which
```

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,193 @@
...
+/// indicates that an element should always be wrapped in parentheses, should be wrapped
+/// only when it's line broken, or should not be wrapped at all.
```
ematipico, line 104: "it's line broken", is it grammatically correct?

crates/rome_js_formatter/src/utils/jsx_utils.rs
ematipico: Sure, but please highlight each option and example with each variant, so it's easier to spot for future developers.

crates/rome_js_formatter/tests/specs/jsx/element.jsx.snap
```diff
@@ -0,0 +1,287 @@
...
+			<TabPanel>
+				<pre className="h-screen overflow-y-scroll">{formatter_ir}</pre>
+			</TabPanel>
+			<TabPanel>
+				<pre className="h-screen overflow-y-scroll">{prettierOutput.ir}</pre>
+			</TabPanel>
+			<TabPanel>
+				<pre
+					className="h-screen overflow-y-scroll whitespace-pre-wrap text-red-500 text-xs"
+				>
```
ematipico, line 235: I am not sure this is correct. I am trying to [replicate in prettier](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=1&typescript=false&jsx=true#CQAJAAkAPABUAGEAYgBQAGEAbgBlAGwAPgAJAAkACQA8AFQAYQBiAFAAYQBuAGUAbAA+AAkACQAJADwAVABhAGIAUABhAG4AZQBsAD4ACQAJAAkAPABUAGEAYgBQAGEAbgBlAGwAPgA8AFQAYQBiAFAAYQBuAGUAbAA+ADwAVABhAGIAUABhAG4AZQBsAD4ACgAJAAkACQAJADwAcAByAGUACgAJAAkACQAJAAkAYwBsAGEAcwBzAE4AYQBtAGUAPQAiAGgALQBzAGMAcgBlAGUAbgAgAGYAZQBlAGYAZgBlAGYAZQBmAGUAZgBlAGYAZQAgAGUAZgAgAGYAZQAgAGYAZQAgAGUAZgAgAGYAZQBmAGUAZgBlAGYAZQBmAGUAIABmAGUAIABmAGUAIABmAGUAIABmAGUAIABmAGUAIABvAHYAZQByAGYAbABvAHcALQB5AC0AcwBjAHIAbwBsAGwAIAB3AGgAaQB0AGUAcwBwAGEAYwBlAC0AcAByAGUALQB3AHIAYQBwACAAdABlAHgAdAAtAHIAZQBkAC0ANQAwADAAIAB0AGUAeAB0AC0AeABzACIACgAJAAkACQAJAD4ACgAJAAkACQAJAAkAewBlAHIAcgBvAHIAcwB9AAoACQAJAAkACQA8AC8AcAByAGUAPgAKAAkACQAJADwALwBUAGEAYgBQAGEAbgBlAGwAPgAJAAkACQA8AC8AVABhAGIAUABhAG4AZQBsAD4ACQAJAAkAPAAvAFQAYQBiAFAAYQBuAGUAbAA+AAkACQAJADwALwBUAGEAYgBQAGEAbgBlAGwAPgAJAAkACQA8AC8AVABhAGIAUABhAG4AZQBsAD4ACQAJAAkAPAAvAFQAYQBiAFAAYQBuAGUAbAA+AA==), but it seems that prettier doesn't break the attribute on a newline, even when there's a high level of indentation.

crates/rome_js_formatter/tests/specs/jsx/element.jsx.snap
```diff
@@ -0,0 +1,287 @@
...
+			</li><li>1 tablespoon minced ginger </li><li>
+				2 tablespoons doubanjiang
+			</li><li>1 tablespoon douchi</li><li>
+				1 tablespoon corn or potato starch
+			</li><li>2 scallions or jiu cai</li><li>
+				6 ounces of ground beef or pork
+			</li>
+		</ul>
+	);
+}
```
ematipico, line 277: This involves `JsxChildList`. We should remove it. In fact, the formatting is not correct

crates/rome_js_formatter/src/formatter.rs
MichaReiser: My refactor will solve this for you. Because you can then build the lists yourself and the logic for gracefully recover if the formatting of a node fails is a single helper function. So you may just want till after the refactor. 

But the other question is. Do you need / want the error recovery? We so far mainly applied it in situations where we are sure that we don't introduce new syntax errors and I'm not sure if that's the case when formatting JSX.

crates/rome_js_formatter/src/utils/jsx_utils.rs
NicholasLYang: What values are non-whitespace? If anything it's a subset of characters that are considered whitespace. In which case perhaps `JSX_WHITESPACE` would make sense. `NON_MEANINGFUL_CHARS` is not quite right because the characters can be meaningful if they do not contain a newline.

crates/rome_js_formatter/src/utils/jsx_utils.rs
NicholasLYang: I think so? "line broken" here is an adjective, although if it's unclear, I can certainly change it

crates/rome_js_formatter/src/jsx/tag/opening_element.rs
```diff
@@ -1,14 +1,47 @@
...
+                    attributes.format(),
+                    line_suffix_boundary(),
```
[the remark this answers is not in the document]
NicholasLYang: They handle single line comments inside an opening element. It's important that they get printed here, as otherwise they might get printed inside the body.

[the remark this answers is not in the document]
ematipico: Could we add a test case so we are covered and we know it's doing what it's supposed to?

crates/rome_js_formatter/tests/specs/jsx/element.jsx.snap
ematipico: Agree to disagree then. 

In my view, the introduced test is not correct. Unless it's explicitly stated somewhere (PR description, comment, TODO, etc.) that this code is there for a reason, then I can't see a reason to keep it.

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,146 @@
...
+/// This is a very special situation where we're returning a JsxElement
+/// from an arrow function that's passed as an argument to a function,
+/// which is itself inside a JSX expression child.
+///
+/// If you're wondering why this is the only other case, it's because
+/// Prettier defines it to be that way.
+///
+/// ```jsx
+///  let bar = <div>
+///    {foo(() => <div> the quick brown fox jumps over the lazy dog </div>)}
```
ematipico, line 120: Could we add a test case for this snippet?

crates/rome_js_formatter/src/utils/jsx_utils.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// should always be wrapped:
+    /// ```jsx
+    /// (<div>Badlands</div>).property
+    /// ```
+    AlwaysWrap,
+}
+
+/// Checks if a JSX Element should be wrapped in parentheses. Returns a [WrapState] which
+/// indicates when the element should be wrapped in parentheses.
+pub fn get_wrap_state(node: &SyntaxNode<JsLanguage>) -> WrapState {
```
ematipico, line 82: Could we add some test case that cover these wrap states? The snippets in the documentation are really helpful, but we should back them with some test.

ematipico (changes requested):
Almost there. We should add some test case for the edge cases covered in this PR (line suffix boundaries and `WrapState`)

crates/rome_js_formatter/src/utils/jsx_utils.rs
NicholasLYang: Added in `element.jsx`

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
```diff
@@ -1,10 +1,54 @@
...
+        if let Some(JsAnyExpression::JsAnyLiteralExpression(
+            JsAnyLiteralExpression::JsStringLiteralExpression(string_literal),
+        )) = &expression
+        {
+            if let Ok(str_token) = string_literal.value_token() {
+                let quote_style = f.context().quote_style();
+                let space_str = match quote_style {
+                    QuoteStyle::Double => "\" \"",
+                    QuoteStyle::Single => "' '",
+                };
```
MichaReiser, line 29: This should work regardless if the programmer used the preferred quotes or not [Prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEBDTBeTAHmwD5gByTcgX0IHpSAdKdLAI3yNOEZE19oMSIADQgIABxgBLaAGdkobACdlEAO4AFFQgUpsqddgCeCsW2XYwAawwBlbAFs4AGWlQ4yAGYG5cc5Y29hJW7gDmyDDKAK7+IHCObHAAJskpLthQYdHYYXAAYhDKjtgwMlnIINjRMBCiIAAWMI6oAOoN0vByIWBwdrqd0gBuncaVYHJmIO5+yjCalmEl3r5xAFZyAB524egAitEQ8CuofmIhyrOVMMYScHJgytJS9RJPsK3SyTANyAAcAAZzmo-K1LBJKm97nBlENPGIAI6HeALSR6KpyAC0HhSKXqyjgSOkBIWuWWSB8pzifkc0kiMWpuzgByOngpqzEMGwbE+31+SAATJzLNJUOEAMIQRzk+JyACs9WifgAKty9FFYmIhrEAJJQNKwOyPZ4wACC+rsN3QJz81GoQA)

`JsStringLiteralExpression` has a `inner_string_text` method (@leops is changing it to return a string slice instead) that gives you the quoted text. You can then test if the inner text matches the empty space.

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
```diff
@@ -1,10 +1,54 @@
...
+
+                if str_token.text() == space_str {
+                    let l_curly_token = l_curly_token?;
+                    let r_curly_token = r_curly_token?;
+                    let empty = empty_element();
+                    let l_curly = format_replaced(&l_curly_token, &empty);
+                    let space = format_replaced(&str_token, &empty);
+                    let r_curly = format_replaced(&r_curly_token, &empty);
```
MichaReiser, line 37: I got confused why the final output contains the `space` and the `JsxSpace::default`. I think it would be easier to understand if you rewrite this to 

```
let space = JsxSpace::default();
format_replaced(&str_token, &space); 
```

You can avoid all the temporary variables by inlining the arguments directly inside the `write` call

```
return write!(f, [
  format_replaced(&l_curly_token, &empty_element()), 
  format_replaced(&str_token, &JsxSpace::default()),
  format_repalced(&r_curly_token, &empty_element())
]);
```

crates/rome_js_formatter/src/jsx/auxiliary/space.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn fmt(&self, formatter: &mut JsFormatter) -> FormatResult<()> {
+        let jsx_space = match formatter.context().quote_style() {
+            QuoteStyle::Double => "{\" \"}",
+            QuoteStyle::Single => "{\' \'}",
+        };
+
+        write![
+            formatter,
+            [
+                if_group_breaks(&format_args![token(jsx_space), soft_line_break()]),
```
MichaReiser, line 34: Nit: You can use a hardline_break here. You already know that the group will break.

crates/rome_js_formatter/src/jsx/tag/opening_element.rs
```diff
@@ -1,10 +1,45 @@
...
+            r_angle_token,
+        } = node.as_fields();
+
+        if attributes.is_empty() && type_arguments.is_none() {
+            write![
+                f,
+                [
+                    l_angle_token.format(),
+                    name.format(),
+                    type_arguments.format(),
```
MichaReiser, line 22: What's the reason you're writing the `type_arguments` and `attributes` if we know that they are `None` / `empty`? 

Overall, I get the impression that it should be possible to simplify this a great deal because the only conditional content seems to be that there must be a `space_token` in front of `attributes` if `attributes`  and `type_arguments` aren't empty

```
write!(f, [
  l_angle_token.format(),
  name.format(),
  type_arguments.format(),
  line_suffix_boundary(),
])?;

if let Some(attributes) = attributes {
    write!(f, [space_token(), attributes, line_suffix_boundary()])?;
}

write!(f, [
    r_angle_token.format()
])
```

MichaReiser (commented):
Nice work! I would like to better understand why the change inside of the `Printer` is necessary before merging and have a few proposal on how the code could be simplified.

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
```diff
@@ -1,10 +1,54 @@
...
+        )) = &expression
+        {
+            if let Ok(str_token) = string_literal.value_token() {
+                let quote_style = f.context().quote_style();
+                let space_str = match quote_style {
+                    QuoteStyle::Double => "\" \"",
+                    QuoteStyle::Single => "' '",
+                };
+
+                if str_token.text() == space_str {
```
MichaReiser, line 31: I think we need to add a check here to test that the string literal has no comments or skipped trivia attached because comments aren't valid inside of JSX text.

crates/rome_formatter/src/builders.rs
```diff
@@ -1638,6 +1638,43 @@ impl<'a, 'buf, Context> FillBuilder<'a, 'buf, Context> {
...
+    }
+
+    /// Adds a new entry to the fill output. If the entry is a [FormatElement::List],
+    /// then adds the list's entries to the fill output instead of the list itself.
+    pub fn flatten_entry(&mut self, entry: &dyn Format<Context>) -> &mut Self {
+        self.result = self.result.and_then(|_| {
+            let mut buffer = VecBuffer::new(self.fmt.state_mut());
+            write!(buffer, [entry])?;
+
+            let item = buffer.into_element();
```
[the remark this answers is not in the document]
ematipico: So I tried to apply your suggestion, but we incur into a regression:

```diff
+`"let array = [1 , 2 ];\n    }\n\n    function func2() {\n        const no_format = () => {};\n    }"`,
-`"let array = [1, 2];\n    }\n\n    function func2() {\n        const no_format = () => {};\n    }"`',
```

The red was is **before** and the green one is **after** applying your suggestion

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
ematipico: It seems that this method `inner_string_text` still returns something different and the methods cover only chars but not strings. For now  I will keep the current implementation

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
ematipico: Could you add a test to cover you suggestion? This would help to understand what I am supposed to do

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -350,14 +350,30 @@ impl<'a> Printer<'a> {
...
+
+            current_content = match items.next() {
+                None => {
+                    // Empty list
+                    return;
+                }
+                Some(item) => item,
+            };
+        }
+
```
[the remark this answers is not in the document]
ematipico: You were right, I removing this code doesn't trigger any regression

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
ematipico: Does the new code address your feedback?

crates/rome_formatter/src/builders.rs
```diff
@@ -1638,6 +1638,43 @@ impl<'a, 'buf, Context> FillBuilder<'a, 'buf, Context> {
         self
     }
 
+    /// Adds an iterator of entries to the fill output, flattening any [FormatElement::List]
+    /// entries by adding the list's elements to the fill output.
+    pub fn flatten_entries<F, I>(&mut self, entries: I) -> &mut Self
```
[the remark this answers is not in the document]
ematipico: What's your suggestion? What's the best course of action?

ematipico:
@MichaReiser I addressed some of you comments. If you're able to provide more info on the others, I might be able to implement them and merge this PR.

ematipico:
!bench_formatter

crates/rome_formatter/src/builders.rs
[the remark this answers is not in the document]
MichaReiser: This doesn't make sense to me. `into_vec` returns a list with the elements and all that line 1664-1668 does is adding all elements while flattening lists. 

You'll have to do some debugging here. I recommend logging `self.items` before/after the change.

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
MichaReiser: I see. Changing that would be good at some time.

But the feedback that this should also work regardless of the used quote still remains.

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
MichaReiser:
```
{ /* comment */ " " }
{
  " "
  /* comment */
}
```

Comments associated with the token itself shouldn't be an issue because `text` includes` the trivia. Ideally, we wouldn't just test for comments but also skipped trivia (I'm unable to come up with an example)

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
MichaReiser: Can we use `format_replaced(&str_token, &JsxSpace::default())`? It has slightly better comment placement.

crates/rome_formatter/src/builders.rs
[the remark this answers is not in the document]
MichaReiser: Hard to say. I'm not very familiar with this is now implemented. My main concern is that this is hard to reason about. The alternative would be to do something similar to binarish (and assignment etc) where we first map the children into a intermediate struct depending on the ways we want to format the entries and then do the formatting.

However, I don't think we should spend too much time on this right now. But could we add some documentation to discourage the use of this method?

crates/rome_js_formatter/src/jsx/auxiliary/expression_child.rs
ematipico: Yes, that had been applied and a test added

MichaReiser (approved):

PR #2543, opened 2022-05-05
feature(playground): Prettier IR tab

NicholasLYang:
## Summary

I found myself wanting to compare with how Prettier represents the code in their IR. Turns out Prettier smartly exposes a function that lets you convert code to their IR.

After the JSON serialization PR is merged, I'll move this to a React JSON View

## Test Plan

Check out the playground link and play around with it

### DIFF
--- a/website/playground/src/utils.ts
+++ b/website/playground/src/utils.ts
@@ -112,19 +112,21 @@ export function formatWithPrettier(
 	},
 ) {
 	try {
-		return prettier.format(
-			code,
-			{
-				useTabs: options.indentStyle === IndentStyle.Tab,
-				tabWidth: options.indentWidth,
-				printWidth: options.lineWidth,
-				parser: getPrettierParser(options.language),
-				plugins: [parserBabel],
-				singleQuote: options.quoteStyle === QuoteStyle.Single,
-			},
-		);
+		const prettierOptions = {
+			useTabs: options.indentStyle === IndentStyle.Tab,
+			tabWidth: options.indentWidth,
+			printWidth: options.lineWidth,
+			parser: getPrettierParser(options.language),
+			plugins: [parserBabel],
+			singleQuote: options.quoteStyle === QuoteStyle.Single,
+		};
+		const formattedCode = prettier.format(code, prettierOptions);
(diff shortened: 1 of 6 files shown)


MichaReiser (commented):
So many tabs ;) 

Prettier's playground also offers an option to show prettier's document (IR). 

How to enable Prettier's IR

I'm not sure where we want to draw the line but in my view showing Prettier's IR is something where we can direct people to use Prettier's playground instead. 

I'm also a bit worried that we're starting to do more and more computation with every keystroke. Prettier now formats each file twice, and the output get's stringified regardless if the tab is visible or not. Have you tested how good the performance is if you test the playground with a 4 times slowdown of the CPU?

NicholasLYang:
We can move the computation into the tab! I haven't tested performance but I suspect for small inputs it's really not that bad. If necessary I can debounce the input.

I mostly wanted prettier's IR because as we start to move the IR closer to Prettier's, it might be good to compare outputs. I know I've been struggling to wrap my head around Prettier's logic for JSX.

ematipico:
> We can move the computation into the tab! I haven't tested performance but I suspect for small inputs it's really not that bad. If necessary I can debounce the input.
> 
> I mostly wanted prettier's IR because as we start to move the IR closer to Prettier's, it might be good to compare outputs. I know I've been struggling to wrap my head around Prettier's logic for JSX.

I share the sentiment, I found myself checking their IRs quite often recenlty

website/playground/src/utils.ts
```diff
@@ -112,19 +112,21 @@ export function formatWithPrettier(
...
+			plugins: [parserBabel],
+			singleQuote: options.quoteStyle === QuoteStyle.Single,
+		};
+		const formattedCode = prettier.format(code, prettierOptions);
+		//@ts-ignore
+		const ir = prettier.__debug.printToDoc(code, prettierOptions);
+		return { code: formattedCode, ir };
 	} catch (err) {
-		return code;
+		console.error(err);
```
MichaReiser, line 128: How about returning it as the `code` instead maybe included in some message?

MichaReiser (approved):

PR #2544, opened 2022-05-06
Revert "fix(rome_js_formatter): elide quotes inside members"

ematipico:
Reverts rome/tools#2502


ematipico:
This PR was overlooked from all of us and it  breaks typescript programs. We revert it and come back to it when the formatter has knowledge of the source file it's formatting.

xunilrj (approved):

PR #2547, opened 2022-05-06
fix(rome_js_formatter): 🐛 wrap some string expression stmt

### LINKED ISSUE #2436: Wrap string literal statements in parentheses
Prettier wraps string literals appearing in statement positions in parentheses to ensure they don't become directives if they happen to be at the top of the program [source](https://github.com/prettier/prettier/issues/7346#issuecomment-574823604). Rome should do the same to have consistent string literal formatting.  [Playground](https://play.rome.tools/?lineWidth=80&indentStyle=space&quoteStyle=double&indentWidth=2&typescript=true&jsx=false&sourceType=module#KAAnAGIAYQByACcAKQA7ACAALwAvACAAcABhAHIAZQBuAHMAIABzAGgAbwB1AGwAZAAgAG4AbwB0ACAAYgBlACAAcgBlAG0AbwB2AGUAZAAgAHQAbwAgAGEAdgBvAGkAZAAgAGIAZQBjAG8AbQBpAG4AZwAgAGEAIABkAGkAcgBlAGMAdABpAHYAZQAKAGAAZgBvAG8AYAA7AAoAJwBiAGEAcgAnADsAIAAvAC8AIABwAGEAcgBlAG4AcwAgAHMAaABvAHUAbABkACAAYgBlACAAYQBkAGQAZQBkACwAIABzAGUAZQAgAGgAdAB0AHAAcwA6AC8ALwBnAGkAdABoAHUAYgAuAGMAbwBtAC8AcAByAGUAdAB0AGkAZQByAC8AcAByAGUAdAB0AGkAZQByAC8AaQBzAHMAdQBlAHMALwA3ADMANAA2ACMAaQBzAHMAdQBlAGMAbwBtAG0AZQBuAHQALQA1ADcANAA4ADIAMwA2ADAANAAKACcAIgAnADsACgA=)

## Input

```ts
('bar'); // parens should not be removed to avoid becoming a directive
`foo`;
'bar'; // parens should be added, see https://github.com/prettier/prettier/issues/7346#issuecomment-574823604
'"';
```

## Prettier

```ts
("bar"); // parens should not be removed to avoid becoming a directive
`foo`;
("bar"); // parens should be added, see https://github.com/prettier/prettier/issues/7346#issuecomment-574823604
('"');
```

## Rome

```ts
("bar"); // parens should not be removed to avoid becoming a directive
`foo`;
"bar"; // parens should be added, see https://github.com/prettier/prettier/issues/7346#issuecomment-574823604
'"';
```

## Expected

Rome's formatting to match Prettier's

IWANABETHATGUY:
## Summary
fix #2436

## Test Plan
The changed snapshot should match prettier, some controversial testcase I have been commented reason.


IWANABETHATGUY:
!bench_formatter

crates/rome_js_formatter/tests/specs/prettier/js/exports/test.js.snap
```diff
@@ -1,8 +1,7 @@
 ---
```
IWANABETHATGUY, line 1: influenced by this pr, because input is invalid syntax,  and the CST matched the target formatted rule

crates/rome_js_formatter/tests/specs/prettier/js/export-extension/export.js.snap
```diff
@@ -1,5 +1,6 @@
 ---
```
IWANABETHATGUY, line 1: influenced by this pr, because input is invalid syntax, and the CST matched the target formatted rule

crates/rome_js_formatter/tests/specs/prettier/js/export-default/escaped/default-escaped.js.snap
```diff
@@ -1,8 +1,7 @@
 ---
```
IWANABETHATGUY, line 1: influenced by this pr, because input is invalid syntax, and the CST matched the target formatted rule

crates/rome_js_formatter/tests/specs/prettier/js/babel-plugins/export-default-from.js.snap
```diff
@@ -1,8 +1,7 @@
 ---
```
IWANABETHATGUY, line 1: influenced by this pr, because input is invalid syntax, and the CST matched the target formatted rule

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -41,6 +41,12 @@ impl FormatNode for JsParenthesizedExpression {
                 group_elements(expression.format(formatter)?),
                 formatter.format_replaced(&r_paren_token?, empty_element()),
             ])
+        } else if JsStringLiteralExpression::can_cast(expression.clone()?.syntax().kind()) {
```
ematipico, line 44: You don't need to `.clone()` here, I think

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -614,6 +614,7 @@ pub(crate) fn format_with_semicolon(
 pub(crate) fn format_string_literal_token(
     token: JsSyntaxToken,
     formatter: &Formatter,
+    needs_parenthesis: bool,
```
MichaReiser, line 617: We shouldn't have to do the formatting inside `format_string_literal_token`. It should be sufficient to implement the wrapping logic for`JsStringLiteralExpression`.

crates/rome_js_formatter/src/js/expressions/string_literal_expression.rs
```diff
@@ -10,7 +12,24 @@ impl FormatNode for JsStringLiteralExpression {
...
+            if let Some(JsSyntaxKind::JS_EXPRESSION_STATEMENT) = parent.clone().map(|p| p.kind()) {
+                // SAFETY: since parent's `kind` is `JsSyntaxKind::JS_EXPRESSION_STATEMENT`, it would not be `None`.
+                let parent_parent = parent.unwrap().parent();
+                matches!(
+                    parent_parent.map(|p| p.kind()),
+                    Some(JsSyntaxKind::JS_BLOCK_STATEMENT | JsSyntaxKind::JS_MODULE_ITEM_LIST)
+                )
+            } else {
+                false
+            };
```
ematipico, line 28: I think this can be written like this:

```rs
        let parent = self.syntax().parent();

        let needs_parenthesis =
            if let Some(expression_statement) = parent.and_then(JsExpressionStatement::cast) {
                let great_parent_kind = expression_statement.syntax().parent().map(|p| p.kind());
                matches!(
                    great_parent_kind,
                    Some(JsSyntaxKind::JS_BLOCK_STATEMENT | JsSyntaxKind::JS_MODULE_ITEM_LIST)
                )
            } else {
                false
            };

```

MichaReiser, line 28: Prettier wraps all string literal expressions in parenthesis, regardless if they are in a block statement or module item list. 

> This is intentional. Prettier wraps any string literal which is in statement position in parentheses because otherwise such strings would get interpreted as directives if they occurred at the top of a function or program, which can change the behavior of the program.

and [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAdKAWABAaiwZgG50SR4BnGVEQkAGhAgAcYBLac5UAQwCdeIAdwAKfBJxTcANoO4BPTgwBGvbmADWcGAGVuAWzgAZVlDjIAZtPJxlqjVu1M1JgObIYvAK42QcPUrgAE0Cgw24oF09uFzgAMQhePW4YNgjkEG5PGAh6EAALGD0pAHU81gonMDhtcXLWADdyuXSwckUQE2teGGFVFySLKx8AK3IAD21XKTgARU8IeEGpawYnXi70pW4AqVymXhMYYtZAmDzkAA4ABlWBa2LVJnT9uC76swYAR3n4XuYJDLkAC0piCQVyvDg31YkN60QGSEsyx81j0rHcXhRU1mPzMiKGDBg22Op3OSAATITVKwpK4AMIQPQI3zkACsuU81gAKtsJEiViB6t4AJJQEKwbRgA4sACCYu0MDk0yW1gAvqqgA)

So this can be simplified to just checking if the parent is a `JsExpressionStatement`. 
```suggestion
            let needs_parenthesis = JsExpressionStatement::can_cast(self.syntax().parent());
```

MichaReiser (changes requested):

crates/rome_js_formatter/src/utils/mod.rs
ematipico, line 617: Could we use an `enum` instead, please? We try to stay away from boolean

crates/rome_js_formatter/src/js/expressions/string_literal_expression.rs
IWANABETHATGUY: If I don't check `JsExpressionStatement`, these testcase would also change, 
image
image
image
1. crates/rome_js_formatter/tests/specs/prettier/js/do/do.js.snap
2. crates/rome_js_formatter/tests/specs/prettier/js/decorators/multiple.js.snap
3. crates/rome_js_formatter/tests/specs/prettier/js/record/spread.js.snap

IWANABETHATGUY: Do you think that was alright?

IWANABETHATGUY: although, these test cases are syntax invalid indeed

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
IWANABETHATGUY: Nope, image

MichaReiser: That should work because `syntax` takes `expression` by reference: `JsStringLiteralExpression::can_cast(expression.syntax().kind())`

crates/rome_js_formatter/src/utils/mod.rs
IWANABETHATGUY: > We shouldn't have to do the formatting inside `format_string_literal_token`. It should be sufficient to implement the wrapping logic for`JsStringLiteralExpression`.

Only handle logic in `JsStringLiteralExpression` is not enough when `JsStringLiteralExpression` has trailing comments 
```diff
 107   107 │ // Prevent strings from being parsed as directives
  108   108 │ // See https://github.com/prettier/prettier/pull/1560#issue-227225960
  109   109 │ 0;
  110   110 │ 
        111 │+(
  111   112 │ // Every string will be changed to double quotes, unless we end up with fewer
  112   113 │ // escaped quotes by using single quotes. (Vice versa if the "singleQuote"
  113   114 │ // option is true).
  114   115 │ //
┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
  116   117 │ // still processed in order to remove unnecessarily escaped quotes within it,
  117   118 │ // for consistency.
  118   119 │ 
  119   120 │ // Simple strings.
        121 │+"abc");
  120   122 │ ("abc");
  121       │-("abc");
  122   123 │ 
        124 │+(
  123   125 │ // Escape.
  124       │-("\0");
        126 │+"\0");
  125   127 │ 
        128 │+(
  126   129 │ // Emoji.
  127       │-("🐶");
        130 │+"🐶");
  128   131 │ 
        132 │+(
  129   133 │ // Empty string.
  130       │-("");
        134 │+"");
  131   135 │ ("");
  132   136 │ 
        137 │+(
  133   138 │ // Single double quote.
  134       │-("\"");
        139 │+"\"");
  135   140 │ ('"');
  136   141 │ 
        142 │+(
  137   143 │ // Single single quote.
  138       │-("'");
        144 │+"'");
  139   145 │ ("\'");
  140   146 │ 
        147 │+(
  141   148 │ // Unnecessary escapes.
  142       │-("\'");
        149 │+"\'");
  143   150 │ ('\"');
  144   151 │ ("\a");
  145   152 │ ("\a");
  146   153 │ ("hol\a");
┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
  156   163 │ ("\a\a"); // consecutive unnecessarily escaped characters
  157   164 │ ("\a\a"); // consecutive unnecessarily escaped characters
  158   165 │ ("escaped \u2030 \‰ (should not stay escaped)");
  159   166 │ 
        167 │+(
  160   168 │ // Meaningful escapes
  161       │-("octal escapes \0 \1 \2 \3 \4 \5 \6 \7");
        169 │+"octal escapes \0 \1 \2 \3 \4 \5 \6 \7");
  162   170 │ ("octal escapes \0 \1 \2 \3 \4 \5 \6 \7");
  163   171 │ ("meaningfully escaped alphabetical characters \n \r \v \t \b \f \u2713 \x61");
  164   172 │ ("meaningfully escaped alphabetical characters \n \r \v \t \b \f \u2713 \x61");
  165   173 │ ("escaped newline \
┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
  168   176 │ ");
  169   177 │ ("escaped \u2028 \ ");
  170   178 │ ("escaped \u2029 \ ");
  171   179 │ 
        180 │+(
  172   181 │ // One of each.
  173       │-("\"'");
        182 │+"\"'");
  174   183 │ ('"\'');
  175   184 │ 
        185 │+(
  176   186 │ // One of each with unnecessary escapes.
  177       │-("\"\'");
        187 │+"\"\'");
  178   188 │ ('\"\'');
  179   189 │ 
        190 │+(
  180   191 │ // More double quotes than single quotes.
  181       │-("\"'\"");
        192 │+"\"'\"");
  182   193 │ ('"\'"');
  183   194 │ 
        195 │+(
  184   196 │ // More single quotes than double quotes.
  185       │-("\"''");
        197 │+"\"''");
  186   198 │ ('"\'\'');
  187   199 │ 
        200 │+(
  188   201 │ // Two of each.
  189       │-("\"\"''");
        202 │+"\"\"''");
  190   203 │ ('""\'\'');
  191   204 │ 
        205 │+(
  192   206 │ // Single backslash.
  193       │-("\\");
        207 │+"\\");
  194   208 │ ("\\");
  195   209 │ 
        210 │+(
  196   211 │ // Backslases.
  197       │-("\"\\\"\\\\\" '\'\\'\\\'\\\\'");
        212 │+"\"\\\"\\\\\" '\'\\'\\\'\\\\'");
  198   213 │ ('\'\\\'\\\\\' "\"\\"\\\"\\\\"');
  199   214 │ 
        215 │+(
  200   216 │ // Somewhat more real-word example.
  201       │-("He's sayin': \"How's it goin'?\" Don't ask me why.");
        217 │+"He's sayin': \"How's it goin'?\" Don't ask me why.");
  202   218 │ ('He\'s sayin\': "How\'s it goin\'?" Don\'t ask me why.');
  203   219 │ 
        220 │+(
  204   221 │ // Somewhat more real-word example 2.
  205       │-("var backslash = \"\\\", doubleQuote = '\"';");
        222 │+"var backslash = \"\\\", doubleQuote = '\"';");
  206   223 │ ('var backslash = "\\", doubleQuote = \'"\';');
  207   224 │ 
  208   225 │ ```
```

MichaReiser: You can call `split_trivia` on the `FormatElement` that formatting the token returns and then place the comments where you want.

IWANABETHATGUY: > You can call `split_trivia` on the `FormatElement` that formatting the token returns and then place the comments where you want.

Thanks, I forgot that api

ematipico (approved):
All the changes in prettier's test suite make total sense. Thank you for the great contribution!

IWANABETHATGUY:
> All the changes in prettier's test suite make total sense. Thank you for the great contribution!

Don't merge, I have the last issue not been addressed

crates/rome_js_formatter/src/utils/mod.rs
IWANABETHATGUY: I have been addressed this issue.

crates/rome_js_formatter/src/js/expressions/string_literal_expression.rs
```diff
@@ -20,19 +19,23 @@ impl FormatNode for JsStringLiteralExpression {
...
-            needs_parenthesis,
-        ))
+        let formatted_element = format_string_literal_token(value_token, formatter);
+        if needs_parenthesis {
+            let (leading_trivia, content, trailing_trivia) = formatted_element.split_trivia();
+            Ok(format_elements![
+                leading_trivia,
+                format_elements![token("("), content, token(")"),],
+                trailing_trivia,
+            ])
```
ematipico, line 36: This is neat!

IWANABETHATGUY:
!bench_formatter

IWANABETHATGUY:
@ematipico The benchmark shows  it's time to merge

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -41,6 +43,13 @@ impl FormatNode for JsParenthesizedExpression {
                 group_elements(expression.format(formatter)?),
                 formatter.format_replaced(&r_paren_token?, empty_element()),
             ])
+            // This branch aiming to align prettier formatting behavior, see https://github.com/rome/tools/pull/2547 for details
+        } else if JsStringLiteralExpression::can_cast(expression.syntax().kind()) {
+            Ok(format_elements![
+                l_paren_token.format(formatter)?,
+                group_elements(expression.format(formatter)?),
```
MichaReiser, line 50: The string is a single token. It shouldn't be needed to group the elements?

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -41,6 +43,13 @@ impl FormatNode for JsParenthesizedExpression {
                 group_elements(expression.format(formatter)?),
                 formatter.format_replaced(&r_paren_token?, empty_element()),
             ])
+            // This branch aiming to align prettier formatting behavior, see https://github.com/rome/tools/pull/2547 for details
```
MichaReiser, line 46: I'm not sure if this comment provides much information to the reader. You can get the same information using git annotate. Why not explain why the branch is necessary without forcing the reader to jump to another window. 

This logic now applies for all string literal expressions, for example also `a[("this here")]`. Should it apply here too or does it require a check if the parent is an `ExpressionStatement`?

crates/rome_js_formatter/src/js/expressions/string_literal_expression.rs
MichaReiser: Checking if the parent is a `JsExpressionStatement` is correct but you don't need to check if the grandparent is of a specific kind.

crates/rome_js_formatter/src/js/module/import_assertion_entry.rs
```diff
@@ -24,7 +24,7 @@ impl FormatNode for JsImportAssertionEntry {
             formatted_key,
             colon_token.format(formatter)?,
             space_token(),
-            format_string_literal_token(value_token?, formatter),
+            format_string_literal_token(value_token?, formatter,),
```
MichaReiser, line 27: Would you mind undoing these changes that added the trailing comma (also in other files)

MichaReiser (changes requested):

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
IWANABETHATGUY: I add test case here, https://github.com/rome/tools/pull/2547/files#diff-f46c5f800731eb225ee6a9ce2c34c209465ca180de9dd763d0892028be35bd87R13-R14.

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -41,6 +43,31 @@ impl FormatNode for JsParenthesizedExpression {
...
+        // "
+        // );
+        // ```
+        // which will not match prettier's formatting behavior, if we add this extra branch to handle this case, it become:
+        // ```js
+        // ("escaped carriage return \
+        // ");
+        // ```
+        // this is what we want
+        } else if JsStringLiteralExpression::can_cast(expression.syntax().kind()) {
```
MichaReiser, line 65: Does this logic apply to all string literal expressions? My understanding is that it should only apply to string literal expressions that are direct children of an expression statement.

crates/rome_js_formatter/tests/specs/js/module/expression/member-chain/computed.js.snap
```diff
@@ -24,3 +24,5 @@ nock(/test/)
 	.matchHeader("Accept", "application/json")[httpMethodNock(method)]("/foo")
 	.reply(200, { foo: "bar" });
 
+a[("test")];
```
MichaReiser, line 27: This shouldn't need parenthesis.

IWANABETHATGUY: Yeah, But I think this is not related to wrap string with parenthesis when `StringLiteralExpression` at statement position, since for now it doesn't remove the parenthesis neither https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#YQBbACgAIgB0AGUAcwB0ACIAKQBdADsA

IWANABETHATGUY: Would like to fix in another pull request

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
MichaReiser: Will be addressed as part of #2564

MichaReiser (approved):

PR #2549, opened 2022-05-06
fix(rome_js_formatter): empty statements in `if/else` and `do/while`

### LINKED ISSUE #2448: 🐛 loop/if with empty statement body

ematipico:
## Summary

Closes #2448

I cherry-picked the implementation of the former PR, which aligns its behaviour to prettier. 

The PR changes how `if/else` and `do/while` statements are formatted where the body is an empty statement, by matching Prettier's behaviour:

```js
// before
if(a); else ;

// now
if (a);
else;
```

## Test Plan

New tests to cover the case

### DIFF
--- a/crates/rome_js_formatter/src/js/statements/do_while_statement.rs
+++ b/crates/rome_js_formatter/src/js/statements/do_while_statement.rs
@@ -19,7 +19,7 @@ impl FormatNodeFields<JsDoWhileStatement> for FormatNodeRule<JsDoWhileStatement>
             semicolon_token,
         } = node.as_fields();
 
-        let head = formatted![formatter, [do_token.format(), space_token(),]]?;
+        let head = formatted![formatter, [do_token.format()]]?;
 
         let tail = formatted![
             formatter,
@@ -40,5 +40,26 @@ impl FormatNodeFields<JsDoWhileStatement> for FormatNodeRule<JsDoWhileStatement>
         ]?;
 
         formatted![formatter, [head, body.format(), tail,]]
+        let body = body?;
+        if matches!(body, JsAnyStatement::JsBlockStatement(_)) {
+            Ok(hard_group_elements(format_elements![
+                head,
+                space_token(),
+                body.format(formatter)?,
+                tail,
+            ]))
+        } else if matches!(body, JsAnyStatement::JsEmptyStatement(_)) {
+            Ok(format_elements![
+                hard_group_elements(format_elements![head, body.format(formatter)?,]),
+                hard_line_break(),
+                tail,
+            ])
+        } else {
+            Ok(format_elements![
+                hard_group_elements(format_elements![head, space_token()]),
+                body.format(formatter)?,
+                hard_group_elements(tail),
+            ])
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


MichaReiser:
It would help reviewers to add some more context to the PR summary. How does it format empty statements differently than before, in what cases. I otherwise have to open the issue, read through all the discussions or infer the logic from your changes which makes it hard to tell if the logic is working as intended.

crates/rome_js_formatter/src/js/statements/do_while_statement.rs
```diff
@@ -39,12 +39,19 @@ impl FormatNode for JsDoWhileStatement {
...
+                space_token(),
                 body.format(formatter)?,
                 tail,
             ]))
+        } else if matches!(body, JsAnyStatement::JsEmptyStatement(_)) {
+            Ok(format_elements![
+                hard_group_elements(format_elements![head, body.format(formatter)?,]),
+                hard_line_break(),
+                tail,
+            ])
```
yassere, line 51: The `tail` here isn't inside a hard group, so this would reposition a block comment from:

```js
do;
while /* comment */ (true); 
```
To:
```js
do;
while (true); /* comment */
```
Speaking of hard groups, I'm not sure why `JsBlockStatement` requires special grouping here. Would this work for everything except `JsEmptyStatement`?
```rs
hard_group_elements(format_elements![
    head,
    space_token(),
    body.format(formatter)?,
    tail,
])
```

yassere: Actually, PR #2597 removes hard groups, so this comment will be out of date once that merges. Looks like inline block comments will be addressed later.

ematipico:
> It would help reviewers to add some more context to the PR summary. How does it format empty statements differently than before, in what cases. I otherwise have to open the issue, read through all the discussions or infer the logic from your changes which makes it hard to tell if the logic is working as intended.

I updated the description with a brief example of the changes.

ematipico:
> Looks good. Would you mind generating the prettier report and posting how the metric changes with your PR?

PR 
```
**File Based Average Prettier Similarity**: 70.84%  
**Line Based Average Prettier Similarity**: 64.85% 
```

`main`
```
**File Based Average Prettier Similarity**: 70.80%  
**Line Based Average Prettier Similarity**: 64.83%  
```

PR #2556, opened 2022-05-08
feat(rome_js_parser): optional variance annotation

IWANABETHATGUY:
## Summary
part of #2400 

**current**: 
**File Based Average Prettier Similarity**: 72.69%  
**Line Based Average Prettier Similarity**: 67.25%  

**main**: 
**File Based Average Prettier Similarity**: 72.65%  
**Line Based Average Prettier Similarity**: 67.13%  

## Test Plan
I took test cases from esbuild https://github.com/evanw/esbuild/pull/2102,  the original test case of Typescript pull request related to semantics, which we don't care too much about. esbuild `Optional variance annotations ` pull request test case is more suitable for us.


IWANABETHATGUY:
!bench_parser

IWANABETHATGUY:
This pull request is pretty big, so I prefer to separate the `optional variance annotation` formatting another pr

xtask/codegen/js.ungram
```diff
@@ -1880,10 +1880,15 @@ TsTypeParameterList = (TsTypeParameter (',' TsTypeParameter)* ','?)
...
 TsTypeParameter =
+	modifier: TsTypeParameterModifier?
 	name: TsTypeParameterName
 	constraint: TsTypeConstraintClause?
 	default: TsDefaultTypeClause?
 
+TsTypeParameterModifier = 
+	'in'? 
+	'out'?
+
```
MichaReiser, line 1891: My understanding is that `MyType<in out T>` isn't valid, right? It should then be sufficient to define this as a union (you can even do this inline)

```
TsTypeParameter = 
  modifier_token: ('in' | 'out')?
```

crates/rome_js_parser/src/syntax/typescript/types.rs
```diff
@@ -146,13 +158,130 @@ impl ParseSeparatedList for TsTypeParameterList {
...
+    parse_ts_type_parameter_modifier(p, could_use_parameter_modifier).ok();
+    let name = parse_ts_type_parameter_name(p);
+    parse_ts_type_constraint_clause(p).ok();
+    parse_ts_default_type_clause(p).ok();
+    if name.is_absent() {
+        m.abandon(p);
+        Absent
+    } else {
+        Present(m.complete(p, TS_TYPE_PARAMETER))
+    }
```
MichaReiser, line 284: This isn't correct. Your implementation now returns `absent` if you encounter something like `T<out>` because it parses the `out` as modifier but then doesn't find a name. 

However, `T<out>` should parse as a type parameter named `out` without a modifier.

xtask/codegen/js.ungram
IWANABETHATGUY: not really, MyType<in out T> is valid, but MyType<out in T> is not valid.

IWANABETHATGUY: https://www.typescriptlang.org/play?ts=4.8.0-dev.20220508#code/C4TwDgpgBACgTgewG4EsAmE4B4EFdhQAqAfFALxQAUAlOaYQNwBQokUAwggHYDOuAtpiwouRUhUoAPAFxFaZUkgTpmraAFkAhmEjYRRADRQ8BAKriqMuXSinV4aPAQBjCDx4I9ok2PKXZhPL09mxOru6eAIw4+FD6JH5SAUFEIdCEbsDRJhY0NiZAA

crates/rome_js_parser/src/syntax/typescript/types.rs
IWANABETHATGUY: https://github.com/rome/tools/pull/2556/commits/9c414d0a3fdda9ad5ceacb788f242b36079c051e#diff-b6c67e053b97ae47ba71fe0d7f76c30828370589a13a1484672a5400a7385182R36-R63, I add a new test case `// type Foo<out> = out` here, the ast should match the expected result. `modifer` is missing, and `TypeParameterName` is `out`

IWANABETHATGUY:
@MichaReiser , @ematipico could we make some progress here ? Since prettier has been landed Ts 4.7, https://github.com/prettier/prettier/pull/12900

xtask/codegen/js.ungram
MichaReiser: the case `MyType<in out T>` isn't supported by your proposed grammar because each type can only have a single modifier or would it nest type constraints in that case? 

The main challenge I continue to see and don't know how to solve right now is that type parameters are only allowed in very specific cases and I'm not sure if we should use two different node types instead to make that clear to analyzers/ formatters etc or not. That's the main reason why I've been hesitant of merging this PR.

crates/rome_js_parser/src/syntax/class.rs
```diff
@@ -249,9 +249,13 @@ fn parse_class(p: &mut Parser, kind: ClassKind) -> CompletedMarker {
     // test ts ts_class_type_parameters
     // class BuildError<A, B, C> {}
     TypeScript
-        .parse_exclusive_syntax(p, parse_ts_type_parameters, |p, type_parameters| {
-            ts_only_syntax_error(p, "class type parameters", type_parameters.range(p))
-        })
+        .parse_exclusive_syntax(
+            p,
+            |p| parse_ts_type_parameters(p, true),
```
MichaReiser, line 254: I find `true`/`false` parameters hard to understand. Having a `parse_ts_type_parameters_impl` function that takes a boolean parameter is fine but I would probably expose a `parse_ts_type_parameters` and a `ts_type_parameters_with_modifiers` function instead to make the difference clear.

crates/rome_js_parser/test_data/inline/ok/type_parameter_modifier_tsx.tsx
```diff
@@ -0,0 +1,7 @@
+<in T></in>
+// <out T></out>
+// <in out T></in>
```
MichaReiser, line 3: Is it intentional that these are all commented out?

IWANABETHATGUY: IIRC, these syntax are not supported by our parser

IWANABETHATGUY: https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#PABvAHUAdAAgAFQAPgA8AC8AbwB1AHQAPgAKADwAaQBuACAAbwB1AHQAIABUAD4APAAvAGkAbgA+AA==

xtask/codegen/js.ungram
MichaReiser: I'm wondering if it makes more sense to inline `TsTypeParameterModifier` into `TsTypeParameter` since it doesn't give any additional guarantees and we should generally avoid nodes that have all optional children.

crates/rome_js_parser/test_data/inline/ok/type_parameter_modifier_tsx.tsx
MichaReiser: You probably want to insert semicolons after each statement to parse each line as its own statement.

MichaReiser: You have to enable JSX [playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=true#PABvAHUAdAAgAFQAPgA8AC8AbwB1AHQAPgA7AAoAPABpAG4AIABvAHUAdAAgAFQAPgA8AC8AaQBuAD4AOwA=)

crates/rome_js_parser/src/syntax/class.rs
IWANABETHATGUY: @MichaReiser， any thought about this?

PR #2557, opened 2022-05-09
refactor(Formatter): `formatted` Macro

MichaReiser:
# # Summary

This PR introduces a new `formatted` macro to format objects that are either `FormatElement`s or implement `Format`. The macro is needed to implement the shared formatter infrastructure #2525 without regressing our API. 

The proposed design in #2525 is that each formatter crate defines its own `AsFormat` and `IntoFormat` trait. Objects implementing either of these traits don't implement `Format` themselves (because of Rust's orphan rule) but expose a `format` and `into_format` methods that return an object that implements the formatting for it. 

The main challenge with this design is that `AstNode.format` doesn't return `FormatResult<FormatElement>` but an object implementing `Format` and `format_elements!` doesn't support combining `FormatElement`s with objects implementing `Format`. 

I considered changing the `AsFormat` trait to have two methods: `as_format` and `format` where `format` has a default implementation that calls `as_format().format(formatter)`. This worked surprisingly well except for the `FormatOptionalTrait` because that works on object implementing `Format`. Formatting an optional object that implements `AsFormat` would then have required to call `if_stmt.else_clause.as_format().format_or_empty(formatter)` which felt excessive. 

That's why this PR introduces a new `formatted!` macro similar to Rust's `write!` macro. The macro accepts a `Formatter` followed by the elements that should be formatted. It's result is `FormatResult<FormatElement>`. It accepts any object that implements the new `IntoFormatElement` trait. `IntoFormatElement` is similar to `Format`, except that it takes ownership of `self`. This new trait is necessary because implementing `Format` for `FormatElement` can't be done in an efficient way because it would be necessary to `clone` the `FormatElement` with each `format` call. 

The introduction of the `formatted` macro starts a shift of our formatter design from a `FormatElement` focused formatting to a logic where the formatting logic centers around the `Format` macro. I believe this will be beneficial over time and address some of the seams that have become visible in our current formatter design where building reusable logic around formatting a sequence of `Format` object felt awkward, because it often time required iterating over them first to collect any potential error that happens when formatting the object, and then returning an iterator over the `FormatElement`s. 

The final design that I've in mind is that our formatter changes so that even `concat`, `group`, `join` accept objects that are formattable AND the IR builders (`space_token`) don't return a `FormatElement` but an object that implements `Format` too. This will also bring us closer to a design where it will ultimately become possible to write to a shared buffer rather than having each `Format` impl return a `FormatElement`. `FormatElement` really is just the IR relevant for printing but should be less important when implementing the formatting logic. 

The main downside of this implementation is that there are now two macros and engineers need to know when to use `format_elements` and when `formatted`. This may even extend to three macros in the future, a `format_args` macro if we decide to implement `space` etc. to return `Format` objects. 

## Alternative designs

#2525 discussed two alternative designed that we agreed to rule out. However, the discussion didn't cover the alternative of copying `Format`, `Formatter`, and `FormatTraits` into every formatter crate. The benefit of such approach is that:

* Significantly smaller complexity (-4 traits or so)
* Simpler API because it only requires one macro

The main downside I see in that is that it prevents us from implementing any shared formatting logic inside of `rome_formatter`. For example, the formatting of separated lists can be done language-agnostic if it is parametrized with the trailing separator behaviour. 

## Review

The design relevant (the other changes are manual refactors to the new design) changes are in:

* Prelude: `rome_formatter` and `rome_js_formatter` now define a `prelude` adding usages for the most commonly used types
* builders: `ConcatBuilder` in `rome_formatter` exposes an imperative way of contacting elements
* traits: `format_extensions` (previously `format_traits`) now return object implementing `Format` rather than `FormatResult<FormatElement>`
* `IntoFormatElement`: Trait that takes ownership of self and return a `FormatResult<FormatElement>`. For now in `rome_js_formatter`, will be moved to `rome_formatter`
* `macros`: Defines the new `formatted` macro. For now in `rome_js_formatter` but will be moved to `rome_formatter` together with  `Format` and `Formatter`.

## Test Plan

`cargo test`


crates/rome_js_formatter/src/format_traits.rs
```diff
@@ -1,263 +0,0 @@
...
-    /// use rome_js_syntax::{JsSyntaxToken};
-    /// use rome_js_formatter::prelude::*;
-    ///
-    /// let formatter = Formatter::default();
-    /// let token: Option<JsSyntaxToken> = None;
-    /// // we wrap the token in [Ok] so we can simulate SyntaxResult.
-    /// let result = token.format_or_empty(&formatter);
-    ///
-    /// assert_eq!(Ok(empty_element()), result)
-    fn format_or_empty(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
```
MichaReiser, line 28: No longer necessary because `Option<Format>` implements `Format`

crates/rome_js_formatter/src/format_extensions.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// let formatter = Formatter::default();
+    ///
+    /// // Wrap the token in [Ok] so we can simulate SyntaxResult.
+    /// let result = Ok(syntax_token);
+    /// let result = result.with(|string_literal| {
+    ///     formatted![&formatter, string_literal, space_token(), token("+")]
+    /// });
+    ///
+    /// assert_eq!(formatted![&formatter, token("'abc'"), space_token(), token("+")], formatted![&formatter, result])
+    fn with<With, WithResult>(&self, with: With) -> FormatItemWith<With, WithResult>
```
MichaReiser, line 160: Now returns an object implementing `Format` rather than a `FormatElement`.

crates/rome_js_formatter/src/generated.rs
```diff
@@ -1,7 +1,6 @@
 //! Generated file, do not edit by hand, see `xtask/codegen`
 
-use crate::{Format, FormatElement, FormatNode, Formatter};
-use rome_formatter::FormatResult;
+use crate::prelude::*;
```
MichaReiser, line 3: I improved our `prelude` and started using it everywhere. This should simplify future refactors.

crates/rome_js_formatter/src/js/any/function.rs
```diff
@@ -15,40 +15,41 @@ impl Format for JsAnyFunction {
...
-                    ]),
-                    token(")"),
-                ])
-            }
+            JsAnyArrowFunctionParameters::JsAnyBinding(binding) => group_elements(formatted![
+                formatter,
+                token("("),
+                soft_block_indent(formatted![
+                    formatter,
+                    binding.format(formatter)?,
```
MichaReiser, line 44: Calling `.format(formatter)` isn't strictly necessary anymore BUT I kept them around because my follow up PR will make them necessary again because `AstNode` then doesn't implement `Format` any longer. Keeping the `format` calls around will hopefully make my refactor a bit easier.

crates/rome_js_formatter/src/macros.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Helper macro that counts the count of elements passed
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __count_elements {
+    () => {0usize};
+    ($ex:expr) => {1usize};
+    ($_head:expr, $($tail:expr),* $(,)?) => {1usize + $crate::__count_elements!($($tail),*)};
+}
```
MichaReiser, line 91: This is a bit of a hack. The problem is that each `IntoFormatElement` has a different memory layout, thus, cannot be stored in a stack-allocated array. 

My plan is to follow up with a `format_args` macro once `space_token` etc return an object that implement `Format` that only works on `Format` and accepts the objects by reference.

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/js/any/class.rs
```diff
@@ -9,34 +9,26 @@ use rome_rowan::AstNode;
...
-            .format_with_or_empty(formatter, |token| format_elements![token, space_token()])?;
+        let abstract_token = self.abstract_token();
+        let id = self.id();
+        let extends = self.extends_clause();
+        let implements_clause = self.implements_clause();
 
-        let id = self
-            .id()
-            .format_with_or_empty(formatter, |id| format_elements![space_token(), id])?;
+        let implements_clause = implements_clause.with_or_empty(|implements_clause| {
```
leops, line 17: Maybe the formatter should be provided as an argument to the closure by the `FormatWith` and `FormatOptional` trait instead of being captured from the outer context, this could be important if we want to use combinator-based formatter implementations (where the formatter used to print a certain token and the formatter to print the entire node may be different instances similar to how this works in `std::fmt` to customize how inner data gets printed) but obviously has the disadvantage of making every use of `with_*` a lot more verbose

crates/rome_js_formatter/src/macros.rs
```diff
@@ -0,0 +1,177 @@
...
+///
+/// let formatted = formatted![&formatter, token("test")].unwrap();
+///
+/// assert_eq!(formatted, token("test"));
+/// ```
+#[macro_export]
+macro_rules! formatted {
+
+    // called for things like formatted![formatter, token("test")]
+    ($formatter:expr, $element:expr) => {
```
leops, line 63: One small issue I have with the syntax of this macro is that the formatter argument isn't clearly separated from the the format elements, so many calls to the macro would end up looking like `formatted![formatter, keyword, ident, node, token]` where the formatter looks like any other element of the list which could be misleading to people unfamiliar with the codebase. I think adding just a few tokens to the macro signature to have it written as `formatted!(formatter, [keyword, ident, node, token])` would make it easier to understand what going on at a glance by making the intent more clear

crates/rome_js_formatter/src/macros.rs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            $crate::IntoFormatElement::into_format_element($element, $formatter)
```
leops, line 65: I think it might be possible for this macro to be owned by `rome_formatter` if it just relied on `$element` having an `into_format_element` method without explicitly naming out the trait (which should be the case if the language-specific `IntoFormatElement` is in that crate's prelude, then it will be in scope for all relevant`FormatNode` implementations), and injected that implementation into the generic `FormatBuilder` using a function pointer / closure

crates/rome_js_formatter/src/macros.rs
leops: Pro tip: when writing macros in Rust there is no such thing as a "hack", there are no rules and everything is permitted as long as it works ...

crates/rome_js_formatter/src/js/any/class.rs
leops: This leaves me wondering if the current combinator-method based design is any good with a formatter architecture where elements get written into a mutable formatter: one could expect these methods to be free of side-effect and simply wrap their returned value, but in practice they mutate the formatter already to track the printed tokens.
At the moment this mutation is order independent (tokens just get added to a set) but if we started writing the tokens to the formatter output buffer then it gets crucial to get the order right, and I think having an explicit `if let Some(implements_clause) = implements_clause { write!(formatter, space_token(), implements_clause); }` might work out better than building up a lazy `Format` object out of functional combinators to have it printed out imperatively at the end anyway

crates/rome_js_formatter/src/format_extensions.rs
```diff
@@ -0,0 +1,283 @@
...
+        WithResult: IntoFormatResult;
+}
+
+pub struct FormatItemWith<'a, With, WithResult>
+where
+    With: Fn(FormatElement) -> WithResult,
+    WithResult: IntoFormatResult,
+{
+    with: With,
+    inner: &'a dyn Format,
```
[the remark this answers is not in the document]
MichaReiser: The sole reason was to not add any more generic types and it doesn't seem to make any significant difference in regards of performance. But that's all. 

The type would need to be `inner: &'a F` where `F` isn't constraint to `Format` except if we change `FormatOptionalTrait` to inherit from `Format`.

crates/rome_js_formatter/src/js/any/class.rs
MichaReiser: > This leaves me wondering if the current combinator-method based design is any good with a formatter architecture where elements get written into a mutable formatter: one could expect these methods to be free of side-effect and simply wrap their returned value, but in practice they mutate the formatter already to track the printed tokens.
At the moment this mutation is order independent (tokens just get added to a set) but if we started writing the tokens to the formatter output buffer then it gets crucial to get the order right, and I think having an explicit if let Some(implements_clause) = implements_clause { write!(formatter, space_token(), implements_clause); } might work out better than building up a lazy Format object out of functional combinators to have it printed out imperatively at the end anyway

This is a great point and something I considered too when exploring a design where the formatter writes to a shared buffer. The reason why I kept this as is for now is because 

a) it's unclear if we ever fully change the design and the current approach is neat for what we have at the moment
b) I wanted to avoid refactors that aren't strictly necessary. 

> Maybe the formatter should be provided as an argument to the closure by the FormatWith and FormatOptional trait instead of being captured from the outer context, this could be important if we want to use combinator-based formatter implementations (where the formatter used to print a certain token and the formatter to print the entire node may be different instances similar to how this works in std::fmt to customize how inner data gets printed) but obviously has the disadvantage of making every use of with_* a lot more verbose

Another great point. I would argue that it isn't required at the time being and that's why I would optimize the API for the current use case.

crates/rome_js_formatter/src/macros.rs
MichaReiser: My plan is to move `IntoFormatElement` to `rome_formatter` together with `Format`, `Formatter` and many of the other crates. Also, isn't the compiler providing better error messages if the type is named explicitly?

crates/rome_js_formatter/src/macros.rs
MichaReiser: Fair point. I'm not very opinionated on this. My main counterpoint would be that `formatted` is similar to Rust's `write!(f, "string", args)` macro. I guess the difference is that the format string helps to separate the elements. 

I would need to find a good regex refactor if I want to make this change. I rather don't want to go through all them manually again :sob:

MichaReiser:
You can see how this is used to move the `Format` to `rome_formatter` in #2559 2559

MichaReiser:
Interesting that this PR deletes more code than it adds :)

crates/rome_js_formatter/src/macros.rs
```diff
@@ -0,0 +1,177 @@
...
+///         &formatter,
+///         token("a"),
+///         space_token(),
+///         token("simple"),
+///         space_token(),
+///         TestFormat
+///     ]
+///     .unwrap();
+///
+///     assert_eq!(
```
NicholasLYang, line 36: nit but the formatting is rather confusing here. It looks like `formatted` is being assigned to the entire chunk of code

crates/rome_js_formatter/src/macros.rs
```diff
@@ -0,0 +1,177 @@
...
+#[macro_export]
+macro_rules! __count_elements {
+    () => {0usize};
+    ($ex:expr) => {1usize};
+    ($_head:expr, $($tail:expr),* $(,)?) => {1usize + $crate::__count_elements!($($tail),*)};
+}
+
+#[doc(hidden)]
+pub struct FormatBuilder {
+    builder: ConcatBuilder,
```
NicholasLYang, line 95: Could we not wrap the ConcatBuilder itself in a Result?

NicholasLYang:
We should consider writing up a document that explains all the different formatting traits and what they're used for. The inline documentation is helpful but it'd be nice to have it all in one place

crates/rome_formatter/src/builders.rs
```diff
@@ -0,0 +1,70 @@
...
+        } else {
+            match element {
+                FormatElement::List(list) => self.elements.extend(list.into_vec()),
+                item => self.elements.push(item),
+            }
+        }
+    }
+
+    #[inline]
+    pub fn size_hint(&mut self, hint: (usize, Option<usize>)) {
```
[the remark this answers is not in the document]
MichaReiser: The `ConcatBuilder` has an optimisation that it tries to reuse the collection in case the first element written is a `List`. That's why it isn't possible to call `reserve` inside `new`. 

It should be possible to rewrite some formatters that internally build up a `vec` and at the very end call `concat_elements(vec)` to use `ConcatBuilder` to avoid one additional allocation. It's often not known in these cases how many elements the vec will contain

crates/rome_js_formatter/src/format_extensions.rs
```diff
@@ -0,0 +1,283 @@
...
+{
+    fn format(&self, formatter: &Formatter) -> FormatResult<FormatElement> {
+        let element = self.inner.format(formatter)?;
+
+        (self.with)(element).into_format_result()
+    }
+}
+
+/// Utility trait to convert [crate::FormatElement] to [FormatResult]
+pub trait IntoFormatResult {
```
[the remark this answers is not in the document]
MichaReiser: I think Rust's orphan rule is preventing this yet again. 

* `Result`: Defined in std
* `Into` Defined in std

So both the type and the trait or foreign to the `rome_js_formatter` crate. 

However, I merged this trait with `IntoFormatElement` in the follow-up PR.

crates/rome_js_formatter/src/macros.rs
MichaReiser: `ConcatBuilder` is used by `format_elements` that doesn't support returning a `Result`. That's why this additional indirection is needed. 

I'm planning to spend some time investing if all the `concat`, `group` can be re-wrrtten to instead implement `Format`. That would then allow merging `Concat` and `Format`Builder.

MichaReiser:
> We should consider writing up a document that explains all the different formatting traits and what they're used for. The inline documentation is helpful but it'd be nice to have it all in one place

That's a good point. Is it OK if I write some module doc as part of my follow up PR that moves the traits to `rome_formatter` so that rebasing will be less painful?

crates/rome_js_formatter/src/js/any/function.rs
leops: I think one of the confusing aspect of this "in-between state" before the followup PR gets merged is the formatter having to be passed to both the `format` method and the `formatted!` macro, as this makes it a bit unclear where the actual formatting is happening (but if I understand correctly after the followup refactor that call will just be `.format()` to turn the AST node into a `FormatWithRule` adapter so that should make it easier to understand)

MichaReiser: Yes, the calls will just be `format`. It's also an open question on the follow up PR if we should either

* `Format.format`. Rename to `fmt` to align with `Debug` and `Display`?
* `AsFormat.format` Rename to `as_format`?

PR #2560, opened 2022-05-09
chore(website): Added careers link

NicholasLYang:
## Summary

I figured we should have a link on our website


MichaReiser (approved):
That's a great idea. Not sure about the ordering. Should we add it last (after playground?)

NicholasLYang:
I kept it alphabetical, but perhaps we could reorder at some point

PR #2562, opened 2022-05-09
feat(website/playground): add switch to change the syntax tree representation

IWANABETHATGUY:
## Summary
 add a switch to control the syntax tree representation 
 

https://user-images.githubusercontent.com/17974631/167458560-d9eca5ed-3866-4b7c-9971-2d5f94abfd32.mp4

## Test Plan
I think it doesn't need to test?


IWANABETHATGUY:
Current Workflow would have formating issue , when `yarn format` would use  formatting binary latest main branch, while our github action use latest release binary,  So I would like to change the `yarn format` to use latest release binary instead run `cargo run -p rome_cli --release -- format --write ./src`

PR #2563, opened 2022-05-09
feature(website): Toggle Tree Style

NicholasLYang:
## Summary

As some have pointed out, the JSON view is not optimal for some situations. This allows users to toggle between the text and JSON view.

## Test Plan
Screen Shot 2022-05-11 at 1 14 29 PM
Screen Shot 2022-05-11 at 1 14 33 PM
Screen Shot 2022-05-11 at 1 15 10 PM
Screen Shot 2022-05-11 at 1 15 15 PM

### DIFF
--- a/website/playground/src/DesktopPlayground.tsx
+++ b/website/playground/src/DesktopPlayground.tsx
@@ -7,7 +7,7 @@ import TreeView from "./TreeView";
 
 export default function DesktopPlayground(
 	{
-		playgroundState: { code, setCode, ...settings },
+		playgroundState: { code, setCode, treeStyle, setTreeStyle, ...settings },
 		prettierOutput,
 		romeOutput: { cst, ast, formatted_code, formatter_ir, errors },
 	}: PlaygroundProps,
@@ -76,10 +76,25 @@ export default function DesktopPlayground(
 							/>
 						</TabPanel>
 						<TabPanel>
+<<<<<<< HEAD
 							<TreeView tree={JSON.parse(cst)} />
 						</TabPanel>
 						<TabPanel>
 							<TreeView tree={JSON.parse(ast)} />
+=======
+							<TreeView
+								treeStyle={treeStyle}
+								setTreeStyle={setTreeStyle}
+								tree={cst}
+							/>
+						</TabPanel>
+						<TabPanel>
+							<TreeView
+								treeStyle={treeStyle}
+								setTreeStyle={setTreeStyle}
+								tree={ast}
+							/>
+>>>>>>> cf4faf5829 (Toggling tree view for json or text)
 						</TabPanel>
 						<TabPanel>
 							<pre className="h-screen overflow-scroll">{formatter_ir}</pre>
--- a/website/playground/src/DesktopPlayground.tsx
+++ b/website/playground/src/DesktopPlayground.tsx
@@ -1,13 +1,14 @@
 import { PlaygroundProps } from "./types";
 import CodeEditor from "@uiw/react-textarea-code-editor";
-import { getLanguage } from "./utils";
+import { createSetter, getLanguage } from "./utils";
 import { Tab, TabList, TabPanel, Tabs } from "react-tabs";
 import { SettingsMenu } from "./SettingsMenu";
 import TreeView from "./TreeView";
 
 export default function DesktopPlayground(
 	{
-		playgroundState: { code, setCode, treeStyle, setTreeStyle, ...settings },
+		setPlaygroundState,
+		playgroundState: { code, treeStyle, ...settings },
 		prettierOutput,
 		romeOutput: { cst, ast, formatted_code, formatter_ir, errors },
 	}: PlaygroundProps,
@@ -17,7 +18,10 @@ export default function DesktopPlayground(
 	return (
 		<div className="divide-y divide-slate-300">
 			<h1 className="p-4 text-xl">Rome Playground</h1>
-			<SettingsMenu settings={settings} />
+			<SettingsMenu
+				settings={settings}
+				setPlaygroundState={setPlaygroundState}
+			/>
 			<div className="box-border flex h-screen divide-x divide-slate-300">
 				<div className="w-1/2 p-5">
 					<CodeEditor
@@ -25,7 +29,10 @@ export default function DesktopPlayground(
 						language={language}
 						placeholder="Enter some code here"
 						onChange={(evn) => {
-							setCode(evn.target.value);
+							setPlaygroundState((state) => ({
+								...state,
+								code: evn.target.value,
+							}));
 						}}
 						style={{
 							fontSize: 12,
@@ -84,14 +91,14 @@ export default function DesktopPlayground(
 =======
 							<TreeView
 								treeStyle={treeStyle}
-								setTreeStyle={setTreeStyle}
+								setTreeStyle={createSetter(setPlaygroundState, "treeStyle")}
 								tree={cst}
 							/>
 						</TabPanel>
 						<TabPanel>
 							<TreeView
 								treeStyle={treeStyle}
-								setTreeStyle={setTreeStyle}
+								setTreeStyle={createSetter(setPlaygroundState, "treeStyle")}
 								tree={ast}
 							/>
 >>>>>>> cf4faf5829 (Toggling tree view for json or text)
--- a/website/playground/src/DesktopPlayground.tsx
+++ b/website/playground/src/DesktopPlayground.tsx
@@ -6,12 +6,14 @@ import { SettingsMenu } from "./SettingsMenu";
 import TreeView from "./TreeView";
 import ReactJson from "react-json-view";
 
-export default function DesktopPlayground({
-	setPlaygroundState,
-	playgroundState: { code, treeStyle, ...settings },
-	prettierOutput,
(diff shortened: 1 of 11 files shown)


website/playground/src/TreeStyleSelect.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface Props {
+	treeStyle: TreeStyle;
+	setTreeStyle: Dispatch<SetStateAction<TreeStyle>>;
+}
+
+export default function TreeStyleSelect({ treeStyle, setTreeStyle }: Props) {
+	return (
+		<div
+			className="group p-0.5 rounded-lg flex bg-gray-200 mb-4 m-2"
+			style={{ width: "fit-content" }}
```
mrkldshv, line 14: Is it possible to use tailwind css `w-fit` class here?

website/playground/src/TreeStyleSelect.tsx
```diff
@@ -0,0 +1,69 @@
...
+						"bg-white shadow-sm ring-1 ring-black ring-opacity-5"
+				)}
+			>
+				<span
+					className={classNames(
+						"p-1.5 lg:pl-2.5 lg:pr-3.5 rounded-md flex items-center text-sm font-medium bg-white shadow-sm ring-1 ring-black ring-opacity-5",
+						treeStyle === TreeStyle.Json &&
+							"bg-white shadow-sm ring-1 ring-black ring-opacity-5"
+					)}
+					style={{ display: "none" }}
```
mrkldshv, line 31: Is it needed?

IWANABETHATGUY:
Could you add some screenshots or videos about the change of this pr?

IWANABETHATGUY:
> Could you add some screenshots or videos about the change of this pr?

Sorry, I didn't see the preview playground

MichaReiser:
> > Could you add some screenshots or videos about the change of this pr?
> 
> Sorry, I didn't see the preview playground

I would still find it useful to have a proper test summary because it safes me time as a reviewer:
* No need to open the playground and go searching for the new option
* It helps me understand how you tested your change. Thish enables me to make suggestion in case I'm aware of something that needs testing too.
* It acts as documentation for the future. Assume Rome continues to grow and some PR that you haven't been aware of breaks one of your features. Proper test cases are useful in this scenario because it makes it easier for you do understand what changed and indeed, check if what you're looking into was tested or should have been tested.

website/playground/src/DesktopPlayground.tsx
```diff
@@ -75,10 +75,18 @@ export default function DesktopPlayground(
 							/>
 						</TabPanel>
 						<TabPanel>
-							<TreeView tree={cst} />
+							<TreeView
+								treeStyle={treeStyle}
+								setTreeStyle={setTreeStyle}
```
MichaReiser, line 80: Events are commonly called `onTreeStyleChanged` to align it with HTML event handlers.

website/playground/src/TreeStyleSelect.tsx
```diff
@@ -0,0 +1,56 @@
+import { classNames } from "./utils";
+import { TreeStyle } from "./types";
+import { Dispatch, SetStateAction } from "react";
+
+interface Props {
+	treeStyle: TreeStyle;
+	setTreeStyle: Dispatch<SetStateAction<TreeStyle>>;
```
MichaReiser, line 7: This type signature may be correct but I would argue that it doesn't help future readers to understand the code. 

```
onTreeStyleChanged(newStyle: TreeStyle): void
```

Is more readable in my view and even less to type (and no need for type imports)

website/playground/src/TreeView.tsx
```diff
@@ -1,12 +1,24 @@
...
+}
 
-export default function TreeView({ tree }: Props) {
+export default function TreeView({ tree, treeStyle, setTreeStyle }: Props) {
 	return (
 		<div className="overflow-scroll">
-			<ReactJson src={JSON.parse(tree)} />
+			<TreeStyleSelect treeStyle={treeStyle} setTreeStyle={setTreeStyle} />
+			{treeStyle === TreeStyle.Json ? (
+				<ReactJson style={{ zIndex: "10 " }} src={JSON.parse(tree)} />
```
MichaReiser, line 18: Why is it needed to set an explicit zIndex?

website/playground/src/types.ts
```diff
@@ -21,6 +22,8 @@ export interface PlaygroundState {
 	setIsTypeScript: (isTypeScript: boolean) => void;
 	isJsx: boolean;
 	setIsJsx: (isJsx: boolean) => void;
+	treeStyle: TreeStyle;
+	setTreeStyle: Dispatch<SetStateAction<TreeStyle>>;
```
MichaReiser, line 26: Nit: Change to plain function

website/playground/src/utils.ts
```diff
@@ -98,6 +103,8 @@ export function usePlaygroundState(): PlaygroundState {
 		setIsJsx,
 		sourceType,
 		setSourceType,
+		treeStyle,
+		setTreeStyle,
```
MichaReiser, line 107: This is becoming hard to navigate. Separating this into a single state object with a reducer that allows changing single fields would remove the fields by about 50%. 

Another alternative is to store the values in a context and having a provider + consumer hook.

MichaReiser (changes requested):
The label seems to be missing on mobile

image

website/playground/src/DesktopPlayground.tsx
MichaReiser: The main reason why I would change it is because `TreeView` now makes very specific assumption in what context its used. The signature implies that `setTreeStyle` must set the tree style. It kind of takes the choice from the parent what it want's to do in that case.

NicholasLYang: I'm confused, how does `TreeView` make an assumption about the context in which it's used?

website/playground/src/utils.ts
MichaReiser: I don't think you need to add Redux just for that. A simple `useContext` should do to have shared state or change it to `useDispatcher`.

website/playground/src/SettingsMenu.tsx
```diff
@@ -2,53 +2,55 @@ import LineWidthInput from "./LineWidthInput";
...
 import SourceTypeSelect from "./SourceTypeSelect";
-import { PlaygroundSettings } from "./types";
+import { PlaygroundSettings, PlaygroundState } from "./types";
+import { Dispatch, SetStateAction } from "react";
+import { createSetter } from "./utils";
 
-interface Props { settings: PlaygroundSettings }
+interface Props {
+	settings: PlaygroundSettings;
+	setPlaygroundState: Dispatch<SetStateAction<PlaygroundState>>;
```
NicholasLYang, line 11: @MichaReiser I'm using the complicated type signature here because I'm making use of the `setPlaygroundState(state => ...)` pattern

website/playground/src/utils.ts
```diff
@@ -38,73 +38,47 @@ export function useWindowSize(): Size {
...
-	);
-	const [isTypeScript, setIsTypeScript] = useState(
-		searchParams.get("typescript") === "true",
-	);
-	const [isJsx, setIsJsx] = useState(searchParams.get("jsx") === "true");
-	const [sourceType, setSourceType] = useState(
-		(searchParams.get("sourceType") as SourceType) ?? SourceType.Module,
-	);
-	const [treeStyle, setTreeStyle] = useState(TreeStyle.Json);
+	const [playgroundState, setPlaygroundState] = useState({
```
MichaReiser, line 46: You may want to move this into a lambda so that the initialize code only runs for the first render:

```
const [..] = useState(() => ({....}))
```

website/playground/src/utils.ts
```diff
@@ -38,73 +38,47 @@ export function useWindowSize(): Size {
...
-		setTreeStyle,
+	return [playgroundState, setPlaygroundState];
+}
+
+export function createSetter(
+	setPlaygroundState: Dispatch<SetStateAction<PlaygroundState>>,
+	field: keyof PlaygroundState,
+) {
+	return function (param: PlaygroundState[typeof field]) {
+		setPlaygroundState((state) => ({ ...state, [field]: param }));
```
MichaReiser, line 83: I think what you're trying to replicate is [`useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer)

The idea is to have one `action` for each field. 

Overall, what you're trying to do here feels complicated with passing down the `setPlaygroundState` and then having a `createSetter`. For someone to understand the code they must first understand what `setPlaygroundState` is, then what `createSetter` does` for something that's otherwise relatively simple. 

I think what you want here is `useContext` which helps you avoid the prop drilling (passing down many props multiple level-deep). 

The provider doesn't need to accept any props because it initializes the state from the URL params. 

They consumer can then expose the state + a method to change a single field 

```
const [{ code }, dispatchPlaygroundAction] = usePlaygroundState();

dispatchPlaygroundAction("change_code", code); 
```

MichaReiser:
We can do the refactor changes as part of another PR. Please update your test plan with a screenshot for Mobile/Desktop.

website/playground/src/utils.ts
NicholasLYang: Ah yeah my bad, forgot to do that

MichaReiser (approved):

PR #2564, opened 2022-05-10
Fix(rome_js_formatter):🐛 fix parenthesis expression bad case

### LINKED ISSUE #2558: Should remove parenthesis, when `member` of ComputedMemberExpression is a `parenthesisExpression`, and `expression` of `parenthesisExpression` is `StringLiteralExpression`
## Input
```js
a[("test")];
```

## Prettier
```js
a["test"];
```
## Rome
```js
a[("test")];
```
https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&typescript=false&jsx=false&sourceType=module#YQBbACgAIgB0AGUAcwB0ACIAKQBdADsA
## Expected
match prettier

IWANABETHATGUY:
## Summary
Fix #2558 

## Test Plan
fix the bad case 
```js
a[("test")]
```

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
@@ -99,24 +99,35 @@ fn is_simple_parenthesized_expression(node: &JsParenthesizedExpression) -> Synta
 fn parenthesis_can_be_omitted(node: &JsParenthesizedExpression) -> SyntaxResult<bool> {
     let expression = node.expression()?;
     let parent = node.syntax().parent();
-    let parent_precedence = FormatPrecedence::with_precedence_for_parenthesis(parent.as_ref());
-    let node_precedence = FormatPrecedence::with_precedence_for_parenthesis(Some(node.syntax()));
 
-    // if expression is a StringLiteralExpression, we could not just return false, here is once case:
+    // if expression is a StringLiteralExpression, we need to check it before precedence comparison, here is an example:
     // ```js
     // a[("test")]
     // ```
+    // if we use precedence comparison, we will get:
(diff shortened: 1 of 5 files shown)


MichaReiser:
I believe this fix is too specific to member expressions. What about computed member assignments? Or any other place where expressions can appear. 

This is prettier's rule for formatting string literal expressions

https://github.com/prettier/prettier/blob/6d02cd9650be29d1b11ab6ff1442eac4f5415c06/src/language-js/needs-parens.js#L507-L528

The way I understand this is that:

* It parenthesis all `StringLiteralExpressions` that are direct children of an `ExpressionStatement`
* Removes parenthesis for all other `StringLiteralExpressions`

This should simplify the logic a bit.

IWANABETHATGUY:
!bench_formatter

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -102,10 +102,23 @@ fn parenthesis_can_be_omitted(node: &JsParenthesizedExpression) -> SyntaxResult<
...
+    // and node_precedence should be `Low` due to expression is StringLiteralExpression.
+    // the parenthesis should be omitted.
+    if parent_precedence > node_precedence
+        && !matches!(
+            expression,
+            JsAnyExpression::JsAnyLiteralExpression(
+                JsAnyLiteralExpression::JsStringLiteralExpression(_)
+            )
+        )
+    {
```
MichaReiser, line 119: The easiest is probably to handle `StringLiteralExpressions` before this line because the precedence isn't relevant for string literals, so there's no point in computing the precedence first.

IWANABETHATGUY:
!bench_formatter

MichaReiser (approved):

Issue #2555, opened 2022-05-07
📎 Prettier Compatibility Metric


MichaReiser:
### Description

Rome's goal is that our formatting matches Prettier's formatting closely. However, it's currently difficult to know if a PR is improving the compatibility or is making things worse. 

## Goal

Define a Prettier compatibility metric and provide means to compute the metric using the current Rome version. 

## Proposal

Percentage of lines that match Prettier's formatting, similar to [git's similarity index](https://stackoverflow.com/a/46258968)

```
compatibility_per_file = matching_lines / MAX(lines_file_1, lines_file_2)
compatibility = SUM(matching_lines) / SUM(MAX(lines_file1, lines_file2))
```

I'm not very proficient at math and the metric might be flawed. Please feel free to propose other metrics. 

## Code Pointers

Our test runner already has an option to generate a Report by setting the `REPORT_PRETTIER` env variable.

https://github.com/rome/tools/blob/db61c4b3a9b3e09811333346ed6d32019bc47dfc/crates/rome_js_formatter/tests/prettier_tests.rs#L270-L297

It should be straightforward to

* Compute the metric for every file and include it in the report
* Compute the metric across all files and print it at the top of the report.

## Stretch

* Create a CI Job similar to the parser conformance that computes the overall metric on the PR branch and on `main` and comments the two numbers together with the difference `PR - main` (percentage the PR improved the compatibility)
* Include a link to the full report (should be possible to commit the report as a gist) in the comment
* Ultimate solution: Compare the metric for every file and report the count of files for which the metric: a) didn't change, b) increased, c): decreased. List the names of the files for which the metric increased/decreased

IWANABETHATGUY:
I would like to have a try.

MichaReiser:
> I would like to have a try.

Awesome. I assigned you the issue. Ping me if something is unclear or if you need some pointers. I recommend approaching this problem step by step (PR by PR) and built out the tools first and test them as CLI before approaching the CI.

IWANABETHATGUY:
I propose another formula to calculate the prettier compatibility, 
```
compatibility_per_file = matching_lines / MAX(lines_file_1, lines_file_2)
compatibility = Sum(compatibility_per_file) / number_of_files
```
There is no conclusion to say which one is better in any case, it dependends.
for example: 
Assume we have three files A, B, C   

---
*match_lines*: 
A: 10000
B: 50
C: 60

---

*MAX(lines_file_1, lines_file_2)*
A: 10010
B: 100
[...]

PR #2574, opened 2022-05-11
feat(rome_js_formatter): 💡 Prettier Compatibility Metric

IWANABETHATGUY:
## Summary
part of #2555


crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -321,7 +329,54 @@ impl DiffReport {
 
             writeln!(report, "```").unwrap();
         }
-
         write("report.md", report).unwrap();
     }
+
+    fn report_metric(&self) {
```
MichaReiser, line 335: Can we write the metric to the normal report instead. For example, emit it at the very top and than in front of every test. 

We may need a different output format if we want to support CI but that's something we can address later.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -321,7 +329,54 @@ impl DiffReport {
...
+            writeln!(report, "# {}", file_name).unwrap();
+            let rome_lines = rome.lines().count();
+
+            let matched_lines = diff_lines(Algorithm::default(), prettier, rome)
+                .iter()
+                .filter(|(tag, _)| matches!(tag, ChangeTag::Equal))
+                .fold(0, |acc, (_, insert_lines)| {
+                    acc + insert_lines.lines().count()
+                });
+            let compatibility_per_file = matched_lines as f64 / rome_lines as f64;
```
MichaReiser, line 352: The definition of the metric is to compute the matching lines over the larger of the two. This is to avoid a 100% match even if the prettier file contains e.g. an additional newline at the end of the file
```suggestion
            let compatibility_per_file = matched_lines as f64 / rome_lines.max(prettier_lines) as f64;
```

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -321,7 +329,54 @@ impl DiffReport {
...
+
+            sum_of_per_compatibility_file += compatibility_per_file;
+            total_line_of_rome += rome_lines;
```
MichaReiser, line 355: Same here, the total lines should take the line count of the larger of the two.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -321,7 +329,54 @@ impl DiffReport {
...
+            total_matched_line_of_rome += matched_lines;
```
MichaReiser, line 356: Nit: The matched line count is the same between rome and prettier. I think `total_matched_lines` should be sufficient.

MichaReiser:
Would you mind creating a Gist and sharing the link of the gist as part of your test plan?

IWANABETHATGUY:
> Would you mind creating a Gist and sharing the link of the gist as part of your test plan?

no, of course

crates/rome_js_formatter/tests/prettier_tests.rs
IWANABETHATGUY: Do you mean  to write the result  to std io  instead file system?

IWANABETHATGUY:
https://gist.github.com/IWANABETHATGUY/41aabdecaa01e684a2ce8e9edd6ee4f2, added here.

crates/rome_js_formatter/tests/prettier_tests.rs
MichaReiser: My suggestion is to merge `report_metric`  and `report_prettier`. Because it should be possible to compute the metric inline and just emit it in addition to the file differences.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -321,7 +329,61 @@ impl DiffReport {
...
+            writeln!(
+                report,
+                "compatibility_per_file: {:.2}%",
+                compatibility_per_file * 100_f64
+            )
+            .unwrap();
+            writeln!(report, "```",).unwrap();
+        }
+        let file_count = state.iter().count();
+        writeln!(
```
leops, line 375: There's going to be a statistical bias here because of how reporting currently works as only the files where our formatter output differs from Prettier are added to the report, so all the files that have 100% compatibility are in fact excluded from this calculation meaning the reported number is lower than what it should actually be

IWANABETHATGUY: So how could I get the actual file_count?

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -300,28 +340,94 @@ impl DiffReport {
...
+                )
+                .unwrap();
+                writeln!(
+                    report,
+                    "compatibility_per_file: {:.2}%",
+                    compatibility_per_file * 100_f64
+                )
+                .unwrap();
+                writeln!(report, "```",).unwrap();
+                writeln!(report, "----").unwrap();
```
MichaReiser, line 405: I would personally not put the metrics into a code block and instead use two paragraphs 

```md
**Prettier Similarity**: 0.50%
```

I also think it's not necessary to print the total number of lines.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -300,28 +340,94 @@ impl DiffReport {
...
-            writeln!(report, "```").unwrap();
+            total_line += rome_lines.max(prettier_lines);
+            total_matched_line += matched_lines;
+            // report prettier compatibility metric for per file
         }
 
+        writeln!(report, "```bash",).unwrap();
+        writeln!(
+            report,
+            "file_based_compatibility: {:.2}%",
```
MichaReiser, line 420: I would use a normal text element here and rename the metric to `Avg Prettier similiarity`. 

```
**Prettier Similarity:: {}
**Average Prettier Similarity**: {}
```


It would be nice if we could print the metric result at the top of the report before it lists any files.

MichaReiser (commented):
Looks good. I've a few suggestions regarding the formatting and naming of the metrics.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -300,28 +340,94 @@ impl DiffReport {
...
+                    }
+                    let line = line.strip_suffix('\n').unwrap_or(line);
+                    writeln!(report, "{}{}", tag, line).unwrap();
+                }
+
+                compatibility_per_file =
+                    matched_lines as f64 / rome_lines.max(prettier_lines) as f64;
+                sum_of_per_compatibility_file += compatibility_per_file;
+                writeln!(report, "```").unwrap();
+                writeln!(report, "----").unwrap();
```
MichaReiser, line 390: I don't think we need the line separator here. The separator is used to separate the results of different files.

IWANABETHATGUY:
https://gist.github.com/IWANABETHATGUY/d67c100cb7dc5c274a497bfb0d16aa68

MichaReiser:
I find the 
> File Based Average Prettier Similarity: 69.29%
> Line Based Average Prettier Similarity: 64.28%

hard to understand. Which is why I would opt for `Prettier similarity` for the "line based" and `Average prettier similarity` for the file based. 

I think these terms are more intuitive because the common definition of an average is that you take the value for each file and divide it by the number of files, which is exactly how the metric is defined. 

The total Prettier Similarity metric then remains a bit more obscure but that's already the case with the per file on which is why I believe that's fine. 

I think it would help to either:
* Include a link to the issue in the report where people can look up the metric definition
* Or include the metric definitions in the report.

IWANABETHATGUY:
> Awesome. This is getting sooo good!
> 
> I find the
> 
> > File Based Average Prettier Similarity: 69.29%
> > Line Based Average Prettier Similarity: 64.28%
> 
> hard to understand. Which is why I would opt for `Prettier similarity` for the "line based" and `Average prettier similarity` for the file based.
> 
> I think these terms are more intuitive because the common definition of an average is that you take the value for each file and divide it by the number of files, which is exactly how the metric is defined.
> 
> The total Prettier Similarity metric then remains a bit more obscure but that's already the case with the per file on which is why I believe that's fine.
> 
> I think it would help to either:
> 
> * Include a link to the issue in the report where people can look up the metric definition
> * Or include the metric definitions in the report.

https://gist.github.com/IWANABETHATGUY/148d1c973aa3e5b604159b6f5890d538

IWANABETHATGUY:
@NicholasLYang , It seems our Ci has failed due to `crates/website/playgroud/lib.rs`, would you mind making a pull request to fix it?

MichaReiser:
Awesome. Thank you. This metric will help me a lot when refactoring some of the printer behaviour to track if the changes are improving compatibility or not. 

Are you interested in tackling the CI integration (writing the information to a JSON file and having a command to compare the results) next?

MichaReiser (approved):

IWANABETHATGUY:
> 
Would the json file need the diff info?

MichaReiser:
I see this as a multip step process, similar to how the parser coverage works

* The `DiffReporter` has an option to emit a JSON instead of a md file
* There's a new `xtask` command that given two paths to such report JSON files prints a nice table with the diff (see `xtask coverage compare`)

PR #2576, opened 2022-05-11
feat(lsp): improve the user experience of the analyzer in the editor

leops:
## Summary

This is small followup PR for #2491 fixing some issues with the general user experience of the analyzer (diagnostics and code fixes) through the editor integration (LSP). Changes include:

- Adding a markup-based message to all code actions and verifying they have the correct categories applied (in the context of the editor integration the markup is converted to plain text as VSCode does not support formatting for diagnostic messages and code action labels, but the markup will be displayed correctly in the eventual CLI integration)
- Moving the parsing of suppression comments from `rome_js_formatter` to `rome_js_syntax` and using it to support suppressions in the analyzer as well. At the moment it only supports suppressing the analysis entirely, but eventually the traversal process will dynamically track a stack of `AnalysisFilter` to support disabling only certain rules in a given subtree
- Filtering out code actions categorized as "refactor" for a given range when there are also actions categorized as "quick fix" for outstanding diagnostics in the same range (this is to make easier for users to apply autofixes through the in-editor code action menu, it avoids having these fixes drowned out in lower priority refactor opportunities)

Screen capture of the in-editor diagnostic + quick fix workflow:
diagnostic_fix


crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -134,239 +135,8 @@ pub(crate) fn format_head_body_statement(
...
-        assert_eq!(
-            parse_suppression_comment("// rome-ignore format lint: explanation")
-                .collect::<Vec<_>>(),
-            vec![Suppression {
-                categories: vec![("format", None), ("lint", None)],
-                reason: "explanation"
-            }],
-        );
-    }
+    has_suppressions_category("format", node)
```
MichaReiser, line 139: Could we make this category an enum?

crates/rome_js_syntax/src/suppression.rs
```diff
@@ -0,0 +1,237 @@
...
+/// line (lint false-positive, specific formatting requirements, ...)
+#[derive(Debug, PartialEq, Eq)]
+pub struct Suppression<'a> {
+    /// List of categories for this suppression
+    ///
+    /// Categories are pair of the category name +
+    /// an optional category value
+    pub categories: Vec<(&'a str, Option<&'a str>)>,
+    /// Reason for this suppression comment to exist
+    pub reason: &'a str,
```
MichaReiser, line 26: Nit: I'm wondering if the category, as well as the reason, should be computed lazily rather than allocating a `Vec`. For example, `Suppression` could have a method `categories()` returning an Iterator and `reason()` returning the string slice. 

The main issue that I see is that `reason()` requires parsing out the categories first. But is the reason even used today? 

The main reason for this suggestion is that the category check is part of the "hot path" in the formatter and in the analyzer and avoiding allocating a vector for each category could improve perf quite a bit. An alternative might be to use a small vec instead because the most common case is that suppressions have a single category.

MichaReiser (approved):

crates/rome_lsp/src/handlers/analysis.rs
```diff
@@ -63,5 +63,29 @@ pub(crate) fn code_actions(
...
+        }
+    });
+
+    if has_fixes {
+        let mut index = 0;
+        while index < result.len() {
+            if let CodeActionOrCommand::CodeAction(action) = &result[index] {
+                if action.diagnostics.is_none() {
+                    result.remove(index);
+                    continue;
```
IWANABETHATGUY, line 82: Would it be better if we use `result.iter().filter())` to filter the action with empty diagnostics, which could reduce `bound checks`.

leops: Even better I can just use `result.retain(|action| action.diagnostics.is_some())`, I didn't use it because I though it was unstable but after checking I just had it confused with `drain_filter`

crates/rome_js_syntax/src/suppression.rs
leops: At the moment I implemented this as a single step for simplicity, however as the parsing code doesn't go into allocating until it positively finds a valid suppression comment and we shouldn't have too many of those in a given file (or that would mean we're doing a pretty bad job as a formatter / linter ...) in practice I wouldn't expect the overhead to be too significant

crates/rome_js_formatter/src/utils/mod.rs
leops: I think I initially thought of this `has_suppressions_category` function as temporary (since the analyzer will eventually want to read out the name of the rule from the comment) so I didn't put too much thought into it, but that doesn't mean I can't make the current version better

crates/rome_lsp/src/handlers/analysis.rs
leops: Yeah that's why `retain` is better than removing in a loop as it should perform the compaction of the underlying memory in a single step instead of repeatedly shifting the tail of the vector

PR #2580, opened 2022-05-12
feat(rome_diagnostic): refactor diagnostics to use markup

leops:
## Summary

This PR changes the internal representation of title, label and suggestion messages in diagnostics to use markup (`MarkupBuf`) instead of plain strings. This change is mostly transparent to the producers of diagnostics (the trait bounds for messages has been changed to `impl rome_console::fmt::Display` which is a superset of the previous `impl Into<String>`) as well as the consumers (diagnostics are already printed as markup, so whether they use strings or markup internally is an implementation detail).

This change removes the need for the intermediate `AnalyzerDiagnostic` struct (analyzer rules now return a plain diagnostic object), however the `AnalyzerAction` struct is still needed as it contains additional information not present in `CodeSuggestion` (but it's now possible to transparently convert the former to the later with `.into()`)

## Test Plan

The tests snapshots for the analyzer have been updated to use the mainline diagnostic printing. The tests snapshots for the parser are unchanged since these diagnostics are represented as markup too but do not make use of it yet.

### DIFF
--- a/crates/rome_analyze/src/analyzers/no_delete.rs
+++ b/crates/rome_analyze/src/analyzers/no_delete.rs
@@ -1,5 +1,5 @@
 use rome_console::markup;
-use rome_diagnostics::Severity;
+use rome_diagnostics::{Applicability, Severity};
 use rome_js_factory::make;
 use rome_js_syntax::{
     JsAnyAssignment, JsAnyAssignmentPattern, JsAnyExpression, JsAnyRoot,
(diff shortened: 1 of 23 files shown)


leops:
!bench_parser

leops:
I would expect this change to incur a slight slowdown of the parser as the markup requires one more nested allocation but the benchmarks number aren't very good at showing if that's the case since I imagine those libraries don't have many diagnostics (if they even have some)

MichaReiser (approved):

PR #2583, opened 2022-05-13
feat(rome_cli): implement the `check` and `ci` CLI commands

leops:
## Summary

This PR introduces two new commands to the CLI:
- `check` parses and runs the analyzer on all the provided files (or recursively traverses directories) and exits with a non-zero error code if diagnostics were emitted with a severity of `Error` or higher
- `ci` performs the same checks but also runs the files through the formatter and errors if the original source code differs from the formatted output (this replaces the previous `rome format --ci` option)

I implemented this by moving out the existing traversal behavior implemented for the `format` command into a generic `traverse` function and expanding the previous `FormatMode` enum into a `TraverseMode` with variants `Check`, `Format` and `CI` to selectively enable the checking behavior, formatting behavior or both in the processing of individual file while keeping most of the boilerplate code around this processing shared.

One shortcoming of the way the CLI currently works is that all formatter options have to be explicitly specified on the `ci` command the same way they already are on the `format` command, but this will eventually go away when we implement a centralized configuration file format.

## Test Plan

I refactored the existing test cases for the CLI and sorted them into modules based on the command they are testing, and created additional tests for the new commands.


crates/rome_cli/src/traversal.rs
```diff
@@ -0,0 +1,571 @@
...
+                        diag.suggestions.push(action.into());
+                    }
+
+                    diagnostics.push(diag);
+                }
+            });
+        }
+
+        // Return early if the file has error diagnostics, there's a good
+        // change it would cause formatting errors
```
xunilrj, line 439: good chance?

crates/rome_cli/src/traversal.rs
```diff
@@ -0,0 +1,571 @@
...
+                },
+            )
+        },
+    );
+
+    let count = processed.load(Ordering::Relaxed);
+    let skipped = skipped.load(Ordering::Relaxed);
+
+    match mode {
+        TraversalMode::Check | TraversalMode::CI { .. } => {
```
xunilrj, line 94: ```TraversalMode::CI``` mode do both, right? check and format/compare.

Doesn't make sense to print both here? Otherwise will be hard to know, by only looking, if the formatter run or not.

```
Checked and Compared 123 files in 00:00:01
```

crates/rome_analyze/src/signals.rs
```diff
@@ -30,19 +30,32 @@ pub struct AnalyzerAction {
     pub applicability: Applicability,
     pub message: MarkupBuf,
     /// Range of the original document being modified by this action
-    ///
-    /// By default this is conservatively set to cover the entire document
-    pub range: TextRange,
+    pub prev_range: TextRange,
+    /// Range of the new document that differs from the original document
+    pub next_range: TextRange,
```
MichaReiser, line 35: Nit: My understanding of `prev` and `next` is normally in context of ordered items, where the `prevous` range is the range of the item coming before the `next` range. This isn't what's meant here. The range's here (as mentioned in the documentation) refer to a nodes `original` position and the transformed/new position. 

That's why I would be leaning towards renaming the variables to `original_range`, and `new_range`/`transformed_range`.

crates/rome_analyze/src/signals.rs
```diff
@@ -30,19 +30,32 @@ pub struct AnalyzerAction {
...
 }
 
 impl From<AnalyzerAction> for CodeSuggestion {
     fn from(action: AnalyzerAction) -> Self {
+        // Only print the relevant subset of tokens
+        let mut code = String::new();
+
+        for token in action.root.syntax().descendants_tokens() {
+            let range = token.text_range();
+            if range.end() <= action.next_range.start() || range.start() >= action.next_range.end()
```
MichaReiser, line 46: Would it be possible to use `range.intersect(action.next_range()).is_some()`?

crates/rome_analyze/src/signals.rs
```diff
@@ -101,14 +114,66 @@ impl<'a, R: Rule> AnalyzerSignal for RuleSignal<'a, R> {
...
+                message: action.message,
+                prev_range,
+                next_range,
+                root: action.root,
+            })
         })
     }
 }
+
+/// Compares the tokens that make up the two trees and find the narrowest text
```
MichaReiser, line 134: My understanding is that tokens are the leave of a tree, so they are "single node" sub-trees. That's why it's unclear to me what the meaning of *two trees* is in this documentation. 

Can we document the invariant that `prev` and `next` must start at the same position.

crates/rome_analyze/src/signals.rs
```diff
@@ -101,14 +114,66 @@ impl<'a, R: Rule> AnalyzerSignal for RuleSignal<'a, R> {
...
+            TextRange::new(start, prev_end),
+            TextRange::new(start, next_end),
+        )),
+        (Some(start), None) => Some((
+            TextRange::new(start, prev.text_range().end()),
+            TextRange::new(start, next.text_range().end()),
+        )),
+
+        (None, None) => None,
+        (None, Some(_)) => unreachable!(),
```
MichaReiser, line 177: Nit: Document why this is unreachable?

crates/rome_cli/src/commands/help.rs
```diff
@@ -3,10 +3,34 @@ use crate::Termination;
...
+USAGE:
+    rome ci [OPTIONS] <INPUTS...>
+
+    INPUTS can be one or more filesystem path, each pointing to a single file or an entire directory to be searched recursively for supported files
+
+OPTIONS:
+    --indent-style <tabs|space>   Determine whether the formatter should use tabs or spaces for indentation (default: tabs)
+    --indent-size <number>        If the indentation style is set to spaces, determine how many spaces should be used for indentation (default: 2)
+    --line-width <number>         Determine how many characters the formatter is allowed to print in a single line (default: 80)
+    --quote-style <single|double> Determine whether the formatter should use single or double quotes for strings (default: double)
```
MichaReiser, line 31: Nit: It would be nice if we could avoid duplicating the documentation for the options. Might be nice to follow up on a way to re-use options as a separate PR

crates/rome_cli/src/commands/help.rs
```diff
@@ -16,7 +40,6 @@ USAGE:
 
 OPTIONS:
     --write                       Write the output of the formatter to the files instead of printing the diff to the console
-    --ci                          Enable CI mode, lock files and exit with an error if the formatter would modify them
```
MichaReiser, line 19: Please make sure to mention this breaking change in our release issue.

MichaReiser (approved):

crates/rome_analyze/src/signals.rs
leops: There's a subtle difference in the condition there: it's a bit hard to read since this check is inverted but it's actually `range.end > action.start && range.start < action.end`, while using `intersect` alone would amount to `range.end >= action.start && range.start <= action.end`. The equality checks are significant as it means the printed string would include an additional token at the start and end of the range, so in order to use `intersect` though I think the condition would have to be `range.intersect(action).filter(|range| !range.is_empty()).is_some()`

PR #2585, opened 2022-05-13
feature(formatter): Group list elements

MichaReiser:
I noticed that Rome doesn't group the items of list. This is problematic
because the printer then breaks the whole list if one element doesn't fit on a single line.

Before:

```js
import {
	// some funky comment
	loooooooooooooooooooong
	as
	moreeeeeeloooooooooooooooooooong,
	loooooooooooooooooooong2
	as
	moreeeeeeloooooooooooooooooooong2,
	// some other comment
} from "test";
```

After:

```
import {
	// some funky comment
	loooooooooooooooooooong as moreeeeeeloooooooooooooooooooong,
	loooooooooooooooooooong2 as moreeeeeeloooooooooooooooooooong2,
	// some other comment
} from "test";
import {
	loooooooooooooooooooong3 as moreeeeeeloooooooooooooooooooong3,
	// some funky comment
} from "test";
```

This PR wraps the formatted element inside of `format_separated` and
  in some other places where Rome does manual list formatting. There may be more places but this should be a good start.

## Test Plan

`cargo test`

I'm a bit surprised that not more files are changing but noticed that there are plenty of cases when I applied these changes to my local PR that changes how our printer determines if some content "fits".

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -728,11 +728,15 @@ pub fn group_elements_with_options(
     content: FormatElement,
     options: GroupElementsOptions,
 ) -> FormatElement {
-    let (leading, content, trailing) = content.split_trivia();
+    if content.is_empty() {
+        content
+    } else {
+        let (leading, content, trailing) = content.split_trivia();
 
-    let group = Group::new(content).with_id(options.group_id);
+        let group = Group::new(content).with_id(options.group_id);
 
-    format_elements![leading, group, trailing]
+        format_elements![leading, group, trailing]
+    }
 }
 
 /// IR element that forces the parent group to print in expanded mode.
@@ -1630,8 +1634,8 @@ impl FormatElement {
 
     /// Splits off the leading and trailing trivias (comments) from this [FormatElement]
     ///
-    /// For [FormatElement::HardGroup] and [FormatElement::Group], the trailing and leading trivias
-    /// are automatically moved  outside of the group. The group itself is then recreated around the
+    /// For [FormatElement::HardGroup], the trailing trivia
+    /// is automatically moved  outside of the group. The group itself is then recreated around the
     /// content itself.
     pub fn split_trivia(self) -> (FormatElement, FormatElement, FormatElement) {
         match self {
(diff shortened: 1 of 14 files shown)


MichaReiser:
!bench_formatter

crates/rome_formatter/src/format_element.rs
```diff
@@ -1524,11 +1529,6 @@ impl FormatElement {
                 (leading, hard_group_elements(content), trailing)
             }
 
-            FormatElement::Group(group) => {
```
MichaReiser, line 1527: This check isn't necessary. `group_elements` moves all trivia out of the `group` inside a `list`. That's why calling `split_trivia` on the group always returns `[empty, group(content), empty]`

This is different for `hard_groupelements` that has the trailing comments as part of the group.

crates/rome_js_formatter/tests/specs/js/module/import/bare_import.js.snap
```diff
@@ -39,15 +39,15 @@ import "short" assert { type: "json" };
 
 import "very_long_import_very_long_import_very" assert {
 	// something good is here
-	"type": "json", /****/
+	"type": /****/ "json",
```
leops, line 42: I think block comments are one of the main reasons why these elements were not wrapped in groups before, since it's what allows the "line suffix if group breaks" behavior to work. But as we want to align with Prettier as much as possible we could decide to simply get rid of that, since I don't think they shuffle comments around as much as we currently do.

MichaReiser: Yeah, comments are a big topic and one of the main issues that I'm facing during adopting Prettier's behaviour and I'm leaning towards removing the special handling of comments inside of `Group` and `HardGroup` because it often happens that comments belong to a new node after formatting which causes issues when formatting the same node a second time because it now uses a different logic. 

My understanding of Prettier's logic right now is that they only distinguish between leading and trailing comments of a node, and not on a per token level. That's easier to reason about than our logic. What I don't understand yet is how they avoid the moving of comments between nodes to prevent that comments get formatted differently in a second format pass.

MichaReiser: Does wrapping the elements into groups cause any stability issues in your view? If so, why is it that this only happens for array elements but not for any other nodes?

MichaReiser:
!bench_formatter

MichaReiser:
Part of #2597

PR #2588, opened 2022-05-16
feature(rome_js_parser): coverage test suite for ts symbols file

xunilrj:
see https://github.com/rome/tools/issues/2488

This PR parses Microsoft/ts .symbols file;

#### Microsoft ts .symbols file

We are running only files without any errors.

```
symbols/msts: Ran 5946 tests in 19.05s
│      5946 │   1613 │   4328 │      5 │    27.13 │
```

All ```parse_*``` functions are simplified versions of ```nom```, just to avoid including another crate.

#### Fake Implementation

```Symbol``` and ```symbols``` are fake and only exist to make the coverage test to compile and run.

### DIFF
--- a/xtask/coverage/src/symbols/msts.rs
+++ b/xtask/coverage/src/symbols/msts.rs
@@ -55,36 +55,39 @@ impl TestCase for SymbolsMicrosoftTestCase {
             .collect();
         actual.sort_by_key(|x| x.range().start());
 
-        if std::env::var("PRINT_CMP").is_ok() {
-            let mut expecteds = expected.symbols.iter();
-            let mut actuals = actual.iter();
-            loop {
-                let e = expecteds.next();
-                let a = actuals.next();
+        // Print to debug! detailed information
+        // on symbols that are different from the
+        // expected
+        let mut expecteds = expected.symbols.iter();
+        let mut actuals = actual.iter();
+        loop {
+            let expected = expecteds.next();
+            let actual = actuals.next();
 
-                if e.is_none() && a.is_none() {
-                    break;
-                }
-
-                if let Some(s) = e {
-                    print!("{}", s.name);
-                }
-
-                print!(" - ");
-
-                if let Some(s) = a {
-                    print!("{:?}", s);
-                }
-
-                match (e, a) {
-                    (Some(e), Some(a)) if e.name != a.name() => {
-                        println!(" <<<<<<<<<<<<<<<<<<<< Diff here")
-                    }
-                    _ => {}
-                }
-
-                println!();
+            if expected.is_none() && actual.is_none() {
+                break;
             }
+
+            let mut debug_text = String::new();
+
+            if let Some(symbol) = expected {
+                debug_text.push_str(&symbol.name);
+            }
+
+            debug_text.push_str(" - ");
+
+            if let Some(symbol) = actual {
+                debug_text.push_str(&format!("{:?}", symbol));
+            }
+
+            match (expected, actual) {
+                (Some(e), Some(a)) if e.name != a.name() => {
+                    debug_text.push_str(" <<<<<<<<<<<<<<<<<<<< Diff here");
+                }
+                _ => {}
+            }
+
+            tracing::debug!("{}", debug_text);
         }
 
         if expected.symbols.len() != actual.len() {
--- a/xtask/coverage/src/symbols/msts.rs
+++ b/xtask/coverage/src/symbols/msts.rs
@@ -160,36 +160,15 @@ struct SymbolsFile {
     symbols: Vec<Symbol>,
 }
 
-fn parse_decl(input: &str) -> Option<(&str, Decl)> {
-    let (input, _) = parse_str(input, "Decl")?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, _) = parse_str(input, "(")?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, file) = parse_until_chr(input, |x| x.is_whitespace() || x == ',')?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, _) = parse_str(input, ",")?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, row_start) = parse_until_chr(input, |x| x.is_whitespace() || x == ',')?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, _) = parse_str(input, ",")?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, col_start) = parse_until_chr(input, |x| x.is_whitespace() || x == ')')?;
-    let (input, _) = parse_whitespace0(input);
-    let (input, _) = parse_str(input, ")")?;
-    Some((
-        input,
-        Decl {
-            file: file.to_string(),
-            row_start: row_start.parse().ok(),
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by MichaReiser** [comment]
I would find it useful if you could write up some overall architecture where the work on symbols and scopes is going. It's otherwise difficult to review the PR from an architectural point and only allows me to point out coding issues. 

My understanding of the problem is and things that are important to know are:

## Symbols

* It should be possible to resolve the symbol for a node, e.g. by querying a semantic model
* A symbol allows to 
  * query all its declarations and accessing the nodes (e.g. in case of overloading)
  * query all its references/uses and accessing the nodes
  * querying its child symbols. For example, to get all members of a class. This raises the point that languages have different symbols. What representation do we use? Somethin similar to `SyntaxNode<L>` and `AstNode` or is `JsSymbol` an enum
  * querying the parent symbol if any

## Scope

I'm deliberately listing scopes independently from `Symbols` because there are different requirements to it and symbols also exist for things that aren't part of a scope. 

* It should be possible to get a node's scope. For example, to suggest another variable name in case of a re-declaration (or to fix a reference if the name was misspelt). 
* It should be possible to get all symbols from a scope. Useful when introducing a new variable to make sure there are no name clashes. 
* It should be possible to get the parent scope
* Maybe: It should be possible to get all child scopes

## Semantic Model

I see it as likely that we need some form of a semantic model that is the entry point for anyone querying such information. The main question here is how does the API look like and how do we intend to build it up (individual method calls or a single method that builds the whole semantic model). 

Having this outline will allow you to describe in the PR summary which problem you are solving and how it helps build the final solution. 

I think it would further help to draft an issue that splits this problem into more tangible sub-problems and outline the work involved for each, so that someone else could help with the work or step in in case you are absent.

**Comment by xunilrj** [comment]
> I would find it useful if you could write up some overall architecture where the work on symbols and scopes is going.

I created this discussion to address the high-level architecture (https://github.com/rome/tools/discussions/2603) and this issue (https://github.com/rome/tools/issues/2488) to track the implementation details on a lower level.

**Comment by xunilrj** [comment]
> What's the difference between `xtask/coverage/src/{mod.rs,msts.rs,utils.rs}` and `xtask/coverage/src/symbols/{mod.rs,msts.rs,utils.rs}` ? It looks like those files are duplicated

You are right. I probably copied and pasted it without realizing it. Fixed.

File: xtask/coverage/src/symbols/msts.rs [hunk 1]
```diff
@@ -0,0 +1,254 @@
...
+        let t = TestCaseFiles::single(self.name.clone(), code.clone(), SourceType::tsx());
+
+        let r = rome_js_parser::parse(&code, 0, SourceType::tsx());
+        let mut actual: Vec<_> = rome_js_parser::symbols::symbols(r.syntax())
+            .into_iter()
+            .filter(|x| !x.name().contains('\"') && !x.name().contains('\''))
+            .collect();
+        actual.sort_by_key(|x| x.range().start());
+
+        if std::env::var("PRINT_CMP").is_ok() {
```
ematipico, line 58: Where and when do we use this environment variable? Maybe we should update the `CONTRIBUTING.md` file?

xtask/coverage/src/symbols/msts.rs
```diff
@@ -0,0 +1,254 @@
...
+                files: t,
+                errors: vec![],
+            }
+        } else {
+            for (expected, actual) in expected.symbols.iter().zip(actual) {
+                let are_names_eq = expected.name == actual.name();
+                // let are_paths_eq = expected.name == actual.name;
+                //TODO check decls
+                if !are_names_eq
+                /*|| !are_paths_eq*/
```
ematipico, line 101: To remove?

xtask/coverage/src/symbols/msts.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        if expected.symbols.len() != actual.len() {
+            TestRunOutcome::IncorrectlyErrored {
+                files: t,
+                errors: vec![],
+            }
+        } else {
+            for (expected, actual) in expected.symbols.iter().zip(actual) {
+                let are_names_eq = expected.name == actual.name();
+                // let are_paths_eq = expected.name == actual.name;
```
ematipico, line 98: To remove?

xtask/coverage/src/symbols/msts.rs
```diff
@@ -0,0 +1,254 @@
...
+                //TODO check decls
```
ematipico, line 99: Will this be done in another PR?

xtask/coverage/src/symbols/msts.rs
```diff
@@ -0,0 +1,254 @@
...
+        input,
+        Decl {
+            file: file.to_string(),
+            row_start: row_start.parse().ok(),
+            col_start: col_start.parse().ok(),
+        },
+    ))
+}
+
+/// see xtask\coverage\Typescript\src\harness\typeWriter.ts
```
ematipico, line 190: Where's this file? Maybe it's best to write a better documentation? At least, for example what this function is doing on a high level?

xtask/coverage/src/symbols/utils.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    if qty > 0 {
+        Some((&input[qty..], &input[0..qty]))
+    } else {
+        None
+    }
+}
+
+pub fn parse_whitespace0(input: &'_ str) -> (&'_ str, &'_ str) {
```
ematipico, line 25: Why the `0` in the name?

ematipico (approved):

xtask/coverage/src/symbols/msts.rs
xunilrj: Only here.
I can rename it and document it.

xtask/coverage/src/symbols/msts.rs
xunilrj: These are more todos from future PRs. But I can remove them for now.

xtask/coverage/src/symbols/msts.rs
xunilrj: Yes.

xtask/coverage/src/symbols/utils.rs
xunilrj: Because it accepts no (zero) whitespaces. It is following nom's convention: https://docs.rs/nom/latest/nom/character/complete/fn.space0.html

xtask/coverage/src/symbols/msts.rs
xunilrj: Instead of having an obscure environment variable that needs documentation, I changed the code to print to ```debug!```. And, of course, it can be configured normally with ```RUST_LOG```, if needed.

I think this makes more sense.

xtask/coverage/src/symbols/msts.rs
xunilrj: Done.

xtask/coverage/src/symbols/msts.rs
xunilrj: Done.

xtask/coverage/src/symbols/msts.rs
xunilrj: Done.

Issue #2579, opened 2022-05-12
📎 Printer & IR Capabilities


MichaReiser:
### Description

The goal of this task is to track missing Formatter IR elements and behaviour differences to Prettier. 

## Different Semantics
* [x] Group printing:`fits` #2598
* [x] Fill

## Rome only IR

Evaluate if indeed necessary

* [x] HardGroup #2597 

## Missing IR Elements
The goal isn't to implement all these IR elements. We should only implement them if necessary 

- [ ] `align`/ `dedent` / `dedentToRoot`, `markAsRoot`
- [ ] `indentIfBreak`: Optimized version of `ifBreak(indent)` unclear what’s optimized about it because it creates an `indent` element inside of the printer.
- [x] Conditional Groups (Alternatives) #2529 #2591
- [x] `LineSuffixBoundary` #2590
- [x] `BreakParent` #2590
    - Forces to break the parent group (except if it’s a conditional group with expanded states)
    - Implemented as a traversal before the print, navigates through the document and sets `group.break` to `propagated` on the parent group
- [x] `ifBreak`/ `group` with `groupId` #2589
    - Allows adding conditional logic depending on if a specific parent group breaks (not necessarily the direct parent)
- [ ] `trim`

MichaReiser:
Here some notes from my research

# Prettier IR / Printer Differences

Tags: Projects

This document describes the differences between Rome’s and Prettier’s printer and IR.

# Differences

## Group

### `fits`

Rome’s `fits` tries to print the whole group including all it’s content on a single line. The group fits if all tokens up the last token of the outmost “flat” group fit inside of the print width.

This is different to Prettier’s `fits` definition because

- A hard-line break doesn’t break a group, instead:
- a group fits if the content of the group and any content following the group up to the first hard line break fits on the line.
- Flat groups can contain expanded groups (with `break: true`).

The definition is the same in regards that if a child group expands, then the parent groups expand and the printer tries to re-print with the outer-most group expanded.

```jsx
let document1 = builders.group([
	"A somewhat long string that must be ",
	builders.softline,
	"grouped",
	builders.hardline,
	builders.group("grouped and should exceed the print with of 86"),
]);
```

Prints to 

```jsx
A somewhat long string that must be grouped
grouped and should exceed the print with of 86
```

vs Rome

```jsx
A somewhat long string that must be 
grouped
grouped and should exceed the print with of 86
```

### `break`

[...]

PR #2589, opened 2022-05-17
refactor(formatter): Refactor `try_print_flat` to `fits_on_line`

MichaReiser:
This PR brings the Printer closer to Prettier's printer by introducing a `fits_on_line` function that measures if printing this element won't exceed the line width.

This is somewhat similar to the existing `try_print_flat` implementation but differs in the sense that it also supports
measuring content that isn't printed in flat mode.

This functionality will be needed to implement alternatives because
the alternatives picks the first variant that fits on the line but
an alternative itself doesn't have to be printed in flat mode.

The PR doesn't change the semantics of `fits` yet so that only fits if the group's content fit as well as the remaining content in the document before the first new line character. 

This PR further introduces the concept of a `GroupId` and adds the ability to use `if_group_breaks` and `if_group_fits_on_line` with a specific group id. This is necessary because this PR aligns the `fill_elements` semantic with Prettier's by introducing an implicit group for each element. The problem of doing this is that the trailing comma insertion no longer worked because the element itself fits on the line (printed in flat mode) but the outer group of the array did break:

```
group([
  '[',
  indent([
    soft_line_break(),
    fill([
      [
        token('2222222'),
        if_group_breaks(',') // Doesn't get printed because the element fits
      ]
  ]),
  soft_line_break(),
])
```


crates/rome_formatter/src/printer/printer.rs
```diff
@@ -0,0 +1,1203 @@
...
+                        PrintMode::Flat
+                    }
+                    // The printer is either in expanded mode or it's necessary to re-measure if the group fits
+                    // because the printer printed a line break
+                    _ => {
+                        // Measure to see if the group fits up on a single line. If that's the case,
+                        // print the group in "flat" mode, otherwise continue in expanded mode
+
+                        let flat_args = args.with_print_mode(PrintMode::Flat);
+                        // TODO replace `ElementCallQueue::default` with the real queue to also measure the rest of the document
```
MichaReiser, line 129: I'll follow up on this TODO in #2579 Passing the rest of the document will make sure that e.g. the following snipped breaks if the `]` doesn't fit on the line:

```
[a, b, c]
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -0,0 +1,1203 @@
...
+                        PrintMode::Flat
+                    }
+                    // The printer is either in expanded mode or it's necessary to re-measure if the group fits
+                    // because the printer printed a line break
+                    _ => {
+                        // Measure to see if the group fits up on a single line. If that's the case,
+                        // print the group in "flat" mode, otherwise continue in expanded mode
+
+                        let flat_args = args.with_print_mode(PrintMode::Flat);
+                        // TODO replace `ElementCallQueue::default` with the real queue to also measure the rest of the document
```
MichaReiser, line 132: This is part of #2579 Passing in the `queue` will make sure that `[a, b, c]` only fits if the `]` also fits on the line even if it's outside of the group.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -0,0 +1,1203 @@
...
+impl PrintElementArgs {
+    pub fn new(indent: u16) -> Self {
+        Self {
+            indent,
+            ..Self::default()
+        }
+    }
+
+    pub fn with_incremented_indent(mut self) -> Self {
+        self.indent += 1;
```
MichaReiser, line 482: Some comments have moved in this PR because `with_incremented_indent` only incremented the `indent` but didn't copy over the other fields like `hard_group`.

MichaReiser:
!bench_formatter

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -0,0 +1,1177 @@
...
+    }
+
+    /// Tries to fit as much content as possible on a single line.
+    /// Each item forms a virtual group that is either printed in flat or expanded mode.
+    /// It handles three different cases:
+    ///
+    /// * The first and second content fit on a single line. It prints the content and separator in flat mode.
+    /// * The first content fits on a single line, but the second doesn't. It prints the content in flat and the separator in expanded mode.
+    /// * Neither the first nor the second content fit on the line. It brings the first content and the separator in expanded mode.
+    fn print_fill(
```
MichaReiser, line 287: @NicholasLYang it should be straightforward for you to change the `fill` semantics so that it supports custom separators. Something I understand you need for JSX. This is prettier's implementation 

https://github.com/prettier/prettier/blob/bb130d2d6e6ea42629b17a57869f5e7e550df6fd/src/document/doc-printer.js#L415-L486

Make sure to also change the fits handling inside `fits_on_line`

MichaReiser:
!bench_formatter

crates/rome_formatter/src/format_element.rs
```diff
@@ -1181,26 +1252,103 @@ impl Deref for List {
...
-pub enum GroupPrintMode {
+#[derive(Clone, Copy, Eq, PartialEq, Hash)]
+pub struct DebugGroupId {
+    id: NonZeroU32,
+    name: &'static str,
+}
+
+impl DebugGroupId {
+    pub fn new(debug_name: &'static str) -> Self {
+        static ID: AtomicU32 = AtomicU32::new(1);
```
leops, line 1293: I think it would be better to have the group ID allocation take place in the `Formatter`, in a long running process like the language server this static counter will eventually overflow and cause the `NonZeroU32` constructor below to panic (it sure would take a huge number of formatting runs for it to reach 2^32, but I'm also probably not the only one that never quits my editor and always just put my laptop to sleep, so it's not entirely unrealistic either ...)

MichaReiser: That makes sense. I forgot that this will be across all files. I considered a `u32` to be sufficient for a single formatting run... So yes, let's move this to the formatter.

crates/rome_formatter/src/printer/mod.rs
MichaReiser: `MeasureQueue` uses an offset into the queue to get the current position. It never actually dequeues.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1689,19 +1839,32 @@ mod tests {
...
-        assert_eq!(24, std::mem::size_of::<List>());
-        // Increasing the size of FormatElement has serious consequences on runtime performance and memory footprint.
-        // Is there a more efficient way to encode the data to avoid increasing its size? Can the information
-        // be recomputed at a later point in time?
-        // You reduced the size of a format element? Excellent work!
-        assert_eq!(32, std::mem::size_of::<FormatElement>());
-    }
 }
+
+#[cfg(target_pointer_width = "64")]
```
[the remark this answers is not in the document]
MichaReiser: The macro is only defined in that configuration but I guess I can remove that restrictions

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -0,0 +1,1177 @@
...
+    pending_space: bool,
+    measured_group_fits: bool,
+    generated_line: usize,
+    generated_column: usize,
+    line_width: usize,
+    has_empty_line: bool,
+    line_suffixes: Vec<PrintElementCall<'a>>,
+    verbatim_markers: Vec<TextRange>,
+    /// Tracks the mode in which groups with ids are printed.
+    group_modes: HashMap<GroupId, PrintMode>,
```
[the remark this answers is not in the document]
MichaReiser: You mean by iterating over the groups instead? The challenge is that there's currently no "exit" event when the printer's done printing a group except if we use `print_all` but that makes the printer recursive which we want to avoid when possible. 

The only other alternative I see is to introduce a new IR element for this which is something I want to defer for now but would play into the *flat element buffer* proposal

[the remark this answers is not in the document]
leops: I purely meant that using a `Vec` to store the group modes by using the GroupId as an index has constant-time lookups (and straightforward amortized insertion), and would be more efficient than using a `HashMap` (but it requires actually having small, densely allocated numerical keys or the memory usage will blow up)

crates/rome_formatter/src/format_element.rs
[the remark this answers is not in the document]
MichaReiser: The checks don't seem to work for wasm. So I'll leave it as-is for now.

MichaReiser:
Addressed the review feedback:
* The `PrinterState` now uses a `Vec<Option<GroupId>>` 
* Moved the creation of `GroupId`s to the `formatter so that the ids are scoped to a single document
* Rewrote `fill` loop to use a `for`

leops (approved):

PR #2590, opened 2022-05-17
feature(formatter): `LineSuffixBoundary` and `ExpandParent` IR

MichaReiser:
## Summary

This PR introduces  two new IR elements:

* `LineSuffixBoundary`: Forces the printer to flush any pending `LineSuffix`. For example necessary to avoid a comment inside of a JSX expression accidentally getting pushed into the content of an outer element. 
* `ExpandParent`: Forces the parent group to be printed as expended. E.g. should allow us to remove the `LineSuffix` special case inside of the printer by emitting a `ExpandParent` for each block comment.

Part of #2579 

## Test Plan

I added examples with integration tests for each new IR element

### DIFF
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -605,7 +605,7 @@ fn fits_on_line<'a>(
         pending_indent: printer.state.pending_indent,
         pending_space: printer.state.pending_space,
         line_width: printer.state.line_width,
-        has_line_suffix: printer.state.line_suffixes.len() > 0,
+        has_line_suffix: !printer.state.line_suffixes.is_empty(),
     };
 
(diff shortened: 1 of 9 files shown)


crates/rome_formatter/src/printer/mod.rs
```diff
@@ -235,6 +240,9 @@ impl<'a> Printer<'a> {
 
                 queue.enqueue(PrintElementCall::new(&verbatim.element, args));
             }
+            FormatElement::ExpandParent => {
+                // No-op, only has an effect on `fits`
```
leops, line 244: Maybe we could have a `debug_assert` here to check the group is being expanded / not printed in flat mode, just in case ?

leops (approved):

PR #2591, opened 2022-05-17
feature(formatter): Add `BestFitting` IR element

MichaReiser:
This PR introduces the new IR element `BestFitting`. The IR matches Prettier's `ConditionalGroupContent`. This IR can be useful if the best representation depends on the available space. `BestFitting` defers the choice of how to print the element to the printer by providing multiple variants. The printer picks the element that best fits given the circumstances (that's where its name is coming from but I'm open to better names).

The printer picks the first variant that fits on the current line and falls back to print the last variant in expanded mode if no variant fits.

Part of #2579

## Test Plan

I added an integration test as part of the documentation of the `best_fitting` function.

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -1395,6 +1395,15 @@ pub struct BestFitting {
 }
 
 impl BestFitting {
+    /// Creates a new best fitting IR with the given variants. The method itself isn't unsafe
+    /// but it is to discourage people from using it because the printer will panic if
+    /// the slice doesn't contain at least the least and most expanded variants.
+    ///
+    /// You're looking for a way to create a `BestFitting` object, use the `best_fitting![least_expanded, most_expanded]` macro.
+    ///
+    /// ## Safety
+    /// The slice must contain at least two variants.
+    #[doc(hidden)]
     pub unsafe fn from_slice_unchecked(variants: &[FormatElement]) -> Self {
         debug_assert!(
             variants.len() >= 2,
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -816,6 +816,108 @@ pub fn hard_group_elements<T: Into<FormatElement>>(content: T) -> FormatElement 
     }
 }
 
+/// Provides multiple different alternatives and the printer picks the first one that fits.
+/// Use this as last resort because it requires that the printer must try all variants in the worst case.
+/// The passed variants must be in the following order:
+/// * First: The variant that takes up most space horizontally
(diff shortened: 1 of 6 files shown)


crates/rome_formatter/src/printer/mod.rs
```diff
@@ -247,6 +247,45 @@ impl<'a> Printer<'a> {
...
+                                    &[variant],
+                                    args.with_print_mode(PrintMode::Expanded),
+                                    &ElementCallQueue::default(),
+                                    self,
+                                ) {
+                                    self.state.measured_group_fits = true;
+
+                                    queue.enqueue(PrintElementCall::new(
+                                        variant,
+                                        args.with_print_mode(PrintMode::Expanded),
```
leops, line 280: Would it make sense to have the first variant printed in flat mode ? Or maybe it wouldn't make any difference

NicholasLYang:
Are we creating a `conditionalGroup` function too? If so, might be good to add documentation like [Prettier's](https://github.com/prettier/prettier/blob/main/commands.md#conditionalgroup). Specifically, warning that it's exponential in time complexity and that the printings are expected to be ordered from least expanded to most expanded.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1374,6 +1484,50 @@ impl PrintMode {
...
+pub struct BestFitting {
+    /// The different variants for this element.
+    /// The first element is the one that takes up the most space horizontally (the most flat),
+    /// The last element takes up the least space horizontally (but most horizontal space).
+    variants: Box<[FormatElement]>,
+}
+
+impl BestFitting {
+    pub fn new(variants: Vec<FormatElement>) -> Self {
+        assert!(variants.len() > 1, "The variants collection must contain at least two variants where the first is the least expanded state and the last element the most expanded option.");
```
yassere, line 1501: Is it necessary to panic when someone tries to construct an invalid `BestFitting` rather than returning a result?
Would it be reasonable to expand `FormatError` to account for errors during IR construction?

If the intended use of `best_fitting` is that it's always passed a `vec!` literal rather than a dynamically built up vector, maybe `best_fitting` could be a macro that ensures there are always at least two variants at compile-time?

MichaReiser: All usages in prettier are indeed using a fixed set of variants. It's unfortunate that rust doesn't allow for var args without using macros.

MichaReiser: I introduced a `best_fitting` macro that checks the invariant and marked the `from_slice_unchecked` as unsafe.

leops (approved):

PR #2592, opened 2022-05-18
chore(playground): Display cleaned prettier IR

MichaReiser:
Prettier has some utility functions to clean up the IR before showing it in the debug panel. I changed our
Playground to also use these functions because it significantly improves readability (less noise)

## Test Plan

**Before**

image

**After**
image


NicholasLYang (approved):
Looks good! One possibility is to use syntax highlighting for the IR output because it looks like it's basically JS. We can open a separate issue for that if necessary.

PR #2594, opened 2022-05-18
fix(playground): state not persisting properly in url

NicholasLYang:
## Summary

We don't load the code properly from urls because I accidentally switched the encoding from using the hash part of the url to using a query parameter named `code`. I moved it back.

## Test Plan
[main vs PR](https://www.loom.com/share/4ccdb106ed2a477aaf73f2f26888532a)

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -1,0 +1,10092 @@
+{
+	"name": "rome-website",
+	"lockfileVersion": 2,
+	"requires": true,
+	"packages": {
+		"": {
+			"name": "rome-website",
+			"dependencies": {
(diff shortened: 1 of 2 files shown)


xunilrj (approved):

Boshen:
pnpm is being used but [package-lock.json](https://github.com/rome/tools/blob/main/website/package-lock.json) got checked in 😢 should probably add a git ignore for it

IWANABETHATGUY:
> pnpm is being used but [package-lock.json](https://github.com/rome/tools/blob/main/website/package-lock.json) got checked in cry we should probably add a git ignore for it

I make a pull request here https://github.com/rome/tools/pull/2596

PR #2595, opened 2022-05-19
fix(rome_rowan): fix the next_token and prev_token methods on SyntaxToken

leops:
## Summary

These methods were behaving incorrectly in the presence of empty nodes, for instance in the following code:
```js
function test() {
  const a = true;
}
```
Calling `prev_token` on the `const` token would return `None` instead of the `{` token because of the empty `JsDirectiveList` at the start of the function body. The methods now exhaustively check all the sibling elements in a given direction before moving on to the ancestor node to avoid false negatives.


MichaReiser (approved):

MichaReiser:
!bench_formatter

PR #2598, opened 2022-05-19
refactor: Printer "fits" definition

MichaReiser:
This PR redefines how the printer decides if some content fits on a line or not.

Up to now, the printer only tested if the content (and the nested groups) fit on the line, but not if any content coming after that group fits too.

For example, if you have the following statement:

```js
if (someLongCondition) {}
```

where `(someLongCondition)` is grouped. The current printer considered that the content fits if printing the `)` didn't exceed the line width.

The new behavour changes the behaviour so that it tries to print any remaining content until it finds the first `\n` or exceeds the line width. This means for the example above that it only considers the content to fit if the `{}` also fit on the line without exceeding the line width.

This change has wide reaching consequences and regresses our formatting in many places because

* Our implementation often wraps content where it shouldn't
* And misses grouping in many places where it should

I don't intend to fix these issues as part of this PR, instead, this should be approached incrementelly by re-visiting every formatting rule and making sure that Rome creates the same IR as prettier.

Landing this change is important because it significantly changes how the printing works and the new behaviour is necessary to be able to adopt an IR output that's closer/similar to Prettier's.

This PR regresses prettier compatibility by a bit:

**File Based Average Prettier Similarity**: 69.25 -> 69.18%  
**Line Based Average Prettier Similarity**: 64.32%  -> 63.60

Part of #2579 

## Test Plan

`cargo test`

// TODO pass correct queue to `fits` when landing this PR after the best fitting PR


crates/rome_formatter/src/printer/mod.rs
```diff
@@ -121,8 +121,7 @@ impl<'a> Printer<'a> {
                         // print the group in "flat" mode, otherwise continue in expanded mode
 
                         let flat_args = args.with_print_mode(PrintMode::Flat);
-                        // TODO replace `ElementCallQueue::default` with the real queue to also measure the rest of the document
-                        if fits_on_line(&[content], flat_args, &ElementCallQueue::default(), self) {
+                        if fits_on_line(&[content], flat_args, queue, self) {
```
MichaReiser, line 124: The change is "trivial" because I did the initial printer rewrite with this change already in mind.

crates/rome_js_formatter/tests/specs/js/module/arrow/call.js.snap
```diff
@@ -80,38 +80,44 @@ expect(
...
-	(result) => result.veryLongVariable.veryLongPropertyName > someOtherVariable ? "ok" : "fail",
+	(
+		result,
+	) => result.veryLongVariable.veryLongPropertyName > someOtherVariable ? "ok" : "fail",
 );
 
 
 ## Lines exceeding width of 80 characters
 
-   25: expect(() => asyncRequest({ type: "foo", url: "/test-endpoint" })).not.toThrowError();
```
MichaReiser, line 113: This is an example where the old printer "printed over" the line width because it only checked if the group (the `{...} argument) fits on the line but not if the rest does so too.

crates/rome_js_formatter/src/js/statements/for_of_statement.rs
```diff
@@ -22,7 +22,7 @@ impl FormatNodeFields<JsForOfStatement> for FormatNodeRule<JsForOfStatement> {
             body,
         } = node.as_fields();
 
-        format_head_body_statement(
+        Ok(group_elements(format_head_body_statement(
```
MichaReiser, line 25: This is one exception that I made. It improves the formatting of `for of` expressions, making use of the new behaviour.

crates/rome_js_formatter/tests/specs/js/module/arrow/arrow_nested.js.snap
```diff
@@ -39,10 +39,19 @@ Indent style: Tab
 Line width: 80
 Quote style: Double Quotes
 -----
-Seq(typeDef.interface.groups).forEach(
-	(group) =>
-		Seq(group.members).forEach(
-			(member, memberName) =>
+Seq(
+	typeDef.interface.groups,
```
leops, line 43: This doesn't look right, although there are definitely some places where we create too many groups and it causes some syntax to break in unexpected ways, here there's no reason for this specific group to be breaking as the line doesn't go over the print width

crates/rome_js_formatter/tests/specs/prettier/js/comments/while.js.snap
```diff
@@ -53,7 +51,10 @@ while (true) {} // comment
 
 while (true) /* comment */ ++x;
 
-while (1) foo(); // Comment
+while (
+  1
+) foo() // Comment
+;
```
leops, line 57: Why is this semicolon getting pushed to the next line now ?

MichaReiser: It's because of the `lineSuffixBoundary` that I had to introduce after the head of a member chain to work around some stability issue. I tightened that test now but I'm somewhat convinced that there's some deeper underlying issue that I'm currently don't fully understand (member chain formatting is wild).

crates/rome_js_formatter/tests/specs/js/module/arrow/arrow_nested.js.snap
MichaReiser: This is one of the cases where `BestFitting` is necessary to make things work because the printer now tries to fit as many groups as possible until it finds a line break and starts expanding groups from the left to the right if this doesn't work. 

I would prefer not having to change the member chain implementation to use alternatives just yet. I think this should be something that is worth doing in its own PR

leops (approved):

PR #2600, opened 2022-05-20
Replace unicode-id table with `unicode-id-start` crate

Boshen:
# Summary

@malobre [pointed out](https://github.com/rome/tools/issues/1930#issuecomment-1131549639) dtolnay published [unicode-ident](https://github.com/dtolnay/unicode-ident), which is better than the widely used `unicode-xid` approach (the one Rome uses) referenced in its README.

I forked, replaced all references of xid to id and published a crate https://crates.io/crates/unicode-id-start

# Test Plan
All test passes.


Boshen:
!bench_parser

MichaReiser (approved):
Nice! The license seems to be OK too. 

I'm not very familiar with our existing unicode tables. What's the difference between your crate and the `unicode-ident` (for a nood :D)

Boshen:
The technical details are detailed here: https://github.com/dtolnay/unicode-ident/blob/master/README.md#unicode-ident-1

tldr: the crate is 2kb smaller, 2 times faster (O(1) table lookup compared to binary search search from the original).

Boshen:
> What's the difference between your crate and the `unicode-ident` (for a nood :D)

The only difference: JavaScript uses the ID_Start and ID_Continue range, the original crate uses XID_Start and XID_Continue.

MichaReiser:
What are your plans of supporting this crate? My main concern about depending on a third party crate for this is that we're now no longer able to update the Unicode tables ourselves if necessary. But I guess, we can go and fork your crate if that ever becomes a problem.

malobre:
It would be great to consolidate efforts between `swc`, `rome`, `rslint` and the like on those kind of things.
However, a crate like this seems important enough to warrant being under an organization (or at least have multiple people able to update the crate) in case the author goes MIA.

Boshen:
We can co-mantain it, I'll add maintainers both on Github and crates.io.
I also don't mind moving it into the Rome org.

xunilrj:
I vote to make a crate that is a generator and not a library.
Why?

The original library is hardcoded for ```XID_Start```and ```XID_Continue``` and you changed yours to ```ID_Start``` and ```ID_Continue```. 

https://github.com/dtolnay/unicode-ident/blob/master/generate/src/ucd.rs#L3
```
ucd-generate property-bool UCD --include XID_Start,XID_Continue
```

If we instead create a crate that generates these tables we will have a more stable crate for all possible uses.
For example, ```uce-generate``` itself lists other categories as examples on its page; and we certainly don't wanna one library per category.

```
ucd-generate property-bool /tmp/ucd-13.0.0 --include Hyphen,Dash,Quotation_Mark --chars
```

So my suggestion is:
- A library that we call passing a list of categories and it generates the tables and the access methods;
- the generated methods are extensions methods on char following the pattern that already exists: ```char::is_unicode_id_start()``` or ```char::is_unicode_emoji```;
- Another library with a lot of categories pre-baked with cargo features to enable them. Ideally everything would be generated with a macro ```gen_table!{ID_Start, ID_Continue}```.

I would be more confident in depending on this second library, for example.

MichaReiser (changes requested):
I like @xunilrj recommendation for this. @Boshen is this something you want to tackle?

Boshen:
> I like @xunilrj recommendation for this. @Boshen is this something you want to tackle?

Unfortunately I don't see any value by making it support other categories, and I don't have the energy or time to change it.

The problem space is "detect ID_Start/ID_Continue range" for JavaScript, and nothing else. 

Feel free to reject this PR :smile:

ematipico:
@xunilrj @MichaReiser considering the last comment, could you please merge or close this PR? It has been in stale for too long

ematipico:
No feedback here, I am going to close the PR.

PR #2601, opened 2022-05-20
feat(rome_rowan): cast and parent to help SyntaxNode and AstNode navigation

xunilrj:
this PR is about this discussion: https://github.com/rome/tools/discussions/2586

I avoided the lifted ```parent``` method because is more unconventional.

### DIFF
--- a/crates/rome_rowan/Cargo.toml
+++ b/crates/rome_rowan/Cargo.toml
@@ -15,8 +15,8 @@ hashbrown = { version = "0.11.2", features = [
 text-size = "1.1.0"
 memoffset = "0.6.5"
 countme = "3.0.0"
-
 serde_crate = { package = "serde", version = "1.0.133", optional = true, default-features = false }
+extension-trait = "1.0.1"
 
 [dev-dependencies]
 quickcheck = "1.0.3"
--- a/crates/rome_rowan/src/ast/mod.rs
+++ b/crates/rome_rowan/src/ast/mod.rs
@@ -4,6 +4,7 @@
 //! from any error and produce an ast from any source code. If you don't want to account for
 //! optionals for everything, you can use ...
 
+use extension_trait::extension_trait;
 #[cfg(feature = "serde")]
 use serde_crate::Serialize;
 use std::error::Error;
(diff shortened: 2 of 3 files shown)


crates/rome_rowan/Cargo.toml
```diff
@@ -15,8 +15,8 @@ hashbrown = { version = "0.11.2", features = [
 text-size = "1.1.0"
 memoffset = "0.6.5"
 countme = "3.0.0"
-
 serde_crate = { package = "serde", version = "1.0.133", optional = true, default-features = false }
+extension-trait = "1.0.1"
```
leops, line 19: I'm not sure about including an additional dependency for this. Ultimately it's just a build-time only procedural macro, but it's also doing something very simple that we could easily do by hand

xunilrj: I agree, but my idea of including this was as an incentive to create more extension methods.

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -76,6 +77,17 @@ pub trait AstNode {
...
     }
+
+    fn parent<T: AstNode<Language = Self::Language>>(&self) -> Option<T> {
+        self.syntax().parent().and_then(T::cast)
+    }
+}
+
+#[extension_trait]
+impl<L: Language> SyntaxNodeCast<L> for SyntaxNode<L> {
+    fn cast<T: AstNode<Language = L>>(self) -> Option<T> {
```
MichaReiser, line 88: Please add some documentation to what `cast` does`.

crates/rome_rowan/Cargo.toml
MichaReiser: I'm leaning towards @leops standpoint

MichaReiser (approved):

MichaReiser:
Any thoughts around a `prelude` that automatically imports this trait?

crates/rome_rowan/src/ast/mod.rs
xunilrj: Done.

crates/rome_rowan/Cargo.toml
xunilrj: Done.

xunilrj:
> Any thoughts around a `prelude` that automatically imports this trait?

At the moment we are already publishing everything from the ```ast``` module.

PR #2602, opened 2022-05-20
fix(rome_formatter): improve the compatibility of range formatting with Prettier

leops:
## Summary

This PR modifies how range formatting is implemented to bring the general results closer to what Prettier outputs:
- First the formatting range is preprocessed to trim the leading and trailing whitespace, so if the user wants to format `\n····statement();\n` what's actually getting formatted is `statement();`
- Instead of starting from two individual tokens, the formatting instead starts from two nodes covering the start and end point of the range respectively and selected using a language-provided predicate (in the case of JavaScript only statements and declarations are allowed for instance)
- The sourcemap emitted by the formatter now includes zero-width markers around formatter-generated tokens (for instance a semicolon added after an expression statement). These are used to extend the range of text output from the range formatting function to include those tokens
- Finally two early exit conditions have been added in case the formatting range is determined to be empty or whitespace-only

Overall these changes improve the Prettier compatibility metrics by the following:
**File Based Average Prettier Similarity**: 69.38% -> 70.38%
**Line Based Average Prettier Similarity**: 64.39% -> 64.57%

## Test Plan

Besides updating the snapshots, the range formatting tests in `rome_js_formatter` have been modified to fit the new range selection logic


crates/rome_formatter/src/lib.rs
```diff
@@ -663,15 +702,27 @@ pub fn format_node<
 ///
 /// It returns a [Formatted] result with a range corresponding to the
 /// range of the input that was effectively overwritten by the formatter
-pub fn format_range<
+pub fn format_range<Options, L, R, P>(
+    options: Options,
+    root: &SyntaxNode<L>,
+    mut range: TextRange,
+    mut predicate: P,
```
MichaReiser, line 709: Please add some additional documentation on what `predicate` is for. I wouldn't know what I need to pass in from reading the API. 

Or introduce a new trait / struct that wrapps `node`, `predicate`, and the formatting rule.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -97,13 +97,24 @@ impl<'a> Printer<'a> {
...
-                    self.state.source_markers.push(SourceMarker {
-                        source: *source,
-                        dest: TextSize::from(self.state.buffer.len() as u32),
-                    });
+                    self.state.source_position = *source;
                 }
 
+                self.state.source_markers.push(SourceMarker {
+                    source: self.state.source_position,
+                    dest: TextSize::from(self.state.buffer.len() as u32),
```
MichaReiser, line 105: Nit
```suggestion
                    dest: self.state.buffer.text_len(),
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -97,13 +97,24 @@ impl<'a> Printer<'a> {
...
 
                 if let Some(source) = token.source_position() {
-                    self.state.source_markers.push(SourceMarker {
-                        source: *source,
-                        dest: TextSize::from(self.state.buffer.len() as u32),
-                    });
+                    self.state.source_position = *source;
                 }
 
+                self.state.source_markers.push(SourceMarker {
```
MichaReiser, line 103: Aren't we now always pushing two markers for every token. One before and one after that token? And wouldn't be the prev_token after being the same as current_token before if the token doesn't have any source position? 

Can't we search in the `format_range` method for the first marker that's one passed the formatted end range instead?

leops: Yes there is now a marker at the start and end of each token. For two consecutive tokens with source positions that means there will be two identical markers but it should be trivial to add a bit of logic checking `source_markers.last()` to avoid creating duplicates.
Searching for the first marker that's outside of the formatting range is what the previous range formatting logic used to do, but that means the selected marker will be at the start of the first token following the formatting range: this has the effect of including all the spaces, line breaks or indentation that may have been inserted between the two tokens by the formatter in the formatting result. This expansion of the formatting range to include all leading and trailing trivias is unfortunately the exact opposite of what Prettier does, contracting the range to trim the leading and trailing whitespace, hence why I changed this logic to tighten the range as much as possible instead but without dropping trailing tokens that may have been inserted by the formatter.

MichaReiser: I think it could  help to add some documentation to this section explaining what it is for and how the mappings would look like. 

I think what's hard to understand is why it's necessary to have more than one mapping for each node. I would expect that even inserted tokens would have a single mapping, just that they point to the last-known source location.

ematipico (commented):
Can we add **new** tests that cover these new changes?

MichaReiser (approved):
!bench_formatter

ematipico (approved):

leops:
!bench_formatter

PR #2609, opened 2022-05-24
feature(formatter): Fill separators

NicholasLYang:
## Summary

Adding a separator option for fill printing. Allows for JSX child list formatting as prettier fill prints with no spaces in between.

## Test Plan

Added a doc test [here](https://github.com/rome/tools/blob/15d2f8d1ac1461ac70f20bb0da7f0bd856f25535/crates/rome_formatter/src/format_element.rs#L377). Existing tests for fill printing should still apply as well.


crates/rome_js_formatter/src/lib.rs
```diff
@@ -499,10 +511,7 @@ mod test {
             file_name: "quick_test",
             format_options: JsFormatOptions::default(),
         });
-        assert_eq!(
-            result.as_code(),
-            r#"(a + (b * c)) > (65 + 5);
-"#
-        );
+
+        println!("{}", result.as_code());
```
IWANABETHATGUY, line 515: Maybe `assert` would be better here, since the function name is `quick_test`.

MichaReiser (changes requested):
Please go through my comments on the other PR and double check to address the review feedback.

crates/rome_formatter/src/format_element.rs
```diff
@@ -374,12 +374,36 @@ where
...
+/// ```rust
+/// use rome_formatter::prelude::*;
+/// use rome_formatter::Formatted;
+/// use std::str::from_utf8;
+///
+/// let a = "<b>Important: </b>";
+/// let b = "Please do not commit memory bugs such as segfaults, buffer overflows, etc. otherwise you ";
+/// let c = "<em>will</em>";
+/// let d = " be reprimanded";
+/// let expr = fill_elements(empty_element(), [token(a), token(b), token(c), token(d)]);
```
MichaReiser, line 386: Please create a test with a spacer that has different "flat" and "expanded" representations and make sure the printer prints all these cases correctly 
* item, spacer, and next item fit
* item and spacer fit, next item doesn't
* only item fits

crates/rome_formatter/src/format_element.rs
```diff
@@ -374,12 +374,36 @@ where
...
+///
+/// assert_eq!(
+///     format!("{a}\n{b}\n{c}{d}"),
+///     Formatted::new(expr, PrinterOptions::default())
+///         .print()
+///         .as_code()
+/// )
+/// ```
+pub fn fill_elements<TSep: Into<FormatElement>>(
```
ematipico, line 395: Should `TSep` be `Separator`?

crates/rome_formatter/src/format_element.rs
```diff
@@ -374,12 +374,36 @@ where
...
+    elements: impl IntoIterator<Item = FormatElement>,
+) -> FormatElement {
     let mut list: Vec<_> = elements.into_iter().collect();
     match list.len() {
         0 => empty_element(),
         1 => list.pop().unwrap(),
-        _ => FormatElement::Fill(List::new(list)),
+        _ => FormatElement::Fill(Box::new(Fill {
+            list: List::new(list),
+            separator: separator.into(),
```
ematipico, line 405: I wonder if we should limit the usage of `separator` to sub set of `FormatElement`? Theoretically, as it is, we could use a `FormatElement::List` as separator, which is not ideal (and it doesn't make much sense either). Should we limit, for now, to have a separator as `FormatElement::Space` and `FormatElement::Empty` as valid options?

ematipico: That's not my concern. My concern is only about the incorrect usage of the feature due to how broad is the type.

I could create a fill element where the separator is another fill element or a group. Is it how we want to use the new IR? To me, it seems an incorrect usage.

MichaReiser: Passing a `group` is something you may want to do. There's nothing wrong with that. For example, you could use a group as separator that uses a `conditional content` to pick a different separator depending if the two neighbouring elements fit on the same line or not.

crates/rome_formatter/src/format_element.rs
NicholasLYang: I asked this in the JSX PR. What should the result be if the separator is printed differently? Because the fill printing should handle line breaking, not the separators themselves. Perhaps we should only allow non breaking separators?

crates/rome_formatter/src/format_element.rs
NicholasLYang: I stole this from one of the other functions. I can rename it if necessaru

crates/rome_formatter/src/format_element.rs
NicholasLYang: Yeah, in fact, reading the [fill printer code](https://github.com/rome/tools/blob/2989555623f9a5460ae4fcdcc6ad6ec8db896c3c/crates/rome_formatter/src/printer/mod.rs#L385), at no point would we print a separator in expanded mode. Either it fits, in which case we print it flat, or we insert a hard line break. In theory we could replace that with printing the separator in expanded mode, but that'd mean users would have to provide a separator that breaks.

crates/rome_js_formatter/src/lib.rs
NicholasLYang: It's not super important considering the quick_test function is ignored by default. It's just a way to test code quickly.

crates/rome_formatter/src/format_element.rs
MichaReiser: Sorry, I didn't see that comment. 

Yes, we would need to change the logic inside of the printer to align it with Prettier's behaviour. The same as you did by allowing any other separator than space. 

I quickly glanced through Prettier's fill use cases and the usage for [`arrays`](https://github.com/prettier/prettier/blob/d13feed42b6478710bebbcd3225ab6f203a914c1/src/language-js/print/array.js#L167-L186) is interesting. 

You can see that there's a use case where they use a `hardline` separator if the next element is separated by an empty line or if the element has a comment and a `softline` (`softline_or_space`) otherwise. This is something that we couldn't support with our current design because we only support a single separator. 

Simply relying on the `flat` and `expanded` mode rather than hardcoding a line break also feels natural. It makes use of the fundamental principals of the printer and should also not be too hard to add. It should only require replacing the `hard_line_break` with the separator.

crates/rome_formatter/src/format_element.rs
```diff
@@ -1631,8 +1676,10 @@ impl FormatElement {
             FormatElement::Indent(indent) => indent.content.will_break(),
             FormatElement::Group(group) => group.content.will_break(),
             FormatElement::ConditionalGroupContent(group) => group.content.will_break(),
-            FormatElement::List(list) | FormatElement::Fill(list) => {
-                list.content.iter().any(FormatElement::will_break)
+            FormatElement::List(list) => list.content.iter().any(FormatElement::will_break),
+            FormatElement::Fill(fill) => {
+                fill.list.content.iter().any(FormatElement::will_break)
+                    || fill.separator.will_break()
```
MichaReiser, line 1682: Nit. Should be safe to remove the `separator` from here as it should never break in flat mode (or using fill is pointless)

MichaReiser (approved):

crates/rome_js_formatter/src/lib.rs
ematipico: Apologies Nicholas, but even though this test is ignored, I use this test to check test things locally using my IDE (and probably other people too). If we remove the assertion, I will have to add it again.

Issue #2506, opened 2022-04-27
☂️ Documentation changes 0.6.0


ematipico:
# Tentative release date, Tuesday 7th of June

List of changes to apply to the changelog and the website:
- [ ] **NEW**: Added the `rome check` CLI command to run the linter on a set of files / directories
- [ ] **BREAKING**: The `rome format --ci` command has been renamed to `rome ci`, and now includes linting as well
- [ ] **NEW**: Added a `--no-colors` global flag to the CLI to disable the formatting of markup
- [ ] Better compatibility with Prettier
	- Better formatting of various statements
	- Closed the majority of typescript formatting issue
	- Great work around string literals

PR #2613, opened 2022-05-25
feat(rome_cli): disable markup formatting when unsupported

leops:
## Summary

This PR disables the formatting of markup in the console output when:
- The CLI is called with the `--no-colors` argument
- The standard output stream is not a Unix TTY / Windows Console handle (checked with the `atty` crate)
- The `TERM=dumb` or `NO_COLOR` environment variables are defined

Additionally I also took the opportunity to update the CLI help to use Markup

### DIFF
--- a/crates/rome_cli/src/lib.rs
+++ b/crates/rome_cli/src/lib.rs
@@ -21,9 +21,12 @@ pub struct CliSession<'app> {
 
 impl CliSession<'static> {
     pub fn from_env() -> Self {
+        let mut args = Arguments::from_env();
+        let no_colors = args.contains("--no-colors");
+
         Self {
-            app: App::from_env(),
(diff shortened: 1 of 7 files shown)


MichaReiser (approved):

crates/rome_cli/src/lib.rs
```diff
@@ -21,9 +21,12 @@ pub struct CliSession<'app> {
 
 impl CliSession<'static> {
     pub fn from_env() -> Self {
+        let mut args = Arguments::from_env();
+        let no_colors = args.contains("--no-colors");
```
ematipico, line 25: It seems that this new argument is not been documented @leops 

Could you please do so?
- add it to the help message
- add a new bullet point to https://github.com/rome/tools/issues/2506

PR #2615, opened 2022-05-25
feat(rome_analyze): add a fix argument to `rome check` command CLI and LSP

leops:
## Summary

This PR adds a new auto-fix workflow to the analysis toolchain, with the following changes:
- The callback injected into the `analyze` entrypoint function for the analyzer now returns an `std::ops::ControlFlow` enum. This lets the callback interrupt the analysis process by returning `ControlFlow::Break(value)`, with the value being returned by the analyze function as an `Option`
- The `Workspace` now has a `fix_file` method that applies all safe fixes to a given file and returns the resulting code. This is implemented by repeatedly calling into `analyze` and breaking for each emitted action signal with `Applicability::Always` to mutate the syntax tree with the code action, then restart the analysis until no more code actions are emitted
- The `check` command of the CLI now has an `--apply` argument reusing the same traversal infrastructure, and calling `fix_file` on each processed file then writing the resulting code back to the disk
- The Language Server can now emit a "Fix all auto-fixable issues" code action when requested by the language client (for instance when using the "Fix all" command in VS Code), the `TextEdit` backing this action is also generated from the result of `fix_file`
- The code action emitted by the `useSingleVarDeclarator` rule has been marked as "always applicable", and the implementation of the fix has been changed to support variable statements in `JsModuleItemList` as well as `JsStatementList`

## Test Plan

New tests have been added to the CLI test suite to check the `rome check --apply` command

### DIFF
--- a/crates/rome_analyze/src/analyzers/use_single_var_declarator.rs
+++ b/crates/rome_analyze/src/analyzers/use_single_var_declarator.rs
@@ -4,10 +4,10 @@ use rome_console::markup;
 use rome_diagnostics::{Applicability, Severity};
 use rome_js_factory::make;
 use rome_js_syntax::{
-    JsAnyRoot, JsAnyStatement, JsStatementList, JsSyntaxToken, JsVariableDeclarationFields,
+    JsAnyRoot, JsModuleItemList, JsStatementList, JsSyntaxToken, JsVariableDeclarationFields,
     JsVariableDeclaratorList, JsVariableStatement, JsVariableStatementFields,
 };
-use rome_rowan::{AstNode, AstNodeExt, AstNodeList, AstNodeListExt, AstSeparatedList};
+use rome_rowan::{AstNode, AstSeparatedList};
 
 use crate::{ActionCategory, RuleCategory};
 
@@ -57,13 +57,19 @@ impl Rule for UseSingleVarDeclarator {
     fn action(root: JsAnyRoot, node: &Self::Query, state: &Self::State) -> Option<JsRuleAction> {
         let (kind, declarators, semicolon_token) = state;
 
-        let prev_parent = JsStatementList::cast(node.syntax().parent()?)?;
+        let prev_parent = node.syntax().parent()?;
+        if !JsStatementList::can_cast(prev_parent.kind())
+            && !JsModuleItemList::can_cast(prev_parent.kind())
+        {
+            return None;
+        }
+
         let index = prev_parent
-            .iter()
-            .position(|slot| slot.syntax() == node.syntax())?;
+            .children()
+            .position(|slot| &slot == node.syntax())?;
 
         let mut is_first = true;
-        let next_parent = prev_parent.clone().splice(
+        let next_parent = prev_parent.clone().splice_slots(
             index..=index,
             declarators.iter().filter_map(|declarator| {
                 let declarator = declarator.ok()?;
@@ -86,15 +92,18 @@ impl Rule for UseSingleVarDeclarator {
                     builder = builder.with_semicolon_token(semicolon_token.clone());
                 }
 
-                Some(JsAnyStatement::from(builder.build()))
+                Some(Some(builder.build().into_syntax().into()))
             }),
         );
 
         Some(JsRuleAction {
             category: ActionCategory::QuickFix,
-            applicability: Applicability::MaybeIncorrect,
+            applicability: Applicability::Always,
             message: markup! { "Break out into multiple declarations" }.to_owned(),
-            root: root.replace_node_discard_trivia(prev_parent, next_parent)?,
+            root: JsAnyRoot::unwrap_cast(
+                root.into_syntax()
+                    .replace_child(prev_parent.into(), next_parent.into())?,
+            ),
         })
     }
 }
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -29,7 +29,7 @@ pub struct AnalysisFilter<'a> {
     pub range: Option<TextRange>,
 }
 
-/// Utility type to be used as a debault value for the `B` generic type on
+/// Utility type to be used as a default value for the `B` generic type on
 /// [analyze] when the provided callback never breaks
 ///
 /// This should eventually get replaced with the `!` type when it gets stabilized
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -36,6 +36,25 @@ pub struct AnalysisFilter<'a> {
 pub enum Never {}
 
 /// Type alias of [ops::ControlFlow] with the `B` generic type defaulting to [Never]
+///
+/// By default the analysis loop never breaks, so it behaves mostly like
+/// `let b = loop {};` and has a "break type" of `!` (the `!` type isn't stable
+/// yet so I'm using an empty enum instead but they're identical for this
+/// purpose)
+///
+/// In practice it's not really a `loop` but a `for` because it's iterating on
+/// all nodes in the syntax tree, so when it reaches the end of the iterator
+/// the loop will exit but without producing a value of type `B`: for this
+/// reason the [analyze] function returns an `Option<B>` that's set to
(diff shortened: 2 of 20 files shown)


crates/rome_analyze/src/analyzers/use_single_var_declarator.rs
```diff
@@ -86,15 +92,18 @@ impl Rule for UseSingleVarDeclarator {
...
-            applicability: Applicability::MaybeIncorrect,
+            applicability: Applicability::Always,
             message: markup! { "Break out into multiple declarations" }.to_owned(),
-            root: root.replace_node_discard_trivia(prev_parent, next_parent)?,
+            root: JsAnyRoot::unwrap_cast(
+                root.into_syntax()
+                    .replace_child(prev_parent.into(), next_parent.into())?,
+            ),
         })
     }
```
xunilrj, line 108: Would be possible to write some unit tests for each rule? Sometimes is not clear what the rule is flagging or doing.

crates/rome_analyze/src/lib.rs
```diff
@@ -27,13 +29,24 @@ pub struct AnalysisFilter<'a> {
     pub range: Option<TextRange>,
 }
 
+/// Utility type to be used as a debault value for the `B` generic type on
```
xunilrj, line 32: "debault" -> "default"

crates/rome_cli/src/commands/fix.rs
```diff
@@ -0,0 +1,9 @@
+use crate::{
+    traversal::{traverse, TraversalMode},
+    CliSession, Termination,
+};
+
+/// Handler for the "fix" command of the Rome CLI
```
ematipico, line 6: Why the `fix` is a command and not an argument of `check` command?

[In rome classic we called it `--apply`]( https://github.com/rome/tools/blob/archived-js/internal/core/server/commands/check.ts#L46-L51) and it was an argument of the command `rome check`. That's because `check` wasn't just about "fixing" the code, but also apply formatting.

ematipico (changes requested):
Sorry to interject your PR but I am a little confused about where we are headed with the commands. 

I would like to understand the requirements first (is there been a discussion?). Could you please create an issue first? 

The reason why I am concerned is because I don't have any idea of where we are headed with the commands.

In Rome classic we had:
- `rome ci`, in line with the rewrite
- `rome format`, in line with the rewrite
- `rome check`, **not in line with the rewrite**. The former command applied linting and formatting. `--apply` was the argument to have safe fixed and formatting applied. Like `--write` in the current `rome format --write`

crates/rome_cli/src/commands/fix.rs
leops: Having two different commands was mostly because the check and fix commands had mostly different backing implementations, but I could as well change it to have the fix behavior be implemented by the `check` command with an `--apply` flag instead to be consistent with how the CLI worked in Rome JS

crates/rome_analyze/src/lib.rs
```diff
@@ -27,13 +29,24 @@ pub struct AnalysisFilter<'a> {
...
+/// [analyze] when the provided callback never breaks
+///
+/// This should eventually get replaced with the `!` type when it gets stabilized
+pub enum Never {}
+
+/// Type alias of [ops::ControlFlow] with the `B` generic type defaulting to [Never]
```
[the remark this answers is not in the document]
leops: By default the analysis loop never breaks, so it behaves mostly like `let b = loop {};` and has a "break type" of `!` (the `!` type isn't stable yet so I'm using an empty enum instead but they're identical for this purpose)
In practice it's not really a `loop` but a `for` because it's iterating on all nodes in the syntax tree, so when it reaches the end of the iterator the loop will exit but without producing a value of type `B`: for this reason the `analyze` function returns an `Option<B>` that's set to `Some(B)` if the callback did break, and `None` if the analysis reached the end of the file.
Most consumers of the analyzer will want to analyze the entire file at once and never break, so using `Never` as the type of `B` in this case lets the compiler know the `ControlFlow::Break` branch will never be taken and can be optimized out, as well as completely remove the `return Some` case (`Option<Never>` has a size of 0 and can be elided, while `Option<()>` has a size of 1 as it still need to store a discriminant)

crates/rome_analyze/src/lib.rs
```diff
@@ -27,13 +29,24 @@ pub struct AnalysisFilter<'a> {
...
 /// then call the `callback` when an analysis rule emits a diagnostic or action
-pub fn analyze<B>(file_id: FileId, root: &JsAnyRoot, filter: AnalysisFilter, mut callback: B)
-where
-    B: FnMut(&dyn AnalyzerSignal),
-{
+pub fn analyze<B>(
+    file_id: FileId,
+    root: &JsAnyRoot,
+    filter: AnalysisFilter,
+    mut callback: impl FnMut(&dyn AnalyzerSignal) -> ControlFlow<B>,
```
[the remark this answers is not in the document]
leops: Most of the arguments for the analysis have already been pushed into the `AnalysisFilter` struct, and I don't expect to add many more arguments to the `analyze` function at this stage except an eventual `&mut self` to let the analyzer keep some state / cache between runs, along some global-level settings

crates/rome_analyze/src/analyzers/use_single_var_declarator.rs
leops: There are snapshots tests for each rules under `crates/rome_analyze/tests/specs`. I went with snapshots since it captures the diagnostic or code action diff in a readable form close to what the end user will actually see on the CLI, rather than running individual asserts on things like the syntax tree resulting from a refactor since I found it to be easily unclear.

crates/rome_analyze/src/lib.rs
```diff
@@ -17,7 +19,7 @@ pub use crate::signals::{AnalyzerAction, AnalyzerSignal};
 
 /// Allows filtering the list of rules that will be executed in a run of the analyzer,
 /// and at what source code range signals (diagnostics or actions) may be raised
-#[derive(Default)]
+#[derive(Default, Clone, Copy)]
```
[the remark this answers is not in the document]
leops: It is currently possible to filter rules by name (that's used for the snapshot tests to enable only the rule being tested for instance), but since the name of each rules are statically known they can be represented with `&'static str` string slices that implement `Copy`

crates/rome_analyze/src/lib.rs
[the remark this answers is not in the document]
MichaReiser: It may make sense to define a type alias for the `Callback` type and copy some of the explanations you gave in the other comments into its documentation so that future me will know what I need to return in the control flow.

crates/rome_analyze/src/registry.rs
```diff
@@ -45,18 +45,22 @@ impl_registry_builders!(
...
     // Run all rules known to the registry associated with nodes of type N
-    pub(crate) fn analyze(
+    pub(crate) fn analyze<B>(
         &self,
         file_id: FileId,
         root: &JsAnyRoot,
         node: JsSyntaxNode,
-        callback: &mut impl FnMut(&dyn AnalyzerSignal),
-    ) {
+        callback: &mut impl FnMut(&dyn AnalyzerSignal) -> ControlFlow<B>,
```
MichaReiser, line 53: Nit: Use the type alias here too?

MichaReiser, line 53: Nit: Use the type alias here too?

MichaReiser (approved):

leops:
> It may make sense to define a type alias for the Callback type and copy some of the explanations you gave in the other comments into its documentation so that future me will know what I need to return in the control flow.

Unfortunately it's not possible to move that declaration in a type alias since `impl Trait` is not stable in type aliases (and it's not possible to alias the `FnMut(&dyn AnalyzerSignal) -> ControlFlow<B>` trait under another name either)

ematipico:
Maybe we should wait to merge this PR until we do the release next week?

ematipico (approved):

Issue #2444, opened 2022-04-14
Remove unnecessary escapes from string literals


MichaReiser:

Boshen:
The lexer from [jsparagus](https://github.com/mozilla-spidermonkey/jsparagus/blob/master/crates/parser/src/lexer.rs#L416) uses a string builder to create an unescaped string and saves it.

Do we do the same or is there a better approach?

MichaReiser:
> The lexer from [jsparagus](https://github.com/mozilla-spidermonkey/jsparagus/blob/master/crates/parser/src/lexer.rs#L416) uses a string builder to create an unescaped string and saves it.
> 
> Do we do the same or is there a better approach?

Feels a bit overkill to me. My gut feeling is that the algorithm must check whenever it sees a `\` if the character following it is in a range that requires escape sequences. If not -> Skip the `\` otherwise keep it.

PR #2624, opened 2022-05-26
fix(rome_js_formatter): choose preferred quotes and reduce escaping

### LINKED ISSUE #2443: Preferred string quotes
Prettier detects the ideal quote to use for a string so that the least amount of escaped quotes are necessary: 

```
function getPreferredQuote(rawContent, preferredQuote) {
  /** @type {{ quote: '"', regex: RegExp, escaped: "&quot;" }} */
  const double = { quote: '"', regex: /"/g, escaped: "&quot;" };
  /** @type {{ quote: "'", regex: RegExp, escaped: "&apos;" }} */
  const single = { quote: "'", regex: /'/g, escaped: "&apos;" };

  const preferred = preferredQuote === "'" ? single : double;
  const alternate = preferred === single ? double : single;

  let result = preferred;

  // If `rawContent` contains at least one of the quote preferred for enclosing
  // the string, we might want to enclose with the alternate quote instead, to
  // minimize the number of escaped quotes.
  if (
    rawContent.includes(preferred.quote) ||
    rawContent.includes(alternate.quote)
  ) {
    const numPreferredQuotes = (rawContent.match(preferred.regex) || []).length;
    const numAlternateQuotes = (rawContent.match(alternate.regex) || []).length;

    result = numPreferredQuotes > numAlternateQuotes ? alternate : preferred;
  }

  return result;
}
```

[Playground](https://play.rome.tools/?lineWidth=80&indentStyle=space&quoteStyle=double&indentWidth=2&typescript=true&jsx=true&sourceType=module#IAAvAC8AIABTAGkAbgBnAGwAZQAgAGQAbwB1AGIAbABlACAAcQB1AG8AdABlAC4ACgAoACcAIgAnACkAOwAKACgAJwAiACcAKQA7AAoAIAAKACAALwAvACAAUwBpAG4AZwBsAGUAIABzAGkAbgBnAGwAZQAgAHEAdQBvAHQAZQAuAAoAKAAiACcAIgApADsACgAoACIAJwAiACkAOwAKAAoALwAvACAATwBuAGUAIABvAGYAIABlAGEAYwBoAAoAIgBcACIAJwAiADsACgAnACIAXAAnACcAOwAKAAoALwAvACAATwBuAGUAIABvAGYAIABlAGEAYwBoACAAdwBpAHQAaAAgAHUAbgBuAGUAYwBlAHMAcwBhAHIAeQAgAGUAcwBjAGEAcABlAHMALgAKACgAIgBcACIAJwAiACkAOwAKACgAIgBcACIAJwAiACkAOwAKACAACgAgAC8ALwAgAE0AbwByAGUAIABkAG8AdQBiAGwAZQAgAHEAdQBvAHQAZQBzACAAdABoAGEAbgAgAHMAaQBuAGcAbABlACAAcQB1AG8AdABlAHMALgAKACgAJwAiAFwAJwAiACcAKQA7AAoAKAAnACIAXAAnACIAJwApADsACgAgAAoAIAAvAC8AIABNAG8AcgBlACAAcwBpAG4AZwBsAGUAIABxAHUAbwB0AGUAcwAgAHQAaABhAG4AIABkAG8AdQBiAGwAZQAgAHEAdQBvAHQAZQBzAC4ACgAoACIAXAAiACcAJwAiACkAOwAKACgAIgBcACIAJwAnACIAKQA7AAoAIAAKACAALwAvACAAVAB3AG8AIABvAGYAIABlAGEAYwBoAC4ACgAoACIAXAAiAFwAIgAnACcAIgApADsACgAoACIAXAAiAFwAIgAnACcAIgApADsACgAKAAoAKAAiAEgAZQAnAHMAIABzAGEAeQBpAG4AJwA6ACAAXAAiAEgAbwB3ACcAcwAgAGkAdAAgAGcAbwBpAG4AJwA/AFwAIgAgAEQAbwBuACcAdAAgAGEAcwBrACAAbQBlACAAdwBoAHkALgAiACkAOwAKACgAIgBIAGUAJwBzACAAcwBhAHkAaQBuACcAOgAgAFwAIgBIAG8AdwAnAHMAIABpAHQAIABnAG8AaQBuACcAPwBcACIAIABEAG8AbgAnAHQAIABhAHMAawAgAG0AZQAgAHcAaAB5AC4AIgApADsACgAgAAoAIAAvAC8AIABTAG8AbQBlAHcAaABhAHQAIABtAG8AcgBlACAAcgBlAGEAbAAtAHcAbwByAGQAIABlAHgAYQBtAHAAbABlACAAMgAuAAoAKAAnAHYAYQByACAAYgBhAGMAawBzAGwAYQBzAGgAIAA9ACAAIgBcAFwAIgAsACAAZABvAHUAYgBsAGUAUQB1AG8AdABlACAAPQAgAFwAJwAiAFwAJwA7ACcAKQA7AAoAKAAnAHYAYQByACAAYgBhAGMAawBzAGwAYQBzAGgAIAA9ACAAIgBcAFwAIgAsACAAZABvAHUAYgBsAGUAUQB1AG8AdABlACAAPQAgAFwAJwAiAFwAJwA7ACcAKQA7AA==)

Rome should do the same when printing string literal tokens. 

## Input

```ts
 // Single double quote.
('"');
('"');
 
 // Single single quote.
("'");
("'");

// One of each
"\"'";
'"\'';

// One of each with unnecessary escapes.
("\"'");
("\"'");
 
 // More double quotes than single quotes.
('"\'"');
('"\'"');
 
 // More single quotes than double quotes.
("\"''");
("\"''");
 
 // Two of each.
("\"\"''");
("\"\"''");


("He's sayin': \"How's it goin'?\" Don't ask me why.");
("He's sayin': \"How's it goin'?\" Don't ask me why.");
 
 // Somewhat more real-word example 2.
('var backslash = "\\", doubleQuote = \'"\';');
('var backslash = "\\", doubleQuote = \'"\';');
```

## Prettier

```ts
// Single double quote.
('"');
('"');

// Single single quote.
("'");
("'");

// One of each
("\"'");
("\"'");

// One of each with unnecessary escapes.
("\"'");
("\"'");

// More double quotes than single quotes.
('"\'"');
('"\'"');

// More single quotes than double quotes.
("\"''");
("\"''");

// Two of each.
("\"\"''");
("\"\"''");

("He's sayin': \"How's it goin'?\" Don't ask me why.");
("He's sayin': \"How's it goin'?\" Don't ask me why.");

// Somewhat more real-word example 2.
('var backslash = "\\", doubleQuote = \'"\';');
('var backslash = "\\", doubleQuote = \'"\';');

```

## Rome

```ts
// Single double quote.
('"');
('"');

// Single single quote.
("'");
("'");

// One of each
"\"'";
'"\'';

// One of each with unnecessary escapes.
("\"'");
("\"'");

// More double quotes than single quotes.
('"\'"');
('"\'"');

// More single quotes than double quotes.
("\"''");
("\"''");

// Two of each.
("\"\"''");
("\"\"''");

("He's sayin': \"How's it goin'?\" Don't ask me why.");
("He's sayin': \"How's it goin'?\" Don't ask me why.");

// Somewhat more real-word example 2.
('var backslash = "\\", doubleQuote = \'"\';');
('var backslash = "\\", doubleQuote = \'"\';');

```

## Expected

Rome's formatting to match prettier's.

ematipico:
## Summary

This PR closes #2443 

In order to close this PR, I had to start some work around https://github.com/rome/tools/issues/2444 , this work was needed because if the algorithm decides to swap quotes, then inside the content we might have some quote that needs escaping.

The rule to swap quotes inside a string literal depends on the literal itself:
- inside directives, it's only eligible if the content itself doesn't contain any quotes at all;
- inside string literals, it depends by the number of quotes inside the content itself; for example, let's say that the default quote is `"`, but then the content itself contains 3 double quotes and 1 quote (e.g. `" content \"\"\" ' "`), then the algorithm says that the preferred quote is `'` because it will reduce the number of escapes inside the string itself;

Considered the previous logic, when swapping a quote, we also need to escape those quotes too. In the previous example, `'` has to become `\'`. So while I was at it, I created a function that is in charge to reduce the number of escapes inside a string. This function, for now, works only on quotes and backslash. It doesn't work on other characters like `\n` or `\r`. And it doesn't work on other cases where the escape can be removed, for example `"b\all"`, where the end result will be `"ball"`.

I documented the algorithm, how it works and the edge cases that I have found.

## Test Plan

I added some cases:
- when the quotes can be swapped
- when the quotes can't be swapped
- when reducing the number of escape, they can be removed
- directives

### Metrics

PR
```
**File Based Average Prettier Similarity**: 71.99%  
**Line Based Average Prettier Similarity**: 65.23%  
```

`main`

```
**File Based Average Prettier Similarity**: 71.96%  
**Line Based Average Prettier Similarity**: 65.15%  
```


ematipico:
!bench_formatter

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,257 @@ pub(crate) fn format_with_semicolon(
...
+    let preferred = quote_style;
+    let alternate = match quote_style {
+        QuoteStyle::Double => QuoteStyle::Single,
+        QuoteStyle::Single => QuoteStyle::Double,
+    };
+
+    if quoteless_content.contains(preferred.as_char())
+        || quoteless_content.contains(alternate.as_char())
+    {
+        let preferred_quotes_count = quoteless_content.matches(preferred.as_char()).count();
```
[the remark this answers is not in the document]
ematipico: Thank you!

crates/rome_js_formatter/src/options.rs
```diff
@@ -82,3 +82,40 @@ impl fmt::Display for QuoteStyle {
...
+            QuoteStyle::Single => "\\'",
+        }
+    }
+
+    pub fn as_bytes(&self) -> u8 {
+        match self {
+            QuoteStyle::Double => b'"',
+            QuoteStyle::Single => b'\'',
+        }
+    }
```
MichaReiser, line 113: Nit: You can do self.as_char() as u8

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+///
+/// But then, we choose the single quote as preferred quote and we would have only one quote that is escaped,
+/// resulting into a string like this:
+/// ```js
+/// (' content """ dont\'t ')
+/// ```
+/// Like this, we reduced the number of escaped quotes.
+fn computed_preferred_quote_style(
+    quoteless_content: &str,
+    formatter: &Formatter<JsFormatOptions>,
```
MichaReiser, line 416: Nit: You only need the formatter options, or more specifically, the quote style. 

I would pass the quoteless content and the preferred quote style only.

```
compute_preferred_quote_style(quoteless_content: &str, preferred_quotes: QuoteStyle) -> (QuoteStyle, QuoteStyle)
```

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+) -> (QuoteStyle, QuoteStyle) {
```
MichaReiser, line 417: What's the meaning of the first and second quote style?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+
+    let preferred = quote_style;
+    let alternate = match quote_style {
+        QuoteStyle::Double => QuoteStyle::Single,
+        QuoteStyle::Single => QuoteStyle::Double,
+    };
+
+    if quoteless_content.contains(preferred.as_char())
+        || quoteless_content.contains(alternate.as_char())
+    {
```
MichaReiser, line 428: `contains` already does a full path. Do you think we could iterate over all characters once and count the preferred and alternate quotes and then determine the preferred style based on the counts?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+/// as opposed to the previous example, we have a signal that says that we should keep the current
+/// character. Then we do so. The third iteration comes along and we find `'`. We still have the
+/// [CharSignal::Keep]. We do so and then we set the signal to [CharSignal::Idle]
+///
+fn reduce_escapes_from_string(
+    raw_content: &str,
+    preferred_quote: QuoteStyle,
+    alternate_quote: QuoteStyle,
+) -> String {
+    let mut reduced_string = String::new();
```
MichaReiser, line 502: Any chance we could do this without allocating a new string? Or is this only called if we want to change the quotes? 

The way I could see this work is that you implement an iterator that returns the next character when next is called. Peeking a single character should be sufficient or not?

Similar to the code at the bottom of https://www.notion.so/rometools/Code-Clean-Text-822f8f2e51c04b00a6daf0728f1053a1

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+pub(crate) enum LiteralType {
+    Directive,
+    String,
+}
+
 pub(crate) fn format_string_literal_token(
     token: JsSyntaxToken,
     formatter: &Formatter<JsFormatOptions>,
+    literal_type: LiteralType,
 ) -> FormatElement {
```
MichaReiser, line 605: Nit. An alternative to this utiltiy functions would be to have a `FormatStringLiteralToken` enum that implements `Format`. 

```
enum FormatStringLiteralToken<'a> {
  Directive(token: &'a JsSyntaxToken),
  String(token: &'a JsSyntaxToken)
}

impl Format for FormatStringLiteralToken<'_> {
...
}
```

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
-    let (primary_quote_char, secondary_quote_char) = match formatter.options().quote_style {
-        QuoteStyle::Double => ('"', '\''),
-        QuoteStyle::Single => ('\'', '"'),
+    let literal = token.text_trimmed();
+
+    let content = match literal_type {
+        LiteralType::Directive => {
+            let raw_content = &literal[1..literal.len() - 1];
+
+            if raw_content.contains(['"', '\'']) {
```
MichaReiser, line 612: This contains check is also O(n). Any chance we could have all this logic inside `normalize_new_lines`?

MichaReiser (commented):
Nice! I havent' went through all the code but I think it would be good if we could keep a solution that returns a Borrowed string if there's no need to change the string. A way I could see this is to have 1 pass to determine the right quotes and then an Iterator that returns character by character on how the string should look like.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Turns out that we don't need `.contains`. The `.fold` iterator is enough!

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Everything is possible 😛

Ideally we could use a single function/enum that does everything: quoting, escaping, newlines normalization, etc. with one single iterator. This might help.

Although, I would like to defer this refactor to later, I think it's best to first fix all the issues we have around quoting, escaping, etc. Doing so we add new tests, trying to cover as much edge cases as possible. Then, we could do a good refactor. So if we break something, the tests won't pass.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Yes, I think it's doable. The solution now covers all cases, so if the refactor blows up, tests won't pass. I'll take a stab at it

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Good call. I will add it to the documentation.

First quote: the preferred one
Second quote: the alternate quote (the one won't use)

crates/rome_js_formatter/src/utils/mod.rs
ematipico: > Or is this only called if we want to change the quotes?

No, this will be called all the times (it's only a coincidence now that it's called for swapping the quotes, but as mentioned in the description, this will be done always) , even when we don't need to change the quotes. For example `"me\at"` will change to `"meat"`. As you can see the quotes are the same, but an escape has been removed.

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: Do we need to return both? How about adding a `other()` fn to `QuoteStyle` that returns the other style.

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: Sorry, that notion is intentionally private but I can share the snipped here


```rust
use std::borrow::Cow;
use std::iter::Peekable;
use std::str::Chars;


pub fn clean_jsx_text(input: &str) -> Cow<str> {
    let mut cleaned = CleanJsxWhitespace::new(input);
    let mut actual = input.chars();

    let mut pos = 0;

    loop {
        let next_cleaned = cleaned.next();
        let next_actual = actual.next();

        match (next_cleaned, next_actual) {
            (Some(cleaned_char), Some(actual_char)) => {
                if cleaned_char == actual_char {
                    // all good
                    pos += cleaned_char.len_utf8();
                } else {
                    // difference, copy what we have so far, then take everything from cleaned
                    let mut result = String::from(&input[..pos]);

                    result.push(cleaned_char);

                    for cleaned_char in cleaned {
                        result.push(cleaned_char);
                    }

                    return Cow::Owned(result);
                }
            }
            (None, Some(_)) => {
                // anything following in actual now can only be truncated away text
                return Cow::Borrowed(&input[..pos]);
            }
            (Some(_), None) => {
                // This should never happen because cleaned only removes
                unreachable!()
            }
            (None, None) => return Cow::Borrowed(input),
        }
    }
}

#[derive(Debug)]
struct CleanJsxWhitespace<'a> {
    inner: Peekable<Chars<'a>>,
    /// Leading/trailing has the same definition as in the Formatter/Parser
    /// trailing: Any whitespace following a character up to, but not including a new line
    /// leading: Any whitespace before the next token, including all line breaks before the last token.
    /// Flag set to true as soon as the iterator starts processing trailing trivia (
    is_trailing_whitespace: bool,

    /// Tracks if there has been any trailing whitespace
    has_trailing_whitespace: bool,

    /// Tracks if there has been a new line
    has_leading_newline: bool,

    /// Tracks if this is the first line
    first_line: bool,
}

impl<'a> CleanJsxWhitespace<'a> {
    pub fn new(input: &'a str) -> Self {
        Self {
            inner: input.chars().peekable(),
            has_trailing_whitespace: false,
            has_leading_newline: false,
            is_trailing_whitespace: false,
            first_line: true,
        }
    }
}

// * First Line:
//      * Renders single space even if the text is all empty
//      * Render single space before first character.

impl Iterator for CleanJsxWhitespace<'_> {
    type Item = char;

    fn next(&mut self) -> Option<Self::Item> {
        loop {
            if let Some(next) = self.inner.peek() {
                match next {
                    ' ' | '\t' => {
                        // Eat over space
                        self.inner.next();

                        if self.first_line {
                            match self.inner.peek() {
                                // `<>    </>` -> `<> </>`
                                None => {
                                    return Some(' ');
                                }
                                // Collapse succeeding whitespace
                                Some(' ' | '\t' | '\n' | '\r') => {}
                                // `<>  a<>` -> `<> a</>`
                                Some(_) => {
                                    return Some(' ');
                                }
                            }
                        }

                        // Track all trailing trivia
                        if self.is_trailing_whitespace {
                            self.has_trailing_whitespace = true;
                        }
                    }

                    '\n' | '\r' => {
                        // Only track the newline if it has been after a non-whitespace character.
                        // (e.g. don't track newlines if the text so far has been empty)
                        if self.is_trailing_whitespace {
                            self.has_leading_newline = true;
                        }

                        self.has_trailing_whitespace = false;
                        self.is_trailing_whitespace = false;
                        self.first_line = false;

                        // Skip new line character
                        self.inner.next();
                    }
                    _ => {
                        // In case there's any whitespace or newline between two characters, print one space
                        // `<>a   a</>` -> `<>a a</>`
                        if self.has_trailing_whitespace || self.has_leading_newline {
                            self.has_trailing_whitespace = false;
                            self.has_leading_newline = false;

                            return Some(' ');
                        }

                        self.is_trailing_whitespace = true;

                        return self.inner.next();
                    }
                }
            } else {
                // `<>a\n b </>` -> `<>a b </>`
                if self.has_trailing_whitespace {
                    self.has_trailing_whitespace = false;
                    return Some(' ');
                }

                return None;
            }
        }
    }
}

#[cfg(test)]
mod tests {
    use crate::clean_jsx_text;

    #[test]
    fn clean_jsx_text_works() {
        assert_eq!("", clean_jsx_text(""));
        assert_eq!(" ", clean_jsx_text(" "));
        assert_eq!("Foo", clean_jsx_text("Foo"));
        assert_eq!(" Foo", clean_jsx_text(" Foo"));
        assert_eq!("Foo", clean_jsx_text("\nFoo"));
        assert_eq!(" Foo", clean_jsx_text("\tFoo"));
        assert_eq!("Foo", clean_jsx_text("\n \t Foo"));
        assert_eq!("Foo", clean_jsx_text("\n \t \n \t\nFoo"));
        assert_eq!(" Foo bar lorem", clean_jsx_text(" Foo bar lorem"));
        assert_eq!("Foo ", clean_jsx_text("Foo "));
        assert_eq!("Foo", clean_jsx_text("Foo\n"));
        assert_eq!("Foo ", clean_jsx_text("Foo\t"));
        assert_eq!("Foo", clean_jsx_text("Foo\t \n "));
        assert_eq!("Foo", clean_jsx_text("Foo\n \t \n \t\n"));
        assert_eq!("Foo Bar", clean_jsx_text("Foo\n \t\t\n \tBar"));
        assert_eq!(
            "Foo Bar",
            clean_jsx_text("\n \t\t\n \tFoo\n \t\t\n \tBar\n \t\t\n \t")
        );
    }
}
```

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: The `Format` trait still exists inside of the `rome_formatter` trait. `FormatRule` is just a "hack" to work around Rust's orphanage rule. But calling `node.format()` returns a struct that implements `Format`. 

> How should I do that?

Use the example code from above. That should work

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: I don't mind adding the tests first but do believe that it's important to keep an eye on performance. 

But this branch might be less of an issue because it only applies for directives, where there are only a few.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+                    quote_style.as_char()
+                );
+
+                Cow::Owned(normalize_newlines(&final_content, ['\r']).into_owned())
+            }
+        }
+
+        LiteralType::String => {
+            // literal, in our syntax, might not have quotes so we need to check if they start with a possible quote
+            // is so, they are eligible for possible string manipulation
```
MichaReiser, line 629:
```suggestion
            // literal, in our syntax, might not have quotes so we need to check if they start with a possible quote
            // and if so, they are eligible for possible string manipulation
```

When is it possible for a string literal to not start with quotes? Would you mind making an example here.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,23 +386,270 @@ pub(crate) fn format_with_semicolon(
...
+                    reduce_escapes_from_string(raw_content, preferred_quote, alternate_quote);
+
+                let final_content = format!(
+                    "{}{}{}",
+                    preferred_quote.as_char(),
+                    polished_raw_content,
+                    preferred_quote.as_char()
+                );
+
+                Cow::Owned(normalize_newlines(&final_content, ['\r']).into_owned())
```
MichaReiser, line 647: I thin it would be good to avoid too many iterations in this branch because it runs for EVERY string literal. 

Currently, this iterates over the string three times and always allocates a new string to where it only used two iterations before and only allocated if the quotes didn't match. 

I think it should be possible to:
* determine the preferred quotes by iterating once
* Clean the new-lines and reduce the escapes in a second pass
* Only change the quotes if they didn't already match.

crates/rome_js_formatter/src/utils/mod.rs
IWANABETHATGUY: > 

Thanks.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: > This contains check is also O(n). Any chance we could have all this logic inside `normalize_new_lines`?

I did a quick check, `normalize_new_lines` is very generic method that belong to `rome_formatter`, it's also used to normalize the text inside the comments. I don't think we can make mix it with the logic of this PR.

ematipico:
!bench_formatter

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: I wouldn't change `normalize_new_lines` but instead, integrate the logic into your iterator

crates/rome_js_formatter/src/utils/mod.rs
ematipico: An example:

```
export { a as "b" };
export { a as b };
```

ematipico:
!bench_formatter

ematipico:
I will probably need to clean the code, but now the solution is quite fast and we do just two iterations in total to normalize the entire string.

crates/rome_js_formatter/src/js/module/import_assertion_entry.rs
```diff
@@ -29,7 +30,7 @@ impl FormatNodeFields<JsImportAssertionEntry> for FormatNodeRule<JsImportAsserti
                 formatted_key,
                 colon_token.format(),
                 space_token(),
-                format_string_literal_token(value_token?, formatter),
+                FormatLiteralStringToken::from_string(&value_token?).format(formatter)?
```
MichaReiser, line 33: No need to call `format` here, `formatted` will do that for you
```suggestion
                FormatLiteralStringToken::from_string(&value_token?)
```

crates/rome_js_formatter/src/js/module/literal_export_name.rs
```diff
@@ -11,6 +11,6 @@ impl FormatNodeFields<JsLiteralExportName> for FormatNodeRule<JsLiteralExportNam
     ) -> FormatResult<FormatElement> {
         let JsLiteralExportNameFields { value } = node.as_fields();
 
-        Ok(format_string_literal_token(value?, formatter))
+        FormatLiteralStringToken::from_string(&value?).format(formatter)
```
MichaReiser, line 14: I would use `formatted` here. More idiomatic. 

```suggestion
        formatted![formatter, [FormatLiteralStringToken::from_string(&value?)]]
```

crates/rome_js_formatter/src/ts/module/import_type.rs
```diff
@@ -27,7 +27,7 @@ impl FormatNodeFields<TsImportType> for FormatNodeRule<TsImportType> {
                     .with_or_empty(|token| formatted![formatter, [token, space_token()]]),
                 import_token.format(),
                 l_paren_token.format(),
-                format_string_literal_token(argument_token?, formatter),
+                FormatLiteralStringToken::from_string(&argument_token?).format(formatter),
```
MichaReiser, line 30:
```suggestion
                FormatLiteralStringToken::from_string(&argument_token?),
```

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
-    };
-    let content =
-        if quoted.starts_with(secondary_quote_char) && !quoted.contains(primary_quote_char) {
-            let s = &quoted[1..quoted.len() - 1];
-            let s = format!("{}{}{}", primary_quote_char, s, primary_quote_char);
-            Cow::Owned(normalize_newlines(&s, ['\r']).into_owned())
+/// This signal is used to tell to the next character what it should do
+enum CharSignal {
+    /// There hasn't been any signal
+    Idle,
```
MichaReiser, line 393: Nit: How about `None`?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
+
+        let content = string_cleaner.clean_text();
+
+        Ok(formatter.format_replaced(
+            token,
+            Token::from_syntax_token_cow_slice(content, token, token.text_trimmed_range().start())
+                .into(),
+        ))
+    }
 }
```
MichaReiser, line 827: Nit: It would help in my view to keep the `struct` definition and its implementation next to each other. Makes it easier to navigate the file.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
+            token,
+            chosen_quote_style,
+            preferred_quote: None,
+            raw_content_has_quotes: None,
+            literal_starts_with_quote: None,
+        }
+    }
+
+    pub fn clean_text(&mut self) -> Cow<str> {
+        // Cow::Owned(String::new())
```
MichaReiser, line 448: Delete?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
+                    literal_starts_with_quote = Some(
+                        current_character == chosen_quote_char
+                            || current_character == alternate_quote_char,
+                    );
+                    return (preferred_quotes_counter, alternate_quotes_counter);
+                } else if index == char_count - 1 {
+                    return (preferred_quotes_counter, alternate_quotes_counter);
+                }
+                if current_character == self.chosen_quote_style.as_char() {
+                    (preferred_quotes_counter + 1, alternate_quotes_counter)
```
MichaReiser, line 520: Nit. Might be easier to read if you use `return` statements in all branches or use `if ... else if ...` branching without returns.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
+    /// Excluding the quotes at the start and beginning, we have three double quote and one single quote.
+    /// If we decided to keep them like this, we would have three escaped quotes.
+    ///
+    /// But then, we choose the single quote as preferred quote and we would have only one quote that is escaped,
+    /// resulting into a string like this:
+    /// ```js
+    /// (' content """ dont\'t ')
+    /// ```
+    /// Like this, we reduced the number of escaped quotes.
+    fn compute_preferred_quote(&mut self) {
```
MichaReiser, line 498: Nit: This could probably be a standalone function that you call from `Format` of `FormatStringLiteralToken`. It allows you to change `preverred_quote` to `QuoteStyle` rather than having an `Option`.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
+        let literal = self.get_token().text_trimmed();
+        let preferred_quote = self
+            .preferred_quote
+            .expect("You must compute the preferred quote style first");
+
+        // literal, in our syntax, might not have quotes so we need to check if they start with a possible quote
+        // and if so, they are eligible for possible string manipulation
+        // For example:
+        // export { a as "b" };
+        // export { a as b };
```
MichaReiser, line 557: But these examples use quotes? What are examples where a string isn't using any quotes?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -386,29 +387,442 @@ pub(crate) fn format_with_semicolon(
...
+
+        for (start, part) in raw_content.match_indices(CHARACTERS_THAT_COULD_KEEP_THE_ESCAPE) {
+            if start - last_end >= 1 {
+                // This is the case where we don't have consecutive characters and if so, we have to reset the signal.
+                // An example is the following: " \\u2028 ' "
+                // After the two backslash, we have a character that is not a quote. So we reset the signal and we
+                // iterate over the single quote, we don't deal with any edge case.
+                signal = CharSignal::Idle;
+            }
+            reduced_string.push_str(&raw_content[last_end..start]);
```
MichaReiser, line 655: I'm probably not understanding this part correctly but isn't this always pushing to the string if the string contains any `'` or `"` even if it already is "correctly" escaped?

MichaReiser (commented):
Nice!, this improved performance a lot

crates/rome_js_formatter/src/utils/mod.rs
ematipico: **This** is the example where a token doesn't use quotes. `b` is a token, and our grammar doesn't differentiate against the one with quotes:

```
// import { a as b } from "c"
//          ^
// export { a as "a-b-c" };
//               ^^^^^^^
JsLiteralExportName = value: ('ident' | 'js_string_literal')

```

crates/rome_js_formatter/src/utils/mod.rs
ematipico: True, and doing I will have to pass `raw_content_has_quotes` and `literal_starts_with_quote` as parameter too (because they are calculated while computing the preferred quote). I'd prefer to keep them inside a struct so I can document them better. 

Otherwise I would have to do something like:

```rs
fn format()  {
  let (preferred_quote, literal_starts_with_quote, raw_content_has_quotes) = self.computed_quote(formatter);
  let string_cleaner = LiteralStringNormaliser::new(self);
  string_cleaner.normalise_string(preferred_quote, literal_starts_with_quote, raw_content_has_quotes);
}
```

crates/rome_js_formatter/src/utils/mod.rs
ematipico: No when we are using `.match_indices`. This function returns us the index and the **matched** pattern of each iteration. So if we match with `"XXX"` on `"helloXXXworldXXX!"`, `.match_indices` will iterate twice where `part` will always be `"XXX"` and `start` is its position (index) inside the original string.

In this particular point of the iteration, with `&raw_content[last_end..start]` we are retrieving the string that we had **before** each iteration. In out example they would be `"hello"` and `"world"`.

In our case, and in case we have consecutive matches, for example in "content \' ", what we store is an empty string because `last_end` and `start` are the same, and the returned slice is an empty string.

crates/rome_js_formatter/src/utils/mod.rs
MichaReiser: Right. But the `LiteralExportName` isn't a string literal... it's an identifier. In which case we shouldn't call this method?

MichaReiser (approved):

MichaReiser:
It could help to have a few unit tests in that file that just test different cases of the string manipulation and make sure that a Borrowed string is returned when there's no need to replace any content.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: Yes, but that's it's the grammar and how's the AST returned. Both are tokens. Unless we change the grammar and the parsing, there isn't much I can do. The check has to stay, regardless where we put it.

crates/rome_js_formatter/src/utils/mod.rs
ematipico: I really don't mind the `expect` and it's a tradeoff I am willing to take, because I really don't like passing around `literal_starts_with_quote` and `raw_content_has_quotes` as boolean, because they can't be documented properly.

MichaReiser: What I had in min is to introduce a new struct that has the `literal_starts_with_quote`, `raw_content_has_quotes` and `preferred_quote` fields. It makes it easier to pass these values around and you have a place to document the fields.

PR #2626, opened 2022-05-29
test(rome_js_formatter): support report prettier metric as a json file

IWANABETHATGUY:
## Summary
part of https://github.com/rome/tools/issues/2555
To achieve printing diff between `main` branch and pull request branch,  a few steps we need to do list below

1. save `prettier compatibility metric` on `main`  branch 
2. switch to `pull request branch` 
3. save `prettier compatibility metric` on `pull request`  branch
4. compare two metrics and echo the diff.

Obviously, `Markdown`  is not a good format for comparing, so we need to add an extra format to save metrics, which is `Json` 
just like `Parser conformance`

This pull request introduces these two changes:
1. Add an option to report the prettier compatibility metric in `json` or `markdown` format.
2. Support choosing file_name to save.

## Test Plan
No need to test

### DIFF
--- a/crates/rome_js_formatter/tests/prettier_tests.rs
+++ b/crates/rome_js_formatter/tests/prettier_tests.rs
@@ -278,32 +278,21 @@ enum ReportType {
     Markdown,
 }
 
-#[derive(Debug, PartialEq, Serialize, Default)]
-struct SingleFileJsonReport {
-    filename: String,
-    single_file_compatibility: f64,
-}
-
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone, Default, Serialize)]
 struct SingleFileMetricData {
     filename: String,
     single_file_compatibility: f64,
+    #[serde(skip)]
     diff: String,
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, Serialize)]
 struct PrettierCompatibilityMetricData {
     file_based_average_prettier_similarity: f64,
     line_based_average_prettier_similarity: f64,
     files: Vec<SingleFileMetricData>,
 }
 
-#[derive(Debug, PartialEq, Serialize, Default)]
-struct ReportJson {
-    file_based_average_prettier_similarity: f64,
-    line_based_average_prettier_similarity: f64,
-    files: Vec<SingleFileJsonReport>,
-}
 impl FromStr for ReportType {
     type Err = String;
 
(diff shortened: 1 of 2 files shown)


ematipico:
Could you add few lines in the description of the PR about:
- why you're adding this to the tests and what kind of problem we're trying to solve;
- a reference to an issue where this new addition was discussed;

yassere (commented):
Thanks for working on this! I agree that it makes sense to have a json option so that changes can be compared more easily in CI.

Currently, this PR duplicates a lot of the report calculations, which makes it more likely to fall out of sync if we modify our metrics in the future. Could you extract those calculations so that the `report_json` and `report_markdown` methods are only directly responsible for the differences that are specific to that report type?

Maybe there could be a method that produces a `PrettierReport` that the individual `report_X` methods know how to print into different formats?

IWANABETHATGUY:
> 

agree

yassere:
We could also update the `.gitignore` to ignore `crates/rome_js_formatter/report.*` rather than just the markdown file.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -270,6 +272,35 @@ enum MatchCategory {
...
 }
 
+#[derive(Debug, PartialEq, Eq)]
+enum ReportType {
+    Json,
+    Markdown,
+}
+
+#[derive(Debug, PartialEq, Serialize, Default)]
+struct PeerFileJsonReport {
```
MichaReiser, line 282: Is this the report per file? `PerFileJsonReport`

MichaReiser (changes requested):
Agree with Yasser. I think it would be good to factor out the computation into a helper function that returns the numbers and the markdown/json functions only print the results differently.

IWANABETHATGUY:
I have refactored the code, I think I am ready for reviewing.

crates/rome_js_formatter/tests/prettier_tests.rs
```diff
@@ -270,6 +272,50 @@ enum MatchCategory {
     Match,
 }
 
+#[derive(Debug, PartialEq, Eq)]
+enum ReportType {
+    Json,
+    Markdown,
+}
+
+#[derive(Debug, PartialEq, Serialize, Default)]
```
MichaReiser, line 281: Nit: It could be an option to merge this struct with `SingleFileMetricData` and use `#[serde(skip)] ` on the `diff` field.

You can then also merge the `PrettierCompatibilityMetricData` struct with the `ReportJson` struct

MichaReiser (approved):
Thanks for doing the refactor. That should help us to keep the implementations in sync in the future. 

I've a small comment that may help to reduce some code. Let me know if you want to address it before merging or prefer to keep it as is.

crates/rome_js_formatter/tests/prettier_tests.rs
IWANABETHATGUY: That's very useful, thanks.

IWANABETHATGUY: I would like to address this.

IWANABETHATGUY: Done

Issue #2425, opened 2022-04-13
Break long object literal members over two lines


MichaReiser:
[...]

PR #2627, opened 2022-05-30
feat(rome_js_formatter): Break long object literal members over two lines #2425

Dominionys:
## Summary
 Fixes https://github.com/rome/tools/issues/2425. 
 
 This is the initial draft for issue "Break long object literal members over two lines".
 Prettier uses [`printAssigment`](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L25-L75) for formatting [AssignmentExpression](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L77), [VariableDeclarator](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L89), [ClassProperty](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/class.js#L195) and [Property](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/property.js#L92).  `printAssigment` can return different IR swhich are based on a layout variants from [`chooseLayout`](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L93) function.
 
-  `"break-after-operator"`
- `"never-break-after-operator"`
- `"fluid"`
- `"break-lhs"`
 - `"chain"`
 - `"chain-tail"`
  - `"chain-tail-arrow-chain"`
   - `"only-left"`

We're going to focus on [`chooseLayout`](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L93-L174) function and branches which are relevant for object literal members. 

- ["only-left"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L97-L99) branch isn't relevant for `JsPropertyObjectMember` because when object member has only left part it's `JsShorthandPropertyObjectMember`.  

- ["chain", "chain-tail-arrow-chain", "chain-tail"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L105-L121) these variants can be only if `shouldUseChainFormatting` is true. It seems that `shouldUseChainFormatting` is true only for match `isAssignmentOrVariableDeclarator`.

- ["break-after-operator"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L122-L129) I'm not sure about this conditional because technically it can be relevant for object member, but I looked prettier [commit](https://github.com/prettier/prettier/commit/0a647a0d31bdbf38230aa256ab1f37062ee98969) when this condition was added and it was for assignment.

- ["never-break-after-operator"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L131-L139) Prettier doesn't add break line if right node is require call. it works for object member. [Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEEBGArTAXk2AB0pNMdqbbqlMoIYAlOARwFcBLAJzgAUASgA05SnUk4G-Ln0GjyAXwDcIESAgAHGN3TJQAQ168IAdwAKxhGmQhDAGzOGAnrY05ehsAGs4MAGVDAFs4ABluKDhkADNHNDgPL19-AK1vSIBzZBheTkSQOGCcOAATUrKwwyhMzkNMuAAxCF5gwxhdGrtDThgIdRAACxhghwB1Qe54NHSwOACbKe4ANymXOzA0dxBIhN4YCy9Mttj4grw0AA8ArIc4AEVOZmikOIcEjXTePbscQxKHAMtLxIjAxtxSjBBsgAEwABk+pgSYy8WjswLge2W0Q0XGeh20thQhjQAFoomUygNZDx+Id6idXmcNAlgtwcnkCmhbg8nvBTu8CjB-uDIdCkDCNLlDNwHFkAMIQYKMwpoACsA04CQAKv8iW8PiBlvkAJJQCqwAJgEE6ACC5oCMBcdwFCSUSiAA). I haven't found test for this case. 

- ["break-lhs"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L141-L148) this variant is for [destructuring](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L227), types ([alias](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L252), [annotation](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L278)) and [arrow function declaration](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L300).

- ["break-after-operator"](https://github.com/prettier/prettier/blob/feb6b519ee0e23eeb272227825ad84ecf1af05ab/src/language-js/print/assignment.js#L176-L223) It uses for right nodes which are:

    1.  `isBinaryish` and `!shouldInlineLogicalExpression`. I haven't added `shouldInlineLogicalExpression` checking but there is a difference in IR between [them](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBDTBeTYAHSkx2woBNqbbqKGlMYAnAVzkwDIvMAKADyUAlgDcA9AD4AlABoSZBtjorKS7E1YduvAIwAmAMzHj8qAF8QskBAAOMYemShsLFhADuABVcI0yEGwAGw9sAE9-awAjFmwwAGs4GABlbABbOAAZYSg4ZAAzYLQ4aNiEpOTbOJyAc2QtEpA4NKi4ajbM7CgatmwauAAxCBY07BgHboDsNhgIKxAACxg0oIB1BeF4NCqwOGS-TbFNsICwNEiQHOKWGC9YmtGCosaAKzQAD2TaoLgARTYIPAnkFitYqixrgEothWkF5rYWDkYKthJQYAtkAAWAAMYPcxVWsVsAQRcGuojy1gAjgD4Hc7P4UNg0ABaXJtNrzFhwGnCbl3PqPJCFEGNYppYT1dhi75-Wl5YXPawwGEotEYpD6ZWxYRBWoAYQgaSFTTQAFZ5mxigAVGGMkWgkCiDgASSglAQKTAiPsAEF3ckYGEfsDiuZzEA). 
    2. `StringLiteralTypeAnnotation`,  
    3. `SequenceExpression`
    4. `ConditionalExpression` + `isBinaryish` and `!shouldInlineLogicalExpression`.
    5. `StringLiteral`
    6. [isPoorlyBreakableMemberOrCallChain](https://github.com/prettier/prettier/blob/feb6b519ee0e23eeb272227825ad84ecf1af05ab/src/language-js/print/assignment.js#L329-L370)

- ["never-break-after-operator"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L162-L171) the variant is for `TemplateLiteral`, `TaggedTemplateExpression`, `BooleanLiteral`, `NumericLiteral`, `ClassExpression` or for a member which key is [short](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L412-L430).

- ["fluid"](https://github.com/prettier/prettier/blob/67c2df0ffe8c1e8bf1425fdd248ee5dd4eb5699f/src/language-js/print/assignment.js#L173) for other cases.

## Open questions
1. I'm not sure about` "break-after-operator"` . Do we need it?

2. Do we need to add `shouldInlineLogicalExpression` conditional for `"break-after-operator"`?

3. I skipped [isPoorlyBreakableMemberOrCallChain](https://github.com/prettier/prettier/blob/feb6b519ee0e23eeb272227825ad84ecf1af05ab/src/language-js/print/assignment.js#L329-L370) for `"break-after-operator"`.

4. There is a problem with array [break-last-call.js.snap](https://github.com/rome/tools/compare/main...Dominionys:break_long_object_literal_members_over_two_lines#diff-ab0460fdc72367f12d0ea47dd51f37874303cf211d9bc2921374f25dca69b443) and function calls [edge_case.js.snap](https://github.com/rome/tools/compare/main...Dominionys:break_long_object_literal_members_over_two_lines#diff-4cea115d8cd21dc63350f69c5b7f843cd12490f3d0b3308e6d05608209a009e8) fluid formatting. I tried to make the same IR that prettier uses for [fruid layout](https://github.com/prettier/prettier/blob/feb6b519ee0e23eeb272227825ad84ecf1af05ab/src/language-js/print/assignment.js#L47-L56).  
```
                formatted![
                    formatter,
                    [
                        group_elements(formatted![formatter, [name.format()]]?),
                        colon_token.format(),
                        group_elements_with_options(
                            indent(soft_line_break_or_space()),
                            GroupElementsOptions {
                                group_id: Some(group_id)
                            }
                        ),
                        line_suffix_boundary(),
                        if_group_with_id_breaks(indent(value.clone()), group_id),
                        if_group_with_id_fits_on_line(value, group_id),
                    ]
                ]
            }
```
Prettier:
```
      const groupId = Symbol("assignment");
      return group([
        group(leftDoc),
        operator,
        group(indent(line), { id: groupId }),
        lineSuffixBoundary,
        indentIfBreak(rightDoc, { groupId }),
      ]);
```
Actually I don't understand how to explain to formatter that it needs to add a break line only if formatted value part + formatted key part exceed max width.


crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+
+    let name = name?;
+    let value = value?;
+
+    let value = match value {
+        JsAnyExpression::TsAsExpression(expression) => expression.expression()?,
+        _ => value,
+    };
+
+    const MIN_OVERLAP_FOR_BREAK: u8 = 3;
```
ematipico, line 100: This can be outside of the function.

ematipico, line 100: This can be outside of the function.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+        name,
+        colon_token: _,
+        value,
+    } = node.as_fields();
+
+    let name = name?;
+    let value = value?;
+
+    let value = match value {
+        JsAnyExpression::TsAsExpression(expression) => expression.expression()?,
```
ematipico, line 96: Could we add a small comment saying why are using a match on `TsAsExpression` ?

ematipico, line 96: Could we add a small comment saying why are using a match on `TsAsExpression` ?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+    }
+
+    if is_never_break_after_operator(&value, has_short_key)? {
+        return Ok(PropertyObjectMemberLayoutMode::NeverBreakAfterOperator);
+    }
+
+    Ok(PropertyObjectMemberLayoutMode::Fluid)
+}
+
+fn is_break_after_operator(value: &JsAnyExpression, has_short_key: bool) -> SyntaxResult<bool> {
```
ematipico, line 116: Can we document what are the criterias of this function?

ematipico, line 116: Can we document what are the criterias of this function?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
     }
+
+    if let JsAnyExpression::JsConditionalExpression(conditional) = &value {
+        let JsConditionalExpressionFields {
+            test,
+            question_mark_token: _,
+            consequent: _,
+            colon_token: _,
+            alternate: _,
+        } = conditional.as_fields();
```
ematipico, line 132: As we are just testing one field, we can just do `conditional.test()?.syntax()`

ematipico, line 132: As we are just testing one field, we can just do `conditional.test()?.syntax()`

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+
+        if is_binaryish_expression(test?.syntax()) {
+            return Ok(true);
+        }
+    }
+
+    if has_short_key {
+        return Ok(false);
+    }
```
ematipico, line 141: This check and the one that we have on lines 170:172 yields two different results, they are not paired with another check and they are used in two different functions. So, do we actually need this check here? It seems that this check is only used as early return. At this point, can't we pair this check with next one? It might make more sense.

And if we do need it, could we document why?

ematipico, line 141: This check and the one that we have on lines 170:172 yields two different results, they are not paired with another check and they are used in two different functions. So, do we actually need this check here? And if we do need, could we document why?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+        JsAnyLiteralExpression::JsStringLiteralExpression(_),
+    ) = &value
+    {
+        return Ok(true);
+    }
+
+    Ok(false)
+}
+
+fn is_never_break_after_operator(
```
ematipico, line 153: Could we document the criteria of this function?

ematipico, line 153: Could we document the criteria of this function?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+    value: &JsAnyExpression,
+    has_short_key: bool,
+) -> SyntaxResult<bool> {
+    if let JsAnyExpression::JsCallExpression(call_expression) = &value {
+        let JsCallExpressionFields {
+            callee,
+            optional_chain_token: _,
+            type_arguments: _,
+            arguments: _,
+        } = call_expression.as_fields();
```
ematipico, line 163: We can call `call_expression.callee()?.syntax().kind()` without the need of extracting all fields.

ematipico, line 163: We can call `call_expression.callee()?.syntax().kind()` without the need of the fields.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +22,175 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
...
+    }
+
+    Ok(false)
+}
+
+fn is_binaryish_expression(node: &JsSyntaxNode) -> bool {
+    matches!(
+        node.kind(),
+        JsSyntaxKind::JS_BINARY_EXPRESSION | JsSyntaxKind::JS_LOGICAL_EXPRESSION
+    )
```
ematipico, line 195: We use the term "binary like" expression. 

Also, we have already `JsAnyBinaryLikeExpression` inside `utils/binary_like_expression.rs`. We could add a function called `pub fn is_binary_like(express: JsAnyExpression)` that applies this logic.

Also, here in Rome `in` and `instanceof` expressions **are not** part of logical expressions (in babel instead, they are part of the logical expressions) and they separate nodes, which means that we need to check them too. That's why I think we could use that `enum` as helper

ematipico, line 195: We use the term "binary like" expression. 

Also, we have already `JsAnyBinaryLikeExpression`. We could add a function called `pub fn is_binary_like(express: JsAnyExpression)` that applies this logic.

Also, here in Rome `in` and `instanceof` expression **are not** part of logical expressions (in babel instead, they are part of the logical expressions), which means that we need to check them too. That's why I think we could use that `enum` as helper

ematipico (changes requested):
It would be great if we could create **new tests** in our test suite, so we know which cases are being covered in this PR.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
Dominionys: Thank you for review and feedback (:

We can use `JsAnyBinaryLikeExpression::can_cast` to check binary like expressions. Do you think that it's better to implement `pub fn is_binary_like(express: JsAnyExpression)` for better semantic?  Is it okay to write this function in `utils/binary_like_expression.rs`? Because this file is already big.

There is a file `parenthesized_expression.rs` which also contains `fn not_binaryish_expression(node: &JsSyntaxNode)`? Should it be replaced too?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
Dominionys: Hm, I'll delete this because I thought that prettier use the same IR for `TsAsExpression` and expression inside `TsAsExpression`, but when I added test to check it, I found that I was wrong :(
[Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAKYAdKACX3KwBOEANiXIUviQIYDOd2AYhIQLYCSdpNMrAwtBjEyFbJhx5cAEhoATOYKKlsAXmw0oAcwCutQgDooECAAcA3FikBfDY20GOS4aXKEldYTrB9CAGis8AhdRSmp6RhZ2Lh5fZxE3cUCpWQV4lXVNXX0jEwtk7GsAqGsASksoLHRA4ISKAEYAJgBmVrb2js6u7p6OqgByXqHm-uKpWtcGluGZ2d6Bua7+u2wOABUAT1M4MaChOsImxePjqjBaBmZWTm5aOP3JwiTJFPlFB5I1DW09GkNjMwVGy7XATUJHE6Q4ZnC6Ra4xO4CD6JCRSGRvdKfTI-HIA-IvQorBxOZEUDxeHysYoYGA0sFuCFQpldGERK7RW68JHKULPGn8mCpd48r5ZX7-PIVAW2ejYYmYtzkwjeXzU2kwelTZlM+os8KXKI3WLckIorACmlCzGinF-XKA83qoqOml-Qg0DZUADa9T8jT8zT8ABY-ABWPz1AAMEb9AeDYYj0d9-sDIfDUZjKfj6aTsdTCYzAF01a7CO6No1vcm42nE5mawXc1naxnq-mc-X23XC0StI4FWToJ5lZT-IEyiA-CAzDAAJZD5CgN0QADuAAU-gg6MgQDQSCuPdupwAjd1gADWcBgAGUaGw4AAZWdQODIABme7oOxAp5oF6v16mH+z5aMgXjfnAbDHnACgwQ+Yo0FocCGrwc7aDuNA6Hwk4gAAFjAbAkAA6rhs7wHQQFgHA15bmRs4AG5kRsO5gAwOHPl+hAwGu7paGwNDvp+34AFZ0AAHteIHkAAijoEDwIJJBflOQGEJxO4wFscB0GAhCzqYMA4aYemwERs5yDAuHIEGABsKnEF+RHuqYO7GdpFD0a+U4AI5yfAPFmNuKD0AAtC+MEwThhBwL5s7RTxiH8YpykgF+bCzmByrfnQUlwLJ8mvkgH5Kd+MA0MeZkWVZSD+iAwg0LOJAgYIbBJSg2mhjhOhfms5VBcVKX0TocAcFAcgIDeun6TAACCY3Xpp5DJXA1jWEAA)

MichaReiser:
> > Actually I don't understand how to explain to formatter that it needs to add a break line only if formatted value part + formatted key part exceed max width.
> 
> @MichaReiser You've been working recently on improving the printer. Is there a better way than `best_fitting` to do this to make sure that the `withGraphQLQuery(` [here](https://github.com/rome/tools/blob/e2c88a2980d6cc0624c8e48ed91f3a912d7c6394/crates/rome_js_formatter/tests/specs/prettier/js/last-argument-expansion/edge_case.js#L14) stays on the same line if it fits?

Isn't the conditionalGroup part in Prettier's IR coming from the call expression formatting? I would refrain from trying to fix all printing in a single PR and instead iterate (fix some printing for node X, then improve node Y, maybe cycle back to node X, etc.)

What I like to do is to open Prettier's playground (or Rome's playground and click on Prettier IR), enable the doc panel in the left bottom corner and inspect Prettier's generated IR. 

The few instructions that I find interesting are:

```
// 1: Indent
group(indent(line), {
  id: Symbol.for("assignment #1"),
}),

/// 2
breakParent,
```

I'm not entirely sure if our printer has the same behaviour as prettier's for the 1 case because we don't print whitespace if the content isn't followed by any "real" token. So if you have `some__d` where `_` is a space then our printer prints `some_d`. 

The second one is that Prettier inserts an explicit `breakParent` (called expand_parent in Rome) that forces the group of the `withGraphQLQuery` to be printed over multiple lines (so that it doesn't try flat printing).

yassere:
> Isn't the conditionalGroup part in Prettier's IR coming from the call expression formatting?

Yes, Prettier isn't using `conditionalGroup` for the behavior I'm referring to, so ideally we'll be able to match it without `best_fitting`.

I was asking how we can get just `render: withGraphQLQuery(` to stay on one line even if the rest of the withGraphQLQuery call expression is printed over multiple lines. But for now, I think this is a small enough regression that it doesn't need to block this PR. As you suggested, this can be iterated on in the future, so I'd be happy to have this merged after @ematipico is satisfied with it.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +20,148 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
             colon_token,
             value,
         } = node.as_fields();
+        let layout = get_object_member_layout(formatter, node)?;
```
MichaReiser, line 23: I love this `layout` approach. It splits the layout decision from the actual formatting which makes it so much easier to reason about the code.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
MichaReiser: I think we can document the rough idea of when we decide to break after an operator but listing all conditions would just repeat what's already written in code.

MichaReiser (approved):
This is an awesome and well-written contribution. Thank you so much! 

Let's add some more documentation and then get this merged :)

Dominionys:
> This is an awesome and well-written contribution. Thank you so much!
> 
> Let's add some more documentation and then get this merged :)

Thank you for your feedback!

- Does the same IR mean the same formatter output? 
 Because I tried to copy Prettier IR but there are still difference between Rome and Prettier output. 
 What is common solution for this? Try to write other IRs or to debug printer? 

- Do I understand correctly that new tests should be inside `test/specs/{js, ts, jsx, tsx}` folders? Do we also include these tests when we compare Prettier and Rome output?
 
- Question out of this issue: There is a function `fits_element_on_line` inside `src/printer/mod.ts` file. 
 There is `FormatElement::Token` variant and the first line 
 `state.line_width += state.pending_indent as usize * options.indent_string.len();` 
 But `FormatElement::Indent` variant in this function doesn't change `state.pending_indent`. 
 Is it okay?

MichaReiser:
>     * Does the same IR mean the same formatter output?
>       Because I tried to copy Prettier IR but there are still difference between Rome and Prettier output.
>       What is common solution for this? Try to write other IRs or to debug printer?

#2645 should get us closer. Ideally, we try to understand what's causing the difference and then decide if we want to align our behaviour with Prettier or not, and if not, document how our IR behaves differently

 
>     * Do I understand correctly that new tests should be inside `test/specs/{js, ts, jsx, tsx}` folders? Do we also include these tests when we compare Prettier and Rome output?

This is correct. To my knowledge, this tests aren't compared with Prettier. 

>     * Question out of this issue: There is a function `fits_element_on_line` inside `src/printer/mod.ts` file.
>       There is `FormatElement::Token` variant and the first line
>       `state.line_width += state.pending_indent as usize * options.indent_string.len();`
>       But `FormatElement::Indent` variant in this function doesn't change `state.pending_indent`.
>       Is it okay?

It's resetting the pending indent at the very last line.

https://github.com/rome/tools/blob/aa9c72801eae6e4327d0b6d9292382c22fc96e5d/crates/rome_formatter/src/printer/mod.rs#L747-L772

Or did you mean something else?

Dominionys:
> Or did you mean something else?

When we have `element` which is `FormatElement::Indent` in `fits_element_on_line` function. Should we increment `state.pending_indent`? Because now `FormatElement::Indent` increment only `args` indent and `args` indent doesn't count in `FormatElement::Token`.

MichaReiser:
> > Or did you mean something else?
> 
> Thank you for your answers!
> 
> When we have `element` which is `FormatElement::Indent` in `fits_element_on_line` function. Should we increment `state.pending_indent`? Because now `FormatElement::Indent` increment only `args` indent and `args` indent doesn't count in `FormatElement::Token`.

Good point! 

`Indent` gets only printed after a line break but line breaks inside of the fits function are either ignored (when soft), or the content fits if it's a hard line break, so the function returns. 

This is different to `print_elements` where the processing continues after a line break and the `indent` value is respected their.

https://github.com/rome/tools/blob/f3b053ecc0dff4f0ca6d803de9a2450b2170e1b0/crates/rome_formatter/src/printer/mod.rs#L212 

The only case where `pending_indent` can be true if the group appears right after a line break, in which case it's important that the fits function respects the indent. But I guess in theory, it would be OK to not increment the indent in args when processing an `Indent` IR element.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
ematipico: > There is a file parenthesized_expression.rs which also contains fn not_binaryish_expression(node: &JsSyntaxNode)? Should it be replaced too?

Good catch!! Yes I think we should replace that function too. 

> We can use JsAnyBinaryLikeExpression::can_cast to check binary like expressions. Do you think that it's better to implement pub fn is_binary_like(express: JsAnyExpression) for better semantic? Is it okay to write this function in utils/binary_like_expression.rs? Because this file is already big.

I think we can just call `cast_cast`. The enum has its own documentation that explains what it does, so using a function `is_binary_like` is not required.

Dominionys: Could I replace it in this RP? 

`JsAnyBinaryLikeExpression::can_cast` looks good! Thank you.

ematipico: Sure go ahead!

Dominionys: Replaced (:

crates/rome_js_formatter/tests/specs/prettier/js/method-chain/break-last-call.js.snap
```diff
@@ -76,62 +76,66 @@ it(
...
-          created: "11/01/2017 13:37",
-        },
-        {
-          message: "test",
-          messageType: "Email",
-          status: "Unknown",
-          created: "11/01/2017 13:37",
-        },
-      ],
+      "11/01/2017 13:36":
```
ematipico, line 79: I am looking at how prettier format this part, and it seems that it keeps the `[` on the same line. Do you think it's caused by something else? Or is there something related with your implementation?

Dominionys: Yes it's related with current PR, it's the same problem which @yassere and @MichaReiser discussed above about printing [withGraphQLQuery](https://github.com/rome/tools/blob/e2c88a2980d6cc0624c8e48ed91f3a912d7c6394/crates/rome_js_formatter/tests/specs/prettier/js/last-argument-expansion/edge_case.js#L14)
I tried to make identical [Prettier IR](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBrTBeTYAHSkzPLKJAAYBOAemoEZ6AmZgdkyY6VYFYqSEhQoBtEaIrFSU0QFs4aNAEMA5nCSYq8KgBpJc8ouXq4AFQCeAB03aQAZQCyD-YaOYMKmAFc0WqgBVKGwoCAB3KDdZD0wwACc4bzgAEwCaBmY2Tm5eAWjYgF8DGNEAXRJCgG4SED0QCGsYAEt0ZFAVePiIgAVOhDRkEBUAG3CVS0H6gCN4lTBsOBgHFUUAGWaoOGQAM1G0OBm5haWHa3nNtWQYeJ9DkDh5adSU1LWVKDUfMwAxCHj5N4Wp8hiofDAIHUQAALGDyEYAdWhzXgaHOYDgDgGKOaADcUZYhmBlFDNgd4jAenM1IDdvt7gArNAADwclxGcAAij4IPA6SMDvVzvFyUNpipniModZ4psYAjmikYNDkAAOahC7oHBFzaxDGVKODxXHbeoARx58CpjUGKBUaAAtFtUqkoYkLc1ElT1LSkHsBfcDvJmtdboH2VzLds-fT6jAJQqlSq+HG5s0RpcAMIQeS+h5ofhQvwWCW2-2CkC4u4ASSgr1gDgSzSaAEE6w4YJYOfyDoVCkA) and Rome IR.

- https://github.com/rome/tools/pull/2627/files#diff-5796188c9832d2142e13bbe9f1b6752a1181a36673a6c45e6f62e00a7ae0fc7cR27-R48

But for some reason printer gives different outputs for some cases.

ematipico: Thank you! I think we can merge once the CI is all green!

ematipico (approved):

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -14,10 +20,154 @@ impl FormatNodeFields<JsPropertyObjectMember> for FormatNodeRule<JsPropertyObjec
             colon_token,
             value,
         } = node.as_fields();
+        let layout = property_object_member_layout(formatter, node)?;
+
+        let name = formatted![formatter, [FormatMemberName::from(name?)]]?;
```
Dominionys, line 25: Could you please help to understand how to handle a case then we have an object key with double quotes and after formatting `FormatMemberName` we have key without quotes? Because it's necessary to know key length after formatting. Is it possible to get length of `FormatElement`?

MichaReiser: There's no way today to get the length of a FormatElement. Doing this in a precise way would require printing the whole tree. But we could change FormatMember to give access to the length (change it to a function that returns the length of the cleaned text)

ematipico: Oh yes, these members goes though a transformation where strings are normalised. We don't just elide quotes, but we also possible escapes too.

We don't have a method at the moment, but I think it can be easily implemented, just for `FormatElement::Token`:

```rs
impl FormatElement {
    pub fn token_size(&self) -> Option<&TextSize> {
        match self { 
            FormatElement::Token(token) => {
                token.source_position()
            },
            _ => None
        }
    }
}
```

MichaReiser: Implementing this on `FormatElement` will have the downside that this method needs to collect the `FormatElement` which requires allocating a new `VecBuffer` after  #2634 gets merged.

In my view, `FormatMember` knows what the length of the emitted string is and can return this if necessary (needs some refactoring because this isn't possible by just implementing `Format`, maybe add an additional method that returns the length as well). 

I'm also not sure if `token.source_position` is what they intend to get. Isn't it returning the position in the source document and not the length of the emitted string?

Dominionys: Could you please take a look. Is it correct to implement normalise length for `FormatMemberName`?

- https://github.com/rome/tools/pull/2627/files#diff-43f1671e76a32109081373edc030c09418af5d234ed82c52faf8b0139e27e15aR436-R452
- https://github.com/rome/tools/pull/2627/files#diff-13c8509b19044485e58df7b8a2ac3eef8012bf37f978217a526cb212fef1bf41R75-R87

I'm not sure how to avoid code duplication with:

- https://github.com/rome/tools/pull/2627/files#diff-13c8509b19044485e58df7b8a2ac3eef8012bf37f978217a526cb212fef1bf41R93-R108

crates/rome_js_formatter/src/utils/string_utils.rs
```diff
@@ -70,6 +71,20 @@ impl<'token> FormatLiteralStringToken<'token> {
     pub fn token(&self) -> &'token JsSyntaxToken {
         self.token
     }
+
+    pub fn normalise_len(&self, formatter: &JsFormatter) -> TextSize {
```
MichaReiser, line 75: This works but has the downside that we now have to do multiple passes over the string. 

Once when we do the actual formatting and once just to compute the length. 

My Recommendation would be to add an additional function to `FormatMember` that formats the string AND returns the length. 

The `FormatMember::format` function can then use that function and discard the length. All places that are interested in the length can use the new function and keep the length.

Dominionys: Oh, yes, I missed it.
What do you think about implementing function  for `FormatLiteralStringToken` with next signature? 
`pub fn format_token(
        &self,
        formatter: &JsFormatter,
    ) -> FormatResult<(FormatElement, Option<TextSize>)> `
  where `Option<TextSize>` is new text size if string's been normalized?

MichaReiser: That works. Would it come at an extra cost if the function also determines the length if the string already was normalized (in which case the string is normalized too)? I'm asking because I think a signature that returns `TextSize` rather than `Option<TextSize>` would be a bit easier to use.

MichaReiser (approved):

crates/rome_js_formatter/src/utils/string_utils.rs
Dominionys: I agree that `TextSize` is easier to use.
Hmm, there are 4 branches, and only 1 branch leads to the new length.
https://github.com/rome/tools/pull/2627/files#diff-43f1671e76a32109081373edc030c09418af5d234ed82c52faf8b0139e27e15aR441-R452
Is it possible to measure this cost?
Maybe if we need a new length in one more place, then refactor to `TextSize`?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
Dominionys: Added code snippets for an enum variant.

crates/rome_js_formatter/src/js/objects/property_object_member.rs
Dominionys: Added code snippets for an enum variant.

Dominionys:
There are some problems which we can fix in other PRs.

These problems are related with fluid layout:

- https://github.com/rome/tools/pull/2627/files#diff-4cea115d8cd21dc63350f69c5b7f843cd12490f3d0b3308e6d05608209a009e8R81-R96
- https://github.com/rome/tools/pull/2627/files#diff-ab0460fdc72367f12d0ea47dd51f37874303cf211d9bc2921374f25dca69b443R79-R85

This a problem with Unicode. [Rome AST playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#YwBvAG4AcwB0ACAAeAAgAD0AIAB7AAoAIAAgACAAIADkU8pOOgAKACAAIAAgACAAIgBoAHQAdABwAHMAOgAvAC8AcAByAGUAdAB0AGkAZQByAC4AaQBvAC8AZABvAGMAcwAvAGUAbgAvAHIAYQB0AGkAbwBuAGEAbABlAC4AaAB0AG0AbAAjAHcAaABhAHQALQBwAHIAZQB0AHQAaQBlAHIALQBpAHMALQBjAG8AbgBjAGUAcgBuAGUAZAAtAGEAYgBvAHUAdAAiACwACgAgACAA5FNTT9eLOgAKACAAIAAgACAAIgBoAHQAdABwAHMAOgAvAC8AcAByAGUAdAB0AGkAZQByAC4AaQBvAC8AZABvAGMAcwAvAGUAbgAvAHIAYQB0AGkAbwBuAGEAbABlAC4AaAB0AG0AbAAjAHcAaABhAHQALQBwAHIAZQB0AHQAaQBlAHIALQBpAHMALQBjAG8AbgBjAGUAcgBuAGUAZAAtAGEAYgBvAHUAdAAiACwACgB9ADsACgA=) shows that the text range for these symbols (古今) is 11-22. For the current algorithm it's a long key. But actually it should be a short key.
- https://github.com/rome/tools/pull/2627/files#diff-251f0a994b59d019f3d94030ef5a16118e03869014199fe8aa43112aa261971eR48-R52

MichaReiser:
> There are some problems which we can fix in other PRs.
> 
> ~These problems are related with fluid layout:~
> 
> ~- https://github.com/rome/tools/pull/2627/files#diff-4cea115d8cd21dc63350f69c5b7f843cd12490f3d0b3308e6d05608209a009e8R81-R96~ ~- https://github.com/rome/tools/pull/2627/files#diff-ab0460fdc72367f12d0ea47dd51f37874303cf211d9bc2921374f25dca69b443R79-R85~
> 
> This a problem with Unicode. [Rome AST playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#YwBvAG4AcwB0ACAAeAAgAD0AIAB7AAoAIAAgACAAIADkU8pOOgAKACAAIAAgACAAIgBoAHQAdABwAHMAOgAvAC8AcAByAGUAdAB0AGkAZQByAC4AaQBvAC8AZABvAGMAcwAvAGUAbgAvAHIAYQB0AGkAbwBuAGEAbABlAC4AaAB0AG0AbAAjAHcAaABhAHQALQBwAHIAZQB0AHQAaQBlAHIALQBpAHMALQBjAG8AbgBjAGUAcgBuAGUAZAAtAGEAYgBvAHUAdAAiACwACgAgACAA5FNTT9eLOgAKACAAIAAgACAAIgBoAHQAdABwAHMAOgAvAC8AcAByAGUAdAB0AGkAZQByAC4AaQBvAC8AZABvAGMAcwAvAGUAbgAvAHIAYQB0AGkAbwBuAGEAbABlAC4AaAB0AG0AbAAjAHcAaABhAHQALQBwAHIAZQB0AHQAaQBlAHIALQBpAHMALQBjAG8AbgBjAGUAcgBuAGUAZAAtAGEAYgBvAHUAdAAiACwACgB9ADsACgA=) shows that the text range for these symbols (古今) is 11-22. For the current algorithm it's a long key. But actually it should be a short key.
> 
>     * https://github.com/rome/tools/pull/2627/files#diff-251f0a994b59d019f3d94030ef5a16118e03869014199fe8aa43112aa261971eR48-R52
> 
> 
> UPD: After #2645 problems with array and function call are solved. @yassere Thank you 🙏🙏

The text range computes the byte length of a text and not the character length. You probably want to use `text.chars().count()` to get the number of characters. 

If this doesn't work then we need to use something like the `unicode-width` which we already use in the cli to compute the width.. @leops is familiar with that crate. 

I don't mind if we tackle this in it's own PR, especially if `chars().count()`. doesn't work

Dominionys:
> I don't mind if we tackle this in it's own PR, especially if `chars().count()`. doesn't work


Yes, it works for current cases. 
Unfortunately if key is grapheme clusters `a̐éö̲` it doesn't work. Prettier [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBrO3MC8mwAOlJpgIaAIwHIIDAEgEMCBMwEpgOQCM3Pvf-AwUO7sANGQqcATAGY2XYYqXKxZAL4BuEKJAQADjACW6ZKEoAncxADuABQsI0yEJQA21ygE8nOgEbnKMFwYAGVKAFs4ABlDKDhkADM3NDg-AKC4UL1A2IBzZBhzAFdUkDhw3zgAEyrqqMooXKLKXLgAMQhzcMoYI0bnSiKYCG0QAAsYcNcAdTHDeDRssDgQx3nDADd5z2cwNB8QWJTzGFsA3O7E5NKAKzQADxC81zgARSKIeCvXFJ1s82Ozl8lEqrlGenMsRg00MVRgY2QABYAJx-KwpaYBPTOCFwY4beI6ACOH3gZ30ThQuPxhJA5jgJMM9LOLUuSCSP1KKXChgKxS5zzepPi7OuOhgIJhcIRSCk4oChlceQAwhBwmyymgAKyjIopAAqIMpHN+IA2JQAklBarAQmBIQYAILWkIwTwvb4pNRqIA) handles this.
I tried `unicode-width` but for some string it has strange length. [unicode-segmentation](https://docs.rs/unicode-segmentation/latest/unicode_segmentation/trait.UnicodeSegmentation.html#tymethod.graphemes) works better.

Снимок экрана 2022-06-04 в 22 38 59

Dominionys:
> I don't mind if we tackle this in it's own PR, especially if `chars().count()`. doesn't work

Thank you!!! `unicode-width` works perfect 🥳

ematipico:
Indeed, chars count won't work in our case, I've been burnt already by that.

Dominionys:
I hope that these changes are ok and we can merge. 🙏

- https://github.com/rome/tools/pull/2627/files#diff-13c8509b19044485e58df7b8a2ac3eef8012bf37f978217a526cb212fef1bf41R75-R106

I'd like to continue investigating in other PRs the following cases (they have difference with prettier print):

- https://github.com/rome/tools/pull/2627/files#diff-bda4cd52acd56ec0915aefe83623ca9dfedcbb06f79eb42b00935a3a9f1e6d62R35-R40

- https://github.com/rome/tools/pull/2627/files#diff-f7ba7d7d511fda68868ddfec8551d3af86f686679c04882a945cc56fcd84587eR194-R204

- Prettier also has a difference in IR for call expression. [Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAbOMAEBDTBeTYAHSk0wCcEATOcpTMbVVACgEYAaTgJgEoOSZSlBqUq9Rsxa8SAXwDcIDiAgAHGAEtoAZ2Shs5chADuABQMJdKJsewBPXcoBG5bGADWGAMrYAtnAAZDSg4ZAAzJm04Z1cPb1U3YIBzZBhyAFdokDhfJzgqGioA7Cgk9OwkuAAxCHJfbBhNUuQQbHSYCCUQAAsYX1QAdW6NeG0EsDgvSxGNADcRuxawbUcQYKjyGFNXJPrwyKyAK20ADy9k9ABFdIh4fdQo5QTyDZanbDzULtVyYJgBjRUGDdZAADgADE8jFEBq5VC0fnANrNQsoAI43eDbNRWVraAC0IXy+S6lAxGko2wqeyQEQeWSivg0qQyDIucGut1CtIOyhgHwBQJBSG4fNcGlQyQAwhBfDTstoAKxddJRAAqHysdMeIFmmQAkiIEDAvGBfuoAIIiLwwOzoe5RWSyIA). I guess it's because of `isPoorlyBreakableMemberOrCallChain`. https://github.com/prettier/prettier/blob/feb6b519ee0e23eeb272227825ad84ecf1af05ab/src/language-js/print/assignment.js#L329-L370

ematipico:
!bench_formatter

MichaReiser:
Excellent, thank you so much for this contribution.

PR #2628, opened 2022-05-30
feature(rome_js_parser): Extraction of variable declarations

xunilrj:
This PR is the part that extracts variable declarations of the main PR (https://github.com/rome/tools/pull/2489).

This PR is the iterator that will extract as many events as possible with just one pass. At the moment I only extract "Declaration" from "JS_IDENTIFIER_BINDING". Will create other PRs until we reach the same 90% of the match of the original PR: https://github.com/rome/tools/pull/2489.

Future PRs will extract references, scopes, hoisting etc...

This PR also introduces the assertions to test the extraction. So we can write tests like 

```js
let a/*#B*/ = 1;
```

And we test if the token "a" was indeed extracted as a "Declaration".

#### Regression

Symbols regression is because I was signalling as ```ok``` tests where we could not find the original source code. Now I recover the source code from the .symbols file itself.


crates/rome_js_parser/src/semantic/events.rs
```diff
@@ -0,0 +1,64 @@
...
+        &code[start as usize..end as usize]
+    }
+}
+
+struct SemanticEventIterator {
+    iter: Preorder<JsLanguage>,
+}
+
+impl SemanticEventIterator {
+    fn extract_event(&self, node: &JsSyntaxNode) -> Option<SemanticEvent> {
```
leops, line 29: Would it be possible to make this method public, or at least expose a push-based model along with the pull-based Iterator for the semantic analysis ? This would be useful in the analyzer since it's already using a `Preorder` iterator to run the syntax-based rules, it could feed the resulting `WalkEvent` into this struct at the same time to merge the syntax analysis pass and semantic model pass in a single iteration over the entire tree

crates/rome_js_parser/src/lib.rs
```diff
@@ -69,7 +69,7 @@ mod state;
 #[cfg(test)]
 mod tests;
 
-pub mod symbols;
+pub mod semantic;
```
leops, line 72: This is something I missed in the previous PR since it was just a placeholder until now, but I'm not sure about having this logic in the `rome_js_parser` crate: the parser is a heavy dependency and I would rather not have it in the requirement chains for all crates that will depend on semantic analysis downstream. Would it be possible to move it to a new `rome_js_semantic` crate ?

xunilrj: Yes, it would. Will do it here.

xunilrj: Done.

crates/rome_js_semantic/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+edition = "2021"
+name = "rome_js_semantic"
+version = "0.3.0"
+authors = ["Rome Tools"]
+license = "MIT"
+description = "An extremely fast ECMAScript parser made for the rslint project"
```
ematipico, line 7:
```suggestion
version = "0.0.0"
authors = ["Rome Tools"]
license = "MIT"
description = "Semantic model for the JavaScript language"
```

xunilrj: Done.

leops (approved):

crates/rome_js_semantic/src/events.rs
```diff
@@ -0,0 +1,73 @@
...
+    stash: VecDeque<SemanticEvent>,
+}
+
+impl SemanticEventExtractor {
+    pub fn extract_from(&mut self, node: &JsSyntaxNode) {
+        use rome_js_syntax::JsSyntaxKind::*;
+        use SemanticEvent::*;
+        if let JS_IDENTIFIER_BINDING = node.kind() {
+            self.stash.push_back(DeclarationFound {
+                range: node.text_trimmed_range(),
```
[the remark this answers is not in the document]
xunilrj: I am back and forth on this one. Probably makes sense to use the cheaper one. I will test it.

PR #2629, opened 2022-05-30
feat(rome_formatter): make formatter language aware

### LINKED ISSUE #2546: Make the formatter language aware

ematipico:
## Summary

Closes #2546 

In order to make the formatter language aware, we need to store information that belong to the current file. In order to do so - after a brief [discussion we had on discord](https://discord.com/channels/678763474494423051/978663597053984808) - we decided to rename `Options` into `Context`, which will be the place where to store different information inside the current instance of the formatter.

Everything around options has been renamed to context, and inside this context we have also options that belong to the language. For example, `quote_style` has been moved inside `JsFormatContext::options` because it's an information that belong to JS only, while `line_width` belongs to context because it's an information that belong to all languages.

Work done:
- changed the codegen with the new code;
- updated all the code inside `rome_js_formatter/src/*` to use the new `JsFormatContext`;
- updated all the tests to use the new `JsFormatContext`;
- updated `rome_services` to consume the new `JsFormatContext`;
- modified the code inside `rome_js_formatter/src/utils/mod.rs` to use the new method to retrieve the `quote_style`;
- updated examples and doc tests;
- updated the code in the playground;
- updated the code in the xtask benchmarking; 

## Test Plan

All existing tests should pass

### DIFF
--- a/crates/rome_js_formatter/src/context.rs
+++ b/crates/rome_js_formatter/src/context.rs
@@ -1,21 +1,52 @@
 use rome_formatter::printer::PrinterOptions;
-use rome_formatter::{FormatOptions, IndentStyle, LineWidth};
+use rome_formatter::{FormatContext, IndentStyle, LineWidth};
+use rome_js_syntax::SourceType;
 use std::fmt;
+use std::fmt::Debug;
 use std::str::FromStr;
 
 #[derive(Debug, Copy, Clone, Default)]
-pub struct JsFormatOptions {
+pub struct JsFormatContext {
     /// The indent style.
     pub indent_style: IndentStyle,
 
     /// What's the max width of a line. Defaults to 80.
     pub line_width: LineWidth,
 
+    /// Options of current instance of the formatter
+    pub options: JsFormatOptions,
+
+    /// Information relative to the current file
+    pub source_type: SourceType,
+}
+
+impl JsFormatContext {
+    pub fn new(
+        indent_style: IndentStyle,
+        line_width: LineWidth,
+        options: JsFormatOptions,
+        source_type: SourceType,
+    ) -> Self {
+        Self {
+            indent_style,
+            line_width,
+            options,
+            source_type,
+        }
+    }
+
+    pub fn quote_style(&self) -> QuoteStyle {
+        self.options.quote_style
+    }
+}
+
+#[derive(Debug, Copy, Clone, Default)]
+pub struct JsFormatOptions {
     // The style for quotes. Defaults to double.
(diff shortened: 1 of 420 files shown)


crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -22,11 +22,12 @@ pub struct JsFormatSettings {
     pub indent_style: Option<IndentStyle>,
     pub line_width: Option<LineWidth>,
     pub quote_style: Option<QuoteStyle>,
+    pub source_type: Option<SourceType>,
```
leops, line 25: It doesn't really make sense to include the `SourceType` in the settings as that's a per-document value, while the `Settings` object is shared between all documents of a certain language (`JsLanguage` here) and is intended to be eventually deserialized straight from the CLI arguments or a `rome.json` configuration file

The `JsFormatContext` struct should instead be constructed in the formatting handlers (the `format`, `format_range` and `format_on_type` functions) from the formatting options derived from the settings, and a `SourceType` derived from the instance of `RomePath` provided in the arguments

ematipico: Please check if the last commit does what you requested

crates/rome_js_formatter/src/context.rs
```diff
@@ -1,21 +1,52 @@
...
+        options: JsFormatOptions,
+        source_type: SourceType,
+    ) -> Self {
+        Self {
+            indent_style,
+            line_width,
+            options,
+            source_type,
+        }
+    }
```
MichaReiser, line 36: This is a rather lengthy constructor. I think using an initializer `JsFormatContext { ... }` or using `JsFormatContext::default().with_indent_style(..)` might be easier to read

crates/rome_js_formatter/src/formatter.rs
```diff
@@ -138,7 +136,7 @@ impl Format for FormatSuppressedNode<'_> {
 
 fn format_verbatim_node_or_token(
     node: &JsSyntaxNode,
-    formatter: &Formatter<JsFormatOptions>,
+    formatter: &Formatter<JsFormatContext>,
```
MichaReiser, line 139: Nit: It could be helpful to have a `JsFormatter` type alias in the root similar to `JsSyntaxNode`. It spares us typing the `<...>` in all the Js code

MichaReiser (approved):
Looks good. Please make sure that you update the formatter codegen too and test it by deleting an `Unknown`, `List`, `Any`, and specific node implementation and re-generating the files

PR #2638, opened 2022-06-01
perf(formatter): Remove `prev_token` traversal from format leading trivia

MichaReiser:
…ivia

The trailing trivia of the previous token can never contain a new line. Any new line is always part of the current tokens leading trivia, thus, this check is not necessary

## Test Plan

`cargo test`


MichaReiser:
!bench_formatter

PR #2639, opened 2022-06-01
feature(rome_js_semantic): Extraction of scope (simple cases)

xunilrj:
This PR is the part that extracts scopes of the main PR (https://github.com/rome/tools/pull/2489).

This PR basically extract ```ScopeStarted``` and ```ScopeEnded``` from blocks and function bodies. There are other more complex scenarios that will be dealt later.

Also this PR introduces the testing and assertion of scopes, like: 

```rust
assert("if (true) {/*START A*/ }/*END A*/");
```


crates/rome_js_semantic/src/events.rs
```diff
@@ -15,12 +15,29 @@ pub enum SemanticEvent {
     /// - Import bindings
     /// - Functions parameters
     DeclarationFound { range: TextRange },
+
+    /// Signifies that a new scope was started
+    /// Currently generated for:
+    /// - Blocks
+    /// - Function body
+    ScopeStarted { range: TextRange },
```
ematipico, line 23: Not sure it's in the scope of this PR, but we should consider having different kind of scopes too. Based on the scope kind, the binding of variables will differ.

Here's what Rome classic had:

```ts
export type ScopeKind = 
         | "root" 
         | "program" 
         | "function" 
         | "block" 
         | "loop" 
         | "conditional-type" 
         | "type-generic" 
         | "class";
```

xunilrj: > Not sure it's in the scope of this PR

That is what I intend to do for the next PR when I will match declarations and scopes. But thank you for the tip. Very nice!

PR #2640, opened 2022-06-02
refactor(formatter): Change binary like expression to format left to right

MichaReiser:
## Summary

Refactors the binary like expression formatting to lazily write the expressions rather than allocating vectors with the sub results. 

The basic idea remains the same. The implementation flattens an expression. What's different to before is that this is now a two-stage process:

1. Flatten the binary expression  and specify for each "part" how it should be formatted (left, right, or a group)
2. Format the flattened expressions 

I think this also improves readability because it makes it clear what concepts exist in this algorithm (left hand side, right hand side, and a sub-group). 

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_js_formatter/src/utils/binary_like_expression.rs
+++ b/crates/rome_js_formatter/src/utils/binary_like_expression.rs
@@ -119,7 +119,7 @@ pub(crate) fn format_binary_like_expression(
             let has_comments = left.syntax().has_comments_direct();
 
             flatten_items.items.push(FlattenItem::new(
-                FlattenedExpression::Left { expression: left },
+                FlattenedBinaryExpressionPart::Left { expression: left },
                 Some(parent_operator),
                 has_comments.into(),
             ));
@@ -133,7 +133,7 @@ pub(crate) fn format_binary_like_expression(
         flatten_items.flatten_binary_expression_right_hand_side(root, None)?;
     }
 
-    let group = FlattenedExpression::Group {
+    let group = FlattenedBinaryExpressionPart::Group {
         current: JsAnyBinaryLikeLeftExpression::JsAnyExpression(current_node.into_expression()),
         expressions_start: flatten_items.current_group_start,
         expressions_end: flatten_items.len(),
@@ -252,8 +252,7 @@ where
     })
 }
 
-/// It tells if the expression can be hard grouped
-fn can_hard_group(flatten_nodes: &[FlattenItem]) -> bool {
+fn keep_on_same_line(flatten_nodes: &[FlattenItem]) -> bool {
     // We don't want to have 1 + 2 to break, for example.
     // But if there are any trailing comments, break it.
     flatten_nodes.len() <= 2 && flatten_nodes.iter().all(|node| !node.has_comments())
@@ -348,7 +347,7 @@ impl FlattenItems {
         let has_comments = right.syntax().has_comments_direct();
 
         let flatten_item = FlattenItem::new(
-            FlattenedExpression::Right {
+            FlattenedBinaryExpressionPart::Right {
                 parent: binary_like_expression,
             },
             parent_operator,
@@ -381,7 +380,7 @@ impl FlattenItems {
         let operator_has_trailing_comments = operator_token.has_trailing_comments();
         let left_parenthesized = needs_parens(operator, &left)?;
         let mut left_item = FlattenItem::new(
-            FlattenedExpression::Group {
+            FlattenedBinaryExpressionPart::Group {
                 current: left,
                 expressions_start: self.current_group_start,
                 expressions_end: self.items.len(),
@@ -406,7 +405,7 @@ impl FlattenItems {
             .map(|operator| operator.has_leading_comments());
 
         let mut right_item = FlattenItem::new(
-            FlattenedExpression::Right {
+            FlattenedBinaryExpressionPart::Right {
                 parent: binary_like_expression,
             },
             parent_operator,
@@ -460,8 +459,8 @@ impl From<bool> for Comments {
 
 /// The left or right sub part of a binary expression.
 #[derive(Debug)]
-enum FlattenedExpression {
-    /// The right hand sie of a binary expression. Needs to format the parent operator and the right expression
+enum FlattenedBinaryExpressionPart {
+    /// The right hand side of a binary expression. Needs to format the parent operator and the right expression
     Right {
         /// The parent expression
         parent: JsAnyBinaryLikeExpression,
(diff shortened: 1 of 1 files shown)


MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
     }
 }
 
+/// The left or right sub part of a binary expression.
+#[derive(Debug)]
+enum FlattenedExpression {
+    /// The right hand sie of a binary expression. Needs to format the parent operator and the right expression
```
ematipico, line 464:
```suggestion
    /// The right hand side of a binary expression. Needs to format the parent operator and the right expression
```

ematipico: Isn't `"right expression"` self in this case? If not, I am really missing the point of this struct

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
     }
 }
 
+/// The left or right sub part of a binary expression.
+#[derive(Debug)]
+enum FlattenedExpression {
```
ematipico, line 463: `FlattenedBinaryExpression` ? If my suggestion is wrong, then I am really missing the point of this struct

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
...
+        /// Whether to parenthesize the expression
+        parenthesized: bool,
+    },
+}
+
+impl FlattenedExpression {
+    fn write(&self, f: &mut JsFormatter, items: &[FlattenItem]) -> FormatResult<()> {
+        match self {
+            FlattenedExpression::Right { parent } => {
+                let right = JsAnyBinaryLikeLeftExpression::JsAnyExpression(parent.right()?);
```
ematipico, line 494: I am not really sure what are we doing here... why do we save to parent and get the right part? Couldn't we save the "right" part straightaway?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
...
+            }
+            FlattenedExpression::Group {
+                current,
+                expressions_start,
+                expressions_end,
+                parenthesized,
+            } => {
+                let expressions = &items[*expressions_start..*expressions_end];
+                let content = format_with(|f| {
+                    let can_hard_group = can_hard_group(expressions);
```
ematipico, line 510: We don't have the concept of "hard group" anymore. Maybe we should change the name of the function and the variable

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
...
+                            [group_elements(&format_once(|f| {
+                                f.join_with(soft_line_break_or_space())
+                                    .entries(groups)
+                                    .finish()
+                            }))]
+                        )
+                    } else if should_indent_if_parent_inlines(current) {
+                        // in order to correctly break, we need to check if the parent created a group
+                        // that breaks or not. In order to do that , we need to create two conditional groups
+                        // that behave differently depending on the situation
```
ematipico, line 550: I think the comment is outdated. Can we update it?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
...
+                                |f| {
+                                    f.join_with(soft_line_break_or_space())
+                                        .entries(groups)
+                                        .finish()
+                                }
+                            )))]
+                        )
+                    } else {
+                        // if none of the previous conditions is met,
+                        // we take take out the first element from the rest of group, then we hard group the "head"
```
ematipico, line 563:
```suggestion
                        // we take out the first element from the rest of group, then we hard group the "head"
```

ematipico: Could we remove the phrase around the hard group?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -529,10 +458,138 @@ impl From<bool> for Comments {
...
+                        // and we indent the rest of the groups in a new line
+                        write!(f, [groups.next().unwrap()])?;
```
ematipico, line 565: Could we add a `// SAFETY` comment?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -209,13 +209,37 @@ fn format_with_or_without_parenthesis(
...
+
+fn format_parenthesized<'a, Inner>(inner: Inner) -> impl Format<JsFormatContext>
+where
+    Inner: Format<JsFormatContext> + 'a,
+{
+    format_with(move |f| {
         let mut buffer = VecBuffer::new(f.state_mut());
+        write!(buffer, [inner])?;
+        let formatted_node = buffer.into_element();
+        let (leading, content, trailing) = formatted_node.split_trivia();
```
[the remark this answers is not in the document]
MichaReiser: We need to intercept the output to be able to get the formatted inner and call `split_trivia` on the format element.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: It's a part of a binary expression but not the full flattened binary expression. It's either the left hand side of the binary expression, the right hand side or a sub group. This is the same as existed before with the kind.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: We also need the operator. An option could be to save the right + operator but the way I remember it is that you then run into lifetime issues. So that was the easiest.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: If you can tell me what it does xD I just ported the code as is. I'm not really sure what the criteria or conceptual idea behind it is

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: Hmm maybe... That's copied 1:1, let me see if I can figure it out

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -209,13 +209,37 @@ fn format_with_or_without_parenthesis(
...
+            write!(f, [sub_expression])
+        }
+    })
+}
+
+fn format_parenthesized<'a, Inner>(inner: Inner) -> impl Format<JsFormatContext>
+where
+    Inner: Format<JsFormatContext> + 'a,
+{
+    format_with(move |f| {
```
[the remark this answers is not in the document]
MichaReiser: This isn't formatter specific. https://doc.rust-lang.org/std/keyword.move.html

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: The point of this struct is that it represents this

https://github.com/rome/tools/blob/6e8a3cb5899f2017332d4e58bfaff63cf9c99bc6/crates/rome_js_formatter/src/utils/binary_like_expression.rs#L50-L53

crates/rome_js_formatter/src/utils/binary_like_expression.rs
ematipico: We can change `can_hard_group` with `can_inline`? Basically we are telling the algorithm that we should not add groups - the expression should be formatted in one line.

crates/rome_js_formatter/src/utils/binary_like_expression.rs
ematipico: I missed this change unfortunately. But here we can safely tell that we always have at least two groups in a binary like expression.

PR #2641, opened 2022-06-02
refactor(formatter): Member chain: left to right formatting

MichaReiser:
This PR refactors the member chain formatting into two stages:

* Building the member chain groups
* Formatting the groups

Doing so allows us to format the elements in left-to-right order.

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
+++ b/crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
@@ -1,8 +1,9 @@
 use crate::prelude::*;
 use rome_formatter::write;
 use rome_js_syntax::{
-    JsAnyExpression, JsCallExpression, JsComputedMemberExpression, JsIdentifierExpression,
-    JsImportCallExpression, JsNewExpression, JsStaticMemberExpression, JsSyntaxNode,
+    JsAnyExpression, JsCallExpression, JsCallExpressionFields, JsComputedMemberExpression,
+    JsComputedMemberExpressionFields, JsIdentifierExpression, JsImportCallExpression,
+    JsNewExpression, JsStaticMemberExpression, JsStaticMemberExpressionFields, JsSyntaxNode,
     JsThisExpression,
 };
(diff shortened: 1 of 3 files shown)


MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
```diff
@@ -138,32 +139,42 @@ impl FlattenItem {
...
-            FlattenItem::ComputedExpression(_, formatted) => {
-                std::write!(f, "ComputedExpression: {:?}", formatted)
+            FlattenItem::ComputedMember(computed_member) => {
+                write!(
+                    f,
+                    [
+                        computed_member.optional_chain_token().format(),
+                        computed_member.l_brack_token().format(),
+                        computed_member.member().format(),
+                        computed_member.r_brack_token().format(),
```
ematipico, line 171: nit: I think we should use the `*Fields` structs. That's the style we are using throughout the formatters

crates/rome_js_formatter/src/utils/member_chain/groups.rs
```diff
@@ -93,43 +92,25 @@ impl Groups {
...
-
-        elements.into_format_element()
+    pub fn write_joined_with_hard_line_breaks(&self, f: &mut JsFormatter) -> FormatResult<()> {
+        f.join_with(hard_line_break())
+            .entries(
+                self.groups
+                    .iter()
+                    .map(|group| format_with(|f| f.join().entries(group.iter()).finish())),
+            )
+            .finish()
```
ematipico, line 103: Any particular reason why we don't consume `self` anymore?

ematipico (approved):

crates/rome_js_formatter/src/utils/member_chain/groups.rs
MichaReiser: There's no reason to consume `self`. This function works fine with a reference of self.

Issue #2642, opened 2022-06-02
☂️ First batch of Linter Rules


leops:
# Description

This umbrella task aims to track the list of lint rules we need to implement in order to get the linter ready for release (aka. provide helpful hints and fixes in real world code)

This list was assembled by collecting lint rules from various linter projects (the Rome JS linter, ESLint and popular plugins, and some concepts from the Clippy linter for Rust that also apply to JS), but more specifically I prioritized rules that are part of the recommended configurations and can be automatically fixed (some rules with no auto fixes but that still provide useful errors are included at the end). I also purposefully excluded most styling-related rules, since we have a formatter for that, as well as most rules that rely on type information (mainly in the ESLint TypeScript rules) as type checking is not part of the plan for the initial release of the analyzer.

Note that while this is an umbrella issue aimed at tracking the implementation work, the list is also open to discussion: I may have accidentally missed some rules (also as of 02 Jun. 2022 I haven't gone over and included rules from the ESLint React plugin yet), and we may also decide that some of these rules aren't required for the initial release and can be de-prioritized.

# If you want to contribute
1. make sure you understand the requirements of the rule you would like to implement;
3. comment on this very issue telling saying which rule you would like to implement;
4. create a new task issue and link to this very umbrella issue (a member of the team will eventually assign it to you);
5. open a PR;

> **Note**: please make sure to comment the issue saying which rule you'd like to implement, this would allow better coordination between contributors. Failing to do so would penalize yourselves against contributors that had followed these simple guildelines

> **Note**: follow the naming convention guidelines: https://github.com/rome/tools/blob/archived-js/CONTRIBUTING.md#naming-patterns

# Fixable Rules

## Rome JS

- [x] #2647
Disallow comparing against `-0`
Example fix:
`1 >= -0` -> `1 >= 0`
- [x] #2651 
Disallow the use of `debugger`
- [x] #2660 
Disallow unnecessary boolean casts
Example fix:
`if (Boolean(foo)) {}` -> `if (foo) {}`
- [x] #2656
Disallow negation in the condition of an if statement if it has an else clause
Example fix:
`if (!true) {consequent;} else {alternate;}` -> `if (true) {alternate;} else {consequent;}`
- [ ] #2754
Disallow the use of single character alternations in regular expressions
Example fix:
`/a|b/` -> `/[ab]/`
- [x] #2652 
Disallow sparse arrays
Example fix:
`[1,,2]` -> `[1,undefined,2]`
- [x] #2746
[...]

PR #2643, opened 2022-06-02
feat(rome_analyze): noDebugger 

### LINKED ISSUE #2651: 📎 noDebugger
### Description

Finish `rome_analyze` rule **noDebugger**

IWANABETHATGUY:
## Summary
1. Add new analyze rule `noDebugger` , part of https://github.com/rome/tools/issues/2642
2. Resolved #2651 

## Test Plan
1. should pass `noDebugger.js`
3. the test case is stole from `ESlint` , https://github.com/eslint/eslint/blob/main/tests/lib/rules/no-debugger.js


NicholasLYang:
We should probably have a guide on proper error message style and conventions. For now, how about we imitate ESLint?
```
Unexpected 'debugger' statement.
```

IWANABETHATGUY:
I copy from here, https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noDebugger.md
```
This is an unexpected use of the debugger statement.
```

IWANABETHATGUY:
> Thanks for this contribution. For the Rome linter we aim to have as many rules as possible coming with automated fixes, Could you add a code action to the rule along with the diagnostic ? Unfortunately the analyzer is still in a very early stage and it's missing many helpers around mutating syntax trees, in this case the fix means removing the debugger statement from it's parent node but I think that might need to be handled through a new helper function to either splice it out from the parent statement list or replace it with an empty statement in the `if(cond) debugger` case for instance

I just found ESlint doesn't have a code action, so I keep the code action as is, which a None. If we want to port other rule, how could I know if it has a code action or not?

leops:
> I just found ESlint doesn't have a code action, so I keep the code action as is, which a None. If we want to port other rule, how could I know if it has a code action or not?

In #2642 I tried to reference rules that have an auto-fix available when possible, so for instance the source for the `noDebugger` rule there is Rome JS since it has a fixable diagnostic

IWANABETHATGUY:
> Thanks for this contribution. For the Rome linter we aim to have as many rules as possible coming with automated fixes, Could you add a code action to the rule along with the diagnostic ? Unfortunately the analyzer is still in a very early stage and it's missing many helpers around mutating syntax trees, in this case the fix means removing the debugger statement from it's parent node but I think that might need to be handled through a new helper function to either splice it out from the parent statement list or replace it with an empty statement in the `if(cond) debugger` case for instance

It seems `root.replace_node` can't replace a `DebuggerStatement`  with a `EmptyStatement` due to they have different type.
```rs
pub fn replace_node<N>(self, prev_node: N, next_node: N) -> Option<Self>
where
    N: AstNode<Language = Self::Language>,
    Self: Sized,
```

leops:
> It seems `root.replace_node` can't replace a `DebuggerStatement` with a `EmptyStatement` due to they have different type.

You can use the `JsAnyStatement` union to replace statements with a different type: `.replace_node(JsAnyStatement::JsDebuggerStatement(prev), JsAnyStatement::JsEmptyStatement(next))`

crates/rome_analyze/src/analyzers/no_debugger.rs
```diff
@@ -0,0 +1,50 @@
...
+            .to_owned(),
+        })
+    }
+
+    fn action(
+        root: rome_js_syntax::JsAnyRoot,
+        node: &Self::Query,
+        _state: &Self::State,
+    ) -> Option<crate::registry::RuleAction> {
+        let root = root.replace_node(
```
leops, line 39: If the parent node for the debugger statement is a list (`JsStatementList` or `JsModuleItemList`) it's better to simply remove the node

crates/rome_analyze/src/analyzers/no_debugger.rs
```diff
@@ -0,0 +1,50 @@
...
+            JsAnyStatement::JsDebuggerStatement(node.clone()),
+            JsAnyStatement::JsEmptyStatement(make::js_empty_statement(make::token(T![;]))),
+        )?;
+        Some(RuleAction {
+            category: ActionCategory::QuickFix,
+            applicability: Applicability::MaybeIncorrect,
+            message: markup! { "Replace with emptyStatement" }.to_owned(),
```
leops, line 46: "emptyStatement" is an internal term users of the linter may not be familiar with, using "Remove debugger statement" as the action message for the quick fix might be easier to understand

crates/rome_analyze/tests/specs/noDebugger.js
```diff
@@ -0,0 +1,3 @@
+const test = { debugger: 1 }; test.debugger;
+
+if (foo) debugger
```
leops, line 3: Can we add additional tests for statement list and module item list ?

crates/rome_analyze/src/analyzers/no_debugger.rs
IWANABETHATGUY: I don't find an api to do that

leops: Right there isn't (yet ?) a method to remove a node from a list, instead it's done by using splice to replace a range with an empty iterator:
```rust
let prev_parent = node.syntax().parent()?;

if JsStatementList::can_cast(prev_parent.kind()) || JsModuleItemList::can_cast(prev_parent.kind()) {
    let index = prev_parent
        .children()
        .position(|slot| &slot == node.syntax())?;

    let next_parent = prev_parent.clone().splice_slots(
        index..=index,
        std::iter::empty(),
    );

    root = JsAnyRoot::unwrap_cast(
        root.into_syntax()
            .replace_child(prev_parent.into(), next_parent.into())?,
    );
}
```

leops (approved):

NicholasLYang:
I'm not really clear on why the error message is better. From my perspective it's a little more verbose, but honestly a minor difference. I can go either way.

A bigger issue I have is that there's no explanation or context around why this rule exists. With linting, I think it should be pretty important to have an explanation of why we are not allowing something.

leops:
> > A bigger issue I have is that there's no explanation or context around why this rule exists. With linting, I think it should be pretty important to have an explanation of why we are not allowing something.
> 
> This is a good point! We should always try to explain why we deny something. Care to venture a suggestion?

This is probably not the best place to discuss this, but besides improving to `RuleDiagnostic` to bring it inline with `rome_diagnostic` (and probably merge all diagnostics into a single unified system eventually) which will allow for richer diagnostics with more and better information, for the release of the analyzer rules will need to come with their documentation. Initially I think this will simply be in the form of documentation pages for each rule on the website, with a `Hyperlink` markup in the diagnostic header to make the code of the lint rule clickable (in terminal clients that support it), but we could also introduce a `rome explain <lint rule>` (similar to the `rustc --explain` command) to let users read the documentation for a rule directly through the command line (in offline usage scenarios for instance)

IWANABETHATGUY:
This pull request is ready to merge @ematipico

crates/rome_analyze/tests/specs/noDebugger.js.snap
```diff
@@ -0,0 +1,82 @@
...
+  │   --------- This is an unexpected use of the debugger statement.
+
+Remove debugger statement
+    | @@ -5,6 +5,5 @@
+4 4 |   debugger;
+5 5 |   
+6 6 |   function test() {
+7   | -   let a = 3;
+8   | -   debugger;
+  7 | +   let a = 3;;
```
ematipico, line 76: There's an additional `;` that should not be there

ematipico (changes requested):

crates/rome_analyze/tests/specs/noDebugger.js.snap
```diff
@@ -0,0 +1,80 @@
...
+  │
+3 │ if (foo) debugger
+  │          -------- This is an unexpected use of the debugger statement.
+
+Remove debugger statement
+    | @@ -1,6 +1,6 @@
+0 0 |   const test = { debugger: 1 }; test.debugger;
+1 1 |   
+2   | - if (foo) debugger
+  2 | + if (foo) ;
```
ematipico, line 33: Is this correct? We're adding a `;`

leops: Yes this change here is correct, the debugger statement is the `consequent` of the parent if statement so it can't be removed (the consequent is required, so it would create an invalid tree with missing children errors), instead it has to be replaced with an empty statement
The same problem exists with other similar statement like for, while, do-while, ... Alternatively we could consider replacing the removed statement with an empty block, or removing the entire if statement (but it may be unsafe if the test condition has side effects)

ematipico: > The same problem exists with other similar statement like for, while, do-while, ... Alternatively we could consider replacing the removed statement with an empty block, or removing the entire if statement (but it may be unsafe if the test condition has side effects)

Thank you for the explanation,  I think it makes sense. I am just curious to know how this will play out with different rules. You have to consider that we will have a rule that will be called `lint/js/useBlockStatements`, which will add blocks instead of empty statements. 

I wonder now if we should prepare a test suite to actually run multiple rules on the same snippet and see how it will be the end result? 

Screenshot 2022-06-13 at 09 12 03

ematipico (approved):

crates/rome_analyze/src/analyzers/no_debugger.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    fn diagnostic(node: &Self::Query, _state: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic {
+            severity: Severity::Warning,
+            range: node.syntax().text_trimmed_range(),
+            message: markup! {
+                "This is an unexpected use of the "<Emphasis>"debugger"</Emphasis>" statement."
+            }
+            .to_owned(),
+        })
```
leops, line 33:
```suggestion
        Some(RuleDiagnostic::warning(
            node.syntax().text_trimmed_range(),
            markup! {
                "This is an unexpected use of the "<Emphasis>"debugger"</Emphasis>" statement."
            },
        ))
```

leops (changes requested):

leops (approved):

PR #2645, opened 2022-06-02
fix(rome_formatter): handle groups correctly when measuring fit

yassere:
## Summary

This seems to bring the printer closer to Prettier’s by changing the `fits_on_line` measurement.

Previously, the `fits_element_on_line` function was measuring all groups with `PrintMode::Flat`, including those from the `rest_queue`. This caused elements to be falsely measured as `Fits::No` even if they were followed by a group that could have broken before the end of the desired line width.

This change causes the test for the `best_fitting` macro to fail, but we decided to just ignore it for now to merge this sooner because the `BestFitting` IR isn't currently used anywhere and there's more than one option for how we can fix it.

## Test Plan

Reviewed the snapshot changes. There are no major blocking regressions, and we decided we should merge this ASAP and then follow up later with additional PRs to match Prettier's behavior.

### DIFF
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -726,10 +726,7 @@ fn fits_element_on_line<'a, 'rest>(
             args.with_incremented_indent(),
         )),
 
-        FormatElement::Group(group) => queue.enqueue(PrintElementCall::new(
-            &group.content,
-            args.with_print_mode(PrintMode::Flat),
-        )),
+        FormatElement::Group(group) => queue.enqueue(PrintElementCall::new(&group.content, args)),
 
         FormatElement::ConditionalGroupContent(conditional) => {
             if args.mode == conditional.mode {
--- a/crates/rome_js_formatter/tests/specs/js/module/object/computed_member.js.snap
+++ b/crates/rome_js_formatter/tests/specs/js/module/object/computed_member.js.snap
@@ -1,5 +1,6 @@
 ---
 source: crates/rome_js_formatter/tests/spec_test.rs
+assertion_line: 257
 expression: computed_member.js
 ---
 # Input
@@ -40,29 +41,11 @@ foo.bar["bar"]["lorem_ispsum"].foo["lorem-ipsum"] = true;
 a[b];
 c?.[d];
 
-x[
-	"very"
-][
-	"long"
-][
-	"chain"
-][
-	"of"
-][
-	"computed"
-][
-	"members"
-][
-	"that"
-][
-	"goes"
-][
-	"on"
-][
+x["very"]["long"]["chain"]["of"]["computed"]["members"]["that"]["goes"]["on"][
 	"for"
-][
-	"ever"
-]["I"]["mean"]["it"]["for"]["ever"]["and"]["even"]["longer"]["than"]["that"];
+]["ever"]["I"]["mean"]["it"]["for"]["ever"]["and"]["even"]["longer"]["than"][
+	"that"
+];
 
 x.very.long.chain.of.static.members.that.goes.on.for.ever.I.mean.it.for.ever[
 	"and"
--- a/crates/rome_js_formatter/tests/specs/prettier/js/async/nested.js.snap
+++ b/crates/rome_js_formatter/tests/specs/prettier/js/async/nested.js.snap
@@ -1,5 +1,6 @@
 ---
 source: crates/rome_js_formatter/tests/prettier_tests.rs
+assertion_line: 182
 expression: nested.js
 ---
 # Input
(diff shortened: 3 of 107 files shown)


crates/rome_formatter/src/printer/mod.rs
```diff
@@ -726,10 +726,7 @@ fn fits_element_on_line<'a, 'rest>(
             args.with_incremented_indent(),
         )),
 
-        FormatElement::Group(group) => queue.enqueue(PrintElementCall::new(
-            &group.content,
-            args.with_print_mode(PrintMode::Flat),
-        )),
+        FormatElement::Group(group) => queue.enqueue(PrintElementCall::new(&group.content, args)),
```
MichaReiser, line 729: After reviewing Prettier's code again I think this is correct. 

Prettier enqueues the rest elements like any other element when processing fits(the same as we do)

https://github.com/prettier/prettier/blob/83ec3c36e34864470a6cb88889d837c46777ab7d/src/document/doc-printer.js#L156-L164

It then enqueues the content with the current mode **except** if `doc.break` is true, a flag our printer currently doesn't support. 

https://github.com/prettier/prettier/blob/83ec3c36e34864470a6cb88889d837c46777ab7d/src/document/doc-printer.js#L196-L209

MichaReiser:
This is looking good. By how much is it improving the prettier compatibility? 

Regarding best fitting. I believe the problem now is that our `BestFitting` doesn't create implicit groups for each variant, thus... `fits_element_on_line` is called with the flat/expanded state of the parent, which, by default is expanded and the result is that inner group of the first variant is now printed in expanded mode too (which fits). See [this comment](https://github.com/rome/tools/pull/2591#discussion_r876986910). 

It may make sense to change this now to match Prettier's behaviour because I'm not aware how we could do this otherwise with the changed semantics except if we also track if a group is from the rest queue but I think this would over complicated things.

yassere:
> By how much is it improving the prettier compatibility?

**File Based Average Prettier Similarity**: 71.99%  -> 72.52%
**Line Based Average Prettier Similarity**: 65.23% -> 66.36%

Hopefully, it will improve even more by the time this merges. And this change should make it much easier to match Prettier’s line-breaking behavior in some cases.

MichaReiser:
!bench_formatter

crates/rome_js_formatter/tests/specs/js/module/object/computed_member.js.snap
```diff
@@ -40,29 +41,11 @@ foo.bar["bar"]["lorem_ispsum"].foo["lorem-ipsum"] = true;
...
-][
-	"members"
-][
-	"that"
-][
-	"goes"
-][
-	"on"
-][
+x["very"]["long"]["chain"]["of"]["computed"]["members"]["that"]["goes"]["on"][
```
ematipico, line 44: This was regression. I remembered I had fixed that. This is correct again.

crates/rome_js_formatter/tests/specs/prettier/js/async/nested.js.snap
```diff
@@ -19,9 +20,9 @@ const getAccountCount = async () =>
 const getAccountCount = async () =>
   (
     await (
-      await (
-        await focusOnSection(BOOKMARKED_PROJECTS_SECTION_NAME)
-      ).findItem("My bookmarks")
+      await (await focusOnSection(BOOKMARKED_PROJECTS_SECTION_NAME)).findItem(
```
ematipico, line 23: This is a regression to look after

MichaReiser: I would recommend merging the PR anyway and then follow up as part of call expression formatting (or whatever node is causing this) because this fix will make it easier for us to match prettier's behaviour. So merging it as soon as possible will improve our all experience writing formatter

ematipico: Yes. I was just helping Yasser to identify possible regressions to look after as he's less familiar with the formatter. I hope it helps.

MichaReiser: It certainly does!

yassere: @ematipico Thanks, I appreciate it!

@MichaReiser If we want to merge this ASAP, do you think I should just `ignore` the `BestFitting` test for now (since that IR isn't used anywhere yet and there's more than one option for how we can fix it) and then follow up later with additional PRs to match Prettier's behavior?

MichaReiser: Sounds good to me

MichaReiser (approved):

PR #2646, opened 2022-06-03
feat(rome_analyze): add a codegen command for the analyzer

leops:
## Summary

This PR adds a `cargo codegen analyzer` to automatically generate the `analyzers.rs` and `assists.rs` index modules in `rome_analyze` as well as the `with_filter` builder on `RuleRegistry` (through the `impl_registry_builders` macro), based on a list of analyzer rules generated automatically by listing all the files in the `analyzers` and `assists` directories


crates/rome_analyze/src/analyzers/no_var.rs
```diff
@@ -1,16 +0,0 @@
-use crate::{signals::DiagnosticExt, Analysis, Analyzer, AnalyzerContext};
```
xunilrj, line 1: Why you don't need this files anymore?

xtask/codegen/src/generate_analyzer.rs
```diff
@@ -0,0 +1,108 @@
+use std::fmt::Write;
```
ematipico, line 1: `generate_analyzers.rs`? (plural)

xtask/codegen/src/generate_analyzer.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use anyhow::{Context, Ok, Result};
+use quote::{format_ident, quote};
+use xtask::{glue::fs2, project_root};
+
+pub fn generate_analyzer() -> Result<()> {
```
ematipico, line 7:
```suggestion
pub fn generate_analyzers() -> Result<()> {
```

ematipico (changes requested):
We should update the `CONTRIBUTING.md` file with this new command.

xtask/codegen/src/generate_analyzer.rs
leops: The rationale was that this is generating the code for the analyzer (well technically the registry), but isn't touching the code for the analyzers / rules themselves

crates/rome_analyze/src/analyzers/no_var.rs
leops: The `no_var` and `use_single_case_statement` files were still written using the previous iteration of the analyzer API and were not being imported into the crate, I'm removing them for now because they would get automatically imported by the codegen script but I'll be adding them back in followup PRs shortly

crates/rome_analyze/src/assists.rs
```diff
@@ -1,3 +1,4 @@
-mod flip_bin_exp;
+//! Generated file, do not edit by hand, see `xtask/codegen`
```
MichaReiser, line 1: Can we name this file `generated` or put into a `generated` directory so that it automatically gets "hidden" in reviews. Or add it to the list in the `gitattributes` file

MichaReiser (approved):
How would you envision dynamic plugins/rules in the future?

xunilrj (approved):

PR #2648, opened 2022-06-03
feat(rome_analyze): implement the useSingleCaseStatement rule

leops:
## Summary

This adds the [useSingleCaseStatement](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/useSingleCaseStatement.md) rule back after it was removed in the analyzer refactor

## Test Plan

I added a snapshot test for the rule reusing the examples for the rule from Rome JS

### DIFF
--- a/crates/rome_analyze/src/analyzers/use_single_case_statement.rs
+++ b/crates/rome_analyze/src/analyzers/use_single_case_statement.rs
@@ -1,0 +1,118 @@
+use std::iter;
+
+use rome_console::markup;
+use rome_diagnostics::{Applicability, Severity};
+use rome_js_factory::make;
+use rome_js_syntax::{
+    JsAnyRoot, JsAnyStatement, JsCaseClause, JsCaseClauseFields, JsSyntaxToken, TriviaPieceKind, T,
+};
+use rome_rowan::{AstNode, AstNodeExt, AstNodeList, TriviaPiece};
+
+use crate::registry::{JsRuleAction, Rule, RuleDiagnostic};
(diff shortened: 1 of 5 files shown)


crates/rome_analyze/src/analyzers/use_single_case_statement.rs
```diff
@@ -1,20 +1,76 @@
...
+        if let Ok(token) = n.case_token() {
+            for piece in token.leading_trivia().pieces() {
+                token_text.push_str(piece.text());
+                leading.push(TriviaPiece::new(piece.kind(), piece.text_len()));
+            }
+        }
+
+        token_text.push('}');
+
+        let root = root.replace_node_discard_trivia(
```
MichaReiser, line 53: Maybe add some documentation that explains why you're handling trivia manually in this case.

MichaReiser (approved):

crates/rome_analyze/tests/specs/useSingleCaseStatement.js.snap
```diff
@@ -0,0 +1,58 @@
...
+source: crates/rome_analyze/tests/spec_tests.rs
+expression: useSingleCaseStatement.js
+---
+# Input
+```js
+switch (foo) {
+    case true:
+    case false:
+        let foo = '';
+        foo;
```
ematipico, line 11: Can we add a case where we have a comment?

```js
case false: // fallback
	let foo = '';
  foo;
```
And make sure we place it correctly?

ematipico (approved):

PR #2649, opened 2022-06-03
feat(rome_analyze): implement the validTypeof rule

leops:
## Summary

This PR implements the [valid-typeof](https://eslint.org/docs/rules/valid-typeof) rule from ESLint verifying the result of `typeof` unary expressions is being compared to a valid expression (either a string literal that's a valid type name or another `typeof` expression)

## Test Plan

I added test cases for the rule based on the examples found in the ESLint docs


crates/rome_analyze/src/analyzers/valid_typeof.rs
```diff
@@ -0,0 +1,198 @@
...
+use rome_rowan::{AstNode, AstNodeExt};
+
+use crate::{
+    registry::{Rule, RuleAction, RuleDiagnostic},
+    ActionCategory, RuleCategory,
+};
+
+pub(crate) enum ValidTypeof {}
+
+impl Rule for ValidTypeof {
```
MichaReiser, line 17: Please add some documentation of what the rule is testing for.

crates/rome_analyze/src/analyzers/valid_typeof.rs
```diff
@@ -0,0 +1,198 @@
...
+    JsBinaryExpressionFields, JsBinaryOperator, JsUnaryOperator,
+};
+use rome_rowan::{AstNode, AstNodeExt};
+
+use crate::{
+    registry::{Rule, RuleAction, RuleDiagnostic},
+    ActionCategory, RuleCategory,
+};
+
+pub(crate) enum ValidTypeof {}
```
MichaReiser, line 15: An empty enum... that's interesting. What's the reason behind using an empty enum over `struct ValidTypeof;`

crates/rome_analyze/src/analyzers/valid_typeof.rs
```diff
@@ -0,0 +1,198 @@
...
+                if unary.operator().ok()? != JsUnaryOperator::Typeof {
+                    return None;
+                }
+
+                if let JsAnyLiteralExpression::JsStringLiteralExpression(literal) = literal {
+                    let literal = literal.value_token().ok()?;
+                    let literal = literal
+                        .text_trimmed()
+                        .trim_start_matches(['"', '\''])
+                        .trim_end_matches(['"', '\''])
```
MichaReiser, line 63: You can do `literal.inner_string_text()`

crates/rome_analyze/src/analyzers/valid_typeof.rs
```diff
@@ -0,0 +1,198 @@
...
+            )
+            | (
+                lit @ JsAnyExpression::JsAnyLiteralExpression(literal),
+                JsAnyExpression::JsUnaryExpression(unary),
+            ) => {
+                if unary.operator().ok()? != JsUnaryOperator::Typeof {
+                    return None;
+                }
+
+                if let JsAnyLiteralExpression::JsStringLiteralExpression(literal) = literal {
```
MichaReiser, line 58: what if the right-hand side is a template literal? 

```
>> typeof "a" ===  `string`
true
```

MichaReiser (approved):

ematipico (changes requested):
This rule should follow the naming convention: https://github.com/rome/tools/blob/main/CONTRIBUTING.md#naming-patterns

Hence, it should be called `useValidTypeof` or `noInvalidTypeof`

crates/rome_analyze/src/analyzers/valid_typeof.rs
```diff
@@ -0,0 +1,198 @@
...
+            expr.clone(),
+            JsAnyExpression::JsAnyLiteralExpression(JsAnyLiteralExpression::from(
+                make::js_string_literal_expression(make::js_string_literal(type_name.as_str())),
+            )),
+        )?;
+
+        Some(RuleAction {
+            category: ActionCategory::QuickFix,
+            applicability: Applicability::MaybeIncorrect,
+            message: markup! { "Compare the result of `typeof` with a valid type name" }.to_owned(),
```
ematipico, line 152: This message should provide the list valid of type names as a suggestion. Essentially, we should suggest the user how can we fix the error

crates/rome_analyze/src/analyzers/valid_typeof.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        Some(None)
+    }
+
+    fn diagnostic(node: &Self::Query, _: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic {
+            severity: Severity::Error,
+            message: markup! {
+                "Invalid typeof comparison value"
```
ematipico, line 132: We should tell **why** the comparison is invalid.

```suggestion
                "Invalid `typeof` comparison value"
```

crates/rome_analyze/src/analyzers/valid_typeof.rs
MichaReiser: Do we have a max message length? Listing all potential types may get rather long.

ematipico: We do, we had a algorithm in Rome classic, here's the diagnostic was: https://github.com/rome/tools/blob/archived-js/internal/compiler/lint/rules/a11y/useValidLang.test.md#3

ematipico: This is the function in charge of creating suggestions: https://github.com/rome/tools/blob/archived-js/internal/diagnostics/helpers.ts#L27

crates/rome_analyze/src/analyzers/valid_typeof.rs
leops: It doesn't make a huge difference, we just need a type to implement the `Rule` trait on but that type will never be instantiated and this is made clear by using a type that actually can't be instantiated at all (while an empty struct can be instantiated, but will be zero-sized)

crates/rome_analyze/src/analyzers/valid_typeof.rs
leops: This is currently unsupported, I could probably add some code to handle the simple case with a single chunk / no interpolation, but it may not be needed eventually depending on how this rule interacts with the [noUnusedTemplateLiteral](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noUnusedTemplateLiteral.md) rule

crates/rome_analyze/src/analyzers/valid_typeof.rs
leops: Here the message is attached to a code action, meaning it will ultimately be displayed either as the name of a quick fix in the editor:
code_action
Or alongside a diff of the actual change to the user like this:
```
Compare the result of `typeof` with a valid type name

12    | - typeof foo === undefined
   12 | + typeof foo === "undefined"
```
So the message is meant more as a general explanation of what the actual action is doing, rather than an exhaustive suggestion on how to fix the code

crates/rome_analyze/src/analyzers/valid_typeof.rs
leops: To make a note on diagnostics here, in the context of a Rust-like diagnostic with rich codespan information it would be possible to attach context-specific messages to the actual label:
```
error[useValidTypeof]: Invalid typeof comparison value
   ┌─ useValidTypeof.js:10:1
   │
10 │ typeof bar !== "fucntion"
   │                ^^^^^^^^^^ not a valid type name
   
error[useValidTypeof]: Invalid `typeof` comparison value
   ┌─ useValidTypeof.js:14:1
   │
14 │ typeof bar == Object
   │               ^^^^^^ not a string literal
```
However I'm not entirely certain how to represent this information in the context of the LSP where we have a lot less control on how these diagnostics are rendered

crates/rome_analyze/src/analyzers/valid_typeof.rs
MichaReiser: There are no current usages of it. Should we change the implementation to return a string slice instead? Sounds like it's the more practical.

leops: Yes that would make sense, the only downside I see to that is that a `&str` slice doesn't have any associated range information but it shouldn't be too much of a problem

crates/rome_analyze/src/analyzers/valid_typeof.rs
ematipico: Probably we should provide two messages/diagnostics: one for the CLI/CI and one for the LSP

ematipico (approved):

PR #2661, opened 2022-06-06
feature(rome_analyze:) lint rule context and typed service bag

xunilrj:
PR for this discussion: https://github.com/rome/tools/discussions/2603

The heart of the PR is the type ```RuleContext```. It acts as a simple façade for everything that we want to offer to lint rules. What we offer varies by language. To see an example look at ```JsRuleContext```, which specifies which methods we allow to be called directly on the context (like extension methods), and ```impl LanguageSpecificServiceBag for JsLanguage``` which specifies which services are stored inside the context for each language.

The ```RuleContextServiceBag``` is typed and not a usual ServiceLocator to avoid all ```get(...)``` methods returning ```Option<...>```. This comes with a little bit of complexity but also guarantees that we will not see any runtime errors when linting.

Another nice side effect is that lint rules for other languages will not compile until someone specifies which services these languages need.

Rules now receive just the context as a parameter

```rust
    ...
}
```

And js rules, for example, have one extra method to give access to the root. With nice code completion and all:

image

# Borrow and Send + Sync

One thing I was not sure about was to constrain all services to be ```Send + Sync```. It seems to make sense, given that we will run rules in a thread pool at some point. But that would eliminate the ```JsAnyRoot```. So I opted to deal with this later.

Also, the context returns a non-mutable borrow, expecting that these services are thread-safe and have interior mutability through normal borrows.

### DIFF
--- a/crates/rome_analyze/src/context.rs
+++ b/crates/rome_analyze/src/context.rs
@@ -1,0 +1,112 @@
+use std::sync::Arc;
+
+use rome_js_syntax::{JsAnyRoot, JsLanguage};
+use rome_rowan::{AstNode, Language};
+
+use crate::{
+    registry::{LanguageRoot, Rule},
+    LanguageOfRule,
+};
+
+/// Specifies which services a language needs
+/// to be able to run all possible lint rules.
+///
+/// Also instantiate all these services.
+pub trait LanguageSpecificServiceBag: Language {
+    type Services;
+
+    fn new(root: LanguageRoot<Self>) -> Self::Services;
+}
+
+/// Easy access to which services a language needs to run all
+/// lint rules.
+type ServicesOf<Language> = <Language as LanguageSpecificServiceBag>::Services;
+
+/// Carries all services a language needs to run all
+/// lint rules
+#[derive(Clone)]
+pub struct RuleContextServiceBag<Language>
+where
+    Language: LanguageSpecificServiceBag,
+{
+    services: Arc<ServicesOf<Language>>,
+}
+
+impl<L> RuleContextServiceBag<L>
+where
+    L: LanguageSpecificServiceBag,
+{
+    pub fn new(root: LanguageRoot<L>) -> Self {
+        Self {
+            services: Arc::new(<L as LanguageSpecificServiceBag>::new(root)),
+        }
+    }
+}
+
+impl<Language> std::ops::Deref for RuleContextServiceBag<Language>
+where
+    Language: LanguageSpecificServiceBag,
+{
+    type Target = ServicesOf<Language>;
+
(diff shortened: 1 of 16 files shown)


crates/rome_analyze/src/context.rs
```diff
@@ -0,0 +1,109 @@
...
+    L: LanguageSpecificServiceBag,
+{
+    pub fn new(root: JsAnyRoot) -> Self {
+        Self {
+            services: Arc::new(<L as LanguageSpecificServiceBag>::new(root)),
+        }
+    }
+}
+
+impl<Language> std::ops::Deref for RuleContextServiceBag<Language>
```
MichaReiser, line 43: It's unclear to me at this point what other services we want to expose. Do you have some more in mind?

How about we go with a fixed set of "services" now and add a more generic approach only once we have the need for it?

crates/rome_analyze/src/context.rs
```diff
@@ -0,0 +1,109 @@
...
+        Self {
+            query_result,
+            services,
+        }
+    }
+
+    pub fn query(&self) -> &<TRule as Rule>::Query {
+        &self.query_result
+    }
+}
```
MichaReiser, line 85: Not sure if I missed this but how do I query a service from the service bag?

crates/rome_analyze/src/lib.rs
```diff
@@ -30,8 +34,13 @@ pub struct AnalysisFilter<'a> {
 /// Run the analyzer on the provided `root`: this process will use the given `filter`
 /// to selectively restrict analysis to specific rules / a specific source range,
 /// then call the `callback` when an analysis rule emits a diagnostic or action
-pub fn analyze<B>(file_id: FileId, root: &JsAnyRoot, filter: AnalysisFilter, mut callback: B)
-where
+pub fn analyze<B>(
+    file_id: FileId,
+    services: RuleContextServiceBag<JsLanguage>,
```
leops, line 39: I think the `file_id` and `root` could be merged into the service bag, conceptually both of these  are "services" the rules may need to access, we just need to statically ensure the analyzer can read them out as well

MichaReiser:
@xunilrj What are your thoughts on limiting this PR to only introduce the `RuleContext` and defer the `ServiceBag` until we have the very first service? I'm asking because I could see us get away with having the `semantic_model` exposed directly on the `RuleContext` in the first version and the decision on how to expose the semantic model will likely decide on how we schedule the lint rules.

xunilrj:
I am cancelling this PR, because there is no good way to implement the typed service bag now that we moved the js specific code to a new crate. I will implement a simplified version where RuleContext do not have a service bag.

PR #2665, opened 2022-06-07
fix(syntax): Remove quotes from tokens in `JsSyntaxKind::to_string`

MichaReiser:
## Summary

`JsSyntaxKind::to_string` should return the token's string representation without any additional quotes.
Quoting should happen at the call side if necessary.

This fixes some double-quoting of parentheses inside of parser diagnostic messages for different parentheses.

Needed for #2447 

## Test Plan

`cargo test`


MichaReiser:
@xunilrj why is it that there's always one symbol test that changes from panic to failed?

PR #2666, opened 2022-06-07
feat(formatter): Treat skipped token trivia as syntax error

MichaReiser:
**Merge AFTER the release cut**

## Summary
The parser uses skipped token trivia to recover from unrecognised syntax. For example, it could skip over the token `5` in `import 5 from "test";` because that allows to continue parsing the rest of the `from` statement.

Up to this point, we decided to format nodes with skipped token as if the nodes don't contain any syntax error because skipped token is, so far, only used to skip over decorator syntax. The issue with that is that our parser doesn't guarantee that the skipped decorators actually are valid and don't contain any syntax errors AND skipped token trivia might be used for real syntax errors at any point in the future. Thus, we shouldn't rely on that skipped token trivia is only used for syntax that Rome doesn't fully support.

This PR renames `FormatError::MissingRequiredChild` to `SyntaxError` and changes the leading trivia formatting to return `Err(FormatError::SyntaxError)` if any piece is a skipped token trivia to align it with how the rest of the formatter handles syntax errors.

This results in a few regressions on decorator formatting. The fix for that is to implement decorator parsing support.

## Test Plan

`cargo test`. Reviewed the updated snapshots.

### DIFF
--- a/crates/rome_js_formatter/src/builders.rs
+++ b/crates/rome_js_formatter/src/builders.rs
@@ -2,8 +2,7 @@ use crate::prelude::*;
 use crate::AsFormat;
 use rome_formatter::{format_args, write, Argument, Arguments, GroupId, PreambleBuffer, VecBuffer};
 use rome_js_syntax::{JsLanguage, JsSyntaxNode, JsSyntaxToken};
-use rome_rowan::syntax::SyntaxTriviaPiecesIterator;
-use rome_rowan::{AstNode, Language, SyntaxTriviaPiece};
+use rome_rowan::{AstNode, Language, SyntaxTriviaPiece, SyntaxTriviaPieceComments};
 
 /// Formats a token without its leading or trailing trivia
 ///
@@ -25,12 +24,12 @@ impl Format<JsFormatContext> for FormatTrimmedToken<'_> {
     }
 }
 
-/// Formats the leading trivia (comments, skipped token trivia) of a token
-pub const fn format_leading_trivia(
+/// Formats the leading comments of a token
+pub const fn format_leading_comments(
     token: &JsSyntaxToken,
     trim_mode: TriviaPrintMode,
-) -> FormatLeadingTrivia {
-    FormatLeadingTrivia { token, trim_mode }
+) -> FormatLeadingComments {
+    FormatLeadingComments { token, trim_mode }
 }
 
 /// Determines if the whitespace separating comment trivias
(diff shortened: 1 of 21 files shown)


MichaReiser:
I'll merge this AFTER the release.

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/builders.rs
```diff
@@ -50,220 +49,92 @@ pub struct FormatLeadingTrivia<'a> {
...
-            Ok(()) => Ok(()),
-            Err(_) => {
-                f.restore_snapshot(snapshot);
-
-                write_leading_trivia_with_skipped_tokens(self.token, self.trim_mode, f)
-            }
-        }
-    }
-}
+        let mut buffer = VecBuffer::new(f.state_mut());
```
MichaReiser, line 52: The "new" code here is the inlined `write_leading_trivia_pieces`. I didn't make any other changes (except changing some things to `self...` rather than using an argument.

leops:
I find this change makes the difference between unknown nodes and skipped trivias a bit unclear, since it seems they're now both used to signal pieces of source text that contain invalid syntax (when previously skipped trivias could be understood as valid syntax the syntax model simply doesn't support). If so then when should one be used over the other ?

MichaReiser:
> I find this change makes the difference between unknown nodes and skipped trivias a bit unclear, since it seems they're now both used to signal pieces of source text that contain invalid syntax (when previously skipped trivias could be understood as valid syntax the syntax model simply doesn't support). If so then when should one be used over the other ?

Skipped token is mainly beneficial if there's a single token that the parser didn't expect but the syntax otherwise is valid. In case of 

```
import 5 from "test";
```

This isn't a valid import statement because the import specifier must be an identifier. 

* Without skipped token trivia: The parser parses the `5` as a number and the syntax factory turns the whole import statement into an unknown statement. This is "fine" for formatting because we probably shouldn't format this statement but has the undesired side effect that analyzers won't be able to provide any suggestions on the rest of the import specifiers EXCEPT if there are unknown nodes for every list element. 
* Skipped token trivia allows the parser to gracefully recover. The way this would be implemented is that the parser looks ahead and tests... OK, I've no clue what the 5 means here but I see that the next token is a `from` token so I know how to continue from there. 

I agree, that it is a thin line and we could use `Unknown` for all cases but the problem is that `Unknown` provides more coarse-grained error recovery where skipped token trivia allows skipping over tokens without invalidating the parent.

ematipico (changes requested):
This has a **HUGE** *drawback*. Now we can't format documents that contain skipped tokens. 

I do remember that some people on Discord requested a way to format/parse code by ignoring the decorators. We implemented that feature but now we are reverting this (in the formatter, not the parser).

The great deal is that we don't format the document and we don't emit diagnostics (the playground doesn't show anything). Before merging, could you please open an issue and discuss how do you plan to solve the issue? I don't think we should discuss this in the PR

MichaReiser:
I didn't expect this to be that controversial. The main goal is to simplify comments formatting to be able to make progress but I see that I may need to overthink this, with the downside, that we may not make any progress soon. 

The issue with skipped token trivia is that they don't have a `JsSyntaxKind` which is necessary to decide on whatever it is necessary to introduce a leading/trailing space before inline comments. We can decide to always insert spaces around those but overall, skipped token trivia make comments even harder. 

> (when previously skipped trivias could be understood as valid syntax the syntax model simply doesn't support).

This wasn't my intention when I introduced skipped token trivia. It just turned out that this is, for now, the only use case for it. Both, in theory, represent a syntax error but where skipped token trivia offers better error recovery because it doesn't invalidate the enclosing node. 

I think the mistake I made back then was to actually implement formatting for skipped token trivia. The parser is free to choose what to use skipped token trivia for. It can do so to skip unsupported syntax but may also do so to recover from a parser error. While formatting might be desired for the first case, I don't think it is for the second. Therefore, the formatter should opt for the more defensive approach because it doesn't know for what purpose the parser used skipped token trivia for. Overall, I abused skipped token trivia by overfitting it for the use case of decorators. 

> This has a **HUGE** _drawback_. Now we can't format documents that contain skipped tokens.
> 
> I do remember that some people on Discord requested a way to format/parse code by ignoring the decorators. We implemented that feature but now we are reverting this (in the formatter, not the parser).
> 
> The great deal is that we don't format the document and we don't emit diagnostics (the playground doesn't show anything). Before merging, could you please open an issue and discuss how do you plan to solve the issue? I don't think we should discuss this in the PR

Formatting documents containing decorators will now work the same as when a document contains an `Unknown` node. It formats that node in verbatim node. That means, it actually aligns the behaviour closer to how the formatting of other nodes with syntax errors works. The issue that we have today is that our formatter would format documents that contain syntax errors (which the parser may have flagged) even though the user specified in the settings that they don't want that. 

I would even argue that our formatter currently doesn't handle skipped token trivia correctly because many formatting rules don't consider that a token has skipped token trivia. We would probably need to extend all checks that test if a node has any comments to also include skipped token trivia. Or in other words, check if there's any non-whitespace (including newline) trivia that the formatter can't influence its formatting.


> The great deal is that we don't format the document and we don't emit diagnostics (the playground doesn't show anything). Before merging, could you please open an issue and discuss how do you plan to solve the issue? I don't think we should discuss this in the PR.

This is a concern of the parser and not of the formatter. We can decide to emit warnings for decorators if that helps but I don't think that's what you want.

PR #2672, opened 2022-06-07
feat(rome_rowan): implement `DoubleEndedIterator` for node lists

ematipico:
## Summary

This PR adds the ability to consume an `AstSeparatedList` from the back by implementing `DoubleEndedIterator`.

The reason why I implemented this is because there have been cases where I needed to retrieve the last elements of a list. This will be useful, for example, inside the formatter when implementing the formatting of call arguments.

Checking a separated list from the back adds a bit more of complexity because the **last** separator is not mandatory, which means that we should handle cases where this is not an error. 

In order to achieve that, I had to make the slots peekable, because sometimes we don't want to consume two slots. An example is when we have `[1, separator, 3, separator, 5]`. We want to consume only `5` in the first iteration, but then we want to consume `separator` and `3`. 

I added the `unreacheable` macro. If we hit that macro, it means that are edge cases we have to handle.

## Test Plan

Created new tests to cover the cases:
- consuming from the back
- consuming from both ends and make sure that once the iterator is consumed from both ends, `None` is returned
- updated all the current tests with a new assertion called `assert_rev_elements`, we call `rev` on all vectors (`rev` needs `next_back`), so we should cover all the edge cases that were previously handled.

### DIFF
--- a/crates/rome_rowan/src/ast/mod.rs
+++ b/crates/rome_rowan/src/ast/mod.rs
@@ -392,44 +392,86 @@ impl<L: Language, N: AstNode<Language = L>> DoubleEndedIterator
     for AstSeparatedListElementsIterator<L, N>
 {
     fn next_back(&mut self) -> Option<Self::Item> {
-        let slot = self.slots.next_back();
+        let first_slot = self.slots.next_back();
+        let mut slots = self.slots.clone().rev().peekable();
+        let second_slot = slots.peek();
 
-        let (node, separator) = match slot {
-            Some(slot) => {
-                match slot {
-                    // we haven't found a token, which means that the last separator is not present
-                    SyntaxSlot::Node(node) => (Ok(N::unwrap_cast(node)), Ok(None)),
-                    // separator found, now let's check if we find a node
-                    SyntaxSlot::Token(separator) => {
-                        // then we should find the node
-                        let node = match self.slots.next_back()? {
-                            // The node for this element is missing if the next child is a token instead of a node.
-                            SyntaxSlot::Token(token) => panic!("Malformed list, node expected but found token {:?} instead. You must add missing markers for missing elements.", token),
-                            // Missing element
-                            SyntaxSlot::Empty => Err(SyntaxError::MissingRequiredChild),
-                            SyntaxSlot::Node(node) => Ok(N::unwrap_cast(node))
-                        };
-
-                        (node, Ok(Some(separator)))
-                    }
-                    SyntaxSlot::Empty => (
-                        Err(SyntaxError::MissingRequiredChild),
-                        Err(SyntaxError::MissingRequiredChild),
-                    ),
-                }
+        let (node, separator) = match (first_slot, second_slot) {
+            // case where the list ends with "1,"
+            (Some(SyntaxSlot::Token(separator)), Some(SyntaxSlot::Node(_))) => {
+                // SAFETY: both unwraps are checked by the previous match
+                let node = self.slots.next_back().unwrap().into_node().unwrap();
+                (Ok(N::unwrap_cast(node)), Ok(Some(separator)))
             }
-            None => {
-                // there wasn't any separator, let's try to go back and see if we find
-                let node = match self.slots.next_back()? {
-                    // The node for this element is missing if the next child is a token instead of a node.
-                    SyntaxSlot::Token(token) => panic!("Malformed list, node expected but found token {:?} instead. You must add missing markers for missing elements.", token),
-                    // Missing element
-                    SyntaxSlot::Empty => Err(SyntaxError::MissingRequiredChild),
-                    SyntaxSlot::Node(node) => Ok(N::unwrap_cast(node))
-                };
 
-                (node, Ok(None))
+            // case where the list ends with ",1"
+            // we dont consume the separator, we assume that we are at the end of the list and the
+            // separator is omitted
+            (Some(SyntaxSlot::Node(node)), Some(SyntaxSlot::Token(_))) => {
+                (Ok(N::unwrap_cast(node)), Ok(None))
             }
+
+            // case where the list ends with "1 1"
+            (Some(SyntaxSlot::Node(_)), Some(SyntaxSlot::Node(node))) => {
+                panic!("Malformed separated list, separator expected but found node {:?} instead. You must add missing markers for missing separators.", node);
+            }
+
+            // case where the list ends with ", ,"
+            (Some(SyntaxSlot::Token(_)), Some(SyntaxSlot::Token(token))) => {
+                panic!("Malformed list, node expected but found token {:?} instead. You must add missing markers for missing elements.", token);
+            }
+
+            // case where the list ends with "(missing sep) ,"
+            (Some(SyntaxSlot::Empty), Some(SyntaxSlot::Token(_))) => {
+                // SAFETY: both unwraps are checked by the previous match
+                let separator = self.slots.next_back().unwrap().into_token().unwrap();
+                (Err(SyntaxError::MissingRequiredChild), Ok(Some(separator)))
+                // panic!("Malformed list, node expected but found token {:?} instead. You must add missing markers for missing elements.", token);
+            }
+
+            // case where the list ends with "(missing node) (missing sep)"
+            (Some(SyntaxSlot::Empty), Some(SyntaxSlot::Empty)) => (
+                Err(SyntaxError::MissingRequiredChild),
+                Err(SyntaxError::MissingRequiredChild),
+            ),
+
+            // case where the list ends with "1 (missing sep)"
+            (Some(SyntaxSlot::Node(node)), Some(SyntaxSlot::Empty)) => (
+                Ok(N::unwrap_cast(node)),
+                Err(SyntaxError::MissingRequiredChild),
+            ),
+            // case where the list ends with "EMPTY ,"
+            (Some(SyntaxSlot::Token(separator)), Some(SyntaxSlot::Empty)) => {
+                self.slots.next_back();
+                (Err(SyntaxError::MissingRequiredChild), Ok(Some(separator)))
+            }
+
+            // case where the list ends with "EMPTY 1"
+            (Some(SyntaxSlot::Node(node)), None) => (
+                Ok(N::unwrap_cast(node)),
+                Err(SyntaxError::MissingRequiredChild),
+            ),
+
+            // case where the list ends with "EMPTY (missing sep)"
+            (Some(SyntaxSlot::Empty), None) => (
+                Err(SyntaxError::MissingRequiredChild),
+                Err(SyntaxError::MissingRequiredChild),
+            ),
+
+            // case where the list ends with "1 (missing sep)"
+            (Some(SyntaxSlot::Empty), Some(SyntaxSlot::Node(_))) => {
+                // SAFETY: covered by the previous match
+                let node = self.slots.next_back().unwrap().into_node().unwrap();
+                (Ok(N::unwrap_cast(node)), Ok(None))
+            }
+
+            (None, _) => {
+                return None;
+            }
+
+            _ => unreachable!(
+                "We should not reach this point. If so, there are some edge case not handled"
+            ),
         };
 
         Some(AstSeparatedElement {
@@ -603,27 +645,60 @@ mod tests {
         SeparatedExpressionList::new(node.into_list())
     }
 
-    fn assert_elements<'a>(
-        actual: impl Iterator<Item = AstSeparatedElement<RawLanguage, LiteralExpression>>,
+    fn map_elements<'a>(
+        actual: impl Iterator<Item = AstSeparatedElement<RawLanguage, LiteralExpression>>
+            + DoubleEndedIterator,
         expected: impl IntoIterator<Item = (Option<f64>, Option<&'a str>)>,
+        revert: bool,
+    ) -> (
+        Vec<(Option<f64>, Option<String>)>,
+        Vec<(Option<f64>, Option<String>)>,
     ) {
-        let actual = actual.map(|element| {
-            (
-                element.node.ok().map(|n| n.text().parse::<f64>().unwrap()),
-                element
-                    .trailing_separator
-                    .ok()
-                    .flatten()
-                    .map(|separator| separator.text().to_string()),
-            )
-        });
+        let actual: Vec<_> = if revert {
+            actual.rev().collect()
+        } else {
+            actual.collect()
+        };
+        let actual = actual
+            .into_iter()
+            .map(|element| {
+                (
+                    element.node.ok().map(|n| n.text().parse::<f64>().unwrap()),
+                    element
+                        .trailing_separator
+                        .ok()
+                        .flatten()
+                        .map(|separator| separator.text().to_string()),
+                )
+            })
+            .collect::<Vec<_>>();
 
         let expected = expected
             .into_iter()
             .map(|(value, separator)| (value, separator.map(|sep| sep.to_string())))
             .collect::<Vec<_>>();
 
-        assert_eq!(actual.collect::<Vec<_>>(), expected);
+        (actual, expected)
+    }
+
+    fn assert_elements<'a>(
+        actual: impl Iterator<Item = AstSeparatedElement<RawLanguage, LiteralExpression>>
+            + DoubleEndedIterator,
+        expected: impl IntoIterator<Item = (Option<f64>, Option<&'a str>)>,
+    ) {
+        let (actual, expected) = map_elements(actual, expected, false);
+
+        assert_eq!(actual, expected);
+    }
+
+    fn assert_rev_elements<'a>(
+        actual: impl Iterator<Item = AstSeparatedElement<RawLanguage, LiteralExpression>>
+            + DoubleEndedIterator,
+        expected: impl IntoIterator<Item = (Option<f64>, Option<&'a str>)>,
+    ) {
+        let (actual, expected) = map_elements(actual, expected, true);
+
+        assert_eq!(actual, expected);
     }
 
     fn assert_nodes(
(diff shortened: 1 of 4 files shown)


MichaReiser:
!bench_formatter

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -187,6 +187,12 @@ impl<L: Language, N: AstNode<Language = L>> ExactSizeIterator for AstNodeListIte
 
 impl<L: Language, N: AstNode<Language = L>> FusedIterator for AstNodeListIterator<L, N> {}
 
+impl<L: Language, N: AstNode<Language = L>> DoubleEndedIterator for AstNodeListIterator<L, N> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        Some(Self::slot_to_node(&self.inner.next_back()?))
+    }
+}
```
MichaReiser, line 194: Nit: Maybe implement `nth_back` too?

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,57 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
 {
 }
 
+impl<L: Language, N: AstNode<Language = L>> DoubleEndedIterator
+    for AstSeparatedListElementsIterator<L, N>
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        let slot = self.slots.next_back();
```
MichaReiser, line 395: You can use the `try` operator here so that the implementation returns immediately if you reached the start of the list
```suggestion
        let slot = self.slots.next_back()?;
```

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,57 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
...
+
+        let (node, separator) = match slot {
+            Some(slot) => {
+                match slot {
+                    // we haven't found a token, which means that the last separator is not present
+                    SyntaxSlot::Node(node) => (Ok(N::unwrap_cast(node)), Ok(None)),
+                    // separator found, now let's check if we find a node
+                    SyntaxSlot::Token(separator) => {
+                        // then we should find the node
+                        let node = match self.slots.next_back()? {
```
MichaReiser, line 405: Using the try operator here isn't safe because you then swallow the separator if there's no preceding node. I don't think this should ever be the case but it's better to be explicit about it.

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,57 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
...
+                            // Missing element
+                            SyntaxSlot::Empty => Err(SyntaxError::MissingRequiredChild),
+                            SyntaxSlot::Node(node) => Ok(N::unwrap_cast(node))
+                        };
+
+                        (node, Ok(Some(separator)))
+                    }
+                    SyntaxSlot::Empty => (
+                        Err(SyntaxError::MissingRequiredChild),
+                        Err(SyntaxError::MissingRequiredChild),
```
MichaReiser, line 417: I don't think this is correct. Let's say you have `[node1, missing_sep node 2]`. Only the separator between `missing_sep` and `node` is missing but `missing_sep` isn't. So we should still return it.

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,57 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
 {
 }
 
+impl<L: Language, N: AstNode<Language = L>> DoubleEndedIterator
+    for AstSeparatedListElementsIterator<L, N>
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        let slot = self.slots.next_back();
+
+        let (node, separator) = match slot {
```
MichaReiser, line 397: I think it would help to structure the code similar to the `next` implementation:

```
let token = match self.slots.next_back()? {
  ...
}

let node = match self.slots.next_back() {
  .. 
}

Some(AstSeparatedElement {
  node, 
  trailing_separator: separator
})
```

It reduces the nesting which simplifies reasoning about the code AND someone who needs to make a change to one method will recognise the structure and know how they need to make the change to the other method too.

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -594,6 +677,57 @@ mod tests {
...
+        let element = iter.next_back().unwrap();
+        assert_eq!(element.node().unwrap().text(), "1");
+        assert_eq!(element.trailing_separator().unwrap().unwrap().text(), ",");
+        // we test next_back once we consumed the whole iterator
+        let element = iter.next_back();
+        assert_eq!(element, None);
+    }
+
+    #[test]
+    fn separated_list_backwards_with_separator() {
```
MichaReiser, line 712: Can we please add some test cases for malformed lists. Similar to `separated_with_two_successive_separators`, `separated_with_leading_separator`, and `separated_with_two_successive_nodes`

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -594,6 +677,57 @@ mod tests {
...
+        assert_eq!(element.trailing_separator().unwrap().unwrap().text(), ",");
+        let element = iter.next_back().unwrap();
+        assert_eq!(element.node().unwrap().text(), "2");
+        assert_eq!(element.trailing_separator().unwrap().unwrap().text(), ",");
+        let element = iter.next_back().unwrap();
+        assert_eq!(element.node().unwrap().text(), "1");
+        assert_eq!(element.trailing_separator().unwrap().unwrap().text(), ",");
+        // we test next_back once we consumed the whole iterator
+        let element = iter.next_back();
+        assert_eq!(element, None);
```
MichaReiser, line 708: Adding a helper similar to `assert_elements` for asserting the elements from the back could greatly improve the readability of the tests.

MichaReiser (changes requested):

xunilrj:
Not sure we are implementing the ```std::iter::DoubleEndedIterator``` contract correctly. For example

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8e1cc1b755357cdd5546c3d270335d4d
```rust 
fn main() {
    let list = vec![1,2,3,4];
    
    let mut iter = list.iter();
    println!("{:?}", iter.next());
    println!("{:?}", iter.next_back());
    
    println!("{:?}", iter.next());
    println!("{:?}", iter.next_back());
    
    println!("{:?}", iter.next());
    println!("{:?}", iter.next_back());
}
```

outputs
```
Some(1)
Some(4)
Some(2)
Some(3)
None
None
```

This happens because: 

> It is important to note that both back and forth work on the same range, and do not cross: iteration is over when they meet in the middle.
https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html


I tried the following test and failed:
```rust
 #[test]
  fn double_iterator_meet_at_middle() {
      let list = build_list(vec![
          (Some(1), Some(",")),
          (Some(2), Some(",")),
          (Some(3), Some(",")),
          (Some(4), None),
      ]);

      let mut iter = list.elements();

      let element = iter.next().unwrap();
      assert_eq!(element.node().unwrap().text(), "1");
      let element = iter.next_back().unwrap();
      assert_eq!(element.node().unwrap().text(), "4");

      let element = iter.next().unwrap();
      assert_eq!(element.node().unwrap().text(), "2");
      let element = iter.next_back().unwrap();
      assert_eq!(element.node().unwrap().text(), "3");

      assert!(iter.next().is_none());  // line 701
      assert!(iter.next_back().is_none());
  }
```

and failed with:
```
---- ast::tests::double_iterator_meet_at_middle stdout ----
thread 'ast::tests::double_iterator_meet_at_middle' panicked at 'assertion failed: iter.next().is_none()', crates\rome_rowan\src\ast\mod.rs:701:9
```

ematipico:
> It is important to note that both back and forth work on the same range, and do not cross: iteration is over when they meet in the middle.
[doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html)

Damn I missed that part! Thank you for looking into this!

crates/rome_rowan/src/ast/mod.rs
ematipico: I can't here, because I might be at the end of the list where the separator isn't there and we have a node instead: `vec![Some(1), Some(","), Some(1)]`.

crates/rome_rowan/src/ast/mod.rs
ematipico: You think I haven't tried? :) 
Using the same code structure from `next` won't fit `next_back` because we are facing different situations that `next` doesn't have. The situation is mainly the fact that consuming a token from the back, especially at the very beginning, can yield a separator or a node and we have to check it (and it's not an error).

MichaReiser: > You think I haven't tried? :)

No, I didn't think that but glad you tried. I wasn't aware of this case. Thanks for explaining.

ematipico:
Please re-read the description, there's more context now

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,99 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
 {
 }
 
+impl<L: Language, N: AstNode<Language = L>> DoubleEndedIterator
+    for AstSeparatedListElementsIterator<L, N>
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        let first_slot = self.slots.next_back();
+        let mut slots = self.slots.clone().rev().peekable();
```
MichaReiser, line 396: What's the reason that you need to peek the next element here? The syntax factory should guarantee that each element takes two slots, with the sole exception of the last node. It should, therefore be safe to retrieve the next node?

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,99 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
...
+            (Some(SyntaxSlot::Token(_)), Some(SyntaxSlot::Token(token))) => {
+                panic!("Malformed list, node expected but found token {:?} instead. You must add missing markers for missing elements.", token);
+            }
+
+            // case where the list ends with "(missing sep) ,"
+            (Some(SyntaxSlot::Empty), Some(SyntaxSlot::Token(_))) => {
+                // SAFETY: both unwraps are checked by the previous match
+                let separator = self.slots.next_back().unwrap().into_token().unwrap();
+                (Err(SyntaxError::MissingRequiredChild), Ok(Some(separator)))
+                // panic!("Malformed list, node expected but found token {:?} instead. You must add missing markers for missing elements.", token);
```
MichaReiser, line 429: I think this case should panic because what I understand from your comment is that this is the case when the trailing separator is missing (marked with a missing marker).

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,99 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
...
+            }
+
+            // case where the list ends with "(missing node) (missing sep)"
+            (Some(SyntaxSlot::Empty), Some(SyntaxSlot::Empty)) => (
+                Err(SyntaxError::MissingRequiredChild),
+                Err(SyntaxError::MissingRequiredChild),
+            ),
```
MichaReiser, line 436: This is a malformed case in my view. There shouldn't be an element if both element and the token are missing.

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -364,6 +388,99 @@ impl<L: Language, N: AstNode<Language = L>> FusedIterator
...
+
+            // case where the list ends with "1 (missing sep)"
+            (Some(SyntaxSlot::Empty), Some(SyntaxSlot::Node(_))) => {
+                // SAFETY: covered by the previous match
+                let node = self.slots.next_back().unwrap().into_node().unwrap();
+                (Ok(N::unwrap_cast(node)), Ok(None))
+            }
+
```
MichaReiser, line 467: I'm a bit surprised that there are so many additional rules that must be handled when navigating backwards where significantly fewer cases are required when navigating forwards. 

The situations I see:

* The last element is a node -> Return the node
* In all other cases, the last element must be the separator token or missing. 
* After the separator, the next element must be the node

```rust
let separator = match self.inner.next_back()? {
  SyntaxSlot::Node(node) => {
    return AstSeparatedElement { node: Ok(Some(separator)), separator: Ok(None)
  },
  SyntaxSlot::Empty => {
    Err(Missing)
  },
  SyntaxSlot::Token(separator) => Ok(separator)
};

let node = match self.inner.next() {
  ...
}
```

The one case I'm not sure about is about the last case. Do we need some internal state tracking if the iterator is positioned at the last element because if not, then the case is invalid.

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -591,9 +749,94 @@ mod tests {
...
+            (Some(3), Some(",")),
+            (Some(4), None),
+        ]);
+
+        assert_eq!(list.len(), 4);
+        assert!(!list.is_empty());
+        assert_eq!(list.separators().count(), 3);
+
+        let mut iter = list.elements();
+        let element = iter.next_back().unwrap();
```
MichaReiser, line 778: Can we rewrite these tests to use the new utility functions?

crates/rome_rowan/src/cursor/node.rs
```diff
@@ -687,17 +691,27 @@ impl SyntaxSlots {
             )),
         }
     }
+
+    /// Checks if the front position is less than the back position.
+    /// When [false], it means that the iterator can't be consumed anymore and it has to return [None]
+    fn assert_can_consume(&self) -> bool {
+        ((self.front_next_position + self.back_next_position) as usize)
+            < self.parent.green().slots().len()
```
MichaReiser, line 699: My understanding is that `next` and `next_back` need to return `None` once the `back` and `front` position crossed in which case testing for `self.front_next_position > self.back_next.position` should be sufficient, I think

MichaReiser (commented):
Looking good. I'm a bit surprised by the amount of special case handling that is necessary and are wondering if we can simplify that a bit. But you will know better because you looked into all edge cases

crates/rome_rowan/src/cursor/node.rs
ematipico: Yes, this is correct if `back_next_position` would start from end of the iterator (its value would be `slots.len() - 1`). But this means that I would need to make `back_next_position` a `i32` because `0` is a correct value.

So I decided to give `back_next_position` the value of `0` and use the [`nth_back`](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#method.nth_back) when picking the next slot. This simplified by a lot the internal logic by making the implementation of `next` and `next_back` almost the same. The only caveat is the line that you pointed!  (I made the same mistake when I wrote it 😆 )

crates/rome_rowan/src/ast/mod.rs
ematipico: It's exactly because of the last node that we can't consume twice regardless.

When consuming the iterator from the front, the operation is simple because the first node of each tuple **must be a node** and the second **can be a separator or be empty**. This means that once we consume the entire iterator keep getting `None` regardless and we safely process the information.

From the back, unfortunately we don't have  the luxury to consume the iterator twice regardless. Let's get these two examples:

```
// 1
["1", separator, "2", separator, "3"]
// 2
["1", separator, "2", separator, "3", separator]
```

If we consume `// 1` from the front with `.next` we should get 3 items, where the last one is `("3", missing_sep)`. This means that we should have the same result when consuming items from the back. If we decided always consume two items at each iteration, we would receive something like this:
```
[separator, "3"] -> [separator, "2"] -> ["1"]
```

Which is not correct, that's why I had to make the second slot a peekable and make to consume the item only when there aren't edge cases.

MichaReiser: Shouldn't it be sufficient to match on the first element

* Node -> You can assume that it was the last element because the list must otherwise have an empty element for the separator -> Return immediately 
* Empty -> The separator is missing, call next again
* Token -> Separator is present, call next again

You can then match on the next value: 
* `None` -> Uhm, that's an invalid state, there should have been a node
* Some(Token) -> Invalid list, should have been a node
* Some(Node) -> Ok 
* Some(Empty) -> Err(Missing Node)

I personally find this a bit easier to reason about the code.

crates/rome_rowan/src/ast/mod.rs
MichaReiser: > Using the try operator would return None and this is not correct, because the end of a list might have a node with a missing trailing separator.

I don't understand why this is so. If calling `self.inner_next_back()` returns `None`, then calling it again will always return from that point onwards (knowing that the inner iterator is a `FusedIterator`). 

A missing separator should have been marked by the syntax factory with `Missing` and shouldn't return `None`.

crates/rome_rowan/src/cursor/node.rs
MichaReiser: Would you mind adding some documentation to `back_next_position` that it's the offset from the end and not the next slot's index.

crates/rome_rowan/src/ast/mod.rs
ematipico: OK, that helps a lot, thank you. Unfortunately that wasn't clear to me

ematipico:
@MichaReiser I addressed your concerns, your help was really helpful, thank you! Now the code should be clear and better to read. I believe we are ready to ship it!

crates/rome_rowan/src/cursor/node.rs
ematipico: It was already there, I added a bold 😆

MichaReiser (approved):
Awesome!

PR #2675, opened 2022-06-07
fix(rome_formatter): measure first variant of BestFitting in flat mode

yassere:
## Summary

This is a minimal fix for the `BestFitting` element that was broken by #2645. It now attempts to fit the first variant in flat mode on a single line. If that doesn't fit, it continues attempting the remaining variants in `PrintMode::Expanded` before defaulting to the `most_expanded` variant. This seems to match the original intention based on the doc comments for the `best_fitting` macro. @MichaReiser?

I don't know if this will be sufficient for our needs, but I believe that matching the exact behavior of Prettier's `conditionalGroup` could involve heavy enough changes to our "fits" measurement and printing that it may be worth trying this out first.

Happy to close this, though, if we'd rather go with a different approach.

## Test Plan

Removed the `ignore` from the broken `best_fitting` doc test.

### DIFF
--- a/crates/rome_formatter/src/macros.rs
+++ b/crates/rome_formatter/src/macros.rs
@@ -236,15 +236,20 @@ macro_rules! format {
 /// * The worst case complexity is that the printer tires each variant. This can result in quadratic
 ///   complexity if used in nested structures.
 ///
-/// ## Prettier
-/// This IR is similar to Prettier's `ConditionalGroupContent` IR. It provides similar functionality but
+/// ## Behavior
+/// This IR is similar to Prettier's `conditionalGroup`. It provides similar functionality but
 /// differs in that Prettier automatically wraps each variant in a `Group`. Rome doesn't do so.
 /// You can wrap the variant content in a group if you want to use soft line breaks.
-/// Unlike in Prettier, a variant (except the first) will be considered to fit if it can be printed without
+/// Unlike in Prettier, the printer will try to fit **only the first variant** in [`Flat`] mode,
+/// then it will try to fit the rest of the variants in [`Expanded`] mode.
+///
+/// A variant that is measured in [`Expanded`] mode will be considered to fit if it can be printed without
 /// overflowing the current line with all of its inner groups expanded. Those inner groups could still end
 /// up being printed in flat mode if they fit on the line while printing. But there is currently no way
 /// to enforce that a specific group inside a variant must be flat when measuring if that variant fits.
 ///
+/// [`Flat`]: crate::format_element::PrintMode::Flat
+/// [`Expanded`]: crate::format_element::PrintMode::Expanded
 #[macro_export]
 macro_rules! best_fitting {
     ($least_expanded:expr, $($tail:expr),+ $(,)?) => {{
@@ -310,10 +315,11 @@ mod tests {
     }
 
     #[test]
-    fn best_fitting_groups() {
+    fn best_fitting_variants_print_as_lists() {
         use crate::prelude::*;
         use crate::{format, format_args, Formatted};
 
+        // The second variant below should be selected when printing at a width of 30
         let formatted_best_fitting = format!(
             SimpleFormatContext::default(),
             [
@@ -356,6 +362,8 @@ mod tests {
         )
         .unwrap();
 
+        // This matches the IR above except that the `best_fitting` was replaced with
+        // the contents of its second variant.
         let formatted_normal_list = format!(
             SimpleFormatContext::default(),
             [
@@ -408,6 +416,8 @@ mod tests {
         .as_code()
         .to_string();
 
+        // The variant that "fits" will print its contents as if it were a normal list
+        // outside of a BestFitting element.
         assert_eq!(best_fitting_code, normal_list_code);
     }
 }
--- a/crates/rome_formatter/src/macros.rs
+++ b/crates/rome_formatter/src/macros.rs
@@ -303,4 +303,106 @@ mod tests {
             ]))
         );
     }
+
+    #[test]
+    fn best_fitting_groups() {
+        use crate::prelude::*;
+        use crate::{format, format_args, Formatted};
+
+        let formatted_best_fitting = format!(
+            SimpleFormatContext::default(),
+            [
+                token("aVeryLongIdentifier"),
+                soft_line_break_or_space(),
+                best_fitting!(
+                    format_args!(token(
+                        "Something that will not fit on a 30 character print width."
+                    ),),
+                    format_args![
+                        token("Start"),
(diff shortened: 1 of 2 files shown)


MichaReiser (commented):
This may fix the test bit I don't think it works as expected because any group inside of the second or later variant always get printed in expanded mode (same as for rest elements). My understanding is that the printing of a variant should work the same as any other content outside of the group where the first group changes the print mode to flat to all its enclosing group and the mode only changes back for the groups from the rest queue

yassere:
> This may fix the test bit I don't think it works as expected because any group inside of the second or later variant always get printed in expanded mode (same as for rest elements). My understanding is that the printing of a variant should work the same as any other content outside of the group where the first group changes the print mode to flat to all its enclosing group and the mode only changes back for the groups from the rest queue

Ah, yeah that makes sense.

yassere:
>  My understanding is that the printing of a variant should work the same as any other content outside of the group

This seems to be the case, unless I'm misunderstanding what you're saying here. I added a test where the best fitting second variant is printed the same as it would be if it were printed directly. Inner groups are still printed flat when possible. Do you have an example where the current implementation doesn't work as intended? Do you think the problem is with how it measures, how it prints, or both?

Also, I'm not sure about this line:
https://github.com/rome/tools/blob/f68757316355bb931c0420a686082297978a370e/crates/rome_formatter/src/printer/mod.rs#L281

What's the purpose of setting that as true when measuring in expanded mode?

MichaReiser:
> > My understanding is that the printing of a variant should work the same as any other content outside of the group
> 
> This seems to be the case, unless I'm misunderstanding what you're saying here. I added a test where the best fitting second variant is printed the same as it would be if it were printed directly. Inner groups are still printed flat when possible. Do you have an example where the current implementation doesn't work as intended? Do you think the problem is with how it measures, how it prints, or both?

It could be but let's say we have the following example:

```
// variant2: 
token("("), group_elements(
  token("["), soft_line_break(), 
  indent(token("abcdef")), 
  soft_line_break(), 
  token("]"))
)
```

My understanding is that the printer calls `fits_on_line` with print_mode: `expanded` for variant 2. The printer then enqueues the group 2 with the print mode it's currently in, the same as for the variant, which is `expanded` here

https://github.com/rome/tools/blob/941faf9164b1f68da660e093a5697fefc2923022/crates/rome_formatter/src/printer/mod.rs#L726

Which will ultimately cause the `soft_line_breaks` to break because the `print_mode` is `Expanded` when the printer measures the first `soft_line_break`. 

 
> Also, I'm not sure about this line:
> 
> https://github.com/rome/tools/blob/f68757316355bb931c0420a686082297978a370e/crates/rome_formatter/src/printer/mod.rs#L281
> 
> What's the purpose of setting that as true when measuring in expanded mode?

The purpose on that line is to cache the fact that everything following now up to the first line break will fit on the line. So that the printer doesn't need to re-measure if the second group contains an inner group:

```
// variant2: 
token("("), group_elements(token("["), token("abcdef"), token("]"))
```

There's no point in re-calculating if the group containing the array fits because the printer already ensured that this is the case when measuring the variant content. 

The boolean gets set to false when printing the next line break and any group following then must be re-measured.

yassere:
> Which will ultimately cause the soft_line_breaks to break because the print_mode is Expanded when the printer measures the first soft_line_break.

They won't necessarily break during printing if that group fits on one line. The second variant will be selected if it fits in expanded mode, but a group that's enqueued for printing in expanded mode will still be printed flat if it fits.

If we have:
```
best_fitting!(
    format_args!(token(
        "Something that will not fit on a 30 character print width."
    )),
    format_args![
        token("("),
        group_elements(&format_args![
            token("["),
            soft_line_break(),
            indent(&token("abcdef")),
            soft_line_break(),
            token("]")
        ])
    ],
    format_args!(token("Most"), hard_line_break(), token("Expanded"))
)
```
Formatting that at width 80 prints the first variant.
Formatting that at width 30 prints the second variant as:
```
([abcdef]
```
Formatting that at width 8 prints the second variant as:
```
([
abcdef
]
```
Formatting that at width 1 defaults to printing the most expanded variant.

Does that address your concern or am I still misunderstanding?

> This should not fit because the printer should try to print the inner group as flat and not as expanded but the printer breaks at the soft_line_breaks and, therefore claims that this variant fits.

That variant would be selected because it fits in expanded mode, but inner groups would still be printed flat if they fit.
If we only want a variant to be selected if it fits entirely in flat mode, we can change that. But it seems less useful to me because if all of the content could fit on a single line, then shouldn't the first variant fit? In real use cases, I assume you won't generally have tokens in the first variant that can be omitted in later variants.

yassere:
> I think where this solution differs from what I thought it does is that inner groups "fit" if the content up to the first line break fits (including soft) whereas what prettier does (I think) is that the group only fits if the whole group in flat mode fits.

Yeah, that's a good point. In this implementation, there's no way to enforce that a specific group inside a variant must be flat when measuring if that variant fits. If that ends up being important, we can follow up with a new approach. I think we'd need to add something like `should_break` to our `Group` to enable more fine-tuned control of the breaking behavior of variants.

crates/rome_formatter/src/macros.rs
```diff
@@ -239,9 +237,14 @@ macro_rules! format {
 ///   complexity if used in nested structures.
 ///
 /// ## Prettier
-/// This IR is similar to Prettier's `ConditionalGroupContent` IR. It provides the same functionality but
+/// This IR is similar to Prettier's `ConditionalGroupContent` IR. It provides similar functionality but
 /// differs in that Prettier automatically wraps each variant in a `Group`. Rome doesn't do so.
 /// You can wrap the variant content in a group if you want to use soft line breaks.
+/// Unlike in Prettier, a variant (except the first) will be considered to fit if it can be printed without
```
ematipico, line 243: Would it be possible to rename the "prettier" title with "Behavior"? Eventually we have to rely to our own algorithms, not others. So us, the consumers, should rely to own documentation. 

For example, making a comparison, a sense forces me to understand where's the difference with prettier, but once we settle on things should work, it doesn't make much sense to check other's documentation.

crates/rome_formatter/src/macros.rs
```diff
@@ -305,4 +308,106 @@ mod tests {
             ]))
         );
     }
+
+    #[test]
+    fn best_fitting_groups() {
```
ematipico, line 313: Can we rename the function, trying to say what are we actually asserting? The name doesn't help, or can we add some comments to help understand which alternatives should be the chosen ones?

ematipico (approved):
Thank you for addressing my nitpicks! Let's 🚢 it!!

PR #2677, opened 2022-06-07
fix(parser): Removed use strict check in parser

NicholasLYang:
## Summary

We have a check in the parser for redundant `"use strict"` directives. This should not be a parser warning but a lint rule.

## Test Plan

The `directives_err` test that was previously `test_err` is now `directives_redundant` and `test`.


MichaReiser (approved):

crates/rome_js_parser/test_data/inline/ok/ts_type_assertion.ts
```diff
@@ -0,0 +1 @@
+let a = <number>b;
```
Boshen, line 1: Where is the source for this 😕

MichaReiser: It's probably a test that was added but someone forgot to run `cargo test`. We should probably add a CI check for this at some point 

https://github.com/rome/tools/blob/b05b3034fe07fd97b64cb31cf1a8400cd33515dd/crates/rome_js_parser/src/syntax/expr.rs#L1305-L1307

PR #2681, opened 2022-06-08
refactor(formatter): Rename `FormatError::MissingRequiredChild` to `SyntaxError`

MichaReiser:
## Summary

Rename `MissingRequiredChild` to `SyntaxError` to account for other syntax errors, for example, when a rest element has a trailing comma.

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_js_formatter/src/builders.rs
+++ b/crates/rome_js_formatter/src/builders.rs
@@ -105,7 +105,7 @@ where
         }
 
         if piece.is_skipped() {
-            return Err(FormatError::MissingRequiredChild);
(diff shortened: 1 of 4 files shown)


crates/rome_js_formatter/src/builders.rs
```diff
@@ -105,7 +105,7 @@ where
         }
 
         if piece.is_skipped() {
-            return Err(FormatError::MissingRequiredChild);
+            return Err(FormatError::SyntaxError);
```
MichaReiser, line 108: This does not change how skipped trivia gets formatted. I removed that change.

PR #2683, opened 2022-06-08
refactor(rome_js_formatter): avoid snapshot when formatting skipped token trivia

MichaReiser:
## Summary

This removes the need to take a snapshot before formatting leading trivia by collecting the comments into a `Vec`. The performance characteristics of the new implementation should be similar to the old because it removes the need to allocate a temporary `VecBuffer` to be able to reverse the written comments in the end. 

I find this solution slightly easier to reason about and results in slightly less code: 

Before:
*  3.5%  10.9% 422.4KiB rome_js_formatter
*  0.1%   0.4%   9.7KiB rome_js_formatter <rome_js_formatter::builders::FormatLeadingTrivia as rome_formatter::Format<rome_js_formatter::context::JsFormatContext>>::fmt

After:

*  3.3%  10.6% 407.1KiB rome_js_formatter
* `FormatLeadingTrivia` no longer shows up. 

It also gives us the option to use a `SmallVec` to store comments if profiling shows that this is beneficial. 

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_js_formatter/src/builders.rs
+++ b/crates/rome_js_formatter/src/builders.rs
@@ -1,15 +1,13 @@
 use crate::prelude::*;
-use crate::{AsFormat, TextRange};
+use crate::AsFormat;
 use rome_formatter::{format_args, write, Argument, Arguments, GroupId, PreambleBuffer, VecBuffer};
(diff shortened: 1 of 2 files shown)


MichaReiser:
!bench_formatter

ematipico (approved):

leops (approved):

PR #2685, opened 2022-06-08
refactor(rome_formatter): Add `Interned` `FormatElement`

MichaReiser:
## Summary

Introduces a new reference-counted `FormatElement`. Useful for situations where the same content is part of the resulting IR twice but with different parent formatting. To this point, it has been necessary to clone the whole sub-tree. The ref counted FormatElement on the other hand only requires bumping the ref count. 

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -5,6 +5,7 @@ use rome_rowan::SyntaxTokenText;
 use std::borrow::Cow;
 use std::fmt::{self, Debug, Formatter};
 use std::ops::Deref;
+use std::rc::Rc;
 
 type Content = Box<FormatElement>;
 
@@ -65,6 +66,10 @@ pub enum FormatElement {
     /// A list of different variants representing the same content. The printer picks the best fitting content.
     /// Line breaks inside of a best fitting don't propagate to parent groups.
     BestFitting(BestFitting),
+
+    /// An interned format element. Useful when the same content must be emitted multiple times to avoid
+    /// deep cloning the IR when using the `best_fitting!` macro or `if_group_fits_on_line` and `if_group_breaks`.
+    Interned(Interned),
 }
 
 #[derive(Clone, Copy, Eq, PartialEq, Debug)]
@@ -144,6 +149,7 @@ impl Debug for FormatElement {
                 best_fitting.fmt(fmt)
             }
             FormatElement::ExpandParent => write!(fmt, "ExpandParent"),
+            FormatElement::Interned(inner) => inner.fmt(fmt),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


leops:
I think with a flat IR this could be implemented using a `(usize, usize)` as a back-reference to a previous range of the buffer, but these would need special handling in operations that need to rewrite the IR

MichaReiser:
> I think with a flat IR this could be implemented using a `(usize, usize)` as a back-reference to a previous range of the buffer, but these would need special handling in operations that need to rewrite the IR

Oh, that's clever.

NicholasLYang:
Yeah IME interning is a little nicer than Rc. That is, unless you need to delete/deallocate.

crates/rome_formatter/src/format_element.rs
```diff
@@ -65,6 +66,10 @@ pub enum FormatElement {
     /// A list of different variants representing the same content. The printer picks the best fitting content.
     /// Line breaks inside of a best fitting don't propagate to parent groups.
     BestFitting(BestFitting),
+
+    /// Reference counted format element content. Useful when the same content must be emitted twice
+    /// because it avoids deep cloning of the inner content and instead only requires bumping a ref counter.
+    Rc(Rc<FormatElement>),
```
ematipico, line 72: Reference count is low level term which doesn't fit well with our current IR naming convention. How about `CanCopy` or `Sharable` (I prefer this one)?  The architecture of the IR is a detail

ematipico:
I have used this commit in my PR, locally, and it definitely improves the caching, especially on heavy workload:

Before:
```
Formatted 1 files in 1436ms
```
After:
```
Formatted 1 files in 43ms
```

For this case, I formatted form the CLI the file: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/tests/specs/prettier/js/performance/nested-real.js

MichaReiser:
> I have used this commit in my PR, locally, and it definitely improves the caching, especially on heavy workload:
> 
> Before:
> 
> ```
> Formatted 1 files in 1436ms
> ```
> 
> After:
> 
> ```
> Formatted 1 files in 43ms
> ```
> 
> For this case, I formatted from the CLI the file: [https://github.com/rome/tools/blob/main/crates/rome_js_formatter/tests/specs/prettier/js/performance/nested-real.js](https://github.com/rome/tools/blob/main/crates/rome_js_formatter/tests/specs/prettier/js/performance/nested-real.js?rgh-link-date=2022-06-14T20%3A03%3A58Z)
> 
> The workload is caused by the fact that each argument of a call expression is memoized. Imagine this for each call argument inside this file.

OMG, 1.4s for an 80 lines file... 

> but these would need special handling in operations that need to rewrite the IR

I still need to figure out a way how to support this new Element in `split_trivia` and `group_elements`

MichaReiser:
> > I have used this commit in my PR, locally, and it definitely improves the caching, especially on heavy workload:
> > Before:
> > ```
> > Formatted 1 files in 1436ms
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After:
> > ```
> > Formatted 1 files in 43ms
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For this case, I formatted from the CLI the file: [https://github.com/rome/tools/blob/main/crates/rome_js_formatter/tests/specs/prettier/js/performance/nested-real.js](https://github.com/rome/tools/blob/main/crates/rome_js_formatter/tests/specs/prettier/js/performance/nested-real.js?rgh-link-date=2022-06-15T06%3A24%3A35Z)
> > The workload is caused by the fact that each argument of a call expression is memoized. Imagine this for each call argument inside this file.
> 
> OMG, 1.4s for an 80 lines file...
> 
> > but these would need special handling in operations that need to rewrite the IR
> 
> I still need to figure out a way how to support this new Element in `split_trivia` and `group_elements`

I removed `split_trivia` in #2709 and #2718 so it should now be safe to wrap elements with an `Interned` element because there are no more usages that mutate the IR.

leops (approved):

ematipico (approved):

PR #2690, opened 2022-06-09
feature(rome_js_semantic): matching declarations and scopes

xunilrj:
This PR is the part that matches declarations and scopes of the main PR (https://github.com/rome/tools/pull/2489).

This is just the first PR for this match, as we are solving only the most naive cases.

This also introduces a new assertion to test if the declaration is actually pointing to the correct scope. The important part here would be the ```@A```, which means that the current declaration should be pointing to a scope named ```A```.
```js
if/*START A*/ (true) { let b/*#b @A*/ = 1; }
```

#### Better Diagnostics

Now the diagnostic specifies more clearly which test failed.

image

### DIFF
--- a/crates/rome_js_semantic/src/tests/mod.rs
+++ b/crates/rome_js_semantic/src/tests/mod.rs
@@ -4,7 +4,7 @@ pub mod scopes;
 
 #[macro_export]
 macro_rules! assert_semantics {
-    ($($name:ident, $code:expr,)*) => {
+    ($(#[$attr:meta])* $($name:ident, $code:expr,)*) => {
         $(
             #[test]
             pub fn $name() {


crates/rome_js_semantic/src/events.rs
```diff
@@ -107,8 +112,44 @@ impl SemanticEventExtractor {
...
             }),
-            JS_BLOCK_STATEMENT | JS_FUNCTION_BODY => self.push_scope(node.text_range()),
+
+            JS_MODULE | JS_SCRIPT => self.push_scope(node.text_range()),
+            JS_FUNCTION_DECLARATION
+            | JS_ARROW_FUNCTION_EXPRESSION
+            | JS_CONSTRUCTOR_CLASS_MEMBER
+            | JS_GETTER_CLASS_MEMBER
+            | JS_SETTER_CLASS_MEMBER
+            | JS_IF_STATEMENT
```
MichaReiser, line 124: Is it the case that the if statement creates a new scope? Isn't it only creating a new scope if it's body is a `JsBlockStatement` (which is missing here)?

crates/rome_js_semantic/src/events.rs
```diff
@@ -119,11 +160,54 @@ impl SemanticEventExtractor {
         use rome_js_syntax::JsSyntaxKind::*;
 
         match node.kind() {
-            JS_BLOCK_STATEMENT | JS_FUNCTION_BODY => self.pop_scope(node.text_range()),
+            JS_MODULE | JS_SCRIPT => self.pop_scope(node.text_range()),
```
MichaReiser, line 163: Do you see a way how we can reduce the duplication here? Seems that what this is really about is defining a set of nodes that introduce a new block.

crates/rome_js_semantic/src/tests/declarations.rs
```diff
@@ -0,0 +1,279 @@
...
+    ok_declaration_class_constructor, ";class A { constructor/*START A*/ (a/*#a @A*/) { let b/*#b @A*/ = 1; } }",
+    ok_declaration_class_getter, ";class A { get/*START A*/ name() { let b/*#b @A*/ = 1;} }",
+    ok_declaration_class_setter, ";class A { set/*START A*/ name(a/*#a @A*/) { let b/*#b @A*/ = 1;} }",
+    ok_declaration_try_catch, ";try {/*START A*/ let a/*#a @A*/ = 1;} catch/*START B*/ (b) { let b/*#b @B*/ = 1; }",
+    ok_declaration_try_catch_finally, ";try {/*START A*/ let a/*#a @A*/ = 1;} catch/*START B*/ (b) { let b/*#b @B*/ = 1; } finally/*START C*/ { let c/*#c @C*/ = 1; }",
+    ok_declaration_with_inner_scopes, r#";
+function/*START SCOPE1*/ f() {
+    let a/*#a1 @SCOPE1*/ = 1;
+    console.log(a);
+    if/*START SCOPE2*/ (true) {
```
MichaReiser, line 37: I'm not sure if that's correct. My understanding is that the block statement starts the new scope

crates/rome_js_semantic/src/events.rs
```diff
@@ -119,11 +160,54 @@ impl SemanticEventExtractor {
...
+            JS_FUNCTION_DECLARATION
```
MichaReiser, line 164: I'm not certain but I believe the `JsFunctionBody` starts yet another scope but that's something you would need to verify with the JS spec.

ematipico (approved):

crates/rome_js_semantic/src/events.rs
ematipico: Yes, you're right. We actually have two scopes: one inside the function arguments and one inside the function body. 

Function name  has to tolerate the global scope. The scope inside the body has to tolerate the global scope and the scope created inside the arguments.

Meaning doing
```js
function f(a) { let a; }
``` 

Will error because of `a` being already declared

ematipico:
> Have you checked the JS spec on what introduces a new scope? I believe this is (partially) the relevant section. [tc39.es/ecma262/#sec-static-semantics-lexicallydeclarednames](https://tc39.es/ecma262/#sec-static-semantics-lexicallydeclarednames)
> 
> Another alternative would be to look at rome classic's implementation. Maybe @ematipico has a pointer for you?

Rome classic had a nice way to test things, although it was via snapshot tests: https://github.com/rome/tools/blob/archived-js/internal/compiler/scope/Scope.test.md

Maybe you can use it as inspiration.

This is where we stored all the globals: https://github.com/rome/tools/blob/archived-js/internal/compiler/scope/globals.ts, they were excluded from being tracked as bindings inside scopes.

This is where the bindings were created, and as you can see they have different kinds: https://github.com/rome/tools/blob/archived-js/internal/compiler/scope/bindings.ts

The creation of the scopes was done at node level by using the evaluators. I believe this is pretty nifty for various reasons:
- each evaluator that we create is bound to a specific node, which means that we know that specific evaluator will create bindings/scopes
- easy to debug

Here's the list of evaluators: https://github.com/rome/tools/tree/archived-js/internal/compiler/scope/evaluators
Here's the example for block statement: https://github.com/rome/tools/blob/archived-js/internal/compiler/scope/evaluators/JSBlockStatement.ts

We use the function `.fork` which creates a new scope but it keeps the same root scope.

crates/rome_js_semantic/src/events.rs
xunilrj: Fixed.

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

crates/rome_js_semantic/src/tests/declarations.rs
xunilrj: Fixed.

crates/rome_js_semantic/src/events.rs
xunilrj: Fixed.

crates/rome_js_semantic/src/tests/declarations.rs
```diff
@@ -0,0 +1,279 @@
...
+use rome_js_syntax::TextRange;
+use rome_js_syntax::WalkEvent;
+use rome_markup::markup;
+use rome_rowan::NodeOrToken;
+
+use super::extract_scope_assertion;
+use super::ScopeAssertionType;
+
+assert_semantics! {
+    ok_declaration_import, "/*START GLOBAL*/ import a/*#a @GLOBAL*/ from 'a'",
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_js_semantic/src/events.rs
```diff
@@ -107,8 +112,44 @@ impl SemanticEventExtractor {
...
+                scope_started_at: self.current_scope_start(),
             }),
-            JS_BLOCK_STATEMENT | JS_FUNCTION_BODY => self.push_scope(node.text_range()),
+
+            JS_MODULE | JS_SCRIPT => self.push_scope(node.text_range()),
+            JS_FUNCTION_DECLARATION
+            | JS_ARROW_FUNCTION_EXPRESSION
+            | JS_CONSTRUCTOR_CLASS_MEMBER
+            | JS_GETTER_CLASS_MEMBER
+            | JS_SETTER_CLASS_MEMBER
```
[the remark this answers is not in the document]
xunilrj: Will be done in another PR.

crates/rome_js_semantic/src/events.rs
```diff
@@ -119,11 +160,54 @@ impl SemanticEventExtractor {
...
         }
     }
 
+    fn current_scope_start(&self) -> TextSize {
+        let started_at = self.scopes.last().map(|x| x.started_at);
+
+        // We should always have, at least, the global scope
+        debug_assert!(started_at.is_some());
+
+        started_at.unwrap_or_else(|| TextSize::of(""))
```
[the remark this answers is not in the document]
xunilrj: I can write this in a way to have an ```unreachable!(...)```, for crazy scenarios.

PR #2691, opened 2022-06-09
feat(rome_rowan): add a `declare_union` macro for one-off union AstNodes

leops:
## Summary

This PR add a new `declare_union` for creating union AST nodes scoped to a local module, without having to add them to `js.ungram` and `rome_js_syntax`. Its usage looks like the following example:

```rust
declare_union! {
    /// Documentation
    #[allow(dead_code)]
    pub(crate) UnionName = NodeType | OtherNodeType
}
```

The resulting union type is an enum with a variant for each node type in the union and implementing:
- the `AstNode` trait
- the `Debug` and `Display` traits
- the `Clone`, `PartialEq`, `Eq` and `Hash` traits with derive
- the `From` trait for each variants
- the `Into` trait for `SyntaxNode` and `SyntaxElement`

The macro is designed to fail to compile when the union would contain node types from different languages, as the `AstNode` trait must have a single `Language` associated type. I'm not really sure if and how this is a restriction we may lift in the future.

## Test Plan

This macro is now used to implement the `JsAnyBinaryLikeExpression` and `JsAnyBinaryLikeLeftExpression` types in the formatting of binary-like expressions

### DIFF
--- a/crates/rome_rowan/src/macros.rs
+++ b/crates/rome_rowan/src/macros.rs
@@ -1,0 +1,116 @@
+use crate::{AstNode, Language};
+
+/// Declares a custom union AstNode type with an ungram-like syntax
+///
+/// # Example
+///
+/// ```ignore
+/// declare_union! {
+///     /// Matches an if statement or a conditional expression
+///     pub(crate) JsAnyConditional = JsIfStatement | JsConditionalExpression
+/// }
+/// ```
+#[macro_export]
+macro_rules! declare_union {
+    ( $( #[$attr:meta] )* $vis:vis $name:ident = $( $variant:ident )|* ) => {
+        $( #[$attr] )*
+        #[allow(clippy::enum_variant_names)]
+        #[derive(Clone, PartialEq, Eq, Hash)]
+        $vis enum $name {
+            $( $variant($variant), )*
+        }
+
+        impl AstNode for $name {
+            type Language = <( $( $variant, )* ) as $crate::macros::UnionLanguage>::Language;
+
+            fn can_cast(kind: <Self::Language as $crate::Language>::Kind) -> bool {
+                $( $variant::can_cast(kind) )||*
+            }
+
+            fn cast(syntax: $crate::SyntaxNode<Self::Language>) -> Option<Self>
+            where
+                Self: Sized,
+            {
+                $( if $variant::can_cast(syntax.kind()) {
+                    return Some(Self::$variant($variant::unwrap_cast(syntax)));
+                } )*
+
+                None
+            }
+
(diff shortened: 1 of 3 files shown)


crates/rome_rowan/src/macros.rs
```diff
@@ -0,0 +1,116 @@
...
+/// # Example
+///
+/// ```ignore
+/// declare_union! {
+///     /// Matches an if statement or a conditional expression
+///     pub(crate) JsAnyConditional = JsIfStatement | JsConditionalExpression
+/// }
+/// ```
+#[macro_export]
+macro_rules! declare_union {
```
ematipico, line 14: nit: could we call `declare_language_union` or `declare_node_union` (this is maybe better, because it applies `AstNode`)? 

The name, right now, is so generic that we could misuse it.

crates/rome_rowan/src/macros.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/// This trait is implemented for tuples of AstNode types of size 1 to 9 if all
+/// node types share the same associated language (which is then aliased as the
+/// `Language` associated type on [UnionLanguage] itself)
+pub trait UnionLanguage {
+    type Language: Language;
+}
+
+macro_rules! impl_union_language {
```
ematipico, line 97: Sorry I can't really understand the purpose of this macro. Could you explain please?

ematipico (approved):

crates/rome_rowan/src/macros.rs
leops: This macro is an implementation detail of `declare_union` (it's not exported from the `macros` module) aimed at making it easier to implement the `UnionLanguage` trait. This trait is implemented for tuples of various arity, but because tuples are heterogeneous the implementation needs to declare one generic type for each element of the tuple. This in turn means it's impossible to implement the trait once for all tuples size, instead implementing the trait would look like this:
```rust
impl<T1> UnionLanguage for (T1,) { ... }
impl<T1, T2> UnionLanguage for (T1, T2) { ... }
impl<T1, T2, T3> UnionLanguage for (T1, T2, T3) { ... }
```
All these impl blocks share the same code with the only difference being the number of generic types, so it was simpler to write that code in a macro so it can easily be repeated for various numbers of generic arguments

crates/rome_rowan/src/macros.rs
leops: I'll change it to `declare_node_union`, I also think it's the better option since the type being declared and all the types it contains must be AST nodes

xunilrj (approved):

PR #2692, opened 2022-06-09
feat(rome_analyze): improve the diagnostics builder in the analyzer

leops:
## Summary

This PR replicates the builder pattern implemented on `rome_diagnostic::Diagnostic` for the `RuleDiagnostic` type used internally by the analyzer rules. It should expose all the capabilities of normal diagnostics (including multiple labels and footers) with the notable exception of:

- `file_id`: at the moment diagnostics are considered to belong to the file currently being analyzed
- `code`: this property has its value set to the name of the rule that emitted the diagnostic
- `suggestions`: rules that emit quick fix suggestions are expected to do so by emitting a code action

Additionally, it adds a new optional `summary` field to diagnostics containing a plain-text description of the issue, aimed at rendering contexts that don't support rich formatting (such as diagnostic popovers in editors)

## Test Plan

Existing rules have been modified to use the new methods, and snapshots were updated accordingly

### DIFF
--- a/crates/rome_analyze/src/analyzers/no_double_equals.rs
+++ b/crates/rome_analyze/src/analyzers/no_double_equals.rs
@@ -37,16 +37,16 @@ impl Rule for NoDoubleEquals {
         let suggestion = if op.kind() == EQ2 { "===" } else { "!==" };
 
         Some(
-            RuleDiagnostic::warning(markup! {
+            RuleDiagnostic::warning(op.text_trimmed_range(),markup! {
                 "Use "<Emphasis>{suggestion}</Emphasis>" instead of "<Emphasis>{text_trimmed}</Emphasis>
             })
-            .primary(op.text_trimmed_range(), markup! {
-                <Emphasis>{text_trimmed}</Emphasis>" is only allowed when comparing against null"
+            .primary( markup! {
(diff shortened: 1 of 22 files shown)


MichaReiser (approved):

crates/rome_analyze/src/analyzers/no_double_equals.rs
```diff
@@ -33,26 +33,32 @@ impl Rule for NoDoubleEquals {
...
+            RuleDiagnostic::warning(markup! {
+                "Use "<Emphasis>{suggestion}</Emphasis>" instead of "<Emphasis>{text_trimmed}</Emphasis>
+            })
+            .primary(op.text_trimmed_range(), markup! {
+                <Emphasis>{text_trimmed}</Emphasis>" is only allowed when comparing against null"
+            })
+            .footer_note(markup! {
+                "Using "<Emphasis>{suggestion}</Emphasis>" may be unsafe if you are relying on type coercion"
+            })
+            .summary(format!("Use {suggestion} instead of {text_trimmed}\n{text_trimmed} is only allowed when comparing against null"))
```
ematipico, line 49:
```suggestion
            .summary(format!("Use {suggestion} instead of {text_trimmed}.\n{text_trimmed} is only allowed when comparing against null"))
```

PR #2694, opened 2022-06-10
fix(rome_analyze):  handle find_diff_range in case tokens were only removed

leops:
## Summary

This PR fixes #2693 by ensuring the start and end of the range returned by `find_diff_range` do not cross. This can happen if the only difference between the previous and new syntax trees are removed tokens: the results of the linear search for the start and end of the range will cross as the new range starts at the start of the new different token from the start of the tree, and ends at the end of the first different token from the end of the tree. In the case of a removal these will be the first token after the removal, and last token before the removal respectively.

## Test Plan

I added a new `diff_range_remove` test to verify this case is handled correctly

PR #2697, opened 2022-06-10
fix(rome_analyze): prevent the range end search of find_diff_range from crossing the start of the range

### LINKED ISSUE #2696: 🐛 diff_print seems still not right
### Environment information

```bash
linux
```

### What happened?

Here is my console for every replaced root
```bash
const test = { debugger: 1 }; test.debugger;

if (foo) ;

debugger;

function test() {
  let a = 3;
  debugger;
}
 --------------------- 
const test = { debugger: 1 }; test.debugger;

if (foo) debugger

function test() {
  let a = 3;
  debugger;
}
 --------------------- 
const test = { debugger: 1 }; test.debugger;

if (foo) debugger

debugger;

function test() {
  let a = 3;
}
 --------------------- 
```

and here is the diff snapshot

```bash
warning[noDebugger]: This is an unexpected use of the debugger statement.
  ┌─ noDebugger.js:9:3
  │
9 │   debugger;
  │   --------- This is an unexpected use of the debugger statement.

Remove debugger statement
    | @@ -5,6 +5,5 @@
4 4 |   debugger;
5 5 |   
6 6 |   function test() {
7   | -   let a = 3;
8   | -   debugger;
  7 | +   let a = 3;;
9 8 |   }


```

```bash
+ let a = 3;;
```
the extra `;` should not be here, since replaced root does not have the double semi colon

### Expected result

diff in snapshot should print the correct result

### Code of Conduct

leops:
## Summary

In #2694 I implemented a simplified solution (clamping the resulting range) to prevent the end of the diff range from crossing the start of the range when calculating the diff of a token removal operation. Unfortunately this solution isn't entirely correct and can result in mismatches between the two ranges (due to the clamping being applied to the new range independently from the original range)

With this PR I implemented a fix that completely removes the root cause of the issue by preventing the search for the endpoint of the modified range from crossing the start of the range in the first place, eliminating the need for the clamping operation on the resulting index

Fixes #2696 

## Test Plan

I also fixed the `diff_range_remove` test case, it turns out I made a mistake when calculating the values the range returned by `find_diff_range` were supposed to have, which in turn led to implementing an incorrect fix

### DIFF
--- a/crates/rome_analyze/src/signals.rs
+++ b/crates/rome_analyze/src/signals.rs
@@ -164,37 +164,42 @@ where
 
     let prev_tokens = prev.descendants_tokens(Direction::Prev);
     let next_tokens = next.descendants_tokens(Direction::Prev);
-    let mut tokens = prev_tokens.zip(next_tokens);
+    let tokens = prev_tokens.zip(next_tokens);
 
-    let range_end = tokens.find_map(|(prev_token, next_token)| {
-        // This compares the texts instead of the tokens themselves, since the
-        // implementation of PartialEq for SyntaxToken compares the text offset
-        // of the tokens (which may be different since we're starting from the
-        // end of the file, after the edited section)
-        // It should still be rather efficient though as identical tokens will
-        // reuse the same underlying green node after an edit, so the equality
-        // check can stop at doing a pointer + length equality without having
-        // to actually check the content of the string
-        if prev_token.text() != next_token.text() {
-            Some((prev_token.text_range().end(), next_token.text_range().end()))
-        } else {
-            None
-        }
-    });
+    let range_end = tokens
+        .take_while(|(prev_token, next_token)| {
+            let prev_range = prev_token.text_range();
+            let next_range = next_token.text_range();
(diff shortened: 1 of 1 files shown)


crates/rome_analyze/src/signals.rs
```diff
@@ -164,37 +164,42 @@ where
...
+
+            // This compares the texts instead of the tokens themselves, since the
+            // implementation of PartialEq for SyntaxToken compares the text offset
+            // of the tokens (which may be different since we're starting from the
+            // end of the file, after the edited section)
+            // It should still be rather efficient though as identical tokens will
+            // reuse the same underlying green node after an edit, so the equality
+            // check can stop at doing a pointer + length equality without having
+            // to actually check the content of the string
+            prev_token.text() == next_token.text()
```
ematipico, line 188: This might be a dump question. Can't we compare the hash of the tokens? 

EDIT: seems not, as we use the `TextSize` as key for the hash

ematipico (approved):

Issue #2410, opened 2022-04-13
Assignment of array/object literals


MichaReiser:
Prettier keeps the opening `[` or `{` on the same line as the assignment expression when it exceeds the line width. Rome formats the whole array or object literal on a new line.

## Input

```ts
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder.arrayOfNumbers = [1, 2, 3, 4, 5];
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder.arrayOfNumbers = { a: 10 };
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder = bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawderArrayNumbes = { a: 10 };
```

## Prettier

```ts
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder.arrayOfNumbers = [
	1, 2, 3, 4, 5,
];
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder.arrayOfNumbers = {
	a: 10,
};
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder =
	bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawderArrayNumbes = {
		a: 10,
	};
```

## Rome

```ts
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder.arrayOfNumbers =
	[1, 2, 3, 4, 5];
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder.arrayOfNumbers =
	{ a: 10 };
bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawder =
	bifornCringerMoshedPerplex.bifornCringerMoshedPerplexSawderArrayNumbes =
		{ a: 10 };
```

## Expected

Rome's formatting should match Prettiers formatting by putting the opening `{` and `[` tokens on the same line as the `=` token.

[...]

Issue #2423, opened 2022-04-13
Variable declarator formatting


MichaReiser:
Rome has no special formatting for variable declarations but Prettier formats them the same way as assignment expressions. 

## Examples

A few examples from the latest Prettier comparison report

* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentissue-10218js
* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentissue-1966js
* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentissue-4094js
* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentissue-7091js
* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentissue-8218js
* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentlone-argjs
* https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentunaryjs

## Expected

Rome's output to match Prettier's

Dominionys:
Hi, 
Could I also take this issue?🙏🏽 
Because it's relative with object member property.

MichaReiser:
Sure thing. I assigned it to you

denbezrukov:
Need to check because Rome and Prettier have different IR for these cases:
`const _id1 = data.createTestMessageWithAReallyLongName.someVeryLongProperty.thisIsAlsoALongProperty._id;`
`const { imStore, showChat, customerServiceAccount } = store[config.reduxStoreName];`
`const pendingIndicators = shield.alarmGeneratorConfiguration.getPendingVersionColumnValues;`
`const pendingIndicatorz = shield.alarmGeneratorConfiguration.getPendingVersionColumnValues();`
`const bifornCringerMoshedPerplexSawderGlyphsHa = someBigFunctionName("foo")(
  "bar",
);`

https://gist.github.com/MichaReiser/77799fe4363bcc48838a159e581d8229#jsassignmentunaryjs

- I guess that it's because of  `isPoorlyBreakableMemberOrCallChain` https://github.com/prettier/prettier/blob/be9d63e543b6ac28aa6eee7ab8dc45849cd8f0b3/src/language-js/print/assignment.js#L215

denbezrukov:
@ematipico @MichaReiser 
[...]

Issue #2416, opened 2022-04-13
Nested assignments


MichaReiser:
[...]

PR #2698, opened 2022-06-11
feat(rome_js_formatter): JsAnyAssignmentLike node union for JsPropertyObjectMember and JsAssignmentExpression

denbezrukov:
## Summary
This PR fixes https://github.com/rome/tools/issues/2410
This is the initial draft for `JsAnyAssignmentLike`. 
I'd like to propose `JsAnyAssignmentLike` node union to handle different cases. 
Prettier uses [printAssignment](https://github.com/prettier/prettier/blob/be9d63e543b6ac28aa6eee7ab8dc45849cd8f0b3/src/language-js/print/assignment.js#L25) for: 

- `ObjectProperty` 
https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/printer-estree.js#L388
- `AssignmentExpression` https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/assignment.js#L79
- `VariableDeclarator` https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/assignment.js#L90
- `TSTypeAliasDeclaration`
https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/type-annotation.js#L88
- `ClassAccessorProperty`, `TSAbstractPropertyDefinition`
https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/class.js#L234

## Open questions
I'm not sure about the name `JsAnyAssignmentLike` because some cases aren't assignment.

## Test Plan

Add new spec tests cases for `JsAssignmentExpression`.


crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -0,0 +1,212 @@
...
+const MIN_OVERLAP_FOR_BREAK: u8 = 3;
+
+impl JsAnyAssignmentLike {
+    fn write_left(&self, f: &mut JsFormatter) -> FormatResult<bool> {
+        match self {
+            JsAnyAssignmentLike::JsPropertyObjectMember(property) => {
+                let width = write_member_name(&property.name()?, f)?;
+                let text_width_for_break =
+                    (f.context().tab_width() + MIN_OVERLAP_FOR_BREAK) as usize;
+                Ok(width < text_width_for_break)
```
denbezrukov, line 120: Prettier checks short name only for `ObjectMember`. 
https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/assignment.js#L413

denbezrukov:
If we choose this way we'll need to add more variants for layout: 
- https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/assignment.js#L37-L74

denbezrukov:
It seems that need to update prettier because [2.6.0 release](https://prettier.io/blog/2022/03/16/2.6.0.html#javascript) treats comment different.

### Input
```
f4 = // Comment 
() => {};
f4 =
	// Comment
	() => {};
```

[2.6.0 Playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzALAAgLyYPR6YDCEAtqQjJgDpQAUAlDgHybAC+A3LRjrdTALEyFWPxiMWbLiAA0ICAAcYAS2gBnZKACGAJ10QA7gAU9CTSm0AbQ9oCem+QCNd2sAGs4MAMraKAGRUoOGRUa3U4Z1cPL29FNyCAc2QYXQBXSJA4Uic4ABM8-P9tKES07US4ADEIXVJtGFVS5BBtNJgIORAACxhSKwB1bpV4dXiwOG9zEZUANxG7FrB1RxAgiN0YY1dE+tDwzIArdQAPbySrOABFNIh4fasI+XjdDZanbVyrLsVdIJgBio8jBush0AAGZ4GCIDVyKFq-OAbWYheQAR1u8G2SgsrXUAFpgvl8l1dHAMSoydsKnskGFHpkIqQVCl0oyLtdMSE6Qd5DBPoDgaCkAAmPmuFRWJIkcjaFpIgCsXTSEQAKp8LPSniBZhkAJJQQqwbxgP7KACChu8MDslweEXY7CAA)

```
f4 = // Comment
  () => {};
f4 =
  // Comment
  () => {};

```

[Rome playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#ZgA0ACAAPQAgAC8ALwAgAEMAbwBtAG0AZQBuAHQAIAAKACgAKQAgAD0APgAgAHsAfQA7AAoAZgA0ACAAPQAKAAkALwAvACAAQwBvAG0AbQBlAG4AdAAKAAkAKAApACAAPQA+ACAAewB9ADsA)
```

f4 = () => {}; // Comment
f4 =
	// Comment
	() => {};
```

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -0,0 +1,216 @@
...
+        } else if is_never_break_after_operator(&right)? {
+            Ok(AssignmentLikeLayout::NeverBreakAfterOperator)
+        } else {
+            Ok(AssignmentLikeLayout::Fluid)
+        }
+    }
+}
+
+pub(crate) fn is_break_after_operator(right: &JsAnyExpression) -> SyntaxResult<bool> {
+    if has_leading_newline(right.syntax()) && right.syntax().has_leading_comments() {
```
denbezrukov, line 169: https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/assignment.js#L126

ematipico:
> I'm not sure about the name JsAnyAssignmentLike because some cases aren't assignment.

I think that's fine, as long as we document the data structure and explain what kind of situations it handles.

denbezrukov:
@ematipico Could you please help me? 🙏🏽
I saw https://github.com/rome/tools/pull/2721.
Should I close this PR?

ematipico:
> @ematipico Could you help me please? 🙏🏽
> I saw https://github.com/rome/tools/pull/2721.
> Should I close this PR?

Oh, please don't. I would actually close mine. The problem is that I didn't see what issues this PR is closing.

Could you please update the description? This is really important to us, it avoids double work.

denbezrukov:
> Could you please update the description? This is really important to us, it avoids double work.

I'm sorry for that. I didn't check the issues before opening PR. My bad. 😓

>  I would actually close mine. 

I'm going to copy `has_new_line_before_comment`  and spec test cases (:

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -0,0 +1,234 @@
...
+use crate::utils::object::write_member_name;
+use crate::utils::JsAnyBinaryLikeExpression;
+use rome_formatter::{format_args, write};
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyLiteralExpression, JsAssignmentExpression, JsPropertyObjectMember,
+    JsSyntaxNode,
+};
+use rome_rowan::{declare_node_union, AstNode, SyntaxResult};
+
+pub(crate) fn format_assignment_like(
```
ematipico, line 12: The result of this function is `FormatResult`, which means that we want to call it `write_assignment_like`. 

Generally, this is the rule:
- returns a `impl Format<JsFormatContext>` -> `format_*`
- returns a `FormatResult` -> `write_*`

crates/rome_js_formatter/src/js/objects/property_object_member.rs
```diff
@@ -1,61 +1,11 @@
...
-                    ]
-                }
-                PropertyObjectMemberLayout::NeverBreakAfterColon => {
-                    write![f, [space_token(), value.format(),]]
-                }
-            }
-        });
-
-        write!(f, [group_elements(&format_content)])
+        format_assignment_like(JsAnyAssignmentLike::JsPropertyObjectMember(node.clone()), f)
```
ematipico, line 9: The macro also implements the `From` for each variant, which means that we can do things like:
- `JsAnyAssignmentLike::from(node.clone())`
- `format_assignment_like(node.clone().into())`

We should use one of these options

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico, line 12: A neat implementation would be to implement `Format` for `JsAnyAssignmentLike`. Here's an example:

https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/utils/binary_like_expression.rs#L856-L867

And then, inside the callers, we can write:

```rs
// using the current code
write!(f, [ JsAnyAssignmentLike::JsPropertyObjectMember(node.clone()).format(f) ])
```

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: True, I noticed that too yesterday

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -0,0 +1,234 @@
...
+    assignment_like: JsAnyAssignmentLike,
+    f: &mut JsFormatter,
+) -> FormatResult<()> {
+    let right = assignment_like.right()?;
+    let format_content = format_with(|f| {
+        let is_left_short = assignment_like.write_left(f)?;
```
ematipico, line 18: Could you add a comment here about the  `is_left_short` please? https://github.com/rome/tools/pull/2721/files#diff-3ab836cde79ac685d6a0e0634875f023483cad5c0adfc79d6866e13aa752b19bR93-R95

I think we should track the information as a comment, we would risk to lose it inside the PR comment

ematipico (commented):
Thank you for the great contribution! 

About the next steps, here's what I had in mind. As you explained, prettier uses this way of formatting for many things, and eventually we will need to check the "left" and "right" part of an assignment like.

Which means that "left" and "right" should be grouped inside another unions, because the "left" part of an assignment like might differ based on what we are formatting. My idea was to create a `LeftAssignmentLike`:

```rs
declare_node_union! {
    pub(crate) LeftAssignmentLike = JsAnyAssignmentPattern | JsAnyObjectMemberName
}
```

If you're OK which this plan, I will work on https://github.com/rome/tools/issues/2416 and you can choose another one (maybe `VariableDeclarator`?). Feel free to comment in the issue and I will assign that to you

ematipico:
> If we choose this approach we'll need to add more variants for layout:
> 
>     * [prettier/prettier@`2316e2f`/src/language-js/print/assignment.js#L37-L74](https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/print/assignment.js#L37-L74)

I can definitely see that. And we should add variants while we implement other assignments. I already started implementing `Chain` and `ChainTail`

denbezrukov:
> If you're OK which this plan, I will work on #2416 and you can choose another one (maybe `VariableDeclarator`?). Feel free to comment in the issue and I will assign that to you

It sounds great. 👍👍👍

 `VariableDeclarator` - https://github.com/rome/tools/issues/2423 ? I've already taken it. After this PR I'll work on it.

denbezrukov:
> I can definitely see that. And we should add variants while we implement other assignments. I already started implementing `Chain` and `ChainTail`

I noticed that to implement this variant we need to know `layout` outside `format_with` in `format_assignment_like` function.
Because chain variants don't have outer group.  Do you have any idea how to get this?

- https://github.com/prettier/prettier/blob/877ae8ec16369dc0cf79ee36019bfdce40429eae/src/language-js/print/assignment.js#L63-L70

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Done

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Done, It looks cleaner now.

May be we can do it for `format_binary_like_expression` also?

crates/rome_js_formatter/src/js/objects/property_object_member.rs
denbezrukov: After implement `Format` trait for `JsAnyAssignmentLike`:

`write![f, [JsAnyAssignmentLike::from(node.clone())]]`

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: Definitely! Maybe in a separate PR but it's surely something we can do :)

denbezrukov: Done.

After implementation `Format` trait for `JsAnyAssignmentLike` this rule makes more sense to me because now it looks like std `Display`.

```

impl Format<JsFormatContext> for JsAnyAssignmentLike {
    fn fmt(&self, f: &mut JsFormatter) -> FormatResult<()> {
        write_assignment_like(self, f)
    }
}

```
```

impl Display for String {
    fn fmt(&self, fmt: &mut Formatter) -> io::Result<()> {
        fmt.write_str(self)
    }
}
```


Thank you for your feedback!

ematipico (approved):
Great! Thank you very much! This contribution is great! We can rebase and merge it :)

ematipico:
> > I can definitely see that. And we should add variants while we implement other assignments. I already started implementing `Chain` and `ChainTail`
> 
> I noticed that to implement this variant we need to know `layout` outside `format_with` in `format_assignment_like` function. Because chain variants don't have outer group. Do you have any idea how to get this?
> 
>     * [prettier/prettier@`877ae8e`/src/language-js/print/assignment.js#L63-L70](https://github.com/prettier/prettier/blob/877ae8ec16369dc0cf79ee36019bfdce40429eae/src/language-js/print/assignment.js#L63-L70)

That's because we assumed that each variant had a group, and we should not. I think the best implementation is:

```rs
fn write_assignment_like() -> FormatResult<()> {
	match layout {}
}
```

And each variant `write!`

ematipico:
I solved like this:

```rs
        match self {
            JsAnyAssignmentLike::JsPropertyObjectMember(_) => {
                write!(f, [group_elements(&format_content)])
            }
            JsAnyAssignmentLike::JsAssignmentExpression(_) => {
                write!(f, [&format_content])
            }
        }
```

denbezrukov:
> I solved like this:
> 
> ```rust
>         match self {
>             JsAnyAssignmentLike::JsPropertyObjectMember(_) => {
>                 write!(f, [group_elements(&format_content)])
>             }
>             JsAnyAssignmentLike::JsAssignmentExpression(_) => {
>                 write!(f, [&format_content])
>             }
>         }
> ```
> 
> At the very end of the function

Hmm, but there are cases then assignment expression has `group` in [prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDABAXnQbQIwA0ATAQMwEAsAugNwgEgQAOMAltAM7KioBOvEAO4AFPgi4pUAG0GoAnlwYAjXqjABrODADKqALZwAMqyhxkAM2kc4y1Rq3amakwHNkMXgFcbIOHqVwACaBQYaoUC6eqC5wAGIQvHqoMGwRyGieMBD0IAAWMHpSAOq5rPAcTmBw2uJlrABuZXLpYByKICbWvDDCqi5JFlY+AFYcAB7arlJwAIqeEPCDUtYMTrxd6UqoAVI5TLwmMEWsgTC5yAAcAAyrAtZFqkzp+3Bd9WYMAI7z8L3MEmgOABaUxBII5Xhwb6sSG9aIDJCWZY+ax6VjuLwoqazH5mRFDBgwbbHU7nJAkEAeVCsKSuADCED0CN8HAArDlPNYACrbCRIlYgereACSUBCsG0YAOLAAgmLtDA5NMltYAL6qoA). It seems that only chain layout variants should be without group. Is it possible to write in temporary buffer before wrapping in a group?

ematipico:
Sorry, I guess I assumed too many things here 😅 

I **had** to move all the code inside `write_assignment_like` inside the `format` function in order to make things work. This is actually way better:
- the implementation needs to have access to the [parent](https://github.com/prettier/prettier/blob/main/src/language-js/print/assignment.js#L106-L113). In this example `path` is the parent node. This way, we can access to the parent without too many traverse.
- most of the functions we have can be moved inside `impl JsAnyAssignmentLike` 
- the only small issue is `is_break_after_operator` because it's used somewhere else

ematipico:
@denbezrukov Please check https://github.com/rome/tools/pull/2728

PR #2700, opened 2022-06-13
refactor(rome_analyze): move the JS-specific code to rome_js_analyze

leops:
## Summary

This PR splits out the analyzer codebase between the generic analysis infrastructure in `rome_analyze`, and the implementation of JS-specific rules in `rome_js_analyze`

## Test Plan

This change is mostly renaming files or moving items between crates with little to no actual code changes, so the code relying on the analyzer should continue building and passing the same tests as before.


ematipico:
How would this infrastructure work when/if we will have rules that touch different languages at the same time? E.g. JSX and HTML

crates/rome_analyze/src/lib.rs
```diff
@@ -57,72 +58,47 @@ pub enum Never {}
...
-
-        if let ControlFlow::Break(b) = registry.analyze(file_id, root, node, &mut callback) {
-            return Some(b);
-        }
-    }
-
-    None
+pub struct Analyzer<L: Language> {
+    pub registry: RuleRegistry<L>,
+    pub has_suppressions: fn(&SyntaxNode<L>) -> bool,
```
MichaReiser, line 63: Nit: I prefer to have pub `new` functions for structs that get constructed by other crates to make it easier to reason about which fields may (or may not) change after construction.

crates/rome_js_analyze/src/registry.rs
```diff
@@ -0,0 +1,36 @@
+//! Generated file, do not edit by hand, see `xtask/codegen`
```
MichaReiser, line 1: This file is still showing up in review mode. Would you mind checking why this is the case?

MichaReiser (approved):

crates/rome_js_analyze/src/registry.rs
leops: I was thinking this was due to having fixed the `.gitattributes` file in the second commit after having already modified the generated files, but they're still showing up in the diff after I squashed the changes in the rebase so it might just be that git doesn't support the `{analyzers,assists,registry}` glob pattern

crates/rome_analyze/src/lib.rs
```diff
@@ -29,6 +23,13 @@ pub struct AnalysisFilter<'a> {
     pub range: Option<TextRange>,
 }
 
+impl AnalysisFilter<'_> {
+    pub fn match_rule<R: Rule>(&self) -> bool {
```
ematipico, line 27: I think we should add some documentation. The name of the function doesn't render justice to what the function does

crates/rome_analyze/src/lib.rs
```diff
@@ -57,72 +58,56 @@ pub enum Never {}
...
-                iter.skip_subtree();
-                continue;
-            }
-        }
-
-        if has_suppressions_category(SuppressionCategory::Lint, &node) {
-            iter.skip_subtree();
-            continue;
-        }
+pub struct Analyzer<L: Language> {
```
ematipico, line 61: Why `Analyzer` and not `Linter`. Does the name entail that it's used for code assists too? If so, maybe some doc should help

ematipico (approved):

crates/rome_analyze/src/lib.rs
leops: Yes the analyzer is intended as an executor for a set of rules (contained in a rule registry), these rules can query the syntax tree (and possibly other data structures in the future like the semantic model) and emit "signals" in response to a query match, with a signal being a generic object containing a diagnostic, a code action or both. It's used to support both linting and code assists at the moment, but all of this is purposefully left very generic in case it may be reused for other workloads in the future

PR #2701, opened 2022-06-13
feature(rome_js_formatter): Inline comment formatting

### LINKED ISSUE #2447: Spacing around block comments

MichaReiser:
This PR improves the formatting of inline comments (`/* comment */`) so that it doesn't insert any space after a *group start* token (any of `(`, `[`, `{`) nor before a group end token (any of `)`, `}`, `]`, `,`, `;`, `EOF`, `.`). 

The implementation introduces new helpers to format tokens:

* `format_inserted`: To format a token that isn't present in the source code. For example, when adding parentheses around an expression. Use this over `token`
* `format_replaced`: To replace the content of a token with some other content
* `format_removed`: To remove a token but still format the tokens trivia
* `format_only_if_breaks`: A combination of `format_replaced` and `format_removed`. Used in the cases where a token should be removed if its enclosing group doesn't break but otherwise gets formatted as is (or with slightly different content). Commonly used when formatting the trailing separator of a list. 

The PR introduces two new formatter states:

* The kind of the last formatted token. Necessary to know whether a leading space must be inserted before the first leading inline comment or not. 
* Whether the last token's last trivia is an inline comment in which case the next token must insert a leading space ONLY IF it is not a group end token. This is necessary to avoid a token looking ahead to guess what next token (which wouldn't work in case of inserted tokens)

This PR moves the comment formatting to `rome_formatter` because there's now nothing JS-specific about formatting comments. 

Closes #2447

## Differences to Prettier's formatting'

I decided to diverge from Prettier for the following case:

```js
// Input
function name/*comment*/   () {
}

// Prettier
function name /*comment*/() {
}

// Rome
function name /*comment*/ () {
}
```

Prettier omits whitespace for inline comments before a `(` token but it doesn't do so for `[` or `{` in which case prettier prints a single whitespace too. In my view, inline comments before start group tokens should either all be separated from the token by a whitespace or the whitespace is omitted for all of them. That's why I didn't align Rome with Prettier's behaviour in this case. 

## Test Plan

`cargo test`. 

Reviewed the changed snapshot, added a few new tests

Increases compatibility by ~0.6% (from 72.65% to 73.25%)


crates/rome_js_formatter/src/lib.rs
```diff
@@ -298,6 +295,45 @@ pub fn format_sub_tree(context: JsFormatContext, root: &JsSyntaxNode) -> FormatR
     rome_formatter::format_sub_tree(context, &root.format())
 }
 
+#[derive(Copy, Clone, Debug)]
+pub struct JsCommentStyle;
+
+impl CommentStyle for JsCommentStyle {
```
MichaReiser, line 301: Ideally, it would be possible that `rome_formatter` defines a new `Language` trait that extends `rome_rowan::Language` and adds these methods to it. Unfortunately, this isn't possible without setup because `rome_js_formatter` isn't able to implement that trait because of the orphan rule: Both the trait and the `JsLanguage` struct are foreign to the crate.

leops (approved):

crates/rome_formatter/src/comments.rs
```diff
@@ -0,0 +1,70 @@
+use rome_rowan::{Language, SyntaxTriviaPieceComments};
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum CommentKind {
+    /// An inline comment that can appear between any two tokens and doesn't contain any line breaks.
+    /// For example, a `/* test */` comment in JavaScript.
+    InlineBlock,
+
+    /// A block comment that can appear between any two tokens and contains at least one line break.
+    /// For example, a `/* first line\nmore content on the second line */` comment in JavaScript.
```
ematipico, line 10: nit: we could take advantage of markdown and showing an actual JS snippet with the comment :)

crates/rome_formatter/src/comments.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    pub const fn is_block(&self) -> bool {
+        matches!(self, CommentKind::Block)
+    }
+
+    pub const fn is_inline_block(&self) -> bool {
+        matches!(self, CommentKind::InlineBlock)
+    }
+
+    pub const fn is_inline(&self) -> bool {
```
ematipico, line 67: Could we add a small doc here please? I think we should explain what "inline" means  to us and why `::Block` variant falls under this.

crates/rome_formatter/src/lib.rs
```diff
@@ -1064,6 +1068,13 @@ impl<L: Language, Context> Format<Context> for SyntaxTriviaPieceComments<L> {
 pub struct FormatState<Context> {
     context: Context,
     group_id_builder: UniqueGroupIdBuilder,
+
+    /// `true` if there's a trailing inline comment before the next token (including its trivia).
```
ematipico, line 1072: Saying `including its trivia` is misleading because a comment **is a trivia**, maybe we should be more descriptive and list what other trivias we mean?

crates/rome_formatter/src/lib.rs
```diff
@@ -1087,11 +1100,39 @@ impl<Context> FormatState<Context> {
...
+        self.trailing_inline_comment = has_comment;
+    }
+
+    /// Returns the kind of the last formatted token.
+    pub fn last_token_kind(&self) -> Option<LastTokenKind> {
+        self.last_token_kind
+    }
+
+    /// Sets the kind of the last formatted token.
+    pub fn set_last_token_kind<Kind: SyntaxKind + 'static>(&mut self, kind: Kind) {
```
ematipico, line 1126: Why are you assigning a `'static` lifetime?

crates/rome_formatter/src/lib.rs
```diff
@@ -1087,11 +1100,39 @@ impl<Context> FormatState<Context> {
...
             context,
             group_id_builder: Default::default(),
+            trailing_inline_comment: false,
+            last_token_kind: None,
             #[cfg(debug_assertions)]
             printed_tokens: Default::default(),
         }
     }
 
+    /// Returns true if there's a trailing inline comment before the next token or the next token's comments.
```
ematipico, line 1110:
```suggestion
    /// Returns `true` if there's a trailing inline comment before the next token or the next token's comments.
```

crates/rome_formatter/src/lib.rs
```diff
@@ -1132,3 +1192,44 @@ impl<Context> FormatState<Context> {
...
+}
+
+pub struct FormatStateSnapshot {
+    trailing_inline_comment: bool,
+    last_token_kind: Option<LastTokenKind>,
+    #[cfg(debug_assertions)]
+    printed_tokens: PrintedTokens,
+}
+
+pub trait CommentStyle: Copy {
```
ematipico, line 1219: Can we add few lines of doc about this trait? Like, what's it for

crates/rome_formatter/src/lib.rs
```diff
@@ -1132,3 +1192,44 @@ impl<Context> FormatState<Context> {
...
+    type Language: Language;
+
+    /// Returns the kind of the comment
+    fn get_comment_kind(&self, comment: &SyntaxTriviaPieceComments<Self::Language>) -> CommentKind;
+
+    /// Returns `true` if a token with the passed `kind` marks the start of a group. Common group tokens are
```
ematipico, line 1225:
```suggestion
    /// Returns `true` if a token with the passed `kind` marks the start of a group. Common group tokens are:
```

ematipico (commented):
Would it be possible to add new tests to show us new changes? This is more needed, I think, especially because we decided to diverge.

crates/rome_formatter/src/token.rs
```diff
@@ -0,0 +1,682 @@
...
+        }
+    }
+}
+
+impl<S, C> Format<C> for FormatInserted<S>
+where
+    S: CommentStyle + 'static,
+{
+    fn fmt(&self, f: &mut Formatter<C>) -> FormatResult<()> {
+        let has_trailing_inline_comment = f.state().has_trailing_inline_comment();
```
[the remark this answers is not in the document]
MichaReiser: This is good feedback and you're right, it's the previous token. The logic behind this is non-trivial. What it tracks is whether the last written comment requires a space before whatever content gets written next (that can be a comment, skipped token, or a regular token). Let's see if I can improve the documentation somehow.

MichaReiser:
> Would it be possible to add new tests to show us the new changes? This is more needed, I think, especially because we decided to diverge.

I added a few more tests to the [`comments`](https://github.com/rome/tools/blob/390836e1957e13ed49da5402c5dfd02d91b62d74/crates/rome_js_formatter/tests/specs/js/module/comments.js) test but good call out. Let me add some tests that explicitly verify the cases where we diverge from prettier's formatting.

crates/rome_formatter/src/lib.rs
MichaReiser: A `'static` lifetime is necessary by `Typeid`. My guess is that this is because `Typeid` only encodes the type but without its lifetime. Thus, comparing two types with different lifetimes would return `true` even though these are (from a compile-time perspective), two different types.

PR #2702, opened 2022-06-13
feat(rome_analyze): noImplicitBoolean

### LINKED ISSUE #2699: 📎 noImplicitBoolean
### Description

rome_analyze rule noImplicitBoolean

IWANABETHATGUY:
## Summary
1. Resolved #2699  
2. https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/jsx/noImplicitBoolean.md

## Test Plan
1. cargo test


crates/rome_analyze/src/analyzers/no_implicit_boolean.rs
```diff
@@ -0,0 +1,105 @@
...
+        );
+        let next_attr = make::jsx_attribute(next_name).with_initializer(
+            make::jsx_attribute_initializer_clause(
+                make::token(T![=]),
+                JsxAnyAttributeValue::JsxExpressionAttributeValue(attr_value),
+            ),
+        );
+        let next_attr = next_attr.build();
+
+        let root = root.replace_node_discard_trivia(n.clone(), next_attr)?;
```
leops, line 97: Using `replace_node` here would remove the need to explicitly clone the trailing trivia from the name node on the `}` token

IWANABETHATGUY: It seems I can't use replace_node without clone, the error is 
```
the trait bound `&rome_js_syntax::JsxAttribute: AstNode` is not satisfied
the following implementations were found:
  <rome_js_syntax::JsxAttribute as AstNode>rustc[E0277](https://doc.rust-lang.org/error-index.html#E0277)
```

leops: Yes the node itself needs to be cloned but `replace_node` will take care of copying the trivia over in the correct place

leops (approved):

PR #2703, opened 2022-06-13
feat(xtask): automatically generate documentation pages for lint rules

leops:
## Summary

This PR adds a new `cargo lintdoc` automatically generating markdown documentation pages on the website for all registered lint rules

Internally this command relies on the content of the doc-comments of each lint rule that's now available as a runtime metadata alongside the name of the rule itself. A new `metadata()` method is available on the `RuleRegistry` to get the name and documentation of each active rule, and this method is aliased as the plain function `rome_js_analyze::metadata()` for the default instance of the JS rule registry.

The `xtask_lintdoc` crate makes use of this metadata by parsing the content of the comments as markdown, running the detected code blocks through the analyzer (similar to the "doctest" feature of `rustdoc`), and inserting snapshots of the emitted diagnostics if the analysis returned errors (using a newly-introduced HTML printer for `rome_console` markup)

## Test Plan

At the moment I've only run the new generation command manually, but it will probably need to be integrated to the automated test suite to ensure the generated files are kept up to date

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
@@ -1,4 +1,6 @@
-use rome_analyze::{context::RuleContext, ActionCategory, Rule, RuleCategory, RuleDiagnostic};
+use rome_analyze::{
+    context::RuleContext, declare_rule, ActionCategory, Rule, RuleCategory, RuleDiagnostic,
+};
 use rome_console::markup;
 use rome_diagnostics::Applicability;
 use rome_js_factory::make;
@@ -14,10 +16,12 @@ pub struct NoCompareNegZeroState {
     left_need_replaced: bool,
     right_need_replaced: bool,
 }
-pub(crate) enum NoCompareNegZero {}
+
+declare_rule! {
+    pub(crate) NoCompareNegZero = "noCompareNegZero"
+}
 
 impl Rule for NoCompareNegZero {
-    const NAME: &'static str = "noCompareNegZero";
     const CATEGORY: RuleCategory = RuleCategory::Lint;
 
     type Query = JsBinaryExpression;
--- a/crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
@@ -18,6 +18,21 @@ pub struct NoCompareNegZeroState {
 }
 
 declare_rule! {
+    /// Disallow comparing against `-0`
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// (1 >= -0)
+    /// ```
+    ///
+    /// ### Valid
+    ///
+    /// ```js
+    /// (1 >= 0)
+    ///```
     pub(crate) NoCompareNegZero = "noCompareNegZero"
 }
 
--- a/website/src/docs/lint/rules/index.md
+++ b/website/src/docs/lint/rules/index.md
@@ -18,35 +18,35 @@ eleventyNavigation:
 	<a href="/docs/lint/rules/noCompareNegZero">noCompareNegZero</a>
 	<a class="header-anchor" href="#noCompareNegZero"></a>
 </h3>
-Disallow comparing against `-0`
+Disallow comparing against <code>-0</code>
 </div>
 <div class="rule">
 <h3 data-toc-exclude id="noDebugger">
 	<a href="/docs/lint/rules/noDebugger">noDebugger</a>
 	<a class="header-anchor" href="#noDebugger"></a>
 </h3>
-Disallow the use of `debugger`
+Disallow the use of <code>debugger</code>
 </div>
 <div class="rule">
 <h3 data-toc-exclude id="noDelete">
 	<a href="/docs/lint/rules/noDelete">noDelete</a>
 	<a class="header-anchor" href="#noDelete"></a>
 </h3>
-Disallow the use of the `delete` operator
+Disallow the use of the <code>delete</code> operator
 </div>
 <div class="rule">
 <h3 data-toc-exclude id="noDoubleEquals">
 	<a href="/docs/lint/rules/noDoubleEquals">noDoubleEquals</a>
 	<a class="header-anchor" href="#noDoubleEquals"></a>
 </h3>
-Require the use of `===` and `!==`
+Require the use of <code>===</code> and <code>!==</code>
 </div>
 <div class="rule">
 <h3 data-toc-exclude id="noImplicitBoolean">
 	<a href="/docs/lint/rules/noImplicitBoolean">noImplicitBoolean</a>
 	<a class="header-anchor" href="#noImplicitBoolean"></a>
 </h3>
-Disallow implicit `true` values on JSX boolean attributes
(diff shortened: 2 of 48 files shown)


ematipico:
>    Add documentation to the remaining rules

I believe this can be done afterwards, we just need to create an issue and we should be fine.

crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
```diff
@@ -14,10 +14,27 @@ pub struct NoCompareNegZeroState {
...
     right_need_replaced: bool,
 }
-pub(crate) enum NoCompareNegZero {}
+
+declare_rule! {
+    /// Disallow comparing against `-0`
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid {#invalid}
```
MichaReiser, line 23: Can we write some documentation for the "documentation" structure and what syntax is allowed. That documentation could either be part of  a `README` or as part of `declare_rule`?

website/src/docs/lint/rules/index.md
```diff
@@ -0,0 +1,111 @@
...
+</h3>
+Enforce the use of `while` loops instead of `for` loops when the
+initializer and update expressions are not needed
+</div>
+<div class="rule">
+<h3 data-toc-exclude id="flipBinExp">
+	<a href="/docs/lint/rules/flipBinExp">flipBinExp</a>
+	<a class="header-anchor" href="#flipBinExp"></a>
+</h3>
+MISSING DOCUMENTATION
```
MichaReiser, line 109: My understanding is that these pages get deployed to the website. Is there already an entry point? If so, what's your plan on handling the missing documentation?

xtask/lintdoc/src/main.rs
```diff
@@ -0,0 +1,353 @@
...
+    markup,
+};
+use rome_diagnostics::{file::SimpleFile, Diagnostic};
+use xtask::{glue::fs2, *};
+
+use rome_analyze::{AnalysisFilter, ControlFlow};
+use rome_js_analyze::{analyze, metadata};
+use rome_js_syntax::SourceType;
+
+fn main() -> Result<()> {
```
MichaReiser, line 15: I think it becomes important to add additional checks to our CI pipeline to make sure that this command is run for all PRs to prevent an outdated homepage. This isn't an issue specific to this command but common to many of our `xtask` commands. 

One way we could accomplish this is by adding a CI step that generates the formatter, grammar, lint rules, etc. and verifies that thereafter are no git changes.

xtask/src/lib.rs
```diff
@@ -10,7 +10,7 @@ use std::{
 
 pub use crate::glue::{pushd, pushenv};
 
-pub use anyhow::{bail, Context as _, Result};
+pub use anyhow::{anyhow, bail, ensure, Context as _, Result};
```
MichaReiser, line 13: Unrelated to this PR. I saw that we are now using `anyhow` and `thiserror`. Any chance that we can reduce our usage to just one of the two?

MichaReiser (approved):

MichaReiser:
What's the code size impact of including the markdown of all rules as part of the linter crate? Do you plan to expose the documentation as part of the CLI?

crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
leops: I'd probably include it in the documentation for `declare_rule` as well as the (future) section on writing lint rules in the contribution docs, but the thing is I'm not entirely sure how should actually work anyway. Currently I've implemented the valid / invalid logic by detecting sections of the markdown document with the id `valid` and `invalid`, but I'm also considering having the expected status of a code block as an attribute on the language tag instead (for example ` ```js,should_fail `) similar to `rustdoc`

website/src/docs/lint/rules/index.md
leops: For the entry point I've added a link to this lint rules index page in the linting section of the main docs similarly to what existed in the JS version, but maybe it would also make sense to it to have an entry in the navigation menu ?
As for the content of the pages, the new generation script generates the individual rule documentation pages and the rules index page at the same place and with the same code as what used to exist in the JS version of Rome, including the "MISSING DOCUMENTATION" message. But I could easily change the syntax of the `declare_rule` macro to make the documentation mandatory, and fail to build if there is no doc-comment on the rule

xtask/src/lib.rs
leops: The usage of two libraries do not currently overlap in the codebase, `thiserror` in the CLI and `anyhow` in the Language Server and `xtask`. This wouldn't really be and issue anyway as those two library often need to coexist (and do so quite easily since any type deriving `thiserror::Error` can be cast into `anyhow::Error`) since they solve slightly different problems.
If we really want to get rid of at least one of them it think it should be somewhat easier to get rid of `thiserror` by implementing the `Error` trait manually instead of deriving it in the CLI (while removing `anyhow` would require writing a new error type that can be instantiated from all the other error types we implicitly cast into it over the codebase)

website/src/docs/lint/rules/noCompareNegZero.md
```diff
@@ -0,0 +1,36 @@
...
+
+# noCompareNegZero
+
+Disallow comparing against `-0`
+
+## Examples
+
+### Invalid
+
+```jsx
```
ematipico, line 14: Is this correct? Should it be `js`?

leops: The language ID emitted in the documentation page is the one used by the documentation generator and may not be exactly the same one specified in the original input: in this case the change is caused by `.js` files being interpreted as JSX [here](https://github.com/rome/tools/blob/6191cb7dd3ceb27022e70d7ced0e758a1e236699/crates/rome_js_syntax/src/source_type.rs#L242).

xtask/lintdoc/src/main.rs
leops: I've added a "Codegen" step to the CI checks run on pull requests, it runs the `grammar`, `analyzer` and `lintdoc` codegen commands before checking the git status is unchanged. A test / example of failing run can be found here: https://github.com/rome/tools/runs/6900285745?check_suite_focus=true

ematipico:
Also, should we allow people to see the rules now? The autofix hasn't been released yet and the new rules too, so I am not sure if it makes sense?

xtask/lintdoc/src/main.rs
MichaReiser: I love it! Thank you @leops for following up on this.

ematipico:
This page rendered in a weird way: https://feature-xtask-lintdoc.tools-8rn.pages.dev/docs/lint/rules/noImplicitBoolean/

ematipico:
> Indeed having auto-generated documentation from the latest revision of the code is going to clash with the workflow we currently have of holding off update to the documentation until the corresponding release is ready. While the linter is unstable it may not be too much of a problem since the main intended use-case for these documentation pages is explaining to users why they are seeing certain lint diagnostics, so if they're using a version of the analyzer that doesn't have those rules they wont need this documentation in the first place. We could also display which version of the toolchain introduced each rule in the documentation, but for the long term it might make sense to expose archives of past and future versions of the full documentation website along with the current one

We can definitely come up with a solution. Definitely not in this PR. Thank you for the insights!

PR #2707, opened 2022-06-14
feat(rome_js_analyze): useSelfClosingElements

### LINKED ISSUE #2706: 📎 useSelfClosingElements
### Description

`rome_analyze` rule [useSelfClosingElements](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/jsx/useSelfClosingElements.md)

IWANABETHATGUY:
## Summary
1. Resolved #2706 

## Test Plan
1. cargo test


IWANABETHATGUY:
@leops , @ematipico would you mind helping with reviewing  ?

crates/rome_js_analyze/src/analyzers/use_self_closing_elements.rs
```diff
@@ -0,0 +1,100 @@
...
+            }
+        } else {
+            let name = name.clone().ok()?;
+            let trailing = name.syntax().last_trailing_trivia();
+            if let Some(trailing) = trailing {
+                !trailing.text().ends_with(' ')
+            } else {
+                true
+            }
+        };
```
leops, line 72: This could be simplified a bit by using `prev_token` to get to the token preceding `r_angle_token` directly (using the original version of the token, not the one with the trivias stripped) along with the `Option` combinator methods:
```suggestion
        let prev_token = r_angle_token.prev_token();
        let need_extra_whitespace = prev_token.as_ref()
            .map(|token| token.syntax())
            .and_then(|token| token.last_trailing_trivia())
            .map_or(true, |trailing| !trailing.text().ends_with(' '));
```

crates/rome_analyze/tests/specs/useSelfClosingElements.jsx
```diff
@@ -0,0 +1,18 @@
...
+<div>child</div>;
+<Component />;
+<Component>child</Component>;
+<Foo.bar />;
+<Foo.bar>child</Foo.bar>;
+
+// invalid
+<div               ></div>;
+<Component></Component>;
+<Foo.bar></Foo.bar>;
```
leops, line 12: Can we add tests for TSX type arguments eg. `<Generic<true>></Generic>` ?

crates/rome_js_analyze/src/analyzers/use_self_closing_elements.rs
IWANABETHATGUY: I found that `r_angle_token.prev_token()` would always be `None`, so the `need_extra_whitespace` would always be true.

leops: That shouldn't be the case, are you calling `prev_token` on the token returned by `as_fields` ? The `with_leading_trivia` method returns a new detached token so it has no previous token, but the original token should definitely have a preceding token

IWANABETHATGUY: > 

You are right, Thanks!!

IWANABETHATGUY: done

leops (approved):

PR #2708, opened 2022-06-14
fix(rome_formatter): `dbg_write` API compatibility with `write`

MichaReiser:
## Summary

This PR ensures that the `buffer` argument of `dbg_write!` is compatible with the `buffer` argument of `write!` so that the two can be swapped without any changes to the debugged code.

The current implementation hasn't been compatible because the `dbg_write!` macro called into `Inspect::new` that expected a `&mut Buffer` which requires the use of `&mut buffer` if an implementation uses a local `Buffer` variable.

This PR changes the macro implementation to use `Buffer.inspect` instead that works for any mutable `Buffer`.

## Test Plan

I changed the `dbg_write` macro test to pass in `buffer` without `&mut`.

### DIFF
--- a/crates/rome_formatter/src/buffer.rs
+++ b/crates/rome_formatter/src/buffer.rs
@@ -391,7 +391,7 @@ pub struct Inspect<'inner, Context, Inspector> {
 }
 
 impl<'inner, Context, Inspector> Inspect<'inner, Context, Inspector> {
-    pub fn new(inner: &'inner mut dyn Buffer<Context = Context>, inspector: Inspector) -> Self {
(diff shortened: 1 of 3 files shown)


crates/rome_formatter/src/buffer.rs
```diff
@@ -433,6 +433,18 @@ pub trait BufferExtensions: Buffer + Sized {
     {
         Inspect::new(self, inspector)
     }
+
+    /// Writes a sequence of elements into this buffer.
```
MichaReiser, line 437: Yeah, I sneaked that into this PR. This allows writing a sequence of elements without going over `buffer.into_element`

ematipico (approved):

leops (approved):

PR #2709, opened 2022-06-14
refactor(rome_formatter): Group leading comments

MichaReiser:
## Summary

It's necessary to move leading comments outside a group to prevent they expand the enclosing group:

```javascript
/* a comment */
[1]
```

The `/* a comment */` belongs to the `[` group token that is part of a group wrapping the whole `[1]` expression. It's important that the comment `/* a comment */` gets moved out of the group element to avoid the `[1]` group expanding because of the line break inserted by the comment.

The `group_elements` implementation so far used `split_trivia` to accomplish this.

This PR changes the semantics of the `Comment` IR if `leading: true` is set to ignore line breaks when considering if an element fits for all leading comments that appear at the beginning of the line. This gives us the same behaviour as with `split_trivia` but without the need to move elements around.

I decided to change `Group` to store a `Box<[FormatElement]>` to avoid boxing List elements.

## Test Plan

`cargo test` because this PR isn't adding any new functionality

### Performance

This change improves performance marginally (split2 is this branch)

```
group                                    split                                  split2
-----                                    -----                                  ------
formatter/checker.ts                     1.04    201.2±1.88ms    12.9 MB/sec    1.00    192.7±1.09ms    13.5 MB/sec
formatter/compiler.js                    1.04    123.7±2.43ms     8.5 MB/sec    1.00    118.5±1.12ms     8.8 MB/sec
formatter/d3.min.js                      1.03     90.0±1.61ms     2.9 MB/sec    1.00     87.6±0.83ms     3.0 MB/sec
formatter/dojo.js                        1.01      6.4±0.10ms    10.7 MB/sec    1.00      6.4±0.02ms    10.8 MB/sec
formatter/ios.d.ts                       1.02    140.4±1.06ms    13.3 MB/sec    1.00    137.6±0.86ms    13.6 MB/sec
formatter/jquery.min.js                  1.01     26.9±0.16ms     3.1 MB/sec    1.00     26.7±0.06ms     3.1 MB/sec
formatter/math.js                        1.03    207.2±1.93ms     3.1 MB/sec    1.00    200.4±3.35ms     3.2 MB/sec
formatter/parser.ts                      1.03      4.4±0.08ms    11.1 MB/sec    1.00      4.2±0.06ms    11.4 MB/sec
formatter/pixi.min.js                    1.05    102.0±1.04ms     4.3 MB/sec    1.00     97.3±0.69ms     4.5 MB/sec
formatter/react-dom.production.min.js    1.04     30.8±0.06ms     3.7 MB/sec    1.00     29.6±0.21ms     3.9 MB/sec
formatter/react.production.min.js        1.02  1518.1±28.17µs     4.1 MB/sec    1.00  1483.7±25.72µs     4.1 MB/sec
formatter/router.ts                      1.02      3.4±0.03ms    18.2 MB/sec    1.00      3.3±0.02ms    18.5 MB/sec
formatter/tex-chtml-full.js              1.03    272.8±6.15ms     3.3 MB/sec    1.00    265.3±3.47ms     3.4 MB/sec
formatter/three.min.js                   1.06    123.5±2.33ms     4.8 MB/sec    1.00    116.5±0.77ms     5.0 MB/sec
formatter/typescript.js                  1.00    783.9±7.12ms    12.1 MB/sec    1.01   790.1±11.15ms    12.0 MB/sec
formatter/vue.global.prod.js             1.02     39.1±0.72ms     3.1 MB/sec    1.00     38.3±0.26ms     3.1 MB/sec
```

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -931,19 +931,29 @@ impl<Context> Format<Context> for GroupElements<'_, Context> {
 
         buffer.write_fmt(Arguments::from(&self.content))?;
 
-        let content = buffer.into_element();
+        let mut content = buffer.into_vec();
 
-        let (leading, content, trailing) = content.split_trivia();
+        // Move the leading comments out of the group to prevent that a leading line comment expands the
+        // token's enclosing group.
+        //
+        // ```javascript
+        // /* a comment */
+        // [1]
+        // ```
+        //
+        // The `/* a comment */` belongs to the `[` group token that is part of a group wrapping the whole
+        // `[1]` expression. It's important that the comment `/* a comment */` gets moved out of the group element
+        // to avoid that the `[1]` group expands because of the line break inserted by the comment.
+        let leading_end = content
+            .iter()
+            .position(|element| !matches!(element, FormatElement::Comment(_)))
+            .unwrap_or(content.len());
+
+        f.write_elements(content.drain(..leading_end))?;
+
         let group = Group::new(content).with_id(self.group_id);
 
-        if !leading.is_empty() {
-            f.write_element(leading)?;
-        }
         f.write_element(FormatElement::Group(group))?;
-
-        if !trailing.is_empty() {
-            f.write_element(trailing)?;
-        }
 
         Ok(())
     }
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -470,7 +470,7 @@ impl<Context> Format<Context> for LineSuffixBoundary {
 ///     SimpleFormatContext::default(),
 ///     [
 ///         group_elements(&format_args![
-///             comments(&format_args![token("// test"), hard_line_break()]).leading(true),
+///             comments(&format_args![token("// test"), hard_line_break()], CommentPosition::Leading),
 ///             token("a"),
 ///             soft_line_break_or_space(),
 ///             token("b")
@@ -484,27 +484,23 @@ impl<Context> Format<Context> for LineSuffixBoundary {
 /// );
 /// ```
 #[inline]
-pub fn comments<Content, Context>(content: &Content) -> FormatComments<Context>
+pub fn comments<Content, Context>(
+    content: &Content,
+    position: CommentPosition,
+) -> FormatComments<Context>
 where
     Content: Format<Context>,
 {
     FormatComments {
         content: Argument::new(content),
-        leading: false,
+        position,
     }
 }
 
 #[derive(Copy, Clone)]
 pub struct FormatComments<'a, Context> {
     content: Argument<'a, Context>,
-    leading: bool,
-}
-
-impl<Context> FormatComments<'_, Context> {
-    pub fn leading(mut self, leading: bool) -> Self {
-        self.leading = leading;
-        self
-    }
+    position: CommentPosition,
 }
 
 impl<Context> Format<Context> for FormatComments<'_, Context> {
@@ -516,7 +512,7 @@ impl<Context> Format<Context> for FormatComments<'_, Context> {
 
         f.write_element(FormatElement::Comments {
             content: content.into_boxed_slice(),
-            leading: self.leading,
+            position: self.position,
         })
     }
 }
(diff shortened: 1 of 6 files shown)


MichaReiser:
!bench_formatter

crates/rome_formatter/src/format_element.rs
```diff
@@ -480,7 +480,7 @@ impl FormatElement {
             FormatElement::Space => false,
             FormatElement::Line(line_mode) => matches!(line_mode, LineMode::Hard | LineMode::Empty),
             FormatElement::Indent(content) => content.will_break(),
-            FormatElement::Group(group) => group.content.will_break(),
+            FormatElement::Group(group) => group.content.iter().any(FormatElement::will_break),
```
ematipico, line 483: I am not even sure if it worth updating this function at this point. It's not used anywhere

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -626,12 +631,16 @@ impl<'a> ElementCallQueue<'a> {
...
+fn fits_on_line<'a, I>(
+    elements: I,
     args: PrintElementArgs,
     queue: &ElementCallQueue<'a>,
     printer: &mut Printer<'a>,
-) -> bool {
+) -> bool
+where
+    I: IntoIterator<Item = &'a FormatElement>,
+    I::IntoIter: DoubleEndedIterator,
```
ematipico, line 642: When do we use `DoubleEndedIterator`?

ematipico (approved):

crates/rome_formatter/src/builders.rs
```diff
@@ -931,20 +943,16 @@ impl<Context> Format<Context> for GroupElements<'_, Context> {
 
         buffer.write_fmt(Arguments::from(&self.content))?;
 
-        let content = buffer.into_element();
+        let content = buffer.into_vec();
+
+        if content.is_empty() && self.group_id.is_none() {
```
MichaReiser, line 948: We need to keep empty groups with group ids because they may be referenced by conditional content.

crates/rome_formatter/src/format_element.rs
MichaReiser: I believe it's still used in the JSX formatting branch.

MichaReiser:
I requested another review because I changed the implementation to extend the `Comment` IR with a custom printer behaviour so that we no longer need to rewrite the IR.

crates/rome_formatter/src/format_element.rs
```diff
@@ -57,7 +57,10 @@ pub enum FormatElement {
     /// a trivia content, and it should only have a limited influence on the
     /// formatting (for instance line breaks contained within will not cause
     /// the parent group to break if this element is at the start of it)
-    Comment(Content),
+    Comments {
+        content: Box<[FormatElement]>,
+        leading: bool,
```
ematipico, line 62: nit: I would argue that it might be better to have a `CommentPosition`/`CommentKind` enum with `Leading` and `Trailing` as variants.

crates/rome_formatter/src/token.rs
```diff
@@ -878,16 +880,19 @@ where
...
         let mut last_inline_comment = false;
 
-        for (index, comment) in comments.enumerate() {
-            if !self.skip_formatted_check && f.state().is_comment_formatted(comment.piece()) {
-                continue;
-            }
+        let mut comments = comments.enumerate().peekable();
+        let is_empty = comments.peek().is_none();
+
+        let format_comments = format_once(|f| {
```
ematipico, line 886: Wouldn't make more sense to move the creation for this `format_comments` after the `!is_empty` check? It **_seems_** to be used only in that case

ematipico (approved):

leops (approved):

MichaReiser:
Closing in favor of #2724 

The issue with this approach is that `will_break` can no longer determine whether a group will break because of a leading comment or not. Being able to do so is important for the call arguments formatting. 

#2724 keeps an implementation that is closer to what we had before that moves out the leading comments of the `group_elements`. This is slightly less performant and adds a bit of additional complexity when it comes to interned elements but seems reasonable to do considering that the functionality is necessary for call arguments.

Issue #2676, opened 2022-06-07
🐛 Formatter adds incorrect unnecessary `,`


RobinMalfait:
### Environment information

```bash
I could reproduce it in the Rome Playground:

https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#ZgBvAG8AKAB7AAoAIAAgAGIAYQByADoAIABiAGEAegAKAH0AKQA=
```

### What happened?

https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#ZgBvAG8AKAB7AAoAIAAgAGIAYQByADoAIABiAGEAegAKAH0AKQA=

I was playing around with Rome in the Tailwind CSS codebase, and noticed an incorrect formatting pattern. I reduced it to the smallest example I could think of. Given this input:

```js
foo({
  bar: baz
})
```

Rome will format it as:
```js
foo({
	bar: baz,
},);
```

Whereas Prettier formats it as:
```js
foo({
	bar: baz,
});
```

### Expected result

I expect the output to be the same as the one from Prettier:
```js
foo({
	bar: baz,
});
```

### Code of Conduct

- [X] I agree to follow Rome's Code of Conduct

leops:
I think this is related to #2421

Issue #2439, opened 2022-04-14
Destructuring arguments


MichaReiser:
[...]

PR #2711, opened 2022-06-14
feat(rome_js_formatter): call arguments

### LINKED ISSUE #2421: Call expression arguments formatting

ematipico:
## Summary

This PR closes #2421 and closes https://github.com/rome/tools/issues/2439

This PR requires #2685 in order to be merged.

The implementation of  the formatting requires the usage of `best_fitting`, which will impact the performance of formatter, in some way.

The usage of `best_fitting` also doesn't play well when using `dbg_write`, which is also understandable.

The implementation is a port of Prettier's algorithm: https://github.dev/prettier/prettier/blob/main/src/language-js/print/call-arguments.js

The implementation doesn't cover ALL the cases, as we don't have a yet a good API/builder for `will_break`, which means that some cases might not be covered.

The final version of the implementation now surgically caches the node/tokens only when we know they need to be printed multiple times. This was a lesson learned, where it's not good doing an early caching regardless, because this was impacting the performance in visible way.

This PR also fixes https://github.com/rome/tools/issues/2676, the trailing comma is manually added only in the cases where we know there's a group that might break.

Here's the work that has been done:
- extracted the logic of trivia formatting out of `format_delimited`, making it a standalone making it a standalone utility. This was needed because there were case where I needed manually replicate the delimiting logic, by customizing  the content. Also, I needed a way to cache the delimiters and print them in different positions based on the condition.
- added specific cases for handling hooks
- added specific cases to handle best fitting by following prettier's source code
- created a new function to print call arguments

## Regressions

A known regression is the absence of `will_break`, which in turns formats anything with comments in the wrong way. Prettier uses `willBreak` to check if some arguments break, and if so, then the formatting is different. Will break is fundamental because  comments (line suffixes to be more precise) mark an element as "yes, it will break". 

So a code like this:

```js
runtimeAgent.getProperties(
    objectId,
    false, // ownProperties
    false, // accessorPropertiesOnly
    false, // generatePreview
    (error, properties, internalProperties) => {
      return 1
    },
);
```

Will print like this:
```js
runtimeAgent.getProperties(objectId, false, false, false, (
	// ownProperties // accessorPropertiesOnly // generatePreview
	error,
	properties,
	internalProperties,
) => {
	return 1;
});
```

A workaround might be to actually check if each argument has comments, but I am not sure how the outcome is going to be for arguments that have deep nested comments. 

## Test Plan

I added new test cases, mostly edge cases. The rest of the formatting can be seen inside the prettier's test suite - I believe they speak by themselves. 

If you feel the need of more tests, please let me know.

This PR brings us closer to Prettier

main

**File Based Average Prettier Similarity**: 73.25%  
**Line Based Average Prettier Similarity**: 67.60%  

PR:

**File Based Average Prettier Similarity**: 73.61%  
**Line Based Average Prettier Similarity**: 68.79%  

I hoped in an higher number but it's still something :)


ematipico:
!bench_formatter

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                                    ]),
+                                    &r_paren
+                                ]
+                            )
+                        })]),
+                        &r_trailing_trivia
+                    ]
+                )
+            })
+            .memoized();
```
MichaReiser, line 209: This `memoized` shouldn't be necessary because `all_arguments_expanded` is only used once`.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                        l_trailing_trivia,
+                        group_elements(&format_args![format_with(|f| {
+                            fmt_arguments_multi_line(separated.iter(), separated.len(), f)
+                        })]),
+                        r_leading_trivia,
+                        r_paren,
+                        r_trailing_trivia
+                    ],
+                    format_args![formatted],
+                    format_args![all_arguments_expanded]
```
MichaReiser, line 226: I believe this should work

```suggestion
                    formatted,
                    all_arguments_expanded
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+        // We now need to allocate a new vector with cached nodes, this is needed because
+        // we can't attempt to print the same node twice without incur in "printed token twice" errors.
+        // We also disallow the trailing separator, we are interested in doing it manually.
+        let separated: Vec<_> = args
+            .format_separated(token(","))
+            .with_options(
+                FormatSeparatedOptions::default().with_trailing_separator(TrailingSeparator::Elide),
+            )
+            .map(|e| e.memoized())
+            .collect();
```
MichaReiser, line 109: `separated` only seems to be used twice if the implementation takes the `should_group_first_argument || should_group_last_argument` branch. It could help performance to move the `memoize` part into that block.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -523,3 +524,51 @@ impl Format<JsFormatContext> for FormatMemberName {
...
+    for (index, element) in iterator.by_ref() {
+        if index == number_of_elements - 1 {
+            join_with.entry(&format_args![&element, &if_group_breaks(&token(","))]);
+        } else {
+            join_with.entry(&element);
+        }
+    }
+
+    join_with.finish()
+}
```
MichaReiser, line 574: Nit: Should we move this into the call arguments formatting?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                                        )
+                                    }))
+                                    .finish()
+                            }),
+                            r_leading_trivia,
+                            r_paren,
+                            r_trailing_trivia
+                        ]
+                    )
+                }
```
MichaReiser, line 173: Took me a while to figure out what the difference between the two branches is. It may help to move out the shared logic out of the if branches:

```
write!(f, [l_leading_trivia, l_paren, l_trailing_trivia])?;

if should_group_first_argument {
 	 // special formatting of the first element
	  let mut iter = separated.iter();
	  // SAFETY: check on the existence of at least one argument are done before
	  let first = iter.next().unwrap();
	  f.join_with(&space_token())
	      .entry(&format_with(|f| {
	          write!(
	              f,
	              [group_elements(&format_args![first, expand_parent()])]
	          )
	      }))
	      .entries(iter)
	      .finish()?;
} else {
  	// special formatting of the last element
	  let mut iter = separated.iter();
	  // SAFETY: check on the existence of at least one argument are done before
	  let last = iter.next_back().unwrap();
	  
	  f.join_with(&space_token())
	      .entries(iter)
	      .entry(&format_with(|f| {
	          write!(
	              f,
	              [group_elements(&format_args![last, expand_parent()])]
	          )
	      }))
	      .finish()?;
}

write!(f, [r_leading_trivia, r_paren, r_trailing_trivia])?;
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+            // This is the version of where all the arguments are broken out
+            let all_arguments_expanded = format_with(|f| {
+                // this formatting structure replicates what we have inside the `format_delimited`
+                // function, but here we use a different way to print the trailing separator
+                write!(
+                    f,
+                    [
+                        &l_leading_trivia,
+                        &l_paren,
+                        &group_elements(&format_args![format_with(|f| {
```
MichaReiser, line 185: Nit
```suggestion
                        &group_elements(&format_with(|f| {
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                                [
+                                    &soft_block_indent(&format_args![
+                                        &l_trailing_trivia,
+                                        format_with(|f| {
+                                            fmt_arguments_multi_line(
+                                                separated.iter(),
+                                                separated.len(),
+                                                f,
+                                            )
+                                        }),
```
MichaReiser, line 197: Nit: It may be slighlty nicer to use if `format_arguments_multi_line` returns an object that implements `Format` rather than doing the formatting right away
```suggestion

                                            format_arguments_multi_line(
                                                separated.iter(),
                                                separated.len(),
                                            ),
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,24 +21,239 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+
+        // tokens on the left
+        let l_leading_trivia = open_delimiter.as_leading_trivia_fmt().memoized();
+        let l_paren = open_delimiter.as_token_fmt().memoized();
+        let l_trailing_trivia = open_delimiter.as_trailing_trivia_fmt().memoized();
+
+        // tokens on the right
+        let r_leading_trivia = close_delimiter.as_leading_trivia_fmt().memoized();
+        let r_paren = close_delimiter.as_token_fmt().memoized();
+        let r_trailing_trivia = close_delimiter.as_trailing_trivia_fmt().memoized();
```
MichaReiser, line 43: I recommend moving the `memoized` passed the early returns to avoid memorizing for the cases where it isn't necessary to do so.

crates/rome_js_formatter/src/js/lists/call_argument_list.rs
```diff
@@ -1,13 +1,23 @@
...
 
 impl FormatRule<JsCallArgumentList> for FormatJsCallArgumentList {
     type Context = JsFormatContext;
 
     fn fmt(node: &JsCallArgumentList, f: &mut JsFormatter) -> FormatResult<()> {
-        f.join_with(&soft_line_break_or_space())
-            .entries(node.format_separated(token(",")))
-            .finish()
+        if node.len() == 0 {
+            return write!(f, [empty_element()]);
```
MichaReiser, line 12: Nit: You don't have to perform a write. You can return `Ok(())` in cases where you don't want to write any output
```suggestion
            return Ok(());
```

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -523,3 +524,51 @@ impl Format<JsFormatContext> for FormatMemberName {
...
+/// This function is in charge to format the call arguments.
+pub(crate) fn fmt_arguments_multi_line<S: Format<JsFormatContext>, I>(
+    separated: I,
+    number_of_elements: usize,
+    f: &mut JsFormatter,
+) -> FormatResult<()>
+where
+    I: Iterator<Item = S>,
+    S: std::fmt::Debug,
+{
```
MichaReiser, line 537: Nit: You can require that `I implements ExactSizeIterator` so that you don't need to pass the length as well
```suggestion
pub(crate) fn fmt_arguments_multi_line<S: Format<JsFormatContext>, I>(
    separated: I,
    f: &mut JsFormatter,
) -> FormatResult<()>
where
    I: Iterator<Item = S> + ExactSizeIterator,
    S: std::fmt::Debug,
{
  let number_of_elements = separated.len();
```

MichaReiser: Or you can do 

```rust
let mut iterator = separated.enumerate().peekable();

while let Some((index, element)) = iterator.next() {
  let last = iterator.peek().is_none();

  if !last {
    join_with.entry(....)
  } else {
    ...
  }
} 
```

crates/rome_js_formatter/tests/specs/prettier/js/arrays/numbers-in-args.js.snap
```diff
@@ -24,7 +25,7 @@ expect(bifornCringerMoshedPerplexSawder.getLongArrayOfNumbers()).toEqual(
 ```js
 expect(bifornCringerMoshedPerplexSawder.getArrayOfNumbers()).toEqual([
   1, 2, 3, 4, 5,
-],);
```
MichaReiser, line 27: :partying_face:

crates/rome_js_formatter/tests/specs/prettier/js/first-argument-expansion/test.js.snap
```diff
@@ -122,68 +123,41 @@ func((args) => {
...
+  thing();
+}, 500);
 
-["a", "b", "c"].reduce(
-  function (item, thing) {
-    return thing + " " + item;
-  },
-  "letters:",
-);
+["a", "b", "c"].reduce(function (item, thing) {
```
MichaReiser, line 130: So much better :)

ematipico:
!bech_formatter

ematipico:
!bench_formatter

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
             r_paren_token,
         } = node.as_fields();
 
-        if is_simple_function_arguments(node)? {
-            return write![
+        let l_paren_token = l_paren_token?;
+        let r_paren_token = r_paren_token?;
+
+        // we now extracts the formatted version of trivias and tokens of the delimiters
```
MichaReiser, line 26: Is this comment still accurate or is it out of date because you moved some lines?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
-        )
-    }
-}
+        // particular formatting for hooks
+        if is_react_hook_with_deps_array(&args)? {
+            let mut list = args.elements();
+            // SAFETY: the function `is_react_hook_with_deps_array` already checks the presence of the
+            // first two arguments, so it's safe un unwrap them
+            let first_argument = list.next().unwrap();
+            let second_argument = list.next().unwrap();
```
MichaReiser, line 45: Nit: Using `unwrap` here is fine but it may lead to issues if someone ever changes `is_react_hook_with_deps_array` and isn't familiar with this call site. 

An alternative would be that `is_react_hook_with_deps_array` returns the first two arguments instead of a bool:

```
if let Some((first_argument, dependencies_argument)) = react_hooks_arguments(args) {
    ...
}
```

where `react_hook_arguments` returns a `Some((JsAnyArgument, JsAnyArgument))`

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                    format_with(|f| {
+                        // we don't want to print the trailing separator, so if it's present, we replace it
+                        // with an empty element
+                        if let Some(separator) = second_argument.trailing_separator()? {
+                            return write!(f, [format_removed(separator)]);
+                        }
 
-    if token_has_comments(&l_paren_token?) || token_has_comments(&r_paren_token?) {
-        return Ok(false);
+                        Ok(())
```
MichaReiser, line 62: Nit: I personally prefer to split the write into multiple blocks if I have conditional logic:

```
write!(f, [l_paren_token.format(), ..., second_argument.node().format()])?;

if let Some(separator) = second_argument.trailing_separator()? {
    write!(f, [format_removed(separator)])?;
}

return write!(f, [r_paren_token.format()]);
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                ]
+            );
+        }
+
+        // we create open a close delimiters
+        let open_delimiter = format_open_delimiter(&l_paren_token);
+        let close_delimiter = format_close_delimiter(&r_paren_token);
+
+        // tokens on the left
+        let l_leading_trivia = open_delimiter.as_leading_trivia_fmt();
```
MichaReiser, line 74: Nit: I started to use `fmt` only for the functions that do the formatting like `Format::fmt` and instead call function that return something formattable `format`. I would probably be fine just calling it `open_delimiter.leading_trivia()` and `open_delimiter.trailing_trivia()` (and `token`) because `open_delimiter` is a `format`. You may want to rename `open_delimiter` to `format_open_delimiter` if you want to be more explicit about (and `leading_trivia` to `format_leading_trivia`)

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                                    write!(
+                                        f,
+                                        [group_elements(&format_args![first, expand_parent()])]
+                                    )
+                                }))
+                                .entries(iter)
+                                .finish()
+                        }),]
+                    )?;
+                } else {
```
MichaReiser, line 139: Nit: This can be simplified to 

```rust
 // special formatting of the first element
                            let mut iter = separated.iter();
                            // SAFETY: check on the existence of at least one argument are done before
                            let first = iter.next().unwrap();
                            f.join_with(&space_token())
                                .entry(&format_with(|f| {
                                    write!(
                                        f,
                                        [group_elements(&format_args![first, expand_parent()])]
                                    )
                                }))
                                .entries(iter)
                                .finish()
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                                .entry(&format_with(|f| {
+                                    write!(
+                                        f,
+                                        [group_elements(&format_args![last, expand_parent()])]
+                                    )
+                                }))
+                                .finish()
+                        }),]
+                    )?;
+                }
```
MichaReiser, line 159: Same here: 

```rust
 // special formatting of the last element
                            let mut iter = separated.iter();
                            // SAFETY: check on the existence of at least one argument are done before
                            let last = iter.next_back().unwrap();

                            f.join_with(&space_token())
                                .entries(iter)
                                .entry(&format_with(|f| {
                                    write!(
                                        f,
                                        [group_elements(&format_args![last, expand_parent()])]
                                    )
                                }))
                                .finish()
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                                            fmt_arguments_multi_line(separated.iter(), f)
+                                        }),
+                                        &r_leading_trivia,
+                                        soft_line_break()
+                                    ]),
+                                    &r_paren
+                                ]
+                            )
+                        })),
+                        &r_trailing_trivia
```
MichaReiser, line 189: Nit:I think you can remove the most outer `format_with`

```rust
&group_elements(&soft_block_indent(&format_args![
                                        &l_trailing_trivia,
                                        format_with(|f| {
                                            fmt_arguments_multi_line(separated.iter(), f)
                                        }),
                                        &r_leading_trivia,
                                        soft_line_break()
                                    ]),
                                    &r_paren
                        )
                )
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                f,
+                [best_fitting![
+                    format_args![
+                        l_leading_trivia,
+                        l_paren,
+                        l_trailing_trivia,
+                        group_elements(&format_args![format_with(|f| {
+                            fmt_arguments_multi_line(separated.iter(), f)
+                        })]),
+                        r_leading_trivia,
```
MichaReiser, line 204: Nit: You can remove the `foramt_args` here because `group_elements` has only a single content

```
group_elements(&format_with(|f| {
                            fmt_arguments_multi_line(separated.iter(), f)
                        })),
```

This could be even written nicer if you change `fmt_arguments_multi_line` to return `impl Format` (and rename to `format_arguments_multi_line`)

```
group_elements(&format_arguments_multi_line(separated.iter())),
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+    Ok(!has_comments
+        && is_function_like
+        && !second_arg_is_function_like
+        && !could_group_argument(second.syntax(), false)?)
+}
+
+/// Checks if the last group requires grouping
+fn should_group_last_argument(list: &JsCallArgumentList) -> SyntaxResult<bool> {
+    let mut iter = list.iter().rev();
+    // SAFETY: checked at the beginning of the function
```
MichaReiser, line 279: I think you can remove this comment. There's no unsafety happening on the next line

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+    let last = iter.next();
+    let penultimate = iter.next();
+
+    if let Some(last) = last {
+        // SAFETY: guaranteed by the syntax factory
```
MichaReiser, line 284: Can you expand on this safety comment. It's unclear what the syntax factory is guaranteeing in this case.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+        let last = last.unwrap();
+        let check_with_penultimate = if let Some(penultimate) = penultimate {
+            // SAFETY: guaranteed by the syntax factory
+            let penultimate = penultimate.unwrap();
```
MichaReiser, line 288: I "think" the safety you're refering to in the previous comment is that the `node` can never be `Err` for the last element in a list. I don't think this holds for the second last element because you can have `f(arg, , last)` in which case the node is missing for the second argument. 

I would recommend using the try operator instead (for both cases) to not rely on that fact and handle whatever you encounter cracefully.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+        Ok(!last.syntax().has_comments_direct()
+            && could_group_argument(last.syntax(), false)?
+            && check_with_penultimate)
+    } else {
+        Ok(false)
     }
+}
+
+/// Checks if the current argument could be grouped
+fn could_group_argument(argument: &JsSyntaxNode, is_arrow_recursion: bool) -> SyntaxResult<bool> {
```
MichaReiser, line 305: I still believe it should be possible to use `JsAnyArgument` for `argument` instead of a `SyntaxNode` if you wrap the `assertion_expression.expression` and `as_expression.expression` in a `JsAnyArgument`

```
could_group_argument(JsAnyArgument::from(assertion_expression.expression()?))
```


Nit: Rename to `can_group_argument`?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                    TsReferenceType::can_cast(any_type.syntax().kind())
+                        || JsFunctionBody::cast(body.syntax().clone()).map_or(
+                            true,
+                            |function_body| {
+                                function_body
+                                    .statements()
+                                    .iter()
+                                    .any(|st| st.syntax().kind() == JS_EMPTY_STATEMENT)
+                            },
+                        )
```
MichaReiser, line 344: Nit: Ideally we should avoid going over `syntax` whenever possible because the compiler gives us less guarantees (let's say we remove `TsReferenceType` from the `AnyType` union, we won't know until someone reviews this code)

```
matches!(any_type, TsAnyReturnType::TsType(TsType::TsReferenceType(_))) || 
match body {
    JsAnyFunctionBody::JsFunctionBody(body) => body.statements().iter().any(|st| matches!(st, JsAnyStatement::JsEmptyStatement(_))),
    _ => false
}
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+
+        let body_is_delimited = matches!(
+            body,
+            JsAnyFunctionBody::JsFunctionBody(_)
+                | JsAnyFunctionBody::JsAnyExpression(JsAnyExpression::JsObjectExpression(_))
+                | JsAnyFunctionBody::JsAnyExpression(JsAnyExpression::JsArrayExpression(_))
+        );
+
+        if let JsAnyFunctionBody::JsAnyExpression(any_expression) = body.clone() {
+            let is_nested_arrow_function =
```
MichaReiser, line 355: I think you can use `&` to avoid the cloning here since you aren't mutating the element

```
if let JsAnyFunctionBody::JsAnyExpression(any_expression) = &body {
    ...
}
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+                JsArrowFunctionExpression::cast(any_expression.syntax().clone())
+                    .and_then(|arrow_function_expression| {
+                        arrow_function_expression
+                            .body()
+                            .ok()
+                            .and_then(|body| could_group_argument(body.syntax(), true).ok())
+                    })
+                    .unwrap_or(false);
```
MichaReiser, line 363: Nit: I would also recommend to match on the AST types here

```
match &any_expression {
    JsAnyExpression::JsArrowFunctionExpression(arrow_function) => match arrow_function.body()? {
      JsAnyFunctionBody::JsExpression(expession) => could_group_argument(JsAnyArgument::JsAnyExpression(expression), true)?,
      _ => false
    },
    _ => false
}
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+
+            body_is_delimited
+                && is_nested_arrow_function
+                && can_group_type
+                && (!is_arrow_recursion
+                    && (is_call_like_expression(&any_expression)
+                        || JsConditionalExpression::can_cast(body.syntax().kind())))
```
MichaReiser, line 370: NIt: Use `matches` with AST types

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+        } else {
+            body_is_delimited && can_group_type
+        }
+    } else {
+        JsFunctionExpression::can_cast(argument.kind())
```
MichaReiser, line 375: Nit: uses `matches!` with AST types

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -14,60 +20,394 @@ impl FormatNodeFields<JsCallArguments> for FormatNodeRule<JsCallArguments> {
...
+            let no_parameters = arrow_function.parameters()?.is_empty();
+            let body = arrow_function.body()?;
+            let is_block = matches!(body, JsAnyFunctionBody::JsFunctionBody(_));
+
+            no_parameters && is_block
+        } else {
+            false
+        };
+
+        let second_node_matches = JsArrayExpression::can_cast(second.syntax().kind());
```
MichaReiser, line 405: Nit: Use `matches!` with AST nodes instead.

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -523,3 +524,28 @@ impl Format<JsFormatContext> for FormatMemberName {
...
 }
+
+/// This function is in charge to format the call arguments.
+pub(crate) fn fmt_arguments_multi_line<S: Format<JsFormatContext>, I>(
+    separated: I,
+    f: &mut JsFormatter,
+) -> FormatResult<()>
+where
+    I: Iterator<Item = S>,
+    S: std::fmt::Debug,
```
MichaReiser, line 535: Nit: I assume the `fmt::Debug` was only needed for debugging and can now be removed?

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -523,3 +524,28 @@ impl Format<JsFormatContext> for FormatMemberName {
         }
     }
 }
+
+/// This function is in charge to format the call arguments.
+pub(crate) fn fmt_arguments_multi_line<S: Format<JsFormatContext>, I>(
+    separated: I,
+    f: &mut JsFormatter,
+) -> FormatResult<()>
```
MichaReiser, line 532: Nit: I would recommend changing the return type to `impl Format<JsFormatContext>` and return a `format_with` instead (and rename to `format_arguments_multi_line`). This has the benefit that the callers don't have to use `format_with` and can instead use `format_args!`

crates/rome_js_syntax/src/parameters_ext.rs
```diff
@@ -0,0 +1,15 @@
...
+        match self {
+            JsAnyArrowFunctionParameters::JsAnyBinding(_) => 1,
+            JsAnyArrowFunctionParameters::JsParameters(parameters) => parameters.items().len(),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
```
MichaReiser, line 15: Nit: Move to `union_ext`?

crates/rome_rowan/src/syntax/node.rs
```diff
@@ -472,12 +472,32 @@ impl<L: Language> SyntaxNode<L> {
         self.has_trailing_comments() || self.has_leading_comments()
     }
 
+    /// It checks if the current node has comments at the edges:
+    /// if first or last tokens contain comments (leading or trailing)
+    pub fn has_comments_at_the_edges(&self) -> bool {
```
MichaReiser, line 477: Nit: Maybe name it `first_or_last_token_has_comments` to keep the terminology that you introduced with the other helper functions. My first intution was that `edges` means if the first token has leading and the last token has trailing comments but it includes both comment kinds for both tokens.

crates/rome_rowan/src/syntax/node.rs
```diff
@@ -472,12 +472,32 @@ impl<L: Language> SyntaxNode<L> {
...
+    }
+
     /// Whether the node contains trailing comments.
     pub fn has_trailing_comments(&self) -> bool {
         self.last_token()
             .map_or(false, |tok| tok.has_trailing_comments())
     }
 
+    /// Whether the last token of a node has comments (leading or trailing)
+    pub fn last_token_has_comments(&self) -> bool {
```
MichaReiser, line 488: Unrelated to this line: You brought up the issue with comments on discord the other day. Do these helpers address the issue you raised on discord?

MichaReiser (commented):
This is great work @ematipico ! I've a few smaller code improvements but that's it. 

My main question now is how we go about to land this since this PR depends on my PR which isn't handling `split_trivia` correctly yet.

MichaReiser:
Compatibility:
**Before**
**File Based Average Prettier Similarity**: 73.25%  
**Line Based Average Prettier Similarity**: 67.60%  

**After**
**File Based Average Prettier Similarity**: 73.61%  
**Line Based Average Prettier Similarity**: 68.79%

crates/rome_js_formatter/src/builders.rs
```diff
@@ -402,3 +381,95 @@ enum DelimitedMode {
...
+            // Not really interested in the pre-amble, but want to know if it was written
+            let mut buffer = VecBuffer::new(f.state_mut());
+
+            write!(buffer, [format_trailing_trivia(self.open_token)])?;
+
+            let trivia = buffer.into_vec();
+
+            if !trivia.is_empty() {
+                f.write_elements(trivia)?;
+                soft_line_break_or_space().fmt(f)?;
```
MichaReiser, line 429: This must be replaced with `soft_line_break`. I changed that in my "inline comments" formatting PR but has probably been difficult to merge. This should fix the inline comments issue (The same is true for the leading trivia in the `CloseDelimiter` implementation

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
ematipico: `r_paren` is outside the `soft_block_indent` call

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: About which part are you confused?

* `Format::fmt`: Does the formatting -> uses `fmt` in the name
* `format_leading_trivia` etc: Returns an object that implements `Format`. 

`as_leading_trivia_fmt` returns an object that implements `Format`. That's why I would either call it: `format_leading_trivia` or just do `let format_leading_trivia = format_open_delimiter.leading_trivia();`

crates/rome_js_formatter/src/utils/mod.rs
ematipico: It seems that this is not possible due to lifetime issue AND borrow checker issues. I will leave it as it is for now

crates/rome_rowan/src/syntax/node.rs
ematipico: Yes, this was the solution I was looking for. It might be an edge case or not, but for now I decided to applying it only to this particular case.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
ematipico: > Nit: Rename to can_group_argument?

I'd prefer to keep this name because it's more in prettier with prettier's naming convention

crates/rome_js_formatter/src/builders.rs
ematipico: Done! The last commit should fix the regressions

crates/rome_js_formatter/src/js/lists/call_argument_list.rs
```diff
@@ -1,13 +1,26 @@
...
+        if node.len() == 0 {
+            return Ok(());
+        }
+
+        write!(
+            f,
+            [&group_elements(&soft_block_indent(&format_with(|f| {
+                let separated = node.format_separated(JsSyntaxKind::COMMA).with_options(
+                    FormatSeparatedOptions::default()
+                        .with_trailing_separator(TrailingSeparator::Elide),
```
MichaReiser, line 20: Nit: I would probably go with `Omit` as `Elide` seems to be specific to pronunciation. [link](https://english.stackexchange.com/questions/226141/elide-versus-omit)

crates/rome_js_formatter/src/utils/mod.rs
```diff
@@ -523,3 +524,27 @@ impl Format<JsFormatContext> for FormatMemberName {
         }
     }
 }
+
+/// This function is in charge to format the call arguments.
+pub(crate) fn format_arguments_multi_line<S: Format<JsFormatContext>, I>(
```
MichaReiser, line 529: Nit: I'm sorry. My previous comment may not have been precise enough. 

I would rename this back to `fmt` or `write_` because it does the formatting and doesn't return an object implementing `Format`. 

* `format_` -> Returns an object implementing `Format`
* `write_` or `fmt` -> If it performs the formatting right away

crates/rome_js_formatter/tests/specs/js/module/arrow/arrow_nested.js.snap
```diff
@@ -62,16 +59,15 @@ const promiseFromCallback = (fn) =>
...
-	objectId,
-	false, // ownProperties
-	false, // accessorPropertiesOnly
-	false, // generatePreview
-	(error, properties, internalProperties) => {
-		return 1;
-	},
-);
+runtimeAgent.getProperties(objectId, false, false, false, (
+	// ownProperties // accessorPropertiesOnly // generatePreview
```
MichaReiser, line 66: Hmm, this not ideal. It's now moving the comments from the arguments to the beginning of the arguments list. I'm not yet sure why this is happening but we should look into this.

Edit: Just saw that you're aware of this.

MichaReiser (approved):

crates/rome_js_formatter/tests/specs/js/module/arrow/arrow_nested.js.snap
ematipico: Tracked here: https://github.com/rome/tools/issues/2720

crates/rome_js_formatter/tests/specs/js/module/comments.js.snap
```diff
@@ -92,15 +93,13 @@ import {
 expression(/* block comment */);
 
 expression(
-	/* block comment */
+/* block comment */
```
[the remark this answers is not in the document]
ematipico: Tracked here https://github.com/rome/tools/issues/2720

crates/rome_js_formatter/src/builders.rs
```diff
@@ -402,3 +381,95 @@ enum DelimitedMode {
...
+
+    /// It extracts the formatted leading trivia of the token, without writing it in the buffer
+    pub(crate) fn format_trailing_trivia(&self) -> impl Format<JsFormatContext> + 't {
+        format_trailing_trivia(self.close_token)
+    }
+
+    /// It extracts the formatted trailing trivia of the token, without writing it in the buffer
+    pub(crate) fn format_leading_trivia(&self) -> impl Format<JsFormatContext> + 't {
+        format_with(|f| {
+            let mut buffer = PreambleBuffer::new(f, soft_line_break_or_space());
```
MichaReiser, line 462: A `soft_line_break` is sufficient here. 
```suggestion
            let mut buffer = PreambleBuffer::new(f, soft_line_break());
```

PR #2714, opened 2022-06-15
feature(rome_analyzer): Rule context

xunilrj:
PR for this discussion: https://github.com/rome/tools/discussions/2603

The heart of the PR is the type RuleContext. It acts as a simple façade for everything that we want to offer to lint rules. What we offer varies by language. To see an example look at JsRuleContext, which specifies which methods we allow to be called directly on the context (like extension methods), and impl LanguageSpecificServiceBag for JsLanguage which specifies which services are stored inside the context for each language.

The RuleContextServiceBag is typed and not a usual ServiceLocator to avoid all get(...) methods returning Option<...>. This comes with a little bit of complexity but also guarantees that we will not see any runtime errors when linting.

Another nice side effect is that lint rules for other languages will not compile until someone specifies which services these languages need.

Rules now receive just the context as a parameter

```rust
    ...
}
```

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_compare_neg_zero.rs
@@ -92,9 +92,9 @@ impl Rule for NoCompareNegZero {
                     ),
                 ),
             )?;
-            root.clone().replace_node(node.clone(), binary)?
+            root.replace_node(node.clone(), binary)?
         } else if state.left_need_replaced {
-            root.clone().replace_node(
+            root.replace_node(
                 node.left().ok()?,
                 JsAnyExpression::JsAnyLiteralExpression(
                     JsAnyLiteralExpression::JsNumberLiteralExpression(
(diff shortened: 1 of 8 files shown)


MichaReiser (approved):

crates/rome_analyze/src/context.rs
```diff
@@ -0,0 +1,26 @@
...
+
+impl<R> RuleContext<R>
+where
+    R: Rule,
+{
+    pub fn new(query_result: <R as Rule>::Query, root: RuleRoot<R>) -> Self {
+        Self { query_result, root }
+    }
+
+    pub fn query_result(&self) -> &<R as Rule>::Query {
```
leops, line 19: Nit: this method could be called `query`, the name is slightly less descriptive but since it's going to get called a lot making it short might be more important ?

crates/rome_analyze/src/context.rs
```diff
@@ -0,0 +1,26 @@
...
+        &self.query_result
+    }
+
+    pub fn root(&self) -> &RuleRoot<R> {
+        &self.root
```
leops, line 24: Since most calls to `root` are going to clone it anyway the call to `.clone()` may be inlined here to make the method return an owned `RuleRoot` directly

leops (approved):

crates/rome_analyze/src/context.rs
xunilrj: Done.

Issue #2716, opened 2022-06-16
📎 (rome_js_analyze): noMultipleSpacesInRegularExpressionLiterals


IWANABETHATGUY:
### Description

https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/regex/noMultipleSpacesInRegularExpressionLiterals.md

PR #2717, opened 2022-06-16
feat(rome_js_analyze):noMultipleSpacesInRegularExpressionLiterals

IWANABETHATGUY:
## Summary
1. Resolved https://github.com/rome/tools/issues/2716

## Test Plan
1. cargo test

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
@@ -69,8 +69,7 @@ impl Rule for NoMultipleSpacesInRegularExpressionLiterals {
         let mut normalized_string_token = String::new();
         let mut previous_start = 0;
 
-        let eg_length = state.iter().fold(0usize, |acc, (s, e)|  acc + *e - *s);
-
+        let eg_length = state.iter().fold(0usize, |acc, (s, e)| acc + *e - *s);
 
         for (start, end) in state.iter() {
             normalized_string_token += &trimmed_token_string[previous_start..*start];
--- a/crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
@@ -1,0 +1,97 @@
+use rome_analyze::{ActionCategory, Rule, RuleCategory, RuleDiagnostic};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_syntax::{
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
@leops , would mind helping with reviewing?

crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
```diff
@@ -0,0 +1,96 @@
...
+        let trimmed_token = node.value_token().ok()?;
+        let trimmed_token_string = trimmed_token.text_trimmed();
+        let mut normalized_string_token = String::new();
+        let mut previous_start = 0;
+
+        let eg_length = state.iter().fold(0usize, |acc, (s, e)| acc + *e - *s);
+
+        for (start, end) in state.iter() {
+            normalized_string_token += &trimmed_token_string[previous_start..*start];
+            normalized_string_token += &format!(" {{{}}}", *end - *start);
```
leops, line 76: You can use the `write!` macro here to append formatted text to a string (and unwrap the result since writing to a string like this cannot fail)

crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
```diff
@@ -0,0 +1,96 @@
...
+        ))
+    }
+
+    fn action(root: JsAnyRoot, node: &Self::Query, state: &Self::State) -> Option<JsRuleAction> {
+        let trimmed_token = node.value_token().ok()?;
+        let trimmed_token_string = trimmed_token.text_trimmed();
+        let mut normalized_string_token = String::new();
+        let mut previous_start = 0;
+
+        let eg_length = state.iter().fold(0usize, |acc, (s, e)| acc + *e - *s);
```
leops, line 72: This sum could be calculated in the following for loop to avoid iterating over the state twice

crates/rome_js_analyze/src/analyzers/no_multiple_spaces_in_regular_expression_literals.rs
IWANABETHATGUY: I am wonder what is the difference between `String` addition and `write!` macro ?

leops: The main difference is that `format!` will allocate an intermediate string then append it to `normalized_string_token` while `write!` will write to the string buffer directly

IWANABETHATGUY: Thanks

leops (approved):

PR #2718, opened 2022-06-16
refactor(rome_js_formatter): Introduce `format_parenthesize`

MichaReiser:
## Summary

This PR replaces the usages of `split_trivia` where it is used to move the leading and trailing trivia of a "to be parenthesised" expression with `format_parenthesize` that manually formats the leading / trailing comments as part of the parentheses. 

```javascript
/* leading */ "string" /* trailing */;

// Becomes
/* leading */ ("string") /* trailing */;
```

Formatting the comments of the content as part of the parentheses requires a way to signal to the formatter that these comments have already been formatted and that it shouldn't format the comments again when formatting the `"string"` token. This is done by storing the "manually" formatted comments inside of the `FormatState` and checking for every comment if the comment has already been formatted. 

The advantage of this approach is that it enables format rules to manually format comments if there's a need to do so without going over `split_trivia`. 

## New Dependencies

This PR adds a dependency on `indexmap` to `rome_formatter`. The reason to use `IndexMap` over `HashMap` is that the new map is part of the `FormatState` that supports taking and restoring snapshots. Thus, it is important that the data structure offers a cheap mechanism to restore the "manual comments" map to a previous state. 

`IndexMap` (or `IndexSet`) offers that because it's possible to truncate the `set` and by doing so, remove all elements that have been added since the snapshot was taken. The built-in `HashMap` doesn't provide this functionality. 

`indexmap` is only a new dependency for `rome_formatter`. It's already used by the parser, `lsp` crate and others.

## Next Steps

Remove the need for `split_trivia` inside of `group_elements`. 

## Test Plan

`cargo test`. I added a few new tests verifying the new behaviour.

### DIFF
--- a/crates/rome_formatter/Cargo.toml
+++ b/crates/rome_formatter/Cargo.toml
@@ -9,3 +9,4 @@ edition = "2021"
 rome_rowan = { path = "../rome_rowan" }
 tracing = { version = "0.1.31", default-features = false, features = ["std"] }
 cfg-if = "1.0.0"
+indexmap = "1.8.2"
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -6,7 +6,6 @@ use crate::{Buffer, VecBuffer};
 use rome_rowan::{Language, SyntaxNode, SyntaxToken, SyntaxTokenText, TextLen};
 use std::borrow::Cow;
 use std::cell::Cell;
-use std::fmt::Debug;
 use std::marker::PhantomData;
 
 /// A line break that only gets printed if the enclosing `Group` doesn't fit on a single line.
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -1,8 +1,6 @@
 use crate::prelude::*;
-use crate::token::{FormatInserted, InsertedToken};
 use crate::{
-    format_element, write, Argument, Arguments, CommentStyle, GroupId, PreambleBuffer, TextRange,
-    TextSize,
+    format_element, write, Argument, Arguments, GroupId, PreambleBuffer, TextRange, TextSize,
 };
 use crate::{Buffer, VecBuffer};
 use rome_rowan::{Language, SyntaxNode, SyntaxToken, SyntaxTokenText, TextLen};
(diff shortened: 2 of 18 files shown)


leops (commented):
Would it be possible to write this helper to make it support wrapping empty nodes in parenthesis transparently, instead of having to unwrap on the first and last token everywhere it's used ?

MichaReiser:
> Would it be possible to write this helper to make it support wrapping empty nodes in parenthesis transparently, instead of having to unwrap on the first and last token everywhere it's used ?

That's a good point. I guess we could do so for `format_parenthesize`.

crates/rome_formatter/Cargo.toml
```diff
@@ -9,3 +9,4 @@ edition = "2021"
 rome_rowan = { path = "../rome_rowan" }
 tracing = { version = "0.1.31", default-features = false, features = ["std"] }
 cfg-if = "1.0.0"
+indexmap = "1.8.2"
```
ematipico, line 12: Could you update the PR's description and add small paragraph of why this crate was chosen to implement your new feature?

crates/rome_formatter/src/lib.rs
```diff
@@ -1075,6 +1076,15 @@ pub struct FormatState<Context> {
...
     last_token_kind: Option<LastTokenKind>,
 
+    /// Tracks comments that have been formatted manually and shouldn't be emitted again
+    /// when formatting the token the comments belong to.
+    ///
+    /// The map stores the absolute position of the manually formatted comments.
+    /// Storing the position is sufficient because comments are guaranteed to not be empty
+    /// (all start with a specific comment sequence) and thus, no two comments can have the same
+    /// absolute position.
+    manually_formatted_comments: IndexMap<TextSize, ()>,
```
ematipico, line 1086: Why `IndexMap` and not `IndexSet`? As we don't store anything as value

crates/rome_formatter/src/token.rs
```diff
@@ -79,6 +80,218 @@ where
...
+    /// The token before which the open paren must be inserted
+    before_token: Option<SyntaxToken<S::Language>>,
+
+    /// The token text of the open paren
+    text: &'static str,
+
+    /// The kind of the open paren
+    kind: <S::Language as Language>::Kind,
+
+    style: S,
```
ematipico, line 113: `comment_style`? I thinks the trait is not sufficient (it's just a type) to explain what this is style is about

crates/rome_formatter/src/token.rs
```diff
@@ -79,6 +80,218 @@ where
...
+            for piece in leading_pieces {
+                if let Some(comment) = piece.as_comments() {
+                    comments.push(SourceComment::leading(comment, lines_before));
+                    lines_before = 0;
+                } else if piece.is_newline() {
+                    lines_before += 1;
+                } else if piece.is_skipped() {
+                    // Keep the skipped trivia inside of the parens. Handled by the
+                    // formatting of the `before_token`.
+                    break;
```
ematipico, line 157: Don't we have cases where we might have comments after a skipped trivia?
E.g.

```js
@Controller() // this is a controller
class {}
```

crates/rome_formatter/src/token.rs
```diff
@@ -79,6 +80,218 @@ where
...
+impl<S> FormatInsertedCloseParen<S>
+where
+    S: CommentStyle,
+{
+    pub fn new<Context>(
+        after_token: &Option<SyntaxToken<S::Language>>,
+        kind: <S::Language as Language>::Kind,
+        text: &'static str,
+        style: S,
+        f: &mut Formatter<Context>,
```
ematipico, line 248: I am not sure this is a correct pattern... I am, passing to `new` an extraneous argument is not stored in the struct. I have never seen this before

Wouldn't it better to have something like:

```
impl FormatInsertedCloseParen {
  #[must_use]
  pub fn new(self) -> Self {}

  pub fn compute(self, f: &mut JsFormatter) -> Self {}
}
```

crates/rome_js_formatter/src/builders.rs
```diff
@@ -52,6 +53,112 @@ pub fn format_inserted(kind: JsSyntaxKind) -> FormatInserted<JsCommentStyle> {
...
+#[derive(Clone)]
+pub struct FormatParenthesize<'content> {
+    grouped: bool,
+    first_content_token: Option<JsSyntaxToken>,
+    content: Argument<'content, JsFormatContext>,
+    last_content_token: Option<JsSyntaxToken>,
+}
+
+impl FormatParenthesize<'_> {
+    pub fn grouped(mut self) -> Self {
```
ematipico, line 127: documentation? Let's make sure to add that this function adds a level of indentation

crates/rome_js_formatter/tests/specs/js/module/comments.js
```diff
@@ -70,4 +70,6 @@ function test /* spacing before parentheses*/ () {}
 a /* before member */.test;
 statement /* comment */;
 
+/* leading */ "test" /*trailing*/ ;
```
ematipico, line 73: Can we add two new cases? One with newlines and one with skipped trivias

ematipico (commented):
No particular concerns, but I would like to set on stone one styling of formatting stuff. We are mixing `write!` and `.fmt()`. I suppose the latter is less to write but I think having "one style" is better, especially for external contributors that see two things and don't know which one to use

crates/rome_formatter/src/builders.rs
```diff
@@ -6,6 +6,7 @@ use crate::{Buffer, VecBuffer};
 use rome_rowan::{Language, SyntaxNode, SyntaxToken, SyntaxTokenText, TextLen};
 use std::borrow::Cow;
 use std::cell::Cell;
+use std::fmt::Debug;
```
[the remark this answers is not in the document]
MichaReiser: Nope... that's the reason why it started to complain that `.fmt` is ambiguous now

crates/rome_formatter/src/lib.rs
MichaReiser: Oh, there's an index map... Didn't know it exists

crates/rome_formatter/src/token.rs
MichaReiser: We do have this case. Let's assume you want to parenthesize 

```javascript
/* leading */ @Controller() // this is a controller
class T {}
```

The formatting should only move the comments before the `(` token but not the skipped token trivia. So we want

```javascript
/* leading */ (
  @Controller() // this is a controller 
  class T {}
)
```

and not 

```javascript
/* leading */ @Controller // this is a controller (
  class T {}
)
```

because that's not valid syntax. 

How this works is that the code inserting the `(` formats the comments up to the first skipped token trivia (the `@Controller`). The skipped token trivia and the comments after the skipped token trivia get formatted when formatting the `class` token.

crates/rome_formatter/src/token.rs
```diff
@@ -79,6 +80,218 @@ where
...
+
+        let is_last_content_inline_comment = f.state().is_last_content_inline_comment();
+
+        if needs_space_between_comments_and_token(
+            &comments,
+            self.kind,
+            is_last_content_inline_comment,
+            self.style,
+        ) {
+            space_token().fmt(f)?;
```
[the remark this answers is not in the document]
MichaReiser: I started using `fmt` because it's less to write when formatting a single element and may produce slightly simpler code. But I can see the point how it might be easier for new contributors.

crates/rome_formatter/src/token.rs
MichaReiser: I can rename the function but I need the formatter and I prefer that users don't have to call a method to make it work correctly.

crates/rome_js_formatter/tests/specs/js/module/comments.js
MichaReiser: I don't think I can come up with a test case where we parenthesize an expression that uses decorators. 

The only possible expression that I can think of is a class expression with decorators, but we don't support decorators in expression positions... 

I added a test for a case where there's a new line between the leading comment and the string literal token.

ematipico: No problem, it's mostly to cover the cases covered by the PR

PR #2719, opened 2022-06-16
fix(rome_js_semantic): remove dbg from semantic tests

xunilrj:
Some tools seems to display ```dbg!(...)``` even when tests succeed (clion is an example). So this PR removes all printing of tests.

### DIFF
--- a/crates/rome_js_semantic/src/tests/assertions.rs
+++ b/crates/rome_js_semantic/src/tests/assertions.rs
@@ -115,11 +115,6 @@ pub fn assert(code: &str, test_name: &str) {
 
     let assertions = SemanticAssertions::from_root(r.tree(), code, test_name);
 
-    // Print info to help debug when a test fail
-
-    dbg!(&events_by_pos);
-    dbg!(&assertions);
-
     // check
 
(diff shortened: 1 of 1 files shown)


crates/rome_js_semantic/src/tests/assertions.rs
```diff
@@ -115,11 +115,6 @@ pub fn assert(code: &str, test_name: &str) {
 
     let assertions = SemanticAssertions::from_root(r.tree(), code, test_name);
 
-    // Print info to help debug when a test fail
-
-    dbg!(&events_by_pos);
```
MichaReiser, line 120: Is there a way to only print if the `assertions.check` fail? Maybe inside of `assertions.check` but I'm not familiar enough with the code

MichaReiser (approved):
Thx

crates/rome_js_semantic/src/tests/assertions.rs
xunilrj: Yes, but I would need to change this a little bit because now I am panicking.
I can try this later.

PR #2721, opened 2022-06-16
feat(rome_js_formatter): apply existing layout strategy to assignments

ematipico:
## Summary

This PR fixes https://github.com/rome/tools/issues/2410

Prettier uses the layout strategy for assignments, which was introduced to format object members. 

This PR changes their name to be more generic. I didn't add the other layouts because they were not needed to fix the cases highlighted in the issue. I will introduce the other layouts in next PRs.

This PR adds an additional check to the condition `is_break_after_colon`, which covers a really specific case which is:

```js
const a = 
	// coment
	something;
```

It checks if a leading trivia comment is lead by newlines.

## Test Plan

Added new cases to cover the `Fluid` and `BreakAfterColon` cases.

PR:
**File Based Average Prettier Similarity**: 73.32%  
**Line Based Average Prettier Similarity**: 67.72%  

`main`
**File Based Average Prettier Similarity**: 73.25%  
**Line Based Average Prettier Similarity**: 67.60%

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/assignment_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/assignment_expression.rs
@@ -17,6 +17,7 @@ impl FormatNodeFields<JsAssignmentExpression> for FormatNodeRule<JsAssignmentExp
         let left = left?;
         let layout = compute_expression_layout(f, None, &right)?;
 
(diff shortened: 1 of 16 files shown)


ematipico:
Closing in favour of https://github.com/rome/tools/pull/2698

PR #2724, opened 2022-06-17
refactor(rome_formatter): `GroupElementsBuffer`

MichaReiser:
It's necessary to move leading comments outside a group to prevent that the comments expand the enclosing group:

```javascript
/* a comment */
[1]
```

The `/* a comment */` belongs to the `[` group token that is part of a group wrapping the whole `[1]` expression. It's important that the comment `/* a comment */` gets moved out of the group element to avoid the `[1]` group expanding because of the line break inserted by the comment.

The `group_elements` implementation so far used `split_trivia` to accomplish this. This PR changes `group_elements` to use a custom buffer that splits the leading trivia while it's being written. 

I decided to change Group to store a Box<[FormatElement]> to avoid boxing List elements.

## Test Plan
`cargo test`

### DIFF
--- a/crates/rome_formatter/src/arguments.rs
+++ b/crates/rome_formatter/src/arguments.rs
@@ -144,10 +144,10 @@ mod tests {
                 FormatElement::Space,
                 FormatElement::Token(Token::Static { text: "a" }),
                 FormatElement::Space,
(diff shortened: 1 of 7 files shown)


MichaReiser:
!bench_formatter

ematipico (approved):

ematipico (approved):

PR #2725, opened 2022-06-17
feature(rome_js_semantic): extraction of reference (read) events 

xunilrj:
This PR is the part that matches declarations and scopes of the main PR (https://github.com/rome/tools/pull/2489).

Now we are matching references and declarations on the naive case.

This is the first time we need to match the content of the code, so to avoid allocating strings and spreading a lifetime everywhere, I am storing only the hash of the ```&str```. 

In theory, we could use interning, but I don´t need the values of string slices (yet), no this is actually faster.


crates/rome_js_semantic/src/events.rs
```diff
@@ -1,16 +1,21 @@
 //! Events emitted by the [SemanticEventExtractor] which are then constructed into the Semantic Model
 
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
```
IWANABETHATGUY, line 3: The std hash algorithm is designed for  protecting against attacks such as HashDoS.Would it be better use hash like `rustc_hash` or `hashbrown` ?

IWANABETHATGUY: https://doc.rust-lang.org/std/collections/struct.HashMap.html

crates/rome_js_semantic/src/events.rs
```diff
@@ -87,31 +101,55 @@ impl SemanticEvent {
 pub struct SemanticEventExtractor {
     stash: VecDeque<SemanticEvent>,
     scopes: Vec<Scope>,
+    declared_names: HashMap<StrHash, TextRange>,
```
IWANABETHATGUY, line 104: What if declarations have same name but  in different scope?

IWANABETHATGUY:
Also I noticed that we have 178 regression here? Is that accetable?

crates/rome_js_semantic/src/events.rs
xunilrj: Let me test to see if there is a difference in performance.

MichaReiser: We use `hashbrown` with `FxHash` in the `TreeBuilder`.

crates/rome_js_semantic/src/events.rs
xunilrj: Do you mean by shadowing one another or in another case? The former I am dealing with.

xunilrj:
> Also I noticed that we have 178 regression here? Is that accetable?

No. Before the refactoring to please ```clippy``` it was actually passing more. Not sure what happened.

crates/rome_js_semantic/src/events.rs
IWANABETHATGUY: Yeah.
what about those cases: 
```js
var a = 3;
var a = 3;
``` 

```js
var a = 2000;
function a() {
}
```

crates/rome_js_semantic/src/events.rs
xunilrj: Using https://github.com/tkaitchuck/ahash was indeed faster. We do not have a benchmark for the semantic model yet, but I will use it anyways. We can always revisit this later.

Nice call!

crates/rome_js_semantic/src/events.rs
xunilrj: Ah... ok. I am ignoring "var" strangeness in this PR.
It will be treated in the next PR together with hoisting.

crates/rome_js_semantic/src/events.rs
leops: Well we actually do want to guard against DOS attacks, for instance if Rome is to be used in CI context to run checks over PRs opened by external contributors we want to make sure it's not possible to craft malicious code that causes the analysis take a huge amount of time to complete and cost CI time to open source projects

crates/rome_js_semantic/src/events.rs
```diff
@@ -182,8 +213,64 @@ impl SemanticEventExtractor {
...
+            scope_started_at,
+        });
+    }
+
+    fn get_declaration_range_of(&self, name_token: &JsSyntaxToken) -> Option<&TextRange> {
+        let name_hash = Self::hash(name_token.text_trimmed());
+        self.declared_names.get(&name_hash)
+    }
+
+    fn hash(name: &str) -> StrHash {
```
leops, line 268: That seems dangerous, the chance for a hash collision is quite small with a 64 bits hash but it exists, and I think if it every happens it's going to be very hard to debug

crates/rome_js_semantic/src/events.rs
xunilrj: Given that we are already using ```rustc_hash```, I will actually use it here also.

It is equivalent in performance with ```ahash```: https://nnethercote.github.io/perf-book/hashing.html

IWANABETHATGUY: > Well we actually do want to guard against DOS attacks, for instance if Rome is to be used in CI context to run checks over PRs opened by external contributors we want to make sure it's not possible to craft malicious code that causes the analysis take a huge amount of time to complete and cost CI time to open source projects

But I can't think of such a scenario, If run rome ci in github, IIRC the first time contributor need member or owner to trigger the CI.

ematipico: The readme of the library says

>  In the compiler we're not really worried about DOS attempts, so we use a fast non-cryptographic hash.

So. is it safe to use it?

crates/rome_js_semantic/src/events.rs
```diff
@@ -87,31 +101,55 @@ impl SemanticEvent {
...
+    declared_names: HashMap<StrHash, TextRange>,
+}
+
+struct ScopeDeclaration {
+    name_hash: StrHash,
 }
 
 struct Scope {
     started_at: TextSize,
+    declared: Vec<ScopeDeclaration>,
```
ematipico, line 113: I think we should add some documentation

crates/rome_js_semantic/src/events.rs
```diff
@@ -87,31 +101,55 @@ impl SemanticEvent {
...
+    shadowed: Vec<(StrHash, TextRange)>,
```
ematipico, line 114: I think we should add some documentation

crates/rome_js_semantic/src/events.rs
```diff
@@ -87,31 +101,55 @@ impl SemanticEvent {
 pub struct SemanticEventExtractor {
     stash: VecDeque<SemanticEvent>,
     scopes: Vec<Scope>,
+    declared_names: HashMap<StrHash, TextRange>,
+}
+
+struct ScopeDeclaration {
```
ematipico, line 107: I think we should call this struct `BindingDeclaration`, which is closer to what it tracks and the name helps to understand what's it for.

crates/rome_js_semantic/src/events.rs
ematipico: `bindings`?

crates/rome_js_semantic/src/events.rs
```diff
@@ -182,8 +213,64 @@ impl SemanticEventExtractor {
...
+        // We should at least have the global scope
+        debug_assert!(!self.scopes.is_empty());
+
+        match self.scopes.last_mut() {
+            None => unreachable!(),
+            Some(scope) => scope,
         }
     }
+
+    fn declare_name(&mut self, name_token: &JsSyntaxToken) {
```
ematipico, line 239: `track_binding` or `create_binding`?

crates/rome_js_semantic/src/events.rs
```diff
@@ -182,8 +213,64 @@ impl SemanticEventExtractor {
...
+        current_scope.shadowed.extend(shadowed);
+        let scope_started_at = current_scope.started_at;
+
+        self.stash.push_back(SemanticEvent::DeclarationFound {
+            range: declaration_range,
+            scope_started_at,
+        });
+    }
+
+    fn get_declaration_range_of(&self, name_token: &JsSyntaxToken) -> Option<&TextRange> {
```
ematipico, line 263: `get_binding_range`?

crates/rome_js_semantic/src/events.rs
```diff
@@ -182,8 +213,64 @@ impl SemanticEventExtractor {
...
+            // return all shadowed names
+            for (name, range) in scope.shadowed {
+                self.declared_names.insert(name, range);
+            }
+        }
+    }
+
+    fn current_scope_mut(&mut self) -> &mut Scope {
+        // We should at least have the global scope
+        debug_assert!(!self.scopes.is_empty());
```
ematipico, line 231: Shouldn't the global scope be tracked by itself? I mean, generally each scope should have access to it, regardless. So how a scope knows how to access to the global scope? 

Sure, it's always be the first one, but is it the correct design?

crates/rome_js_semantic/src/events.rs
```diff
@@ -19,14 +24,22 @@ pub enum SemanticEvent {
         scope_started_at: TextSize,
     },
 
+    /// Signifies that a symbol value is being read.
```
ematipico, line 27:
```suggestion
    /// It tracks when a symbol is read.
```

crates/rome_js_semantic/src/events.rs
```diff
@@ -19,14 +24,22 @@ pub enum SemanticEvent {
...
+    /// Generated for:
+    /// - All reference identifiers
+    Read {
+        range: TextRange,
+        declaration_at: Option<TextRange>,
```
ematipico, line 32: Can we add some documentation? 

Also, how would this work for cases like:
```js
var a;
b = a;
c = a;
```

Are we going to have two `SemanticEvent::Read`? If so, in this case `declaration_at` of each event will have the same value of `TextRange` which is not ideal...  wouldn't it make sense to have an `Rc` of the `TextRange` of where the binding was declared?

crates/rome_js_semantic/src/events.rs
leops: > But I can't think of such a scenario, If run rome ci in github, IIRC the first time contributor need member or owner to trigger the CI.

That's one layer of defense but it can be defeated if the attack is made innocent looking, hidden in a large refactor or included in a third-party dependency as the JS ecosystem has already demonstrated vulnerabilities to supply chain attacks in the past.
Anyway DOS attacks are probably not the most problematic issue, but DOS protection also happens to make crafting predictable hash collisions a lot harder, and since hash collisions would cause the semantic analysis to misbehave this could also be a potential attack surface as it might make our eventual compiler / bundler misinterpret a malicious piece of code and generates an output that executes differently.

crates/rome_js_semantic/src/events.rs
ematipico: For example, the root scope should contain all the globals (`window`, `Array`, etc.) and they don't have any `declared_at`, still, they are bindings and they can be read.

crates/rome_js_semantic/src/events.rs
xunilrj: I am using ```&str``` directly through ```SyntaxTokenText```. It is going to be a little bit slower, but we can think about optimizations later.

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

crates/rome_js_semantic/src/events.rs
xunilrj: Went with ```push_binding_at_current_scope```; hopefully is more clear now.

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

crates/rome_js_semantic/src/events.rs
xunilrj: Yes, but not this ```Scope``` struct. This one is just a helper to iterate the tree.
After this iteration, references like ```window```, for example, will be "unresolved". Then another process will link the current file with a global definition file. And only then we will be able to match them.

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

crates/rome_js_semantic/src/events.rs
xunilrj: > Are we going to have two SemanticEvent::Read?

Yes

> will have the same value of TextRange which is not ideal... wouldn't it make sense to have an Rc of the TextRange of where the binding was declared?

Not sure I understand why this is not ideal.
Any specific reason to have a ```Rc```?

Also, I am trying to make events as simple as possible, so they don´t allocate. Ideally, they all will be ```Copy```.

For example. These are the events currently being generated for this case.

```
ScopeStarted { range: 0..46 }
DeclarationFound { range: 4..11, scope_started_at: 0 }
Read { range: 17..28, declaration_at: Some(4..11) }
Read { range: 34..45, declaration_at: Some(4..11) }
ScopeEnded { range: 0..46, started_at: 0 }
```

crates/rome_js_semantic/src/events.rs
ematipico: Very clear! Thank you!

crates/rome_js_semantic/src/events.rs
ematipico: > Read { range: 17..28, declaration_at: Some(4..11) }
Read { range: 34..45, declaration_at: Some(4..11) }

So, the question was about these two `Some(4..11)`. In this case, do we allocate two times the same information? 

Since this information comes from:

> DeclarationFound { range: 4..11, scope_started_at: 0 }

I was wondering why we can't have a reference to `range: 4..11` instead. 

Not trying to suggest a solution, just wondering if it's feasible to have such architecture or not. No need to implement it here and now, it it makes sense.

leops: The `TextRange` is a simple tuple of two `u32` meaning on 64-bits architectures it has the same size as `&TextRange` (or `Rc<TextRange>`), so storing the range behind a reference would only increase the memory usage

IWANABETHATGUY: > 

I am not sure if it is because of memory alignment, I tested two versions of `Read`, https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=434f627fce3bc349543fcf0daf69f7eb, `Rc` would be smaller.

IWANABETHATGUY: > Ah. I get it now.
> 
> Today, `SemanticEvent` is:
> 
> ```
> print-type-size type: `events::SemanticEvent`: 24 bytes, alignment: 4 bytes
> print-type-size     discriminant: 4 bytes
> print-type-size     variant `Read`: 20 bytes
> print-type-size         field `.range`: 8 bytes
> print-type-size         field `.declaration_at`: 12 bytes
> print-type-size     variant `DeclarationFound`: 12 bytes
> print-type-size         field `.range`: 8 bytes
> print-type-size         field `.scope_started_at`: 4 bytes
> print-type-size     variant `ScopeEnded`: 12 bytes
> print-type-size         field `.range`: 8 bytes
> print-type-size         field `.started_at`: 4 bytes
> print-type-size     variant `ScopeStarted`: 8 bytes
> print-type-size         field `.range`: 8 bytes
> 
> print-type-size type: `std::option::Option<rome_js_syntax::TextRange>`: 12 bytes, alignment: 4 bytes
> print-type-size     discriminant: 4 bytes
> print-type-size     variant `Some`: 8 bytes
> print-type-size         field `.0`: 8 bytes
> print-type-size     variant `None`: 0 bytes
> ```
> 
> This means that `TextRange` is already 8 bytes. Anything sharing will be or equal or bigger. The same problem will appear with text/code. That is why I return range everywhere. Don´t allocate, nor intern anything.

Would you mind telling me how could print the memory layout like this?

ematipico: Thank you for clarifying!

xunilrj: > Would you mind telling me how could print the memory layout like this?

```
RUSTFLAGS=-Zprint-type-sizes cargo +nightly build -p rome_js_semantic --release > types.txt
```

ematipico (approved):

crates/rome_js_semantic/src/events.rs
IWANABETHATGUY: > > Would you mind telling me how could print the memory layout like this?
> 
> ```
> RUSTFLAGS=-Zprint-type-sizes cargo +nightly build -p rome_js_semantic --release > types.txt
> ```

Thanks!!

Issue #2424, opened 2022-04-13
Object pattern formatting


MichaReiser:
[...]

PR #2729, opened 2022-06-17
feat(rome_js_formatter): object pattern formatting

ematipico:
## Summary

This PR advances works around "assignments like", by partially implementing #2424

Here's the main works done:
- prettier formats the object patterns `const { b } = a;` and `const a = { b }` using the same function, this is because in their AST they are treated with the same AST nodes. Rome used two different nodes: `JsObjectAssignmentPattern` and `JsObjectBindingPattern`. In this PR I created a new union called `JsObjectPatternLike` where we apply the same formatting for both patterns;
- implemented two new layout modes: `BreakLeftHandSide` and `ChainTailArrowChain`. The first had required some changes inside the object pattern formatting, which are included in this PR, the second one won't have the desidered changes because of how things are built. ~When `ChainTailArrowChain`, prettier signals this layout to another function, which reads it and applies another layout. [Here's the code that reads the layout](https://github.com/prettier/prettier/blob/main/src/language-js/print/function.js#L257-L259). I'd prefer to figure this out later, outside of this PR (fortunately we don't break any existing code)~ Turns out the implementation was easier than expected, and it is now implemented!
- created a new union inside `JsAnyAssignmentLike`, called `RightAssignmentLike`. This was needed because the right hand side of an assignment like can be an `JsAnyExpression` or `JsAnyAssignmentPattern`;
- This PR **doesn't** cover **all the cases** of when deducting when to apply `BreakLeftHandSide`, because the rest of the checks are done on other types of nodes (type aliases and type parameters), which are not covered in this PR.

## Test Plan

I created new test cases that should cover the new layouts


crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -205,28 +382,47 @@ impl JsAnyAssignmentLike {
...
+            }
         }
 
         Ok(false)
     }
+
+    /// Particular function that checks if the left hand side of a [JsAnyAssignmentLike] should
+    /// be broken on multiple lines
+    fn should_break_left_hand_side(&self) -> SyntaxResult<bool> {
+        // TODO: here we have to add the check for variable declarator too
```
denbezrukov, line 413: Added in https://github.com/rome/tools/pull/2737

crates/rome_js_formatter/src/utils/object_pattern_like.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        if should_break_properties {
+            write!(
+                f,
+                [
+                    &self.l_curly_token()?.format(),
+                    group_elements(&block_indent(&format_with(|f| {
+                        self.write_properties(f)
+                    }),),),
+                    &self.r_curly_token()?.format(),
```
ematipico, line 114: At the moment prettier uses `group({ shouldBreak: true })` to format this branch, but we don't have such IR at the moment, so I thought this was the only way to match it.

Not sure how it will play out.

MichaReiser: Does the behaviour differ if you don't add a group? Because the `block_indent` will always force the content to break. 

If the behaviour does differ (would be interesting to understand how), then you can do `group_elements(expand_parent()... )`

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -122,6 +243,32 @@ impl JsAnyAssignmentLike {
...
+            JsAnyAssignmentLike::JsPropertyObjectMember(property) => {
+                let value = property.value()?;
+                write!(f, [value.format()])
+            }
+            JsAnyAssignmentLike::JsAssignmentExpression(assignment) => {
+                let right = assignment.right()?;
+                write!(f, [space_token(), right.format()])
+            }
+            JsAnyAssignmentLike::JsObjectAssignmentPatternProperty(property) => {
+                let pattern = property.pattern()?;
```
MichaReiser, line 264: Do I understand this correctly that the formatting will no longer call into `property.format()`? This might accidentally by-pass the suppression formatting. 

Can you add a few test cases with `rome-ignore` comments that start right before a property to make sure they get still ignored.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -310,27 +508,50 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
                         f,
                         [group_elements(&indent(&format_args![
                             soft_line_break_or_space(),
-                            right.format()
+                            format_with(|f| { self.write_right(f) }),
```
MichaReiser, line 511: Nit: You can avoid some of the repetition by declaring 

```
let right = format_with(|f| self.write_right(f))`
```

Outside of the match so that you can then use `right` instead of the `format_with`... everywhere (and `right.memoized()`)

ematipico: >Outside of the match so that you can then use right instead of the format_with... everywhere (and right.memoized())

I tried this approach, but using `format_with` inside the branches of a match gives a compiler error about "branches have different return types", using the `format_with` outside the match, will trigger some other compiler error

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: Well, it's also true for the rest of the branches... and you raised a good point. I will add some test

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -188,10 +374,29 @@ impl JsAnyAssignmentLike {
...
-                Some(AssignmentLikeLayout::ChainTail)
-            } else {
+            if !right_is_tail {
                 Some(AssignmentLikeLayout::Chain)
+            } else {
+                match right {
+                    RightAssignmentLike::JsAnyExpression(
+                        JsAnyExpression::JsArrowFunctionExpression(arrow),
+                    ) => {
+                        dbg!("here");
```
MichaReiser, line 384: No dbg here ;)

crates/rome_js_formatter/tests/specs/js/module/assignment/assignment.js
```diff
@@ -116,3 +116,17 @@ bifornCringerMoshedPerplex = bifornCringerMoshedPerplexSawder = arrayOfNumb = a
...
+a = {
+    a: { t: c = b }, loreum, ipsurm
+} = {}
+
+a =
+// rome-ignore format: test
+{
+    a: { t: c = b }, loreum, ipsurm
+} = {}
+
```
MichaReiser, line 132: Can you add a test where you only ignore a property and not the whole object. 

```
a =
{
    // rome-ignore format: test
    a: { t: c = b }, loreum, ipsurm
} = {}
```

MichaReiser (approved):

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -310,25 +533,39 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
                         ])),]
                     ]
                 }
                 AssignmentLikeLayout::NeverBreakAfterOperator => {
-                    write![f, [space_token(), right.format(),]]
+                    write![f, [space_token(), right,]]
+                }
+
+                AssignmentLikeLayout::BreakLeftHandSide => {
+                    write![f, [space_token(), group_elements(right),]]
```
denbezrukov, line 544: I've noticed that Prettier for `BreakLeftHandSide` variant doesn't wrap left part with `group`.

- https://github.com/prettier/prettier/blob/b0201e01ef99db799eb3716f15b7dfedb0a2e62b/src/language-js/print/assignment.js#L59

Do we need the same behaviour?

ematipico: We're grouping the right part, which is Prettier's behaviour. At this point, the left part was already formatted/written

ematipico: Yes we do. Now, this problem is that we are stuck. In order to NOT to have a group in the left hand side, we need to check the layout, but if want the layout, we need to format the left hand side so we can compute the length of the key 😆

denbezrukov: 🤯🤯🤯
Is it possible to write in temp buffer left side and get length of the key for layout?

ematipico: Yes it's possible. I believe we have to apply this approach. It's gets trickier because by using a custom buffer, we are stranded and we have to make sure that we write tokens in the right order. But we do this in few places already.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -502,14 +502,40 @@ pub(crate) fn has_new_line_before_comment(node: &JsSyntaxNode) -> bool {
...
+            // We create a temporary buffer because the left hand side has to conditionally add
+            // a group based on the layout, but the layout can be computed when the left hand
+            // side is actually formatted.
+            //
+            // 1. we crate a temporary buffer
+            // 2. we write the left hand side into the buffer and retrieve the `is_left_short` info
+            // 3. we compute the layout
+            // 4. we write the left node inside the main buffer based on the layout
+            let mut buffer = VecBuffer::new(f.state_mut());
+            let is_left_short = self.write_left(&mut buffer)?;
```
MichaReiser, line 514: You can do

```
write!(buffer, [format_with(|f| self.write_left(f)])?;
```

to get a formatter for a `buffer`

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -502,14 +502,40 @@ pub(crate) fn has_new_line_before_comment(node: &JsSyntaxNode) -> bool {
...
             // If not (for example, left is not an identifier), then let's fallback to false,
             // so we can continue the chain of checks
-            let is_left_short = self.write_left(f)?;
-            self.write_operator(f)?;
-
             let layout = self.layout(is_left_short)?;
 
+            let formatted_element = buffer.into_element();
+
+            if layout == AssignmentLikeLayout::BreakLeftHandSide {
```
MichaReiser, line 523: This works fine but has the downside (isn't specific to here, we have this all over the place), that the implementation needs to copy all elements that have been written to the `VecBuffer`. I'll spend a few minutes looking into a `InternBuffer` that returns an interned element.

PR #2731, opened 2022-06-18
feat(rome_js_analyze): noEmptyPattern

IWANABETHATGUY:
## Summary
1. Resolved https://github.com/rome/tools/issues/2730

## Test Plan
1. cargo test


crates/rome_js_analyze/src/analyzers/no_empty_pattern.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// ```js
+    /// function foo({a = {}}) {}
+    /// ```
+    ///
+    /// ```js
+    /// function foo({a = []}) {}
+    /// ```
+    /// ```js
+    /// var [a] = foo;
+    /// ```
```
ematipico, line 30: I believe all these cases can we wrapped inside a single `js` code block.

crates/rome_js_analyze/src/analyzers/no_empty_pattern.rs
```diff
@@ -0,0 +1,117 @@
...
+use rome_console::markup;
+use rome_js_syntax::{JsArrayBindingPattern, JsObjectBindingPattern};
+use rome_rowan::{declare_node_union, AstNode, AstSeparatedList};
+
+declare_rule! {
+    /// Disallows empty destructuring patterns.
+    /// ## Examples
+    ///
+    /// ### Valid
+    ///
```
ematipico, line 11:
```suggestion
    /// ### Valid
    /// The following cases are valid because they create new bindings.
```

crates/rome_js_analyze/src/analyzers/no_empty_pattern.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// function foo({a = []}) {}
+    /// ```
+    /// ```js
+    /// var [a] = foo;
+    /// ```
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// var {} = foo;
```
ematipico, line 35: I believe we can reduce the cases to just a few:

```js
var {} = foo;
var {a: {}} = foo;
function foo({}) {};
```

I believe these two are sufficient to explain the use case

ematipico (approved):

crates/rome_js_analyze/src/analyzers/no_empty_pattern.rs
IWANABETHATGUY: Done

leops (approved):

PR #2733, opened 2022-06-18
feat (rome_js_analyze): noAsyncPromiseExecutor

### LINKED ISSUE #2732: 📎 noAsyncPromiseExecutor
### Description

https://eslint.org/docs/rules/no-async-promise-executor

IWANABETHATGUY:
## Summary
1. Resolved #2732 

## Test Plan
1. `cargo test`

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
@@ -1,0 +1,83 @@
+use rome_analyze::{
+    context::RuleContext, declare_rule, ActionCategory, Rule, RuleCategory, RuleDiagnostic,
+};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_factory::make;
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyStatement, JsArrowFunctionExpression, JsForStatement,
+    JsForStatementFields, JsFunctionExpression, JsNewExpression, JsNewExpressionFields, T,
+};
+use rome_rowan::{declare_node_union, AstNode, AstNodeExt, AstSeparatedList};
+
+use crate::JsRuleAction;
+
+declare_rule! {
+    /// Disallows using an async function as a Promise executor.
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// for (; x.running;) {
+    ///     x.step();
+    /// }
+    /// ```
+    pub(crate) NoAsyncPromiseExecutor = "noAsyncPromiseExecutor"
+}
+
+impl Rule for NoAsyncPromiseExecutor {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = JsNewExpression;
+    type State = JsAnyFunctionExpressionLike;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let node = ctx.query();
+        let JsNewExpressionFields {
+            new_token: _,
+            callee,
+            type_arguments: _,
+            arguments,
+        } = node.as_fields();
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
@leops , @ematipico , would you mind helping for reviewing the code?

crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// ```
+    /// ```js
+    ///   new Promise((resolve, reject) => {}, async function unrelated() {})
+    /// ```
+    /// ```js
+    ///   new Foo(async (resolve, reject) => {})
+    /// ```
+    /// ```js
+    /// new Foo((( (resolve, reject) => {} )))
+    /// ```
```
ematipico, line 26: One single block `js` should be enough

crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
```diff
@@ -0,0 +1,111 @@
+use rome_analyze::{context::RuleContext, declare_rule, Rule, RuleCategory, RuleDiagnostic};
+use rome_console::markup;
+use rome_js_syntax::{
+    JsAnyExpression, JsArrowFunctionExpression, JsFunctionExpression, JsNewExpression,
+    JsNewExpressionFields,
+};
+use rome_rowan::{declare_node_union, AstNode, AstSeparatedList};
+
+declare_rule! {
+    /// Disallows using an async function as a Promise executor.
```
ematipico, line 10: Maybe we should expand more on **why** this rule exists. Eslint has a great a explanation, maybe we should take something from their website: https://eslint.org/docs/rules/no-async-promise-executor

crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
```diff
@@ -0,0 +1,111 @@
...
+        } else {
+            None
+        }
+    }
+
+    fn diagnostic(_: &RuleContext<Self>, state: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic::warning(
+            state.range(),
+            markup! {
+                "Promise executor functions should not be async."
```
ematipico, line 81:
```suggestion
                "Promise executor functions should not be `async`."
```

ematipico (approved):

crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
```diff
@@ -0,0 +1,111 @@
...
+            new_token: _,
+            callee,
+            type_arguments: _,
+            arguments,
+        } = node.as_fields();
+        let callee = callee.ok()?;
+        let is_promise_constructor = callee
+            .as_js_identifier_expression()
+            .and_then(|ident| ident.name().ok())
+            .map_or(false, |name| name.syntax().text() == "Promise");
```
leops, line 59: This should use `text_trimmed` here instead of `text` to avoid including the trivia for the token in the check

crates/rome_js_analyze/src/analyzers/no_async_promise_executor.rs
```diff
@@ -0,0 +1,111 @@
...
+        JsAnyExpression::JsParenthesizedExpression(expr) => {
+            let inner_expression = expr.expression().ok()?;
+            get_async_function_expression_like(&inner_expression)
+        }
+        _ => None,
+    }
+}
+
+declare_node_union! {
+    pub(crate) JsAnyFunctionExpressionLike = JsFunctionExpression | JsArrowFunctionExpression
```
leops, line 107: Instead of declaring a new union this could use `JsAnyFunction`, which has an `async_token` method to lookup whether the function is async already

leops (approved):

PR #2734, opened 2022-06-18
feat (xtask): add docgen for rome_analyze

IWANABETHATGUY:
## Summary
1. add xtask for rome_js_analyze docgen
2. Currently, we need to use `declare_rule` for automatically lint doc generation, for example
```rs
declare_rule! {
    /// Disallow comparing against `-0`
    ///
    /// ## Examples
    ///
    /// ### Invalid
    ///
    /// ```js,expect_diagnostic
    /// (1 >= -0)
    /// ```
    ///
    /// ### Valid
    ///
    /// ```js
    /// (1 >= 0)
    ///```
    pub(crate) NoCompareNegZero = "noCompareNegZero"
}
```
This would increase our workload since we need to split our test file item by item and adjust the cargo doc format.
3. With this xtask you could generate the doc in `declare_rule` automatically
 4. For example, run `cargo docgen` and paste the test code below into terminal :
```js
// valid
/foo {2}bar/;
/foo bar baz/;
/foo bar\tbaz/;
/foo /;
// invalid
/   /;
/  foo/;
/foo   /;
/foo  bar/;
/foo   bar    baz/;
/foo [ba]r  b(a|z)/;
```
the xtask would generate markdown below:
~~~
### Valid

```js
// valid
/foo {2}bar/;
```
```js
/foo bar baz/;
```
```js
/foo bar\tbaz/;
```
```js
/foo /;
```
### Invalid

```js,expect_diagnostic
// invalid
/   /;
```
```js,expect_diagnostic
/  foo/;
```
```js,expect_diagnostic
/foo   /;
```
```js,expect_diagnostic
/foo  bar/;
```
```js,expect_diagnostic
/foo   bar    baz/;
```
```js,expect_diagnostic
/foo [ba]r  b(a|z)/;
```
~~~

## Test Plan
1. run `cat crates/rome_js_analyze/tests/specs/noNegationElse.js | cargo docgen`


ematipico:
I am not sure what's the issue, but if you intend to generate documentation from the tests, we don't want that.

We used to do that, but realised that it's not ideal for the users. Many times, our tests are used to cover edge cases and regressions, which could lead to very strange code sometimes. The user is not interested in edge case, but they are interested in understanding straightaway the rule.

That's why we decided, now, to keep the documentation and the tests separate. 

I know it's more work for us, but this comes at the cost of user experience and documentation.

IWANABETHATGUY:
> I am not sure what's the issue, but if you intend to generate documentation from the tests, we don't want that.

Actually, it is just a helper crate. The input is stdio from the terminal, so we could use `cat` pipeline or other copy and paste to terminal way to generate doc. Feel free to close it, I would like to use it for myself if it is not so generic.

ematipico:
But am I correct about it works? 

You would take a file as a source (`.js` in this case) and pipe it into this command, and the command will generate in the command line the markdown?

I may see the value of this command, but would you explain to the users how to use this command? (because we have to document it somewhere, and it seems it's a command **done only for the linter**).  And how would you tell the user to use it only for some snippets?

I am not against the PR :) It might actually help, but how much? Why not writing things by hand is so tedious?

ematipico:
Closing, as we don't need this task at the moment.

PR #2737, opened 2022-06-20
feat(rome_js_formatter): variable declarator formatting

denbezrukov:
## Summary
This PR partially implements https://github.com/rome/tools/issues/2423

This PR adds 
- `OnlyLeft` layout to the list of layouts. This layout is hit, usually, when variable declarator doesn't have initializer.
- New conditional branch for `should_break_left_hand_side` when variable declaration `has_complex_type_annotation`. Prettier [hasComplexTypeAnnotation](https://github.com/prettier/prettier/blob/fde0b49d7866e203ca748c306808a87b7c15548f/src/language-js/print/assignment.js#L278-L298).
- New conditional branch for `should_not_indent_if_parent_indents`. Don't indent if `great_parent` is `JS_VARIABLE_DECLARATOR`.

## Open questions
There are problems with `BindExpression` and `RecordExpression`. I described them in comments.

## Test Plan

Add new tests specs


denbezrukov:
Will be ready for review after https://github.com/rome/tools/pull/2729

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -81,19 +88,70 @@ impl AnyObjectPattern {
...
+            LeftAssignmentLike::JsAnyBindingPattern(
+                JsAnyBindingPattern::JsObjectBindingPattern(node),
+            ) => Some(AnyObjectPattern::from(node.clone())),
             _ => None,
         }
     }
 }
 
+impl AnnotationLike {
+    fn has_complex_type_annotation(&self) -> SyntaxResult<bool> {
```
denbezrukov, line 108: https://github.com/prettier/prettier/blob/fde0b49d7866e203ca748c306808a87b7c15548f/src/language-js/print/assignment.js#L278

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -1,29 +1,36 @@
...
+        JsVariableDeclarator
+}
 
+declare_node_union! {
+    pub(crate) LeftAssignmentLike = JsAnyAssignmentPattern | JsAnyObjectMemberName | JsAnyBindingPattern
 }
 
 declare_node_union! {
-    pub(crate) LeftAssignmentLike = JsAnyAssignmentPattern | JsAnyObjectMemberName
+    pub(crate) AnnotationLike = TsAnyVariableAnnotation
```
denbezrukov, line 29: I'm not sure that we need new union for `TsAnyVariableAnnotation`.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -81,19 +88,70 @@ impl AnyObjectPattern {
...
+                            let is_complex_type = argument
+                                .as_ts_reference_type()
+                                .and_then(|reference_type| reference_type.type_arguments())
+                                .map_or(false, |type_arguments| {
+                                    type_arguments.ts_type_argument_list().len() > 0
+                                });
+
+                            is_complex_type
+                        });
+                    Some(argument_list_len > 1 && has_at_least_a_complex_type)
```
denbezrukov, line 135: Maybe it's better to move `argument_list_len` above `is_complex_type`?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -260,16 +260,24 @@ fn is_inside_parenthesis(current_node: &JsSyntaxNode) -> bool {
...
-            true
+    let parent = current_node.syntax().parent();
+    let parent_kind = current_node.syntax().parent().map(|node| node.kind());
+
+    let great_parent = parent.and_then(|parent| parent.parent());
+    let great_parent_kind = great_parent.map(|node| node.kind());
+
+    match (parent_kind, great_parent_kind) {
+        (Some(JsSyntaxKind::JS_PROPERTY_OBJECT_MEMBER), _)
+        | (Some(JsSyntaxKind::JS_INITIALIZER_CLAUSE), Some(JsSyntaxKind::JS_VARIABLE_DECLARATOR)) => {
```
denbezrukov, line 271: https://github.com/prettier/prettier/blob/fde0b49d7866e203ca748c306808a87b7c15548f/src/language-js/print/binaryish.js#L110

crates/rome_js_formatter/tests/specs/prettier/js/babel-plugins/record-tuple-record.js.snap
```diff
@@ -16,14 +16,14 @@ const record2 = #{...record1, b: 5};
 
 # Output
 ```js
-const record1 = #
```
denbezrukov, line 19: `RecordExpression`

I don't know what it's better options to handle this case. It seems that Rome has wrong AST for that.
[Rome playground](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=false&jsx=false#YwBvAG4AcwB0ACAAcgBlAGMAbwByAGQAMQAgAD0AIAAjAHsAYQA6ACAAMQAsACAAYgA6ACAAMgB9ADsA)
[Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAnOlsAmAjJgLyYDEwAhkpkQDSYBGdATAL4DcIDIEABxgBLdMlDVs2CAHcACpIRpkIagBsZ1AJ7K+zbNTABrODADK1ALZwAMsKhxkAM3Vo4eg8dNmBh+wHNkGGwAV3cQOEtmOAICGJtqKH8Q6n84ADEIbEtqGBEklWoQmAheEAALGEs1AHVy4Xg0XzA4MyUG4QA3Bq0VMDRdEHs3bBg5A38c51dwgCs0AA8zALU4AEUQiHhptTc+X2wRlWZqaLUygWx7GBrhAhhy5AAOAAZ96TcagwEVS7gRzqOPgAR028HGgmUKGoaAAtA4YjEyrhQcJcONUlMkC5duE3JZhEFQniVuswY5sTM+DBTrd7o8kGxqQZhGoAgBhCCWLERNAAVjKITcABVTlCcXsQJ0wgBJKBxWBmMBXIQAQXlZhgWlWOzcHA4QA)

ematipico: We don't support this syntax (check the errors tab) because it's not official yet. Prettier tends to support unsupported syntax

crates/rome_js_formatter/tests/specs/prettier/js/bind-expressions/long_name_method.js.snap
```diff
@@ -19,7 +19,7 @@ class X {
 ```js
 class X {
   constructor() {
-    this.testLongNameMethodAndSomethingElseLallala =
```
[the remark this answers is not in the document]
ematipico: You can ignore this syntax, we don't support it

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: In this case we don't need the union, so we can discard it. The macro would generate code that is essentially not needed

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: Like doing an early return? Yeah maybe it makes sense. I'll do the same in my PR too

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Removed

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Done

crates/rome_js_formatter/tests/specs/prettier/js/babel-plugins/record-tuple-record.js.snap
denbezrukov: Thank you (:

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -260,16 +260,24 @@ fn is_inside_parenthesis(current_node: &JsSyntaxNode) -> bool {
...
-
-    match parent_kind {
-        Some(JsSyntaxKind::JS_PROPERTY_OBJECT_MEMBER) => current_node
-            .as_expression()
-            .and_then(|expression| should_break_after_operator(expression).ok())
-            .unwrap_or(false),
-        Some(JsSyntaxKind::JS_RETURN_STATEMENT | JsSyntaxKind::JS_ARROW_FUNCTION_EXPRESSION) => {
-            true
+    let parent = current_node.syntax().parent();
+    let parent_kind = current_node.syntax().parent().map(|node| node.kind());
```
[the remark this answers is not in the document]
denbezrukov: Done

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -87,16 +90,71 @@ impl AnyObjectPattern {
...
             ) => Some(AnyObjectPattern::from(node.clone())),
+            LeftAssignmentLike::JsAnyBindingPattern(
+                JsAnyBindingPattern::JsObjectBindingPattern(node),
+            ) => Some(AnyObjectPattern::from(node.clone())),
             _ => None,
         }
     }
 }
 
+pub(crate) fn is_complex_type_annotation(
```
denbezrukov, line 106: https://github.com/prettier/prettier/blob/fde0b49d7866e203ca748c306808a87b7c15548f/src/language-js/print/assignment.js#L278

crates/rome_js_formatter/src/utils/binary_like_expression.rs
IWANABETHATGUY: I think you could add this reference above your code, so other contributors could know why would us write like this even without reviewing this pull request.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -87,16 +90,71 @@ impl AnyObjectPattern {
 }
 
 impl LeftAssignmentLike {
-    fn as_object_assignment_pattern(&self) -> Option<AnyObjectPattern> {
+    fn as_object_pattern(&self) -> Option<AnyObjectPattern> {
```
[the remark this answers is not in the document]
ematipico: I believe the change makes sense 💯

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -115,8 +177,13 @@ impl Format<JsFormatContext> for RightAssignmentLike {
 /// Assignment like are:
 /// - Assignment
 /// - Object property member
+/// - Variable declaration
 #[derive(Debug)]
 pub(crate) enum AssignmentLikeLayout {
+    /// This is a special layout usually used for variable declarations
+    /// This layout is hit, usually, when variable declarator doesn't have initializer:
```
ematipico, line 184: I believe you wanted to add an example? The phrase finishes with `:`
```suggestion
    /// This is a special layout usually used for variable declarations.
    /// This layout is hit, usually, when a [variable declarator](JsVariableDeclarator) doesn't have initializer:
```

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -183,14 +252,19 @@ pub(crate) enum AssignmentLikeLayout {
 }
 
 impl JsAnyAssignmentLike {
-    fn right(&self) -> SyntaxResult<RightAssignmentLike> {
+    fn right(&self) -> Option<RightAssignmentLike> {
```
ematipico, line 255: Is there a particular reason why we changed this into an `Option`? The reason why we want `SyntaxResult`, is to usually bubble up any syntax error and prevent formatting nodes that are syntactically wrong.

And if the right hand side is missing, it's usually a syntax error.

crates/rome_js_formatter/tests/specs/js/module/declarations/variable_declaration.js
```diff
@@ -0,0 +1,27 @@
+//break left-hand side layout
```
ematipico, line 1: Would you mind adding a test case with a suppression comment? We should make sure that the layout formatting is not applied to the node that is suppressed

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Yes, you're right. But for `JsVariableDeclarator` variant a right part is `initializer` and it has `Option<JsInitializerClause>` type. We can try to keep `SyntaxResult`:

-  make `fn right(&self) -> SyntaxResult<Option<RightAssignmentLike>>`. I have concerns that it will be more difficult to use.
- unwrap `Option<JsInitializerClause>` here and introduce `should_only_left` function which which will be a guarantee of safety.

ematipico: Yeah, I agree that having `SyntaxResult<Option<RightAssignmentLike>` would add a lot of branches/indirection/safety comments. Let's try you second suggested solution. Documenting `should_only_left` should help developers understand why we're doing this

denbezrukov: https://github.com/rome/tools/pull/2737/commits/12ada03d687baf8565042c825ae77645615a20b3

crates/rome_js_formatter/tests/specs/js/module/declarations/variable_declaration.js
denbezrukov: https://github.com/rome/tools/pull/2737/commits/c5223b95780c5fc5f668d8b193a0e7d8e75ebc27
Is it correct?
Could you please explain why we need this test? How could we by-pass a suppression comment?

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -188,10 +508,29 @@ impl JsAnyAssignmentLike {
...
-                Some(AssignmentLikeLayout::ChainTail)
-            } else {
+            if !right_is_tail {
                 Some(AssignmentLikeLayout::Chain)
+            } else {
+                match right {
+                    RightAssignmentLike::JsAnyExpression(
+                        JsAnyExpression::JsArrowFunctionExpression(arrow),
+                    ) => {
+                        dbg!("here");
```
MichaReiser, line 518: This shouldn't be in here :O

MichaReiser:
Small note. This PR seems to have added a roughly 10% performance regression. I would say this is kind of expected as it has to do many checks but I'm wondering if there are any hotspots that could be improved. We need to be careful to not slowly regress performance by a large margin.

crates/rome_js_formatter/src/utils/assignment_like.rs
MichaReiser: Too bad, I hoped that the performance regression comes from the `dbg` expression :disappointed:

denbezrukov: Do we have tools to diagnostic the performance issue?
How do you usually debug this?🙏🏼

MichaReiser: You can run `cargo bench_formatter` to profile the formatter. The command has two useful options

* `--save-baseline=<name>`: Stores the result under the given name. Useful to e.g. run the benchmarks on main first so that you can then compare against these base results. To compare benchmark, use `critcmp <name1> <name2>` (requires [critcmp](https://github.com/BurntSushi/critcmp))
* --filter=<name>` filters the files to benchmarks. I often go with `tex-chtml-full` because it seems to be big enough that performance improvements are visible but doesn't take as long to run as e.g. typescript. 

Finding bottlenecks inside the formatter can be difficult and the available tools vary depending on the platform you use:

* Mac: I very much like [cargo-instruments](https://crates.io/crates/cargo-instruments). 
* CLion has a built-in profiler for Linux and I believe MacOS. 

The [rust book](https://nnethercote.github.io/perf-book/profiling.html) lists some more available tools as well as how you can enable debug information in release builts (you want that to be able to make sense of the output)

MichaReiser: I just found [Firefox Profiler]. You can do profiling with (assuming linux and that perf is installed)


```bash
perf record -F999 --call-graph dwarf cargo bench_formatter --filter tex
perf script -F +pid > /tmp/test.perf
```

Then upload the file on the website. They also have [great documentation](https://profiler.firefox.com/docs/#/)

PR #2469, opened 2022-04-20
fix(rome_js_formatter): Fix stackoverflow on Windows for nested seque…

MichaReiser:
…nce expressions

Formatting deeply nested sequence expressions is causing a stack overflow on Windows because they are parsed left-recursive:

```js
a, b,c,d
```

Parses to

```yaml
JsExpressionStatement {
    expression: JsSequenceExpression {
        left: JsSequenceExpression {
            left: JsSequenceExpression {
                left: JsIdentifierExpression {
                    name: JsReferenceIdentifier {
                        value_token: IDENT@0..1 "a" [] [],
                    },
                },
                comma_token: COMMA@1..3 "," [] [Whitespace(" ")],
                right: JsIdentifierExpression {
                    name: JsReferenceIdentifier {
                        value_token: IDENT@3..4 "b" [] [],
                    },
                },
            },
            comma_token: COMMA@4..5 "," [] [],
            right: JsIdentifierExpression {
                name: JsReferenceIdentifier {
                    value_token: IDENT@5..6 "c" [] [],
                },
            },
        },
        comma_token: COMMA@6..7 "," [] [],
        right: JsIdentifierExpression {
            name: JsReferenceIdentifier {
                value_token: IDENT@7..8 "d" [] [],
            },
        },
    },
    semicolon_token: missing (optional),
}
```

This PR changes the formatting of sequence expression to unwind the recursion by using a loop
that first finds the left most sequence expression and then concatenates the expressions.

## test

```
cargo bench_formatter --filter d3 --criterion=false
```

No longer panics on Windows

PR #2738, opened 2022-06-20
fix(rome_js_formatter): Fix suppression command handling for `if-else` clause

MichaReiser:
The suppression comments are checked as part of a node's formatting. It is, therefore, required that the formatter calls `format` on every node or the formatter won't respect the suppression comment.

This hasn't been the case for else clauses where the `if` statement formatting also formatted its else clauses.

This PR changes this by moving the logic for formatting the `else` clause into the `else_clause` formatting and calling it from the `if` stmt formatting.

This PR further ensures that this requirement upholds for static and computed members as well as sequence expressions. It isn't strictly required for these (yet) because the suppression comment will be associated with the inner expression of that node but it would become important if comments get attached to nodes.

## Test Plan

Added a few more tests where it's possible to test the impression. I plan to add a new formatter state in debug that verifies that all nodes were handled but I can only do so after most call sites have been fixed. 

Verified that #2469 doesn't get re-introduced by using recursion to format sequence expressions


crates/rome_js_formatter/src/js/any/mod.rs
```diff
@@ -1,45 +1,45 @@
 //! Generated file, do not edit by hand, see `xtask/codegen`
 
-mod array_assignment_pattern_element;
```
MichaReiser, line 3: I changed the visibility to `pub(crate)` so that it becomes possible to share utility functions across modules (e.g. else clause uses the same alternate formatting as the if statement)

MichaReiser:
!bench_formatter

ematipico (commented):
Could you tell us more why we don't have an overflow of memory anymore? I mean, we restored some code that was leaking memory, and it seems all fine?

MichaReiser:
> Could you tell us more why we don't have an overflow of memory anymore? I mean, we restored some code that was leaking memory, and it seems all fine?

It wasn't the case that some code was leaking memory, just that we hit a recursion limit (on windows only). I'm not entirely sure. One thing I could think of is that Binary expression now formats the elements lazily when the whole tree has been flattened where we previously called into the formatting right away (in a potentially already deep stack). 

The other reason could be that the format refactoring helped to flatten the stack but my intention would be that it was the other way round.

ematipico (approved):

PR #2739, opened 2022-06-20
fix(rome_formatter): Group ID in fits function

MichaReiser:
## Summary

#2589 introduced `GroupId`s that add support to refer to a specific group when using `if_group_breaks` and `if_group_fits_on_line`.  However, the implementation doesn't yet respect the `group_id` when measuring if a group (or element) fits on a line. 

This PR ensures that the printer retrieves the id of the referenced group when measuring if the content of a group fits. 

## Test Plan

You can see now that the [numbers-trailing-comma.js.snap](https://github.com/rome/tools/compare/fix/group-id-in-fits?expand=1#diff-a094715543707300788a39ec8ce850d8e19d2c1fe84692e1337425ebf5162100) now no longer exceeds the line. The issue was that printing the last element and the `,` exceeded the print width.


ematipico (approved):

ematipico (commented):
Can we add a test case in **our** test suite?

PR #2742, opened 2022-06-20
feature(rome_js_formatter): Class formatting (including heritage)

MichaReiser:
This PR brings Rome's class formatting closer to Prettier's.

## Test

Manually verified the output of each updated snapshot test with prettier's to make sure they match (except for comment positioning)

**Before**:
File Based Average Prettier Similarity: 74.79%
Line Based Average Prettier Similarity: 69.89%

**After**:

File Based Average Prettier Similarity: 74.99%
Line Based Average Prettier Similarity: 70.40%

### DIFF
--- a/crates/rome_js_formatter/src/js/any/class.rs
+++ b/crates/rome_js_formatter/src/js/any/class.rs
@@ -13,6 +13,9 @@ impl FormatRule<JsAnyClass> for FormatJsAnyClass {
         let id = node.id();
         let extends = node.extends_clause();
         let implements_clause = node.implements_clause();
+        let id = id?;
+
+        let group_mode = should_group_class(node)?;
 
         if let Some(abstract_token) = abstract_token {
             write!(f, [abstract_token.format(), space_token()])?;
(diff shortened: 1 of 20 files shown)


ematipico (approved):
It looks good to me! Would you mind adding new test cases to cover in our test suite?

MichaReiser:
> It looks good to me! Would you mind adding new test cases to cover in our test suite?

I'm not sure how much value they bring because the prettier test suite already covers them well. I would literally end up copy pasting the test cases from prettier to ours and I believe that this only ends up generating noise.

ematipico:
> > It looks good to me! Would you mind adding new test cases to cover in our test suite?
> 
> I'm not sure how much value they bring because the prettier test suite already covers them well. I would literally end up copy pasting the test cases from prettier to ours and I believe that this only ends up generating noise.

Simply because one day in the future we might want to rely on our test suite because we don't want to follow prettier anymore. And if that happens (it will for many things), having our test suites is more reliable, and we cover us from possible regressions.

ematipico (approved):

PR #2745, opened 2022-06-20
feat(rome_js_analyze): useShorthandType

### LINKED ISSUE #2744: useShorthandArrayType
### Description

https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/ts/preferShorthandArrayType.md

IWANABETHATGUY:
## Summary
1. Resolved #2744 

## Test Plan
1. cargo test


crates/rome_js_analyze/tests/specs/useShorthandArrayType.ts
```diff
@@ -0,0 +1,5 @@
+let valid: Array<foo>;
```
ematipico, line 1: These are valid cases:

```ts
let valid: Array<Foo | Bar>;
let valid: Array<keyof Bar>;
let valid: Array<foo | bar>;

```

xtask/lintdoc/src/main.rs
```diff
@@ -332,7 +332,10 @@ fn assert_lint(
                 "analysis returned multiple diagnostics"
             );
         } else {
```
IWANABETHATGUY, line 334: This is used for telling author which inline doc caused an issue

ematipico (approved):

crates/rome_js_analyze/src/analyzers/use_shorthand_array_type.rs
```diff
@@ -0,0 +1,150 @@
...
+            let element_type = match param {
+                TsType::TsUnionType(_) => continue,
+                TsType::TsTypeOperatorType(_) => continue,
+                TsType::TsReferenceType(ty)
+                    if is_array_reference(&ty).unwrap_or(false)
+                        && ty.type_arguments().is_some() =>
+                {
+                    // SAFETY: We have checked the `ty.type_arguments` is `Some` in match guard
+                    convert_to_array_type(ty.type_arguments().unwrap())
+                }
```
ematipico, line 108: Wouldn't it be better to use 

```rs
if let Some(type_arguments) = ty.type_arguments() {
	if  is_array_reference(&ty).unwrap_or(false) {
		convert_to_array_type(ty.type_arguments())
	}
}
```

So we extract and check `type_arguments` without using the unsafe `unwrap`

crates/rome_js_analyze/src/analyzers/use_shorthand_array_type.rs
```diff
@@ -0,0 +1,150 @@
...
+        let name = identifier.value_token().ok()?;
+        Some(name.text_trimmed() == "Array")
+    })
+}
+
+fn convert_to_array_type(type_arguments: TsTypeArguments) -> Option<TsType> {
+    if type_arguments.ts_type_argument_list().len() > 0 {
+        let mut array_types = Vec::new();
+
+        for param in type_arguments.ts_type_argument_list().iter() {
```
ematipico, line 97: I believe we can use `filter_map` here, so we don't need to create the new vector `array_types`. 

The reason why Rome classic did have a separate array is mostly because there's no `filterMap` there 😛

ematipico (changes requested):
CI is failing

ematipico (approved):

IWANABETHATGUY:
> CI is failing

Addressed

leops (approved):

IWANABETHATGUY:
@ematipico , I think this pr is ready to merge.

PR #2757, opened 2022-06-21
refactor(rome_js_formatter): Support `FormatRule`s with options

MichaReiser:
## Summary

To this point, it hasn't been possible to customize the formatting of a `FormatRule`. The only way to pass additional data has been by creating an additional method on the struct that takes additional parameters and call that method instead. Doing so comes with the risk that this will accidentally by-pass the node's suppression check. 

This PR adds a way to `FormatRule` s to be parametrized by options which can be useful to e.g. pass a group id from one rule to another. I did so by

* Changing the `fmt` signature of `FormatRule` to take `self` by reference (so that the rule is able to access the options)
* Add a new `FormatRuleWithOptions` trait that rules with options can implement and that `FormatOwnedWithRule` and `FormatRefWithRule` automatically implement if the inner rule supports options. This makes it possible to do write `elements.format().with_options(options)` rather than using `format_with(|f| MyFormatRule::new(options).fmt(elements, f))` 
* Refactor `FormatNodeRule` to a trait so that each rule has its own struct that can have additional options

## Test Plan

`cargo test` 

`FormatJsArrayElementList` now uses the new options infrastructure.


crates/rome_js_formatter/src/js/lists/array_element_list.rs
```diff
@@ -1,36 +1,38 @@
...
 
-use crate::generated::FormatJsArrayElementList;
 use crate::utils::has_formatter_trivia;
 use rome_js_syntax::{JsAnyExpression, JsArrayElementList, JsSyntaxKind};
 use rome_rowan::{AstNode, AstSeparatedList};
 
-impl FormatRule<JsArrayElementList> for FormatJsArrayElementList {
-    type Context = JsFormatContext;
+#[derive(Debug, Clone, Default)]
+pub struct FormatJsArrayElementList {
```
MichaReiser, line 12: Format rule with custom options

MichaReiser:
!bench_formatter

MichaReiser:
!bench_formatter

MichaReiser:
> 
> ```
> formatter/checker.ts                     1.00    375.7±2.66ms     6.9 MB/sec    1.04    390.6±1.49ms     6.7 MB/sec
> formatter/compiler.js                    1.00    233.8±1.18ms     4.5 MB/sec    1.03    240.7±1.81ms     4.4 MB/sec
> formatter/d3.min.js                      1.00    187.7±0.82ms  1430.3 KB/sec    1.03    193.9±1.15ms  1384.5 KB/sec
> formatter/dojo.js                        1.00     13.1±0.06ms     5.2 MB/sec    1.02     13.4±0.15ms     5.1 MB/sec
> formatter/ios.d.ts                       1.00    240.4±1.47ms     7.8 MB/sec    1.05    252.3±1.13ms     7.4 MB/sec
> formatter/jquery.min.js                  1.00     50.7±0.17ms  1669.2 KB/sec    1.02     51.6±0.17ms  1640.5 KB/sec
> formatter/math.js                        1.00    363.6±2.40ms  1823.8 KB/sec    1.03    373.9±2.56ms  1773.4 KB/sec
> formatter/parser.ts                      1.00      8.0±0.02ms     6.0 MB/sec    1.03      8.3±0.02ms     5.8 MB/sec
> formatter/pixi.min.js                    1.00    203.8±1.34ms     2.2 MB/sec    1.04    211.6±0.98ms     2.1 MB/sec
> formatter/react-dom.production.min.js    1.00     62.2±0.20ms  1895.3 KB/sec    1.02     63.6±0.21ms  1853.3 KB/sec
> formatter/react.production.min.js        1.00      3.0±0.01ms     2.1 MB/sec    1.02      3.0±0.01ms     2.0 MB/sec
> formatter/router.ts                      1.00      6.1±0.03ms     9.7 MB/sec    1.03      6.3±0.03ms     9.4 MB/sec
> formatter/tex-chtml-full.js              1.00    467.6±1.20ms  1995.7 KB/sec    1.03    483.7±2.44ms  1929.3 KB/sec
> formatter/three.min.js                   1.00    237.3±0.59ms     2.5 MB/sec    1.05    249.0±3.71ms     2.4 MB/sec
> formatter/typescript.js                  1.00   1513.5±7.92ms     6.3 MB/sec    1.03   1556.1±7.23ms     6.1 MB/sec
> formatter/vue.global.prod.js             1.00     79.9±0.23ms  1543.4 KB/sec    1.02     81.6±0.28ms  1511.3 KB/sec
> ```

This is interesting. I also noticed that the `rome_js_formatter` crate increases by roughly 10Kb to 465KB. The only causes that I can think of are:

* `FormatNodeRule` is now a trait. But that shouldn't matter much because the compiler unrolled the generic node parameter before. 
* `FormatRule::fmt` now takes `self` as a parameter but I would have expected that this doesn't make a difference for the majority of `FormatRule`s because they are all zero sized types. 

@leops @xunilrj  do you have any idea where this regression might be coming from?

ematipico:
I noticed that you force pushed and the benchmarker is not decent. What did you change between the two benchs?

crates/rome_js_formatter/src/ts/types/void_type.rs
```diff
@@ -1,10 +1,13 @@
 use crate::prelude::*;
-use crate::FormatNodeFields;
+
 use rome_formatter::write;
 use rome_js_syntax::{TsVoidType, TsVoidTypeFields};
 
-impl FormatNodeFields<TsVoidType> for FormatNodeRule<TsVoidType> {
-    fn fmt_fields(node: &TsVoidType, f: &mut JsFormatter) -> FormatResult<()> {
+#[derive(Debug, Clone, Default)]
+pub struct FormatTsVoidType;
```
ematipico, line 7: Should everything be `pub`? Isn't `pub(crate)` enough?

crates/rome_js_formatter/docs/implement_the_formatter.md
```diff
@@ -63,24 +69,27 @@ This will automatically build and open a browser tab to the documentation.
    3. `format_extensions.rs`: with these traits, we give the ability to nodes and tokens to implements certain methods
    that are exposed based on its type. If you have a good IDE support, this feature will help you. For example:
    ```rust
-      impl FormatNodeFields<JsExportDefaultExpressionClause> for FormatNodeRule<JsExportDefaultExpressionClause> {
-   			fn fmt_fields(node: &JsExportDefaultExpressionClause, f: &mut JsFormatter) -> FormatResult<()> {
+   #[derive(Debug, Clone, Default)]
+   pub struct FormatJsExportDefaultExpressionClause;
```
ematipico, line 73: So now the structs are created inside the single file and in the `generated.rs` we have only the implementation of the traits?

ematipico (approved):

MichaReiser:
> I noticed that you force pushed and the benchmarker is now in a decent shape. What did you change between the two benchs?

I added an `#[inline]` to the `FormatRefWithRule` and `FormatOwnedWithRule` constructors.

crates/rome_js_formatter/src/ts/types/void_type.rs
MichaReiser: I used `pub(crate)` first with the same motivation you just mentioned but the problem is that the types are used as associated types in `AsFormat` and `IntoFormat` 

```
error[E0446]: crate-private type `FormatJsScript` in public interface
  --> crates/rome_js_formatter/src/generated.rs:12:5
   |
12 | /     type Format = FormatRefWithRule<
13 | |         'a,
14 | |         rome_js_syntax::JsScript,
15 | |         crate::js::auxiliary::script::FormatJsScript,
16 | |     >;
   | |______^ can't leak crate-private type
   |
  ::: crates/rome_js_formatter/src/js/auxiliary/script.rs:9:1
   |
9  |   pub(crate) struct FormatJsScript;
   |   --------------------------------- `FormatJsScript` declared as crate-private
```

crates/rome_js_formatter/docs/implement_the_formatter.md
MichaReiser: That's correct. The definition in the implementation file is necessary so that you can customise the rules fields without the codegen overriding it.

leops (approved):

PR #2759, opened 2022-06-21
feat(rome_js_analyze): useSimplifiedLogicExpression

### LINKED ISSUE #2749: useSimplifiedLogicalExpression

IWANABETHATGUY:
## Summary
1. Resolved #2749 

## Test Plan
1. cargo test


IWANABETHATGUY:
@ematipico , @leops , would you mind helping with reviewing?

crates/rome_js_analyze/src/analyzers/use_simplified_logic_expression.rs
```diff
@@ -0,0 +1,227 @@
...
+use rome_rowan::{AstNode, AstNodeExt};
+
+use crate::JsRuleAction;
+
+declare_rule! {
+    /// This rule simplified your logic expressions.
+    ///
+    /// ## Examples
+    ///
+    /// ### Valid
```
leops, line 20: The convention is to have the invalid cases first since those are the most helpful for users to understand what the rule is actually guarding for (and what the suggested fix is with the diagnostic snapshot). The valid cases can come later and could be treated as optional, as they're probably only needed to document unconventional cases a user might think the rule applies to when it does not

IWANABETHATGUY: Done

crates/rome_js_analyze/src/analyzers/use_simplified_logic_expression.rs
```diff
@@ -0,0 +1,226 @@
...
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyLiteralExpression, JsBooleanLiteralExpression, JsLogicalExpression,
+    JsUnaryExpression, JsUnaryOperator, T,
+};
+use rome_rowan::{AstNode, AstNodeExt};
+
+use crate::JsRuleAction;
+
+declare_rule! {
+    /// This rule simplified your logic expressions.
```
leops, line 16: Rome JS had "Discard redundant terms from logical expressions", this would be a more specific explanation of what the rule is doing (and could work as a message for the code action as well)

PR #2760, opened 2022-06-21
fix(rome_js_formatter): don't add trailing operator on rest arguments

### LINKED ISSUE #2756: 🐛 Typescript: error TS1013 raised with rest parameter in multiline function declaration
### Environment information

```bash
rome 0.6.1-next
tsc 8.3.0
```
```


### What happened?

### Summary
rome can add an invalid training comma, raising `error TS1013: A rest parameter or binding pattern may not have a trailing comma.`

### Reproduction
Given:
```typescript
// sample.ts
export class Task {
    args: any[];

    constructor(
        public script: string,
        public duration: number,
        public threadCount: number,
        ...args: any[]
    ) {
        this.args = args;
    }
}

```
`rome format --write .` will generate invalid syntax by placing a commas after `...args: any[]`

### Notes
1. The arguments in the function must be long enough to be split into multiple lines
2. The last argument must be a `rest` parameter using the `...argName` notation

### Expected result

A rest parameter in a multiline function parameter declaration will **not** have a trailing comma added.

### Code of Conduct

ematipico:
## Summary

This PR fixes #2756

It checks if the last argument is not a rest operator. And if so, it doesn't add the trailing separator.

FYI, I copied the exact same check we have in `parameters_list.rs`

## Test Plan

Added two new tests, one for TS and one for JS

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/call_arguments.rs
+++ b/crates/rome_js_formatter/src/js/expressions/call_arguments.rs
@@ -106,7 +106,6 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
                 // which means that if one is `false`, then the other is `true`.
                 // This means that in this branch we format the case where `should_group_first_argument`,
                 // in the else branch we format the case where `should_group_last_argument` is `true`.
-
                 write!(f, [l_leading_trivia, l_paren, l_trailing_trivia,])?;
                 if should_group_first_argument {
                     // special formatting of the first element
@@ -132,7 +131,6 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
                         }))
                         .finish()?;
                 }
-
                 write!(f, [r_leading_trivia, r_paren, r_trailing_trivia])
             });
 
--- a/crates/rome_js_formatter/src/js/lists/constructor_parameter_list.rs
+++ b/crates/rome_js_formatter/src/js/lists/constructor_parameter_list.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use rome_js_syntax::{JsConstructorParameterList, JsSyntaxKind};
+use rome_js_syntax::{JsAnyConstructorParameter, JsConstructorParameterList, JsSyntaxKind};
 
 #[derive(Debug, Clone, Default)]
 pub struct FormatJsConstructorParameterList;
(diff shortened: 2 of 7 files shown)


IWANABETHATGUY:
!bench_formatter

crates/rome_js_formatter/src/separated.rs
```diff
@@ -30,19 +30,22 @@ where
...
             if self.is_last {
                 match self.options.trailing_separator {
                     TrailingSeparator::Allowed => {
-                        // Use format_replaced instead of wrapping the result of format_token
-                        // in order to remove only the token itself when the group doesn't break
-                        // but still print its associated trivia unconditionally
-                        format_only_if_breaks(separator, &separator.format())
-                            .with_group_id(self.options.group_id)
-                            .fmt(f)?;
+                        if !node_is_rest_operator {
```
MichaReiser, line 41: You need to call `format_removed` if the separator is present OR return an `Err` (because it is invalid syntax)

crates/rome_js_formatter/src/separated.rs
```diff
@@ -30,19 +30,22 @@ where
         } else {
             group_elements(&node.format()).fmt(f)?;
         }
+        let node_is_rest_operator = matches!(node.syntax().kind(), JsSyntaxKind::JS_REST_PARAMETER);
```
MichaReiser, line 33: I like that this approach solves this genericly but the downside is that this is now preventing us from moving the formatting logic to `rome_formatter` (all that is needed to move it today is to add a `CommentStyle` parameter and change `JsLanguage` to `L`. 

It would be nice if we can parametrise this over the `options` instead.

crates/rome_js_formatter/src/separated.rs
```diff
@@ -61,14 +64,18 @@ where
...
+                        write!(
+                            f,
+                            [if_group_breaks(&format_inserted(self.separator))
+                                .with_group_id(self.options.group_id)]
+                        )?;
+                    }
                 }
                 TrailingSeparator::Mandatory => {
-                    format_inserted(self.separator).fmt(f)?;
+                    if !node_is_rest_operator {
```
MichaReiser, line 76: This seems a bit conflicting. My understanding of mandatory is that it is a syntax error if it is missing, thus, omitting it here would lead to a syntax error. Should the call site instead use `Allowed`? 

If not, then we should also change the logic in the branch where the token is present.

crates/rome_js_formatter/src/separated.rs
ematipico: Or maybe move the option in the callsite... probably it makes more sense

PR #2763, opened 2022-06-22
feat(rome_js_analyze): implement control flow analysis

leops:
## Summary

This PR adds a new `noDeadCode` rule to the analyzer (might be renamed to `noUnusedCode` or `noInaccessibleCode`) that emits "unnecessary code" warning diagnostics for code that is statically determined to be unreachable

It does so by running a simple reachability analysis over the control flow graph of each function in the program, a new data structure defined in `rome_control_flow` and generated in `rome_js_analyzer` as part of the analysis traversal of the syntax tree. 

This control flow graph is an auxiliary data structure to the syntax tree representing the execution order of statements and expressions in the function as a graph of "basic blocks" comprised of a flattened list of instruction. There are currently 3 kind of instructions: `Statement` denotes the execution of a certain syntax node that may have some side effects but does not explicitly impact the control flow, `Jump` signals an edge from the current block to another block in the function (the edge may be conditional in which case it will generally be associated with an expression), and finally `Return` represents an exit point for the function (these includes return statements but also throw statements).

As an example, considering the following function:

```js
function exampleLoop(count) {
    for(let i = 0; i < count; i++) {
        if(count - i === 2) {
            continue;
        }

        if(i % 3 === 0) {
            callExpression(i + 1);
        } else {
            callExpression(i);
        }
    }

    callExpression();
}
```

The following control flow graph would be generated:

```mermaid
flowchart TB
    block_0["Statement(JS_VARIABLE_DECLARATION)<br/>Jump { block: 1 }"]
    block_1["Jump { condition: JS_BINARY_EXPRESSION, block: 4 }<br/>Jump { block: 3 }"]
    block_2["Statement(JS_POST_UPDATE_EXPRESSION)<br/>Jump { block: 1 }"]
    block_3["Statement(JS_EXPRESSION_STATEMENT)"]
    block_4["Jump { condition: JS_BINARY_EXPRESSION, block: 5 }<br/>Jump { block: 6 }"]
    block_5["Jump { block: 2 }<br/>Jump { block: 6 }"]
    block_6["Jump { condition: JS_BINARY_EXPRESSION, block: 7 }<br/>Jump { block: 8 }"]
    block_7["Statement(JS_EXPRESSION_STATEMENT)<br/>Jump { block: 9 }"]
    block_8["Statement(JS_EXPRESSION_STATEMENT)<br/>Jump { block: 9 }"]
    block_9["Jump { block: 2 }"]

    block_7 --> block_9
    block_5 --> block_6
    block_4 -- JS_BINARY_EXPRESSION --> block_5
    block_4 --> block_6
    block_0 --> block_1
    block_2 --> block_1
    block_6 -- JS_BINARY_EXPRESSION --> block_7
    block_1 --> block_3
    block_6 --> block_8
    block_1 -- JS_BINARY_EXPRESSION --> block_4
    block_8 --> block_9
    block_5 --> block_2
    block_9 --> block_2
```

(Note that the generated graph contains extraneous jumps and edges, these could be avoided in the generation pass but ultimately it does not influence the end result as everything following an unconditional jump is treated as unreachable)

In order to generate this data, the infrastructure of the analyzer has been refactored to allow for any number of `Visitor` to be registered to analyzer. These visitors receive language-specific `WalkEvent` during the syntax traversal pass and may use the opportunity to mutate their own state, as well as mutably access the shared `VisitorContext` object. This context holds a reference to the `RuleRegistry` through which the visitors may dispatch `QueryMatch` objects to trigger the rules associated with a certain query: for instance `rome_analyze` defines a `SyntaxVisitor` that dispatched an `Ast` query match for each node it enters, which is the query type used by all currently existing lint rules as they all currently work on AST nodes. This PR introduces a new `ControlFlowGraph` queryable type that's emitted by the `ControlFlowVisitor` when the traversal exits a function (and it did not contain syntax errors that caused the control flow graph generation to fail).

The `ControlFlowVisitor` is implemented using a new `merge_node_visitors` macro that creates a single struct implementing `Visitor` over a collection of type implementing the `NodeVisitor` helper trait. Unlike the global `Visitor`, node visitors are transient: they get instantiated when the traversal enters the corresponding node and destroyed when the node is exited. They are intended as a building blocks for creating and managing the state of complex visitors by allowing the implementation to be split over multiple smaller components.

TODO:
- Add documentation comments to the code
- Should the refactor of the analyzer infrastructure be split off into its own PR ?

## Test Plan

Currently all these changes are only being tested through the integration tests of the analyzer as a whole, but it could be useful for the control flow generation code and general analyzer infrastructure to have their own unit test to check for issues and detect regressions at a finer level


crates/rome_js_analyze/src/analyzers/no_dead_code.rs
```diff
@@ -0,0 +1,123 @@
...
+    type Query = ControlFlowGraph;
+    type State = TextRange;
+    type Signals = Vec<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let mut signals = Vec::new();
+
+        let cfg = ctx.query();
+
+        let mut reachable_blocks = BTreeSet::new();
```
IWANABETHATGUY, line 55: Why do we use a `BTreeSet` instead of a `HashSet` here? I don't see any in order iteration operation, maybe use `HashSet` would faster ?

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -15,16 +15,69 @@ use text_size::TextRange;
 mod mutation;
 
 use crate::syntax::{SyntaxSlot, SyntaxSlots};
-use crate::{Language, SyntaxList, SyntaxNode, SyntaxToken};
+use crate::{Language, RawSyntaxKind, SyntaxKind, SyntaxList, SyntaxNode, SyntaxToken};
 pub use mutation::{AstNodeExt, AstNodeListExt, AstSeparatedListExt};
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct SyntaxKindSet<L: ?Sized + Language>([u128; 4], PhantomData<L>);
```
xunilrj, line 22: Have you checked something like this? https://github.com/RoaringBitmap/roaring-rs

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
leops: `BTreeSet` is generally faster than `HashSet` for numerical keys, at least with the default `RandomState` hasher (`FxHashSet` however should be faster than `BTreeSet`)

xunilrj: Given that we are just using  "insert" and "contains", can we use https://github.com/RoaringBitmap/roaring-rs?

crates/rome_analyze/src/visitor.rs
```diff
@@ -0,0 +1,151 @@
...
+
+/// Visitors are the main building blocks of the analyzer: they receive syntax
+/// [WalkEvent]s, process these events to build secondary data structures from
+/// the syntax tree, and emit rule query matches through the [RuleRegistry]
+pub trait Visitor<B> {
+    type Language: Language;
+
+    fn visit(
+        &mut self,
+        event: WalkEvent<SyntaxNode<Self::Language>>,
```
MichaReiser, line 24: Nit: Can visitors work on references to the walk events?

crates/rome_control_flow/src/builder.rs
```diff
@@ -0,0 +1,59 @@
...
+        self.block_cursor
+    }
+
+    pub fn set_cursor(&mut self, block: BlockId) {
+        debug_assert!(block.index < self.result.blocks.len());
+        self.block_cursor = block;
+    }
+
+    pub fn append_instruction(&mut self, inst: Instruction<L>) {
+        self.result.blocks[self.block_cursor.index]
```
MichaReiser, line 51: Nit: Would it be possible to have one single instruction vector over all blocks and only store the start/end indices per block or is this not possible because the order in which instructions are appended isn't known when building the blocks?

crates/rome_analyze/src/lib.rs
```diff
@@ -1,16 +1,57 @@
...
+
+    pub fn add_visitor<V>(&mut self, visitor: V)
+    where
+        V: Visitor<B, Language = L> + 'static,
+    {
+        self.visitors.push(Box::new(visitor));
+    }
+
+    pub fn run(mut self, ctx: &mut VisitorContext<L, B>) -> Option<B> {
+        for event in ctx.root.syntax().preorder() {
```
xunilrj, line 44: Not sure here if we need 

```rust
for event in ctx.root.syntax().preorder() {
    for visitor in &mut self.visitors {
        ...
    }
}
```

or 


```rust
for visitor in &mut self.visitors {
    for event in ctx.root.syntax().preorder() {
        ...
    }
}
```

The second would support the semantic visitor.

Visitor 1 - Syntax <- Run rules depending on the AST
Visitor 2 - Semantic <- Generate Semantic Model
Visitor 3 - Semantic + CFG <- Run Rules depending on CFG and/or Semantic Model

MichaReiser: The downside of running `preorder` for every visitor is that we traverse the whole tree multiple times. 

Would it be possible to change the semantic model so that it supports a `visit` API vs doing the traversal itself?

crates/rome_analyze/src/registry.rs
```diff
@@ -1,119 +1,171 @@
...
-        R::Query: AstNode<Language = L>,
+        R::Query: Queryable<Language = L> + Clone,
     {
-        self.rules.push(RegistryRule::of::<R>());
+        match <R::Query as Queryable>::KEY {
+            QueryKey::Syntax(key) => {
+                for kind in key.iter() {
+                    let index = usize::from(kind.to_raw().0);
+
+                    if self.syntax.len() <= index {
```
xunilrj, line 39: We can pre-allocate here using: ```JsSyntaxKind::LAST``` or even just using the limit of JsSyntaxKindSet.

crates/rome_control_flow/src/builder.rs
leops: Blocks are often created and filled out of order (for instance the block following a loop needs to be created before generating the code for the body of said loop in order to provide a jump target for break statements) so allocating the instructions in a single arena vector might require `insert` calls and range fixups (whether that would improve or regress performance compared to having to drop nested vectors is a little uncertain to me though, I'd need to test it out)

crates/rome_js_syntax/src/generated/nodes.rs
```diff
@@ -30424,6 +31281,40 @@ impl From<TsVoidType> for TsType {
 }
 impl AstNode for TsType {
     type Language = Language;
+    const KIND_SET: SyntaxKindSet<Language> = TsAnyType::KIND_SET
+        .with_union(TsArrayType::KIND_SET)
```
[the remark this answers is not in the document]
leops: The name "union" is referring to the concept in set theory, not to union nodes in the grammar: the `with_union` method returns the union of the two sets `TsAnyType::KIND_SET` and `TsArrayType::KIND_SET`

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -15,16 +15,69 @@ use text_size::TextRange;
...
+{
+    pub fn of(kind: L::Kind) -> Self {
+        Self::from_raw(kind.to_raw())
+    }
+
+    pub const fn from_raw(kind: RawSyntaxKind) -> Self {
+        Self(kind.as_bits(), PhantomData)
+    }
+
+    pub const fn with_union(self, other: Self) -> Self {
```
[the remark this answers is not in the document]
leops: Yes that's what it's based on, with some small modifications to make the methods `const`

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
```diff
@@ -0,0 +1,123 @@
...
+        let mut queue = VecDeque::new();
```
IWANABETHATGUY, line 56: According code below, the `length` of `queue` should at least larger than `length` of `cfg.entry_blocks`, So use `VecDeque::with_capicity( cfg.entry_blocks.len()` should be faster.

crates/rome_analyze/src/lib.rs
xunilrj: > Would it be possible to change the semantic model so that it supports a visit API vs doing the traversal itself?

We already have that.

The problem with the semantic model is that its rules can only run after the whole tree is iterated. For example, you want to ask for all bindings of a block inside the block and want to see "future bindings".

I can try to alleviate this by calling the semantic rules on the "exit" of the visitor, but this would still not solve for hoisting and other more complex cases.

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
xunilrj: Or at least a fraction of it. Like 10% of ```cfg.entry_blocks.len()```.

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
IWANABETHATGUY: How much `cfg.entry_blocks` would we have in realworld code case ?

crates/rome_js_syntax/src/generated/nodes.rs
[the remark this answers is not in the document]
MichaReiser: I see. I would, in that case, rename the method to `union` as it performs a "union" operation (same as `.add`) and doesn't change the field `union` to have the given value

crates/rome_analyze/src/lib.rs
MichaReiser: > We already have that.

Great! 

> The problem with the semantic model is that its rules can only run after the whole tree is iterated. For example, you want to ask for all bindings of a block inside the block and want to see "future bindings".

My understanding is that you can run any additional logic at the end of the traversal when retrieving the result from the visitor. 

Rules that need access to the semantic model need to wait for the semantic model to be ready before they call into `analyze` again. But we can, for the beginning, schedule all lint rules in a second pass.

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
leops: Running the analysis on the libraries we use for benchmarking, the maximum number of `entry_blocks` seems to be 16 (it's very close to 1 on average as most functions only have one entry point, the only nodes that create additional entry paths into a function are `catch` or `finally` clauses on `try` statements)
For the number of actual `blocks` in a function, it's approximately 3.8 on average with a maximum of 1884 blocks in a single function
Finally, the maximum length attained by the blocks queue is 168, for a function that has 171 blocks (and a single entry block)

crates/rome_analyze/src/registry.rs
leops: Interestingly I tried running a microbenchmark on this, and preallocating the maximum capacity seems to be consistently slightly slower than resizing the vector on demand. Of course this is only about 3 or 4 µs so that's largely within the noise margin, and is is probably largely dependent on the underlying allocator (and how it can reuse the same memory block between iterations of the benchmark loop). Overall on Windows at least is doesn't seem to make any significant difference

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
IWANABETHATGUY: Here is my benchmark result on `FxhashSet`, `std::HashSet` and `BTreeSet`, 
```bash
Gnuplot not found, using plotters backend
BTreeSet insert 10      time:   [204.89 ns 206.40 ns 208.10 ns]                               
                        change: [-1.9987% -0.8072% +0.3246%] (p = 0.18 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe

HashSet insert 10       time:   [684.25 ns 689.45 ns 694.37 ns]                               
                        change: [+1.5628% +2.7069% +3.7221%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

FxHashSet insert 10     time:   [507.02 ns 512.64 ns 518.88 ns]                                 
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe

BTreeSet contains 10    time:   [20.194 ns 20.396 ns 20.668 ns]                                  
                        change: [-18.211% -17.056% -15.953%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe

HashSet contains 10     time:   [58.242 ns 59.526 ns 60.982 ns]                                
                        change: [-1.2754% +0.4989% +2.2110%] (p = 0.58 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe

FxHashSet contains 10   time:   [13.146 ns 13.208 ns 13.273 ns]                                   
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  6 (6.00%) high severe

```

IWANABETHATGUY: I have to say you are right, `BTreeSet` is a better choice when dataset is around `10 - 20`

crates/rome_analyze/src/registry.rs
xunilrj: Probably because of the cost of initializing the "empty vector". If we could ```memset(0)``` would probably be faster. But we can stash this for later then.

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+#[derive(Debug, Clone)]
+pub struct ControlFlowGraph<L: Language> {
+    /// List of blocks that make up this function
+    pub blocks: Vec<BasicBlock<L>>,
+    /// Indices of the block that act as "entry point" for the function,
+    /// defined as points that the control flow may arbitrarily jump into
+    ///
+    /// Currently this is used for the first block in the function, as well as
+    /// `catch` and `finally` clauses to model the exception control flow
+    /// possibly diverging into these
```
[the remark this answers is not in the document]
ematipico: "to model the exception control flow possibly diverging into these" I find this phrase hard to read... I can't understand it

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+}
+
+/// A basic block represents an atomic unit of control flow, a flat list of
+/// instruction that will be executed linearly when the function is run
+///
+/// Note however that while the instructions that comprise a basic block are
+/// guaranteed to be executed in order from the start towards the end, the
+/// block may not be executed entirely if a jump or return instruction is
+/// encountered
```
ematipico, line 43:
```suggestion
/// Note, however, that while the instructions that comprise a basic block are
/// guaranteed to be executed in order from the start towards the end, the
/// block may not be executed entirely if a jump or return instruction is
/// encountered.
```

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+impl<L: Language> BasicBlock<L> {
+    fn new() -> Self {
+        Self {
+            instructions: Vec::new(),
+        }
+    }
+}
+
+/// Instructions are used to represent statements or expression being executed
+/// for side effects, as well as potential divergence points for the control flow
```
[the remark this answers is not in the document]
ematipico:
```suggestion
/// Instructions are used to represent statements or expressions being executed
/// for side effects, as well for potential divergence points for the control flow.
```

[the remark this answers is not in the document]
ematipico: We have a `as` and a `for`. We should align the prepositions

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+///
+/// Each node has an associated kind, as well as an optional syntax node: the
+/// node is useful to emit diagnostics but does not have a semantic value, and
+/// is optional as the code generating the control flow graph may emit
+/// instructions that do not correspond to any node in the syntax tree to model
+/// the control flow of the program accurately
```
ematipico, line 64:
```suggestion
/// node is useful to emit diagnostics but does not have a semantic value, and
/// is optional. The code generating the control flow graph may emit
/// instructions that do not correspond to any node in the syntax tree, to model
/// the control flow of the program accurately.
```

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/// The different types of supported [Instruction]
+#[derive(Copy, Clone, Debug)]
+pub enum InstructionKind {
+    /// Indicates the [SyntaxNode] associated with this instruction is to be
+    /// evaluated at this point in the program
+    Statement,
+    /// This instruction may cause the control flow to diverge towards `block`,
+    /// either unconditionally if `conditional` is set to false, or after
```
ematipico, line 78:
```suggestion
    /// either unconditionally if `conditional` is set to `false`, or after
```

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// evaluating the associated syntax node otherwise
```
ematipico, line 79: I believe "syntax node" is kinda misleading. Should we say "associated block" instead? We don't know if a block can have an associated node, as they can be optional

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    Jump { conditional: bool, block: BlockId },
+    /// This instruction causes the control flow to unconditionally abort the
+    /// execution of the function, either due to a `return` or `throw` statement
```
ematipico, line 82:
```suggestion
    /// execution of the function, for example is JavaScript this can be triggered by a `return` or `throw` statement
```

crates/rome_control_flow/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    Return,
+}
+
+/// The Display implementation for ControlFlowGraph prints a flowchart in
```
ematipico, line 86:
```suggestion
/// The Display implementation for [ControlFlowGraph] prints a flowchart in
```

crates/rome_control_flow/src/builder.rs
```diff
@@ -0,0 +1,75 @@
...
+        self.result.blocks.push(BasicBlock::new());
+        BlockId { index }
+    }
+
+    /// Get the [BlockId] at the current position of the cursor
+    pub fn cursor(&self) -> BlockId {
+        self.block_cursor
+    }
+
+    /// Move the cursor to the end of `block`
```
ematipico, line 59: Why can't we assign a new block that is lower than the current one? I believe we should document it or add an explanation in the assertion.

crates/rome_control_flow/src/builder.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn set_cursor(&mut self, block: BlockId) {
+        debug_assert!(block.index < self.result.blocks.len() as u32);
+        self.block_cursor = block;
+    }
+
+    /// Insert an instruction at the current position of the cursor
+    pub fn append_instruction(&mut self, inst: Instruction<L>) {
+        let index = self.block_cursor.index as usize;
+        self.result.blocks[index].instructions.push(inst);
```
ematipico, line 68: How is this safe? Can we add a `// SAFETY` comment?

ematipico (commented):
First batch of comments around `rome_control_flow`

crates/rome_control_flow/src/builder.rs
ematipico: My question is more about index overflow. What if `index` is bigger than `blocks.len()`? That is covered inside a `debug_assert` inside `append_instruction`, which goes away on a release. That's why I was wondering why it's safe to do so.

ematipico:
It would be great if we can split this PR in multiple and smaller ones, mostly because there are different things going on here, and it won't be easy to review all of them:

- one with only `rome_control_flow`
- one with codegen (it makes lot of noise) 
- one the implementation of the control flow (if the codegen and control flow are tight together, then one with both)
- one with the new lint rules 

This PR is too big, I think

crates/rome_control_flow/src/lib.rs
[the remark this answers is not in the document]
leops: I also had a hard time coming up with it, the idea I'm trying to explain here is that since throwing an exception is represented as a `Return` instruction that aborts the execution of the function, catch and finally blocks are represented as alternate entry points the control flow can conceptually resume from in case an exception is thrown. This isn't entirely accurate as it means `catch` and `finally` clauses are never marked as unused event if the `try` statement as a whole is unreachable but it works well enough as a very simplified representation of exception control flow

crates/rome_control_flow/src/lib.rs
leops: I meant "syntax node" as in the `node` field of the `Instruction` that has its kind set to `Jump`. It isn't a hard requirement that a jump instruction with `conditional: true` has an associated node, and the analysis will still work correctly if it doesn't have once, but since generally evaluating said condition will require running a certain part of the code having it associated with the jump instruction allows it to be marked as unnecessary in case the instruction is found to be unreachable

crates/rome_control_flow/src/builder.rs
leops: The assert is checking that the index of the provided `BlockId` is lower than the number of blocks in the function. Since the only way to obtain a `BlockId` for a user of the `FunctionBuilder` is to call `append_block` (or `cursor`) this assert should always pass (the builder does not emit IDs for blocks that do not exist), except if a `BlockId` generated by one instance of the function builder is used to write into a different function: this is what this assert is partially trying to catch, but since writing a block to the wrong builder is a programming error (it's a bug in the code generating the control flow graph) we don't really need to run that check in release mode

crates/rome_control_flow/src/builder.rs
leops: As I explained above, since the `BlockId` type is opaque for users of the `FunctionBuilder` it's not possible to manually construct an invalid block ID that moves the cursor to an out-of-bounds index, the only way of doing that is by generating block IDs with one builder and using them to write another one but that's clearly a bug

leops:
> It would be great if we can split this PR in multiple and smaller ones

I've rebased the branch and split it into multiple commits, so I could easily submit those as separate PRs:

- expose the kind of AstNode as the `KIND_SET` constant
- refactor `rome_analyze` to introduce visitors and queryables
- add the `rome_control_flow` crate
- implement the control flow analysis for JS

Does that make sense ?

xunilrj:
> Does that make sense ?

I am interested in number 2. Maybe the first two can be created in parallel.

ematipico:
> > It would be great if we can split this PR in multiple and smaller ones
> 
> I've rebased the branch and split it into multiple commits, so I could easily submit those as separate PRs:
> 
>     * expose the kind of AstNode as the `KIND_SET` constant
> 
>     * refactor `rome_analyze` to introduce visitors and queryables
> 
>     * add the `rome_control_flow` crate
> 
>     * implement the control flow analysis for JS
> 
> 
> Does that make sense ?

I am not a big fun of big PRs in general, even though we have good commits, new commits will add up because of PR reviews and how would that work? It's really hard to maintain and for us reviewers to keep track of the changes

crates/rome_js_analyze/src/control_flow/nodes.rs
```diff
@@ -0,0 +1,1143 @@
+use std::any::TypeId;
```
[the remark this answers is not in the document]
leops: Agreed, in fact this was pretty much the entire point of introducing `NodeVisitor` and `merge_node_visitors`

ematipico (commented):
Wow this is really neat!! 😍

crates/rome_js_analyze/src/control_flow/visitor.rs
```diff
@@ -0,0 +1,239 @@
...
+        ctx: &mut VisitorContext<JsLanguage, B>,
+        _: &mut ControlFlowVisitor<B>,
+    ) -> ControlFlow<B> {
+        if let Some(builder) = self.builder {
+            return ctx.match_query(&QueryMatch::ControlFlowGraph(builder.finish()));
+        }
+
+        ControlFlow::Continue(())
+    }
+}
```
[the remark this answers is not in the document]
leops: This visitor is different from the other as it implements the `NodeVisitor` trait from `rome_analyze` instead of the one declared just above in `rome_js_analyze`. The role of this visitor is to provide the `FunctionBuilder` used by all the other visitor implementations, but it doesn't really contribute to building the control flow graph itself

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
```diff
@@ -0,0 +1,315 @@
...
+        // number and position of secondary labels
+        match state.terminators.as_slice() {
+            // The CFG didn't contain enough informations to determine a cause
+            // for this range being unreachable
+            [] => {}
+            // A single node is responsible for this range being unreachable
+            [node] => {
+                if node.start() < state.text_trimmed_range.start() {
+                    diagnostic = diagnostic
+                        .secondary(*node, "This statement will abort control flow ...")
```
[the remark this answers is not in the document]
leops: I tried to push this even further by providing specialized explanations messages depending on the type of the underlying node (eg. "return from the function" for `JsReturnStatement` or "continue the loop" for `JsContinueStatement`), using your suggestion as the default / fallback reason

crates/rome_js_analyze/src/control_flow/nodes/if_stmt.rs
```diff
@@ -0,0 +1,116 @@
...
+            self.consequent_end.unwrap_or_else(|| builder.cursor()),
+        );
+
+        let alt_block = self.alt_block;
+
+        let next_block = builder.append_block();
+
+        builder.set_cursor(self.entry_block);
+
+        if let Some((alt_start, alt_end)) = alt_block {
```
ematipico, line 55: It took me a bit to understand that `alt_block` is set by the `ElseVisitor`. Would it be possible to add a comment here?

crates/rome_js_analyze/src/control_flow/nodes/switch_stmt.rs
```diff
@@ -0,0 +1,102 @@
...
+            break_block,
+            ..
+        } = self;
+
+        builder.append_jump(false, break_block);
+
+        builder.set_cursor(entry_block);
+        builder.append_jump(false, break_block);
+
+        builder.set_cursor(break_block);
```
ematipico, line 57: The `break_block` is added twice, is it correct?

crates/rome_js_analyze/src/control_flow/nodes/switch_stmt.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::control_flow::{
+    visitor::{NodeVisitor, StatementStack},
+    FunctionBuilder,
+};
+
+pub(in crate::control_flow) struct SwitchVisitor {
+    entry_block: BlockId,
+    pub(super) label: Option<JsSyntaxToken>,
+    pub(super) break_block: BlockId,
+    is_first: bool,
```
ematipico, line 14: What's this used for? I find it difficult to understand and the usage doesn't help much. As I read it, it's telling me if a `switch` statement is "first". First in what?

crates/rome_js_analyze/tests/specs/noDeadCode.js
```diff
@@ -0,0 +1,234 @@
+// JsLabeledStatement
```
ematipico, line 1: nit: not sure if it's already possible, but I think it would be more helpful to have a directory called `noDeadCode` and have each scenario inside a separate file. This would help to see the diagnostic that belongs to the case we are testing.

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+
+            let catch_block = builder.append_block();
+            builder.add_entry_block(catch_block);
+            builder.set_cursor(catch_block);
+
+            Ok(Self { next_block: None })
+        } else {
+            // Cursor is at the end of the try block, jump to the next block
+            let next_block = builder.append_block();
+            builder.append_jump(false, next_block);
```
ematipico, line 53: What's a "next block" in this case? Don't we have `try` and ` catch` block in this case? Should `next_block` be `try_block`?

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+
+            builder.append_jump(false, *finally_block);
+
+            let catch_block = builder.append_block();
+            builder.add_entry_block(catch_block);
+            builder.set_cursor(catch_block);
+
+            Ok(Self { next_block: None })
+        } else {
+            // Cursor is at the end of the try block, jump to the next block
```
ematipico, line 51: Can we add a comment say that we are covering the `JsTryStatement` node here? Asking because we don't it at all among the visitors.

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+            let next_block = builder.append_block();
+            builder.append_jump(false, next_block);
+
+            // Create the catch block, mark is as entry point and start writing
```
ematipico, line 55:
```suggestion
            // Create the catch block, mark it as entry point and start writing
```

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+        _: Self::Node,
+        builder: &mut FunctionBuilder,
+        stack: StatementStack,
+    ) -> SyntaxResult<()> {
+        let Self { next_block } = self;
+
+        let try_finally_stmt = stack.read_top::<TryFinallyVisitor>();
+
+        if let Ok(try_finally_stmt) = try_finally_stmt {
+            builder.append_jump(false, try_finally_stmt.finally_block.unwrap());
```
ematipico, line 77: Let's add a `// SAFETY` comment

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+
+            Ok(())
+        } else {
+            let next_block = next_block.unwrap();
```
ematipico, line 81: `// SAFETY` comment

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    fn enter(_: Self::Node, _: &mut FunctionBuilder, _: StatementStack) -> SyntaxResult<Self> {
+        Ok(Self {
+            finally_block: None,
+        })
+    }
+}
+
+pub(in crate::control_flow) struct CatchVisitor {
+    next_block: Option<BlockId>,
```
ematipico, line 25: `next_block` seems to be not descriptive enough. Considering we are inside a `try/catch/finally` scenario, maybe we can give it a name that suite more the situation

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
```diff
@@ -0,0 +1,134 @@
...
+    type Node = JsFinallyClause;
+
+    fn enter(
+        _: Self::Node,
+        builder: &mut FunctionBuilder,
+        stack: StatementStack,
+    ) -> SyntaxResult<Self> {
+        let try_finally_stmt = stack.read_top::<TryFinallyVisitor>()?;
+
+        let has_catch = try_finally_stmt.finally_block.is_some();
```
ematipico, line 106: Why having `finally_block` as `Some` determines the presence of the catch (`has_catch`)? 

```js
try {} catch (e) {} finally {};
try {} finally {};
```

These are both parsed to `JsTryFinallyStatement` and the `catch` is present in the first statement.

ematipico (commented):
I went through the majority of the visitors, and I think we should document some of these structs, in particular some properties.

For example:
- cases where we have tuples `Option<()>`. We should document each item;
- labels, they are not actively used inside the visitor where it's encountered, we should document **where** they are used;

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
leops: As a convention in all statement visitors `next_block` refers to the block that contains whatever comes after this statement in the control flow, it could be any other statement or nothing at all so it can't really refer to something specific

crates/rome_js_analyze/src/control_flow/nodes/try_catch.rs
leops: If the try-finally statement has a catch clause then `CatchVisitor::enter` will have been called before going into `FinallyVisitor::enter`: since the `finally_block` is initialized to `None` but is allocated by the catch visitor if it's part of a try-finally statement, being set to `Some` when the finally clause is reached means a catch clause has been encountered

crates/rome_js_analyze/tests/specs/noDeadCode/JsLabeledStatement.js.snap
```diff
@@ -0,0 +1,23 @@
...
+    label: while (true) {
+        while (true) {
+            if (true) {
+                continue label;
+            } else {
+                break label;
+            }
+        }
+
+        afterLabelJump();
```
ematipico, line 17: Here we should have two diagnostics: one at `break label` and the second at `afterLabelJump`

crates/rome_js_analyze/tests/specs/noDeadCode/JsSwitchStatement.js.snap
```diff
@@ -0,0 +1,74 @@
...
+        case 0:
+            case0();
+            break;
+        default:
+            caseDefault();
+            break;
+        case 1: // afterDefault
+            afterDefault();
+    }
+}
```
ematipico, line 18: I don't think this statement should have any diagnostics. If `value` is `1`, then the `case 1` is actually executed. The only way to statically know if `case 1` won't be executed is by knowing if `value` won't ever be `1`

ematipico (changes requested):

crates/rome_js_analyze/tests/specs/noDeadCode/JsLabeledStatement.js.snap
leops: With this initial implementation of control flow analysis there is no constant folding or any kind or expression evaluation, the CFG encodes the two branches of the if-statement but has no way of inspecting the condition expression and thus can't determine the condition is always true and the else branch will never be taken

crates/rome_js_analyze/tests/specs/noDeadCode/JsSwitchStatement.js.snap
leops: Good catch, that's I regression I probably introduced when I factored out the control flow generation for switch statements into multiple node visitors

crates/rome_js_analyze/tests/specs/noDeadCode/JsLabeledStatement.js.snap
ematipico: Either way, if **not having** diagnostics is a known limitation, then we should not have the snippet. Instead, if we should have on diagnostic, we should fix it.

leops: To be clear that was only about `break label`, this snippet is verifying the reachability analysis can correctly reason about labeled break or continue statement, even when they're nested inside multiple other nodes like loops or conditions. There should be a diagnostic emitted for this snapshot, but only for the `afterLabelJump();` expression statement, and it being missing is a bug

ematipico (approved):

PR #2769, opened 2022-06-23
fix(rome_js_formatter): custom formatting for test calls

ematipico:
## Summary

This PR implements gets us closer to how call arguments are formatted.

Prettier has a particular formatting just for call expressions that are used by testing frameworks, such as

```js
describe("Component X", () => {});
it.only("it does want I want", t => {});
```

This PR brings their algorithm into our code base and it uses it to do an early formatting if the criteria are met. The function explains what the criteria are.

While doing so, I added an option to `JsCallArguments` to accept an option called `callee`, which is essentially the `callee` that belong to its parent. The reason why I added this option is because this check should not be used for call expressions of dynamic import. So instead of calling the parent, checking its type and then extracting the`callee`, I decided to pass a reference of the `callee` from the parent.

It addresses a comment from closed PR, by limiting the usage of a custom buffer to a single call, instead of dragging the whole buffer across functions;

## Test Plan

`cargo test`  and made sure that the prettier testing suite format as prettier's

PR 

**File Based Average Prettier Similarity**: 76.09%  
**Line Based Average Prettier Similarity**: 71.53% 

`main`
**File Based Average Prettier Similarity**: 76.00%  
**Line Based Average Prettier Similarity**: 70.84%


crates/rome_formatter/src/builders.rs
```diff
@@ -1892,3 +1892,110 @@ impl<Context> Format<Context> for BestFitting<'_, Context> {
...
+///     
+///     let element_will_break = format_with(|f| {
+///         write!(f, [
+///             token("hello"),
+///             hard_line_break(),
+///             token("world!")
+///         ])
+///     });
+///
+///     let will_break = will_break(&element_will_break, f)?;
```
MichaReiser, line 1923: This will already write the content into `f`. So there's no need to print `element_will_break` inside of the `will_break` branch again.

MichaReiser (commented):
I think it would be good to split out the `WillBreakBuffer` into its own PR as the other logic in this PR isn't using it.

crates/rome_formatter/src/builders.rs
```diff
@@ -1892,3 +1892,110 @@ impl<Context> Format<Context> for BestFitting<'_, Context> {
...
+///     }   
+///    
+/// })]).unwrap();
+///
+/// assert_eq!(
+///     "break\nhello\nworld",
+///     formatted.print().as_code()
+/// );
+/// ```
+pub fn will_break<Context>(
```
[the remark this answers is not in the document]
ematipico: This will make it's usage less intuitive, are you sure about it?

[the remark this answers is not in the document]
MichaReiser: Returning a `bool` might not be sufficient in many cases because they may want to defer the formatting of the node. `will_break` as a Buffer will make the API more flexible. We can still support additional helpers on top of that.

ematipico:
> I think it would be good to split out the `WillBreakBuffer` into its own PR as the other logic in this PR isn't using it.

Done, that's been removed

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -642,7 +642,14 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
             // 3. we compute the layout
             // 4. we write the left node inside the main buffer based on the layout
             let mut buffer = VecBuffer::new(f.state_mut());
-            let is_left_short = self.write_left(&mut buffer)?;
+            let mut is_left_short = false;
```
MichaReiser, line 645: Nit: Or you can do

```suggestion
            let mut is_left_short = self.write_left(Formatter::new(buffer);
```

ematipico: Is this document somewhere? Because if not, I'd prefer to not use it because it seems more something that the internals should do, but not the consumers.

crates/rome_js_formatter/src/utils/member_chain/flatten_item.rs
```diff
@@ -155,18 +155,21 @@ impl Format<JsFormatContext> for FlattenItem {
...
                     // Formatted as part of the previous item
-                    callee: _,
+                    callee,
                     optional_chain_token,
                     type_arguments,
                     arguments,
                 } = call_expression.as_fields();
 
+                let arguments = arguments?;
+                let arguments = arguments.format().with_options(Some(callee?));
```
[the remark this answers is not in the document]
ematipico: The main reason was that I needed to check the `kind` of the parent (`JsImportCallExpression` for example is not entitled). But biggy, I can revert the change and use the usually `parent()`

ematipico:
I pushed a small refactor and it uses only references, and uses in a better way the checks on the arguments for the custom formatting (hooks and tests). It should improve slightly the performances.

ematipico:
!bench_formatter

crates/rome_formatter/src/buffer.rs
```diff
@@ -1,9 +1,7 @@
 use super::{write, Arguments, FormatElement};
 use crate::format_element::List;
-use std::any::{Any, TypeId};
-
 use crate::{Format, FormatResult, FormatState};
-
+use std::any::{Any, TypeId};
```
MichaReiser, line 4: Nit: There are no real changes in this file. Can we revert them?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -35,25 +34,68 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+        let third_argument = iter.next();
+
+        if let (Some(first_argument), Some(second_argument)) = (first_argument, second_argument) {
+            let first_argument = first_argument?;
+            let second_argument = second_argument?;
+
+            if let Some(call_expression) = node
+                .syntax()
+                .parent()
+                .as_ref()
```
MichaReiser, line 49: I don't think the `as_ref` will give you much here because `syntax.parent()` always returns an owned `SyntaxNode` and you aren't using the same parent in any of the other calls (keeping hold of the parent can be useful if used in multiple places).
```suggestion
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -35,25 +34,68 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+                    return write!(
+                        f,
+                        [
+                            l_paren_token.format(),
+                            format_with(|f| {
+                                let separated =
+                                    args.format_separated(JsSyntaxKind::COMMA).with_options(
+                                        FormatSeparatedOptions::default()
+                                            .with_trailing_separator(TrailingSeparator::Omit),
+                                    );
```
MichaReiser, line 70: Nit: I simplified this API a bit. You can now do
```suggestion
                                    args.format_separated(JsSyntaxKind::COMMA).with_trailing_separator(TrailingSeparator::Omit);
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -35,25 +34,68 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+                                f.join_with(space_token()).entries(separated).finish()
+                            }),
+                            r_paren_token.format()
+                        ]
+                    );
```
MichaReiser, line 75: Nit: You can avoid the inner closure by splitting the `write` into multiple blocks. Up to you which style you prefer (my preference is to reduce nesting and closures)

```suggestion
					write!(f, [l_paren_token.format()])?;

					let separated = args.format_separated(JsSyntaxKind::COMMA)
						.with_trailing_separator(TrailingSeparator::Omit);
					f.join_with(space_token()).entries(separated).finish()?;
					
                    return write!(f, [r_paren_token.format()]);
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -35,25 +34,68 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+                }
             }
 
-            return write!(f, [r_paren_token.format()]);
+            // particular formatting for hooks
+            if is_react_hook_with_deps_array(&first_argument, &second_argument)? {
```
MichaReiser, line 80: Is this the same formatting as for test frameworks? If so, can we unify the two branches:

```
let is_test_call = ... ;
let is_react_hook_call = is_react_hook_with_deps_array(...)

if is_test_call || is_react_hook_call {
   write!... 
}
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -389,46 +431,279 @@ fn could_group_argument(
...
-            let no_parameters = arrow_function.parameters()?.is_empty();
-            let body = arrow_function.body()?;
-            let is_block = matches!(body, JsAnyFunctionBody::JsFunctionBody(_));
-
-            no_parameters && is_block
-        } else {
-            false
-        };
+fn is_react_hook_with_deps_array(
+    first_argument: &JsAnyCallArgument,
```
MichaReiser, line 440: Nice! No need for unwrap

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -389,46 +431,279 @@ fn could_group_argument(
...
+                }
+                _ => return Ok(false),
+            };
+        } else {
+            return Ok(false);
+        }
+    }
+
+    Ok(false)
+}
```
MichaReiser, line 636: Calling `member_expression.member()` creates a new heap allocated `Rc` every time `matches_test_call` is called. That's why I think it could be more performant to extract the member names once and then try to test against the patterns:

`fn matches_test_call(callee: &JsAnyExpression) -> String` where the returned `String` is the pattern of the call expression, e.g. `describe` or `test.skip`. 

You can then test with `matches_test_call(callee:...).matches(patterns)` 

Another alternative that's probably out of scope is to implement this using a prefix search where `matches_test_call` tries to match all patterns in a single go. But this would require that the patterns are represented in a tree structure 

```
describe -> only
               -> skip
it -> only
    -> skip
test 
	-> only
    -> skip
    -> step
	-> describe	-> only
						-> skip   
````

Each node in the tree represents a "valid" pattern. But this is probably something for later.

File: crates/rome_js_formatter/src/utils/assignment_like.rs (hunk 3 above)
Comment (MichaReiser) [reply crates/rome_js_formatter/src/utils/assignment_like.rs:645]: There's some, not very helpful, documentation on `Formatter::new` but in the end, this is exactly what `write!` is doing under the hood (or `write`). 

I agree that we should prefer `write!` whenever possible but it seems to simplify things here quiet a bit

**Review by MichaReiser — APPROVED** [review]
Looks good. I've one comment around testing the `match_test_call` where I believe it should be possible to improve the performance by avoiding iterating O(patterns) times overall member names for any non-test method.

File: crates/rome_js_formatter/src/utils/object.rs [hunk 12]
```diff
@@ -1,36 +1,35 @@
...
+    f: &mut JsFormatter,
 ) -> FormatResult<usize> {
     match name {
         name @ JsAnyObjectMemberName::JsLiteralMemberName(literal) => {
             let value = literal.value()?;
 
             if value.kind() == JS_STRING_LITERAL {
                 let format = FormatLiteralStringToken::new(&value, StringLiteralParentKind::Member);
-                let cleaned = format.clean_text(buffer.context());
+                let cleaned = format.clean_text(f.context());
```
MichaReiser, line 20: Can we remove the `context` method from `VecBuffer` too. Feels weird that it is the only buffer having it (otherwise move it to `Buffer` with a default implementation but it isn't used.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
ematipico: This has the downside that we don't don't how deep we have to go retrieve the names, e.g. `test.only.something.but.i.like.adding.staff` will be considered fine and will query the trees 8 times deep. 

Also, by looking at the name each time, we are able to bail early.

crates/rome_formatter/src/buffer.rs
ematipico: It's just `rustfmt`

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: We know all patterns so we can hard code a "max-depth" at which the check should terminate.

ematipico: Check the last commit and let me know what you think

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -385,46 +398,286 @@ fn could_group_argument(
...
+///     });
+/// })
+///
+/// test.only("", testSomething);
+/// ```
+///
+/// This function should accept the `callee` of [JsCallExpression] and the
+/// string pattern to test against. For example "test", "test.only"
+fn matches_test_call(callee: &JsAnyExpression) -> SyntaxResult<String> {
+    const MAX_DEPTH: u8 = 4;
```
MichaReiser, line 573: We should go one level deeper than the max level tested for to make sure that the call isn't `test.describe.parallel.only.NOT`

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -385,46 +398,286 @@ fn could_group_argument(
...
+                .as_str(),
+            "test"
+        );
+
+        assert_ne!(
+            matches_test_call(&call_expression.callee().unwrap())
+                .unwrap()
+                .as_str(),
+            "testtt"
+        );
```
MichaReiser, line 645: I believe this is already covered by the test for equality. It's guaranteed that the result of `matches_test_call` doesn't equal `"testtt" if it's content equals "test" as asserted on line 637

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -385,46 +398,286 @@ fn could_group_argument(
...
+/// string pattern to test against. For example "test", "test.only"
+fn matches_test_call(callee: &JsAnyExpression) -> SyntaxResult<String> {
+    const MAX_DEPTH: u8 = 4;
+    let mut test_call = Vec::new();
+    let mut current_node = callee.clone();
+    for _ in 0..MAX_DEPTH {
+        if let JsAnyExpression::JsIdentifierExpression(identifier) = &current_node {
+            let value_token = identifier.name()?.value_token()?;
+            let value = value_token.text_trimmed();
+            test_call.push(value.to_string());
```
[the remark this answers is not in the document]
MichaReiser: I agree that it's harder to read but is probably as fast as it can get. The first one still iterates 23 times over the match where the second one does it once.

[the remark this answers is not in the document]
MichaReiser: You could also link to this chapter to explain how it works: 

https://craftinginterpreters.com/scanning-on-demand.html#tries-and-state-machines

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -385,46 +398,296 @@ fn could_group_argument(
...
+                    current_node = member_expression.object()?;
+                }
+                _ => break,
+            };
+        } else {
+            break;
+        }
+    }
+
+    Ok(test_call.into_iter().rev().collect())
```
MichaReiser, line 625: You can do `test_call.reverse()` instead. Doing `into_iter()` + `collect()` creates a new vector

PR #2770, opened 2022-06-23
refactor(rome_js_formatter): Remove unnecessary groups from `format_separated`

MichaReiser:
`format_separated` used to group all elements by default. This is often undesired and also comes with a performance cost (printer needs to test if the elements fit).

This PR changes the default of `format_separated` so that it doesn't group the elements except if enabled explicitly using `nodes_grouped`.

This PR further fixes the formatting of named import specifiers to more closely match Prettier's formatting.

## Test Plan

`cargo test` and I compared the snapshot changes with Prettier to make sure they match.


MichaReiser:
!bench_formatter

ematipico (approved):

MichaReiser:
!bench_formatter

PR #2771, opened 2022-06-23
feat(rome_formatter): will_break utility

ematipico:
## Summary

This PR adds two new APIs to the formatter:

- `inspect_will_break`: this API creates a new buffer where the element gets written in the main buffer, but it also uses the `FormatElement::will_break` to track if the element written in the buffer breaks. The consumer has to call `.finish()` to retrieve that information. This means that the information can be retrieved only after the element is written in the buffer.
- `inspect_null`: this API creates a buffer where all the formatted elements are ignored. This can be useful to compute information **before the elements are written in the main buffer**.

I created a couple of examples that work with the "will break" functionality.

The PR also adds a couple of bug fixes around call arguments. The checks come from prettier, and they are not aligned with it; here's the source code: https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/call-arguments.js#L235-L244

## Test Plan

`cargo test` and checked that the snapshots matches prettier

PR 
**File Based Average Prettier Similarity**: 76.41%  
**Line Based Average Prettier Similarity**: 71.98%  

`main`
**File Based Average Prettier Similarity**: 76.38%  
**Line Based Average Prettier Similarity**: 71.92%


crates/rome_formatter/src/builders.rs
```diff
@@ -1892,3 +1892,111 @@ impl<Context> Format<Context> for BestFitting<'_, Context> {
...
+///     "break\nhello\nworld!",
+///     formatted.print().as_code()
+/// );
+/// ```
+pub fn will_break<Context>(
+    element: impl Format<Context>,
+    f: &mut Formatter<Context>,
+) -> FormatResult<bool> {
+    let mut will_break = WillBreakBuffer::new(f);
+    let snapshot = will_break.snapshot();
```
MichaReiser, line 1943: Taking snapshots is rather expensive and may also fail if the caller doesn't call `memoize` on the `element` first (because it then happens to format tokens twice). It further wastes time by writing the formatted content into `f` to then discard all the work again at the end. 

I do see two different use cases:

**Decide layout of later content** 

Let's say we format something like a binary expression and the layout of the right-hand side depends on whatever the left hand side breaks. In pseudo code:

```
let mut buffer = f.inspect_will_break(); 
write!(buffer [left.format()])?;
let will_break = buffer.finish();

if will_break {
  write!(f, [group_elements(&right.format())])?;
} else {
  write!(f, [right.format()])?;
```

This case doesn't require any rewinding because the layout decision only affects elements that will be written at a later point in time. This use case works well with the extension method. 

**Deciding its own layout**

There are cases where the layout of the item itself depends on whatever the formatted item breaks. I believe this is the case for call arguments where the layout of the call arguments depends on whatever any of the arguments break. 

My proposal here is to introduce a new `NullBuffer` that does nothing... 

```rust
impl<Context> Buffer for NullBuffer<Context> {
  type Context = Context;
  
  fn write_element(&mut self, element: FormatElement) {
    // drop  the element 
  }
  
  fn snapshot(&self) -> BufferSnapshot {
    BufferSnapshot::Position(0)
  }
  
  fn restore(&mut self, snapshot: BufferSnapshot) {
     // buffer has no state, drop snapshot
  }
}
```

I don't have yet a good proposal on how to wrap this in a utility function but call arguments could use it like this

```
// Memoize to avoid formatting tokens more than once
let format_arguments = arguments.format().memoized();
let will_break = {
  let mut buffer = NullBuffer::new(f.state_mut());
  let mut buffer = buffer.inspect_will_break();
  
  write!(buffer, [arguments])?;
  
 buffer.finish()  
};

if will_break {
  write!(f, [group
}
```

This has the advantage that the arguments only get formatted once, as well as that the elements are only written once to a buffer.

The thing that I still have in mind but haven't been able to come up with a good design yet is if `memoized` would give you access to the formatted element or allow you to inspect it

```
let format_arguments = arguments.format().memoized();

let will_break = format_arguments.cache(f)?.will_break();
```

I think this would be the most ergonomic API and doesn't have the shortcomings of the two other approaches because:
* It interns the `FormatElement`: Writing it to the final output only involves writing a single element (and not the whole nested list)
* The written elements don't get discarded

I'll spend a little time to see if I can come up with an API that would also work in a world with a FlatIR.

MichaReiser (changes requested):

crates/rome_formatter/src/formatter.rs
```diff
@@ -176,7 +176,7 @@ impl<'buf, Context> Formatter<'buf, Context> {
 impl<Context> Formatter<'_, Context> {
     /// Take a snapshot of the state of the formatter
     #[inline]
-    pub fn snapshot(&self) -> FormatterSnapshot {
+    pub fn state_snapshot(&self) -> FormatterSnapshot {
```
ematipico, line 179: These two functions were going in conflict with the ones created by the `Buffer` trait. While using them in some cases, the compiler wasn't able to compute the correct types.

PR #2772, opened 2022-06-24
Feature/inspect memoized

MichaReiser:
## Summary

This PR adds the `inspect` method to `Memoized` that allows inspecting the formatted content.

This can be useful in scenarios where some formatting must determine if some content will break to decide the best formatting.

This is different from https://github.com/rome/tools/pull/2771 because the buffer should be preferred in cases where the fact whatever some content breaks is only used to determine the formatting of another element because it avoids interning/memoizing the inner content.

The API shown here should be preferred when the formatting of the element itself depends on whatever its content breaks where using the `WillBreakBuffer` would need to rewind to the previous state.

## Test Plan

I added a new doc test showing how to use the new API


crates/rome_formatter/src/format_extensions.rs
```diff
@@ -159,43 +160,117 @@ where
...
+    ///
+    /// ```
+    pub fn inspect(&mut self, f: &mut Formatter<Context>) -> FormatResult<&FormatElement> {
+        let cache = self.memory.get_mut();
+        if cache.is_none() {
+            Self::format_and_store(cache, &self.inner, f)?;
+        }
+
+        // SAFETY: Safe because `format_and_store` populates the cache
+        match cache.as_ref().unwrap() {
```
leops, line 222: It might be possible to use `cache.get_or_insert_with()` here to read the cache and lazily fill it without having to unwrap

leops (approved):

crates/rome_formatter/src/format_extensions.rs
MichaReiser: Much better, thanks!

MichaReiser:
@ematipico please merge the PR if you are happy with it as I'll be on PTO.

MichaReiser:
@denbezrukov this or #2771 might be useful for you when testing if an element has a specific label. See the summary of this PR to when to use the `memoized.inspect` vs the `WillBreakBuffer`

crates/rome_formatter/src/format_extensions.rs
```diff
@@ -159,43 +160,86 @@ where
...
+    /// use rome_formatter::{format, write};
+    /// use rome_formatter::prelude::*;
+    /// use rome_rowan::TextSize;
+    ///
+    /// #[derive(Default)]
+    /// struct Counter {
+    ///   value: Cell<u64>
+    /// }
+    ///
+    /// impl Format<SimpleFormatContext> for Counter {fn fmt(&self, f: &mut Formatter<SimpleFormatContext>) -> FormatResult<()> {
```
denbezrukov, line 183: It seems that need to add new line here🙂

crates/rome_formatter/src/format_extensions.rs
```diff
@@ -159,43 +160,86 @@ where
...
+    ///             hard_line_break()
+    ///         ])?;
+    ///         self.value.set(current + 1);
+    ///         Ok(())
+    ///     }
+    /// }
+    ///
+    /// let content = format_with(|f| {
+    ///     let mut counter = Counter::default().memoized();
+    ///     let counter_content = counter.inspect(f)?;
```
denbezrukov, line 199: This API looks great! 🙀

ematipico (approved):
I've started to ponder if we should rename `VecBuffer` into `ElementBuffer`. `Vec` is way too generic, while it's actually the buffer where we want to write `FormatElement`

MichaReiser:
> I've started to ponder if we should rename `VecBuffer` into `ElementBuffer`. `Vec` is way too generic, while it's actually the buffer where we want to write `FormatElement`

I'd like to keep `VecBuffer` because it best describes its data structure. The fact that we use `into_element` is more a hack and I ultimately want to get to a place where the `Formatter` uses an internal `Interner` that is shared between all `memoized` elements.

PR #2773, opened 2022-06-24
feature(rome_js_formatter): TS Interface formatting

MichaReiser:
## Summary

This PR improves the formatting of TS interface declarations. 

> **Note**: ~~There's an unresolved stability issue around formatting comments which will require #2768 to be resolved~~

## Test Plan

Verified the updated snapshots manually and compared them with prettier

**File Based Average Prettier Similarity**: 76.01%  -> 76.26%  
**Line Based Average Prettier Similarity**: 70.85%  -> 71.23%

### DIFF
--- a/crates/rome_js_formatter/src/ts/declarations/interface_declaration.rs
+++ b/crates/rome_js_formatter/src/ts/declarations/interface_declaration.rs
@@ -1,6 +1,6 @@
 use crate::prelude::*;
 
-use rome_formatter::write;
+use rome_formatter::{format_args, write};
 use rome_js_syntax::{TsInterfaceDeclaration, TsInterfaceDeclarationFields};
 
 #[derive(Debug, Clone, Default)]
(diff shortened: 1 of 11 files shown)


MichaReiser:
@ematipico I'll leave it to you to decide if you want to ignore the problematic lines and add it to the Tasks section of #2768 or leave the PR open

MichaReiser:
I resolved the instable comment formatting by manually handling comments inside of the interface formatter. This isn't ideal but lets us move forward

crates/rome_js_formatter/src/ts/declarations/interface_declaration.rs
```diff
@@ -18,26 +18,129 @@ impl FormatNodeRule<TsInterfaceDeclaration> for FormatTsInterfaceDeclaration {
...
+            Ok(())
+        });
+
+        write![f, [interface_token.format(), space_token()]]?;
+
+        // Manually handle the trailing comments and push them into the members block
+        // to prevent that a comment gets formatted on the same line as the opening `{`
+        let last_node = match (&type_parameters, &extends_clause) {
+            (_, Some(extends_clause)) => extends_clause.syntax(),
+            (Some(type_parameters), None) => type_parameters.syntax(),
```
ematipico, line 74: nit: Both `None` of both `_`? Just to be consistent

ematipico (approved):

crates/rome_js_formatter/src/ts/declarations/interface_declaration.rs
MichaReiser: In the first case it can be `Some(_) | None`, that's why I used `_`. In the second case it can only ever be `None` (guaranteed by the match above)

PR #2774, opened 2022-06-24
feat(rome_rowan): expose the kind of AstNode as a constant

leops:
## Summary

This PR adds a new `KIND_SET` associated constant to the `AstNode`, representing the set of `SyntaxKind` values that can be cast into this node (the `can_cast` function would return true for these values). This allows code to statically list out all node types that match a certain node type, to optimize query matching in the analyzer for instance.

## Example

This code uses the new feature to list the variants of `JsSyntaxKind` that can be cast into `JsAnyRoot`:

```rust
let root_kinds: Vec<_> = JsAnyRoot::KIND_SET.iter().collect();
assert_eq!(root_kinds.as_slice(), &[
    JsSyntaxKind::JS_MODULE,
    JsSyntaxKind::JS_SCRIPT,
    JsSyntaxKind::JS_EXPRESSION_SNIPPED,
]);
```

## Test Plan

This is mostly a codegen change, and with most operations on `SyntaxKindSet` being `const` almost all these operations are checked at compile time (adding enough variants to `JsSyntaxKind` to make it overflow the 64 bytes of the bitfield would cause a compilation error for instance).
The changes to the rule registry are tested indirectly through the integration tests of the analyzer, ensuring that queries are still matched correctly

### DIFF
--- a/crates/rome_analyze/src/registry.rs
+++ b/crates/rome_analyze/src/registry.rs
@@ -1,7 +1,7 @@
-use std::{iter::Map, vec::IntoIter};
+use std::collections::HashSet;
 
 use rome_diagnostics::file::FileId;
-use rome_rowan::{AstNode, Language, SyntaxNode};
+use rome_rowan::{AstNode, Language, SyntaxKind, SyntaxNode};
 
 use crate::{
     context::RuleContext,
@@ -11,12 +11,12 @@ use crate::{
 
 /// The rule registry holds type-erased instances of all active analysis rules
 pub struct RuleRegistry<L: Language> {
-    rules: Vec<RegistryRule<L>>,
+    nodes: Vec<KindRules<L>>,
 }
 
 impl<L: Language> RuleRegistry<L> {
     pub fn empty() -> Self {
-        Self { rules: Vec::new() }
+        Self { nodes: Vec::new() }
     }
 
     pub fn push<R>(&mut self)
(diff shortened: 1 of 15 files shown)


crates/rome_rowan/src/ast/mod.rs
```diff
@@ -15,16 +15,80 @@ use text_size::TextRange;
...
 pub use mutation::{AstNodeExt, AstNodeListExt, AstSeparatedListExt};
 
+/// Represents a set of [SyntaxKind] as a bitfield, with each bit representing
+/// whether the corresponding [RawSyntaxKind] value is contained in the set
+///
+/// This is similar to the `TokenSet` struct in `rome_js_parser`, with the
+/// bitfield here being twice as large as it needs to cover all nodes as well
+/// as all token kinds
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct SyntaxKindSet<L: ?Sized + Language>([u128; 4], PhantomData<L>);
```
MichaReiser, line 28: Would the solution be to increase the array length if it happens that we encounter a union that has more than 32 (4*128/32) variants?

crates/rome_rowan/src/green.rs
```diff
@@ -18,6 +18,20 @@ pub(crate) use self::node_cache::NodeCacheNodeEntryMut;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct RawSyntaxKind(pub u16);
 
+impl RawSyntaxKind {
+    pub(crate) const fn as_bits(self) -> [u128; 4] {
```
MichaReiser, line 22: Can you add some documentation to this part that explains what is happening here? I wouldn't know what to make of this function or how to change it in the future.

MichaReiser (approved):

crates/rome_rowan/src/ast/mod.rs
leops: The length of the array would need to be increased if any `SyntaxKind` ever got more than 512 (128*4) variants, currently the largest one is `JsSyntaxKind` with 481 variants

crates/rome_rowan/src/green.rs
leops: I've inlined the `as_bits` function in the 2 places it's used in `SyntaxKindSet` to make it easier to understand what it's doing in context. I also modified how the function is implemented to make easier to extend the size of the bitfield without having to introduce additional branches (I wasn't sure Rust's constant folding would be able to statically check it for error like this, but it turned out to work just as well)

crates/rome_analyze/src/registry.rs
```diff
@@ -11,31 +11,52 @@ use crate::{
...
+            .into_iter()
+            .flat_map(|node| node.rules)
+            .map(|rule| (rule.name, rule.docs))
+            .filter(move |(name, _)| unique.insert(name.as_ptr() as u64))
     }
 }
 
-pub type MetadataIter<L> =
-    Map<IntoIter<RegistryRule<L>>, fn(RegistryRule<L>) -> (&'static str, &'static str)>;
+struct KindRules<L: Language> {
```
ematipico, line 51: What's a `KindRules`? Should it be a `RuleKind`? Could you add some doc?

ematipico (changes requested):
Could you please add some example, somewhere, of how to use this new feature?

ematipico:
> > Could you please add some example, somewhere, of how to use this new feature?
> 
> The new feature is being used in [registry.rs](https://github.com/rome/tools/blob/7ff535932b53e1a3b294646a166c9a5bb271a632/crates/rome_analyze/src/registry.rs#L29) to store rules in a flat vector, with each index of the vector being associated with the corresponding `RawSyntaxKind` value allowing for constant-time lookup of lint rules during traversal. That's a real-world use case though, I don't know if a more synthetic example would be easier to understand

That's exactly why I think having a documented example would be beneficial for us. I looked at it and honestly I didn't understand some parts of it (for example the `.to_raw().0` seemed quite hacky, and a documentation that explains it might help).

crates/rome_analyze/src/registry.rs
ematipico: @leops I think you missed this comment. I believe this should be `RulesKind`. `KindRules` had a different meaning 😝

leops: Interesting, I added some documentation to this struct but it's not showing in the code block attached to the comment in the conversation tab.
I think `KindRules` feels more natural to read left-to-right since it maps a single `SyntaxKind` to multiple `Rule`, whereas `RulesKind` feels more ambiguous: the `-Kind` suffix is generally used for enums, so it could be understood as "the kind of a collection of rules" when it's in fact a struct holding said collection of rules

ematipico: It's more about english grammar here 😅 "kind rules" is like "kind person", which is the adjective to address a person that is "sweet, gentle, etc.", which is not the meaning we want to have. I'd suggest to find another wording

leops: I guess then `SyntaxKindRules` would work, it makes it clearer that "kind" refers to the `SyntaxKind` trait

PR #2779, opened 2022-06-24
refactor(rome_js_formatter): Remove `Copy` from `JsFormatContext`

MichaReiser:
## Summary

Remove the `Copy` from `JsFormatContext` to allow storing non-copyable data in the context. The idea is to store the extracted comments as an `Rc<Comments>` but this isn't possible today because `Rc` (nor `Comments`) implements `Copy`.

## Test Plan

The rust compiler is so kind to point me towards all places where we now need to call `clone` (only in tests).

### DIFF
--- a/crates/rome_js_formatter/src/context.rs
+++ b/crates/rome_js_formatter/src/context.rs
@@ -8,25 +8,24 @@ use std::fmt;
 use std::fmt::Debug;
 use std::str::FromStr;
 
-#[derive(Debug, Copy, Clone, Default)]
+#[derive(Debug, Clone, Default)]
 pub struct JsFormatContext {
     /// The indent style.
-    pub indent_style: IndentStyle,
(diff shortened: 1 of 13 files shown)


crates/rome_js_formatter/src/context.rs
```diff
@@ -8,25 +8,24 @@ use std::fmt;
...
     /// The indent style.
-    pub indent_style: IndentStyle,
+    indent_style: IndentStyle,
 
     /// What's the max width of a line. Defaults to 80.
-    pub line_width: LineWidth,
+    line_width: LineWidth,
 
-    /// Options of current instance of the formatter
-    pub options: JsFormatOptions,
```
MichaReiser, line 20: I decided to inline this type as it only holds one field.

leops (approved):

PR #2780, opened 2022-06-24
 feature(rome_js_formatter): Track suppression handling 

MichaReiser:
## Summary

This PR adds a debug-only tracking that verifies if the formatter checked the suppression comments for every node. Doing so is useful in cases where our implementation bypasses the node's `FormatNodeRule` implementation that otherwise handles suppressions automatically.

## Test Plan

Fixed the existing tests that started failing because of missing suppression checks.

### DIFF
--- a/crates/rome_js_formatter/src/context.rs
+++ b/crates/rome_js_formatter/src/context.rs
@@ -1,8 +1,11 @@
+#[cfg(debug_assertions)]
+use indexmap::IndexSet;
 use rome_formatter::printer::PrinterOptions;
 use rome_formatter::{
     CommentContext, CommentKind, CommentStyle, FormatContext, IndentStyle, LineWidth,
 };
-use rome_js_syntax::{JsLanguage, JsSyntaxKind, SourceType};
+use rome_js_syntax::suppression::{has_suppressions_category, SuppressionCategory};
+use rome_js_syntax::{JsLanguage, JsSyntaxKind, JsSyntaxNode, SourceType};
 use rome_rowan::SyntaxTriviaPieceComments;
 use std::fmt;
 use std::fmt::Debug;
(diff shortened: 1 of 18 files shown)


MichaReiser:
I won't be able to finish this before my PTO. I believe that having this in place is valuable to make sure we don't accidentally format nodes that the user explicitly ignored. This infrastructure will also allow us to identify the places where we manually need to call into `format_leading_comments` and `format_trailing_comments` if we decide to format node comments differently (there will probably be a few more places, but it's a good start). @leops , @ematipico , @yassere: would anyone of you be able to finish this wile I'm on leave?

crates/rome_js_formatter/src/context.rs
```diff
@@ -21,6 +24,8 @@ pub struct JsFormatContext {
 
     /// Information relative to the current file
     source_type: SourceType,
+
+    checked_node_suppressions: IndexSet<JsSyntaxNode>,
```
MichaReiser, line 28: Ideally, this would be part of `FormatState` but the suppression comments is a `rome_js_syntax` only feature :(

ematipico: I think it makes sense though? The comments suppression is language specific

ematipico (commented):
Maybe we should document somewhere this new usage. Maybe inside the `.format()` function, or inside the `formatter` module. 

The thing is, the usage of `is_suppressed` is a bit hidden and it requires few steps in order to use it, e.g. `f.context_mut().is_suppressed()`.

crates/rome_js_formatter/src/context.rs
MichaReiser: The extraction is but not what the formatter does in that case.

ematipico:
> My current thinking was that raised panic will hint you towards the method if you run into a situation where you apply custom formatting

True. Still we suggest this only if the error occurs. We should also document the API with some example. 


> The only other option that I see is to document it inside of a readme but that would need to be specific to rome_js_formatter because the current implementation (this will likely change) is JS specific (and is the reason why I can't mention it inside `Format::fmt`.

We could deploy the rust documentation of `rome_js_formatter` and document it there. We decided to use rust documentation to explain all the development details. Gradually we will update all the `README`/docs to point to the rust documentation.

PR #2783, opened 2022-06-24
feat(rome_js_formatter): add format element label

denbezrukov:
## Summary

This PR introduces the new IR element `Label`. This IR matches Prettier's `label` command.
- https://github.com/prettier/prettier/blob/main/commands.md#label

This IR can be useful if representation depends on different representation of child content. 
E.g., to decide how to print an assignment expression, we might want to know whether its right-hand side has been printed as a method call chain, not as a plain function call. 

## Test Plan

Added a new doc showing how to use the new IR

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -523,6 +523,88 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
     }
 }
 
+/// Marks some content with a label.
+///
+/// This does not directly influence how this content will be printed, but some
+/// parts of the formatter may inspect the `FormatElement::Label` and chose to handle this element in a specific way
+///
+/// # Example
+///
+/// ```rust
+/// use rome_formatter::prelude::*;
+/// use rome_formatter::{format, write};
+///
+/// #[derive(Default)]
+/// struct Labelled;
+///
+/// impl Format<SimpleFormatContext> for Labelled {
+///     fn fmt(&self, f: &mut Formatter<SimpleFormatContext>) -> FormatResult<()> {
+///         write!(f, [labelled("some-label", &token("labelled"))])?;
+///         Ok(())
+///     }
+/// }
+///
+/// let content = format_with(|f| {
+///     let mut counter = Labelled::default().memoized();
+///     let counter_content = counter.inspect(f)?;
+///
+///     let is_labelled = match counter_content {
+///         FormatElement::Label(label) => label.label() == "some-label",
+///         _ => false,
+///     };
+///
+///     if is_labelled {
+///         write!(f, [token("This is "), &counter])
+///     } else {
+///         write!(f, [token("This is not "), &counter])
+///     }
+/// });
+///
+/// let formatted = format!(SimpleFormatContext::default(), [content]).unwrap();
+/// assert_eq!("This is labelled", formatted.print().as_code())
+/// ```
+#[inline]
+pub fn labelled<'a, Content, Context>(
+    label: &'static str,
+    content: &'a Content,
+) -> FormatLabelled<'a, Context>
+where
+    Content: Format<Context>,
+{
+    FormatLabelled {
+        label,
+        content: Argument::new(content),
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct FormatLabelled<'a, Context> {
+    label: &'static str,
+    content: Argument<'a, Context>,
+}
+
+impl<Context> Format<Context> for FormatLabelled<'_, Context> {
+    fn fmt(&self, f: &mut Formatter<Context>) -> FormatResult<()> {
+        let mut buffer = VecBuffer::new(f.state_mut());
+
+        buffer.write_fmt(Arguments::from(&self.content))?;
+        let content = buffer.into_vec();
+
+        let label = Label::new(self.label, content);
+
+        f.write_element(FormatElement::Label(label))
+    }
+}
+
+impl<Context> std::fmt::Debug for FormatLabelled<'_, Context> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Label")
+            .field(&"{{name}}")
+            .field(&"{{content}}")
+            .finish()
+    }
+}
+
 /// Inserts a single space. Allows to separate different tokens.
 ///
 /// # Examples
--- a/crates/rome_formatter/src/format_element.rs
+++ b/crates/rome_formatter/src/format_element.rs
@@ -70,6 +70,11 @@ pub enum FormatElement {
     /// An interned format element. Useful when the same content must be emitted multiple times to avoid
     /// deep cloning the IR when using the `best_fitting!` macro or `if_group_fits_on_line` and `if_group_breaks`.
     Interned(Interned),
+
+    /// Special semantic element marking the content with a label.
+    /// This does not directly influence how the content will be printed.
+    /// See [crate::labelled] for documentation.
+    Label(Label),
 }
 
 #[derive(Clone, Copy, Eq, PartialEq, Debug)]
@@ -150,6 +155,10 @@ impl Debug for FormatElement {
             }
             FormatElement::ExpandParent => write!(fmt, "ExpandParent"),
             FormatElement::Interned(inner) => inner.fmt(fmt),
+            FormatElement::Label(label) => {
+                write!(fmt, "Label")?;
+                label.fmt(fmt)
+            }
         }
     }
 }
@@ -352,6 +361,34 @@ impl Deref for Interned {
     }
 }
 
+#[derive(Clone, PartialEq, Eq)]
+pub struct Label {
+    pub(crate) content: Box<[FormatElement]>,
+    label: &'static str,
+}
+
+impl Label {
+    pub fn new(label: &'static str, content: Vec<FormatElement>) -> Self {
+        Self {
+            content: content.into_boxed_slice(),
+            label,
+        }
+    }
+
+    pub fn label(&self) -> &'static str {
+        self.label
+    }
+}
+
+impl Debug for Label {
+    fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
+        fmt.debug_struct("")
+            .field("label", &self.label)
+            .field("content", &self.content)
+            .finish()
+    }
+}
+
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct ConditionalGroupContent {
     pub(crate) content: Content,
@@ -509,7 +546,9 @@ impl FormatElement {
             FormatElement::Space => false,
             FormatElement::Line(line_mode) => matches!(line_mode, LineMode::Hard | LineMode::Empty),
             FormatElement::Indent(content) => content.will_break(),
-            FormatElement::Group(Group { content, .. }) | FormatElement::Comment(content) => {
+            FormatElement::Group(Group { content, .. })
+            | FormatElement::Comment(content)
+            | FormatElement::Label(Label { content, .. }) => {
                 content.iter().any(FormatElement::will_break)
             }
             FormatElement::ConditionalGroupContent(group) => group.content.will_break(),
--- a/crates/rome_formatter/src/lib.rs
+++ b/crates/rome_formatter/src/lib.rs
@@ -47,7 +47,7 @@ pub use arguments::{Argument, Arguments};
 pub use buffer::{Buffer, BufferExtensions, BufferSnapshot, Inspect, PreambleBuffer, VecBuffer};
 pub use builders::{
     block_indent, comment, empty_line, get_lines_before, group_elements, hard_line_break,
-    if_group_breaks, if_group_fits_on_line, indent, line_suffix, soft_block_indent,
+    if_group_breaks, if_group_fits_on_line, indent, labelled, line_suffix, soft_block_indent,
     soft_line_break, soft_line_break_or_space, soft_line_indent_or_space, space_token, token,
     BestFitting,
 };
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -295,6 +295,12 @@ impl<'a> Printer<'a> {
                 }
             }
             FormatElement::Interned(content) => queue.enqueue(PrintElementCall::new(content, args)),
+            FormatElement::Label(label) => queue.extend(
+                label
+                    .content
+                    .iter()
+                    .map(|element| PrintElementCall::new(element, args)),
+            ),
         }
     }
 
@@ -832,6 +838,12 @@ fn fits_element_on_line<'a, 'rest>(
             }
         }
         FormatElement::Interned(content) => queue.enqueue(PrintElementCall::new(content, args)),
+        FormatElement::Label(label) => queue.extend(
+            label
+                .content
+                .iter()
+                .map(|element| PrintElementCall::new(element, args)),
+        ),
     }
 
     Fits::Maybe


crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,88 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
...
+///
+/// impl Format<SimpleFormatContext> for Labelled {
+///     fn fmt(&self, f: &mut Formatter<SimpleFormatContext>) -> FormatResult<()> {
+///         write!(f, [labelled("some-label", &token("labelled"))])?;
+///         Ok(())
+///     }
+/// }
+///
+/// let content = format_with(|f| {
+///     let mut counter = Labelled::default().memoized();
```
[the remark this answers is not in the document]
denbezrukov: Done

crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,88 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
...
+        let label = Label::new(self.label, content);
+
+        f.write_element(FormatElement::Label(label))
+    }
+}
+
+impl<Context> std::fmt::Debug for FormatLabelled<'_, Context> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Label")
+            .field(&"{{name}}")
```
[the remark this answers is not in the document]
denbezrukov: Done

denbezrukov:
@MichaReiser Could you help please?
Hm,
There is an assertion with `static_assert!(std::mem::size_of::<crate::FormatElement>() == 32usize);`
May be we can:

- use `Label(Box<Label>)` instead `Label(Label)`.
- use
```
#[derive(Clone, PartialEq, Eq)]
pub struct Label {
    pub(crate) content: Box<[FormatElement]>,
    label: Box<&'static str>,
}
```
- use another type for lable? `u16`?

MichaReiser:
I don't think using dynamic strings is a good idea as these labels drive the formatting logic. Therefore, using a static string sounds good, in which case the size is 24bytes, and 32 for FormatElement. Or is the assertion triggering with a static string?

An alternative is to do something similar to GroupId where it is a wrapper for a usize but holds a name in debug builds. But I don't think we need this optimisation just yet

MichaReiser:
Or you could try to make Label a thin wrapper around TypeId. That has the nice added benefit that the compiler assigns compile time constants for each label. 

Each label would the have its own Type (zero type struct or enum).

It would probably make sense to add a debug only name field, which you can automatically derive by using type_name

'''Rust
#[derive(Eq, PartialEq, Copy, Clone)]
pub struct Label {
  id: TypeId,
  #[cfg(debug_assertions)]
  label: &'static str
}

impl Label { 
  pub fn of<T>() -> Self {
    Self {
      id: TypeId::of<T>(),
      #[cfg(debug_assertions)]
      label: type_name::<T>()
    }
  }
'''

crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,90 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
...
+        let label = Label::new(self.label_id, content);
+
+        f.write_element(FormatElement::Label(label))
+    }
+}
+
+impl<Context> std::fmt::Debug for FormatLabelled<'_, Context> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Label")
+            .field(&self.label_id)
```
denbezrukov, line 604: It looks:
```
label_id: LabelId {
        id: TypeId {
            t: 4451653586406524753,
        },
        label: "rome_formatter::arguments::tests::test_nesting::SomeChain",
    },
```

crates/rome_formatter/src/format_element.rs
```diff
@@ -352,6 +364,51 @@ impl Deref for Interned {
...
 
+#[derive(Eq, PartialEq, Copy, Clone, Debug)]
+pub struct LabelId {
+    id: TypeId,
+    #[cfg(debug_assertions)]
+    label: &'static str,
+}
+
+impl LabelId {
+    pub fn of<T: ?Sized + 'static>() -> Self {
```
denbezrukov, line 375: It's unstable api to make this function `const`.

- https://github.com/rust-lang/rust/issues/77125

crates/rome_formatter/src/format_element.rs
```diff
@@ -352,6 +364,51 @@ impl Deref for Interned {
...
+
+    pub fn label_id(&self) -> LabelId {
+        self.label_id
+    }
+}
+
+impl Debug for Label {
+    fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
+        fmt.debug_struct("")
+            .field("label_id", &self.label_id)
```
denbezrukov, line 406: It looks:
```
label_id: LabelId {
        id: TypeId {
            t: 4451653586406524753,
        },
        label: "rome_formatter::arguments::tests::test_nesting::SomeChain",
    },
```

denbezrukov:
>  Or is the assertion triggering with a static string?

Yes, it is. I had two solutions how to handle this.

-  use `Label(Box<Label>)` as a variant for `FormatElement`.  
-  use `Box<&'static str>` as a field for `label`.



> Or you could try to make Label a thin wrapper around TypeId. 

This way is great 😻 Thank you! I've edited the example of using this new IR.

- https://github.com/rome/tools/pull/2783/files#diff-ef3073c2b45df5f202b850b362d40cb413b493036a5f726cb172fe25cf56645eR533-R569

Only one problem that we can't use `const` in `pub const fn of<T>() -> Self`.  Because this API is unstable. I attached the link in the comment.

crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,90 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
     }
 }
 
+/// Marks some content with a label.
+///
+/// This does not directly influence how this content will be printed, but some
+/// parts of the formatter may inspect the `FormatElement::Label` and chose to handle this element in a specific way
```
ematipico, line 529: What do you mean by "handle this element in a specific way"? Would it be better to explain it by saying "and choose a different formatting layout".

Also, could you explain if a formatter should check the **value of a label** or not? This will help to explain the usage of the feature

crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,90 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
...
+/// use rome_formatter::{format, write, LabelId};
+///
+/// enum SomeLabelId {}
+///
+/// #[derive(Default)]
+/// struct Labelled;
+///
+/// impl Format<SimpleFormatContext> for Labelled {
+///     fn fmt(&self, f: &mut Formatter<SimpleFormatContext>) -> FormatResult<()> {
+///         let label_id = LabelId::of::<SomeLabelId>();
```
ematipico, line 544: I think we can improve the usage of `LabelId` by adding a new function to `BufferExtensions`:

```rs
    fn label_of<T: ?Sized + 'static>() -> LabelId {
        LabelId {
            id: TypeId::of::<T>(),
            #[cfg(debug_assertions)]
            label: type_name::<T>(),
        }
    }
```

And then the buffer can create it for us:

```rs
let label_id = f.label_of::<SomeLabelId>();
```

Doing so we might be able to keep `LabelId` private

crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,90 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
...
+///
+/// let content = format_with(|f| {
+///     let mut labelled = Labelled::default().memoized();
+///     let labelled_content = labelled.inspect(f)?;
+///     let label_id = LabelId::of::<SomeLabelId>();
+///
+///     let is_labelled = match labelled_content {
+///         FormatElement::Label(labelled) => labelled.label_id() == label_id,
+///         _ => false,
+///     };
```
ematipico, line 559: I am not sure this is the correct usage. This example shows that the consumer is forced to extract manually the `FormatElement` which is a low-level API for our formatter.

I wonder instead if we should provide an API, something like `f.label_assert_of`/`label_id.assert_of` (it asserts if the type of the label), which returns a boolean for us.

The suggestion is vague I still need to understand how we do want to use the label and it can be used inside a real world example.

crates/rome_formatter/src/builders.rs
denbezrukov: Nice catch! 
I'll try to add.

crates/rome_formatter/src/builders.rs
denbezrukov: I've noticed this prettier API then I looked into `isPoorlyBreakableMemberOrCallChain` for variable assignment.

- https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/assignment.js#L329

They use label to understand that `CallExpression` has been printed as a `member-chain`. 
I've just realized that it seems they traverse tree to extract `label`.  Because they call `isPoorlyBreakableMemberOrCallChain` with properties array. 

- https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/assignment.js#L216

ematipico: We would need to come up with a different solution here, I suppose. I am not sure if `printCallExpression` actually prints in their main buffer or not (it seems not, which works for their logic). Regardless, they are able to extract the IR for they right-hand side of the assignment like expression and then decide the layout. 

Our formatter works left-to-right, which means that once we write the right-hand side, it's there, unless we write it in a temporary buffer (which can be expensive, so we should avoid it). 

I would suggest another solution for the assignment case. They create the label only when they actually create a member chain. They don't create the member chain in a specific case: https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/member-chain.js#L342-L347

That condition is this one: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/utils/member_chain/groups.rs#L42-L48

If we are able to use that logic inside the assignment like formatting, we might be able to not use the label.

Otherwise, the only solution that I can see is the write the right-hand side inside a temporary buffer, than inspect that buffer and check the label. But as said before, this has a big impact on memory usage.

ematipico: I am not sure if that would work. Mainly because at the end of the format phase, where that function is used, the logic might add parenthesis to the binary expression, and the indentation has to stay inside the parenthesis. That's why checking the AST is better in this case.

denbezrukov: Should we close this PR?:)

ematipico: I think it still makes sense, especially with the new utilities that I implemented in #2771. With that, maybe we can emulate prettier's solution with ease.

It adds a new `f.inspect_null`, where it's possible to dump formatters without writing them in the main buffer. This would allow us to inspect elements "ahead-of-time" and take layout decisions before printing elements.

denbezrukov: Do I understand correctly that after #2771 we can try to implement a buffer which can inspect all elements? E.g. `InspectBuffer` or `InspectLabelBuffer` uses a closure to search labelled elements in IR tree?

ematipico: Yes, that's the idea :)

denbezrukov: Okey,
I'll wait when that PR will be merged.:)

denbezrukov: @ematipico 🙏🏽🙏🏽
Could you please explain difference between two approaches with `will_break`?  
Are they the same? 

```
let mut element = format_with(|f| {
    write!(f, [token("hello"), hard_line_break(), token("world!")])
})
.memoized();

 let will_break =  {
     let mut null_buffer =  f.inspect_null();
     let mut buffer = null_buffer.inspect_will_break();
     write!(buffer, [element])?;
     buffer.will_break()
};
```

```
let mut element = format_with(|f| {
    write!(f, [token("hello"), hard_line_break(), token("world!")])
})
.memoized();

let will_break = {
    element.inspect(f)?.will_break()
};
```

ematipico: In the second case, the `element` gets written in the main buffer. In the first case `null_buffer` works as a sink where the elements are dropped and not written anywhere.

denbezrukov: Got it! Thank you!
So to emulate prettier `label` functionality we can add new extension `inspect_label`, `InspectLabelBuffer` and method for `FormatElement` `has_label` which traverses IR tree and search expected label?

ematipico: > has_label which traverses IR tree and search expected label?

Is this what Prettier does? Does it traverse ALL the IR in order to find a label? I thought it just checks the first element

crates/rome_formatter/src/builders.rs
```diff
@@ -523,6 +523,52 @@ impl<Context> std::fmt::Debug for FormatComment<'_, Context> {
...
 
+/// Marks some content with a label.
+///
+/// This does not directly influence how this content will be printed, but some
+/// parts of the formatter may inspect the [labelled element](FormatElement::Label)
+/// using `inspect_is_labelled` buffer extension and choose a different formatting layout.
+///
+/// See [inspect_is_labelled](BufferExtensions::inspect_is_labelled) for documentation.
+#[inline]
+pub fn labelled<Content, Context>(label_id: LabelId, content: &Content) -> FormatLabelled<Context>
```
denbezrukov, line 534: I hope that we can add here `T: ?Sized + 'static` and remove `label_id`.

```
pub fn labelled<T: ?Sized + 'static, Content, Context>(content: &Content) -> FormatLabelled<Context>
where
    Content: Format<Context>,
{
    FormatLabelled {
        label_id: LabelId::of::<T>(),
        content: Argument::new(content),
    }
}
```

But I'm not sure how to make convenient on call side because we have to write all generics:
`labelled::<SomeLabelId, _, _>(&token("labelled"))`

ematipico: Do we need to pass `Content`? Isn't `&'static str` sufficient?

ematipico:
```rs
pub fn labelled(label_id: &'static -> FormatLabelled<Context> {
	let content = dynamic_token(&std::format!("{label_id}"), TextSize::default());
	FormatLabelled {
        label_id,
        content: Argument::new(content),
    }
}
```

denbezrukov: I thought that the idea behind `labelled` builder is wrapping a `content` with a `label` which we can later  look up in another part of formatter and choose another layout. 
I don't know how to market content with` labelled(label_id: &'static str)` . How do you see it? Will it be combination with `group_elements` ?

ematipico: Ah, then I misunderstood things, sorry! I supposed what got me confused this `labelled::<SomeLabelId, _, _>(&token("labelled")`, but I understand now that the label is actually the type, not the token... which is actually the content. 

I would go for the most straightforward solution for now.

crates/rome_formatter/src/buffer.rs
```diff
@@ -487,6 +487,60 @@ pub trait BufferExtensions: Buffer + Sized {
         WillBreakBuffer::new(self)
     }
 
+    /// It emits a custom buffer called [IsLabelledBuffer], which tracks
+    /// labelled elements written in the main buffer, it does so by
+    /// checking if [element](FormatElement) is [label](FormatElement::Label)
+    /// with expected [label_id](LabelId).
+    ///
+    /// This functionality can be used only one element and only after the element
```
ematipico, line 495:
```suggestion
    /// This functionality can be used only on one element and only after the element
```

ematipico (approved):

leops (approved):

denbezrukov:
@ematipico Could I try to implement `isPoorlyBreakableMemberOrCallChain` after this PR?
https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L329

ematipico:
> @ematipico Could I try to implement `isPoorlyBreakableMemberOrCallChain` after this PR?
> https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L329

Sure go ahead. I think it's the last piece for completing assignments

denbezrukov:
> > @ematipico Could I try to implement `isPoorlyBreakableMemberOrCallChain` after this PR?
> > https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L329
> 
> Sure go ahead. I think it's the last piece for completing assignments

🥳🥳

There is one more piece 
- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L145

I guess that we can implement `canBreak` the same way as `willBreak`.

crates/rome_formatter/src/builders.rs
denbezrukov: > I would suggest another solution for the assignment case. They create the label only when they actually create a member chain. They don't create the member chain in a specific case: https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/member-chain.js#L342-L347
> 
> That condition is this one: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/utils/member_chain/groups.rs#L42-L48
> 

Could you please help me?🙏🏽 I can't find this case in Rome code:

- https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/member-chain.js#L342-L347

Because it seems this conditional:

- https://github.com/rome/tools/blob/main/crates/rome_js_formatter/src/utils/member_chain/groups.rs#L42-L4

is prettier:

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/member-chain.js#L301-L304

PR #2784, opened 2022-06-25
feat(rome_parse): try to extract common code from `rome_js_parse` to `rome_parse`

IWANABETHATGUY:
## Summary
1. We want to extract common patterns from `rome_js_parse` so that we could use this code for other parsers.

## Test Plan
1. cargo test
2. Benchmark should not have regression.


IWANABETHATGUY:
!bench_parser

PR #2785, opened 2022-06-26
feat(rome_json_parser): first version of json parser

IWANABETHATGUY:
## Summary
1. Part of #2786 
2. Implementing JSON parser with easy error recovery.
3. Adding basic errory recover for `trailing comma`
4. Adding some valid basic JSON test cases copy from `prettier`  [json test case](https://github.com/prettier/prettier/tree/main/tests/format/json/json)
5. Adding some invalid `trailing comma` test cases 
6. Extracting common parser logic into a new crate `rome_parse` 

## Test Plan
1. cargo test
2. Check if the new generated `*.rast` is correct.


IWANABETHATGUY:
> Have you tested this parser with any open source JSON test suites? For example: https://github.com/nst/JSONTestSuite

Not yet. (only use prettier json test case.)
The rome member suggested me to use https://github.com/rome/tools/blob/archived-js/internal/codec-config/json/parser-test262.test.ts instead prettier test case, so I think I would like to use `rome` classic json test case.
Also, thanks for sharing the json test case repo

ematipico (changes requested):
Thank you for your contribution! I think we structure the implementation of the parser in a better way.

You should make smaller PRs, so they are better to review. This one is too big, and it contains a lot of information  and it's easier to miss bugs. Even if some code is copied.

Here's some pointers to better structure the implementation:
- no refactors for to extract generic logic (for now)
- fix the grammar (I saw some fixes) with new generated code
- **basic** implementation of the lexer (no rewind/relex logic, as it's not needed)
- **basic** implementation of the parser (happy path)
- implementation of the error paths, with diagnostics
- error recovery

Between each step, if there's need to extract logic from the JS parser to a generic crate, then it makes sense to do it only for that code that is actually needed.

Also, some other pointer when creating a PR. In this particular case, you created a big PR but the description of the PR is very small compared to the work you've done. This won't help the reviewers because you missed lot of information, leaving the reviewers to spend more time to review it. For example you forgot to mentioned:
-  to explain why you added a new crate (`logos`);
- which code you copied from the existing parser and why;
- which code you copied and changed because it wasn't needed;
- which code you made generic and why;

PR #2787, opened 2022-06-27
feat(rome_js_formatter): type alias as assignment like

ematipico:
## Summary

This adds `TsTypeAliasDeclaration` to the list of assignment like expressions. Part of #2424

It also adds some additional condition, around this new variant, for the `BreakLeftHandSide`. 

I had to change a bit how `should_break_after_operator` is used. The function accepts a `JsAnyExpression`, but with TS we have a right hand side that is not an expression. So I just created a thin wrapper to call `has_new_line_before_comment` for every right node. 

At the moment there's a small regression around the indentation of type unions, that's because at the moment these unions are not formatted as a binary like expression. 

I will make a separate PR to include them in the `JsAnyBinaryLikeExpression` union.

I also took the liberty to merge the formatting logic of `JsObjectExpression` and `TsObjectType` into a new union type. This will make sure that we don't diverge too much, and if we need to fix some logic, it's applied to both nodes.

## Test Plan

Current test suite, and checked that the new snapshots are correct.

PR
**File Based Average Prettier Similarity**: 76.38%  
**Line Based Average Prettier Similarity**: 71.92%   

`main`
**File Based Average Prettier Similarity**: 76.34%  
**Line Based Average Prettier Similarity**: 71.91%


crates/rome_js_formatter/src/ts/declarations/type_alias_declaration.rs
```diff
@@ -1,37 +1,25 @@
...
-        let TsTypeAliasDeclarationFields {
-            type_token,
-            binding_identifier,
-            type_parameters,
-            eq_token,
-            ty,
-            semicolon_token,
-        } = node.as_fields();
-
+        let type_token = node.type_token()?;
```
denbezrukov, line 11: Does `type_token` belong to 'left' in `JsAnyAssigmentLike`?

- https://github.com/prettier/prettier/blob/9dd761a6e491ffff3856eea47fb10b4573b351a6/src/language-js/print/type-annotation.js#L77

ematipico: I maintained the same logic we have introduced for variable declarations. There, the `kind` doesn't belong to left. Hence, here `type_token` doesn't belong to left.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -151,6 +153,8 @@ impl RightAssignmentLike {
             RightAssignmentLike::JsAnyExpression(expression) => Some(expression.clone()),
             RightAssignmentLike::JsInitializerClause(initializer) => initializer.expression().ok(),
             RightAssignmentLike::JsAnyAssignmentPattern(_) => None,
+            // TODO: check here
```
leops, line 156: What's this TODO about ?

leops (approved):

PR #2788, opened 2022-06-27
Feat/json lexer

IWANABETHATGUY:
## Summary
1. Add new crate `rome_json_parser`
2. Implementing `json_lexer` with logos


crates/rome_json_parser/src/lexer.rs
```diff
@@ -0,0 +1,542 @@
+use std::ops::Range;
+
+use logos::Logos;
+use rome_diagnostics::Diagnostic;
+use rome_json_syntax::{JsonSyntaxKind, TextRange, TextSize, T};
+
+#[derive(Logos, Debug, PartialEq, Clone, Copy)]
+pub enum TokenKind {
```
ematipico, line 8: In these tokens we're missing double quotes

crates/rome_json_parser/src/lexer.rs
```diff
@@ -0,0 +1,542 @@
...
+    Error,
+    // Root,
+
+    // Array,
+    // Object,
+}
+
+impl From<TokenKind> for JsonSyntaxKind {
+    fn from(kind: TokenKind) -> Self {
+        match kind {
```
ematipico, line 58: I think there's no need to repeat this code

xtask/codegen/js.ungram
```diff
@@ -2017,7 +2017,7 @@ TsMappedTypeAsClause =
 	type: TsType
 
 TsMappedTypeOptionalModifierClause =
-	operator_token: ('+' | '-')
+	operator_token: ('+' | '-')?
```
ematipico, line 2020: We can remove this change

crates/rome_json_parser/src/lexer.rs
IWANABETHATGUY: We don't need to lex the double quotes token separately, since the double quote should always be part of `StringLiteral` token.

ematipico: Do you mean `TokenKind::String`?

IWANABETHATGUY: > String

Yes

MichaReiser:
What's the total size of these new dependencies and the impact on build size? 

How does the new lexer raise unicode and escape sequence errors? How flexible is logos when it comes to error recovery?

IWANABETHATGUY:
> What's the total size of these new dependencies and the impact on build size?
> 
> How does the new lexer raise unicode and escape sequence errors? How flexible is logos when it comes to error recovery?

How could I inspect the bundle size? since the `rome_bin` has not dependent `rome_json_parser` yet

IWANABETHATGUY:
> What's the total size of these new dependencies and the impact on build size?
> 
> How does the new lexer raise unicode and escape sequence errors? How flexible is logos when it comes to error recovery?

This may be a temporary solution.
`logos` provide a basic error recovery solution, it marks the first invalid `char` as an `Error` token, and tries to relexing from next char, until `EOF`. We will eventually replace it with a handwritten lexer. what do you think?

MichaReiser:
> How could I inspect the bundle size? since the rome_bin has not dependent rome_json_parser yet

What's the size of the new crate on its own? What's the size of its dependencies?

>  logos provide a basic error recovery strategy, it marks the first invalid char as an Error token, and tries to relexing from next char, until EOF. We will eventually replace it with a handwritten lexer. what do you think?

My main question is if the lexer of this PR implements the full JSON spec or if there are things massing, e.g. unicode escape sequences, and if so, does logos support adding the missing functionality? 

> We will eventually replace it with a handwritten lexer. what do you think?

I think if that's the ultimate goal then I would recommend copying our existing JavaScript lexer and strip out everything that isn't needed instead. This should be straightforward and already gives us the manual lexer. We can remove duplication in later PRs (e.g. extract a `Source` type) if we think this is valuable. 

My main concern is that contributors now need to understand two lexer that fundamentally are different in design which makes it harder to contribute, review PRs, etc.

IWANABETHATGUY:
> 

make sense.

PR #2789, opened 2022-06-28
feature(rome_js_semantic): more docs for semantic events

xunilrj:
This PR addresses all the comments of https://github.com/rome/tools/pull/2761.

### DIFF
--- a/crates/rome_js_semantic/src/events.rs
+++ b/crates/rome_js_semantic/src/events.rs
@@ -370,11 +370,11 @@ impl SemanticEventExtractor {
     /// }
     /// ```
     ///
-    /// ```a``` declaration will be hoisted to the scope of
-    /// function ```f```.
+    /// `a` declaration will be hoisted to the scope of
+    /// function `f`.
     ///
-    /// This method when called inside the ```if``` scope will return
-    /// the ```f``` scope index.
+    /// This method when called inside the `f` scope will return
+    /// the `f` scope index.
     fn scope_index_to_hoist_declarations(&mut self) -> usize {
         // We should at least have the global scope
         // that do not hoist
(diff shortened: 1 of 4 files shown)


crates/rome_js_semantic/src/events.rs
```diff
@@ -374,26 +354,45 @@ impl SemanticEventExtractor {
...
+    ///         var a;
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// ```a``` declaration will be hoisted to the scope of
+    /// function ```f```.
+    ///
+    /// This method when called inside the ```if``` scope will return
+    /// the ```f``` scope index.
```
ematipico, line 377:
```suggestion
    /// `a` declaration will be hoisted to the scope of
    /// function `f`.
    ///
    /// This method when called inside the `f` scope will return
    /// the `f` scope index.
```

crates/rome_js_semantic/src/events.rs
```diff
@@ -374,26 +354,45 @@ impl SemanticEventExtractor {
...
+        let idx = self.scopes.iter().rev().position(|scope| {
+            matches!(scope.hoisting, ScopeHoisting::DontHoistDeclarationsToParent)
+        });
 
         match idx {
-            None => unreachable!(),
             Some(idx) => idx,
+            // Worst case this will fallback to the global scope
+            // which will be idx = 0
+            None => unreachable!(),
```
ematipico, line 395: Actually, I meant something like this:

```suggestion
            None => unreachable!("We must have a least of scope."),
```

So in the unfortunate case we reach this branch, the compiler will show us this string

ematipico (approved):

PR #2795, opened 2022-06-29
fix(rome_js_formatter): indentation before arrow fn is JSX

ematipico:
## Summary

This PR enables formatting of `jsx` and `tsx` files inside the VS Code extension.

## Test Plan

I tried locally by installing it on my VS Code. The user has to manually enable Rome as default formatter for these files

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -3,11 +3,13 @@
 	"publisher": "rome",
 	"displayName": "Rome",
 	"description": "Rome LSP VS Code Extension",
-	"version": "0.8.0",
+	"version": "0.9.0",
 	"icon": "icon.png",
 	"activationEvents": [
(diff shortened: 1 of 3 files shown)


editors/vscode/package.json
```diff
@@ -3,11 +3,13 @@
 	"publisher": "rome",
 	"displayName": "Rome",
 	"description": "Rome LSP VS Code Extension",
-	"version": "0.8.0",
+	"version": "0.9.0",
```
leops, line 6: I think this change shouldn't have been committed since bumping the version number will trigger a release workflow, and the `0.9.0` version number is reserved for pre-releases of the extension

ematipico: Shoot! Good catch! Yes, it should be here

leops (approved):

PR #2797, opened 2022-06-29
feat(rome_js_formatter): class property members as assignment like

ematipico:
## Summary

This adds `TsPropertySignatureClassMember` and `JsPropertyClassMember` to be formatted as "assignment like"

This PR implements https://github.com/rome/tools/issues/2424

The next fix on the table would be to finish up certain conditions to trigger the correct layout, but with this PR, we completed the number of nodes that require this type of formatting.

## Test Plan

Made sure that the updated snapshots are correct

PR 
**File Based Average Prettier Similarity**: 77.22%  
**Line Based Average Prettier Similarity**: 72.43%  

`main`
**File Based Average Prettier Similarity**: 77.16%  
**Line Based Average Prettier Similarity**: 72.41%

### DIFF
--- a/crates/rome_js_formatter/src/js/classes/property_class_member.rs
+++ b/crates/rome_js_formatter/src/js/classes/property_class_member.rs
@@ -1,38 +1,18 @@
 use crate::prelude::*;
-use rome_formatter::{format_args, write};
-
-use crate::utils::FormatWithSemicolon;
-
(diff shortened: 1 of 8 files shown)


leops (approved):

IWANABETHATGUY:
!bench_formatter

PR #2800, opened 2022-06-30
 fix(rome_js_formatter): breaking logic for member chains

ematipico:
## Summary

This PR adds again a check on the breaking logic of the member chain. The check is around comments. I also fixed the previous logic, where the check for certain comments was incorrectly done.

## Test Plan

`cargo test` and checked that snapshots are correctly updated

PR 
**Line Based Average Prettier Similarity**: 72.59%  

`main`
**File Based Average Prettier Similarity**: 77.25%  
**Line Based Average Prettier Similarity**: 72.45%

### DIFF
--- a/crates/rome_js_formatter/src/utils/member_chain/groups.rs
+++ b/crates/rome_js_formatter/src/utils/member_chain/groups.rs
@@ -1,8 +1,7 @@
+use crate::context::TabWidth;
 use crate::prelude::*;
 use crate::utils::member_chain::flatten_item::FlattenItem;
 use crate::utils::member_chain::simple_argument::SimpleArgument;
-
-use crate::context::TabWidth;
 use rome_js_syntax::JsCallExpression;
 use rome_rowan::{AstSeparatedList, SyntaxResult};
 use std::mem;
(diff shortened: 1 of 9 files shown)


crates/rome_js_formatter/src/utils/member_chain/groups.rs
```diff
@@ -90,7 +93,50 @@ impl Groups {
         let call_expressions_are_not_simple =
             calls_count > 2 && self.call_expressions_are_not_simple()?;
 
-        Ok(call_expressions_are_not_simple)
+        // TODO: add here will_break logic
```
leops, line 96: What exactly is this todo here about ?

leops (approved):

crates/rome_js_formatter/src/utils/member_chain/groups.rs
ematipico: A reminder for me to add more logic here. The logic should decide to break the group if some member of the chain will break. I decided to have a `TODO` here because it's not a small implementation and it requires its own PR.

PR #2803, opened 2022-07-01
feat(rome_js_analyze): useTemplate

### LINKED ISSUE #2750: useTemplate

IWANABETHATGUY:
## Summary
1. Resolved #2750 

## Test Plan
1. cargo test


crates/rome_js_analyze/src/utils.rs
```diff
@@ -1,10 +1,30 @@
-use rome_js_syntax::JsAnyExpression;
```
IWANABETHATGUY, line 1: `is_boolean_literal` isn't used anywhere so I removed it.

crates/rome_js_analyze/src/analyzers/use_template.rs
```diff
@@ -0,0 +1,268 @@
...
+                        JsAnyExpression::JsAnyLiteralExpression(
+                            rome_js_syntax::JsAnyLiteralExpression::JsStringLiteralExpression(_)
+                        )
+                    )
+                }) {
+                    return Some(collections);
+                }
+            }
+            None
+        })
```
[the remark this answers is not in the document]
IWANABETHATGUY: Done

crates/rome_js_analyze/Cargo.toml
```diff
@@ -17,6 +17,7 @@ rome_console = { path = "../rome_console" }
 rome_diagnostics = { path = "../rome_diagnostics" }
 roaring = "0.9.0"
 rustc-hash = "1.1.0"
+snailquote = "0.3.1"
```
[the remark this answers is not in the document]
IWANABETHATGUY: Done

crates/rome_js_analyze/src/analyzers/use_template.rs
```diff
@@ -0,0 +1,280 @@
...
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsBinaryExpression>;
+    type State = Vec<JsAnyExpression>;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let binary_expr = ctx.query();
+        // Avoiding recursive diagnostics.
+        if binary_expr.parent::<JsBinaryExpression>().is_some() {
```
leops, line 62: Doesn't this mean the rule would not trigger for `1 + "2" - 3` ?

IWANABETHATGUY: Would trigger on `1 + "2" - 3`, but would not trigger on `"2" - 3`

IWANABETHATGUY: I think I misunderstood your advice in https://github.com/rome/tools/pull/2803#discussion_r913922567

leops: What I had in mind was to return `None` if the parent node is a binary expression that would emit a signal (`UseTemplate::run` returns `Some`), but skipping the match for all parent binary expressions will result in false negatives

IWANABETHATGUY: But how could I know if parent emit a `Some` signal by `UseTemplate::run` ?

IWANABETHATGUY: This method is nice, but it increases the times to execute `run` method, although it is still `O(n)`. 
But `run` will execute `2 * n` times compare to `n` times before. We may want to optimize this until we found this is a performance bottleneck in production usage.

IWANABETHATGUY: Done

leops (approved):

IWANABETHATGUY:
@ematipico , I guess this pull request is ready to merge, I have addressed all issues. Would you mind having a look?

ematipico (changes requested):
A couple of things before merging:
1. could you rebase your fork? we fixed the CI failures and it should be fine now;
2. could you add invalid case where strings have spaces?
	  
  An example:
	  ```js
	  0;
	  foo() + " bar"; // before
	  `${foo()} bar`; // after
	  ```

IWANABETHATGUY:
> 

I have rebased the main branch and added an extra test case.

ematipico (approved):

IWANABETHATGUY:
@xunilrj would you mind helping review this pr?

crates/rome_js_analyze/src/utils.rs
```diff
@@ -1,10 +1,30 @@
...
-
-pub fn is_boolean_literal(expr: JsAnyExpression) -> bool {
-    matches!(
-        expr,
-        JsAnyExpression::JsAnyLiteralExpression(
-            rome_js_syntax::JsAnyLiteralExpression::JsBooleanLiteralExpression(_)
-        )
-    )
+#[derive(Debug, PartialEq)]
+pub enum MyError {
```
xunilrj, line 2: Can we have a better name?

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -0,0 +1,286 @@
...
+/// ## Example
+/// - from: `1 + 2 + 3 + (1 * 2)`
+/// - to: `[1, 2, 3, (1 * 2)]`
+fn collect_binary_add_expression(node: &JsBinaryExpression) -> Option<Vec<JsAnyExpression>> {
+    let mut result = vec![];
+    match node.left().ok()? {
+        JsAnyExpression::JsBinaryExpression(left)
+            if matches!(left.operator().ok()?, JsBinaryOperator::Plus) =>
+        {
+            result.extend(collect_binary_add_expression(&left)?);
```
xunilrj, line 269: I think you can ```result.append(...)``` here, which will be faster.
SAme thing below.

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -0,0 +1,286 @@
...
+                        [],
+                    )),
+                );
+                reduced_exprs.push(chunk_element);
+            }
+            JsAnyExpression::JsTemplate(template) => {
+                reduced_exprs.extend(flatten_template_element_list(template.elements())?);
+            }
+            _ => {
+                // drop the trivia of original expression make the generated `JsTemplate` a little nicer, if we don't do this
```
xunilrj, line 145: Can we drop trivia if they are comments?

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -0,0 +1,286 @@
...
+        rome_js_syntax::JsAnyExpression::JsBinaryExpression(binary) => {
+            if is_un_necessary_string_concat_expression(&binary) == Some(true) {
+                return Some(true);
+            }
+        }
+        rome_js_syntax::JsAnyExpression::JsTemplate(_) => return Some(true),
+        rome_js_syntax::JsAnyExpression::JsAnyLiteralExpression(
+            rome_js_syntax::JsAnyLiteralExpression::JsStringLiteralExpression(string_literal),
+        ) => {
+            if interpret_escaped_string(&string_literal.text())
```
xunilrj, line 246: How about moving all this to a function and calling it here with a clearer name?

```rust
if has_newline_or_tick(...) {
    ...
}
```

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -0,0 +1,286 @@
...
+            applicability: Applicability::MaybeIncorrect,
+            message: markup! { "Use a "<Emphasis>"TemplateLiteral"</Emphasis>"." }.to_owned(),
+            root: next_root,
+        })
+    }
+}
+/// Extract main logic of `UseTemplate::run`
+/// Aiming to avoid recursive diagnostics.
+fn emits_signal(expr: &JsBinaryExpression) -> Option<Vec<JsAnyExpression>> {
+    is_un_necessary_string_concat_expression(expr).and_then(|result| {
```
xunilrj, line 104: You can exchange the ```and_then``` by the ```?``` operator.

crates/rome_js_analyze/src/analyzers/js/use_template.rs
IWANABETHATGUY: I think we should keep the comments.

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -0,0 +1,286 @@
...
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsBinaryExpression>;
+    type State = Vec<JsAnyExpression>;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let binary_expr = ctx.query();
+        // Avoiding recursive diagnostics.
+        if binary_expr
```
xunilrj, line 62: I think this call needs to be cheaper. For example;

This is what I mean. We are analyzing the red ```BinaryExpr```. Then we call ```emit_signal``` on our parent, the green node. This recursively processes the red node again.

image

If this call returns ```Some(...)``` we throw everything away and do everything again when the lint rule calls the green node, which probably already happened, given that it is higher in the tree.

In deeper expressions, this will grow exponentially. One way to check this is to print something on the ```run``` and something in the ```is_un_necessary_string_concat_expression``` and other recursive calls and use something like https://github.com/nnethercote/counts to see if their count spread exponentially as the expression grows.

Can´t we just check if the parent is a ```BinaryExpr``` and assume that it was dealt with?

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -0,0 +1,286 @@
...
+                        ret.push(element);
+                    }
+                }
+            }
+        }
+    }
+    Some(ret)
+}
+
+fn is_un_necessary_string_concat_expression(node: &JsBinaryExpression) -> Option<bool> {
```
xunilrj, line 210: I think ```is_unnecessary_string_concat_expression``` is fine.
Also, some comments explaining when is unnecessary would be fine.

crates/rome_js_analyze/src/utils.rs
```diff
@@ -1,10 +1,30 @@
...
+                Some('\\') => Ok('\\'),
+                Some(c) => Err(MyError::InvalidEscapedChar(c)),
+            },
+            c => Ok(c),
+        })
+    }
+}
+
+/// unescape string  
+pub fn interpret_escaped_string(s: &str) -> Result<String, MyError> {
```
xunilrj, line 28: I would name this as the comments. Easier to understand what is doing.

xunilrj:
> @xunilrj would you mind helping review this pr?

Done. I would also test this with a very big binary expression, because we already had problems with recursive algorithms.
I will search later, I think the Typescript test suite has one file to test this.

IWANABETHATGUY:
> > @xunilrj would you mind helping review this pr?
> 
> Done. I would also test this with a very big binary expression, because we already had problems with recursive algorithms. I will search later, I think the Typescript test suite has one file to test this.

Thanks, your review mesage is very helpful, I will have a look later on.

IWANABETHATGUY:
!bench_analyzer

crates/rome_js_analyze/src/analyzers/js/use_template.rs
IWANABETHATGUY: Done, this should be a linear algorithm `is_unnecessary_string_concat_expression` should be executed `2 * N` times, 
`N` is the number of `BinaryExpression`. 
I will paste the output later.

IWANABETHATGUY:
!bench_analyzer

ematipico (approved):

PR #2805, opened 2022-07-01
chore: fix broken rust documentation

ematipico:
## Summary

This PR fixes all the warnings emitted by the command `cargo doc` inside the crate `rome_formatter`

## Test Plan

No warnings when `cargo doc` is run

### DIFF
--- a/crates/rome_rowan/src/cursor/trivia.rs
+++ b/crates/rome_rowan/src/cursor/trivia.rs
@@ -52,7 +52,7 @@ impl SyntaxTrivia {
     }
 
     /// Get the number of TriviaPiece inside this trivia
-    fn len(&self) -> usize {
+    pub(crate) fn len(&self) -> usize {
         self.green_trivia().len()
     }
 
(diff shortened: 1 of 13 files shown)


crates/rome_rowan/src/cursor/trivia.rs
```diff
@@ -52,7 +52,7 @@ impl SyntaxTrivia {
     }
 
     /// Get the number of TriviaPiece inside this trivia
-    fn len(&self) -> usize {
+    pub(crate) fn len(&self) -> usize {
```
leops, line 55: Why does this visibility need to change ?

ematipico:
> We probably should add `#![deny(rustdoc::broken_intra_doc_links)]` to the root of all the crates, and run `cargo doc` at least in the check workflows to ensure we don't re-introduce this kind of issues in the future

I plan to add the workflow in the next iteration. I will crate another PR

crates/rome_rowan/src/cursor/trivia.rs
ematipico: Reverted

PR #2806, opened 2022-07-01
ci: add documentation check on CI

ematipico:


xunilrj:
Do you think makes sense to upload the documentation together with the playground?

xunilrj (approved):

ematipico:
> Do you think makes sense to upload the documentation together with the playground?

Yes, although we should figure out first where to upload the documentation. The playground uses a mix of s3 and cloudflare. And unfortunately AWS is not correctly set up for everyone yet.

Issue #2743, opened 2022-06-20
☂️ Rome classic lint rules, with scoping


ematipico:
# Note

***This rules should be implemented once https://github.com/rome/tools/issues/2488 is closed***.

This is because in order to correctly implement the rule, there's need to have access to the scope.

# If you want to contribute
1. make sure you understand the requirements of the rule you would like to implement;
3. comment on this very issue telling saying which rule you would like to implement;
4. create a new task issue and link to this very umbrella issue (a member of the team will eventually assign it to you);
5. open a PR;

> **Note**: please make sure to comment the issue saying which rule you'd like to implement, this would allow better coordination between contributors. Failing to do so would penalize yourselves against contributors that had followed these simple guildelines

> **Note**: follow the naming convention guidelines: https://github.com/rome/tools/blob/archived-js/CONTRIBUTING.md#naming-patterns

# List of rules that might require scoping resolution

- [x] #2903
- [x] `noUnusedVariables`: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noUnusedVariables.md
- [x] #3292
- [x] #3267
- [x] #3204
- [x] #3313
- [x] ~`noRedundantShouldComponentUpdate`: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react`/noRedundantShouldComponentUpdate.md~ Deprecated
- [x] #3248
- [x] #3254
- [x] #3272
- [x] #3233
- [x] #3278
- [x] ~`useInterfaces`: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/ts/useInterfaces.md 
	This rule was removed from Rome classic. We should consider it with care~ too pedantic, not needed
- [x] `noArguments`: #2807
- [x] https://github.com/rome/tools/issues/2848
- [x] #2878
- [x] `noImportAssign`: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noImportAssign.md
- [x] #2837
- [x] #3341
- [x] #3338
- [x] #2897
- [x] #2836
- [x] #3281
- [x] ~`noFindDOMNode`: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noFindDOMNode.md~

> **Note**: this rule requires access to the filename, which is not implemented yet
- [x] ~[useDefaultExportBasename](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/useDefaultExportBasename.md)~
[...]

PR #2807, opened 2022-07-01
feature(rome_js_analyzer): semantic rules and "no arguments" lint rule

xunilrj:
This PR creates the foundation and implements the first semantic/scoping rule of https://github.com/rome/tools/issues/2743. Undoubtedly the ```SemanticModel``` will need improvements. We can do them as we see need from the rules.

Relevant changes

#### Query return node directly

```rust
 let node = ctx.query(); //instead of  let Ast(node) = ctx.query();
```

#### Semantic model is accessed through RuleContext

but the method ```model()``` exist only if the ```Rule``` asked for the semantic model.

```rust
let reference = ctx.query();

let name = reference.syntax().text_trimmed();
if name == "arguments" {
    let model = ctx.model();
    ...
}
```

#### New query type

Rules can ask to run with the semantic model with

```
type Query = Semantic<JsReferenceIdentifier>;
```

#### rome_js_analyze run the in two phases

Today everything is run serially. But the opportunity to run in MT is right there!
I did a quick test and the only major problem I saw was ```Rc``` inside rome_rowan.

```rust
pub fn analyze<F, B>(...) {
    // Phase Syntax
    // Phase Semantic
}

```

#### performance

Speed is almost the same without the semantic model. I saw a minor increase (1/2ms) in the ```input.js``` of the ```rome_cli``` folder.

As soon as we land more semantic lint rules, we can focus on optimization.

#### playground

I added two small improvements to the playground.
First, colors!

image

Second, timing:

this time in on the debug build. I am not good to advertise features! :smile:
image


crates/rome_analyze/src/registry.rs
```diff
@@ -132,6 +167,7 @@ type RuleExecutor<L, B> = for<'a> fn(
     FileId,
     &'a LanguageRoot<L>,
     &'a QueryMatch<L>,
+    &'a ServiceBag,
     &'a mut dyn FnMut(&dyn AnalyzerSignal<L>) -> ControlFlow<B>,
```
ematipico, line 171: I wonder if at this point it's better to have a `struct` here, to better to document each parameter

crates/rome_analyze/src/rule.rs
```diff
@@ -106,6 +106,10 @@ pub trait Rule: RuleMeta {
     /// analyzer
     type Signals: IntoIterator<Item = Self::State>;
 
+    fn phase() -> Phases {
```
ematipico, line 109: This seems to return the `syntax_phase`. I think we should change its name accordingly

crates/rome_analyze/src/services.rs
```diff
@@ -0,0 +1,50 @@
...
+        self.0.as_ref()
+    }
+}
+
+impl TryFrom<ServiceBag> for () {
+    type Error = CannotCreateServicesError;
+    fn try_from(_: ServiceBag) -> Result<Self, Self::Error> {
+        Ok(())
+    }
+}
```
ematipico, line 50: This implementation is a bit weird, it doesn't return any error. Is it on purpose?

crates/rome_analyze/src/services.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use crate::CannotCreateServicesError;
+
+#[derive(Default)]
+pub struct ServiceBagData {
+    services: HashMap<TypeId, Box<dyn Any>>,
+}
+
+impl ServiceBagData {
+    pub fn insert<T: 'static + Clone>(&mut self, svc: T) {
```
ematipico, line 16: `svc` => `service`?

ematipico, line 16: `insert_service`?

crates/rome_analyze/src/services.rs
```diff
@@ -0,0 +1,50 @@
...
+        let id = TypeId::of::<T>();
+        self.services.insert(id, Box::new(svc));
+    }
+
+    pub fn get<T: 'static + Clone>(&self) -> Option<T> {
```
ematipico, line 21: `get_service`?

crates/rome_analyze/src/syntax.rs
```diff
@@ -14,7 +14,7 @@ pub struct SyntaxVisitor<L: Language, F> {
     skip_subtree: Option<SyntaxNode<L>>,
 }
 
-impl<L: Language, F> SyntaxVisitor<L, F>
+impl<L: Language, F: Send> SyntaxVisitor<L, F>
```
ematipico, line 17: I assume we are preparing this to be MT-able? If so, wouldn't it be better to delay the change to when we actually want to try this?

crates/rome_js_analyze/src/lib.rs
```diff
@@ -39,23 +42,43 @@ where
     F: FnMut(&dyn AnalyzerSignal<JsLanguage>) -> ControlFlow<B>,
     B: 'static,
 {
+    let mut registry = build_registry(&filter, callback);
+
     let mut analyzer = Analyzer::<JsLanguage, B>::empty();
 
+    // Phase Syntax
```
ematipico, line 49:
```suggestion
    // Syntax phase
```

crates/rome_js_analyze/tests/specs/noArguments.js.snap
```diff
@@ -0,0 +1,38 @@
...
+
+```
+warning[noArguments]: Use the rest parameters instead of arguments.
+  ┌─ noArguments.js:7:17
+  │
+7 │     console.log(arguments);
+  │                 ---------
+
+
+```
```
ematipico, line 36: Is this diagnostic correct? The snippet was under "valid" section

ematipico:
@xunilrj there's an empty file that is not supposed to be there. It's called `services.rs`

crates/rome_js_analyze/src/lib.rs
```diff
@@ -39,23 +42,43 @@ where
...
+    let mut services = ServiceBagData::default();
+    services.insert(model);
+    let services = ServiceBag::new(services);
+
+    analyzer.run(VisitorContext {
+        file_id,
+        root: root.clone(),
+        range: filter.range,
+        match_query: Box::new(|file_id, root, query_match| {
+            registry.match_query(Phases::Semantic, file_id, root, query_match, &services)
```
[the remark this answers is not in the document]
xunilrj: It is possible, but a little bit inconvenient. At least for now. We can reuse this as an optimization later.

The major issue is that ServiceBag is read-only, to be MT safe. And for that, I chose ```Arc```. The only way to get the data out of it is using https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap, which can fail. So I would need to deal with it failing for now.

crates/rome_js_analyze/tests/specs/noArguments.js.snap
xunilrj: I have to delete this file.

website/playground/src/lib.rs
```diff
@@ -69,6 +75,9 @@ struct ErrorOutput(Vec<u8>);
 
 impl io::Write for ErrorOutput {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        // let str = std::str::from_utf8(buf).unwrap();
+        // let html = ansi_to_html::convert_escaped(str).unwrap();
+        // self.0.write(html.as_str().as_bytes())
         self.0.write(buf)
```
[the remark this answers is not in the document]
xunilrj: No. Good catch.

crates/rome_analyze/src/context.rs
```diff
@@ -1,26 +1,48 @@
...
 where
     R: Rule,
 {
-    pub fn new(query_result: &'a <R as Rule>::Query, root: &'a RuleRoot<R>) -> Self {
-        Self { query_result, root }
+    pub fn new(
+        query_result: &'a <<R as Rule>::Query as Queryable>::Output,
+        root: &'a RuleRoot<R>,
+        services: ServiceBag,
+    ) -> Result<Self, <<<R as Rule>::Query as Queryable>::Services as TryFrom<ServiceBag>>::Error>
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_analyze/src/context.rs
```diff
@@ -1,26 +1,48 @@
-use crate::{registry::RuleRoot, Rule};
+use crate::{registry::RuleRoot, services::ServiceBag, Queryable, Rule};
+use std::ops::Deref;
 
 pub struct RuleContext<'a, R>
 where
     R: ?Sized + Rule,
 {
-    query_result: &'a <R as Rule>::Query,
+    query_result: &'a <<R as Rule>::Query as Queryable>::Output,
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_analyze/src/query.rs
```diff
@@ -1,11 +1,20 @@
...
 
-use crate::registry::NodeLanguage;
+use crate::{
+    registry::{NodeLanguage, Phase},
+    services::ServiceBag,
+};
+
+pub enum CannotCreateServicesError {
+    MissingServices(&'static [&'static str]),
+}
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_analyze/src/registry.rs
```diff
@@ -6,23 +6,38 @@ use rome_rowan::{AstNode, Language, RawSyntaxKind, SyntaxKind};
...
+    Semantic = 1,
+}
+
+pub trait Phase {
+    fn phase() -> Phases {
+        Phases::Syntax
+    }
+}
+
+impl Phase for () {}
```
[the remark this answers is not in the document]
xunilrj: For rules that do not demand any services. I improved the documentation.

crates/rome_analyze/src/registry.rs
```diff
@@ -6,23 +6,38 @@ use rome_rowan::{AstNode, Language, RawSyntaxKind, SyntaxKind};
 use crate::{
     context::RuleContext,
     query::{QueryKey, QueryMatch, Queryable},
+    services::ServiceBag,
     signals::{AnalyzerSignal, RuleSignal},
     ControlFlow, Rule,
 };
 
+#[repr(usize)]
+pub enum Phases {
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_analyze/src/registry.rs
```diff
@@ -6,23 +6,38 @@ use rome_rowan::{AstNode, Language, RawSyntaxKind, SyntaxKind};
...
+}
+
+impl Phase for () {}
+
 /// The rule registry holds type-erased instances of all active analysis rules
 pub struct RuleRegistry<'a, L: Language, B> {
     /// Holds a collection of rules for each [SyntaxKind] node type that has
-    /// lint rules associated with it
-    ast_rules: Vec<SyntaxKindRules<L, B>>,
+    /// lint rules associated with it for each phase
```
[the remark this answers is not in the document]
xunilrj: I actually improved the documentation of the RuleRegistry. I think makes more sense now.

crates/rome_js_analyze/src/lib.rs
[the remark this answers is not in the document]
ematipico: I understand now, thank you for the explanation!

crates/rome_analyze/src/registry.rs
```diff
@@ -34,6 +49,8 @@ impl<'a, L: Language, B> RuleRegistry<'a, L, B> {
         R: Rule + 'static,
         R::Query: Queryable<Language = L> + Clone,
     {
+        let phase = R::phase() as usize;
```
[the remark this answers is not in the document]
xunilrj: But "syntax phase" is just one of the possible phases. 

You insert the rule. We use the trait system to see what that Rule demands, in terms of services, and see when we can run it. 

This is what this ```phase()```method do.

crates/rome_analyze/src/registry.rs
```diff
@@ -6,23 +6,38 @@ use rome_rowan::{AstNode, Language, RawSyntaxKind, SyntaxKind};
 use crate::{
     context::RuleContext,
     query::{QueryKey, QueryMatch, Queryable},
+    services::ServiceBag,
     signals::{AnalyzerSignal, RuleSignal},
     ControlFlow, Rule,
 };
 
+#[repr(usize)]
```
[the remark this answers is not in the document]
xunilrj: Just because I will index an array with this value.

crates/rome_analyze/src/registry.rs
xunilrj: Yes. I think we would need a small refactor in the future to make things faster and simpler.
But I would postpone this a little bit yet.

crates/rome_analyze/src/rule.rs
xunilrj: I improved the documenration.
This method actually can return any phase. Depending on what the rule is demanding, in terms of services.

crates/rome_analyze/src/services.rs
xunilrj: Yes. If you don´t need any services, this cannot fail.
The same thing does not happen in other phases. For example, if I am going to start the semantic phase, the ```ServiceBag``` must have a ```SemanticModel```, and thus can fail.

crates/rome_analyze/src/services.rs
xunilrj: Done.

xunilrj: Done.

crates/rome_analyze/src/services.rs
xunilrj: Done.

crates/rome_js_analyze/src/lib.rs
xunilrj: Done.

crates/rome_js_analyze/src/semantic.rs
```diff
@@ -0,0 +1,59 @@
+use rome_analyze::{
```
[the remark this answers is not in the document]
xunilrj: Done.

crates/rome_analyze/src/registry.rs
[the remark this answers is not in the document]
ematipico: I suppose that the absence of documentation didn't help me to understand this.

crates/rome_analyze/src/rule.rs
ematipico: Thanks, this helps a lot

crates/rome_analyze/src/registry.rs
```diff
@@ -63,13 +92,27 @@ impl<'a, L: Language, B> RuleRegistry<'a, L, B> {
...
     /// Returns an iterator over the name and documentation of all active rules
     /// in this instance of the registry
-    pub fn metadata(self) -> impl Iterator<Item = (&'static str, &'static str)> {
+    pub fn metadata(&self) -> Vec<(&'static str, &'static str)> {
         let mut unique = HashSet::new();
-        self.ast_rules
-            .into_iter()
-            .flat_map(|node| node.rules)
+
+        let [phase0, phase1] = &self.phases;
```
ematipico, line 98: `phase0` => `syntax_phase`
`phase1` => `semantic_phase`

Better names, I suppose

ematipico (approved):

crates/rome_analyze/src/syntax.rs
xunilrj: Yes. Fixed.

crates/rome_js_analyze/tests/specs/noArguments.cjs
```diff
@@ -0,0 +1,8 @@
+function f() {
```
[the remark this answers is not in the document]
xunilrj: There is actually: I want to declare a variable named ```arguments```.

crates/rome_analyze/src/registry.rs
xunilrj: Done.

PR #2809, opened 2022-07-03
feat(rome_js_formatter): poorly breakable member or call chain

denbezrukov:
## Summary

This PR partly implements prettier functionality  `isPoorlyBreakableMemberOrCallChain` for `assignment_like` nodes. These nodes have `BreakAfterOperator` layout.

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L329

Also this PR fixes `assignment_like` content wrapping. See more details in comments.

## Test Plan

Add new test cases.

Current:
**File Based Average Prettier Similarity**: 77.64%  
**Line Based Average Prettier Similarity**: 72.78%  

Main:
**Line Based Average Prettier Similarity**: 72.59%

### DIFF
--- a/crates/rome_js_formatter/src/js/lists/variable_declarator_list.rs
+++ b/crates/rome_js_formatter/src/js/lists/variable_declarator_list.rs
@@ -16,7 +16,7 @@ impl FormatRule<JsVariableDeclaratorList> for FormatJsVariableDeclaratorList {
 
         let mut declarators = node.elements().enumerate().map(|(index, element)| {
             format_with(move |f| {
-                write!(f, [group_elements(&element.node().format())])?;
+                write!(f, [&element.node().format()])?;
 
                 match element.trailing_separator()? {
                     None => {
--- a/crates/rome_js_formatter/src/utils/assignment_like.rs
+++ b/crates/rome_js_formatter/src/utils/assignment_like.rs
@@ -1,19 +1,21 @@
 use crate::prelude::*;
+use crate::utils::member_chain::is_member_call_chain;
 use crate::utils::object::write_member_name;
 use crate::utils::JsAnyBinaryLikeExpression;
 use rome_formatter::{format_args, write, VecBuffer};
 use rome_js_syntax::{
-    JsAnyAssignmentPattern, JsAnyBindingPattern, JsAnyClassMemberName, JsAnyExpression,
-    JsAnyFunctionBody, JsAnyObjectAssignmentPatternMember, JsAnyObjectBindingPatternMember,
-    JsAnyObjectMemberName, JsAssignmentExpression, JsInitializerClause, JsLiteralMemberName,
-    JsObjectAssignmentPattern, JsObjectAssignmentPatternProperty, JsObjectBindingPattern,
-    JsPropertyClassMember, JsPropertyClassMemberFields, JsPropertyObjectMember, JsSyntaxKind,
-    JsVariableDeclarator, TsAnyVariableAnnotation, TsIdentifierBinding,
-    TsPropertySignatureClassMember, TsPropertySignatureClassMemberFields, TsType,
-    TsTypeAliasDeclaration,
+    JsAnyAssignmentPattern, JsAnyBindingPattern, JsAnyCallArgument, JsAnyClassMemberName,
+    JsAnyExpression, JsAnyFunctionBody, JsAnyObjectAssignmentPatternMember,
+    JsAnyObjectBindingPatternMember, JsAnyObjectMemberName, JsAnyTemplateElement,
+    JsAssignmentExpression, JsInitializerClause, JsLiteralMemberName, JsObjectAssignmentPattern,
+    JsObjectAssignmentPatternProperty, JsObjectBindingPattern, JsPropertyClassMember,
+    JsPropertyClassMemberFields, JsPropertyObjectMember, JsSyntaxKind, JsVariableDeclarator,
+    TsAnyVariableAnnotation, TsIdentifierBinding, TsPropertySignatureClassMember,
+    TsPropertySignatureClassMemberFields, TsType, TsTypeAliasDeclaration, TsTypeArguments,
 };
 use rome_js_syntax::{JsAnyLiteralExpression, JsSyntaxNode};
 use rome_rowan::{declare_node_union, AstNode, SyntaxResult};
+use std::iter;
 
 declare_node_union! {
     pub(crate) JsAnyAssignmentLike =
@@ -514,7 +516,11 @@ impl JsAnyAssignmentLike {
 
     /// Returns the layout variant for an assignment like depending on right expression and left part length
     /// [Prettier applies]: https://github.com/prettier/prettier/blob/main/src/language-js/print/assignment.js
-    fn layout(&self, is_left_short: bool) -> FormatResult<AssignmentLikeLayout> {
+    fn layout(
+        &self,
+        is_left_short: bool,
+        f: &mut Formatter<JsFormatContext>,
+    ) -> FormatResult<AssignmentLikeLayout> {
         if self.has_only_left_hand_side() {
             return Ok(AssignmentLikeLayout::OnlyLeft);
         }
@@ -523,6 +529,12 @@ impl JsAnyAssignmentLike {
 
         if let Some(layout) = self.chain_formatting_layout()? {
             return Ok(layout);
+        }
+
+        if let Some(JsAnyExpression::JsCallExpression(call_expression)) = &right {
+            if call_expression.callee()?.syntax().text() == "require" {
+                return Ok(AssignmentLikeLayout::NeverBreakAfterOperator);
+            }
         }
 
         if self.should_break_left_hand_side()? {
@@ -537,8 +549,21 @@ impl JsAnyAssignmentLike {
             return Ok(AssignmentLikeLayout::NeverBreakAfterOperator);
         }
 
+        // Before checking `BreakAfterOperator` layout, we need to unwrap the right expression from `JsUnaryExpression` or `TsNonNullAssertionExpression`
+        // [Prettier applies]: https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L199-L211
+        // Example:
+        //  !"123" -> "123"
+        //  void "123" -> "123"
+        //  !!"string"! -> "string"
+        let right_expression = iter::successors(right, |expression| match expression {
+            JsAnyExpression::JsUnaryExpression(unary) => unary.argument().ok(),
+            JsAnyExpression::TsNonNullAssertionExpression(assertion) => assertion.expression().ok(),
+            _ => None,
+        })
+        .last();
+
         if matches!(
-            right,
+            right_expression,
             Some(JsAnyExpression::JsAnyLiteralExpression(
                 JsAnyLiteralExpression::JsStringLiteralExpression(_)
             )),
@@ -546,9 +571,31 @@ impl JsAnyAssignmentLike {
             return Ok(AssignmentLikeLayout::BreakAfterOperator);
         }
 
-        if self.should_never_break_after_operator()? {
+        let is_poorly_breakable = match right_expression {
+            Some(expression) => is_poorly_breakable_member_or_call_chain(expression, f)?,
+            None => false,
+        };
+
+        if is_poorly_breakable {
+            return Ok(AssignmentLikeLayout::BreakAfterOperator);
+        }
+
+        if matches!(
+            self.right()?.as_expression(),
+            Some(
+                JsAnyExpression::JsClassExpression(_)
+                    | JsAnyExpression::JsTemplate(_)
+                    | JsAnyExpression::JsAnyLiteralExpression(
+                        JsAnyLiteralExpression::JsBooleanLiteralExpression(_),
+                    )
+                    | JsAnyExpression::JsAnyLiteralExpression(
+                        JsAnyLiteralExpression::JsNumberLiteralExpression(_)
+                    )
+            )
+        ) {
             return Ok(AssignmentLikeLayout::NeverBreakAfterOperator);
         }
+
         Ok(AssignmentLikeLayout::Fluid)
     }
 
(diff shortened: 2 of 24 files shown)


crates/rome_js_formatter/src/js/lists/variable_declarator_list.rs
```diff
@@ -16,7 +16,7 @@ impl FormatRule<JsVariableDeclaratorList> for FormatJsVariableDeclaratorList {
 
         let mut declarators = node.elements().enumerate().map(|(index, element)| {
             format_with(move |f| {
-                write!(f, [group_elements(&element.node().format())])?;
+                write!(f, [&element.node().format()])?;
```
denbezrukov, line 19: This `group_elements` is unnecessary. Because `assignment_like` wraps it's content with a group.

denbezrukov, line 19: This `group_elements` is unnecessary. Because `assignment_like` wraps it's content with a group.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -525,6 +531,12 @@ impl JsAnyAssignmentLike {
             return Ok(layout);
         }
 
+        if let Some(JsAnyExpression::JsCallExpression(call_expression)) = &right {
```
denbezrukov, line 534: Need to extract this condition from function. Because prettier checks this before `break_left_hand_side`.

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L138

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -537,18 +549,39 @@ impl JsAnyAssignmentLike {
             return Ok(AssignmentLikeLayout::NeverBreakAfterOperator);
         }
 
+        let node = iter::successors(right, |node| match node {
```
denbezrukov, line 552: https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L199-L211

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +916,192 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+                )
+            })
+            .unwrap_or(false);
+
+        let will_break = {
+            // let mut null_buffer = f.inspect_null();
+            // let mut buffer = null_buffer.inspect_will_break();
+            // write!(buffer, [type_arguments.format()]);
+            // buffer.will_break()
+            false
```
denbezrukov, line 1094: I kept this comments because I'm not sure how to implement it. Is it right? 
This implementation leads to double formatting node.

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L454

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +916,192 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+            JsAnyExpression::JsUnaryExpression(unary_expression) => {
+                is_signed_numeric_literal(unary_expression)?
+            }
+            JsAnyExpression::JsAnyLiteralExpression(literal) => match literal {
+                JsAnyLiteralExpression::JsRegexLiteralExpression(regex) => {
+                    let token = regex.value_token()?;
+                    let text_trimmed = token.text_trimmed();
+
+                    // SAFETY: a valid regex literal must have a end slash
+                    let end_slash_pos = text_trimmed.rfind('/').unwrap();
```
denbezrukov, line 1011: Do we have more convenient way to extract a pattern from `JsRegexLiteralExpression`?

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +916,196 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+    let threshold = f.context().line_width().value() / 4;
+
+    let mut expression = expression;
+    let mut is_deep = false;
+
+    while let Some(node) = expression.take() {
+        match node {
+            JsAnyExpression::JsCallExpression(call_expression) => {
+                let is_member_chain = {
+                    // let mut null_buffer = f.inspect_null();
```
denbezrukov, line 934: I kept this comments because I'm not sure how to implement it. Is it right? 
This implementation leads to double formatting node.

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L346

ematipico: This is, unfortunately, a limitation of the architecture as it is now (and we should probably look into it to make easier to handle).

Whenever you plan to format a formatter,  in this case the `call_expression.format`, inside **any buffer**, that has be memoized, otherwise we trigger the double formatting.

crates/rome_js_formatter/src/utils/member_chain/groups.rs
```diff
@@ -20,7 +20,7 @@ pub(crate) struct Groups {
 
     /// This is a threshold of when we should start breaking the groups
     ///
-    /// By default, it's 2, meaning that we start breaking after the second group.
+    /// By default, it's 1, meaning that we start breaking after the first group.
```
denbezrukov, line 23: Do I understand correctly that prettier uses only one array for all groups and Rome uses two structs (`HeadGroup` and `Groups`)?  `cutoff` value could be `2` and `3` depends on `should_merge`.
It seems that it always should be `1`. Because Rome uses two structs and then `should_merge` is true it mutates `Groups` vec.

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/member-chain.js#L301-L304

ematipico: Yes, I went for two structs because it was difficult to reason with prettier's logic. 

Essentially there's an head and there's a tail. The head can be become bigger is `shouldMerge` is `true` and in that case the first group of the tail joins the head.  

Prettier relies often on slicing the whole `group` based on `shouldMerge`, which is not ideal for us in terms of memory. In Rome we went for a different solution. That's where `HeadGroup::expand_group` comes into play.

denbezrukov: I have an assumption that `cutoff` must always be 1 for the behavior to be the same as Prettier.

Prettier:
[headGroup, secondGroup, thirdGroup, fourthGroup]

shouldMerge = false -> cutoff = 2 -> thirdGroup
shouldMerge = true-> cutoff = 3 -> fourthGroup

Rome:
`headGroup` and `groups`: [secondGroup, thirdGroup, fourthGroup, fifthGroup, sixthGroup]

shouldMerge  = false -> cutoff = 2; -> `fourthGroup`
shouldMerge = true 
We move `secondGroup` to `headGroup`
 `groups`: [thirdGroup, fourthGroup, fifthGroup, sixthGroup] cutoff = 3 -> `sixthGroup`

ematipico: Yeah you probably right, we should lower the number of cutoff, to match prettier's cutoff

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +923,193 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+            expression => {
+                return Ok(is_deep
+                    && matches!(
+                        expression,
+                        JsAnyExpression::JsIdentifierExpression(_)
+                            | JsAnyExpression::JsThisExpression(_)
+                    ));
+            }
+        }
+    }
```
ematipico, line 991: I wonder if we can actually invert the logic on this one.

This is a loop that might not have any ending unless we hit the last branch at the end, and each time we go a level deeper, we do a lot of checks.

So my idea is first to do the deep check, while we do it, we save a reference of a call expression in a vector. 

And then, we do check on call expressions (member chain, call arguments, etc.).

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +923,193 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+
+    Ok(false)
+}
+
+/// [Prettier applies]: https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L374
+fn is_lone_short_argument(argument: JsAnyCallArgument, threshold: u16) -> SyntaxResult<bool> {
+    if argument.syntax().has_leading_comments() || argument.syntax().has_trailing_comments() {
```
ematipico, line 998: you can use `has_comments_direct`

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: No  we don't, at the moment. You're welcome to add any function if you want. The file `expr_ext.rs` inside `rome_js_syntax` we `impl` all the nodes with utility functions

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +923,193 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+
+                    len <= threshold as usize
+                }
+                JsAnyLiteralExpression::JsStringLiteralExpression(string) => {
+                    string.value_token()?.text_trimmed().len() <= threshold as usize
+                }
+                _ => true,
+            },
+            JsAnyExpression::JsTemplate(template) => {
+                let elements = template.elements();
```
ematipico, line 1028: `JsTemplate` has a function called `template_range` which gives you a `Option<TextRange>`. From there you can check if the range is bigger of the `threshold`.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +923,193 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+            _ => false,
+        };
+        Ok(is_lone_short_argument)
+    } else {
+        Ok(false)
+    }
+}
+
+/// [Prettier applies]: https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/utils/index.js#L231
+fn is_signed_numeric_literal(expression: JsUnaryExpression) -> SyntaxResult<bool> {
```
ematipico, line 1051: Documentation for this function

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +923,193 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+    );
+
+    let is_numeric_literal = matches!(
+        argument,
+        JsAnyExpression::JsAnyLiteralExpression(JsAnyLiteralExpression::JsNumberLiteralExpression(
+            _
+        ))
+    );
+    let has_comments =
+        argument.syntax().has_leading_comments() || argument.syntax().has_trailing_comments();
```
ematipico, line 1066: `has_comments_direct`

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: This would require to memoize the actual arguments. I would remove anything around `will_break` and think about it later. This would require to basically memoize the whole right hand side of the assignment.

ematipico:
!bench_formatter

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Added `pub fn pattern(&self) -> SyntaxResult<String>` for `JsRegexLiteralExpression`.
I wasn't sure about `String` but I found `name` function for `JsLiteralMemberName`.

https://github.com/rome/tools/blob/main/crates/rome_js_syntax/src/expr_ext.rs#L69-L79

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Done.
Also added TODO and link to prettier code.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -537,18 +550,45 @@ impl JsAnyAssignmentLike {
             return Ok(AssignmentLikeLayout::NeverBreakAfterOperator);
         }
 
+        // Before checking `BreakAfterOperator` layout, we need to unwrap the node expression
+        // [Prettier applies]: https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L199-L211
+        // Example:
+        //  !"123" -> "123"
+        //  void "123" -> "123"
+        //  !!"string"! -> "string"
+        let node = iter::successors(right, |node| match node {
```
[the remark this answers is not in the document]
denbezrukov: Renamed.

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: done

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Actually Prettier has a more complex condition. It checks that there is no any expression in template.
For Rome it means we can check that we have only one element  and  this element is  `JsTemplateChunkElement` .

- https://github.com/prettier/prettier/blob/a043ac0d733c4d53f980aa73807a63fc914f23bd/src/language-js/print/assignment.js#L402-L405

crates/rome_js_formatter/src/utils/assignment_like.rs
denbezrukov: Done. Also moved it to `expr_ext.rs`

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -906,3 +923,193 @@ impl Format<JsFormatContext> for JsAnyAssignmentLike {
...
+fn is_poorly_breakable_member_or_call_chain(
+    expression: Option<JsAnyExpression>,
+    f: &mut Formatter<JsFormatContext>,
+) -> SyntaxResult<bool> {
+    let threshold = f.context().line_width().value() / 4;
+
+    let mut expression = expression;
+
+    // We want to keep track of whether we are in a call chain or member chain
+    // If we only have one level, we don't have to do anything
```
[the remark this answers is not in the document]
denbezrukov: I tried to paraphrase =)

crates/rome_js_formatter/src/utils/assignment_like.rs
ematipico: Yes, member call chain is actually a better naming. Spot on!

PR #2811, opened 2022-07-04
feat(rome_js_analyze): useTsExpectError

### LINKED ISSUE #2753: useTsExpectErrorPrefer

IWANABETHATGUY:
## Summary
1. Resolved #2753 

## Test Plan
1. Checking the new snapshot result


crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
```diff
@@ -0,0 +1,218 @@
...
+};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_syntax::{JsAnyStatement, JsSyntaxToken};
+use rome_rowan::{AstNode, AstNodeExt, TriviaPiece};
+
+use crate::JsRuleAction;
+
+declare_rule! {
+    /// Enforce use `@ts-expect-error` instead of `@ts-ignore` to get notified when suppression is no longer necessary.
```
ematipico, line 12:
```suggestion
    /// Enforces the use of `@ts-expect-error` instead of `@ts-ignore` to get notified when a suppression is no longer necessary.
```

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// ```
+    pub(crate) UseTsExpectError = "useTsExpectError"
+}
+
+impl Rule for UseTsExpectError {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsAnyStatement>;
+    /// The `@ts-ignore` index of trailing trivia, and the corresponding replaced_string using `@ts-expect-error`
+    type State = Vec<(usize, String)>;
```
ematipico, line 69: Why not an `HashMap`?

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
```diff
@@ -0,0 +1,218 @@
...
+        n.syntax().first_token().and_then(|token| {
+            let mut ts_ignore_index_vec = vec![];
+            token
+                .leading_trivia()
+                .pieces()
+                .enumerate()
+                .for_each(|(i, item)| match item.kind() {
+                    rome_js_syntax::TriviaPieceKind::SingleLineComment => {
+                        let original = item.text();
+                        // I use this `strip_prefix` just because of clippy not happy
```
ematipico, line 83: This comment should actually say *why* clippy is not happy, so we can understand why `strip_prefix` is used instead. What did you try to use before?

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
ematipico: Creating a new string for each `@ts-ignore` is really expensive. Wouldn't be better to return, instead, all the information needed to compute the string? 

`@ts-expect-error` should be a `const` of this file instead.

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
```diff
@@ -0,0 +1,218 @@
...
+                                    "{}@ts-expect-error{}",
+                                    &original[0..original.len() - finalized.len()],
+                                    stripped
+                                ),
+                            ));
+                        }
+                    }
+                    rome_js_syntax::TriviaPieceKind::MultiLineComment => {
+                        let original = item.text();
+                        //
```
ematipico, line 107: remove?

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
```diff
@@ -0,0 +1,218 @@
...
+        let Ast(n) = ctx.query();
+        n.syntax().first_token().and_then(|token| {
+            let mut ts_ignore_index_vec = vec![];
+            token
+                .leading_trivia()
+                .pieces()
+                .enumerate()
+                .for_each(|(i, item)| match item.kind() {
+                    rome_js_syntax::TriviaPieceKind::SingleLineComment => {
+                        let original = item.text();
```
ematipico, line 82: Why using `.replace()` is not enough?

ematipico (changes requested):
Would it be possible to have some description in the PR that explains the implementation please? There's a lot going on and the description is poor compared to the actual logic. Without context, the review is really hard to do

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
IWANABETHATGUY: > Why not an `HashMap`?

Because the iteration of comment is always ordered, whether checking phase or replacing phase.

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
IWANABETHATGUY: I write code like this before:
```rs
if let Some(stripped) = original.starts_with("/**") {
    &original[3..]
} else if original.starts_with("//") || original.starts_with("/*") {
    &original[2..]
} else {
    original
}
.trim_start();
```

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
IWANABETHATGUY: > 

Do you mean deferring the string replacement until the user triggers the code action?

ematipico: > > 
> 
> Do you mean deferring the string replacement until the user triggers the code action?

Yes exactly

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
IWANABETHATGUY: What do you mean by `replace`,  `Trivia` seems not to be replaceable

IWANABETHATGUY: > OK, probably I need to wait for a better description of the PR. It's really hard to understand what's going on

Yeah, I am writing the doc

crates/rome_js_analyze/src/analyzers/use_ts_expect_error.rs
```diff
@@ -0,0 +1,233 @@
...
+const TS_EXPECT_ERROR_SUPPRESSION: &str = "@ts-expect-error";
+
+impl Rule for UseTsExpectError {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsAnyStatement>;
+    /// Because `ts-ignore` could be used in multiple places, we need to track the state of each `@ts-ignore`.
+    /// The first element of tuple is the index of the original leading trivia pieces that need to replace,
+    /// the second element of tuple is a [Vector] of the range of `@ts-ignore` in the original trivia piece text,
+    /// we use a [Vector] to store the range because the `@ts-ignore` could exist multiple time in single trivia piece.
```
ematipico, line 75:
```suggestion
    /// `ts-ignore` could be used in multiple places inside a comment, we need to track the state of each `@ts-ignore`.
    /// In order to track these comments, we shape the state with a [Vec] of tuples, where a tuple contains:
    /// - the first element is the position  of the original leading trivia that needs to be replaced;
    /// - the second element of the tuple is a [vector](Vec) of the [ranges](Range) of the string `"@ts-ignore"` found in the original trivia piece text;
    /// 
    /// We use a [Vector] because the string `"@ts-ignore"` can be present multiple times inside a comment.
```

1. "track the state" is a bit generic, what is "a state"? Maybe we can say "track the position"

ematipico (changes requested):
Sorry to reject the PR, but the whole algorithm should be reviewed.

The suppression comment works only on single line comments: https://www.typescriptlang.org/play?ssl=7&ssc=19&pln=1&pc=1#code/PQKhCgQAgAQFwM4FoCmAPADigxnVAnfAe30mHABsU4oBDALigCMiiraA7KAXigGZw4YMFiJUmHHhSESlas0Ys2KTj35A

ematipico:
@IWANABETHATGUY are there any progress on this PR?

IWANABETHATGUY:
> @IWANABETHATGUY are there any progress on this PR? Do you need any help?

Nope, I will address issues this week.

IWANABETHATGUY:
!bench_analyzer

crates/rome_js_analyze/tests/specs/js/noLabelVar.js.snap
```diff
@@ -0,0 +1,55 @@
+---
```
IWANABETHATGUY, line 1: move this test file to the right place.

crates/rome_js_analyze/src/analyzers/ts/use_ts_expect_error.rs
```diff
@@ -0,0 +1,236 @@
...
+            .pieces()
+            .enumerate()
+            .for_each(|(i, piece)| {
+                let text = piece.text();
+                if state.0 == i {
+                    let mut normalized_string = String::with_capacity(text.len());
+                    // Replace `@ts-ignore` to `@ts-expect-error`
+                    normalized_string.push_str(&text[0..state.1]);
+                    normalized_string.push_str(TS_EXPECT_ERROR_SUPPRESSION);
+                    normalized_string.push_str(&text[(state.1 + 10)..]);
```
ematipico, line 201: What's 10 in this case? Can we store it in a `const` so we know what's the meaning of it?

ematipico (approved):
@leops @xunilrj could you please add your review?

crates/rome_js_analyze/src/analyzers/ts/use_ts_expect_error.rs
```diff
@@ -0,0 +1,236 @@
...
+                        normalized_string.len() as u32,
+                    ));
+                    next_leading_trivia_string.push_str(&normalized_string);
+                } else {
+                    next_leading_trivia.push(TriviaPiece::new(piece.kind(), text.len() as u32));
+                    next_leading_trivia_string.push_str(text);
+                }
+            });
+        // Copy trailing trivia
+        first_token.trailing_trivia().pieces().for_each(|item| {
```
leops, line 213: Nit: use a `for-in` loop here instead of `for_each`

crates/rome_js_analyze/src/analyzers/ts/use_ts_expect_error.rs
```diff
@@ -0,0 +1,236 @@
...
+        let mut next_leading_trivia_string = String::new();
+        let mut next_trailing_trivia_string = String::new();
+        first_token
+            .leading_trivia()
+            .pieces()
+            .enumerate()
+            .for_each(|(i, piece)| {
+                let text = piece.text();
+                if state.0 == i {
+                    let mut normalized_string = String::with_capacity(text.len());
```
leops, line 197: This text could be pushed directly into the `next_leading_trivia_string` buffer without having to allocate an intermediate `String` (or even better, only allocate a single string holding the data for `next_leading_trivia_string`, `first_token.text_trimmed()` and `next_trailing_trivia_string`)

leops (commented):
I think the main issue I have with this rule is that it only matches on `JsAnyStatement`, so it would catch these cases:

```typescript
const
    // @ts-ignore
    value1: boolean = 3;
    
const value2 = 
    // @ts-ignore
    true + 4;
```

Unfortunately in order to detect these, the rule would have to inspect the leading trivia of all the tokens in the file, and the performance cost of doing that seems prohibitively high at the moment

ematipico:
> I think the main issue I have with this rule is that it only matches on `JsAnyStatement`, so it would not catch these cases:
> 
> ```ts
> const
>     // @ts-ignore
>     value1: boolean = 3;
>     
> const value2 = 
>     // @ts-ignore
>     true + 4;
> ```
> 
> Unfortunately in order to detect these, the rule would have to inspect the leading trivia of all the tokens in the file, and the performance cost of doing that seems prohibitively high at the moment

Given this premise, should we put on hold this PR? I doesn't make much sense to release an half working rule.

IWANABETHATGUY:
> 

Unfortunately, in order to detect these, the rule would have to inspect the leading trivia of all the tokens in the file, and the performance cost of doing that seems prohibitively high at the moment. As @leops said, the main trade-off is the performance,.
If you agree to the trade-off for the performance, I would like to fix these edge issues.

leops:
> Unfortunately, in order to detect these, the rule would have to inspect the leading trivia of all the tokens in the file, and the performance cost of doing that seems prohibitively high at the moment. As @leops said, the main trade-off is the performance,. If you agree to the trade-off for the performance, I would like to fix these edge issues.

At the moment the linter lacks two capabilities requires to implement this, the first and most important is the ability for lint rules (and visitors) to inspect syntax tokens (currently lint rules can only query AST nodes), while a secondary one would be to allow rules to be enabled or disabled on a per-language basis: currently the existing JSX and TS rules are enabled all the time which isn't really a problem since they only query node for their language and thus never match in incompatible files, but if this rule needs to inspect all tokens we would want it to be disabled on non-TS files.

ematipico:
I think we should have this PR on hold until we figure out a better way to implement it. 

> we would want it to be disabled on non-TS files.

Unfortunately this is not always true. It's possible to avail of the TS compiler even in JS files, there's option called `checkJs` in the `tsconfig.json` file that enables the compiler to check JS files. Regardless, it seems we don't have enough tools to implement this rule at the moment.

ematipico:
@leops should we just close the PR until we have the tools to be able to implement this rule?

leops:
> @leops should we just close the PR until we have the tools to be able to implement this rule?

Yes I think in general it could make sense to close PRs that are on hold, it's probable that by the time we resume working on these it'll be easier to start again from scratch than trying to rebase the previous attempt and handle all the possible merge conflicts

PR #2812, opened 2022-07-04
fix(rome_js_formatter): regression of call arguments

### LINKED ISSUE #2720: Fix regression on call arguments with `will_break`
The PR #2711 introduced a known regression, where there are cases where we have to inspect the arguments and check if they break. This will affect the layout of the formatting

Regressions:

```js
expression(
	/* block comment */
);

runtimeAgent.getProperties(
    objectId,
    false, // ownProperties
    false, // accessorPropertiesOnly
    false, // generatePreview
    (error, properties, internalProperties) => {
      return 1
    },
);

```

ematipico:
## Summary

This PR fixed #2720 

The new logic checks if any arguments breaks and if so, it does a custom printing.

Doing so, it triggered a regression of the new code that we didn't see before:

This snippet:

```js
test.expect(t => {
	t.true(a);
}, false /* inline comment */ );
```

Would be formatted

```js
test.expect ((t) => {
	t.true(a);
}, false /* inline comment */ );
```

Note the space added before the first `(`. This regression is caused by a bug inside the function `write_space_between_comment_and_token`, which triggers a false positive condition and it adds a space before the token.

I will have to look into it later. In order to avoid to commit wrong code, I changed the snippet that was causing the false positive.

I will open an issue to track this bug

## Test Plan

`cargo test` and checked that the regression is not fixed

PR 

**File Based Average Prettier Similarity**: 77.40%  
**Line Based Average Prettier Similarity**: 72.66%  

`main`
**Line Based Average Prettier Similarity**: 72.61%

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/call_arguments.rs
+++ b/crates/rome_js_formatter/src/js/expressions/call_arguments.rs
@@ -179,6 +179,33 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
                 )
             });
 
+            let mut null_buffer = f.inspect_null();
+            let mut arguments_break = separated.iter().map(|element| {
+                let mut will_break_buffer = null_buffer.inspect_will_break();
+                write!(will_break_buffer, [element]).ok();
+                will_break_buffer.will_break()
+            });
+            let any_breaks = arguments_break.any(|will_break| will_break);
+            let an_argument_breaks = arguments_break.enumerate().any(|(index, will_break)| {
+                return if should_group_first_argument && index > 0
+                    || (should_group_last_argument && index < args.len() - 1)
+                {
+                    will_break
+                } else if index > 0 && index < args.len() - 1 {
+                    will_break
+                } else {
+                    false
+                };
+            });
+
+            if an_argument_breaks {
+                return write!(f, [all_arguments_expanded]);
+            }
+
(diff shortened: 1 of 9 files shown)


crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -179,6 +179,31 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
                 )
             });
 
+            let mut null_buffer = f.inspect_null();
+            let mut arguments_break = separated.iter().map(|element| {
+                let mut will_break_buffer = null_buffer.inspect_will_break();
+                write!(will_break_buffer, [element]).ok();
+                will_break_buffer.will_break()
+            });
+            let any_breaks = arguments_break.any(|will_break| will_break);
```
leops, line 188: If I remember correctly calling `.any` will mutate the `arguments_break` iterator in place, so all the iterators items consumed to initialize `any_breaks` will not be processed by the next call to `.any` to initialize `an_argument_breaks`, is this really what's intended here ?

ematipico: That's really a good catch, and I was wondering why I could not replicate a reformat issue that I was encountering some time ago.

leops (approved):

PR #2819, opened 2022-07-05
chore: better package.json and readme

ematipico:
## Summary

This PR does the following:
- creates a new `README.md` file inside the package that we ship on `npm`; the text is taken from the original read me that have at root of the repository, and added few instructions of how to use it; 
- made the  `README.md` at the root of the repository a symbolic link to the new file created;
- changed the URL to the images by using the full URL to github, doing so we will be able to see the image also on the npm page of the library;
- updated the `package.json` with some useful metadata;


npm/rome/package.json
```diff
@@ -5,11 +5,22 @@
...
     ],
+    "keywords": [
+        "JavaScript",
+        "TypeScript",
+        "format",
+        "lint",
+        "toolchain"
+    ],
     "engines": {
         "node": ">=14.18.0"
```
xunilrj, line 25: Is there a specific reason for this version? Or we could go below it?

xunilrj (approved):

npm/rome/package.json
ematipico: Initially it was there because of things like `require("node:child_process")` (`node` prefix), but we removed it. I suppose we can lower it to be `14.*`

PR #2821, opened 2022-07-05
feat(rome_control_flow): introduce explicit exception control flow

leops:
## Summary

This PR replaces the (temporary) initial implementation of try-catch-finally statements in the control flow analysis (based on the concept of multiple entry points) with a more complete modelling solution for the exception control flow. The `BasicBlock` struct that make up the control flow graph now also contain a list of "exception handlers" and "cleanup handlers": these handlers point to a set of blocks to be executed in order when an exception is thrown (either with an explicit throw instruction or as a possible side-effect of a statement) or when the function returns respectively.

For JavaScript, the exceptions handlers of a block are all the `finally` clauses found going upward in the syntax tree until a `catch` clause is found (the block corresponding to this clause is the last handler in the list), while the cleanup handlers are made of all the `finally` clauses going up to the root level of the function (and ignoring all catch clauses)

The last change I've made is introducing a new `finally_fallthrough` flag on `Jump` instructions. This flag is used specifically to mark the implicit jump instruction inserted by the control flow generator at the end of a `finally` block towards the next statement: when the finally clause gets executed as an exception or cleanup handler instead of going through the normal control flow the target block of this jump gets modified to the next handler in the chain if it exists, or as a return instruction otherwise.

## Test Plan

I've added additional test cases for `JsTryStatement` and `JsTryFinallyStatement` under the snapshots for the `noDeadCode` rule

### DIFF
--- a/crates/rome_control_flow/src/builder.rs
+++ b/crates/rome_control_flow/src/builder.rs
@@ -1,6 +1,9 @@
 use rome_rowan::{Language, SyntaxElement};
 
-use crate::{BasicBlock, ControlFlowGraph, Instruction, InstructionKind};
+use crate::{
+    BasicBlock, ControlFlowGraph, ExceptionHandler, ExceptionHandlerKind, Instruction,
+    InstructionKind,
+};
 
 /// Identifier for a block in a [ControlFlowGraph]
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
@@ -20,6 +23,7 @@ impl BlockId {
 /// should be added
 pub struct FunctionBuilder<L: Language> {
     result: ControlFlowGraph<L>,
+    exception_target: Vec<ExceptionHandler>,
     block_cursor: BlockId,
 }
 
@@ -27,6 +31,7 @@ impl<L: Language> Default for FunctionBuilder<L> {
     fn default() -> Self {
         Self {
             result: ControlFlowGraph::new(),
+            exception_target: Vec::new(),
             block_cursor: BlockId { index: 0 },
         }
     }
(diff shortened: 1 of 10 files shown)


crates/rome_control_flow/src/builder.rs
```diff
@@ -87,7 +125,19 @@ impl<L: Language> FunctionBuilder<L> {
     }
 
     pub fn append_jump(&mut self, conditional: bool, block: BlockId) -> InstructionBuilder<L> {
-        self.append_instruction(InstructionKind::Jump { conditional, block })
+        self.append_instruction(InstructionKind::Jump {
+            conditional,
+            block,
+            finally_fallthrough: false,
```
IWANABETHATGUY, line 131: I'm not sure if it's proper to put a language-specific property into "rome_control_flow", which seems to be a language-irrelevant crate

crates/rome_control_flow/src/lib.rs
```diff
@@ -79,12 +80,32 @@ pub enum InstructionKind {
...
     /// either unconditionally if `conditional` is set to `false`, or after
     /// evaluating the associated syntax node otherwise
-    Jump { conditional: bool, block: BlockId },
+    Jump {
+        conditional: bool,
+        block: BlockId,
+        /// Set to `true` for the terminating jump instruction out of a
+        /// `finally` clause, the target block can be reinterpreted to the next
+        /// exception handler instead if the control flow is currently unwinding
+        finally_fallthrough: bool,
```
ematipico, line 89: We could call it `fall_)through` (more generic) and use `finally` **as an example**

crates/rome_control_flow/src/builder.rs
```diff
@@ -87,7 +125,19 @@ impl<L: Language> FunctionBuilder<L> {
...
+        })
+    }
+
+    pub fn append_finally_fallthrough(&mut self, block: BlockId) -> InstructionBuilder<L> {
```
ematipico, line 135: `append_fall_through`?

crates/rome_control_flow/src/builder.rs
leops: `finally` clauses are a concept found in a few other languages than JavaScript (like Java, C#, Python, ...), languages that don't use it simply won't emit instructions with this flag (I specifically made it possible to create an instruction with this flag through the specialized `append_finally_fallthrough` method so it doesn't pollute all the other call sites of `append_jump` and languages that don't use it can transparently ignore it)

IWANABETHATGUY: > 

Thanks for explaining

PR #2822, opened 2022-07-05
feat(rome_analyze): implement per-rule tracking of suppression comments

leops:
## Summary

This PR refactors the handling of suppression comments in the analyzer to introduce support for suppressing individual rules. This is implemented at the analyzer traversal level by deferring the emission of analyzer signals (diagnostics and code actions) until enough tokens in the tree have been visited to ensure the line associated with the signal does not have any associated suppression comments for the rule.

I took the opportunity to refactor some parts of the analyzer infrastructure and decouple individual elements: the most important change is that while the visitors and the rule registry retain the capability to interrupt the analysis process by returning `ControlFlow::Break` there is no longer an associated type `B` associated to the break value (it's instead forced to `()`). This generic type previously had to be threaded through the entire codebase as it could be returned by the user-provided `emit_signal` callback, but the execution of the `emit_signal` is now deferred until later in the traversal process and handled by the analyzer itself, removing the need to "contaminate" the whole codebase with this generic.

Incidentally this change modifies the way the `RuleRegistry` stores rule metadata (since it now needs to be able to recognize and validate rule names in suppression comments) which fixes the documentation generation for the `noDeadCode` rule

## Test Plan

The test for suppression comments has been extended, but this PR should otherwise not change the way the analyzer is run (in terms of observable behavior) and so all existing tests should continue to pass

### DIFF
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -14,13 +14,16 @@ mod visitor;
 
 pub use crate::categories::{ActionCategory, RuleCategories, RuleCategory};
 pub use crate::query::{Ast, CannotCreateServicesError, QueryKey, QueryMatch, Queryable};
-pub use crate::registry::{LanguageRoot, Phase, Phases, RuleRegistry};
+pub use crate::registry::{LanguageRoot, RuleKey, RuleRegistry};
+pub use crate::registry::{Phase, Phases};
 pub use crate::rule::{Rule, RuleAction, RuleDiagnostic, RuleMeta};
 pub use crate::services::{ServiceBag, ServiceBagData};
 pub use crate::signals::{AnalyzerAction, AnalyzerSignal};
 pub use crate::syntax::SyntaxVisitor;
 pub use crate::visitor::{NodeVisitor, Visitor, VisitorContext};
-use rome_rowan::{AstNode, Language, TextRange};
+use rome_diagnostics::file::FileId;
+use rome_rowan::{AstNode, Language, SyntaxKind, SyntaxNode, TextRange, WalkEvent};
+use rustc_hash::FxHashMap;
 
 /// The analyzer is the main entry point into the `rome_analyze` infrastructure.
 /// Its role is to run a collection of [Visitor]s over a syntax tree, with each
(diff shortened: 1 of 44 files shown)


crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +15,294 @@ mod syntax;
...
 /// diagnostics, code actions or both
-pub struct Analyzer<L, B> {
-    visitors: Vec<Box<dyn Visitor<B, Language = L>>>,
+pub struct Analyzer<'a, L: Language, Matcher, Break> {
+    /// List of visitors being run by this instance of the analyzer
+    visitors: Vec<Box<dyn Visitor<Language = L> + 'a>>,
+    /// Executor for the query matches emitted by the visitors
+    query_matcher: Matcher,
+    /// Queue for pending analyzer signals
+    signal_queue: BinaryHeap<SignalEntry<L>>,
```
xunilrj, line 45: Why do you need a ```BinaryHeap```?
Are we generating signals out of order, or it is here just for safety?

leops: Yes rules can emit signals out of order, the default position for a signal is the primary label of its diagnostic and this may be anywhere in the file (for instance the `noDeadCode` rule generates diagnostics covering previous ranges of the file once the traversal reaches the end of a function)

crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +16,297 @@ mod syntax;
...
+    }
+
+    fn handle_comment(&mut self, text: &str, range: TextRange) {
+        let mut suppress_all = false;
+        let mut suppressions = Vec::new();
+
+        for (category, rule) in (self.parse_suppression_comment)(text) {
+            if category != "lint" {
+                continue;
+            }
```
ematipico, line 240: Is this temporary? It seems odd to have a category hardcoded like this inside a generic analyzer

crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +16,297 @@ mod syntax;
...
+
+                visitor.visit(&node_event, ctx);
             }
         }
 
         None
     }
+
+    /// Process the text for a single token, parsing suppression comments and
+    /// flushing pending query matches
```
ematipico, line 151: Would you mind add more details to the method? Mostly how the algorithm works. Or if you think it's not needed, at least in the description. It would be great to give a detailed description of what algorithm is trying to do and how.

ematipico (commented):
I am not in a position yet to completely review the algorithm, probably because I am tired. I left few questions (trivial). I will review it next week with a pair of fresh eyes and a fresh brain

crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +16,311 @@ mod syntax;
...
+
+    /// Check a piece of source text (token or trivia) for line breaks and
+    /// increment the line index accordingly, extending the range of the
+    /// current suppression as required
+    fn bump_line_index(&mut self, text: &str, range: TextRange) {
+        let mut did_match = false;
+        for (index, _) in text.match_indices('\n') {
+            if let Some(suppression) = self.line_suppressions.last_mut() {
+                if suppression.line_index == self.line_index {
+                    let index = TextSize::try_from(index).expect("integer overflow");
```
ematipico, line 288:
```suggestion
                    let index = TextSize::try_from(index).expect("integer overflow while converting a suppression line to `TextSize`");
```

crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +16,311 @@ mod syntax;
...
+/// This function receives the text content of a comment and returns a list of
+/// lint suppressions as an optional lint rule (if the lint rule is `None` the
+/// comment is interpreted as suppressing all lints)
+///
+/// # Examples
+///
+/// - `// rome_ignore format` -> `vec![]`
+/// - `// rome_ignore line` -> `vec![None]`
+/// - `// rome_ignore lint(js/useWhile)` -> `vec![Some("js/useWhile")]`
+/// - `// rome_ignore lint(js/useWhile) lint(js/noDeadCode)` -> `vec![Some("js/useWhile"), Some("js/noDeadCode")]`
```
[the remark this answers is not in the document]
leops: That was a typo, it was supposed to be `rome-ignore lint`

crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +16,311 @@ mod syntax;
...
+        if !suppress_all && suppressions.is_empty() {
+            return;
+        }
+
+        // Suppression comments apply to the next line
+        let line_index = self.line_index + 1;
+
+        // Ensure proper ordering of the `line_suppressions` buffer
+        if let Some(suppression) = self.line_suppressions.last() {
+            assert!(suppression.line_index < line_index);
```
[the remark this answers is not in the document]
leops: This assert was actually there as a placeholder to catch a special case in the algorithm that wasn't implemented yet (merging multiple contiguous suppression comments), it works now

crates/rome_analyze/src/lib.rs
```diff
@@ -13,52 +16,311 @@ mod syntax;
...
+            } else {
+                suppressions.clear();
+                suppress_all = true;
+                // If this if a "suppress all lints" comment, no need to
+                // parse anything else
+                break;
+            }
+        }
+
+        if !suppress_all && suppressions.is_empty() {
```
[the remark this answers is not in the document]
leops: There's a call to `suppressions.clear()` during the parsing of comments when `suppress_all` is set to true (to avoid storing a list of suppressed rules when they're all disabled anyway), so `suppressions.is_empty()` is always true if `!suppress_all` is false, and the function would always return if the `||` operator was used in that case

PR #2834, opened 2022-07-06
fix performance.measure returning undefined on firefox

xunilrj:
Work around the nuisance that Firefox returns undefined for ```performance.measure(...)```.
See: https://stackoverflow.com/questions/69666408/how-to-use-performance-measure-in-firefox

image


IWANABETHATGUY:
> 

The Ci still failed after I rebase the main branch.

PR #2842, opened 2022-07-07
perf(rome_analyze): optimize code actions diff generation

leops:
## Summary

The generation of code action diffs shows up a lot when profiling the analyzer, so I've optimized it a bit by making it skip over irrelevant subtrees. For `find_diff_range` this implies comparing the green nodes between the old and new version of the tree, so I've made the `key` method public on `SyntaxNode` / `SyntaxToken` / `SyntaxElement`

## Test Plan

The snapshot tests for the analyzer still pass, as well as the unit tests for `find_diff_range` although I had to change the expected values since the function now compares the green nodes instead of the text content

### DIFF
--- a/crates/rome_analyze/src/signals.rs
+++ b/crates/rome_analyze/src/signals.rs
@@ -1,11 +1,14 @@
-use std::marker::PhantomData;
+use std::{collections::VecDeque, marker::PhantomData};
 
 use rome_console::MarkupBuf;
 use rome_diagnostics::{
     file::{FileId, FileSpan},
     Applicability, CodeSuggestion, Diagnostic, SuggestionChange, SuggestionStyle,
 };
-use rome_rowan::{AstNode, Direction, Language, SyntaxNode, TextRange};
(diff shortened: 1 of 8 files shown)


ematipico:
!bench_analyzer

ematipico:
The actions died: https://github.com/rome/tools/runs/7249709125?check_suite_focus=true

IWANABETHATGUY:
I was wondering any reason why we don't use a string-based diff algorithm?

leops:
> I was wondering any reason why we don't use a string-based diff algorithm?

This is technically all about converting a CST-based diff into a text-based diff, but for the initial version the diff only consist of a single range-replace operation since it was easier to just implement that first. We do have the capability to eventually represent more complex string diffs with the `rome_text_edit` crate, and the switch could happen sooner rather than later if the theoretically simpler algorithm we have now continues to require some tweaking to run correctly

ematipico (approved):

PR #2847, opened 2022-07-08
chore(xtask): bench analyzer on better files

ematipico:
## Summary

The analyzer at the moment works on files that are minified. While this might be useful for performance, this creates a lot of false positives and it stresses the reporting in terms of diagnostics. 

Minified files generate code that usually goes against the lint rules. This PR creates two new files where we list non-minified files to analyze. This is a better real world example.

I couldn't find big files to lint, if you have more, please post them here

## Test Plan

Run the bench command and it should work

### DIFF
--- a/xtask/bench/src/analyzer-libs-js.txt
+++ b/xtask/bench/src/analyzer-libs-js.txt
@@ -1,0 +1,3 @@
+https://github.com/prettier/prettier/blob/main/src/language-js/utils/index.js
+https://raw.githubusercontent.com/jquery/jquery/main/src/css.js
+https://raw.githubusercontent.com/fastify/fastify/main/lib/configValidator.js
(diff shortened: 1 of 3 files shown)


ematipico:
!bench_analyzer

leops (commented):
One additional change I would add to this is changing the `AnalysisFilter` for the benchmarks to only enable `RuleCategories::SYNTAX | RuleCategories::LINT` as categories. Currently the benchmarks are being run with the `RuleCategory::Action` rules turned on, but these are not supposed to be enabled for whole-file analysis as they generate context specific refactor opportunities

ematipico:
!bench_analyzer

xunilrj (approved):

ematipico:
!bench_analyzer

ematipico:
!bench_analyzer

ematipico:
!bench_analyzer

ematipico:
!bench_analyzer

leops:
Unfortunately until this PR is merged to main the analyzer benchmark command will remain very slow (since the workflow runs the benchmarks on main too and they haven't been fixed there yet)

PR #2849, opened 2022-07-08
feat(rome_js_analyze): noCatchAssign

### LINKED ISSUE #2848: 📎 noCatchAssign
### Description

https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noCatchAssign.md

IWANABETHATGUY:
## Summary
1. Resolve #2848 

## Test Plan
1. cargo test

### DIFF
--- a/website/src/docs/lint/rules/index.md
+++ b/website/src/docs/lint/rules/index.md
@@ -28,11 +28,11 @@ Disallow the use of <code>arguments</code>
 Disallows using an async function as a Promise executor.
 </div>
 <div class="rule">
-<h3 data-toc-exclude id="noCommentText">
-	<a href="/docs/lint/rules/noCommentText">noCommentText</a>
-	<a class="header-anchor" href="#noCommentText"></a>
+<h3 data-toc-exclude id="noCatchAssign">
+	<a href="/docs/lint/rules/noCatchAssign">noCatchAssign</a>
+	<a class="header-anchor" href="#noCatchAssign"></a>
 </h3>
-Prevent comments from being inserted as text nodes
+Disallow reassigning exceptions in catch clauses
 </div>
 <div class="rule">
 <h3 data-toc-exclude id="noCompareNegZero">
--- a/website/src/docs/lint/rules/noCatchAssign.md
+++ b/website/src/docs/lint/rules/noCatchAssign.md
@@ -20,8 +20,8 @@ try {
 }
 ```
 
-{% raw %}<pre class="language-text"><code class="language-text"><span style="color: Orange;">warning</span><span style="color: Orange;">[</span><span style="color: Orange;">noCatchAssign</span><span style="color: Orange;">]</span><em>: </em><em> Do not </em><em><em>reassign catch parameters.</em></em><em></em>
-  <span style="color: rgb(38, 148, 255);">┌</span><span style="color: rgb(38, 148, 255);">─</span> noCatchAssign.js:5:3
+{% raw %}<pre class="language-text"><code class="language-text"><span style="color: Orange;">warning</span><span style="color: Orange;">[</span><span style="color: Orange;">js/noCatchAssign</span><span style="color: Orange;">]</span><em>: </em><em> Do not </em><em><em>reassign catch parameters.</em></em><em></em>
+  <span style="color: rgb(38, 148, 255);">┌</span><span style="color: rgb(38, 148, 255);">─</span> js/noCatchAssign.js:5:3
   <span style="color: rgb(38, 148, 255);">│</span>
 <span style="color: rgb(38, 148, 255);">3</span> <span style="color: rgb(38, 148, 255);">│</span> } catch (e) {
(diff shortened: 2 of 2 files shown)


crates/rome_js_analyze/src/semantic_analyzers/no_catch_assign.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        (!invalid_assign.is_empty()).then(|| invalid_assign)
+    }
+
+    fn diagnostic(ctx: &RuleContext<Self>, state: &Self::State) -> Option<RuleDiagnostic> {
+        let node = ctx.query();
+
+        let mut diagnostic = RuleDiagnostic::warning(
+            node.syntax().text_trimmed_range(),
```
leops, line 81: Instead of using the span of the catch clause as the primary range for the diagnostic, it would be more readable to emit a new diagnostic signal for each assignment and set the primary label to that node, with a secondary label on the declaration of the catch clause

IWANABETHATGUY: But that means I need check every `JsAssignment`, that would be pretty inefficient.

IWANABETHATGUY: I agree with you.

IWANABETHATGUY: Done

xunilrj:
> This pull request could be reviewed until this #2854 pull request has been merged.

Merged.

crates/rome_js_analyze/src/semantic_analyzers/no_catch_assign.rs
leops: Sorry I should have responded to that earlier, I didn't mean to suggest to rewrite the query or run function entirely (indeed processing all assignments instead of only matching on catch clauses seems less efficient), instead my proposal was just to change the return type of the `run` method to `Vec<JsSyntaxNode>` instead of `Option<Vec<JsSyntaxNode>>`

xunilrj:
Should we underline the whole assignment?

image

xunilrj:
Do you think makes sense to improve the semantic model to support:
1 - all_references from JsAnyBindings;
2 - methods to test if a reference is "read" or "write".

In this case this lint would be:

```rust
let references = decl.binding().ok()?.all_references(model);

for assignment in references.filter(|r| r.is_write())
{
      // generate warning
}
```

IWANABETHATGUY:
> 

I just keep the same behavior as rome-classic https://github.com/rome/tools/blob/archived-js/internal/compiler/lint/rules/js/noCatchAssign.test.md, @leops what do you think?

IWANABETHATGUY:
> Do you think makes sense to improve the semantic model to support: 1 - all_references from JsAnyBindings; 2 - methods to test if a reference is "read" or "write".
> 
> In this case this lint would be:
> 
> ```rust
> let references = decl.binding().ok()?.all_references(model);
> 
> for assignment in references.filter(|r| r.is_write())
> {
>       // generate warning
> }
> ```
> 
> Or more succinctly
> 
> ```rust
> for assignment in decl.binding().ok()?.all_writes(model);
> {
>       // generate warning
> }
> ```

That's pretty nice.

IWANABETHATGUY:
> 

image

xunilrj:
> The whole statement could be quite big, how will the diagnostic be printed in this case?

This can be a problem and our diagnostic do not handle this cas well.

leops (approved):
I think having the primary label cover only the identifier is fine, it focuses the attention of the user on the binding being written to and not the expression being assigned to it

xunilrj:
> Do you think makes sense to improve the semantic model to support: 1 - all_references from JsAnyBindings; 2 - methods to test if a reference is "read" or "write".
> 
> In this case this lint would be:
> 
> ```rust
> let references = decl.binding().ok()?.all_references(model);
> 
> for assignment in references.filter(|r| r.is_write())
> {
>       // generate warning
> }
> ```
> 
> Or more succinctly
> 
> ```rust
> for assignment in decl.binding().ok()?.all_writes(model);
> {
>       // generate warning
> }
> ```

See https://github.com/rome/tools/pull/2874

IWANABETHATGUY:
> > Do you think makes sense to improve the semantic model to support: 1 - all_references from JsAnyBindings; 2 - methods to test if a reference is "read" or "write".
> > In this case this lint would be:
> > ```rust
> > let references = decl.binding().ok()?.all_references(model);
> > 
> > for assignment in references.filter(|r| r.is_write())
> > {
> >       // generate warning
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or more succinctly
> > ```rust
> > for assignment in decl.binding().ok()?.all_writes(model);
> > {
> >       // generate warning
> > }
> > ```
> 
> See #2874

Would like to have a try when #2874 is merged.

PR #2850, opened 2022-07-08
feat(rome_analyze): add rule groups

leops:
## Summary

This PR introduces the concept of "rule group" to the analyzer. Groups care simply a collection rules under a given name, and are generated automatically based on the filesystem structure of the lint rules, They serve an organizational purpose, allowing rules to be classified in the documentation, but they also have a more practical purpose as the analyzer allows entire groups to be enabled or disabled using suppression comments in the code or an `AnalysisFilter` object provided to the analyzer (in the future this filter may be generated from the configuration file, allowing rules or groups to be enabled or disabled for the entire workspace).

Since this change adds additional logic to the processing of suppression comments I've based it on top of #2822, in addition to allowing these comments to suppress an entire group I've also added a warning that's now emitted by the analyzer when a suppression comment refers to a rule or group that's not known to the linter

## Test Plan

I've moved all the snapshot tests for the lint rules to the directory corresponding to the group for each rule, other than including the rule group name in the diagnostics the snapshots should not differ in their content. I've also added an additional test for `rome_analyze` that checks the suppression comment parsing logic on the `RawLanguage`, in addition to the existing tests in `rome_js_analyze` doing the same thing for the `JsLanguage`


crates/rome_analyze/src/lib.rs
```diff
@@ -237,14 +245,56 @@ where
...
+                                    "Unknown lint rule "{group}"/"{rule}" in suppression comment"
+                                },
+                            ),
+                            None => Diagnostic::warning(
+                                file_id,
+                                "Linter",
+                                markup! {
+                                    "Unknown lint rule group "{rule}" in suppression comment"
+                                },
+                            ),
```
ematipico, line 291: We should add a suggestion with the list of available groups. Do you plan to have it for another PR?

ematipico (approved):

PR #2851, opened 2022-07-08
feat(rome_cli): read configuration file from CLI

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/2826 and https://github.com/rome/tools/issues/2824

This PR adds support to read the configuration file from the `FileSystem` trait.

The reason why it was implemented there, it's because it allows for a better testing, allowing to programmatically inject configuration files inside the tests, using the `MemoryFileSystem`, while the `OsFileSystem` will used an hardcoded path seek the configuration file and read it.

The configuration file is then merged with the formatter options.

The formatter command now bails if the formatter is disabled.

## Test Plan

Added a couple of test to simulate the loading of a configuration.

I also run the current CLI using the command we have inside the `package.json` of the VSCode extension. I can confirm that no errors were thrown.


leops (commented):
I'm not sure about having the path to the configuration file stored in the filesystem, configuration seems like an information the filesystem shouldn't be aware of (due to testing, I would rather keep the filesystem as small as possible since everything else builds on top of it). Instead of explicitly setting a "configuration path", the tests that use a custom configuration files could just create a file called `rome.json` in the memory filesystem and let the rest of the code load it as it normally would

ematipico:
> I'm not sure about having the path to the configuration file stored in the filesystem, configuration seems like an information the filesystem shouldn't be aware of (due to testing, I would rather keep the filesystem as small as possible since everything else builds on top of it). Instead of explicitly setting a "configuration path", the tests that use a custom configuration files could just create a file called `rome.json` in the memory filesystem and let the rest of the code load it as it normally would

Done, the code should be more simpler, although I had to tweak a bit some tests. But overall it should match your suggestions.

leops (approved):

Issue #2846, opened 2022-07-08
📎 （rome_playground): Support saving  format configuration into local storage


IWANABETHATGUY:
## Description
Currently, the default format of CST or AST is `JSON`. But sometimes `Text` would be more useful, saving user prefer format choice to local storage could avoid developers re configuring playground every time.

## Plan
We want to support saving all configurations into local storage, not only the format of `*ST`

PR #2853, opened 2022-07-09
feat(playground): Sync rome config to local storage

IWANABETHATGUY:
## Summary
1. Resolve https://github.com/rome/tools/issues/2846

## Test Plan
1. Check out if the new feature of the playground is expected.


IWANABETHATGUY:
https://user-images.githubusercontent.com/17974631/178101372-68fdc7ac-b07a-4b02-84d9-91555963164f.mp4

website/playground/src/utils.ts
```diff
@@ -194,3 +217,73 @@ function fromBinary(binary: string) {
...
+	return mergeRomeConfig(config, defaultRomeConfig);
+}
+
+// Change the romeConfig in local storage seems to be meaningless, so we simplify the validation phase.
+// only use default value if some key is empty
+function mergeRomeConfig(
+	config: Partial<RomeConfiguration>,
+	defaultConfig: RomeConfiguration,
+): RomeConfiguration {
+	let interested_keys: Array<keyof RomeConfiguration> = ['treeStyle'];
```
ematipico, line 244: We already have `persistKeys` which has the same type. We should not repeat code. Also, `const` is enough.

ematipico (changes requested):

website/playground/src/utils.ts
```diff
@@ -17,7 +17,9 @@ export function classNames(
 ): string {
 	return classes.filter(Boolean).join(" ");
 }
-export const persistKeys: Array<keyof PlaygroundState> = ["treeStyle"];
+// Excluding `playgroundState.code` reduce the content to save,
+// Only save prop of `PlaygroundState` that we interested, making `localStorage.setItem` which is file IO a bit faster
```
ematipico, line 21: This line should have stayed where we use `localStorage.setItem` and you have to add "faster than..." something. I don't know what though... you have to tell us 😛

IWANABETHATGUY: Done

ematipico (approved):

PR #2860, opened 2022-07-11
perf(rome_analyze): fallback to a simple analysis in noDeadCode for complex functions

leops:
## Summary

This changes the implementation of the `noDeadCode` lint rule to select between two different implementations of the reachability analysis (simple and fine) based on a "complexity score" for the function.

This score is an arbitrary number, roughly related to the cyclomatic complexity of the function: the formula might look similar but the actual value being returned is slightly different as our `ControlFlowGraph` doesn't exactly match the strict definition of a control flow graph of basic blocks, and the algorithm doesn't detect connected components for efficiency.

If that score is determined to be over 20 for a given function the rule will only run the "simple" analysis that calculates reachability for all instructions in the function but doesn't attempt to determine a "terminator instruction" for each path and yields very simple diagnostics instead of the detailed ones returned by the "fine" version.

## Test Plan
I added a new test case (`HighComplexity.js`) declaring a function with an artificially high complexity and ensuring only a basic diagnostic is being emitted (created by the simple analysis instead of the fine algorithm)

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/no_dead_code.rs
+++ b/crates/rome_js_analyze/src/analyzers/no_dead_code.rs
@@ -55,10 +55,10 @@ impl Rule for NoDeadCode {
 
         let cfg = ctx.query();
 
-        if estimate_complexity(cfg) <= COMPLEXITY_THRESHOLD {
-            analyze_fine(cfg, &mut signals)
-        } else {
+        if exceeds_complexity_threshold(cfg) {
             analyze_simple(cfg, &mut signals)
+        } else {
+            analyze_fine(cfg, &mut signals)
         }
 
         signals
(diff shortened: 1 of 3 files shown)


ematipico:
!bench_analyzer

crates/rome_js_analyze/src/analyzers/no_dead_code.rs
```diff
@@ -227,6 +186,195 @@ impl Rule for NoDeadCode {
...
+
+        for inst in &block.instructions {
+            if let InstructionKind::Jump { conditional, .. } = inst.kind {
+                edges += 1;
+
+                if conditional {
+                    conditionals += 1;
+                }
+            }
+        }
```
ematipico, line 211: Isn't it there a way to bail from the loop early? I mean, it seems that we have to calculate **all** the jumps, before making a decision. Can't we make a decision while we loop?

ematipico:
> analyzer/statement.ts     30.39  263.4±10.89ms   137.9 KB/sec    1.00      8.7±0.01ms     4.1 MB/sec

That's the stuff!

xunilrj (approved):

PR #2866, opened 2022-07-12
feat(rome_service, rome_cli): linter via configuration

### LINKED ISSUE #2862: implement the linter configuration, consume it across the application

ematipico:
## Summary

Closes #2862 

This PR introduces a new tool section inside the configuration structure. For now only the field is `enabled`. The rest of the fields will be added later, once we have support for groups for lint rules.

I updated the `TraversalMode`, which now has information about the enabled/disabled tools (formatter and linter for now). These information are passed from the configuration file.

I also refactored a bit the `TraversalMode`, by joining the `Check` and `Fix` in a single one. They contain the same information and their difference is really minimal. A new field called `should_fix` is set to true if `--apply` argument is passed via CLI.

The configuration is now merged inside the `Workspace` trait, which now also have the fields `enabled`. 

## Test Plan

Added new tests that checks that the linter won't analyze the files if in the configuration we set `"enabled": false`

### DIFF
--- a/crates/rome_cli/src/commands/check.rs
+++ b/crates/rome_cli/src/commands/check.rs
@@ -1,20 +1,19 @@
+use crate::commands::format::parse_format_options;
 use crate::{
     traversal::{traverse, TraversalMode},
     CliSession, Termination,
 };
 use rome_diagnostics::MAXIMUM_DISPLAYABLE_DIAGNOSTICS;
+use rome_service::configuration::Configuration;
 use rome_service::load_config;
 use rome_service::load_config::ConfigurationType;
+use rome_service::settings::{LinterSettings, WorkspaceSettings};
+use rome_service::workspace::UpdateSettingsParams;
 
 /// Handler for the "check" command of the Rome CLI
 pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
     let configuration = load_config(&session.app.fs, ConfigurationType::Root)?;
-    let formatter_disabled = configuration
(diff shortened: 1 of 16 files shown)


crates/rome_cli/src/traversal.rs
```diff
@@ -327,14 +379,14 @@ impl<'ctx, 'app> TraversalOptions<'ctx, 'app> {
         self.workspace.supports_feature(SupportsFeatureParams {
```
leops, line 379: Since the configuration is also injected into the workspace, it might be easier to change the `supports_feature` method to return false if the corresponding feature is disabled. This way we wouldn't need to pass explicit boolean flags all the way through in the traversal context, and turning formatting or linting off in the configuration should also mostly work "out of the box" once we add reading the configuration file in the LSP

crates/rome_cli/src/commands/check.rs
```diff
@@ -3,34 +3,54 @@ use crate::{
...
+            max_diagnostics,
+        }
+    } else {
+        TraversalMode::Check {
+            max_diagnostics,
+            linter_disabled,
+            formatter_disabled,
+            should_fix: false,
+        }
     };
```
leops, line 54: This `if` could be collapsed:
```suggestion
    let mode = TraversalMode::Check {
        linter_disabled,
        formatter_disabled,
        should_fix: session.args.contains("--apply"),
        max_diagnostics,
    };
```

PR #2868, opened 2022-07-12
feat(rome_js_analyze): noExtraBooleanCast

### LINKED ISSUE #2660: noExtraBooleanCast
### Description

From umbrella ticket #2642 
>[noExtraBooleanCast](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noExtraBooleanCast.md)
Disallow unnecessary boolean casts
Example fix:
`if (Boolean(foo)) {} -> if (foo) {}`

IWANABETHATGUY:
## Summary
1. Resolved #2660 

## Test Plan
1. cargo test


crates/rome_js_analyze/tests/specs/noExtraBooleanCast.js.snap
```diff
@@ -0,0 +1,268 @@
...
+0 0 |   // invalid
+1 1 |   if (Boolean(foo)) {
+2 2 |   }
+3   | - if (!!Boolean(foo)) {
+  3 | + if (Boolean(foo)) {
+4 4 |   }
+5 5 |   if (!Boolean(foo)) {
+6 6 |   }
+
+=  note: It is not necessary to use double-negation when a value will already be coerced to a boolean.
```
xunilrj, line 79: Is this true in all cases?
I thought there are cases in which ```!!``` is not redundant.

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// ```
+    pub(crate) NoExtraBooleanCast = "noExtraBooleanCast"
+}
+
+fn is_in_boolean_context(node: &JsSyntaxNode, parent: &JsSyntaxNode) -> Option<bool> {
+    let parent = parent.clone();
+    match parent.kind() {
+        JsSyntaxKind::JS_IF_STATEMENT => parent
+            .cast::<JsIfStatement>()
+            .and_then(|stmt| Some(stmt.test().ok()?.syntax() == node)),
```
xunilrj, line 72: Do you need ```and_then``` here? You are returning ```Option<>```.

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,261 @@
...
+/// arguments: JsCallArguments?
+/// // call expression
+/// JsCallExpression =
+/// callee: JsAnyExpression
+/// optional_chain: '?.'?
+/// type_arguments: TsTypeArguments?
+/// arguments: JsCallArguments
+/// ```
+fn is_boolean_constructor_call(node: &JsSyntaxNode) -> Option<bool> {
+    if JsCallArgumentList::can_cast(node.kind()) {
```
xunilrj, line 111: Same thing here. I think you can use "?" and simplify the logic.

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,261 @@
...
+    /// !x;
+    /// ```
+    ///
+    /// ```js
+    /// !!x;
+    /// ```
+    pub(crate) NoExtraBooleanCast = "noExtraBooleanCast"
+}
+
+fn is_in_boolean_context(node: &JsSyntaxNode, parent: &JsSyntaxNode) -> Option<bool> {
```
xunilrj, line 67: This is the only method without comments.

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,261 @@
...
+    })
+}
+
+/// Check if the SyntaxNode is a Negate Unary Expression
+/// ## Example
+/// ```js
+/// !!x
+/// ```
+fn is_negation(node: &JsSyntaxNode) -> Option<bool> {
+    JsUnaryExpression::cast(node.clone()).and_then(|expr| {
```
xunilrj, line 149: Same thing with ```and_then```

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,261 @@
...
+        let parent_syntax = syntax.parent()?;
+
+        // Check if parent `SyntaxNode` in any boolean `Type Coercion` context,
+        // reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean
+        let parent_node_in_boolean_cast_context = is_in_boolean_context(&syntax, &parent_syntax)
+            .unwrap_or_default()
+            || is_boolean_constructor_call(&parent_syntax).unwrap_or_default()
+            || is_negation(&parent_syntax).unwrap_or_default()
+            || is_boolean_call(&parent_syntax).unwrap_or_default();
+        // Convert `!!x` -> `x` if parent `SyntaxNode` in any boolean `Type Coercion` context
```
xunilrj, line 176: Silly, but would you match ```!(!a)```?

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,261 @@
...
+        if parent_node_in_boolean_cast_context {
+            if is_negation(&syntax).unwrap_or_default() {
```
xunilrj, line 178: Makes sense to extract to ```is_double_negation``` method?

crates/rome_js_analyze/tests/specs/noExtraBooleanCast.js.snap
```diff
@@ -0,0 +1,268 @@
...
+source: crates/rome_js_analyze/tests/spec_tests.rs
+assertion_line: 98
+expression: noExtraBooleanCast.js
+---
+# Input
+```js
+// invalid
+if (Boolean(foo)) {
+}
+if (!!Boolean(foo)) {
```
ematipico, line 11: @leops @xunilrj quick question on how the fixes are displayed/applied. In this example, as you may noticed from the snapshots, we have two diagnostics with two difference code fixes. Although, the snapshot shows the fixes **for the single** diagnostic but it's hard to understand **the final result of the code**.

For example, Rome classic was better at displaying that: https://github.com/rome/tools/blob/archived-js/internal/compiler/lint/rules/js/noExtraBooleanCast.test.md#1 (eventually only `foo`) should remain.

So my question is: what's the final result of the code?

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
IWANABETHATGUY: Yes, I do. Because I need to get the `test` of `JsIfStatement`, and I need to use this `?` for early returning.
Do you mean I could write this code like this:
```rs
JsSyntaxKind::JS_IF_STATEMENT => {
  let stmt = parent.cast::<JsIfStatement>()?;
  Some(stmt.test().ok()?.syntax() == node)
},
```

crates/rome_js_analyze/tests/specs/noExtraBooleanCast.js.snap
leops: Unfortunately merging diagnostics and code actions isn't supported yet at the analyzer infrastructure level. It could be possible to support this by applying code actions on the fly to mutate the tree in place as its being analyzed, but that will probably require some work to get it right. In the meantime it's possible to handle this at the rule level by manually checking for the recursive case in the run function, this is what the `useTemplate` rule does for instance

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
xunilrj: Yes.

IWANABETHATGUY: LGTM

xunilrj: I am so old, I had to look at what "LGTM" stands for! :smile:

IWANABETHATGUY: It means **Looks Good To Me**

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
IWANABETHATGUY: Nope, I am still not sure if checking this case is necessary. Using double-negation is more commongly, but I have nevery saw someone write code like this. 🤣

IWANABETHATGUY: @ematipico @leops , any thought about this? Do we need to support this?

crates/rome_js_analyze/tests/specs/noExtraBooleanCast.js.snap
IWANABETHATGUY: I prefer handle this in infrastructure level. 
I have a simple algorithm for now, maybe not efficient.
1. run `rome_js_analyze`, this should emit some diagnostic with code action.
2. If there are any code action emitted by `rome_js_analyze`, apply them and go to `1`.
3. If there are no code action found , break.
Also, we need to set a threshhold like `10`, avoiding infinite loop.

crates/rome_js_analyze/src/analyzers/no_extra_boolean_cast.rs
IWANABETHATGUY: Got it, I will fix this.

ematipico (changes requested):

crates/rome_js_analyze/tests/specs/noExtraBooleanCast.js.snap
ematipico: > So what is the conclusion?

I'd suggest to leave the diagnostic like this and not trying to improve it. This is something that has to be fixed in the core of the analzyer.

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+/// Check if the SyntaxNode is a Negate Unary Expression
+/// ## Example
+/// ```js
+/// !!x
+/// ```
+fn is_negation(node: &JsSyntaxNode) -> Option<bool> {
+    Some(JsUnaryExpression::cast(node.clone())?.operator().ok()? == JsUnaryOperator::LogicalNot)
```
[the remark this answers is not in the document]
IWANABETHATGUY: That what how i wrote first time, you could see https://github.com/rome/tools/pull/2868#discussion_r920068039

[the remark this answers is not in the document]
ematipico: I missed that comment :)

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let n = ctx.query();
+        let syntax = n.syntax().clone();
+        let parent_syntax = syntax.parent()?;
+
+        // Check if parent `SyntaxNode` in any boolean `Type Coercion` context,
+        // reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean
+        let parent_node_in_boolean_cast_context = is_in_boolean_context(&syntax, &parent_syntax)
+            .unwrap_or_default()
```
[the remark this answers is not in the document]
IWANABETHATGUY: I guess this because is clippy not happy, let me check it.

IWANABETHATGUY:
!bench_analyzer

ematipico:
@IWANABETHATGUY are there any progress on this PR? Do you need any help?

IWANABETHATGUY:
> @IWANABETHATGUY are there any progress on this PR? Do you need any help?

Nope, I will address issues this week.

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+
+/// Check if this node is in the position of `test` slot of parent syntax node.
+/// ## Example
+/// ```js
+/// if (!!x) {
+///     ^^^ this is a boolean context
+/// }
+/// ```
+fn is_in_boolean_context(node: &JsSyntaxNode, parent: &JsSyntaxNode) -> Option<bool> {
+    let parent = parent.clone();
```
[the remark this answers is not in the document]
IWANABETHATGUY: 1. Done,
2. It is not redundant, because the `node` pass into `is_in_boolean_context` is just a `JsAnyExpression`, the expression could be any where, 
**Example** 
```
while (test) {
          ^^^^ Only here is valid, so we need to check 
                   if the expression SyntaxNode is equal to the parent's `test` clause SyntaxNode
}
```

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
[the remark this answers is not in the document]
IWANABETHATGUY: Done

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let n = ctx.query();
+        let syntax = n.syntax().clone();
+        let parent_syntax = syntax.parent()?;
+
+        // Check if parent `SyntaxNode` in any boolean `Type Coercion` context,
+        // reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean
+        let parent_node_in_boolean_cast_context = is_in_boolean_context(&syntax, &parent_syntax)
```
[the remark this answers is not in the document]
IWANABETHATGUY: No, we can't have this early return.
Because the `is_in_boolean_context` only check the scenario that `SyntaxNode` equals to the parent's `test` clause `SyntaxNode` like :
```js
if (xxx) {
    ^^^
}
while (xxx) {
          ^^^
}
//...............and so on.
```
But it could be possible to have an extra boolean cast like: 
```js
new Boolean(xxx);
                      ^^^
// Because when you write something like this
new Boolean(!!xxx)
// the new Boolean `NewExpression` is redundant
```

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+                            .ok()
+                            .map(|argument| (argument, ExtraBooleanCastType::DoubleNegation));
+                    }
+                    _ => {
+                        return None;
+                    }
+                }
+            }
+
+            // Convert `Boolean(x)` -> `x` if parent `SyntaxNode` in any boolean `Type Coercion` context
```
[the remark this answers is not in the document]
IWANABETHATGUY: That would be another case, which I will add below

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+            // Only if `Boolean` Call Expression have one `JsAnyExpression` argument
+            return JsCallExpression::cast(syntax).and_then(|expr| {
```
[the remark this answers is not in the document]
IWANABETHATGUY: Done

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,262 @@
...
+    /// !x;
+    /// ```
+    ///
+    /// ```js
+    /// !!x;
+    /// ```
+    pub(crate) NoExtraBooleanCast = "noExtraBooleanCast"
+}
+
+/// Check if this node is in the position of `test` slot of parent syntax node.
```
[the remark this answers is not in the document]
IWANABETHATGUY: image
Would it be better if I rename it to `test clause of the parent `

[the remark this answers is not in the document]
xunilrj: I think ```JsWhileStatement::test``` would be quite clear.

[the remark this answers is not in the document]
ematipico: It's not about the name of the node, it's about the whole comment. When I read it without checking the code of function, I don't understand what it means.

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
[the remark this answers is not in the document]
ematipico: I understand now, thank you.

IWANABETHATGUY:
!bench_analyzer

IWANABETHATGUY:
I think there should be a more performant way to implement this, let me try.

ematipico:
@xunilrj @leops could you please review and merge once you see fit?

ematipico (approved):

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,292 @@
...
+        let node = ctx.query();
+        let (_, extra_boolean_cast_type) = state;
+        let (title, note) = match extra_boolean_cast_type {
+			ExtraBooleanCastType::DoubleNegation => ("Avoid redundant double-negation.", "It is not necessary to use double-negation when a value will already be coerced to a boolean."),
+			ExtraBooleanCastType::BooleanCall => ("Avoid redundant `Boolean` call", "It is not necessary to use `Boolean` call when a value will already be coerced to a boolean."),
+		};
+        Some(
+            RuleDiagnostic::warning(
+                node.range(),
+                markup! {
```
leops, line 229: Nit: markup is redundant here since the `RuleDiagnostic` constructors accept any type implementing `Display` as a message, just pass `title` directly instead

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,292 @@
...
+            ExtraBooleanCastType::BooleanCall => "Remove redundant `Boolean` call",
+        };
+        let root = ctx
+            .root()
+            .replace_node(node.clone(), node_to_replace.clone())?;
+
+        Some(JsRuleAction {
+            category: ActionCategory::QuickFix,
+            applicability: Applicability::MaybeIncorrect,
+            message: markup! { ""{message}"" }.to_owned(),
```
leops, line 251: Nit: The two empty string literals `""` are not needed, those will just create formatting nodes with an empty content

leops (approved):

crates/rome_js_analyze/src/analyzers/js/no_extra_boolean_cast.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// ```
+    ///
+    /// ### Valid
+    /// ```js
+    /// Boolean(!x);
+    /// !x;
+    /// !!x;
+    /// ```
+    pub(crate) NoExtraBooleanCast {
+        version: "0.8.0",
```
ematipico, line 59: This should be `0.9.0` now. Once updated, we can merge it

IWANABETHATGUY: Done

PR #2870, opened 2022-07-13
fix(rome_js_analyze): only emit control flow side-effects for variable declarators with an initializer

### LINKED ISSUE #2869: 🐛 (rome_js_analyze): false positive (node_dead_code)
### Environment information

```bash
chrome
```

### What happened?

*rome_js_analyze* report a diagnostic: 
the `var a;` is unreachable
```
warning[noDeadCode]: This code is unreachable
  ┌─ main.js:3:3
  │
2 │   return ;
  │   -------- This statement will return from the function ...
3 │   var a;
  │   ------ ... before it can reach this code

``` 
```js
function test() {
  return ;
  var a;
}
```

The diagnostic is false positive becuase `Var` declaration will hoist.
https://developer.mozilla.org/en-US/docs/Glossary/Hoisting#var_hoisting

### Expected result

```js
function test() {
  return ;
  var a;
}
```
The code above `rome_js_analyze` should not  emit a diagnostic, because line 3 is a `declaration` without an initialization

```js
function test() {
  return ;
  var a = 10;
  ^^^^^^^  ------------ the code is unreachable
}
```
The code above `rome_js_analyze` should emit a diagnostic, because line 3 is a `declaration` with a initialization

### Code of Conduct

leops:
## Summary

This PR modifies how control flow instructions are emitted for variable statements: instead of considering the entire statement as a side effect, each initializer expression now creates it own instruction. This means variable declarations without an initializer are not considered to have any side effects in terms of control flow, checking whether the variable itself it unused will be part of the semantic analysis

Fixes #2869

## Test Plan

I've added additional test cases in `JsVariableStatement.js` to check the correct warnings are being emitted

### DIFF
--- a/crates/rome_js_analyze/tests/specs/noDeadCode/JsVariableStatement.js.snap
+++ b/crates/rome_js_analyze/tests/specs/noDeadCode/JsVariableStatement.js.snap
@@ -1,0 +1,103 @@
+---
+source: crates/rome_js_analyze/tests/spec_tests.rs
+assertion_line: 98
+expression: JsVariableStatement.js
+---
+# Input
+```js
+function JsVariableStatement1() {
+    return;
+    var variable;
+}
+
+function JsVariableStatement2() {
+    return;
+    var variable = initializer();
+}
+
+function JsVariableStatement3() {
+    return;
+    let variable;
+}
+
+function JsVariableStatement4() {
+    return;
+    let variable = initializer();
+}
+
+function JsVariableStatement5() {
+    return;
+    const variable = initializer();
+}
+
+function JsVariableStatement6() {
+    return;
+    var variable1 = initializer(),
+        variable2 = initializer();
+}
+
+```
+
+# Diagnostics
+```
+warning[noDeadCode]: This code is unreachable
+  ┌─ JsVariableStatement.js:8:20
+  │
+7 │     return;
+  │     ------- This statement will return from the function ...
+8 │     var variable = initializer();
+  │                    ------------- ... before it can reach this code
+
+
+```
+
+```
+warning[noDeadCode]: This code is unreachable
+   ┌─ JsVariableStatement.js:18:20
+   │
+17 │     return;
+   │     ------- This statement will return from the function ...
+18 │     let variable = initializer();
+   │                    ------------- ... before it can reach this code
+
+
+```
+
+```
+warning[noDeadCode]: This code is unreachable
+   ┌─ JsVariableStatement.js:23:22
+   │
+22 │     return;
+   │     ------- This statement will return from the function ...
+23 │     const variable = initializer();
+   │                      ------------- ... before it can reach this code
+
+
+```
+
+```
+warning[noDeadCode]: This code is unreachable
+   ┌─ JsVariableStatement.js:28:21
+   │
+27 │     return;
+   │     ------- This statement will return from the function ...
+28 │     var variable1 = initializer(),
+   │                     ------------- ... before it can reach this code
+
+
+```
+
+```
+warning[noDeadCode]: This code is unreachable
+   ┌─ JsVariableStatement.js:29:21
+   │
+27 │     return;
+   │     ------- This statement will return from the function ...
+28 │     var variable1 = initializer(),
+29 │         variable2 = initializer();
+   │                     ------------- ... before it can reach this code
+
+
+```
+
+
--- a/crates/rome_js_analyze/src/control_flow/nodes.rs
+++ b/crates/rome_js_analyze/src/control_flow/nodes.rs
@@ -11,6 +11,7 @@ mod statement;
 mod switch_stmt;
 mod throw_stmt;
 mod try_catch;
+mod variable;
 mod while_stmt;
 
 pub(super) use block::*;
@@ -26,4 +27,5 @@ pub(super) use statement::*;
 pub(super) use switch_stmt::*;
 pub(super) use throw_stmt::*;
 pub(super) use try_catch::*;
+pub(super) use variable::*;
 pub(super) use while_stmt::*;
--- a/crates/rome_js_analyze/src/control_flow/nodes/statement.rs
+++ b/crates/rome_js_analyze/src/control_flow/nodes/statement.rs
@@ -1,6 +1,4 @@
-use rome_js_syntax::{
-    JsDebuggerStatement, JsEmptyStatement, JsExpressionStatement, JsVariableStatement,
-};
+use rome_js_syntax::{JsDebuggerStatement, JsEmptyStatement, JsExpressionStatement};
 use rome_rowan::{declare_node_union, AstNode, SyntaxResult};
 
 use crate::control_flow::{
@@ -9,7 +7,7 @@ use crate::control_flow::{
 };
 
 declare_node_union! {
-    pub(in crate::control_flow) JsSimpleStatement = JsDebuggerStatement | JsEmptyStatement | JsExpressionStatement | JsVariableStatement
+    pub(in crate::control_flow) JsSimpleStatement = JsDebuggerStatement | JsEmptyStatement | JsExpressionStatement
 }
 
 pub(in crate::control_flow) struct StatementVisitor;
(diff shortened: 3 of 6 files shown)


crates/rome_js_analyze/tests/specs/noDeadCode/JsVariableStatement.js.snap
```diff
@@ -0,0 +1,103 @@
...
+
+function JsVariableStatement2() {
+    return;
+    var variable = initializer();
+}
+
+function JsVariableStatement3() {
+    return;
+    let variable;
+}
```
ematipico, line 21: This case should throw a diagnostic, even when it gets assigned. Same for `const`.

[Examples](https://eslint.org/play#eyJ0ZXh0IjoibGV0IGEgPSBmdW5jdGlvbiBmKCkge1xuICByZXR1cm47XG4gIGxldCBhO1xufVxuXG5sZXQgZmYgPSBmdW5jdGlvbiBmKCkge1xuICByZXR1cm47XG4gIGxldCBzb21ldGhpbmcgPSAyO1xufVxuXG5cblxubGV0IGMgPSBmdW5jdGlvbiBmKCkge1xuICByZXR1cm47XG4gIGNvbnN0IGEgPSBiO1xufVxuXG5cbmEoKVxuYygpXG5mZigpIiwib3B0aW9ucyI6eyJwYXJzZXJPcHRpb25zIjp7ImVjbWFWZXJzaW9uIjoibGF0ZXN0Iiwic291cmNlVHlwZSI6InNjcmlwdCIsImVjbWFGZWF0dXJlcyI6e319LCJydWxlcyI6eyJjb25zdHJ1Y3Rvci1zdXBlciI6MiwiZm9yLWRpcmVjdGlvbiI6MiwiZ2V0dGVyLXJldHVybiI6Miwibm8tYXN5bmMtcHJvbWlzZS1leGVjdXRvciI6Miwibm8tY2FzZS1kZWNsYXJhdGlvbnMiOjIsIm5vLWNsYXNzLWFzc2lnbiI6Miwibm8tY29tcGFyZS1uZWctemVybyI6Miwibm8tY29uZC1hc3NpZ24iOjIsIm5vLWNvbnN0LWFzc2lnbiI6Miwibm8tY29uc3RhbnQtY29uZGl0aW9uIjoyLCJuby1jb250cm9sLXJlZ2V4IjoyLCJuby1kZWJ1Z2dlciI6Miwibm8tZGVsZXRlLXZhciI6Miwibm8tZHVwZS1hcmdzIjoyLCJuby1kdXBlLWNsYXNzLW1lbWJlcnMiOjIsIm5vLWR1cGUtZWxzZS1pZiI6Miwibm8tZHVwZS1rZXlzIjoyLCJuby1kdXBsaWNhdGUtY2FzZSI6Miwibm8tZW1wdHkiOjIsIm5vLWVtcHR5LWNoYXJhY3Rlci1jbGFzcyI6Miwibm8tZW1wdHktcGF0dGVybiI6Miwibm8tZXgtYXNzaWduIjoyLCJuby1leHRyYS1ib29sZWFuLWNhc3QiOjIsIm5vLWV4dHJhLXNlbWkiOjIsIm5vLWZhbGx0aHJvdWdoIjoyLCJuby1mdW5jLWFzc2lnbiI6Miwibm8tZ2xvYmFsLWFzc2lnbiI6Miwibm8taW1wb3J0LWFzc2lnbiI6Miwibm8taW5uZXItZGVjbGFyYXRpb25zIjoyLCJuby1pbnZhbGlkLXJlZ2V4cCI6Miwibm8taXJyZWd1bGFyLXdoaXRlc3BhY2UiOjIsIm5vLWxvc3Mtb2YtcHJlY2lzaW9uIjoyLCJuby1taXNsZWFkaW5nLWNoYXJhY3Rlci1jbGFzcyI6Miwibm8tbWl4ZWQtc3BhY2VzLWFuZC10YWJzIjoyLCJuby1uZXctc3ltYm9sIjoyLCJuby1ub25vY3RhbC1kZWNpbWFsLWVzY2FwZSI6Miwibm8tb2JqLWNhbGxzIjoyLCJuby1vY3RhbCI6Miwibm8tcHJvdG90eXBlLWJ1aWx0aW5zIjoyLCJuby1yZWRlY2xhcmUiOjIsIm5vLXJlZ2V4LXNwYWNlcyI6Miwibm8tc2VsZi1hc3NpZ24iOjIsIm5vLXNldHRlci1yZXR1cm4iOjIsIm5vLXNoYWRvdy1yZXN0cmljdGVkLW5hbWVzIjoyLCJuby1zcGFyc2UtYXJyYXlzIjoyLCJuby10aGlzLWJlZm9yZS1zdXBlciI6Miwibm8tdW5kZWYiOjIsIm5vLXVuZXhwZWN0ZWQtbXVsdGlsaW5lIjoyLCJuby11bnJlYWNoYWJsZSI6Miwibm8tdW5zYWZlLWZpbmFsbHkiOjIsIm5vLXVuc2FmZS1uZWdhdGlvbiI6Miwibm8tdW5zYWZlLW9wdGlvbmFsLWNoYWluaW5nIjoyLCJuby11bnVzZWQtbGFiZWxzIjoyLCJuby11bnVzZWQtdmFycyI6Miwibm8tdXNlbGVzcy1iYWNrcmVmZXJlbmNlIjoyLCJuby11c2VsZXNzLWNhdGNoIjoyLCJuby11c2VsZXNzLWVzY2FwZSI6Miwibm8td2l0aCI6MiwicmVxdWlyZS15aWVsZCI6MiwidXNlLWlzbmFuIjoyLCJ2YWxpZC10eXBlb2YiOjJ9LCJlbnYiOnt9fX0=)

ematipico (changes requested):

crates/rome_js_analyze/tests/specs/noDeadCode/JsVariableStatement.js.snap
leops: As I mentioned in the PR description, I expect a diagnostic for this case will be emitted by a future `noUnusedVariable` lint rule based on the semantic analysis instead. Any access to the variable before the return statement is invalid and will emit an error (since reading or writing a `let` variable before its declared is not permitted), while any access to the variable coming after the declaration will also be marked as unreachable and should be ignored by the `noUnusedVariable` rule. Since the variable will be marked as unused by the semantic analysis I wanted to avoid a duplicate diagnostic flagging it as unreachable again in the control flow analysis, so the control flow diagnostic only covers the initializer expression

ematipico: I thought you were just covering the `var` declarations, because also the `let` don't need an initializers. Thank you

xunilrj: > while any access to the variable coming after the declaration will also be marked as unreachable and should be ignored by the noUnusedVariable rule.

Do we want that? This would imply a rule that needs the ```SemanticModel```and the ```CFG```. This means that we need to check if every reference is reachable or not.

crates/rome_js_analyze/tests/specs/noDeadCode/JsVariableStatement.js.snap
```diff
@@ -0,0 +1,103 @@
...
+
+```
+warning[noDeadCode]: This code is unreachable
+   ┌─ JsVariableStatement.js:29:21
+   │
+27 │     return;
+   │     ------- This statement will return from the function ...
+28 │     var variable1 = initializer(),
+29 │         variable2 = initializer();
+   │                     ------------- ... before it can reach this code
```
xunilrj, line 98: I was expecting this PR to do something like:

```
warning[noDeadCode]: This code is unreachable
   ┌─ JsVariableStatement.js:29:21
   │
27 │     return;
   │     ------- This statement will return from the function ...
28 │     var variable1 = initializer(),
   │                     ------------- ... before it can reach this code
             variable2,
29 │         variable3 = initializer();
   │                     ------------- ... before it can reach this code
```

leops: This is due to a limitation in the Language Server Protocol: when the diagnostic is displayed in the editor only the primary range will be rendered as "unnecessary", and since diagnostics can only have a single primary label the linter needs to emit a new diagnostic for each non-contiguous range of code that's unreachable

ematipico (approved):

xunilrj (approved):

PR #2874, opened 2022-07-14
feat(rome_js_semantic): Enrich the semantic model with read/write information

xunilrj:
## Summary

This PR enriches the Semantic Model with information about if references are "reading" or "writing". This can really simplify rules like https://github.com/rome/tools/pull/2849#issuecomment-1181544741

One interesting point is what to do with initializers. Today I **do not** flag initializers as writes.

## Test Plan

```
> cargo test -p rome_js_semantic -- ok_semantic_model
```

### DIFF
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -60,7 +60,7 @@ impl SemanticModelData {
         &self,
         range: &TextRange,
     ) -> std::slice::Iter<'_, (ReferenceType, TextRange)> {
-        if let Some(v) = self.declaration_all_references.get(&range) {
+        if let Some(v) = self.declaration_all_references.get(range) {
             v.iter()
         } else {
             [].iter()
@@ -71,7 +71,7 @@ impl SemanticModelData {
         &self,
         range: &TextRange,
     ) -> std::slice::Iter<'_, (ReferenceType, TextRange)> {
-        if let Some(v) = self.declaration_all_reads.get(&range) {
+        if let Some(v) = self.declaration_all_reads.get(range) {
             v.iter()
         } else {
             [].iter()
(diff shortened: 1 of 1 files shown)


crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -48,7 +48,46 @@ struct SemanticModelData {
...
+
+impl SemanticModelData {
+    pub fn all_references_iter(
+        &self,
+        range: &TextRange,
+    ) -> std::slice::Iter<'_, (ReferenceType, TextRange)> {
+        if let Some(v) = self.declaration_all_references.get(range) {
+            v.iter()
+        } else {
+            [].iter()
```
IWANABETHATGUY, line 66: I am not sure if this `[]` array would introduce a onstack allocation, I think `std::iter::empty()` maybe more proper here ?

IWANABETHATGUY: Ignore me please, I did't notice the type.

xunilrj: The problem is that I need the method to return an ```std::slice::Iter``` for now.
Because extensions methods need to return a know type:

```
error[E0562]: `impl Trait` only allowed in function and inherent method return types, not in trait method return
   --> crates/rome_js_semantic/src/semantic_model.rs:533:59
    |
533 |     fn all_writes<'a>(&self, model: &'a SemanticModel) -> impl Iterator<Item = Reference>
    |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0562`.
error: could not compile `rome_js_semantic` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `rome_js_semantic` due to previous error
```

IWANABETHATGUY: Got it.

ematipico (approved):

PR #2882, opened 2022-07-15
refactor(rome_analyze): merge analyzer phases

leops:
## Summary

This change merges the execution of the multiple phases of the analyzer into the `Analyzer` struct itself. This allows some derived data like the line suppression cache to be reused between phases, generally improving the performances of the analysis.

In order to support this I've modified the way the way the semantic model is generate to use a `Visitor`, and added a new `finish` hook to visitors called once the entire file has been visited to give them an opportunity to mutate the service bag and inject additional services for subsequent phases.

In turn, in order to provide the visitors with a mutable access to the service bag I changed the way ownership of the bag is managed from using reference counting to using normal Rust references tied to an explicit `'phase` lifetime: the service bag is immutably borrowed by the visitors, rules and signals for the duration of the phase, then mutably borrowed by the visitors once the phase completes.

I've also included a few other optimizations that were found while profiling, namely marking some methods in `rome_js_semantic` as inlinable, and changing the `NoArguments` rule to read the text of the token directly instead of the binding node

## Test Plan

This is an infrastructure change, it should not have any noticeable side-effects and all the tests for the analyzer should continue to pass

### DIFF
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -114,10 +114,13 @@ where
                 range: ctx.range,
             };
 
+            // The first phase being run will inspect the tokens and parse the
+            // suppression comments, then subsequent phases only needs to read
+            // this data again since it's already cached in `line_suppressions`
             let result = if index == 0 {
-                runner.run_initial_phase()
+                runner.run_first_phase()
             } else {
-                runner.run_cached_phase()
+                runner.run_remaining_phases()
             };
 
             if let ControlFlow::Break(br) = result {
@@ -127,7 +130,7 @@ where
             // Finish all the active visitors, this is executed outside of the
             // phase runner as it needs mutable access to the service bag (the
             // runner borrows the services for the entire phase)
-            for mut visitor in visitors {
+            for visitor in visitors {
                 visitor.finish(VisitorFinishContext {
                     root: &ctx.root,
                     services: &mut ctx.services,
@@ -189,7 +192,7 @@ where
 {
     /// Runs phase 0 over nodes and tokens to process line breaks and
     /// suppression comments
-    fn run_initial_phase(mut self) -> ControlFlow<Break> {
+    fn run_first_phase(mut self) -> ControlFlow<Break> {
         let iter = self.root.syntax().preorder_with_tokens(Direction::Next);
         for event in iter {
             let node_event = match event {
@@ -229,7 +232,7 @@ where
 
     /// Runs phases 1..N over nodes, since suppression comments were already
     /// processed and cached in `run_initial_phase`
-    fn run_cached_phase(mut self) -> ControlFlow<Break> {
+    fn run_remaining_phases(mut self) -> ControlFlow<Break> {
         for event in self.root.syntax().preorder() {
             // Run all the active visitors for the phace on the event
             for visitor in self.visitors.iter_mut() {
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -2,7 +2,6 @@
 
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, BinaryHeap};
-use std::mem::swap;
 use std::ops;
 
 mod categories;
@@ -18,14 +17,14 @@ mod visitor;
 
 pub use crate::categories::{ActionCategory, RuleCategories, RuleCategory};
 pub use crate::matcher::{QueryMatcher, RuleKey, SignalEntry};
-pub use crate::query::{Ast, CannotCreateServicesError, QueryKey, QueryMatch, Queryable};
+pub use crate::query::{Ast, QueryKey, QueryMatch, Queryable};
 pub use crate::registry::{LanguageRoot, Phase, Phases, RuleMetadata, RuleRegistry};
 pub use crate::rule::{GroupLanguage, Rule, RuleAction, RuleDiagnostic, RuleGroup, RuleMeta};
-pub use crate::services::{ServiceBag, ServiceBagData};
+pub use crate::services::{CannotCreateServicesError, FromServices, ServiceBag};
 use crate::signals::DiagnosticSignal;
 pub use crate::signals::{AnalyzerAction, AnalyzerSignal};
 pub use crate::syntax::SyntaxVisitor;
-pub use crate::visitor::{NodeVisitor, Visitor, VisitorContext};
+pub use crate::visitor::{NodeVisitor, Visitor, VisitorContext, VisitorFinishContext};
 use rome_console::markup;
 use rome_diagnostics::file::FileId;
 use rome_diagnostics::Diagnostic;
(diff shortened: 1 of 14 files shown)


leops:
!bench_analyzer

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -573,10 +573,12 @@ impl SemanticModelBuilder {
         }
     }
 
+    #[inline]
     pub fn push_node(&mut self, node: &JsSyntaxNode) {
         self.node_by_range.insert(node.text_range(), node.clone());
     }
 
+    #[inline]
```
IWANABETHATGUY, line 581: Does this `inline` really suitable? Since the function has a big chunk of code, which may increase the binary .

leops: The `inline` attribute only instructs rustc to emit the metadata required for the function to be inlined across crates, the final decision of whether to inline the function or not is still left to the optimization heuristics of rustc and LLVM.
Previously these methods were called from the `rome_js_semantic` crate itself (in `semantic_model`), so it was pretty much guaranteed that rustc would be able to inline them, but since they're now being called from `rome_js_analyze` I added the `inline` attribute to ensure rustc would be able to apply the same optimizations to the new callsite

crates/rome_analyze/src/lib.rs
```diff
@@ -130,29 +207,56 @@ where
...
         }
 
-        None
+        // Flush all remaining pending events
+        self.flush_matches(None)
+    }
+
+    /// Runs phases 1..N over nodes, since suppression comments were already
+    /// processed and cached in `run_initial_phase`
+    fn run_cached_phase(mut self) -> ControlFlow<Break> {
```
MichaReiser, line 232: What's the difference of `cached` phases to normal phases? 

Would the formatters comments (and suppression?) extraction be a cached or uncached phase?

MichaReiser (approved):

crates/rome_analyze/src/lib.rs
leops: The only real difference between uncached and cached phases is that uncached phases inspect the tokens and update the state of `line_suppressions` as suppression comments get visited, while cached phase only inspect nodes and suppress signals by reading the state of and existing instance of `line_supressions`.
As such the extraction of comments would run in the uncached phase since it needs to access tokens, and is technically about building up a representation of comments similar to what the uncached phase infrastructure is doing with `line_supressions`

leops: To be honest I don't really like the "cached" name either, but the only other idea I had was to call the method `run_followup_phase` and it didn't feel much more explicit about what's going on

MichaReiser: For now, we only have two phases: 
- syntax
- semantic

How about naming the phases `run_syntax_phase` and `run_semantic_phase`? 

What would be nice is if it would be possible to move the custom line number and comments handling out into a visitor.

Doing so could be beneficial if someone only runs the formatter in which case the suppression comments aren't needed (they are computed in a formatter specific way).

crates/rome_analyze/src/lib.rs
```diff
@@ -89,39 +67,138 @@ where
...
+            phases,
+            mut query_matcher,
+            parse_suppression_comment,
+            mut emit_signal,
+        } = self;
+
+        let mut line_index = 0;
+        let mut line_suppressions = Vec::new();
+
+        for (index, (phase, mut visitors)) in phases.into_iter().enumerate() {
```
ematipico, line 101: The iteration seems a bit unnatural to me. Because of the fact that `0` is always the uncached phase, I believe we should restructure the code in a way that makes that more explicit.

Something like:

```rs
let iter = phases.into_iter();
let uncached_phase = iter.next();
if let Some(uncached_phase) = uncached_phase {
	// .. do stuff
}

// loop here for cached phase, consume the rest of the iterator
```

crates/rome_analyze/src/lib.rs
leops: I didn't want to named the "initial" and "followup" phases with specific names since technically a consumer of the Analyzer is allowed to only add visitors to the semantic phase, then the syntax phase would not be run and the semantic phase would become the initial phase (for instance if the syntax tree hasn't changed but the semantic model has been update with external information coming from other files in the workspace, and we want to only re-run the semantic lint rules)

Moving the suppression comment logic to a visitor is how I originally started implementing this, but this significantly complicated the implementation at the time so I ended up moving all this logic to the analysis infrastructure itself. Maybe this decision could now be revisited since the implementation has evolved a bit from my initial prototype, however I still want to carefully how to allow visitors to access individual tokens since this incurs a significant slowdown on traversal.

crates/rome_analyze/src/lib.rs
leops: This is how I originally had it written, but I had to duplicate the `PhaseRunner` constructor, control flow break and visitor finish loop in both blocks, which I found made the small difference between the two (calling `run_initial_phase` vs. `run_cached_phase`) easy to miss

crates/rome_js_analyze/src/semantic_services.rs
```diff
@@ -57,3 +58,49 @@ where
...
+                self.extractor.leave(node);
+            }
+        }
+
+        while let Some(e) = self.extractor.pop() {
+            self.builder.push_event(e);
+        }
+    }
+
+    fn finish(&mut self, ctx: VisitorFinishContext<JsLanguage>) {
```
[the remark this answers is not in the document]
leops: The trait isn't allowed to have `self` methods since it's stored as a `Box<dyn Visitor>` trait object, but I think changing the method signature to `self: Box<Self>` might work

crates/rome_analyze/src/lib.rs
```diff
@@ -349,7 +461,7 @@ where
         let mut did_match = false;
         for (index, _) in text.match_indices('\n') {
```
MichaReiser, line 462: But I see that this is challenging because there are some syntax elements like template literals and JSX that can contain line breaks that are not represented as newline trivia. 

The main issue with hard coding line breaks to `\n` is that this won't work with e.g. `\r`. I don't know if there are any language-specific line break characters that are worth checking.

leops: > But I see that this is challenging because there are some syntax elements like template literals and JSX that can contain line breaks that are not represented as newline trivia.

The analyzer tracks an arbitrary `line_index` instead of the actual line number specifically so that it could be allowed deviate from the line number when handling multiline token. This isn't the case currently since the text of tokens is also checked for line breaks, but it could be beneficial to consider a JSX text or template literal to be on the same "line index" even if it's not actually on the same line within the file. This would allow a suppression comment on the line preceding the token to suppress diagnostics for the whole literal instead of just the first line, which may be what the user actually wants considering it's not possible to add a suppression comment in the middle of a token

crates/rome_analyze/src/lib.rs
MichaReiser: I see. Last recommendation. `run_first_phase` and run `run_remaining_phases` or `run_head_phase` and `run_tail_phases`

crates/rome_js_analyze/src/lib.rs
```diff
@@ -59,41 +58,17 @@ where
...
-    services.insert_service(model);
-
-    let mut analyzer = Analyzer::new(
-        build_registry(&filter),
-        parse_linter_suppression_comment,
-        &mut emit_signal,
-    );
-
-    analyzer.add_visitor(SyntaxVisitor::default());
+    analyzer.add_visitor(Phases::Semantic, SyntaxVisitor::default());
```
MichaReiser, line 65: Always adding a `SyntaxVisitor` to the syntax and semantic phases has the downside that the analyzer always does at least two passes even if there are only syntax xor semantic visitors. Do you have something in mind to build up the visitors more on demand?

crates/rome_js_analyze/src/lib.rs
```diff
@@ -59,41 +58,17 @@ where
         &mut emit_signal,
     );
 
-    analyzer.add_visitor(make_visitor());
+    analyzer.add_visitor(Phases::Syntax, make_visitor());
+    analyzer.add_visitor(Phases::Syntax, SyntaxVisitor::default());
+    analyzer.add_visitor(Phases::Syntax, SemanticVisitor::new(root));
```
MichaReiser, line 63: Nit: It can be misleading why the `SemanticVisitor` runs in the `Syntax` phase. How about renaming it to `BuildSemanticModelVisitor`?

crates/rome_js_analyze/src/lib.rs
leops: What I have in mind is expose methods on the `RuleRegistry` to check whether it has any rules for a certain phase or query kind (AST node, control flow graph, ...), but ultimately since the visitors are allowed to have side-effects it will be up to the consumer of the Analyzer to use this information to decide whether or not to register a certain visitor

PR #2883, opened 2022-07-15
fix(playground): malformed prettier formating result

IWANABETHATGUY:
## Summary
1. Downgrade `vite` to `2.9.5`

## Test Plan
1. No need to test I guess, or you could run the code locally.

### DIFF
--- a/website/playground/package.json
+++ b/website/playground/package.json
@@ -34,7 +34,7 @@
 		"rome": "^0.7.0-next",
 		"tailwindcss": "^3.0.19",
 		"typescript": "^4.4.4",
-		"vite": "^3.0.0",
+		"vite": "2.9.5",
 		"vite-plugin-svgr": "^2.2.0"
 	},
 	"engines": {
--- a/website/playground/pnpm-lock.yaml
+++ b/website/playground/pnpm-lock.yaml
@@ -19,7 +19,7 @@ specifiers:
   rome: ^0.7.0-next
   tailwindcss: ^3.0.19
   typescript: ^4.4.4
-  vite: ^3.0.0
+  vite: 2.9.5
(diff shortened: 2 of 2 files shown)


vramana:
How can Vite be the problem here?

IWANABETHATGUY:
> How can Vite be the problem here?



> How can Vite be the problem here?

So what do you think the problem is?

MichaReiser:
Thanks @IWANABETHATGUY for looking into this. Can you provide some additional information on what the exact problem is that you're solving? E.g. what's your test plan to verify that this change is indeed fixing the problem (and maybe add some additional description on what problem this PR is solving in the first place)? 

Is this solving the [object Object] that occasionally appears in the playground? If so, then I expect this to be a race condition inside of our react code. I would suspect some misbehaving between initializing the context from the URL params and the first time the `effect` updating the URL runs.

ematipico:
> Thanks @IWANABETHATGUY for looking into this. Can you provide some additional information on what the exact problem is that you're solving? E.g. what's your test plan to verify that this change is indeed fixing the problem (and maybe add some additional description on what problem this PR is solving in the first place)?
> 
> Is this solving the [object Object] that occasionally appears in the playground? If so, then I expect this to be a race condition inside of our react code. I would suspect some misbehaving between initializing the context from the URL params and the first time the `effect` updating the URL runs.

The issue that this fix solves  happens only in production, only on the prettier output. All the lines after the first one have an additional level of indentation.

IWANABETHATGUY:
Because I don't have permission to deploy playground,so I add a screenshot:
## Current  (production build and serve the static `/dist` via `live-server`)
image
## Main
image

MichaReiser (approved):

PR #2885, opened 2022-07-16
feat(vscode): add a trace channel for better debugging

IWANABETHATGUY:
## Summary
1. Add a new channel to vscode client, help for debugging the language server.

## Test Plan
1. You could build and install locally , start tracing `rome_lsp` by adding `"rome_lsp.trace.server": "verbose"` to your `setting.json`


IWANABETHATGUY:
here is a sample of the tracing output:
```bash

[Trace - 17:47:24] Received response 'textDocument/codeAction - (44)' in 1ms.
Result: [
    {
        "edit": {
            "changes": {
                "file:///home/Documents/rust/github/tools/editors/vscode/src/main.ts": [
                    {
                        "newText": "true",
                        "range": {
                            "end": {
                                "character": 17,
                                "line": 125
                            },
                            "start": {
                                "character": 8,
                                "line": 125
                            }
                        }
                    }
                ]
            }
        },
        "kind": "quickfix",
        "title": "Discard redundant terms from the logical expression."
    }
]

```

editors/vscode/package.json
```diff
@@ -59,6 +59,22 @@
 			"title": "Rome",
 			"type": "object",
 			"properties": {
+				"rome_lsp.trace.server": {
```
IWANABETHATGUY, line 62: Why using `rome_lsp` instead of `rome` ?  see //reference  https://github.com/Microsoft/vscode-languageserver-node/issues/384

ematipico:
We already have some debugging already, could you please explain what's the difference from what we have now and what you added?

IWANABETHATGUY:
Also, this `traceChannel` could catch all the `language-server-protocol` content sent by the client and server.

ematipico:
Would you mind sharing some screenshot of the information sent between client an server of this PR?

IWANABETHATGUY:
> 

image

editors/vscode/src/main.ts
```diff
@@ -31,6 +31,8 @@ export async function activate(context: ExtensionContext) {
 		transport: TransportKind.stdio,
 	};
 
+	const traceOutputChannel = window.createOutputChannel("rome_lsp trace");
```
ematipico, line 34: What's the label for?

IWANABETHATGUY: image
trace chanel label

ematipico: Can we give a better name then? Like `"Rome Server Trace"`

IWANABETHATGUY: Yes

IWANABETHATGUY: Done

ematipico:
> The Ci is broken because our CI uses `preview`, and we use `rome@next-0.7.0` in package.json

Actually, this is the change that we wanted. We just need to update the `package.json` with the preview, but I guess we can do it in another PR. Thank you!

ematipico (approved):

PR #2888, opened 2022-07-18
refactor(playground): Worker code handling

MichaReiser:
This PR improves some of the playground's worker related code:

* Use `new Worker` instead of `?worker` as [recommended by vite](https://vitejs.dev/guide/features.html#import-with-constructors)
* Handle timeouts inside react as this is a UI concern
* Log unknown messages
* Make `init` a command

## Test Plan

I built the playground locally in test and production build and verified that it correctly formats the code


website/playground/src/App.tsx
```diff
@@ -66,8 +77,13 @@ function App() {
 		if (loadingState !== LoadingState.Success) {
 			return;
 		}
-		romeWorkerRef.current?.postMessage({ type: "format", playgroundState });
-		prettierWorkerRef.current?.postMessage({ type: "format", playgroundState });
+
+		let timeout = setTimeout(() => {
```
ematipico, line 81: What's the timeout for? Could you please add a comment? Usually if we have a `setTimeout` it means there's something wrong. A comment should explain why it's there.

ematipico (changes requested):
The CI is failing because there are some linting issues (they are not blocking)

website/playground/src/App.tsx
ematipico: Oh, so maybe having a debounce would make more sense?

ematipico (approved):

Issue #2741, opened 2022-06-20
☂️ Proposed lint rules, from other sources


ematipico:
# Note

**The teams decided that they prefer to tackle this rules only after #2642 is closed**

# Description

These are ***proposed lint rules*** that come from other sources. They are proposed, they might require a brief discussion for the following reasons:
- they might come from `eslint`, which have a "configuration" per rule. Rome doesn't have a configuration per rule, which means that we might not be able to implement it of we might need to write a default version of that rule, that's why a discussion/decision might be needed;
- they might come from a different language, where some cases might not apply to TS/JS. So it requires to understand the rule and try to apply it to JS/TS

# If you want to contribute
1. make sure you understand the requirements of the rule you would like to implement;
3. comment on this very issue telling saying which rule you would like to implement;
4. create a new task issue and link to this very umbrella issue (a member of the team will eventually assign it to you);
5. open a PR;

> **Note**: please make sure to comment the issue saying which rule you'd like to implement, this would allow better coordination between contributors. Failing to do so would penalize yourselves against contributors that had followed these simple guildelines

> **Note**: follow the naming convention guidelines: https://github.com/rome/tools/blob/archived-js/CONTRIBUTING.md#naming-patterns

## ESLint

- [x] #2649 
Enforces comparing `typeof` expressions against valid strings
Can provide auto-fixes if the result of `typeof` is being compared to a well-known constant
Example fixes:
`typeof a === String` -> `typeof a === "string"`
`typeof a === undefined` -> `typeof a === "undefined"`
- [ ] [dot-notation](https://eslint.org/docs/rules/dot-notation)
Enforces dot notation instead of square bracket notation for member access with static names
Example fix:
`foo["bar"]` -> `foo.bar`
- [ ] [no-else-return](https://eslint.org/docs/rules/no-else-return)
Disallow else blocks for if statements ending with a return
Example fix:
`if (a) { return 1; } else { return 2; }` -> `if (a) { return 1; } return 2;`
- [ ] [no-extra-bind](https://eslint.org/docs/rules/no-extra-bind)
Disallow calling `.bind()` on functions that never access `this`
- [ ] [no-extra-label](https://eslint.org/docs/rules/no-extra-label)
Disallow using single-level deep labelled breaks
Example fix:
`label: while(true) { break label; }` -> `while(true) { break; }`
- [ ] [no-implicit-coercion](https://eslint.org/docs/rules/no-implicit-coercion)
Disallow shorthand type conversions
Example fixes:
[...]

PR #2890, opened 2022-07-18
feat(rome_service): auto generate configuration for rules

### LINKED ISSUE #2887: auto generate lint configuration for disabling rules

ematipico:
## Summary

Closes #2887 

This PR implements a way to auto generate the configuration of the `rome.json` of the rules using the metadata defined in each rule. 

The configuration, in this PR, is now shaped in this way:

```json
{
  "linter": {
    "enabled": true,
    "globals": ["$"],
    "rules": {
        "js": {
            "noDeadCode": "off",
            "useSimplifiedLogicExpression": "warn"
        }
    }
  }
}
```

There's a new field called `globals`. We will use this field to store global references to the root scope of the semantic model cc @xunilrj 

There's a new field called `rules` which will contain the rules. The rules divided by groups. This all auto generate from the metadata, so it will change accordingly cc @leops .

Each rule is defined by a `RuleConfiguration`, where we can:
- turn off a rule
- make a rule to emit warnings
- make a rule to emit errors

A rule is **not serialized** if it's configured to emit diagnostics, because it's the default. This could change if we will use a recommendation system.

Consuming these defaults **is out of scope**, this will be implemented in other PRs. 

As for now, all the rules are enabled and their default is `RuleConfiguration::Error`, although since they are not consumed yet, it won't make a difference for now. 

The idea is to enable certain rules as "recommended", but this is also out of scope and it will be implemented in another PR.

I also fixed an error on the max diagnostics printer, the message was incorrectly emitted every  time.

## Test Plan

Because the configuration is not consume it yet, there's no practical test. I added for unknown field on the rule, which demonstrates that the configuration is correctly deserialized by `serde`.


crates/rome_service/src/configuration/linter/rules.rs
```diff
@@ -0,0 +1,106 @@
+//! Generated file, do not edit by hand, see `xtask/codegen`
+
+use serde::{Deserialize, Serialize};
+#[derive(Deserialize, Serialize, Debug, Eq, PartialEq, Clone)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub enum RuleConfiguration {
+    Warn,
+    Error,
+    Off,
+}
```
MichaReiser, line 10: Is it possible to move `RuleConfiguration` out of the generated file? It seems to me that its implementation doesn't depend on any generated code.

crates/rome_service/src/settings.rs
```diff
@@ -55,16 +61,27 @@ impl Default for FormatSettings {
...
+/// Linter settings for the entire workspace
 #[derive(Debug)]
 pub struct LinterSettings {
     /// Enabled by default
     pub enabled: bool,
+
+    /// A list of global bindings that should be ignored by the analyzers
+    ///
+    /// If defined here, they should not emit diagnostics.
+    pub globals: Option<Vec<String>>,
```
MichaReiser, line 73: Is there a semantic difference between an empty vector and `None`? If not, then I would recommend removing the `Option` wrapping of the vector.

MichaReiser (commented):
Looks good. I do have a few questions (and I like @leops point on the suggestions). 

* Does Rome-classic support changing the warning level of a diagnostic? 
* 'error' is likely not the right default for all lint rules, especially rules for which no sound static analysis is possible due to JS's dynamic nature. How do you intend rules that, by default, should emit warnings?

crates/rome_service/src/settings.rs
ematipico: That should work I think, we can use `Vec::is_empty` to avoid the serialization

leops:
> I would like to evaluate the PR with this approach. We are always in time for changing it.

Absolutely, this is just something I thought about while reviewing this and wanted to raise a possible discussion about, but it's not an issue for merging this PR

> > 'error' is likely not the right default for all lint rules, especially rules for which no sound static analysis is possible due to JS's dynamic nature. How do you intend rules that, by default, should emit warnings?
> 
> That's a nice question, and I don't have an answer yet. The thing is, for now all rules emit warnings or errors and we don't have a more finer way of doing that per rule, not ahead of time at least. Each rule emits a diagnostic and we can't know ahead of time which type, unless we execute them.

This is something we can revisit in more detail once this configuration actually gets integrated with the analyzer, but it would be possible to make the severity on `RuleDiagnostic` optional and let it be filled with the severity associated with the `RuleConfiguration` for that rule by default, unless the rule specified an explicit severity for the diagnostic.

An alternative / complement to this would be to consider the `RuleConfiguration` enum as setting a "maximum level" at which a certain rule can emit diagnostics:
- if a rule is set to `RuleConfiguration::Error` and emits a diagnostic with a severity of `Error`, `Warning` or `Note` the diagnostic is emitted as-is
- if a rule is set to `RuleConfiguration::Warn` and emits a diagnostic with a severity of `Error` or `Warning` it gets emitted as a warning in both case, a severity of `Note` is still emitted as-is
- if a rule is set to `RuleConfiguration::Off` no diagnostic gets emitted

This way if a rule is not guaranteed to be sound it could emit a warning, and the diagnostic will still be emitted as a warning even if the rule is set to `RuleConfiguration::Error` (this is potentially very counterintuitive though)

ematipico:
> I agree on the philosophy of providing good defaults and avoiding any unnecessary options. The philosophy could even be extended to not have support for overriding the lint rule level by arguing that we pick the appropriate defaults.

True, this is a good point. This is something we wanted to do time ago: https://github.com/rome/tools/issues/1529

> I think it would be worth spending some time exploring configuration formats that support per-rule options to avoid a breaking change in the very near future OR at least go through the rules with planned and check if we can implement all of them without having per-rule option support.

This is part of #2741 , where we will evaluate rules that require a configuration and discuss them, and see how we can go from there. 

I don't mind breaking changes because we are still in `0.*`. It allows us to experiment, change things and whatnot without worrying too much. We can start the exploration after we release `0.8.0`.

ematipico:
@leops do you have any idea why the workflow for the codegen is failing? Can't understand why

MichaReiser (approved):

ematipico:
> > I don't mind breaking changes because we are still in 0.*. It allows us to experiment, change things and whatnot without worrying too much. We can start the exploration after we release 0.8.0.
> 
> That's true but comes at the risk that other things will be more important in the upcoming 0.9 and then 10.0 release and we then missed our chance. That's why I think it would be valuable to invest a couple of minutes to post a few options on how we could support rule-specific options. For example, one option that isn't a breaking change is to define the rule value as `string | { level: "warn" | "off", options: {} }` or maybe we find a format that avoids the need for a union type (what do linters of other languages use?)

I believe we can use a mixed approach using what clippy is doing, which by using a custom function to deserialize the rules. Clippy doesn't allow to turn off rules at configuration level though, you can only provide options to rules. Clippy uses macros to turn off rules in a global way.

I will spend some time and propose a solution

Issue #2835, opened 2022-07-06
☂️ Documentation changes 0.8.0


ematipico:
# Tentative date, 2th August 2022

- NEW: configuration file available for formatter
- NEW: `--max-diagnostics` for `rome check`
- NEW: `rome init` command
- NEW: configuration file available for linter
- NEW: renaming symbol feature available VSCode

# Nice to have features
- [x] rules divided per category (`js`, `jsx`, `react`, etc.);
- [x] add the version number to the documentation of rules, so we can have something like:
	```md
	# noUnusedVariable (since v0.8.0)
	```
- [x] semantic model to use inside lint rules;
- [x] limit diagnostics to 10 by default, have an argument for `rome check` to change the number of diagnostics: `--max-diagnostics`

PR #2891, opened 2022-07-18
chore(rome_analyzer): declare rules with version

ematipico:
## Summary

This adds once of the nice to have listed in #2835 

Proposed on discord: https://discord.com/channels/678763474494423051/978209399120216076/987343929231941673

The PR changes how the `declare_rule` is defined. Now the shape is different and it has a struct-like syntax:

```rs
declare_rule! {
    ///  Disallow labels that share a name with a variable
    pub(crate) NoLabelVar {
        version: "0.7.0",
        name: "noLabelVar"
    }
}
```

It also accepts another field, which is a possible deprecation reason. This will be useful in the future, when we want to ditch or change rules. It's not used anywhere yet, other than being part of the metadata:

```rs
declare_rule! {
    ///  Disallow labels that share a name with a variable
    pub(crate) NoLabelVar {
        version: "0.7.0",
        name: "noLabelVar",
		deprecated: "This rule wasn't safe, please use `noLalbelInVar`"
    }
}
```

The small downside of this approach is that the version is hardcoded, so the developer has to make sure to put the correct next version (usually the next minor). But, for now, I think it's a small price to pay compared on setting workflows and branches for multiple websites.

I also tweaked a bit the `documentation` cargo alias to include also `rome_js_formatter`, which will be deployed once the PR is merged.

I also moved the codegen of the configuration in another command called `cargo codegen configuration`. Having it in a separate command fixed an issue on CI, now some rules that were "missed" locally are now correctly added. I plan to add this command in the CI in another PR, to see if actually fixed the issue.

## Test Plan

Check the rendered result


crates/rome_analyze/src/rule.rs
```diff
@@ -76,11 +78,12 @@ pub trait RuleMeta {
 /// diagnostic in the resulting documentation page
 #[macro_export]
 macro_rules! declare_rule {
-    ( $( #[doc = $doc:literal] )+ $vis:vis $id:ident = $name:literal ) => {
+    ( $version:literal, $( #[doc = $doc:literal] )+ $vis:vis $id:ident = $name:literal ) => {
```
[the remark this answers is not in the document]
ematipico: I wanted to keep the changes to a minimum, but I suppose the suggestion makes sense. I am sure we will add more stuff in the future

crates/rome_js_analyze/src/analyzers/js/no_async_promise_executor.rs
```diff
@@ -4,6 +4,7 @@ use rome_js_syntax::{JsAnyExpression, JsAnyFunction, JsNewExpression, JsNewExpre
 use rome_rowan::{AstNode, AstSeparatedList};
 
 declare_rule! {
+    "0.7.0",
```
MichaReiser, line 7: The meaning of the "0.7.0" is only obvious to me as it is part of this PR. I think it would help to use a more explicit naming: 

```rust
declare_rule! {
  /// documentation
since_version: "0.7.0",
}
```

or 

```rust
declare_rule!{
	/// documentation
	since_version = 0.7.0,
}
```

crates/rome_analyze/src/rule.rs
```diff
@@ -11,6 +11,12 @@ use crate::registry::RuleLanguage;
 use crate::{AnalysisFilter, LanguageRoot, Phase, Phases, Queryable, RuleRegistry};
 
 pub trait RuleMeta {
+    /// When a rule is deprecated, a reason should be provided
+    const REASON: &'static str;
+    /// It marks if a rule is deprecated or not
+    const DEPRECATED: bool;
```
leops, line 17: Nit: why not make this an `Option<&'static str>` directly ?

leops (approved):

crates/rome_analyze/src/rule.rs
ematipico: I thought `Option` could not be a const, but good to know!

Issue #2878, opened 2022-07-14
noFunctionAssign


xunilrj:
See https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noFunctionAssign.md

PR #2895, opened 2022-07-19
feature(rome_js_analyze): no function reassign lint rule

xunilrj:
## Summary

Closes https://github.com/rome/tools/issues/2878

## Test Plan

```
> cargo rome-cli check crates/rome_js_analyze/tests/specs/js/noFunctionAssign.js 
```

### DIFF
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_function_assign.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_function_assign.rs
@@ -1,4 +1,4 @@
-use crate::{semantic_services::Semantic, JsRuleAction};
+use crate::semantic_services::Semantic;
 use rome_analyze::{context::RuleContext, declare_rule, Rule, RuleCategory, RuleDiagnostic};
 use rome_console::markup;
 use rome_js_semantic::{AllReferencesExtensions, Reference};
@@ -93,7 +93,10 @@ declare_rule! {
     ///     var foo = bar;
     /// }
     /// ```
-    pub(crate) NoFunctionAssign = "noFunctionAssign"
+    pub(crate) NoFunctionAssign {
+        version: "0.7.0",
+        name: "noFunctionAssign"
+    }
 }
 
 pub struct State {
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_function_assign.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_function_assign.rs
@@ -18,7 +18,9 @@ declare_rule! {
     /// ```
     ///
     /// ```js,expect_diagnostic
-    /// function foo() { foo = bar; }
+    /// function foo() {
+    ///     foo = bar;
+    ///  }
     /// ```
     ///
     /// ```js,expect_diagnostic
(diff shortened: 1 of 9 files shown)


crates/rome_js_analyze/src/semantic_analyzers/js/no_function_assign.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// var foo = function() {}; foo = bar;
+    /// ```
+    ///
+    /// ```js
+    /// var foo = function() { foo = bar; };
+    /// ```
+    ///
+    /// ```js
+    /// import bar from 'bar';
+    /// function foo() { var foo = bar; }
```
MichaReiser, line 74: Nit: I recommend to use our formatters formatting for the example scripts (maybe something we can add to the code gen in the future?) because I find it more difficult to read 

```
var foo = function () {}; foo = bar;
```

vs 
```
var foo = function () {} 
foo = bar;
```

crates/rome_js_analyze/src/semantic_analyzers/js/no_function_assign.rs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        let diag = diag.footer_note(markup! {"Use a local variable instead."});
+
+        Some(diag)
+    }
+
+    fn action(_: &RuleContext<Self>, _: &Self::State) -> Option<JsRuleAction> {
+        None
+    }
```
ematipico, line 149: This can be omitted, because it's basically the default implementation

crates/rome_js_analyze/tests/specs/js/noFunctionAssign.js.snap
```diff
@@ -0,0 +1,129 @@
...
+```
+
+```
+warning[js/noFunctionAssign]: Do not reassign a function declaration.
+  ┌─ noFunctionAssign.js:7:10
+  │
+6 │ foo3 = bar;
+  │ ---- Reassigned here.
+7 │ function foo3() { };
+  │          ----
```
ematipico, line 68: Is this correct? Because the `foo3` is not reassigned by `bar` but by `function foo3`. And the end value will be `bar`

ematipico (commented):
I left some concern about the diagnostics messaging and the correctness of the rule

crates/rome_js_analyze/tests/specs/js/noFunctionAssign.js.snap
xunilrj: I think this is correct because the function binding is not only hoisted but also initialized.

image

ematipico: Although your example is slightly different, you use `foo3` before assign it. In that case it makes sense. But the example in the test is different. Our example:

Firefox:

Screenshot 2022-07-19 at 11 27 34

Chrome: 
Screenshot 2022-07-19 at 11 27 55


In your example we should have a diagnostic on `foo3 = 1` saying that `foo3` was reassigned, because `function foo3` gets hoisted.

ematipico: Or, if we want to put in a different way, we should have a better diagnostic for these kind of cases. Because `function f()` gets hoisted. We can't just assume that people know what's going on

ematipico: Yeah, it helps a little bit more, I think!

ematipico (approved):

PR #2898, opened 2022-07-19
fix(rome_js_analyze/useTemplate):  comments in trivia when merge `Vec<JsAnyExpression>` into a `JsTemplate`

IWANABETHATGUY:
## Summary
1. Handle case comments in trivia when merging `Vec<JsAnyExpression>` into a `JsTemplate`
2. For more details, you could see https://github.com/rome/tools/pull/2803#discussion_r920826576

## Test Plan
1. cargo test

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/js/use_template.rs
+++ b/crates/rome_js_analyze/src/analyzers/js/use_template.rs
@@ -9,7 +9,7 @@ use rome_js_syntax::{
     JsAnyExpression, JsAnyLiteralExpression, JsBinaryExpression, JsBinaryOperator, JsSyntaxKind,
     JsSyntaxToken, JsTemplate, JsTemplateElementList, T,
 };
-use rome_rowan::{AstNode, AstNodeExt, AstNodeList, SyntaxToken};
+use rome_rowan::{AstNode, AstNodeExt, AstNodeList, SyntaxToken, TriviaPiece};
 
 use crate::{utils::escape_string, JsRuleAction};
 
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
!bench_analyzer

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -142,22 +142,54 @@ fn convert_expressions_to_js_template(exprs: &Vec<JsAnyExpression>) -> Option<Js
...
+                        .leading_trivia()
+                        .pieces()
+                        .skip_while(|item| item.is_newline() || item.is_whitespace())
+                        .map(|item| TriviaPiece::new(item.kind(), item.text_len()))
+                        .collect::<Vec<_>>(),
+                    first_token
+                        .trailing_trivia()
+                        .pieces()
+                        .map(|item| TriviaPiece::new(item.kind(), item.text_len())),
+                );
```
ematipico, line 169: Would it be possible to restructure the code in a more readable way? Something like this:

```rs
let next_first_token = {
	let token_kind = first_token.kind();
	// ..
	JsSyntaxToken::new_detached(token_kind, second, third, fourth)
};
```
The reason of the change is mostly because like this we assign things to variable with proper names, which will help to understand what we are computing.

crates/rome_js_analyze/src/analyzers/js/use_template.rs
```diff
@@ -142,22 +142,54 @@ fn convert_expressions_to_js_template(exprs: &Vec<JsAnyExpression>) -> Option<Js
...
+                next_last_token_trailing.reverse();
+                let next_last_token = JsSyntaxToken::new_detached(
+                    last_token.kind(),
+                    last_token.text().trim_end(),
+                    last_token
+                        .leading_trivia()
+                        .pieces()
+                        .map(|item| TriviaPiece::new(item.kind(), item.text_len())),
+                    next_last_token_trailing,
+                );
```
ematipico, line 190: Same here, let's restructure

ematipico (approved):
Nice fix! I left some nitpick, but it seems good to me

crates/rome_js_analyze/src/analyzers/js/use_template.rs
IWANABETHATGUY: Done

PR #2899, opened 2022-07-19
docs: move documentation inside crates

ematipico:
## Summary

This PR moves our internal documentation inside the rust docs.

I didn't do much other than doing copy and paste, and removing the reference to the old files.

## Test Plan

No plan


crates/rome_js_parser/src/lib.rs
```diff
@@ -55,6 +55,252 @@
...
+//!
+//!  p.expect_required(T![if]);
+//!  p.expect_required(T!['(']);
+//!  parse_any_expression(p).or_missing_with_error(p, js_parse_errors::expeced_if_statement);
+//!  p.expect_required(T![')']);
+//!  parse_block_statement(p).or_missing_with_error(p, js_parse_errors::expected_block_statement);
+//! // the else block is optional, so we mark it as "missing" in case it's absent
+//!  parse_else_clause(p).or_missing();
+//!
+//!  Present(m.complete(p, JS_IF_STATEMENT));
```
MichaReiser, line 158:
```suggestion
//!  p.expect(T![if]);
//!  p.expectT!['(']);
//!  parse_any_expression(p).or_add_diagnostic(p, js_parse_errors::expeced_if_statement);
//!  p.expectT![')']);
//!  parse_block_statement(p).or_add_diagnostic(p, js_parse_errors::expected_block_statement);
//! // the else block is optional, handle the marker by using `ok`
//!  parse_else_clause(p).ok();
//!
//!  Present(m.complete(p, JS_IF_STATEMENT));
```

crates/rome_js_parser/src/lib.rs
```diff
@@ -55,6 +55,252 @@
...
+//! }
+//! ```
+//!
+//! Hold on, what are these *missing* markers? Rome's AST facade uses fixed offsets to retrieve a particular child from a node.
+//! For example, the 3rd child of the if statement is the condition. However, the condition would become the second element
+//! if the opening parentheses `(` isn't present in the source text. That's where missing elements come into play.
+//! Missing elements (added by calling `p.missing()`) represent placeholders for syntax that isn't present in the source text to guarantee that the children always appear in the same order.
```
MichaReiser, line 165: I think we can delete this section since missing markers are now handled automatically

crates/rome_js_parser/src/lib.rs
```diff
@@ -55,6 +55,252 @@
...
+//!
+//! You can convert the `ConditionalParsedSyntax` to a regular `ParsedSyntax` by calling `or_invalid_to_unknown`, which wraps the whole parsed `with` statement in an `UNKNOWN` node if the parser is in strict mode and otherwise returns the unchanged `with` statement.
+//!
+//! What if there's no `UNKNOWN` node matching the node of your parse rule? You must then return the `ConditionalParsedSyntax` without making the `or_invalid_to_unknown` recovery. It's then up to the caller to recover the potentially invalid syntax.
+//!
+//!
+//! ## Summary
+//!
+//! * Parse rules are named `parse_rule_name`
+//! * The parse rules should return a `ParsedSyntax` or `ConditinalParsedSyntax`
```
MichaReiser, line 299:
```suggestion
//! * The parse rules should return a `ParsedSyntax` 
```

crates/rome_js_parser/src/lib.rs
```diff
@@ -55,6 +55,252 @@
...
+//! The start of the rule is the same as for any other rule. The exciting bits start with
+//!
+//! ```rust, ignore
+//! let conditional = StrictMode.excluding_syntax(p, with_stmt, |p, marker| {
+//!  p.err_builder("`with` statements are not allowed in strict mode")
+//!   .primary(marker.range(p), "")
+//! });
+//! ```
+//!
+//! The `StrictMode.excluding_syntax` creates a conditional syntax that is invalid in strict mode and adds an error if the parser is currently in strict mode.
```
MichaReiser, line 289:
```suggestion
//! The `StrictMode.excluding_syntax` converts the parsed syntax to an unknown node and uses the diagnostic builder to create a diagnostic if the feature is not supported.
```

PR #2901, opened 2022-07-19
fix(CI): broken format result

IWANABETHATGUY:
## Summary
1. Update `rome` to `next` instead of `next-0.7.0`, keep the same as our CI

## Test Plan
1. Pass all CI


ematipico (approved):
That's great! Thank you!!

PR #2902, opened 2022-07-19
refactor(rome_formatter): Pre-compute suppressions

MichaReiser:
## Summary

This PR changes how the formatter extracts formatter suppression comments. The existing implementation lazily traversed to a node's first token and it checks if the token has any leading suppression comments inside of `context.is_suppressed()`.

This PR iterates over the CST before starting the actual formatting to extract all suppressed nodes and it stores them in an index that can be queried during formatting.

This is mainly preparing for the migration to extract leading and trailing node comments with the re-work of the comments handling. Determining if a node is suppressed then becomes as simple as iterating over the leading comments of a node and determining if it has any suppression comment. 

This change also shows that the current implementation that calls `node.first_token` is roughly as expensive as iterating the whole tree ahead of time and computing the (suppression) comments. 

This change has the added benefit that suppressions are now a core formatter concept.

Part of #2768 

## Other changes
This PR further fixes a few places where the formatter didn't correctly check if a node has any suppression comment and if so, prevents it from changing the formatting.

### Why add suppressions to the `FormatContext`
The main reason for storing the suppressions on the context rather than on the `state` is that the context, unlike the `FormatState`, is language specific, and thus, it's possible to have a `Language` type parameter on the `Suppressions` type.

The one thing I dislike about adding it to the context is that the context is changed AFTER its creation.

## Performance

The performance of the two approaches is roughly the same.

### Future improvement
It would be possible to improve Rome's overall performance if the workspace provides a way to combine the visitors run for the linter and the visitor for computing the suppression comments.

## Test Plan

`cargo test`

### DIFF
--- a/crates/rome_formatter/src/comments.rs
+++ b/crates/rome_formatter/src/comments.rs
@@ -197,7 +197,9 @@ impl<L: Language> Comments<L> {
                         }
                     }
                 }
-                WalkEvent::Leave(SyntaxElement::Token(_)) => {}
+                WalkEvent::Leave(SyntaxElement::Token(_)) => {
+                    // Token already handled as part of the enter event.
+                }
             }
         }
 
--- a/crates/rome_formatter/src/comments.rs
+++ b/crates/rome_formatter/src/comments.rs
@@ -1,5 +1,12 @@
-use crate::CommentStyle;
-use rome_rowan::{Language, SyntaxTriviaPieceComments};
+use crate::FormatContext;
+use rome_rowan::{
+    Direction, Language, SyntaxElement, SyntaxKind, SyntaxNode, SyntaxTriviaPieceComments,
+    WalkEvent,
+};
+#[cfg(debug_assertions)]
+use std::cell::RefCell;
+use std::collections::HashSet;
+use std::rc::Rc;
 
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
 pub enum CommentKind {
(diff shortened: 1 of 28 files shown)


MichaReiser:
!bench_formatter

MichaReiser:
!bench_formatter

ematipico:
> This is mainly preparing for the migration to extract leading and trailing node comments with the re-work of the comments handling.

Can someone read somewhere what you plan to do? If this is a refactor per preparing something else, I think the reviewers should know what's coming next, so they have more context and everything is out in the open. We can add some comment in the original issue to explain what you have in mind.

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
+                            .filter_map(|piece| piece.as_comments())
+                        {
+                            if context.comment_style().is_suppression(comment.text()) {
+                                suppressed_nodes.insert(current_node);
+                                break;
+                            }
+                        }
+                    }
+                }
+                WalkEvent::Leave(SyntaxElement::Token(_)) => {}
```
ematipico, line 216: Could you add a comment explaining why this is a no-op?

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
+            suppressed_nodes,
+            #[cfg(debug_assertions)]
+            checked_suppressions: RefCell::default(),
+        }
+    }
+
+    /// Returns `true` if the passed `node` has a leading suppression comment.
+    ///
+    /// Suppression comments only apply if they at the start of a node and they suppress the most
+    /// outer node.
```
ematipico, line 230:
```suggestion
    /// Suppression comments only apply if they are at the start of a node and they suppress the most
    /// outer node.
```

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
+    ///
+    /// # Examples
+    ///
+    /// Returns `true` for the expression statement but `false` for the call expression because the
+    /// call expression is nested inside of the expression statement.
+    ///
+    /// ```javascript
+    /// // rome-ignore format: Reason
+    /// console.log("Test");
+    /// ```
```
ematipico, line 240: The snippet should go first, and then the explanation of the snippet.

crates/rome_formatter/src/lib.rs
```diff
@@ -729,14 +714,17 @@ where
 ///
 /// It returns a [Formatted] result, which the user can use to override a file.
 pub fn format_node<
-    Context: FormatContext,
+    Context: CommentContext<L>,
```
ematipico, line 717: This is food for documentation: why did we change the context instead of adding the new context?

MichaReiser:
> > This is mainly preparing for the migration to extract leading and trailing node comments with the re-work of the comments handling.
> 
> Can someone read somewhere what you plan to do? If this is a refactor per preparing something else, I think the reviewers should know what's coming next, so they have more context and everything is out in the open. We can add some comment in the original issue to explain what you have in mind.

The goal is to address #2768 by distinguishing between leading/node comments and dangling comments as described in [this notion](https://www.notion.so/rometools/Comment-Formatting-849c24813c90437ea4c30c1b595b2790) document.

crates/rome_formatter/src/lib.rs
```diff
@@ -1039,7 +1031,7 @@ where
 ///
 /// It returns a [Formatted] result
 pub fn format_sub_tree<
-    C: FormatContext,
+    C: FormatContext + CommentContext<L>,
```
ematipico, line 1034: This is food for documentation: why did we add the new context instead of replacing it?

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
 
     fn comment_style(&self) -> Self::Style;
+
+    /// Returns a ref counted comments. The use of a [Rc] is necessary so that [Comments] has a different
+    /// lifetime than the [crate::Formatter] to support the case where some formatter
+    /// iterates over all comments of a node and writes them to the formatter in the loop body.
+    fn comments(&self) -> Rc<Comments<L>>;
+
+    /// Consumes this context and returns a new context that uses the provided `comments` information.
+    fn with_comments(self, comments: Comments<L>) -> Self;
```
ematipico, line 130: I think `comments` should be `Rc<Comments<L>>`. 

Mostly because an implementation of this function could "forget" to do `Rc::new()` like you did inside the implementation of the JS formatter. In this way the type is the same of what we have in the function `comments`, and it's less loose.

ematipico (commented):
I have mostly nipicks

> This PR further fixes a few places where the formatter didn't correctly check if a node has any suppression comment and if so, prevents it from changing the formatting.

Would it be possible to add new tests to make we demonstrate the bugs you found?

crates/rome_formatter/src/comments.rs
MichaReiser: Not sure what I should explain here. There's nothing that needs to be done and the arm exists so that the match handle all variants.

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
+
+    /// Returns `true` if a token with the passed `kind` marks the start of a group. Common group tokens are:
+    /// * left parentheses: `(`, `[`, `{`
+    fn is_group_start_token(&self, kind: L::Kind) -> bool;
+
+    /// Returns `true` if a token with the passed `kind` marks the end of a group. Common group end tokens are:
+    /// * right parentheses: `)`, `]`, `}`
+    /// * end of statement token: `;`
+    /// * element separator: `,` or `.`.
+    /// * end of file token: `EOF`
```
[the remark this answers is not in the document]
MichaReiser: These are comments that I only moved and are no longer needed after the refactor.

A semicolon separates statements and a comma elements in array, in arguments, etc.. You can think of each statement or element as its own group where the semicolon / comma mark the "end" of the group. 

EOF marks the end of the file and it groups all preceding statements.

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
+    fn is_group_end_token(&self, kind: L::Kind) -> bool;
+}
+
+/// Type that stores the comments of a tree and gives access to:
+///
+/// * whether a node should be formatted as is because it has a leading suppression comment.
+/// * a node's leading and trailing comments
+/// * the dangling comments of a token
```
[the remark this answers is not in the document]
MichaReiser: I plan to write the module comments explaining leading/trailing and dangling comments once they get properly introduced in my next PR.

crates/rome_formatter/src/comments.rs
```diff
@@ -109,8 +116,177 @@ impl CommentKind {
...
+    /// * right parentheses: `)`, `]`, `}`
+    /// * end of statement token: `;`
+    /// * element separator: `,` or `.`.
+    /// * end of file token: `EOF`
+    fn is_group_end_token(&self, kind: L::Kind) -> bool;
+}
+
+/// Type that stores the comments of a tree and gives access to:
+///
+/// * whether a node should be formatted as is because it has a leading suppression comment.
```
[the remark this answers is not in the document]
MichaReiser: I think it's useful because any comment that isn't a leading comment but adheres to `rome ignore format:` isn't handled as a suppression comment.

crates/rome_formatter/src/lib.rs
MichaReiser: I hope it's clearer with the new name

crates/rome_formatter/src/comments.rs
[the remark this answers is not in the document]
ematipico: > A semicolon separates statements and a comma elements in array, in arguments, etc.. You can think of each statement or element as its own group where the semicolon / comma mark the "end" of the group.

Thank you for the explanation, here's a follow up question (if it's valid after the refactor): how does the grouping behaves when we have two statements where there aren't semicolons between them?

Example:

```js
console.log("something")
console.log("something")
```

crates/rome_formatter/src/comments.rs
ematipico: "The algorithm doesn't need to handle this event"?

crates/rome_formatter/src/comments.rs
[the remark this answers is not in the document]
MichaReiser: `is_group_end` is used to determine if a space should be inserted after a trailing comment, it doesn't perform any grouping itself:

```
/* a */;
        ^- No space here because `;` is a group end token

/* a */ console.log("something");
         ^- inserts a space to separate the comment from `console` 
```

crates/rome_formatter/src/comments.rs
MichaReiser: This isn't providing more information in my view because the match arm wouldn't be empty if the algorithm must handle that variant.

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -368,6 +368,14 @@ impl JsAnyAssignmentLike {
         match self {
             JsAnyAssignmentLike::JsPropertyObjectMember(property) => {
                 let name = property.name()?;
+
+                // It's safe to mark the name as checked here because it is at the beginning property
```
ematipico, line 372:
```suggestion
                // It's safe to mark the name as checked here because it is at the beginning of the property
```

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -379,7 +387,16 @@ impl JsAnyAssignmentLike {
                 Ok(false)
             }
             JsAnyAssignmentLike::JsObjectAssignmentPatternProperty(property) => {
-                let width = write_member_name(&property.member()?.into(), f)?;
+                let member_name = property.member()?;
+
+                // It's safe to mark the name as checked here because it is at the beginning property
```
ematipico, line 392:
```suggestion
                // It's safe to mark the name as checked here because it is at the beginning of  the property
```

PR #2906, opened 2022-07-20
fix(playground): potential xss attack

### LINKED ISSUE #2904: (rome_playground): XSS attack
### Environment information

```bash
Chrome
```

### What happened?

https://play.rome.tools/?code=%3Cscript%3Ealert%28%27test%27%29%3C%2Fscript%3E%3B%3Ca+href%3D%22gogole.com%22%3E%3C%2Fa%3E%3B&lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&typescript=false&jsx=true#PABzAGMAcgBpAHAAdAA+AGEAbABlAHIAdAAoACcAdABlAHMAdAAnACkAPAAvAHMAYwByAGkAcAB0AD4AOwA8AGEAIABoAHIAZQBmAD0AIgBnAG8AZwBvAGwAZQAuAGMAbwBtACIAPgA8AC8AYQA+ADsA

this code snippet could insert a `script` tag into our playground html.Also the diagnostic result is wrong.
image
HTML,

### Expected result

The dangerous tag should be escaped.
Diagnostic should display the right result.

### Code of Conduct

IWANABETHATGUY:
## Summary
1. Fix #2904 

## Test Plan
1. The diagnostic should display the right result, and escape the tag that could cause xss attack.

### DIFF
--- a/website/playground/src/lib.rs
+++ b/website/playground/src/lib.rs
@@ -173,13 +173,9 @@ pub fn run(
     let root_element = formatted.into_format_element();
     let formatter_ir = format!("{:#?}", root_element);
 
-    let mut diagnostic = String::new();
+    let mut html = HTML(Vec::new());
     for diag in parse.diagnostics() {
-        diagnostic.push_str(&diagnostic_to_string(
-            simple_files.name(main_file_id).unwrap(),
-            &code,
-            &diag,
-        ));
+        diagnostic_to_string(&simple_files, main_file_id, &diag, &mut html);
     }
 
     mark("rome::analyze::begin");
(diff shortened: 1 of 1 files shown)


website/playground/src/lib.rs
```diff
@@ -267,6 +220,23 @@ pub fn run(
...
+fn markup_to_string(markup: Markup) -> String {
+    let mut write = HTML(Vec::new());
+    let mut fmt = Formatter::new(&mut write);
+    fmt.write_markup(markup).unwrap();
+
+    String::from_utf8(write.0).unwrap()
+}
+
+fn diagnostic_to_string(name: &str, source: &str, diag: &Diagnostic) -> String {
+    let file = SimpleFile::new(name.into(), source.into());
```
leops, line 236: I think this recalculate line numbers for every diagnostic that gets printed, it would be better to use the existing values in `simple_files` instead

website/playground/src/lib.rs
```diff
@@ -223,11 +173,13 @@ pub fn run(
     let root_element = formatted.into_format_element();
     let formatter_ir = format!("{:#?}", root_element);
 
-    let mut errors = ErrorOutput(Vec::new());
+    let mut diagnostic = String::new();
```
leops, line 176: It would be more efficient to keep the `HTML<Vec<u8>>` object around and only call `String::from_utf8` at the end to only keep a single byte buffer around instead of allocating a new one for each diagnostic

leops (approved):

PR #2907, opened 2022-07-20
refactor(rome_service): move `globals`, refactor `rule` configuration

ematipico:
## Summary

This PR does four things

### `globals` field

This field has been moved inside the JavaScript language, it is now like this

```json
{
	"root": true,
  	"javascript": {
        "globals": ["$"],
    	"formatter": {
      		"quoteStyle": "double"
    	},
  	}
}
```

### `rules` settings

Now the rules can be configured in two different way. The old one and a new one, in case we will need to configure some rules:

```json
{
    "noLabelVar": {
        "level": "warn",
        "options": "test_option"
    },
    "useTemplate": {
        "level": "error",
        "options": [5]
    }
}
```

`options` is defined as a `Value` for now, which means that anything can be assigned to it. I don't have a clear idea how this field will be de-serialized/configured and documented inside our code generation, but at least this data structured gives us an edge where we don't need to apply breaking changes, because `options` is now a wildcard.

### `rome init` command

Now the command will turn off the `linter` by default, as we at our last meeting. I also added more text to the command. This text is exactly the same text that the Rome classic used to show. I didn't change anything, other than some link.

### Configuration validation

In my last PR I didn't add the attribute `deny_unknown_fields` to all the structs. I was under the impression that the attribute worked for all the `struct` once defined in the main one, but nope! It should work now

## Test Plan

Updated the existing cases to match the structure of the new configuration file.

Here's how the `rome init` is shown. Unfortunately the current `markup` macro is a bit difficult to tame when it comes to add tabbing and newlines. 

Screenshot 2022-07-20 at 13 55 19


crates/rome_service/src/configuration/linter/mod.rs
```diff
@@ -35,26 +29,43 @@ impl From<&LinterConfiguration> for LinterSettings {
...
+    Warn,
+    Error,
+    Off,
+}
+
+#[derive(Deserialize, Serialize, Debug, Eq, PartialEq, Clone)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct RuleWithOptions {
+    level: RulePlainConfiguration,
+    options: Value,
```
leops, line 70: Alternatively to `serde_json::Value` it's possible to use `Box<serde_json::value::RawValue>` (the difference is that `Value` gets deserialized into a dynamic representation while `RawValue` is just skipped over to be deserialized later)

crates/rome_service/src/configuration/mod.rs
```diff
@@ -53,6 +53,18 @@ impl Default for Configuration {
 }
 
 impl Configuration {
+    /// This function is used to generate configuration for the `rome init` command
+    pub fn new_init() -> Self {
```
leops, line 57: I'm not sure if the "init configuration" should diverge from what we consider to be the "default configuration", the fact that these two similar-sounding concepts may have difference values seems counterintuitive

ematipico: While this is true, we should consider also that `rome` can be used without using the configuration, which means that someone should be able to hack straightaway. This is also true in CI (and our CI), where we don't have a configuration file yet. So the real question is, should be Rome be used without configuration?

crates/rome_service/src/configuration/linter/mod.rs
ematipico: This is now done, although I had to change the test and remove the `options`. This is because serde couldn't serialize it while trying to match the expected content.

Another thing. `Eq` and `PartialEq` had to be removed from the structs because of this change.

crates/rome_service/src/configuration/javascript.rs
```diff
@@ -27,3 +31,13 @@ impl Default for PlainQuoteStyle {
...
 }
+
+#[derive(Default, Debug, Deserialize, Serialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase", default)]
+pub struct JavascriptLinter {
+    /// A list of global bindings that should be ignored by the analyzers
+    ///
+    /// If defined here, they should not emit diagnostics.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub globals: Vec<String>,
```
MichaReiser, line 42: This could be a `Set` as order doesn't matter and defining the same global twice has the same result.

(Order may matter if we intend to read the config, update it and then serialize it again, in which case we can use `IndexSet`)

crates/rome_service/src/configuration/mod.rs
MichaReiser: In my view, running:

```
rome init
rome check
```

or 

```
rome check
```

without a configuration


Should bove yield the same result.

MichaReiser (commented):
## globals

```
{
	"root": true,
  	"javascript": {
    	"formatter": {
      		"quoteStyle": "double"
    	},
    	"linter": {
        	"globals": ["$"]
    	}
  	}
}
```

* How do you intend to represent globals for other languages like HTML (global web component names), or CSS (global styles)? 
* I would move `globals` out of `linters` because knowing the globals is also something that is useful in the context of modification where globals should never be mangled. Tough, closure uses explicit [`extern` files](https://developers.google.com/closure/compiler/docs/externs-and-exports) that also provide type annotations for better results.

## Init command
What's the migration strategy if Rome's default change between two versions? Do users have to explicitly upgrade their configuration or should new defaults be applied automatically? 

Let's say we release the linter and enable it by default. Do now all Rome users have to enable the linter manually to make use of it?

crates/rome_service/src/configuration/javascript.rs
ematipico: > (Order may matter if we intend to read the config, update it and then serialize it again, in which case we can use IndexSet)

It does matter, the configuration has to be serialized too. I will use `IndexSet`

crates/rome_service/src/configuration/mod.rs
xunilrj: Agree with @MichaReiser

ematipico: Done.

ematipico:
@leops @MichaReiser 

The configuration of `rome init` command is now enabled by default. I added a `TODO` for adding the recommendation settings.

@MichaReiser 

> I would move globals out of linters 

Done. It makes total sense.

> How do you intend to represent globals for other languages like HTML (global web component names), or CSS (global styles)?

I am not sure I fully understand the question. But for CSS I would see something like `globals: ["--external-variable"]` and then our analyzer infrastructure will pick it up when running the rules.

> What's the migration strategy if Rome's default change between two versions? Do users have to explicitly upgrade their configuration or should new defaults be applied automatically?

> Let's say we release the linter and enable it by default. Do now all Rome users have to enable the linter manually to make use of it?

Migrations for configuration files was in plan for Rome classic. There are two ideas on the table, which can leave together. Having a `version` field, and when that is lower than the current one, we can run transformations to upgrade it to the new defaults.

The other idea is to map transformations of certain sections of the configuration from one version to another. For example, let's say we decided to rename `foo` from `bar`, and this change occurs at version `10.5.0`, then we create a migration for that field to rename itself. This requires semantic versioning and parsing/transforming of the JSON language.

leops (approved):

PR #2909, opened 2022-07-20
feat(rome_analyze): allow rules to dynamically emit local suppressions

leops:
## Summary

This PR adds a new `suppressed_nodes` method to the `Rule` trait. This hook allows lint rules to emit dynamic node suppressions for their own query, for instance to prevent the rule from matching again on a node that would be removed by a previously-emitted code action.

## Test Plan

I've modified the `UseTemplate` rule to use this new method instead of recursively checking for nested binary expressions, this rule has tests checking the rule does not trigger again on recursive string concatenation expressions

### DIFF
--- a/crates/rome_analyze/src/registry.rs
+++ b/crates/rome_analyze/src/registry.rs
@@ -1,4 +1,4 @@
-use std::{borrow, cell::RefCell, collections::BTreeMap, rc::Rc};
+use std::{borrow, collections::BTreeMap};
 
 use rome_rowan::{AstNode, Language, RawSyntaxKind, SyntaxKind, SyntaxNode};
 use rustc_hash::FxHashSet;
@@ -55,6 +55,8 @@ struct PhaseRules<L: Language> {
     /// lint rules associated with it
     ast_rules: Vec<SyntaxKindRules<L>>,
     control_flow: Vec<RegistryRule<L>>,
+    /// Holds a list of states for all the rules in this phase
+    rule_states: Vec<RuleState<L>>,
 }
 
 impl<L: Language + Default> RuleRegistry<L> {
@@ -73,7 +75,7 @@ impl<L: Language + Default> RuleRegistry<L> {
         let phase = R::phase() as usize;
         let phase = &mut self.phase_rules[phase];
 
-        let rule = RegistryRule::of::<G, R>();
+        let rule = RegistryRule::new::<G, R>(phase.rule_states.len());
 
         match <R::Query as Queryable>::KEY {
             QueryKey::Syntax(key) => {
@@ -93,13 +95,15 @@ impl<L: Language + Default> RuleRegistry<L> {
                     // Insert a handle to the rule `R` into the `SyntaxKindRules` entry
                     // corresponding to the SyntaxKind index
                     let node = &mut phase.ast_rules[index];
-                    node.rules.push(rule.clone());
+                    node.rules.push(rule);
                 }
             }
             QueryKey::ControlFlowGraph => {
                 phase.control_flow.push(rule);
             }
         }
+
+        phase.rule_states.push(RuleState::default());
 
         self.metadata.insert(
             MetadataKey {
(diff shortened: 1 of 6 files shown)


leops:
!bench_analyzer

crates/rome_analyze/src/registry.rs
```diff
@@ -224,23 +225,35 @@ pub struct RuleMetadata {
...
-// These need to be implemented manually because the implementations generated
-// by `derive(Copy, Clone)` would require `where: L: Copy + Clone: Copy + Clone`
-impl<L: Language> Copy for RegistryRule<L> {}
-impl<L: Language> Clone for RegistryRule<L> {
-    fn clone(&self) -> Self {
-        *self
+/// Internal state for a given rule
+#[derive(Clone, Default)]
+struct RuleState<L: Language> {
+    suppressions: Rc<RefCell<RuleSuppressions<L>>>,
```
MichaReiser, line 237: Would it be an alternative to using an `Rc` + `RefCell` to split `node.rules` into

* `node.rules`: Storing the rule + an index into the state array
* `states`: A `Vec` storing the states of all Rules. 

I don't understand the code well enough to tell if this isn't leading to any other borrow checker issues but may be an alternative to express that the state is shared and that `node.rules` is just a data structure to get faster rule lookup.

MichaReiser (approved):

crates/rome_analyze/src/registry.rs
ematipico, line 237: Could you explain me what `Rc` and `RefCell` are used for here?

crates/rome_analyze/src/rule.rs
```diff
@@ -239,6 +239,18 @@ pub trait Rule: RuleMeta {
         Self::diagnostic(ctx, state).map(|diag| diag.span())
     }
 
+    /// Allows the rule to suppress a set of syntax nodes to prevent them from
+    /// matching the `Query`. This is useful for rules that implement a code
+    /// action that recursively modifies multiple nodes at once, this hook
+    /// allows these rules to avoid matching on those nodes again.
+    fn suppressed_nodes(
```
ematipico, line 246: Could you add an example - code snippet - of how to use this new function into the documentation?

crates/rome_analyze/src/registry.rs
leops: In order to improve query matching performance, a given `Rule` may actually be represented as multiple `RegistryRule` withing a registry instance: for instance the `NoEmptyPattern` will have an entry matching `JsArrayBindingPattern` and another one matching `JsObjectBindingPattern` nodes.
Now although these are two distinct entries into the registry, they should share the same underlying `RuleState` since they belong to the same rule, and `Rc` allows that by implementing a reference-counted smart pointer managing a single instance of `RuleSuppressions` shared between multiple entries.
However since the state is now shared and the reference count is a dynamic runtime property, Rust cannot statically check how many outstanding references to the state exists and will not allow it to be mutated, so `RefCell` gets around that limitation by pushing this verification to the runtime, keeping a counter of outstanding mutable and immutable references and panicking if the code tries to do both at the same time.
Now as @MichaReiser mentioned [here](https://github.com/rome/tools/pull/2909#discussion_r926289959) it would also be possible to manually manage how the state is shared by storing a vector of `RuleState` in `PhaseRules`, and having each `RegistryRule` simply store an index into that vector instead of an explicit reference-counted pointer. This would allow the Rust compiler to statically check the mutable access to the state and may actually be easier to understand so I'll try implementing this approach instead

xunilrj:
I am wondering what would be the performance impact of not putting this in the "rule framework", but making it easy for rules to use ```RuleSuppressions``` directly.

We would be avoiding 99% of ```contains(...)``` on an empty set.

leops:
> We would be avoiding 99% of `contains(...)` on an empty set.

I expected this cost to be quite low since the implementation of `HashSet::contains` has a "fast path" and aborts immediately if the underlying hash table is empty, but it's incredibly easy to make incorrect assumptions about performance so I could be wrong about that ...

Issue #2903, opened 2022-07-20
useCamelCase


xunilrj:
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/useCamelCase.md

IWANABETHATGUY:
Should we close this issue since #2926 has been merged?

PR #2910, opened 2022-07-20
feat(rome_lsp): renaming symbols

xunilrj:
## Summary

This PR is the first step to implementing https://github.com/rome/tools/issues/2903.
It introduces an ```rename``` extension method to the ```BatchMutation```.

I still need to understand what to do when the cursor is not at a symbol.

## Test Plan

```
> cargo test -p rome_js_analyze -- rename
```

or directly using the lsp.

image

### DIFF
--- a/crates/rome_js_analyze/src/utils.rs
+++ b/crates/rome_js_analyze/src/utils.rs
@@ -2,30 +2,6 @@ pub mod rename;
 #[cfg(test)]
 pub mod tests;
 
-#[macro_export]
-macro_rules! assert_rename_ok {
-    ($(#[$attr:meta])* $($name:ident, $before:expr, $expected:expr,)*) => {
-        $(
-            #[test]
-            pub fn $name() {
-                $crate::utils::tests::assert_rename_ok($before, $expected);
-            }
-        )*
-    };
-}
-
-#[macro_export]
-macro_rules! assert_rename_nok {
-    ($(#[$attr:meta])* $($name:ident, $before:expr,)*) => {
-        $(
-            #[test]
-            pub fn $name() {
-                $crate::utils::tests::assert_rename_nok($before);
-            }
-        )*
-    };
-}
-
 #[derive(Debug, PartialEq)]
 pub(crate) enum EscapeError {
     EscapeAtEndOfString,
--- a/crates/rome_js_analyze/src/utils.rs
+++ b/crates/rome_js_analyze/src/utils.rs
@@ -27,7 +27,7 @@ macro_rules! assert_rename_nok {
 }
 
 #[derive(Debug, PartialEq)]
-pub enum EscapeError {
+pub(crate) enum EscapeError {
     EscapeAtEndOfString,
     InvalidEscapedChar(char),
 }
@@ -54,6 +54,6 @@ impl<'a> Iterator for InterpretEscapedString<'a> {
 
 /// unescape   
 ///
-pub fn escape_string(s: &str) -> Result<String, EscapeError> {
+pub(crate) fn escape_string(s: &str) -> Result<String, EscapeError> {
     (InterpretEscapedString { s: s.chars() }).collect()
 }
--- a/crates/rome_js_analyze/src/utils.rs
+++ b/crates/rome_js_analyze/src/utils.rs
@@ -1,3 +1,31 @@
+pub mod rename;
+#[cfg(test)]
+pub mod tests;
+
+#[macro_export]
+macro_rules! assert_rename_ok {
+    ($(#[$attr:meta])* $($name:ident, $before:expr, $expected:expr,)*) => {
+        $(
+            #[test]
+            pub fn $name() {
+                crate::utils::tests::assert_rename_ok($before, $expected);
+            }
+        )*
+    };
+}
+
+#[macro_export]
+macro_rules! assert_rename_nok {
+    ($(#[$attr:meta])* $($name:ident, $before:expr,)*) => {
+        $(
+            #[test]
+            pub fn $name() {
+                crate::utils::tests::assert_rename_nok($before);
+            }
+        )*
+    };
+}
+
 #[derive(Debug, PartialEq)]
 pub enum EscapeError {
     EscapeAtEndOfString,
--- a/crates/rome_js_analyze/src/utils.rs
+++ b/crates/rome_js_analyze/src/utils.rs
@@ -8,7 +8,7 @@ macro_rules! assert_rename_ok {
         $(
             #[test]
             pub fn $name() {
-                crate::utils::tests::assert_rename_ok($before, $expected);
+                $crate::utils::tests::assert_rename_ok($before, $expected);
             }
         )*
     };
@@ -20,7 +20,7 @@ macro_rules! assert_rename_nok {
         $(
             #[test]
             pub fn $name() {
-                crate::utils::tests::assert_rename_nok($before);
+                $crate::utils::tests::assert_rename_nok($before);
             }
         )*
     };
--- a/crates/rome_js_analyze/src/utils/rename.rs
+++ b/crates/rome_js_analyze/src/utils/rename.rs
@@ -5,81 +5,89 @@ use rome_js_syntax::{
 };
 use rome_rowan::{AstNode, BatchMutation, SyntaxNodeCast, TriviaPiece};
 
-pub trait CanBeRenamed {
+pub trait RenamableNode {
     fn declaration(&self, model: &SemanticModel) -> Option<JsSyntaxNode>;
 }
 
-impl CanBeRenamed for JsIdentifierBinding {
+impl RenamableNode for JsIdentifierBinding {
     fn declaration(&self, _: &SemanticModel) -> Option<JsSyntaxNode> {
         Some(self.syntax().clone())
     }
 }
 
-impl CanBeRenamed for JsReferenceIdentifier {
+impl RenamableNode for JsReferenceIdentifier {
     fn declaration(&self, model: &SemanticModel) -> Option<JsSyntaxNode> {
         Some(model.declaration(self)?.syntax().clone())
     }
 }
 
-impl CanBeRenamed for JsIdentifierAssignment {
+impl RenamableNode for JsIdentifierAssignment {
     fn declaration(&self, model: &SemanticModel) -> Option<JsSyntaxNode> {
         Some(model.declaration(self)?.syntax().clone())
     }
 }
 
-pub enum JsAnyCanBeRenamed {
+pub enum JsAnyRenamebleDeclaration {
     JsIdentifierBinding(JsIdentifierBinding),
     JsReferenceIdentifier(JsReferenceIdentifier),
     JsIdentifierAssignment(JsIdentifierAssignment),
 }
 
-impl CanBeRenamed for JsAnyCanBeRenamed {
+impl RenamableNode for JsAnyRenamebleDeclaration {
     fn declaration(&self, model: &SemanticModel) -> Option<JsSyntaxNode> {
         match self {
-            JsAnyCanBeRenamed::JsIdentifierBinding(node) => CanBeRenamed::declaration(node, model),
-            JsAnyCanBeRenamed::JsReferenceIdentifier(node) => {
-                CanBeRenamed::declaration(node, model)
+            JsAnyRenamebleDeclaration::JsIdentifierBinding(node) => {
+                RenamableNode::declaration(node, model)
             }
-            JsAnyCanBeRenamed::JsIdentifierAssignment(node) => {
-                CanBeRenamed::declaration(node, model)
+            JsAnyRenamebleDeclaration::JsReferenceIdentifier(node) => {
+                RenamableNode::declaration(node, model)
+            }
+            JsAnyRenamebleDeclaration::JsIdentifierAssignment(node) => {
+                RenamableNode::declaration(node, model)
             }
         }
     }
 }
 
 pub enum RenameError {
-    CannotBeRenamed,
+    CannotFindDeclaration,
+    CannotBeRenamed {
+        original_name: String,
+        new_name: String,
+    },
 }
 
-impl TryFrom<JsSyntaxNode> for JsAnyCanBeRenamed {
+impl TryFrom<JsSyntaxNode> for JsAnyRenamebleDeclaration {
     type Error = RenameError;
 
     fn try_from(node: JsSyntaxNode) -> Result<Self, Self::Error> {
         match node.kind() {
             JsSyntaxKind::JS_IDENTIFIER_BINDING => node
                 .cast::<JsIdentifierBinding>()
-                .map(JsAnyCanBeRenamed::JsIdentifierBinding)
-                .ok_or(Self::Error::CannotBeRenamed),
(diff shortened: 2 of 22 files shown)


IWANABETHATGUY:
Could our LSP rename handle the case that reference across the file?

xunilrj:
> Could our LSP rename handle the case that reference across the file?

Not yet.

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,210 @@
...
+        let mut changes = Vec::with_capacity(all_references.len());
+
+        for reference in all_references.iter() {
+            let scope = reference.scope();
+            if scope
+                .ancestors()
+                .filter_map(|scope| scope.get_binding(new_name))
+                .next()
+                .is_some()
+            {
```
MichaReiser, line 151: Unrelated to the PR itself but I was surprised to learn that `scope.get_binding(new_name)` only checks the current scope but not the parent scopes as well to retrieve any binding with that name. 

I think, ideally, `scope.get_binding(new_name)` should check the whole ancestor chain to see if there's any binding with the given name but `scope` can have a `scope.get_own_binding(new_name)` method that only checks the current scope. That would most closely match my expectations of `get_binding`.

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,210 @@
...
+            let prev_token = match reference.node().kind() {
+                JsSyntaxKind::JS_REFERENCE_IDENTIFIER => reference
+                    .node()
+                    .clone()
+                    .cast::<JsReferenceIdentifier>()
+                    .and_then(|node| node.value_token().ok()),
+                JsSyntaxKind::JS_IDENTIFIER_ASSIGNMENT => reference
+                    .node()
+                    .clone()
+                    .cast::<JsIdentifierAssignment>()
```
MichaReiser, line 164: This last clone should not be necessary if you use `all_references.into_iter()`

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,210 @@
...
+                _ => None,
+            };
+
+            if let Some(prev_token) = prev_token {
+                let next_token = token_with_new_text(&prev_token, new_name);
+                changes.push((prev_token, next_token));
+            }
+        }
+
+        // Now it is safe to pish changes to the batch mutation
```
MichaReiser, line 175: pish -> push?

crates/rome_js_analyze/src/utils.rs
```diff
@@ -1,5 +1,33 @@
+pub mod rename;
+#[cfg(test)]
+pub mod tests;
+
+#[macro_export]
```
MichaReiser, line 5: Is it possible to limit these test helpers to test builds only or is this something Rust doesn't support?

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -56,6 +56,18 @@ struct SemanticModelData {
 }
 
 impl SemanticModelData {
+    pub fn scope(&self, range: &TextRange) -> usize {
```
MichaReiser, line 59: Nit: No `pub` as this is an internal type?

crates/rome_rowan/src/ast/batch.rs
```diff
@@ -122,6 +122,24 @@ where
         });
     }
 
+    pub fn replace_token(&mut self, prev_token: SyntaxToken<L>, next_token: SyntaxToken<L>) {
```
MichaReiser, line 125: Can we add some documentation to these new public methods?

crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -260,3 +263,34 @@ fn format_on_type(
...
+    if let Some(node) = parse
+        .syntax()
+        .descendants_tokens(Direction::Next)
+        .find(|token| token.text_range().contains(symbol_at))
+        .and_then(|token| token.parent())
+    {
+        match node.try_into() {
+            Ok(node) => {
+                let mut batch = root.begin();
+                batch.rename_with_any_can_be_renamed(&model, node, &new_name);
```
MichaReiser, line 287: Should we return an error if the renaming fails in case `new_name` is already used?

IWANABETHATGUY: I found that vscode will not emit an error when the new name is already used.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: I have no problem with this approach, but as soon as we start adding more methods, we will need to keep consistency: "all_reads", "all_writes", "all_references" etc... will also navigate upwards. Which means we will end up with two sets of methods.

And in the end ```.ancestors().filter_map(...)``` solves the problem.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: I still need it because ```Reference::node(...)``` returns a borrow.
But I indeed can consume the ```all_references```.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: ops... and "pish" is not a nice word! :smile:

crates/rome_js_analyze/src/utils.rs
xunilrj: This was annoying me also. I think I managed to improve it.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_rowan/src/ast/batch.rs
xunilrj: Done.

crates/rome_service/src/file_handlers/javascript.rs
xunilrj: Yes, we should. Done.

IWANABETHATGUY:
I think we  should add an option like"rome.analysis.enableRenameSymbol": true,. Because in vscode javascript-language-service is built in which have more stable functionality on SymbolRename, I am not sure if there would be some side-effects when there are two language server providing SymbolRename service.

xunilrj:
> I think we should add an option like `"rome.analysis.enableRenameSymbol": true`. Because in vscode javascript-language-service is built in which have more stable functionality on SymbolRename, I am not sure if there would be some side-effects when there are two language server providing SymbolRename service.

I had to disable the built-in to test ours. I will research later if there is another option.

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,214 @@
+use rome_js_semantic::{AllReferencesExtensions, SemanticModel};
+use rome_js_syntax::{
+    JsIdentifierAssignment, JsIdentifierBinding, JsLanguage, JsReferenceIdentifier, JsSyntaxKind,
+    JsSyntaxNode, JsSyntaxToken,
+};
+use rome_rowan::{AstNode, BatchMutation, SyntaxNodeCast, TriviaPiece};
+
+pub trait CanBeRenamed {
+    fn declaration(&self, model: &SemanticModel) -> Option<JsSyntaxNode>;
+}
```
ematipico, line 10: I believe this trait should go in `rome_analyze` and make it generate via `L: Language`. This way we will have the rename feature for other languages too.

Also, let's make sure to add some documentation

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+impl CanBeRenamed for JsIdentifierAssignment {
+    fn declaration(&self, model: &SemanticModel) -> Option<JsSyntaxNode> {
+        Some(model.declaration(self)?.syntax().clone())
+    }
+}
+
+pub enum JsAnyCanBeRenamed {
```
ematipico, line 30: `JsAnyCanBeRenamed` => `JsAnyRenamebleDeclaration`

crates/rome_service/src/lib.rs
```diff
@@ -45,6 +46,8 @@ pub enum RomeError {
 
     /// Error thrown when validating the configuration. Once deserialized, further checks have to be done.
     Configuration(ConfigurationError),
+
+    RenameError(RenameError),
```
ematipico, line 50: Documentation

crates/rome_service/src/lib.rs
```diff
@@ -105,6 +109,11 @@ impl Display for RomeError {
             RomeError::DirtyWorkspace => {
                 write!(f, "Uncommitted changes in repository")
             }
+            RomeError::RenameError(error) => match error {
+                RenameError::CannotBeRenamed => {
+                    write!(f, "encountered an error while renaming a symbol",)
```
ematipico, line 114: This diagnostic could store the original name of the binding we tried to rename, so we can show it in the message.

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,214 @@
...
+            JsSyntaxKind::JS_IDENTIFIER_ASSIGNMENT => node
+                .cast::<JsIdentifierAssignment>()
+                .map(JsAnyCanBeRenamed::JsIdentifierAssignment)
+                .ok_or(Self::Error::CannotBeRenamed),
+            _ => Err(Self::Error::CannotBeRenamed),
+        }
+    }
+}
+
+pub trait RenameSymbolExtensions {
```
ematipico, line 76: Could you add some documentation of how this trait should be used in this crate?

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn rename(&mut self, model: &SemanticModel, node: impl CanBeRenamed, new_name: &str) -> bool;
+
+    fn rename_with_any_can_be_renamed(
+        &mut self,
+        model: &SemanticModel,
+        node: JsAnyCanBeRenamed,
```
ematipico, line 82: Do you think it's possible to have a `TryInto` instead? As we can we take advantage of the `TryFrom`

ematipico: Shouldn't we add a constrain like this `where node: CanBeRenamed`?

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -0,0 +1,214 @@
...
+    let trailing = token
+        .trailing_trivia()
+        .pieces()
+        .map(|item| TriviaPiece::new(item.kind(), item.text_len()))
+        .collect::<Vec<_>>();
+
+    JsSyntaxToken::new_detached(JsSyntaxKind::IDENT, new_text.as_str(), leading, trailing)
+}
+
+impl<N: AstNode<Language = JsLanguage>> RenameSymbolExtensions for BatchMutation<JsLanguage, N> {
```
ematipico, line 111: Not sure if this is permitted  with the current architecture, but it seems `RenameSymbolExtensions` can be made generic over `L: Language`

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -56,6 +56,18 @@ struct SemanticModelData {
...
 impl SemanticModelData {
+    fn scope(&self, range: &TextRange) -> usize {
+        let scopes = self
+            .scope_by_range
+            .find(range.start().into(), range.end().into());
+
+        match scopes.last() {
+            Some(interval) => interval.val,
+            // We always have at least one scope, the global one.
+            None => unreachable!(),
```
ematipico, line 67: I would move this comment inside the unreachable: 

```suggestion
            None => unreachable!("We always have at least one scope, the global one."),
```

crates/rome_rowan/src/syntax/trivia.rs
```diff
@@ -498,6 +498,10 @@ impl<L: Language> SyntaxTriviaPiece<L> {
     pub fn token(&self) -> SyntaxToken<L> {
         SyntaxToken::from(self.raw.token().clone())
     }
+
+    pub fn piece(&self) -> TriviaPiece {
```
ematipico, line 502: Documentation

crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -260,3 +263,38 @@ fn format_on_type(
     let printed = rome_js_formatter::format_sub_tree(context, &root_node)?;
     Ok(printed)
 }
+
+fn rename(
```
ematipico, line 267: `rename_declaration` or `rename_binding`. I thinks `rename` feels too generic. The suggestion tells us that the renaming occurs on a **single** binding/declaration

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: This method was created only to help type inference.
That is why I can do this without any explicit type. If I change to an "impl Trait" will be exactly like the other method.

https://github.com/rome/tools/pull/2910/files#diff-ef61b071e3781d8ac7027695d028fb3097b6a75c1363b93ff0be873b2777c5dbR284

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: The problem is that we will fall into the orphan rule. Trait at ```rome_analyze```, struct at ```rome_js_syntax```.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: Done.

crates/rome_service/src/lib.rs
xunilrj: Done.

crates/rome_service/src/lib.rs
xunilrj: Done.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: Done.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_rowan/src/syntax/trivia.rs
xunilrj: I will actually undo this method.

crates/rome_service/src/file_handlers/javascript.rs
xunilrj: the problem is that you can pass any ```RenamableNode```, not just declarations, or bindings. ```rename_node_declaration``` helps?

xunilrj: Done.

ematipico (approved):

xunilrj:
Given that this changed the LSP, I would include it in the change log of https://github.com/rome/tools/issues/2835

PR #2911, opened 2022-07-21
feat(rome_console): add an hyperlink markup element

leops:
## Summary

I originally left out the `Hyperlink` markup element from the initial implementation of `rome_console` since there wasn't really any use case for it at the time compare to the added complexity (requires support for properties in the `rome_markup` proc-macro and the `Markup` struct, and the implementation depends on the underlying platform)

I think there's now an interesting use case for it (turning linter error codes into links to the corresponding lint rule's documentation page), and the implementation should be easier to review in isolation

As part of this I changed the `code` property of diagnostics from a `String` to a `MarkupBuf` (and requiring `impl Display` instead of `impl Into<String>`). This required changing a few code paths that read out the code for a diagnostic, and raises the interesting question of whether having "inspectable codes" on diagnostics is something we actually want to support (in this case I would probably change the code back to a string and store the link in a secondary `code_link: Option<String>` on the `Diagnostic` struct)

## Test Plan

I added additional test cases for the `markup` macro checking both passing and failing syntax, and a test for the `Termcolor` ANSI emitter


ematipico:
> and raises the interesting question of whether having "inspectable codes" on diagnostics is something we actually want to support

Could you explain to me what it means, please? What can we do with this feature?

ematipico (approved):

crates/rome_console/src/write/termcolor.rs
```diff
@@ -74,18 +76,44 @@ where
...
 
+    let mut reset_link = false;
+    if let Some(href) = link {
+        // `is_synchronous` is used to check if the underlying writer
+        // is using the Windows Console API, that does not support ANSI
+        // escape codes. Generally this would only be true when running
+        // in the legacy `cmd.exe` terminal emulator, since int modern
+        // clients like the Windows Terminal ANSI is used instead
+        if writer.supports_color() && !writer.is_synchronous() {
+            write!(writer, "{ESC}]8;;{href}{ESC}\\")?;
```
xunilrj, line 106: In this case, I actually prefer "\x1b" directly. Easier to find. And everyone knows what it is.

crates/rome_js_analyze/src/lib.rs
```diff
@@ -110,10 +111,17 @@ mod tests {
         let mut error_ranges = Vec::new();
         analyze(0, &parsed.tree(), AnalysisFilter::default(), |signal| {
             if let Some(diag) = signal.diagnostic() {
-                let code = diag.code.as_deref().unwrap();
+                let code = diag.code.as_ref().unwrap();
                 let primary = diag.primary.as_ref().unwrap();
 
-                if code == "js/noDoubleEquals" {
+                let expect_code = markup! {
```
xunilrj, line 117: Why we have this check here?

leops:
> But I would be interested in hearing your thoughts on either of these approaches compared to changing the type of `code` on markup.

For now I've rolled back the change of `code` from string to markup, and added an additional `code_link` field to the diagnostics instead

PR #2913, opened 2022-07-21
docs: adds `rome_rowan` and the `rome_js_parser` to the deployed crates

ematipico:
## Summary

This PR adds `rome_rowan` and `rome_js_parser` to the uploaded crates for our technical documentation.

I fixed some warning but I am not sure if they are correct

## Test Plan

Merge and check the deployed result.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,7 +3,21 @@ lint = "clippy --workspace --all-targets --verbose -- --deny warnings"
 format = "fmt --all --verbose"
 codegen = "run -p xtask_codegen --"
 lintdoc = "run -p xtask_lintdoc --"
-documentation = "doc -p rome_*formatter -p rome_control_flow -p rome_analyze -p rome_*_syntax -p rome_*_factory -p rome_diagnostics -p rome_service -p rome_cli -p rome_console --no-deps"
+documentation = """
+    doc \
+    -p rome_*formatter \
+    -p rome_control_flow \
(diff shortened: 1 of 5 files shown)


.cargo/config.toml
```diff
@@ -3,7 +3,21 @@ lint = "clippy --workspace --all-targets --verbose -- --deny warnings"
 format = "fmt --all --verbose"
 codegen = "run -p xtask_codegen --"
 lintdoc = "run -p xtask_lintdoc --"
-documentation = "doc -p rome_*formatter -p rome_control_flow -p rome_analyze -p rome_*_syntax -p rome_*_factory -p rome_diagnostics -p rome_service -p rome_cli -p rome_console --no-deps"
+documentation = """
+    doc \
+    -p rome_*formatter \
```
MichaReiser, line 8: Any chance we can use `rome_*` or are there any rome crates that are excluded?

MichaReiser (approved):

.cargo/config.toml
ematipico: There are few that should be hidden: `lsp`, `text_edit`, `markup`

MichaReiser: What's the reason for hiding those?

ematipico: No particular reason, but if have to think about it:
- LSP is purely an implementation over a trait of another crate
- same for text edit, it doesn't add much more
- not sure about markup

PR #2915, opened 2022-07-21
feat(rome_analyze): refactor code actions to use batch mutations

leops:
## Summary

This PR refactors the `RuleAction` and `AnalyzerAction` to have lint rules provide code actions in the form of `BatchMutation` objects instead of returning a mutated syntax tree. In addition to possible improvements to performances unlocked by batching the mutations, this change allows the set of changes to be inspected in order to efficiently generate smaller text diffs that directly represent the mutation operations that were executed on the syntax tree using `rome_text_edit` (in fact for most use-cases code actions can directly be turned into a text-based diff operation without having to go through the mutation commit operation to build the mutated syntax tree)

In order to support this I had to extend the `BatchMutation` to support additional mutation operations that were used by existing lint rules, as well as replicating the "automatic trivia transfer" feature of the `replace_node` and `replace_token` method of the `AstNode` mutation extensions

## Test Plan

The lint rules test suite has been adapted to the new internal representation of code actions with no changes to the content of snapshots, from a user perspective all rules continue to emit the exact same code actions as before the change

### DIFF
--- a/crates/rome_analyze/src/signals.rs
+++ b/crates/rome_analyze/src/signals.rs
@@ -6,7 +6,6 @@ use rome_diagnostics::{
     Applicability, CodeSuggestion, Diagnostic, SuggestionChange, SuggestionStyle,
 };
 use rome_rowan::{BatchMutation, Language};
-use rome_text_edit::Indel;
 
 use crate::{
     categories::ActionCategory,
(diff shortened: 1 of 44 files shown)


leops:
!bench_analyzer

crates/rome_analyze/src/signals.rs
```diff
@@ -58,61 +56,52 @@ where
...
-            }
-        }
+impl<L> From<AnalyzerAction<L>> for CodeSuggestion
+where
+    L: Language,
+{
+    fn from(action: AnalyzerAction<L>) -> Self {
+        let indels = action.as_indels();
+
+        let range = indels.iter().fold(None, |state, indel| match state {
```
xunilrj, line 95: Makes sense to move this closer to the ```BatchMutation```? 

I can see this being helpful when renaming inside the LSP, for example.
Today I return the whole file, instead of the snippets that changed.

ematipico (approved):

PR #2924, opened 2022-07-25
feat(rome_service): support for recommendation of rules

### LINKED ISSUE #2912: enable recommendation setting for rules
Here's a list of recommended rules (up for discussion and change):

- `noArguments`
- `noAsyncPromiseExecutor`
- `noCatchAssign`
- `noCompareNegZero`
- `noDebugger`
- `noDelete`
- `noDoubleEquals`
- `noEmptyPattern`
- `noFunctionAssign`
- `noLabelVar`
- `noNegationElse`
- `noSparseArray`
- `noUnnecessaryContinue`
- `noUnsafeNegation`
- `noUnusedTemplateLiteral`
- `useBlockStatements`
- `useSimplifiedLogicExpression`
- `useSingleCaseStatement`
- `useSingleVarDeclarator`
- `useTemplate`
- `useValidTypeof`
- `useWhile`
- `noCommentText`
- `noImplicitBoolean`
- `useSelfClosingElements`
- `useShorthandArrayType`

ematipico:
## Summary

Closes #2912 

This PR enables a new feature in the configuration, which is the possibility to opt-in the recommended rules suggested by the Rome team.

### Configuration changes

It's possible to opt-in rules in two different way:

- Globally
```json
{
	"linter": {
		"recommended": true
	}
}
```

- Group level

```json
{
	"linter": {
		"rules": {
			"js" {
				"recommended": true
			},
			"jsx" {
				"recommended": false
			}
		}
	}
}
```

### De-serialization of rules

While we desarialize rules, we also validate that the rules we have in the configuration rules are valid for the current group. If not, we emit a `ConfigurationError`. We do so by saving an array of all the rules that belong to a certain group.

**The serialization is out of scope, because we don't have an use case yet**. But the serialization will need to be custom, because we want to keep the order of the original map. 

### Code generation

All the new code (configuration shape, de-serialization and methods) are code generated from the metadata of the rules. 

- all rules that belong to a group, are stored into an array 
- the recommended rules are also stored into a `const` inside the struct, and they represented as an array of `RuleFilter`. The name of the rule is taken from `Self::GROUP_RULES`, so we don't repeat the static strings
- the struct `Rules` has now a method called `as_analysis_filters` which will emit two `IndexSet` of enabled filters and disabled. The reason why `IndexSet` was used is because the enabled filters needs to be computed from the difference with the disabled filters. Here's an example:

```json
{
  "root": true,
  "linter": {
    "rules": {
        "recommended": true,
        "js": {
            "rules": {
                "noDebugger": "off"
            }
        }
    }
  }
}
```
In this example we need pull all the recommended rules, but then we have to remove the rules that have been turned off. `IndexSet` allows us to do so with a simple `.difference` function.

### Computation of `AnalysisFilter`

Here's there have been some limitations, due to how `RuleFilter` is designed (all references), and because of that the creation of the `AnalysisFilter` object must be done in place, right before being used. This caused a bit of repetition of code. If you have a solution to avoid this repetition of code, it would be great

## Test Plan

I updated some existing test and added new ones, some happy path and error path.

### DIFF
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -560,13 +560,9 @@ impl<'analysis> AnalysisFilter<'analysis> {
             })
     }
 
-    pub fn new_from_filters(
-        enabled_rules: Option<&'analysis [RuleFilter<'analysis>]>,
-        disabled_rules: Option<&'analysis [RuleFilter<'analysis>]>,
-    ) -> Self {
+    pub fn from_enabled_rules(enabled_rules: Option<&'analysis [RuleFilter<'analysis>]>) -> Self {
         Self {
             enabled_rules,
-            disabled_rules,
             ..AnalysisFilter::default()
         }
     }
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -506,7 +506,7 @@ type SuppressionParser = fn(&str) -> Vec<Option<&str>>;
 type SignalHandler<'a, L, Break> = &'a mut dyn FnMut(&dyn AnalyzerSignal<L>) -> ControlFlow<Break>;
 
 /// Allow filtering a single rule or group of rules by their names
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum RuleFilter<'a> {
     Group(&'a str),
     Rule(&'a str, &'a str),
@@ -540,7 +540,7 @@ pub struct AnalysisFilter<'a> {
     pub range: Option<TextRange>,
 }
 
-impl AnalysisFilter<'_> {
+impl<'analysis> AnalysisFilter<'analysis> {
     /// Return `true` if the rule `R` matches this filter
     pub fn match_rule<G, R>(&self) -> bool
     where
@@ -558,6 +558,17 @@ impl AnalysisFilter<'_> {
                     .iter()
                     .any(|filter| filter.match_rule::<G, R>())
             })
+    }
+
+    pub fn new_from_filters(
+        enabled_rules: Option<&'analysis [RuleFilter<'analysis>]>,
+        disabled_rules: Option<&'analysis [RuleFilter<'analysis>]>,
+    ) -> Self {
+        Self {
+            enabled_rules,
+            disabled_rules,
+            ..AnalysisFilter::default()
+        }
     }
 }
 
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -560,16 +560,10 @@ impl<'analysis> AnalysisFilter<'analysis> {
             })
     }
 
+    /// It creates a new filter with the set of [enabled rules](RuleFilter) passed as argument
     pub fn from_enabled_rules(enabled_rules: Option<&'analysis [RuleFilter<'analysis>]>) -> Self {
         Self {
             enabled_rules,
-            ..AnalysisFilter::default()
-        }
-    }
-
-    pub fn from_disabled_rules(disabled_rules: Option<&'analysis [RuleFilter<'analysis>]>) -> Self {
-        Self {
-            disabled_rules,
             ..AnalysisFilter::default()
         }
     }
--- a/crates/rome_analyze/src/lib.rs
+++ b/crates/rome_analyze/src/lib.rs
@@ -566,6 +566,13 @@ impl<'analysis> AnalysisFilter<'analysis> {
             ..AnalysisFilter::default()
         }
     }
+
+    pub fn from_disabled_rules(disabled_rules: Option<&'analysis [RuleFilter<'analysis>]>) -> Self {
+        Self {
+            disabled_rules,
+            ..AnalysisFilter::default()
+        }
+    }
 }
 
 /// Utility type to be used as a default value for the `B` generic type on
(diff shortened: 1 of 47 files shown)


leops (commented):
The code building up the filter could be simplified by only creating the `enabled_rules` list instead of both the enabled and disabled set (the information in `disabled_rules` is redundant since those rules could simply not be included in the enabled set)

ematipico:
> The code building up the filter could be simplified by only creating the `enabled_rules` list instead of both the enabled and disabled set (the information in `disabled_rules` is redundant since those rules could simply not be included in the enabled set)

I can change the code generation, but now I have a question: why do we have `disabled_filters` in the first place?

Also, don't you think that having the disabled rules would able more features?  An example that come up to my mind is the possibility to emit a warning when we find a suppression for a rule that was disabled in the configuration.

leops:
> I can change the code generation, but now I have a question: why do we have `disabled_filters` in the first place?

These two filters solve different problems, the idea was that `enabled_rules` can be used to enable only one or a few rules, while `disabled_rules` can be use to turn off a few rules while running all the others (in general the idea is to use the filter that would need a smaller number of entries since it's more efficient to check the rules against a small set)

> Also, don't you think that having the disabled rules would able more features? An example that come up to my mind is the possibility to emit a warning when we find a suppression for a rule that was disabled in the configuration.

This should already be the case, rules that are filtered out are not added to the `RuleRegistry` used for analysis, so a warning will be emitted if any comment tries to suppress a disabled rule since the analyzer considers that the rule does not exist

crates/rome_service/src/configuration/linter/rules.rs
```diff
@@ -1,100 +1,397 @@
...
 
-use crate::RuleConfiguration;
+use crate::{ConfigurationError, RomeError, RuleConfiguration};
+use indexmap::{IndexMap, IndexSet};
+use rome_analyze::RuleFilter;
 use serde::{Deserialize, Serialize};
 #[derive(Deserialize, Serialize, Debug, Clone)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct Rules {
+    #[doc = r" It enables a preset of rules of any group recommended by Rome. `true` by default."]
```
MichaReiser, line 10:
```suggestion
    #[doc = r" It enables the lint rules recommended by Rome. `true` by default."]
```

crates/rome_service/src/configuration/linter/rules.rs
```diff
@@ -1,100 +1,397 @@
...
+        &self,
+    ) -> (Option<IndexSet<RuleFilter>>, Option<IndexSet<RuleFilter>>) {
+        let mut enabled_rules = IndexSet::new();
+        let mut disabled_rules = IndexSet::new();
+        if self.is_recommended() {
+            enabled_rules.extend(Js::RECOMMENDED_RULES);
+            enabled_rules.extend(Jsx::RECOMMENDED_RULES);
+            enabled_rules.extend(Regex::RECOMMENDED_RULES);
+            enabled_rules.extend(Ts::RECOMMENDED_RULES);
+        }
```
MichaReiser, line 51: Do we have any validation of the configuration. Or what happens if the user sets `recomended: true` but also has `js: { recommended: false }`?

crates/rome_service/src/configuration/linter/rules.rs
```diff
@@ -1,100 +1,397 @@
...
+        if let Some(group) = self.js.as_ref() {
```
MichaReiser, line 52: I would prefer the term `language` or `language_group` over `group` or do you forsee other groups that aren't specific to language?

crates/rome_analyze/src/lib.rs
```diff
@@ -559,6 +559,20 @@ impl AnalysisFilter<'_> {
                     .any(|filter| filter.match_rule::<G, R>())
             })
     }
+
+    pub fn from_enabled_rules(enabled_rules: Option<&'analysis [RuleFilter<'analysis>]>) -> Self {
```
MichaReiser, line 563: Can you add some documentation why it is necessary to have `enabled` and `disabled` rules? The thing I'm concerned about the two filters is, what if `enabled` and `disabled` disagree about a particular rule. Is it then disabled or enabled?

crates/rome_service/src/configuration/linter/rules.rs
ematipico: Groups are not specific to a language. `a11y` is an example

crates/rome_service/src/configuration/linter/rules.rs
ematipico: Oh yeah, good point. I changed the code and added a test case. In your example we should enable all the recommended rules, exception for the recommended rules of the group `js`.

 > Do we have any validation of the configuration
 
 Only the one that belongs to serde during serialization/deserialization

crates/rome_analyze/src/lib.rs
ematipico: > what if enabled and disabled disagree about a particular rule. Is it then disabled or enabled?

This is more a question for @leops 

For now I reverted the change and limit myself only to `enabled_rules`. I will optimize it later

leops: The exact semantic of the `AnalysisFilter` is that a rule is included if it's present in the `enabled_rules` (or this field is set to `None`) and not present in `disabled_rules`, this means if a rule is present in both sets it will be disabled

crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -137,15 +126,29 @@ fn debug_print(_rome_path: &RomePath, parse: AnyParse) -> String {
     format!("{tree:#?}")
 }
 
-fn lint(rome_path: &RomePath, parse: AnyParse, categories: RuleCategories) -> Vec<Diagnostic> {
+fn lint(
+    rome_path: &RomePath,
+    parse: AnyParse,
+    categories: RuleCategories,
+    rules: Option<&Rules>,
```
leops, line 133: The `WorkspaceServer` should probably send in an `AnalysisFilter` object directly to avoid having too many arguments on this function

leops (approved):

PR #2925, opened 2022-07-25
release: 0.8.0

ematipico:
## Summary

Closes #2835

PR #2926, opened 2022-07-25
feat(rome_js_analyze): useCamelCase

xunilrj:
## Summary

Implements https://github.com/rome/tools/issues/2903.

## Test Plan

```
> cargo rome-cli check ./crates/rome_js_analyze/tests/specs/js/useCamelCase.js
```

### Camel Case

This PR implements only the happy path for camel case. It does not support nice camel case for abbreviations and acronyms.

### Renaming and conflicts

Given that we rename different bindings to possibly the same name, they can conflict. The renaming is correctly dealing with this. So if we have:

```js
let snake_case
snake_case = 1;
let _snake_case
console.log(_snake_case);
```

After renaming everything in one go, we have:

```js
let snakeCase
snakeCase = 1;
let snakeCase2
console.log(snakeCase2);
```

But this is not reflected in the diagnostic phase. Because there is no conflict there. This is something we may need to address in another PR.

```
warning[js/useCamelCase]: Prefer symbols names in camel case.
  ┌─ ./crates/rome_js_analyze/tests/specs/js/useCamelCase.js:1:5
  │
1 │ let snake_case;
  │     ----------
2 │ snake_case = 1;
  │ ---------- Used here.

Safe fix: Rename this symbol to camel case
    | @@ -1,5 +1,5 @@
0   | - let snake_case;
1   | - snake_case = 1;
  0 | + let snakeCase;
  1 | + snakeCase = 1;
2 2 |   let _snake_case;
3 3 |   console.log(_snake_case);
4 4 |   


warning[js/useCamelCase]: Prefer symbols names in camel case.
  ┌─ ./crates/rome_js_analyze/tests/specs/js/useCamelCase.js:3:5
  │
3 │ let _snake_case;
  │     -----------
4 │ console.log(_snake_case);
  │             ----------- Used here.

Safe fix: Rename this symbol to camel case
    | @@ -1,7 +1,7 @@
0 0 |   let snake_case;
1 1 |   snake_case = 1;
2   | - let _snake_case;
3   | - console.log(_snake_case);
  2 | + let snakeCase;
  3 | + console.log(snakeCase);
4 4 |   
5 5 |   function f(snake_case, PascalCase) {}
6 6 |  
```

### Performance

I compared this PR to https://github.com/SkylerLipthay/case. But focusing on being faster when everything is ok. Now we are only slower when we actually need to transform the string. We run fewer instructions, but unfortunately, we access the memory more times.

```
rome_all_lowercase
  Instructions:                 194 
  L1 Accesses:                  226
  L2 Accesses:                    4 
  RAM Accesses:                   7
  Estimated Cycles:             491

case_all_lowercase
  Instructions:                 613
  L1 Accesses:                  768
  L2 Accesses:                    5 
  RAM Accesses:                  12
  Estimated Cycles:            1213 

rome_already_camel_case
  Instructions:                 194 
  L1 Accesses:                  226 
  L2 Accesses:                    3 
  RAM Accesses:                   8 
  Estimated Cycles:             521 

case_already_camel_case
  Instructions:                 613 
  L1 Accesses:                  769 
  L2 Accesses:                    4 
  RAM Accesses:                  12 
  Estimated Cycles:            1209 

rome_pascal_case
  Instructions:                 574 
  L1 Accesses:                  767 
  L2 Accesses:                    7 
  RAM Accesses:                  25 
  Estimated Cycles:            1677 

case_pascal_case
  Instructions:                 613 
  L1 Accesses:                  771 
  L2 Accesses:                    3 
  RAM Accesses:                  11 
  Estimated Cycles:            1171 
```

### DIFF
--- a/crates/rome_js_analyze/Cargo.toml
+++ b/crates/rome_js_analyze/Cargo.toml
@@ -19,9 +19,15 @@ rome_console = { path = "../rome_console" }
 rome_diagnostics = { path = "../rome_diagnostics" }
 roaring = "0.9.0"
 rustc-hash = "1.1.0"
+iai = "0.1.1"
 
 [dev-dependencies]
 tests_macros = { path = "../tests_macros" }
 rome_text_edit = { path = "../rome_text_edit" }
 insta = { version = "1.10.0", features = ["glob"] }
 countme = { version = "3.0.0", features = ["enable"] }
+case = "1.0.0"
+
+[[bench]]
+name = "iai"
+harness = false
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
@@ -27,6 +27,9 @@ declare_rule! {
     ///
     /// ```js,expect_diagnostic
     /// let snake_case;
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
     /// let PascalCase;
     /// ```
     ///
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
@@ -1,8 +1,6 @@
-use std::borrow::Cow;
-
 use crate::{
     semantic_services::Semantic,
-    utils::{rename::RenameSymbolExtensions, to_camel_case, ToCamelCase},
+    utils::{rename::RenameSymbolExtensions, ToCamelCase},
     JsRuleAction,
 };
 use rome_analyze::{
@@ -11,12 +9,9 @@ use rome_analyze::{
 use rome_console::markup;
 use rome_diagnostics::Applicability;
 use rome_js_semantic::{AllReferencesExtensions, Reference};
-use rome_js_syntax::{
-    JsAnyExpression, JsAnyLiteralExpression, JsAnyRoot, JsFormalParameter, JsIdentifierBinding,
-    JsIdentifierExpression, JsLanguage, JsParameterList, JsStringLiteralExpression,
-    JsVariableDeclaration, JsVariableDeclarator, JsVariableDeclaratorList, JsVariableStatement,
-};
-use rome_rowan::{AstNode, AstSeparatedList, BatchMutation, BatchMutationExt, SyntaxNodeCast};
+use rome_js_syntax::{JsFormalParameter, JsIdentifierBinding, JsVariableDeclarator};
+use rome_rowan::{AstNode, BatchMutationExt};
+use std::borrow::Cow;
 
 declare_rule! {
     /// Enforce camel case naming convention.
@@ -90,6 +85,7 @@ impl Rule for UseCamelCase {
     }
 
     fn action(ctx: &RuleContext<Self>, state: &Self::State) -> Option<JsRuleAction> {
+        let model = ctx.model();
         let mut batch = ctx.root().begin();
 
         // Avoid renaming conflicts
@@ -100,7 +96,7 @@ impl Rule for UseCamelCase {
             } else {
                 format!("{}{}", state.0, suffix)
             };
-            if batch.rename_node_declaration(&ctx.model(), ctx.query().clone(), &new_name) {
+            if batch.rename_node_declaration(model, ctx.query().clone(), &new_name) {
                 return Some(JsRuleAction {
                     category: ActionCategory::Refactor,
                     applicability: Applicability::Always,
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
@@ -1,0 +1,112 @@
+use std::borrow::Cow;
+
+use crate::{
+    semantic_services::Semantic,
+    utils::{rename::RenameSymbolExtensions, to_camel_case, ToCamelCase},
+    JsRuleAction,
+};
+use rome_analyze::{
+    context::RuleContext, declare_rule, ActionCategory, Rule, RuleCategory, RuleDiagnostic,
+};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_semantic::{AllReferencesExtensions, Reference};
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyLiteralExpression, JsAnyRoot, JsFormalParameter, JsIdentifierBinding,
+    JsIdentifierExpression, JsLanguage, JsParameterList, JsStringLiteralExpression,
+    JsVariableDeclaration, JsVariableDeclarator, JsVariableDeclaratorList, JsVariableStatement,
+};
+use rome_rowan::{AstNode, AstSeparatedList, BatchMutation, BatchMutationExt, SyntaxNodeCast};
+
+declare_rule! {
+    /// Enforce camel case naming convention.
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// let snake_case;
+    /// let PascalCase;
+    /// ```
+    ///
+    /// ## Valid
+    ///
+    /// ```js
+    /// let camelCase;
+    /// ```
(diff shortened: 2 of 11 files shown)


xunilrj:
Relevant for: https://github.com/rome/tools/pull/2925

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -88,6 +88,36 @@ pub trait RenameSymbolExtensions {
...
+    ///
+    /// A usual usecase is to append a suffix to a variable name.
+    ///
+    /// ```ignore
+    /// let new_name = "new_name";
+    /// let candidates = (2..).map(|i| format!("{}{}", new_name, i).into());
+    /// let candidates = once(Cow::from(new_name)).chain(candidates);
+    /// batch.try_rename_node_declaration_until_success(model, node, candidates);
+    /// ```
+    fn try_rename_node_declaration_until_success<S, I>(
```
ematipico, line 102: I think using `try_*` should hint to return `Result`, at least that's what the convention seems to be.

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -88,6 +88,36 @@ pub trait RenameSymbolExtensions {
         new_name: &str,
     ) -> bool;
 
+    /// Succesively try to rename a symbol using an iterator
+    /// as source for candidate names.
```
ematipico, line 92: What "Successively try" mean in this case? These two words seem to conflict with each other. Maybe we can reword it:

```suggestion
    /// It tries to rename a symbol using an iterator
    /// as source for candidate names. It returns a [bool], which is
    /// the result of the operation. `false` means that the renaming didn't occur. 
```

crates/rome_js_analyze/src/utils/rename.rs
```diff
@@ -88,6 +88,36 @@ pub trait RenameSymbolExtensions {
...
+    ///
+    /// A usual usecase is to append a suffix to a variable name.
```
ematipico, line 94:
```suggestion
    /// A usual use case is to append a suffix to a variable name.
```

crates/rome_js_analyze/src/utils/rename.rs
ematipico: The name of the function seems different from what it actually does. If inside `candidates` we have more than one eligible for renaming, only the first gets renamed and the rest don't. I think we should review the logic or the name of the function.

crates/rome_js_analyze/tests/specs/js/useCamelCase.js.snap
```diff
@@ -0,0 +1,113 @@
...
+2 │ snake_case = 1;
+  │ ---------- Used here.
+
+Safe fix: Rename this symbol to camel case
+    | @@ -1,5 +1,5 @@
+0   | - let snake_case
+1   | - snake_case = 1;
+  0 | + let snakeCase
+  1 | + snakeCase = 1;
+2 2 |   let _snake_case
```
ematipico, line 40: Bindings that start with underscore should not emit diagnostics. I think this exception was added because it's common to add the underscore for local/unused variables.

Rome classic was taking this into consideration.

crates/rome_js_analyze/tests/specs/js/useCamelCase.js.snap
```diff
@@ -0,0 +1,113 @@
...
+```
+
+# Diagnostics
+```
+warning[js/useCamelCase]: Prefer symbols names in camel case.
+  ┌─ useCamelCase.js:1:5
+  │
+1 │ let snake_case
+  │     ----------
+2 │ snake_case = 1;
```
ematipico, line 31: I think we don't need to show where variable is used:
- it could be used in **many** places, and this could bloat the console with lot of info;
- it doesn't add any contextual information;

Rome classic diagnostic didn't add any second message: https://github.com/rome/tools/blob/archived-js/internal/diagnostics/descriptions/lint.ts#L759-L762

crates/rome_js_analyze/src/utils.rs
```diff
@@ -33,3 +35,130 @@ impl<'a> Iterator for InterpretEscapedString<'a> {
...
+        to_camel_case(self)
+    }
+}
+
+/// Return the camel case form of the input parameter.
+/// If it is already in camel case, nothing is done.
+///
+/// This method do not address abbreviations and acronyms.
+pub fn to_camel_case(input: &str) -> Cow<str> {
+    pub enum ForceNext {
```
IWANABETHATGUY, line 58: This enum does not need to modify with `pub`, which is only used for internal function

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
```diff
@@ -0,0 +1,104 @@
...
+        version: "0.8.0",
+        name: "useCamelCase"
+    }
+}
+
+impl Rule for UseCamelCase {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Semantic<JsIdentifierBinding>;
+    type State = (String, Vec<Reference>);
```
IWANABETHATGUY, line 46: Would it be better to defer this reference collection action to the `action` phase ? This should be a little more performant

crates/rome_js_analyze/Cargo.toml
```diff
@@ -19,8 +19,14 @@ rome_console = { path = "../rome_console" }
 rome_diagnostics = { path = "../rome_diagnostics" }
 roaring = "0.9.0"
 rustc-hash = "1.1.0"
+iai = "0.1.1"
```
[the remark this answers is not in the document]
xunilrj: Yes. Wrong place.

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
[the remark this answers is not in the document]
xunilrj: Changed for a struct.

xunilrj: Done.

crates/rome_js_analyze/src/utils.rs
xunilrj: Done.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: I renamed the method and changed the comments.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: Done.

crates/rome_js_analyze/src/utils/rename.rs
xunilrj: I renamed this function.

crates/rome_js_analyze/tests/specs/js/useCamelCase.js.snap
xunilrj: Done.

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
```diff
@@ -0,0 +1,108 @@
...
+    ///
+    /// ## Valid
+    ///
+    /// ```js
+    /// let camelCase;
+    /// ```
+    pub(crate) UseCamelCase {
+        version: "0.8.0",
+        name: "useCamelCase",
+        recommended: true
```
ematipico, line 39: What do you think if we *don't* recommend the rule? Mostly because it's not finished as you suggested. But I don't mind having it recommended too. No strong opinions, as long as the current implementation is reliable.

crates/rome_js_analyze/tests/specs/js/useCamelCase.js.snap
xunilrj: Done.

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
xunilrj: Done.

leops (approved):

IWANABETHATGUY:
!bench_analyzer

PR #2928, opened 2022-07-25
feat(rome_workspace): ensure all parameters and result structs on the workspace interface are serializable

leops:
## Summary

This PR adds a new `serde_workspace` feature to `rome_service` that automatically derives the `Serialize` and `Deserialize` traits from `serde` on all the types that transit through the `Workspace` interface (including types declared in other crates and transitively included in this interface). This feature is currently disabled by default but prepares the way for the workspace potentially being callable across IPC or FFI boundaries, and in the meantime still imposes some limitations on how data can transit across the `Workspace` trait that forces it to work across languages (code actions for instance are now represented as language-agnostic text diffs, instead of language-specific syntax trees)

## Test Plan

This code is currently untested since the feature is disabled by default, however since it all relies on the procedural derive macros from `serde` simply running `cargo build -p rome_service --features serde_workspace` is enough to verify that everything compiles correctly, maybe this command could be added to the suite of tests being run on GitHub Actions ?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1352,6 +1352,7 @@ dependencies = [
  "rome_js_syntax",
  "rome_rowan",
  "rustc-hash",
+ "serde",
 ]
 
 [[package]]
@@ -1383,6 +1384,7 @@ dependencies = [
  "atty",
  "lazy_static",
  "rome_markup",
+ "serde",
  "similar",
  "termcolor",
  "text-size",
@@ -1439,6 +1441,7 @@ dependencies = [
  "cfg-if",
  "indexmap",
  "rome_rowan",
+ "serde",
  "tracing",
 ]
 
@@ -1450,6 +1453,7 @@ dependencies = [
  "indexmap",
  "parking_lot",
  "rayon",
+ "serde",
  "tracing",
 ]
 
@@ -1665,6 +1669,7 @@ dependencies = [
 name = "rome_text_edit"
 version = "0.0.0"
 dependencies = [
+ "serde",
  "text-size",
 ]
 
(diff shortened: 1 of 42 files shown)


ematipico:
Re: test plan

I presume that once we start using the new feature internally, we should start seeing its benefits/repercussions?

MichaReiser (approved):
Looks good. 

Does that mean that we limit ourself to a `Workspace` facade that only uses serializable types? Do you see this as a limiting factor e.g. in a long-running CLI service that isn't restricted by the fact that the returned result types must be serializable? Would it be an option to have a specific `Workspace` facade for consumers that require a serializable/message based interface or is this something you can still see us do in the future/

leops:
> I presume that once we start using the new feature internally, we should start seeing its benefits/repercussions?

Besides laying out the groundwork for future work on a client-server system with a shared daemon process, one of the immediate application of this change could be in the JavaScript or WebAssembly bindings. Since this change transitively ensures that most of the "public API" types are serializable it opens the possibility of using serde for transmitting objects across FFI boundaries

> Does that mean that we limit ourself to a `Workspace` facade that only uses serializable types?

Yes that's the general intent for the moment, obviously a lot of concepts for our internal client-server model still need to be fleshed out so the idea I'm currently going with is to have this interface as constrained as possible to facilitate future experimentation.
For instance even in the context of a long running CLI process like monitoring a running bundler development server, I think it could still be beneficial to use a Workspace instance shared with a Language Server process, so that both the editor experience and bundler execution can benefit from hot caches primed by the other process.
In general since the Workspace is a high-level, language-agnostic abstraction most of the data going through this interface is going to be strings intended to be directly displayed to the user, either in the form of console messages or editor UI annotations, so requiring everything to be serializable doesn't seem like a constraint that will be hard to keep

ematipico (approved):

PR #2929, opened 2022-07-26
fix(rome_cli): better printing of message

ematipico:
## Summary

This PR fixes the printing of a message and it sets the limit of printable diagnostics to a higher limit

## Test Plan

Screenshot 2022-07-26 at 08 08 23

### DIFF
--- a/crates/rome_cli/src/commands/check.rs
+++ b/crates/rome_cli/src/commands/check.rs
@@ -13,7 +13,7 @@ pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
     let configuration = load_config(&session.app.fs)?;
     let mut workspace_settings = WorkspaceSettings::default();
 
-    let max_diagnostics: Option<u8> = session
+    let max_diagnostics: Option<u16> = session
         .args
         .opt_value_from_str("--max-diagnostics")
         .map_err(|source| Termination::ParseError {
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
https://github.com/rome/tools/pull/2929/files#diff-fb912ee1fd0e24af6fd125034ed4321c2d09d098bc73a8b93bb5895dbe2fc639R27 
You should change this line to 200 as well

IWANABETHATGUY:
What if I want to show all the diagnostics? How could I do that?

Boshen:
Does this fixes the problem where it keeps printing "The number of diagnostics exceeds the number allowed by Rome."? The actual diagnostics will get cleared away. I'm expecting the message to only display once.

ematipico:
> Does this fixes the problem where it keeps printing "The number of diagnostics exceeds the number allowed by Rome."? The actual diagnostics will get cleared away. I'm expecting the message to only display once.

I don't know, it would be great to be able to reproduce the issue somehow

leops:
> > Does this fixes the problem where it keeps printing "The number of diagnostics exceeds the number allowed by Rome."? The actual diagnostics will get cleared away. I'm expecting the message to only display once.
> 
> I don't know, it would be great to be able to reproduce the issue somehow

My guess is that this issue is cause by the line that prints the "diagnostics not shown" message being inside of the receiver loop, this will cause it to be printed again for every file that gets checked

> What if I want to show all the diagnostics? How could I do that?

I think at the moment this isn't allowed since we haven't seen any value to showing more than 200 diagnostics, it just seems like a lot of noise. Of course if there's a compelling enough use case to include this feature we could consider it

leops (approved):

PR #2933, opened 2022-07-26
fix(rome_js_syntax): Don't panic on unclosed block comment

MichaReiser:
This fixes an issue where Rome panicked when parsing suppression comments from a block comment without the closing `*/`.

This PR gracefully handles the case where a block comment has no closing `*/` by trying to remove a trailing `*` or `/` at the end of the comment and otherwise take the comment as is.

## Tests

I verified that running rome over rome classic no longer panics and added a unit test to the suppression comment parsing.

### DIFF
--- a/crates/rome_js_syntax/src/suppression.rs
+++ b/crates/rome_js_syntax/src/suppression.rs
@@ -166,6 +166,32 @@ mod tests {
             }],
         );
     }
+    #[test]
+    fn parse_unclosed_block_comment_suppressions() {
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


leops:
The tests for language-specific suppression comments are embedded directly in `rome_js_analyze`: https://github.com/rome/tools/blob/0e7aaa505470b32c6d7766e8352accd1d5af228b/crates/rome_js_analyze/src/lib.rs#L86

ematipico (approved):

leops (approved):

Issue #2932, opened 2022-07-26
🐛 Format On Paste Timing Issue


leops:
### Environment information

```bash
VSCode 1.69.2
Rome LSP Debug Build (ba8d17d)
```

### What happened?

1. Enable "Format On Paste" in the VSCode settings
2. Create or open a file in a language Rome supports for range formatting
3. Paste a block of code at the end or near the end of the file

The Rome LSP server will panic in `LineIndex::offset` or `SyntaxNode::token_at_offset` depending on where the text was pasted

This seems to be caused by VSCode sending the `textDocument/rangeFormatting` request for the Format On Paste feature before the `textDocument/didChange` notification for the newly pasted text is dispatched. This can be observed by setting the `rome_lsp.trace.server` setting of the extension to `messages`, after the paste operation the `Rome Trace` output channel will contain these events:

```text
[Trace - 15:55:47] Sending request 'textDocument/rangeFormatting - (5)'.
[Trace - 15:55:47] Sending notification 'textDocument/didChange'.
```

### Expected result

This is possibly a timing issue that needs to be resolved in the editor or the `vscode-languageclient` package, I opened an issue for it here: https://github.com/microsoft/vscode-languageserver-node/issues/1042
In any case the language server process should not crash when provided with invalid input ranges.

### Code of Conduct

- [X] I agree to follow Rome's Code of Conduct

leops:
Fixed by #2936 and #2935

PR #2935, opened 2022-07-26
fix(rome_lsp): fix panics on out of bounds formatting range

leops:
## Summary

This is a partial fix for #2932, it doesn't fix the root cause and format-on-paste will still cause range formatting errors but at least it will not crash the language server process. I've added additional checks at two levels: one is at the start of the `format_range` function in the formatter itself, while the other is in the `LineIndex` struct that's used to translate line + column positions into text offsets.

I've also made the `LineIndex` type crate-private, this makes it easier to detect unused code in order to review all possible "unsafe" vector indexing operation this type is doing ("unsafe" as "out of bounds index triggers a panic", there's no actual unsafety in the Rust language sense of the term)

Finally while I was debugging the issue I've added additional tracing instrumentation and filters to make the console output of the language server more useful, since that generally improves the observability of the code and doesn't impact release builds I decided to just leave it in.

## Test Plan
I've added a new test for `LineIndex` verifying it returns `None` if the provided line number is out of bounds, and a test for the `format_range` wrapper in `rome_js_formatter` checking it returns an error if the range does not fit the provided root node.

### DIFF
--- a/crates/rome_lsp/Cargo.toml
+++ b/crates/rome_lsp/Cargo.toml
@@ -25,7 +25,7 @@ tower-lsp = { version = "0.17.0"}
 tokio = { version = "1.15.0", features = ["full" ] }
 tracing = { version = "0.1.31", default-features = false, features = ["std", "max_level_trace", "release_max_level_warn"] }
 tracing-tree = "0.2.0"
-tracing-subscriber = "0.3.5"
+tracing-subscriber = { version = "0.3.5", features = ["env-filter"] }
 parking_lot = "0.12.0"
 futures = "0.3"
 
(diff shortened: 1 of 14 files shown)


crates/rome_lsp/Cargo.toml
```diff
@@ -25,7 +25,7 @@ tower-lsp = { version = "0.17.0"}
 tokio = { version = "1.15.0", features = ["full" ] }
 tracing = { version = "0.1.31", default-features = false, features = ["std", "max_level_trace", "release_max_level_warn"] }
 tracing-tree = "0.2.0"
-tracing-subscriber = "0.3.5"
+tracing-subscriber = { version = "0.3.5", features = ["env-filter"] }
```
ematipico, line 28: Does this remove all the jargon coming from the traces of other crates?

ematipico (approved):

crates/rome_lsp/Cargo.toml
leops: Yes it's used to set a filter here: https://github.com/rome/tools/blob/4ac2cacb38d37271c848cb3602805323243fc44b/crates/rome_lsp/src/main.rs#L11-L14
As explained in the comments this only prints info-level tracing spans and events for external crates, and only print trace-level items for a few internal crates

Issue #1042, opened 2020-08-12
🐛 UNLICENSED license not recognized


nstepien:
I work on an internal package at my job, we publish it as a private npm package, it's not public/open-source so it has "UNLICENSED" in the "license" field. Rome does not consider it a valid value.

From the npm docs:
https://docs.npmjs.com/files/package.json

>Finally, if you do not wish to grant others the right to use a private or unpublished package under any terms:
>```json
>{ "license": "UNLICENSED" }
>```

## Environment

```
$ rome rage --summary
Environment Variables
  Object {
    COLORTERM: "truecolor"
    LANG: "en_US.UTF-8"
    TERM_PROGRAM: "vscode"
  }

User Config
  unset

Rome Version
  10.0.4-beta

Node Version
  14.8.0

Platform
  win32 x64 10.0.19041

Terminal Features
  Object {
    background: "unknown"
    colorDepth: 24
    columns: 187
    cursor: true
    hyperlinks: true
    isTTY: true
    unicode: false
  }

Client Flags
  Object {
    clientName: "cli"
    cwd: "D:\\xxx"
    realCwd: "D:\\xxx"
    silent: false
  }

Server Status
  Object {
    projects: Array [
      Object {id: 0}
      Object {id: 1}
    ]
    server: Object {
      heapTotal: 27_451_392
      pid: 19_348
      uptime: 0.1582676
    }
    workers: Array [
      Object {
        astCacheSize: 0
        heapTotal: 27_451_392
        ownedBytes: 0
        ownedFileCount: 0
        pid: 19_348
        uptime: 0.1582196
      }
    ]
  }
```

## Steps To Reproduce

1. Have `{ "license": "UNLICENSED" }` in your package.json
2. Run `npx rome init --apply`

## Expected Results

`UNLICENSED` should be valid.

nstepien:
NVM I misread the Rome error report.

PR #2936, opened 2022-07-26
chore(vscode): upgrade dependencies

ematipico:


leops (approved):
I checked this out locally and it seems like updating `vscode-languageserver-node` fixes the root cause for #1042

PR #2939, opened 2022-07-27
refactor(vscode): remove beta formatter options

### LINKED ISSUE #2938: remove formatter configuration from VSCode

ematipico:
## Summary

Closes #2938 .

Now that we have a configuration file, users should use that as source to change Rome's defaults. The options in the VSCode were always a beta, with the idea that they were going to be removed once the configuration file was available.

Now it's the time to remove them.

I enabled the serialization of `format_with_errors` and assigned its value to the `WorkspaceSettings`. This field it hasn't been used yet in the CLI. I will make a further PR to enable it.

## Test Plan

Current tests should pass. I manually built the VSCode extension and verified that the options don't exist anymore.

### DIFF
--- a/crates/rome_lsp/Cargo.toml
+++ b/crates/rome_lsp/Cargo.toml
@@ -13,7 +13,7 @@ serde = { version = "1.0.133", features = ["derive"] }
 anyhow = "1.0.52"
 indexmap = "1.8.0"
 rome_fs = {  path = "../rome_fs" }
-rome_service = { path = "../rome_service", features = ["serde_workspace"] }
+rome_service = { path = "../rome_service" }
 rome_js_formatter = { path = "../rome_js_formatter" }
 rome_formatter = { path = "../rome_formatter" }
 rome_analyze = { path = "../rome_analyze" }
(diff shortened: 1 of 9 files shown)


leops (commented):
I think we could also remove the `rome.analysis.enableDiagnostics` and `rome.analysis.enableCodeActions` now that the linter can also be enabled or disabled in the configuration file

crates/rome_lsp/Cargo.toml
```diff
@@ -13,7 +13,7 @@ serde = { version = "1.0.133", features = ["derive"] }
 anyhow = "1.0.52"
 indexmap = "1.8.0"
 rome_fs = {  path = "../rome_fs" }
-rome_service = { path = "../rome_service" }
+rome_service = { path = "../rome_service", features = ["serde_workspace"] }
```
leops, line 16: Is this feature really required now ? It doesn't seem like any type from the workspace is getting serialized

leops (approved):

crates/rome_lsp/Cargo.toml
ematipico: Thank you for the good catch. It was some old code.

PR #2941, opened 2022-07-27
feat(rome_service): diagnostic severity for rules

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/2946

This PR adds new code generated methods to compute the `Severity` based on the configuration of the rule. This allows to inspect it and decide of the `Diagnostic` emitted by the linting phase should be changed.

The change of severity is done at `Workspace` level. This should have an added benefit to be language-agnostic and to work on different clients (CLI and LSP).

### Architectural changes

**The severity of the  diagnostics is now computed by workspace**, this means that when creating a rule, we won't be able to decide there the severity of rule.

The golden rule is the following:
- if the rule is recommended, by default it will emit and error;
- if the rule is not recommended, by default it will emit a warning;

In order to achieve that, I created a new data structure `AnalyzerDiagnostic` (suggested by Leo) which is simply a middle man that holds all the information to create a fully fledged `Diagnostic`, but it doesn't have the severity. When calling `into_diagnostic`, a severity must be passed and it will emit a `Diagnostic`. `into_diagnostic` is called by the workspace when requesting the diagnostics of a file. 

In order to correctly build the documentation of the rules, the codegen now creates a slim workspace and via `WorkspaceExt` we are now able to compute the correct severity of the diagnostics.

### Fixes 

This PR also fixes a bug where the recommeded rules were incorrectly computed in some cases.

The rule `useSemplifiedLogicExpression` is not recommended anymore because of https://github.com/rome/tools/issues/2948

### Pending works

This change unlocks something really interesting for our adoption. Now that a workspace is needed to compute these diagnostics correctly, the playground is now is not able to correctly compute this information.

In fact, the PR changes also the playground crate. When computing the diagnostic (via `signal.diagnostic()`), we default to `Severity::Error`. This **is correct** for now because the playground only runs the recommended rules, which will emit errors by design. Once the playground will be able to tweak the configuration of rules, then we will need to create a workspace instance just for the playground. 

## Test Plan

Created new test cases.

I also manually checked the LSP.

Given the following configuration:
```json
{
  "linter": {
    "enabled": true,
    "rules": {
      "recommended": true,
      "js": {
        "rules": {
          "noDeadCode": "error"
        }
      }
    }
  }
}

```
notice that `noDeadCode` is an error.
Screenshot 2022-07-28 at 10 24 55

Instead, given the following configuration:
```json
{
  "linter": {
    "enabled": true,
    "rules": {
      "recommended": true,
      "js": {
        "rules": {
          "noDeadCode": "warn"
        }
      }
    }
  }
}
```

Notice now how the LSP shows a warning instead.

Screenshot 2022-07-28 at 10 24 34

Updated documentation for the website and the snapshot tests.


crates/rome_js_analyze/src/analyzers/js/use_simplified_logic_expression.rs
```diff
@@ -127,7 +127,7 @@ impl Rule for UseSimplifiedLogicExpression {
     fn diagnostic(ctx: &RuleContext<Self>, _: &Self::State) -> Option<RuleDiagnostic> {
         let node = ctx.query();
 
-        Some(RuleDiagnostic::warning(
+        Some(RuleDiagnostic::error(
```
IWANABETHATGUY, line 130: This one probably should not change to severity `error`, as we discuss in discord

ematipico: Is it because it's buggy?

IWANABETHATGUY: > Is it because it's buggy?

Yes

leops (commented):
Since the workspace now overrides the severity of diagnostics emitted by the analyzer, should we remove the `severity` entirely from the analyzer by returning a partial `AnalyzerDiagnostic` object with no severity to be converted into an actual `Diagnostic` by the workspace by injecting the severity ? It seems weird to me that rules are even able to set a severity if we're going to overwrite it later on anyway

crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -126,18 +127,44 @@ fn debug_print(_rome_path: &RomePath, parse: AnyParse) -> String {
...
+                        rule_severity
+                    } else {
+                        Severity::Error
+                    }
+                } else {
+                    Severity::Error
+                }
+            } else {
+                Severity::Error
+            };
```
leops, line 159: This if-else chain could be simplified using combinator methods:
```suggestion
            let severity = rules.as_ref()
                .and_then(|rules| {
                    rules.get_severity_from_code(diagnostic.code()?.as_str())
                })
                .unwrap_or(Severity::Error);
```

crates/rome_analyze/src/lib.rs
```diff
@@ -571,6 +573,91 @@ impl<'analysis> AnalysisFilter<'analysis> {
...
 
+/// It holds onto the information necessary to eventually create a [Diagnostic]
+pub enum AnalyzerDiagnosticBag {
+    /// Inside the bag we have a diagnostic created by the linter
+    Rule(RuleDiagnostic),
+    /// We have raw information to create a basic [Diagnostic]
+    Raw(Diagnostic),
+}
+
+pub struct AnalyzerDiagnostic {
```
leops, line 584: The since the `file_id`, `code` and `code_link` fields are only set for rule diagnostics, the `AnalyzerDiagnostic` type could be an enum directly, with the `Rule` variant holding all the optional fields along with the inner diagnostic (this would avoid having to unwrap the file ID in `into_diagnostic`)

crates/rome_service/src/workspace.rs
```diff
@@ -400,3 +401,11 @@ impl<'app, W: Workspace + ?Sized> Drop for FileGuard<'app, W> {
             .ok();
     }
 }
+
+/// Convenience trait with utility functions to query the workspace
+pub trait WorkspaceExt: Workspace {
```
leops, line 406: The convention for `-Ext` traits is that they are implemented in an `impl<T: Workspace> WorkspaceExt for T {}` block, but this doesn't seem to be the case here. It should probably have a name that's more descriptive of the actual "capability" for its underlying workspace type

xtask/lintdoc/src/workspace.rs
```diff
@@ -0,0 +1,93 @@
...
+        unimplemented!()
+    }
+
+    fn rename(&self, _params: RenameParams) -> Result<RenameResult, RomeError> {
+        unimplemented!()
+    }
+}
+
+impl WorkspaceExt for WorkspaceTest {
+    fn get_severity_from_rule_code(&self, code: &str) -> Severity {
```
leops, line 82: This could probably be a simple helper function over `WorkspaceSettings` without having to implement the whole `Workspace` trait

xtask/lintdoc/src/main.rs
```diff
@@ -423,19 +427,31 @@ fn assert_lint(
     } else {
         let root = parse.tree();
 
+        let mut settings = WorkspaceSettings::default();
+        settings.linter.rules = Some(Rules::default());
+
+        let workspace = WorkspaceTest {
```
leops, line 433: I'm still unsure about why this code needs to create a fake workspace instance that's only used as a wrapper for the settings. The analyzer callback could read the `settings` variable above directly, without having to wrap it in a mutex

ematipico: True!

leops (approved):

PR #2949, opened 2022-07-28
refactor(rome_rowan): change mutation methods to return detached nodes

leops:
## Summary

The `splice_slots` and `replace_child` methods in `rome_rowan` were initially implemented to rebuild the entire tree above the mutated node, but no code was actually relying on this outside of the implementation of `replace_child` itself. I've modified `splice_slots` and moved `replace_child` into the `cursor` API to return a new detached node instead of respining the entire tree, this greatly simplifies the algorithm since it doesn't need to rebuild the green tree above the current node and doesn't need to rebuild the red tree at all.

## Test Plan

This change is internal to rowan, while it does change the semantics of `splice_slots` and `replace_child` no code relies on this specific behavior and all the tests for the analyzer should continue to pass

### DIFF
--- a/crates/rome_rowan/src/ast/batch.rs
+++ b/crates/rome_rowan/src/ast/batch.rs
@@ -176,6 +176,18 @@ where
         self.push_change(prev_element, Some(next_element))
     }
 
+    /// Push a change to replace the "prev_node" with "next_node".
+    ///
+    /// Changes to take effect must be commited.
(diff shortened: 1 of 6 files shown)


leops:
!bench_analyzer

MichaReiser (approved):

ematipico (approved):

PR #2950, opened 2022-07-28
chore(cargo): always enable LTO in release mode

leops:
## Summary

Link-Time Optimization was previously enabled only in production builds (running on CI) since it makes the compilation times longer, but release builds take a long time anyway so the difference seems marginal while enabling whole-project optimization brings in important improvements to performance (for instance the various functions making up the control flow graph visitor can be almost entirely inlined)

Measured cold compilation times for the CLI in release mode:
Without LTO: `105.1s (1m 45.1s)`
With LTO: `113.2s (1m 53.2s)`

## Test Plan

This is a change to the optimization options of the compiler, it shouldn't have any observable side effects (besides improving performances)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -24,3 +24,6 @@ bench_analyzer = "run -p xtask_bench --release -- --feature analyzer"
 coverage = "run -p xtask_coverage --profile=release-with-debug --"
 rome-cli = "run -p rome_cli --release --"
 rome-cli-dev = "run -p rome_cli --"
+
+[profile.release]
+lto = true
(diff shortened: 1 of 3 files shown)


leops:
!bench_analyzer

ematipico:
Does that mean that now all the bench commands will run with the `lto` enabled?

leops:
> Does that mean that now all the bench commands will run with the `lto` enabled?

Yes, in fact that's probably the main point of the change: it doesn't actually "improve" the results of the benchmarks but brings them in line with the actual performance of production builds

MichaReiser:
!bench_formatter

MichaReiser:
!bench_parser

MichaReiser (approved):

ematipico (approved):

PR #2952, opened 2022-07-28
feat(rome_service): `--apply-suggested` argument to `rome check`

ematipico:
### `--apply-suggested` argument

This PR adds a new argument `--apply-suggested` to the `check`.

As opposed to `--apply`, which only applies safe fixes, this argument tells to the workspace to apply fix actions that are marked with suggested fixes, marked with `Applicability::MaybeIncorrect` **and** safe fixes.

This argument is temporary and it will be replaced with another argument called `--review`, which will allow users to interactively choose the safe fixes.

### Settings passed to format requests

Until now we were passing the `IndentStyle` coming from the editor, which would then override the one coming from the configuration. The PR removes the need to sending the indentation coming from the LSP. We now have a configuration file and Rome should rely on what the user specified there.

### Messaging of the `rome check --apply` command

When using `--apply`, the workspace now tracks the fixes marked as suggested, and at the end it will present a message with the amount of suggested fixes that were not applied, and it will suggest the new argument added in this PR

## Test Plan

I added new tests

### DIFF
--- a/crates/rome_cli/src/commands/check.rs
+++ b/crates/rome_cli/src/commands/check.rs
@@ -50,7 +50,8 @@ pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
     traverse(
         TraversalMode::Check {
             max_diagnostics,
-            should_fix: session.args.contains("--apply"),
+            should_apply_safe_fixes: session.args.contains("--apply"),
+            should_apply_suggested_fixes: session.args.contains("--apply-suggested"),
         },
         session,
     )
--- a/crates/rome_cli/src/commands/check.rs
+++ b/crates/rome_cli/src/commands/check.rs
@@ -50,14 +50,14 @@ pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
     let apply = session.args.contains("--apply");
     let apply_suggested = session.args.contains("--apply-suggested");
 
-    let fix_file_mode = if apply == true && apply_suggested == true {
+    let fix_file_mode = if apply && apply_suggested {
         return Err(Termination::IncompatibleArguments(
             "--apply",
             "--apply-suggested",
         ));
-    } else if apply == false && apply_suggested == false {
+    } else if !apply && !apply_suggested {
         None
-    } else if apply == true && apply_suggested == false {
+    } else if apply && !apply_suggested {
         Some(FixFileMode::SafeFixes)
     } else {
         Some(FixFileMode::SafeAndSuggestedFixes)
--- a/crates/rome_cli/src/commands/check.rs
+++ b/crates/rome_cli/src/commands/check.rs
@@ -6,7 +6,7 @@ use crate::{
 use rome_diagnostics::MAXIMUM_DISPLAYABLE_DIAGNOSTICS;
 use rome_service::load_config;
 use rome_service::settings::WorkspaceSettings;
-use rome_service::workspace::UpdateSettingsParams;
+use rome_service::workspace::{FixFileMode, UpdateSettingsParams};
 
 /// Handler for the "check" command of the Rome CLI
 pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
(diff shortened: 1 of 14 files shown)


crates/rome_cli/src/traversal.rs
```diff
@@ -483,11 +504,20 @@ fn process_file(ctx: &TraversalOptions, path: &Path, file_id: FileId) -> FileRes
...
                 .with_file_id_and_code(file_id, "Lint")?;
 
+            if fixed.skipped_suggested_fixes > 0 {
+                ctx.push_message(TraversalWarning {
+                    title: format!("Skipped {} suggested fixes", fixed.skipped_suggested_fixes),
+                    paragraph:
+                        "If you wish to apply the suggested fixes, use --apply-suggested argument"
+                            .to_string(),
+                })
+            }
```
ematipico, line 519: I really don't like this way to push messages via context. @leops do you have a better way to so?

crates/rome_cli/src/traversal.rs
```diff
@@ -290,7 +295,12 @@ pub(crate) enum TraversalMode {
     Check {
         max_diagnostics: u16,
         /// `true` when running the command `check` with the `--apply` argument
-        should_fix: bool,
+        should_apply_safe_fixes: bool,
```
leops, line 298: These two fields could be combined into a single `Option<FixFileMode>` (and maybe the code reading these values from the command line should print an error if both `--apply` and `--apply-suggested` are specified)

crates/rome_cli/src/traversal.rs
```diff
@@ -483,11 +504,20 @@ fn process_file(ctx: &TraversalOptions, path: &Path, file_id: FileId) -> FileRes
...
 
-        if ctx.mode.should_fix() {
+        if let Some(fix_mode) = ctx.mode.as_fix_file_mode() {
             let fixed = file_guard
-                .fix_file()
+                .fix_file(fix_mode)
                 .with_file_id_and_code(file_id, "Lint")?;
 
+            if fixed.skipped_suggested_fixes > 0 {
+                ctx.push_message(TraversalWarning {
```
leops, line 513: I think instead of a generic `TraversalWarning` message we should be using a specific `SkippedFixes` containing the value of `skipped_suggested_fixes`, then the console printing code could accumulate these values and print a single count for all the checked files at once (like we already do for unprinted diagnostics)

crates/rome_service/src/settings.rs
```diff
@@ -155,9 +154,11 @@ pub struct LanguageSettings<L: Language> {
 
 /// Handle object holding a temporary lock on the workspace settings until
 /// the deferred language-specific options resolution is called
+#[derive(Debug)]
 pub(crate) struct SettingsHandle<'a, E> {
     inner: RwLockReadGuard<'a, WorkspaceSettings>,
     /// Additional per-request state injected by the editor
+    #[allow(dead_code)]
```
leops, line 161: If this field is now unused let's remove it

MichaReiser:
> Until now we were passing the IndentStyle coming from the editor, which would then override the one coming from the configuration. The PR removes the need to sending the indentation coming from the LSP. We now have a configuration file and Rome should rely on what the user specified there.

What if there's no configuration in place? Shouldn't Rome then use the editor defaults?

crates/rome_cli/src/commands/check.rs
```diff
@@ -50,7 +50,8 @@ pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
     traverse(
         TraversalMode::Check {
             max_diagnostics,
-            should_fix: session.args.contains("--apply"),
+            should_apply_safe_fixes: session.args.contains("--apply"),
+            should_apply_suggested_fixes: session.args.contains("--apply-suggested"),
```
MichaReiser, line 54: Nit: This could also be represented as enum to a have fewer boolens: 

```
enum ApplyFixes { Safe, Suggested, Both }
```

MichaReiser (commented):
What do you think of changing the `--apply-suggested` argument to `--apply=suggested`?

ematipico:
> What do you think of changing the `--apply-suggested` argument to `--apply=suggested`?

From the description:

> This argument is temporary and it will be replaced with another argument called --review, which will allow users to interactively choose the safe fixes.

`--review` comes form Rome classic, and it was the "way to go" to apply suggested fixes. So I am not sure if it's safe to change the `--apply` argument behaviour, to change it back when we will introduce `--review`

ematipico:
> > Until now we were passing the IndentStyle coming from the editor, which would then override the one coming from the configuration. The PR removes the need to sending the indentation coming from the LSP. We now have a configuration file and Rome should rely on what the user specified there.
> 
> What if there's no configuration in place? Shouldn't Rome then use the editor defaults?

We should not use editor's configuration, because they would go in conflict with commands like `rome ci`. If we can't keep the configuration stable across environments, we risk to create errors.

crates/rome_service/src/settings.rs
ematipico: Are you sure about it? It's definitely something that we will need in the future if we decided to have options in the editors.

leops: At least at the moment I don't really see any use case for being able to provide configuration override into the workspace on a per-request basis

leops (approved):

PR #2968, opened 2022-07-29
test(rome_js_formatter): Only snapshot not matching prettier tests

MichaReiser:
The goal of Rome's formatter is to match Prettier's formatting for most cases. That's why this PR changes the prettier snapshot tests to:

* Only create a snapshot if Rome's and Prettier's formatting doesn't match
* Print the diff of Rome's and Prettier's formatting for tests with different output.
* Delete snapshots of tests for matching output

The overreaching goal of this PR is to simplify reviewing formatter changes so that it becomes easier to tell if a change is a regression:

* Regresssions:
  * New snapshot file is added
  * New lines are added to the "diff" section of the snapshot
* Improvement
  * Snapshot file gets deleted
  * Lines are removed from the diff section
* Neutral
  * Snapshot file doesn't exist or doesn't change
  * Existing lines have been changed.

## Test Plan

* Verified that it only deletes snapshots of tests with matching output
* Verified that new snapshots are created if I change the formatting


ematipico:
So now the test suite works in the other way around? Meaning that if we don't have snapshots in the prettier testing suite, it means we are doing a good job?

MichaReiser:
> So now the test suite works in the other way around? Meaning that if we don't have snapshots in the prettier testing suite, it means we are doing a good job?

Yes, that's how it works. Not sure if I would call this the "opposite" way because we always had both snapshots before.

ematipico (approved):

PR #2969, opened 2022-07-29
feat(playground): display the control flow graph in the playground

leops:
## Summary

This PR add a new "Control Flow Graph" tab in the playground displaying the control flow graph for the currently selected function (the function that contains either the cursor or the start of the current selection) using mermaid.js

In order to extract the CFG from the analyzer I introduced a concept of "matcher layer" that can wrap an existing query matcher and intercept incoming query matches from the visitors. Calling the analyzer through the existing `analyze` entry point simply uses an empty layer function, this is all statically dispatched so it should all get inlined and come out as zero-cost in the end.

### DIFF
--- a/crates/rome_analyze/src/matcher.rs
+++ b/crates/rome_analyze/src/matcher.rs
@@ -113,12 +113,16 @@ impl<'phase, L: Language> PartialEq for SignalEntry<'phase, L> {
     }
 }
 
+/// Adapter type wrapping a [QueryMatcher] type with a `layer` function that
+/// can be used to inspect the query matches emitted by the analyzer
 pub struct MatcherLayer<F, I> {
     layer: F,
     inner: I,
 }
 
 impl<F, I> MatcherLayer<F, I> {
+    ///  Create a new instance of [MatcherLayer] from an existing [QueryMatcher]
+    /// object and a wrapper layer function
     pub fn new<L>(inner: I, layer: F) -> Self
     where
         L: Language,
--- a/crates/rome_analyze/src/matcher.rs
+++ b/crates/rome_analyze/src/matcher.rs
@@ -113,6 +113,42 @@ impl<'phase, L: Language> PartialEq for SignalEntry<'phase, L> {
     }
 }
 
+pub struct MatcherLayer<F, I> {
+    layer: F,
+    inner: I,
+}
+
+impl<F, I> MatcherLayer<F, I> {
+    pub fn new<L>(inner: I, layer: F) -> Self
+    where
+        L: Language,
+        F: FnMut(&MatchQueryParams<L>),
+        I: QueryMatcher<L>,
+    {
+        Self { layer, inner }
+    }
+}
+
+impl<L, F, I> QueryMatcher<L> for MatcherLayer<F, I>
+where
+    L: Language,
+    F: FnMut(&MatchQueryParams<L>),
+    I: QueryMatcher<L>,
+{
+    fn find_group(&self, group: &str) -> Option<GroupKey> {
+        self.inner.find_group(group)
+    }
+
+    fn find_rule(&self, group: &str, rule: &str) -> Option<RuleKey> {
+        self.inner.find_rule(group, rule)
+    }
+
+    fn match_query(&mut self, params: MatchQueryParams<L>) {
+        (self.layer)(&params);
+        self.inner.match_query(params);
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use rome_console::codespan::Severity;
--- a/crates/rome_analyze/src/matcher.rs
+++ b/crates/rome_analyze/src/matcher.rs
@@ -113,27 +113,27 @@ impl<'phase, L: Language> PartialEq for SignalEntry<'phase, L> {
     }
 }
 
-/// Adapter type wrapping a [QueryMatcher] type with a `layer` function that
-/// can be used to inspect the query matches emitted by the analyzer
-pub struct MatcherLayer<F, I> {
-    layer: F,
+/// Adapter type wrapping a [QueryMatcher] type with a function that can be
+/// used to inspect the query matches emitted by the analyzer
+pub struct InspectMatcher<F, I> {
+    func: F,
     inner: I,
 }
 
-impl<F, I> MatcherLayer<F, I> {
-    ///  Create a new instance of [MatcherLayer] from an existing [QueryMatcher]
-    /// object and a wrapper layer function
-    pub fn new<L>(inner: I, layer: F) -> Self
+impl<F, I> InspectMatcher<F, I> {
+    ///  Create a new instance of [InspectMatcher] from an existing [QueryMatcher]
+    /// object and an inspection function
+    pub fn new<L>(inner: I, func: F) -> Self
     where
         L: Language,
         F: FnMut(&MatchQueryParams<L>),
         I: QueryMatcher<L>,
     {
-        Self { layer, inner }
+        Self { func, inner }
     }
 }
 
-impl<L, F, I> QueryMatcher<L> for MatcherLayer<F, I>
+impl<L, F, I> QueryMatcher<L> for InspectMatcher<F, I>
 where
     L: Language,
     F: FnMut(&MatchQueryParams<L>),
@@ -148,7 +148,7 @@ where
     }
 
     fn match_query(&mut self, params: MatchQueryParams<L>) {
-        (self.layer)(&params);
+        (self.func)(&params);
         self.inner.match_query(params);
     }
 }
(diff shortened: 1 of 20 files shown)


xunilrj (approved):

ematipico:
The new tab can't been seen on mobile

crates/rome_analyze/src/matcher.rs
```diff
@@ -113,6 +113,42 @@ impl<'phase, L: Language> PartialEq for SignalEntry<'phase, L> {
...
+
+    fn find_rule(&self, group: &str, rule: &str) -> Option<RuleKey> {
+        self.inner.find_rule(group, rule)
+    }
+
+    fn match_query(&mut self, params: MatchQueryParams<L>) {
+        (self.layer)(&params);
+        self.inner.match_query(params);
+    }
+}
```
ematipico, line 150: We should document the struct and the functions. And how to use them.

crates/rome_analyze/src/matcher.rs
```diff
@@ -113,6 +113,42 @@ impl<'phase, L: Language> PartialEq for SignalEntry<'phase, L> {
...
+{
+    fn find_group(&self, group: &str) -> Option<GroupKey> {
+        self.inner.find_group(group)
+    }
+
+    fn find_rule(&self, group: &str, rule: &str) -> Option<RuleKey> {
+        self.inner.find_rule(group, rule)
+    }
+
+    fn match_query(&mut self, params: MatchQueryParams<L>) {
```
ematipico, line 146: How does this function differ from `match_query` of `QueryMatcher`?

crates/rome_js_analyze/src/lib.rs
```diff
@@ -30,13 +31,15 @@ pub fn metadata(filter: AnalysisFilter) -> impl Iterator<Item = RegistryRuleMeta
 /// Run the analyzer on the provided `root`: this process will use the given `filter`
 /// to selectively restrict analysis to specific rules / a specific source range,
 /// then call `emit_signal` when an analysis rule emits a diagnostic or action
-pub fn analyze<'a, F, B>(
+pub fn analyze_with_matcher_layer<'a, V, F, B>(
```
ematipico, line 34: I think the documentation should be updated, now that this function has been renamed and it accepts a new argument.

website/playground/src/DesktopPlayground.tsx
```diff
@@ -59,6 +68,16 @@ export default function DesktopPlayground(
...
 
+	const onUpdate = useCallback((viewUpdate: ViewUpdate) => {
+		const cursorPosition = viewUpdate.state.selection.ranges[0]?.from ?? 0;
+		setPlaygroundState(
+			(state) =>
+				state.cursorPosition !== cursorPosition ? {
+					...state,
+					cursorPosition,
+				} : state,
+		);
```
ematipico, line 79: nit: I think there's no need to update the state if the cursor hasn't changed, meaning I think we should do like:
```js
if state.cursorPosition !== cursorPosition {
	setPlaygroundState()
}
```

ematipico (changes requested):

crates/rome_analyze/src/matcher.rs
leops: The `MatcherLayer` struct is an "adapter type" that implements the `QueryMatcher` trait on top of another, generic type also implementing this trait. It's similar to `Iterator` adapters like `Inspect<T>` that implements the iterator trait by calling through to the underlying `T: Iterator` object, then passing the items returned by `next` to the provided callback (the difference is that `MatcherLayer` calls the `layer` callback function before calling the implementation of `match_query` for the underlying matcher type)

website/playground/src/DesktopPlayground.tsx
leops: This would require to get rid of the memoization on the `onUpdate` function, or at least reattach a new event handler every time the state changes. This should have more or less the same effect anyway, it's my understanding of React that it will abort a state update if the updater function returns the same object unmodified but I could be wrong about that

ematipico (approved):

crates/rome_analyze/src/matcher.rs
```diff
@@ -113,6 +113,46 @@ impl<'phase, L: Language> PartialEq for SignalEntry<'phase, L> {
     }
 }
 
+/// Adapter type wrapping a [QueryMatcher] type with a `layer` function that
+/// can be used to inspect the query matches emitted by the analyzer
+pub struct MatcherLayer<F, I> {
+    layer: F,
+    inner: I,
+}
```
MichaReiser, line 121: Nit: `Inspect` seems to be an important concept, how about naming this an `InspectMatcher`?

crates/rome_js_analyze/src/lib.rs
```diff
@@ -54,7 +60,7 @@ where
     }
 
     let mut analyzer = Analyzer::new(
-        build_registry(&filter),
+        MatcherLayer::new(build_registry(&filter), matcher_layer),
```
MichaReiser, line 63: Would it be an option to add this functionality instead on `Analyzer` as `inspect`?

ematipico:
@leops could you check why the deployment on CloudFlare fails? Hopefully it's their fault

website/playground/src/MermaidGraph.tsx
```diff
@@ -0,0 +1,32 @@
...
+	graph: string;
+}
+
+export default function MermaidGraph({ graph }: MermaidGraphProps) {
+	const element = useRef<HTMLDivElement>(null);
+
+	useLayoutEffect(() => {
+		if (element.current) {
+			element.current.removeAttribute("data-processed");
+			mermaid.contentLoaded();
```
MichaReiser, line 18: How much processing time does this add for large files?

MichaReiser: An alternative could be to link to [mermaid's playground](https://mermaid.live/edit#pako:eNpVkE1qw0AMha8itEohvoAXhcZOsgm00Ow8WQiPnBma-WEsU4Ltu3ccU0i0kvS-JwmN2AbNWOI1UTRwrpWHHB9NZZLtxVF_gaJ4n44s4ILn-wS7zTFAb0KM1l_fVn63QFCNpwVjEGP9z7xK1cP_6XmCujlRlBAvz8r5N0ywb-yXyeNfFZM4uw5NR2VHRUsJKkoPBLfoODmyOp8-Lh2FYtixwjKnmjsabqJQ-TmjQ9QkvNdWQsJS0sBbpEHC9923__XK1JbyIxzmfbee5z-A61yS) (or take the JPG link) but
* likely to be unstable because it's not a real API
* unclear what params must be passed to the playground

MichaReiser (approved):

website/playground/src/MermaidGraph.tsx
leops: I tried to paste some of the very complex test cases we have for the CFG into the playground, with a large graph the processing can block the page for approximately 1 second. The alternative that GitHub or Notion seem to be using is to display the graph in a dedicated page embedded using an iframe, this way the rendering happens in its own dedicated process isolated from the main app

crates/rome_js_analyze/src/lib.rs
leops: That would work, but in the end the `rome_js_analyze` library would probably still need to expose two functions since the internal `Analyzer` is not part of the public API

website/playground/src/MermaidGraph.tsx
MichaReiser: Does this also block if you're tying and the CFG page is open? I'm not too concerned about the impact if it only happens on the CFG page (and ideally, doesn't affect typing).

ematipico (changes requested):
I think we have to do something with the build process (or the upload of static files). The playground is not able to load the iframe:

Screenshot 2022-08-04 at 12 46 13

leops:
> I think we have to do something with the build process (or the upload of static files). The playground is not able to load the iframe:

Ah I think it's an issue with the temporary deployment (it tries to load the content page for the iframe at `play.rome.tools/mermaid.html` when it should be `play.rome.tools/10cc478/mermaid.html`), it would correctly on the main playground deployment but I'll try to fix it so it works in PRs too

leops:
The CI check is failing on false positives for the `noUnusedVariables` rule introduced in the latest release, but this seems like an unrelated issue and this PR could be merged anyway ?

ematipico:
> The CI check is failing on false positives for the `noUnusedVariables` rule introduced in the latest release, but this seems like an unrelated issue and this PR could be merged anyway ?

I don't mind merging the PR

ematipico (approved):

PR #2970, opened 2022-07-29
feat(rome_js_formatter): Loop formatting

MichaReiser:
## Summary

This PR re-visits the formatting of loops. I verified that Rome generates the same IR (except for comments). 

I tried to match the same formatting for the semicolon's leading comments of `continue` and `break` statements but wasn't able to match the behaviour without 

a) adding a serious amount of complexity 
b) get a stable format.

That's why I left the "default" comments formatting in place. 

## Test Plan

Verified the updated snapshots.


ematipico (approved):

ematipico:
Could you add the metrics please?

MichaReiser:
> Could you add the metrics please?



File Based Average Prettier Similarity: 77.98% -> 77.97%
Line Based Average Prettier Similarity: 73.23% => 73.11%


My understanding of this small regression is that it is mainly coming from how we format comments. I wouldn't give this too much weight for now as we are aware that this is an open issue and requires more fundamental architecture changes

PR #2972, opened 2022-07-30
feat(rome_js_analyze): rule `noDupeArgs`

### LINKED ISSUE #2735: 📎 (rome_js_analyze): noDupeArgs
### Description

https://eslint.org/docs/rules/no-dupe-args

IWANABETHATGUY:
## Summary
1. Resolved #2735 

## Test Plan
1. Should pass all test cases newly added.


IWANABETHATGUY:
Wait #2973 to be merged.

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                    JsAnyFunction::JsFunctionExpression(func) => func.parameters().ok()?,
+                };
+                if let Some(binding) = args
+                    .syntax()
+                    .descendants()
+                    .find_map(JsIdentifierBinding::cast)
+                {
+                    let mut set = FxHashSet::default();
+                    let model = ctx.model();
```
xunilrj, line 67: Do you need the Semantic Model? 

I thought iterating the names of the arguments was enough in this case. Maybe just ```BTreeSet::insert```ing them and checking the return value.

xunilrj (approved):

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
IWANABETHATGUY: Iterate the arguments may be a method to handle this issue, but you need to check the scenario below:
```js
function test(a, {a, b, ...a}, [a, b]) {
  
}

function test2(a, ...args ) {
  
}
```
So that's why I am using the semantic model.

IWANABETHATGUY:
!bench_analyhzer

IWANABETHATGUY:
!bench_analyzer

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,92 @@
...
+        name: "noDupeArgs",
+        recommended: true,
+    }
+}
+
+impl Rule for NoDupeArgs {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Semantic<JsAnyFunction>;
+    type State = JsSyntaxNode;
```
ematipico, line 44: Using `JsSyntaxNode` is not recommended, because we can't understand which nodes are are involved. We should use an union is we have to include different nodes.

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,92 @@
...
+                    }
+                    JsAnyFunction::JsFunctionDeclaration(func) => func.parameters().ok()?,
+                    JsAnyFunction::JsFunctionExportDefaultDeclaration(func) => {
+                        func.parameters().ok()?
+                    }
+                    JsAnyFunction::JsFunctionExpression(func) => func.parameters().ok()?,
+                };
+                if let Some(binding) = args
+                    .syntax()
+                    .descendants()
```
ematipico, line 63: This is expensive... is it needed? Imagine if we had something like this:

```js
function f(
 	a = () => false,
	{ b = "something" },
) {}
```

It means that we have traverse all nodes contained in the arguments?

ematipico (changes requested):

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
IWANABETHATGUY: Nope. Actually, it use `find_map` so it will terminate when it meet the first `JsidentifierBinding`.
That means it will stop at `a` in this case, which pretty fast.

ematipico: But it won't if it doesn't find a `JsidentifierBinding`. This is a case where we might have `JsUnknownParameter` instead of `JsFormatParameter`. Because of the fact that it's an `unknown` node, it might contain more valid nodes that could cause false positives.

I think it's best to surgically query the nodes you need.

```rust
args.map(|arg| {
	// here check the type of argument and query the biding
});
```

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,92 @@
...
+    ///
+    /// ```js
+    /// function i(i, b, c) {}
+    /// var j = function (j, b, c) {};
+    /// function k({ k, b }, { c, d }) {}
+    /// function l([, l]) {}
+    /// function foo([[a, b], [c, d]]) {}
+    /// ```
+    pub(crate) NoDupeArgs {
+        version: "0.8.0",
```
ematipico, line 34: This should be `0.9.0` now

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
IWANABETHATGUY: The `SyntaxNode` is cloned from `binding.syntax()`, and `binding` has no method to return a specific `JsAstNode`

ematipico: If you know the type you need, you can use `cast`, `can_cast`, `cast_unwrap` to retrieve a typed node from a `JsSyntaxNode`

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
IWANABETHATGUY: Your example is a function call expression, not Any JsFunction

IWANABETHATGUY: As I reply to @xunilrj , https://github.com/rome/tools/pull/2972#discussion_r933942507, we could iterate the args list, but the implementation would be pretty complex since we need to handle all kinds of parameter, 
```
function test(a, {a, b, ...a}, [a, b]) {
  
}

function test2(a, ...args ) {
  
}
function foo([[a, b], [c, d]]) {}
function foo([[[a, b], [c, d]]]) {} // the array binding pattern could be recursively, 
function foo([[[a = function(d, g, f) {}, b], [c, d]]]) {} // the array binding pattern could be recursively, 
```
And also, according to the benchmark, my implementation is not bad.
Maybe we could optimize this rule after we found it has performance issues. What do you think?

ematipico: Unfortunately we can't rely on the current benchmark that much, because it doesn't reflect the benches we are interested with. (this is a separate issue, we're working on it)

What about this real case: https://github.com/rome/tools/blob/main/crates/rome_js_formatter/tests/specs/prettier/js/performance/nested-real.js

ematipico: > Maybe we could optimize this rule after we found it has performance issues. What do you think?

I would then move it `recommended: false`, unfortunately I am not really confident about the current implementation

MichaReiser: It could even be that this yields incorrect results. Let's say you have

```
function test(a  = function(a) {}) {}
```

`descendents` descents into the inner `function`, reporting both `a` identifiers.

IWANABETHATGUY:
!bench_analyzer

IWANABETHATGUY:
@ematipico I rewrite the algorithm as you said before, using a recursive pattern matching algorithm to find the duplicate params. Would you mind helping with reviewing?

leops:
So this may be more of a discussion we could have in the corresponding issue for this rule rather than in the PR, but I'm wondering if there's a way we could make the implementation for this rule more generic, into a `noNameShadowing` rule that would disallow duplicate binding names in the same scope whether that's two arguments, variables, functions, object / class member ...

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+        Some(RuleDiagnostic::new(
+            binding_syntax_node.syntax().text_trimmed_range(),
+            markup! {
+                "Duplicate argument name"
+            },
+        ))
+    }
+}
+
+/// Traverse the parameter recursively and check if it is duplicated.
```
ematipico, line 86: Could you update the comment and explain what "traverse a parameter recursively" mean, please? It's clear, because it's not something that a developer does every day. Maybe adding some snippet in the documentation can help to understand

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+    p: JsAnyParameter,
+    set: &mut FxHashSet<String>,
+    res: &mut Vec<JsIdentifierBinding>,
+) -> Option<()> {
+    match p {
+        JsAnyParameter::JsAnyFormalParameter(p) => match p {
+            rome_js_syntax::JsAnyFormalParameter::JsFormalParameter(parameter) => {
+                traverse_binding(parameter.binding().ok()?, set, res);
+            }
+            rome_js_syntax::JsAnyFormalParameter::JsUnknownParameter(_) => {}
```
ematipico, line 97: We can remove `rome_js_syntax` here, as we already use it.

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+            },
+        ))
+    }
+}
+
+/// Traverse the parameter recursively and check if it is duplicated.
+fn traverse_parameter(
+    p: JsAnyParameter,
+    set: &mut FxHashSet<String>,
+    res: &mut Vec<JsIdentifierBinding>,
```
ematipico, line 90: Let's avoid abbreviations, at  least for `set` and `res`. Sometimes it's difficult for someone's brain to understand what they stand for

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+        },
+        JsAnyParameter::JsRestParameter(rest_parameter) => {
+            traverse_binding(rest_parameter.binding().ok()?, set, res);
+        }
+        JsAnyParameter::TsThisParameter(_) => {}
+    }
+    Some(())
+}
+
+fn traverse_binding(
```
ematipico, line 107: Could you add some documentation to this function and explain what it does?

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+                    JsAnyObjectBindingPatternMember::JsUnknownBinding(_) => {}
+                }
+            }
+        }
+    }
+    Some(())
+}
+
+#[inline]
+fn check_binding(
```
ematipico, line 179: `check_binding` doesn't do any check... Maybe we can call it `track_binding`? And document what happens, we should explain why sometimes we do `res.push` or `set.insert`. By looking at the code I can't understand it.

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+                    }
+                    JsAnyFunction::JsFunctionDeclaration(func) => func.parameters().ok()?,
+                    JsAnyFunction::JsFunctionExportDefaultDeclaration(func) => {
+                        func.parameters().ok()?
+                    }
+                    JsAnyFunction::JsFunctionExpression(func) => func.parameters().ok()?,
+                };
+                let mut set = FxHashSet::default();
+                // Traversing the parameters of the function in preorder and checking for duplicates,
+                // reuse the `HashSet` and `Vec` to avoid extra allocations.
```
ematipico, line 64:
```suggestion
                // reuse the `FxHashSet` and `Vec` to avoid extra allocations.
```

crates/rome_js_analyze/tests/specs/js/noDupeArgs.js
```diff
@@ -0,0 +1,26 @@
+// valid
```
ematipico, line 1: Would you mind adding more tests, please?
- arrow functions
- class methods
- object methods

We can replace the ones you already added, for example:

`function d(a, b, a) {}` => `var d = (a, b, a) => {}`. So we test the error case with arrow functions

Could you add this cases too:

```js
export function f(a, b, c = (a, b, b) => {}) {
  
}
```

It should fail in the arrow function

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+impl Rule for NoDupeArgs {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsAnyFunction>;
+    type State = JsIdentifierBinding;
+    type Signals = Vec<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Vec<Self::State> {
+        let function = ctx.query();
+        Some(vec![])
```
MichaReiser, line 50: What's the reason for using `Some(vec![])` here? Seems like it's always used?

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,190 @@
...
+) -> Option<()> {
+    match binding {
+        JsAnyBindingPattern::JsAnyBinding(inner_binding) => match inner_binding {
+            JsAnyBinding::JsIdentifierBinding(id_binding) => {
+                check_binding(id_binding, set, res);
+            }
+            JsAnyBinding::JsUnknownBinding(_) => {}
+        },
+        JsAnyBindingPattern::JsArrayBindingPattern(inner_binding) => {
+            for ele in inner_binding.elements().into_iter() {
```
MichaReiser, line 120:
```suggestion
            for element in inner_binding.elements().into_iter() {
```

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
IWANABETHATGUY: Because I want to use the handy `?`, but the return type of `run` is `Vec<Self::State>` which I can't use the `?` operator.

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+    Some(())
+}
+
+#[inline]
+/// If the name of binding has been seen in set, then we push the `JsIdentifierBinding` into `identifier_vec`.
+/// Else we mark the name of binding as seen.
+fn track_binding(
+    id_binding: JsIdentifierBinding,
+    binding_set: &mut FxHashSet<String>,
```
ematipico, line 193: Ok, with the documentation now I understand better what we're doing. I would suggest to change the name of the variables: `binding_set` => `tracked_bindings` and `identifier_vec` => `duplicated_arguments` (the name reminds the name of the rule)

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,207 @@
...
+    identifier_vec: &mut Vec<JsIdentifierBinding>,
+) {
+    let binding_text = id_binding.text();
+    if binding_set.contains(&binding_text) {
+        identifier_vec.push(id_binding);
```
ematipico, line 198: Don't we need to remove `id_binding` from `binding_set` here? The old rule bailed once it found just one.

crates/rome_js_analyze/src/semantic_analyzers/js/no_dupe_args.rs
```diff
@@ -0,0 +1,207 @@
...
+impl Rule for NoDupeArgs {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsAnyFunctionAndMethod>;
+    type State = JsIdentifierBinding;
+    type Signals = Vec<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Vec<Self::State> {
+        let function = ctx.query();
+        Some(vec![])
```
ematipico, line 52: Do we need to track the names in a vector? The old rule used to bail once it found at least one duplicate. Can't we do the same here or there's a reason why we should do it for all the parameters?

IWANABETHATGUY: Typescript using this behavior 
image

IWANABETHATGUY: I am ok with both.

IWANABETHATGUY: @ematipico, which one do you prefer? This maybe the last issue this pr has

ematipico: I would go for the simplest approach. It also requires less AST navigation

IWANABETHATGUY: Done

IWANABETHATGUY:
!bench_analyzer

ematipico (approved):
@xunilrj would you mind a give a second view and merge if you think it's fine?

xunilrj:
> @xunilrj would you mind a give a second view and merge if you think it's fine?

Done.

PR #2974, opened 2022-07-30
fix(playground): rome ast codemirror grammar

IWANABETHATGUY:
## Summary
1. Fixing wrong Rome ast codemirror grammar
2. Screenshot:
**Main**
image
**Current Branch**
image
3. For more details, you could see patch pr https://github.com/IWANABETHATGUY/lang-rome-ast/commit/8134a49cb672290c930a8e6993c599833086ac37

## Test Plan
1. Running the playground locally or see my screenshot.

### DIFF
--- a/website/playground/package.json
+++ b/website/playground/package.json
@@ -15,7 +15,7 @@
 	},
 	"dependencies": {
 		"@codemirror/lang-javascript": "^6.0.1",
-		"codemirror-lang-rome-ast": "0.0.3",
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):
Nice! 
How hard would it be to add a grammar for Rome's formatter IR? Would be awesome to have collapsing of elements. 


have you had any time looking into why the grammar doesn't work when embedding it locally?

PR #2980, opened 2022-08-01
fix(rome_js_analyze): fix the removal of nodes from syntax lists in batch mutations

### LINKED ISSUE #2966: 🐛 `noDebugger` errors while trying a safe fix
### Environment information

```bash
Current codebase on `main`
```

### What happened?

Given the following code 

```js

function f() {
    debugger;
    return {}
}
```

The rule `noDebugger` emits an invalid CST. I discovered this issue while I was trying to format the emitted CST by the action of the rule.

The stack trace:

```block

thread '<unnamed>' panicked at 'Node isn't permitted to contain empty slots', /Users/emanuele/www/rometools/rome/crates/rome_rowan/src/ast/mod.rs:239:34
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:142:14
   2: rome_rowan::ast::AstNodeListIterator<L,N>::slot_to_node
             at /Users/emanuele/www/rometools/rome/crates/rome_rowan/src/ast/mod.rs:239:34
   3: <rome_rowan::ast::AstNodeListIterator<L,N> as core::iter::traits::iterator::Iterator>::next
             at /Users/emanuele/www/rometools/rome/crates/rome_rowan/src/ast/mod.rs:254:14
   4: <rome_js_formatter::js::lists::statement_list::FormatJsStatementList as rome_formatter::FormatRule<rome_js_syntax::generated::nodes::JsStatementList>>::fmt
             at /Users/emanuele/www/rometools/rome/crates/rome_js_formatter/src/js/lists/statement_list.rs:13:26
```

This is a long stack trace, I just copied the first part.

### Expected result

I would expect to not throw this kind of error

### Code of Conduct

leops:
## Summary

This PR fixes #2966 by changing the behavior of the `remove_node` batch mutation method in syntax lists, making it actually remove the corresponding slot from the list instead of replacing the previous node with an empty slot.

In order to make this work the batch mutation now has to track the number of removed slots in a node to adjust the indices of other modifications being applied to the same node, in turn this means I had to change the ordering semantics of mutation changes slightly so that modifications that apply to the same node are sorted in increasing order by the index of the slot they are modifying.

## Test Plan

To test this change and protect from future regressions I modified the analyzer tests to perform additional checks on the syntax tree emitted by code actions. These checks include:
- verifying that applying the text edits returned by `as_text_edits` and printing the syntax tree returned by `commit` result in the same source code
- verifying the resulting syntax tree has no unknown nodes, empty slots or missing required children
- verifying that re-parsing the returned source code does not yield any syntax errors

Incidentally these new checks required changes in a few existing tests that had syntax errors in the input file

### DIFF
--- a/crates/rome_js_analyze/src/utils.rs
+++ b/crates/rome_js_analyze/src/utils.rs
@@ -1,3 +1,6 @@
+use rome_js_factory::make;
+use rome_js_syntax::{JsAnyRoot, JsAnyStatement, JsLanguage, JsModuleItemList, JsStatementList, T};
+use rome_rowan::{AstNode, BatchMutation};
 use std::borrow::Cow;
 
 pub mod rename;
@@ -114,6 +117,28 @@ pub fn to_camel_case(input: &str) -> Cow<str> {
 
         Cow::Owned(output)
     }
+}
+
+pub(crate) fn remove_statement<N>(
+    mutation: &mut BatchMutation<JsLanguage, JsAnyRoot>,
+    node: &N,
+) -> Option<()>
+where
+    N: AstNode<Language = JsLanguage>,
+    JsAnyStatement: From<N>,
+{
+    let parent = node.syntax().parent()?;
+
+    if JsStatementList::can_cast(parent.kind()) || JsModuleItemList::can_cast(parent.kind()) {
+        mutation.remove_node(node.clone());
+    } else {
+        mutation.replace_node(
+            JsAnyStatement::from(node.clone()),
+            JsAnyStatement::JsEmptyStatement(make::js_empty_statement(make::token(T![;]))),
+        );
+    }
+
+    Some(())
 }
 
 #[test]
--- a/crates/rome_js_analyze/src/utils.rs
+++ b/crates/rome_js_analyze/src/utils.rs
@@ -119,13 +119,15 @@ pub fn to_camel_case(input: &str) -> Cow<str> {
     }
 }
 
+/// Utility function to remove a statement node from a syntax tree, by either
+/// removing the node from its parent if said parent is a statement list or
+/// module item list, or by replacing the statement node with an empty statement
 pub(crate) fn remove_statement<N>(
     mutation: &mut BatchMutation<JsLanguage, JsAnyRoot>,
     node: &N,
 ) -> Option<()>
 where
-    N: AstNode<Language = JsLanguage>,
-    JsAnyStatement: From<N>,
+    N: AstNode<Language = JsLanguage> + Into<JsAnyStatement>,
 {
     let parent = node.syntax().parent()?;
 
@@ -133,7 +135,7 @@ where
         mutation.remove_node(node.clone());
     } else {
         mutation.replace_node(
-            JsAnyStatement::from(node.clone()),
+            node.clone().into(),
             JsAnyStatement::JsEmptyStatement(make::js_empty_statement(make::token(T![;]))),
         );
     }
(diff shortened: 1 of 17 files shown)


crates/rome_js_analyze/src/utils.rs
```diff
@@ -116,6 +119,28 @@ pub fn to_camel_case(input: &str) -> Cow<str> {
     }
 }
 
+pub(crate) fn remove_statement<N>(
```
xunilrj, line 122: In this case would not be better to name this ```replace_with_empty_statement```?

xunilrj, line 122: In this case, I think makes sense to document when it removes and when it replaces.

xunilrj:
I suppose we will need to be able to insert nodes in the future. Right?

xunilrj (approved):

leops:
> I suppose we will need to be able to insert nodes in the future. Right?

At the moment it's possible to get around it by first creating a mutated copy of the parent list with the new node inserted, then adding a replacement operation for that list node to the batch mutation (this is what `useSingleVarDeclarator` does for instance). But it certainly would be nicer to have insertion supported directly in the batch mutation API

crates/rome_js_analyze/src/utils.rs
```diff
@@ -116,6 +119,28 @@ pub fn to_camel_case(input: &str) -> Cow<str> {
...
+    mutation: &mut BatchMutation<JsLanguage, JsAnyRoot>,
+    node: &N,
+) -> Option<()>
+where
+    N: AstNode<Language = JsLanguage>,
+    JsAnyStatement: From<N>,
```
MichaReiser, line 128: My understanding is that it is preferred to use `Into<JsAnyStatement`> over `From`. What's your reasoning for constraining on `From` here?

crates/rome_js_parser/src/test_utils.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use rome_diagnostics::{file::SimpleFile, termcolor::Buffer, Emitter};
+use rome_js_syntax::{JsLanguage, JsSyntaxNode};
+use rome_rowan::{AstNode, SyntaxKind, SyntaxSlot};
+
+use crate::Parse;
+
+// This check is used in the parser test to ensure it doesn't emit
+// unknown nodes without diagnostics, and in the analyzer tests to
+// check the syntax trees resulting from code actions are correct
```
MichaReiser, line 11:
```suggestion
/// This check is used in the parser test to ensure it doesn't emit
/// unknown nodes without diagnostics, and in the analyzer tests to
/// check the syntax trees resulting from code actions are correct
```

MichaReiser (approved):

crates/rome_js_parser/src/test_utils.rs
```diff
@@ -0,0 +1,55 @@
...
+            return descendant
+                .slots()
+                .any(|slot| matches!(slot, SyntaxSlot::Empty));
+        }
+
+        false
+    })
+}
+
+pub fn assert_errors_are_absent<T>(program: &Parse<T>, path: &Path)
```
ematipico, line 29: Can we also document this function?

crates/rome_js_analyze/src/utils.rs
leops: I added a constraint on `From` because that's the trait I'm using for conversion in the function body, and while `A: From<B>` implies `B: Into<A>` the opposite is not true. The reason I'm using `From` is because it explicitly spells out the type `N` should be converted into (while calling `.into()` is implicit), however in practice the type inference in Rust would probably be able to figure this out automatically

PR #2985, opened 2022-08-02
feat(rome_js_analyze): add an `inlineVariable` code action

leops:
## Summary

This PR introduces a simple `inlineVariable` code assist, available on all variable declarations that only have have read-only references. The rule inlines the initializer expression for that variable into all the places it's used and removes the original variable declaration node, using logic that's largely shared with the existing `noShoutyConstants` rule

## Test Plan

I've added a test file for the rule verifying that "read-only variables" can be inlined while writable variables are not

### DIFF
--- a/crates/rome_js_analyze/src/assists/js/inline_variable.rs
+++ b/crates/rome_js_analyze/src/assists/js/inline_variable.rs
@@ -1,0 +1,90 @@
+use rome_analyze::{context::RuleContext, declare_rule, ActionCategory, Rule, RuleCategory};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_semantic::{AllReferencesExtensions, Reference};
+use rome_js_syntax::{
+    JsAnyBinding, JsAnyBindingPattern, JsAnyExpression, JsIdentifierExpression,
+    JsVariableDeclarator,
+};
+use rome_rowan::{BatchMutationExt, SyntaxNodeCast};
+
+use crate::{semantic_services::Semantic, utils::remove_declarator, JsRuleAction};
+
+declare_rule! {
+    /// Provides a refactor to inline variables
+    ///
+    /// ## Examples
+    ///
+    /// ```js
+    /// let variable = expression();
+    /// statement(variable);
+    /// ```
+    pub(crate) InlineVariable {
+        version: "0.8.0",
+        name: "inlineVariable",
+        recommended: false,
+    }
+}
+
+impl Rule for InlineVariable {
+    const CATEGORY: RuleCategory = RuleCategory::Action;
+
+    type Query = Semantic<JsVariableDeclarator>;
+    type State = (Vec<Reference>, JsAnyExpression);
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let semantic_model = ctx.model();
+        let declarator = ctx.query();
+
+        let id = declarator.id().ok()?;
+        let binding = match id {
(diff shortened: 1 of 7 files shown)


xunilrj:
Should we limit this to only one read reference?
Because if the "initializer" has side effects, we will be changing the code semantic.

leops:
> Should we limit this to only one read reference? Because if the "initializer" has side effects, we will be changing the code semantic.

I think for refactors we can be less strict about semantics and side-effects compared to quick fixes: they never get applied automatically, and don't have the same implied "urgency" since they're not attached to an outstanding diagnostic. The code action only gets executed in response to a very clear action intent from the user (select the variable, click on "Refactor", click on "Inline Variable") so at this point we can trust that this is really what they want to do, and they know what they're doing

crates/rome_js_analyze/src/assists/js/inline_variable.rs
```diff
@@ -0,0 +1,90 @@
...
+        name: "inlineVariable",
+        recommended: false,
+    }
+}
+
+impl Rule for InlineVariable {
+    const CATEGORY: RuleCategory = RuleCategory::Action;
+
+    type Query = Semantic<JsVariableDeclarator>;
+    type State = (Vec<Reference>, JsAnyExpression);
```
ematipico, line 33: Could you document the `State` please?

crates/rome_js_analyze/tests/specs/js/inlineVariable.js
```diff
@@ -0,0 +1,8 @@
+let inlinable = "value1";
```
ematipico, line 1: Would it be possible to add more cases to cover this rule? For example we could add to the test case the snippet we added in the documentation

ematipico:
> > Should we limit this to only one read reference? Because if the "initializer" has side effects, we will be changing the code semantic.
> 
> I think for refactors we can be less strict about semantics and side-effects compared to quick fixes: they never get applied automatically, and don't have the same implied "urgency" since they're not attached to an outstanding diagnostic. The code action only gets executed in response to a very clear action intent from the user (select the variable, click on "Refactor", click on "Inline Variable") so at this point we can trust that this is really what they want to do, and they know what they're doing

While this is true, there's also the fact that we can't preview a refactor. It's my understanding that, as for today, these refactors are available only via LSP? 

I share @xunilrj 's concerns here, and I think we should add more test cases, to make sure that we emit the correct code without breaking semantics.

leops:
> While this is true, there's also the fact that we can't preview a refactor. It's my understanding that, as for today, these refactors are available only via LSP?

Yes refactor actions can only be used through the language server integration, since they're emitted contextually for the code the user has currently selected. How these actions are presented to the user depends on the actual editor being used, but VSCode has a "Refactor Preview" panel that shows the changes about to be applied for instance:

code_action_preview

> I share @xunilrj 's concerns here, and I think we should add more test cases, to make sure that we emit the correct code without breaking semantics.

I added a few more test cases (and ended up exposing a bug in the way lint rules are filtered in the code actions query)

crates/rome_js_analyze/src/assists/js/inline_variable.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Provides a refactor to inline variables
+    ///
+    /// ## Examples
+    ///
+    /// ```js
+    /// let variable = expression();
+    /// statement(variable);
+    /// ```
+    pub(crate) InlineVariable {
+        version: "0.8.0",
```
ematipico, line 23: This should be `0.9.0` now

ematipico (approved):

PR #2986, opened 2022-08-02
feat(rome_js_formatter): Format `throw` and `return` argument

MichaReiser:
This PR ensures that `throw` and `return` add parentheses around the argument for binary like expressions, sequence expressions and if the argument has any leading comments.

## Testing

 I manually reviewed all snapshot cases. The differences to prettier are around

 * formatting of binary like expressions -> separate PR
 * when to add or omit parentheses -> not part of this PR
 * comments -> requires the comments refactoring.

File Based Average Prettier Similarity: 77.98% -> 78.02%
Line Based Average Prettier Similarity: 73.23% -> 73.20%

The main reason why line based regresses is that the parenthesized / comments issues tend to spawn over multiple lines. However, these must be addressed as part of their own PR as they require significant architectural changes.

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
@@ -1,5 +1,8 @@
 use crate::prelude::*;
-use crate::utils::{is_simple_expression, FormatPrecedence};
+use crate::utils::{
+    binary_argument_needs_parens, is_simple_expression, FormatPrecedence,
+    JsAnyBinaryLikeLeftExpression,
+};
 use rome_formatter::write;
 
 use crate::utils::JsAnyBinaryLikeExpression;
@@ -36,7 +39,7 @@ impl FormatNodeRule<JsParenthesizedExpression> for FormatJsParenthesizedExpressi
                 write![f, [l_paren_token.format()]]?;
             };
 
-            write![f, [expression.format(),]]?;
+            write![f, [expression.format()]]?;
 
             if parenthesis_can_be_omitted {
                 write!(f, [format_removed(&r_paren_token?)])?;
@@ -49,7 +52,7 @@ impl FormatNodeRule<JsParenthesizedExpression> for FormatJsParenthesizedExpressi
                 f,
                 [
                     format_removed(&l_paren_token?),
-                    group(&expression.format()),
+                    expression.format(),
                     format_removed(&r_paren_token?),
                 ]
             ]?;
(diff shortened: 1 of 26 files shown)


crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -156,34 +172,27 @@ fn parenthesis_can_be_omitted(node: &JsParenthesizedExpression) -> SyntaxResult<
...
-            Ok(!JsAnyBinaryLikeExpression::can_cast(left.syntax().kind())
-                && !JsAnyBinaryLikeExpression::can_cast(right.syntax().kind()))
-        }
 
-        JsAnyExpression::JsLogicalExpression(expression) => {
-            let left = expression.left()?;
-            let right = expression.right()?;
+    if let Some(parent) = &parent {
+        if JsAnyBinaryLikeExpression::can_cast(parent.kind()) {
+            let binary_like = JsAnyBinaryLikeExpression::unwrap_cast(parent.clone());
```
ematipico, line 185: What's the reason of using the combo `can_cast`/`unwrap_cast` VS 
```rs
if let Some(binary_like) = JsAnyBinaryLikeExpression::cast(parent.clone()) {}
```
?

crates/rome_js_formatter/src/js/expressions/sequence_expression.rs
```diff
@@ -10,61 +10,58 @@ pub struct FormatJsSequenceExpression;
...
         } = node.as_fields();
 
-        let format_content = format_with(|f| {
-            write!(f, [left.format(), comma_token.format()])?;
+        let mut is_nested = false;
+        let mut first_non_sequence_or_paren_parent = None;
 
-            let format_right =
-                format_with(|f| write!(f, [soft_line_break_or_space(), right.format()]));
+        for parent in node.syntax().ancestors().skip(1) {
```
ematipico, line 22: Would you mind adding a comment explaining the `skip(1)`?

crates/rome_js_formatter/src/js/statements/return_statement.rs
```diff
@@ -15,32 +19,94 @@ impl FormatNodeRule<JsReturnStatement> for FormatJsReturnStatement {
...
+                    if_group_breaks(&text(")"))
+                ])]
+            )
+        } else {
+            write!(f, [argument.format()])
+        }
+    }
+}
+
+fn has_argument_leading_comments(argument: &JsAnyExpression) -> SyntaxResult<bool> {
```
ematipico, line 81: nit: `argument_has_leading_comments` / `has_leading_comments_argument` , it took me a bit to understand the name of the function

ematipico (approved):

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
MichaReiser: It avoids cloning the node in the cases where the parent isn't a binary-like expression. But let me see if this is even visible in profiling or if rust is able to optimize the clone away.

crates/rome_js_formatter/src/js/statements/return_statement.rs
MichaReiser: I can see how this is hard to understand but I prefer the name as it is because it starts with `has` and I do find `has_leading_comments_argument` more difficult to understand. But I'll add a comment explaining what the function is about. I hope that will help.

Issue #2979, opened 2022-08-01
noUnusedVariables


xunilrj:
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noUnusedVariables.md

xunilrj:
This rule is incorrectly flagging:

- [ ] Parameters in Typescript abstract classes and other signature methods
- [ ] Exported functions

xunilrj:
Given some comments on https://github.com/rome/tools/pull/3071. We may break ```noUnusedVariables``` into two rules: one checking names/bindings, another one checking values.

ematipico:
I am going to close this rule as implemented, and take note that in the next iteration we should split the rule in two.

PR #2987, opened 2022-08-02
feat(rome_js_analyze): `noUnusedVariable` rule

xunilrj:
## Summary

Closes https://github.com/rome/tools/issues/2979.

## Test Plan

```
>  cargo run -p rome_cli -- check crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js 
```

### DIFF
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
@@ -4,14 +4,12 @@ use rome_analyze::{
 };
 use rome_console::markup;
 use rome_diagnostics::Applicability;
-use rome_js_semantic::{AllReferencesExtensions, Reference, SemanticScopeExtensions};
+use rome_js_semantic::{AllReferencesExtensions, SemanticScopeExtensions};
 use rome_js_syntax::{
-    JsAnyExpression, JsAnyLiteralExpression, JsAnyRoot, JsFormalParameter, JsFunctionDeclaration,
-    JsIdentifierBinding, JsIdentifierExpression, JsLanguage, JsStringLiteralExpression,
-    JsSyntaxKind, JsVariableDeclaration, JsVariableDeclarator, JsVariableDeclaratorList,
-    JsVariableStatement,
+    JsFormalParameter, JsFunctionDeclaration, JsIdentifierBinding, JsSyntaxKind,
+    JsVariableDeclarator,
 };
-use rome_rowan::{AstNode, AstSeparatedList, BatchMutation, BatchMutationExt, SyntaxNodeCast};
+use rome_rowan::{AstNode, BatchMutationExt};
 
 declare_rule! {
     /// Disallow unused variables.
@@ -116,7 +114,7 @@ impl Rule for NoUnusedVariables {
                 // Another possibility is if all its references are "inside" the same declaration
                 if let Some(declarator) = declarator.as_ref() {
                     let node = declarator.syntax();
-                    if r.node().ancestors().find(|n| n == node).is_some() {
+                    if r.node().ancestors().any(|n| n == *node) {
                         continue;
                     }
                 }
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
@@ -1,0 +1,181 @@
+use crate::{semantic_services::Semantic, utils::batch::JsBatchMutation, JsRuleAction};
+use rome_analyze::{
+    context::RuleContext, declare_rule, ActionCategory, Rule, RuleCategory, RuleDiagnostic,
+};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_semantic::{AllReferencesExtensions, Reference, SemanticScopeExtensions};
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyLiteralExpression, JsAnyRoot, JsFormalParameter, JsFunctionDeclaration,
+    JsIdentifierBinding, JsIdentifierExpression, JsLanguage, JsStringLiteralExpression,
+    JsSyntaxKind, JsVariableDeclaration, JsVariableDeclarator, JsVariableDeclaratorList,
+    JsVariableStatement,
+};
+use rome_rowan::{AstNode, AstSeparatedList, BatchMutation, BatchMutationExt, SyntaxNodeCast};
+
+declare_rule! {
+    /// Disallow unused variables.
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// const a = 4;
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// let a = 4;
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// function foo(myVar) {
+    ///     console.log('foo');
+    /// };
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// function foo(myVar) {
+    ///     console.log('foo');
+    /// }
+    /// foo();
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// const foo = (b) => {
+    ///     console.log('foo');
+    /// };
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// function foo() {
+    ///     foo();
+    /// }
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// const foo = () => {
+    ///     foo();
+    ///     console.log(this);
+    /// };
+    /// ```
+    ///
+    /// # Valid
+    ///
+    /// ```js
+    /// function foo(, b) {
+    ///     console.log(b)
+    /// };
+    /// foo();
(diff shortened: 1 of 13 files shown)


xunilrj:
!bench_analyzer

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
```diff
@@ -0,0 +1,249 @@
...
+  │          --
+
+Suggested fix: Remove dead code.
+    | @@ -4,10 +4,6 @@
+3 3 |   
+4 4 |   function f1() {}
+5 5 |   
+6   | - function f2() {
+7   | - 	f2();
+8   | - }
```
ematipico, line 119: I think this incorrect? `f2` is actually used, inside itself (recursion)

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
```diff
@@ -0,0 +1,249 @@
...
+ 7  7 |   	f2();
+ 8  8 |   }
+ 9  9 |   
+10    | - function f3() {
+11    | - 	function g() {
+12    | - 		f3();
+13    | - 	}
+14    | - 	g();
+15    | - }
+16    | -
```
ematipico, line 148: `f3` is actually used, so we can't remove it.

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
```diff
@@ -0,0 +1,249 @@
...
+=  note: Unused variables usually are result of incomplete refactoring, typos and other source of bugs.
+
+
+```
+
+```
+warning[js/noUnusedVariables]: This parameter is unused.
+   ┌─ noUnusedVariables.js:21:17
+   │
+21 │ function f42(a, b) {
```
ematipico, line 185: Could we add a test case where we have a trailing comma? `function f(a,b,){}`

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
```diff
@@ -0,0 +1,249 @@
...
+28 │ const f6 = () => {
+   │       --
+
+Suggested fix: Remove dead code.
+      | @@ -24,7 +24,3 @@
+23 23 |   f42();
+24 24 |   
+25 25 |   const f5 = () => {};
+26    | - 
+27    | - const f6 = () => {
```
ematipico, line 240: `f6` is actually used

ematipico (changes requested):
There are few issues with functions, we have diagnostics when in reality they are valid cases.

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -0,0 +1,177 @@
...
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let binding = ctx.query();
+        let model = ctx.model();
+
+        let all_references = binding.all_references(model);
+
+        if all_references.count() == 0 {
+            Some(())
```
ematipico, line 88: Why `Some` instead of `None`?

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
xunilrj: This is on purpose because is not called from the outside.

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
xunilrj: Same thing here. ```f3``` is never called from the outside.

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
xunilrj: If there are no references, so it is unused.

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -0,0 +1,177 @@
...
+        if let Some(declaration) = binding.parent::<JsFunctionDeclaration>() {
+            batch.remove_node(declaration)
+        } else if let Some(variable_declarator) = binding.parent::<JsVariableDeclarator>() {
+            batch.remove_js_variable_declarator(&variable_declarator);
+        } else if let Some(formal_parameter) = binding.parent::<JsFormalParameter>() {
+            batch.remove_js_formal_parameter(&formal_parameter);
+        }
+
+        Some(JsRuleAction {
+            category: ActionCategory::Refactor,
```
ematipico, line 171: Why is this a `Refactor` and not a `QuickFix`? Just asking because I don't know yet the difference

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -0,0 +1,177 @@
...
+            applicability: Applicability::Unspecified,
+            message: markup! { "Remove dead code." }.to_owned(),
```
ematipico, line 173: The message might be misleading, because we also have a rule called `noDeadCode`. Would it be possible to have better messaging?

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
xunilrj: I create more tests for the "remove_*" methods, instead of polluting this one here.
https://github.com/rome/tools/pull/2987/files#diff-0902796fee286e39e6f27f612b14a08a77e0af0dee55c59871e1a0292707c0faR110

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -0,0 +1,177 @@
...
+            batch.remove_node(declaration)
+        } else if let Some(variable_declarator) = binding.parent::<JsVariableDeclarator>() {
+            batch.remove_js_variable_declarator(&variable_declarator);
+        } else if let Some(formal_parameter) = binding.parent::<JsFormalParameter>() {
+            batch.remove_js_formal_parameter(&formal_parameter);
+        }
+
+        Some(JsRuleAction {
+            category: ActionCategory::Refactor,
+            applicability: Applicability::Unspecified,
```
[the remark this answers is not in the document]
xunilrj: > what does it mean for the end user? Can an user apply the change via CLI?
No.

I am torn apart between ```Unspecified```and ```MaybeIncorrect```. I assumed that removing unused code is not the "best" solution.

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
xunilrj: You are right. I think "Quickfix" makes more sense.

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
xunilrj: Done.

crates/rome_js_analyze/tests/specs/js/noUnusedVariables.js.snap
xunilrj: Same thing as above. It is only used internally. Which means it is never called.

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// # Valid
+    ///
+    /// ```js
+    /// function foo(b) {
+    ///     console.log(b)
+    /// };
+    /// foo();
+    /// ```
+    pub(crate) NoUnusedVariables {
+        version: "0.8.0",
```
ematipico, line 68: This should be `0.9.0` now

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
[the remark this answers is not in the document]
leops: The semantics of `Applicability` is something we'll need to review when we rework the diagnostics, the current version of the enum is inherited directly from `rslint` and indirectly `rustc`, but those tools are using text-based suggestions instead of tree-based.
In this case I think the `Unspecified` applicability comes from `rustc`, where the concept of "applicability" was added when the compiler already has a bunch of existing suggestions, so all those diagnostics were set to `Unspecified` in the refactor to be manually reviewed later on.
In our case though we should always document if a suggestion is safe to apply or not, so in practice the only variants of `Applicability` we should be using a `MaybeIncorrect` and `Always` (the difference is that `Unspecified` can never be applied automatically, while `MaybeIncorrect` can after the has the user has reviewed it)

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
xunilrj: Done.

ematipico (approved):
Looks good to me. We should address the applicability of the rule.

PR #2954, opened 2022-07-28
fix(rome_js_parser): Better error recover for jsx attribute #2944

IWANABETHATGUY:
## Summary
1. Resolved #2944

## Test Plan
1. Should pass all test cases newly added.

PR #2989, opened 2022-08-02
fix: broken ci

IWANABETHATGUY:
## Summary
1. Disable the case that has syntax error due to broken CI
2. Diable for now, recover the case later if we have solution to testing analyzer that have syntax error.

## Test Plan
1. CI should pass.


ematipico:
Where did we introduce this error?

IWANABETHATGUY:
> Where did we introduce this error?

https://github.com/rome/tools/pull/2954 here, 
I want to verify the unexpected code action is no more exist, so I add this test case that has syntax error.

ematipico (approved):

PR #2990, opened 2022-08-02
feat(rome_js_formatter): Function parameter & return type grouping

MichaReiser:
## Summary

This PR refines the formatting of function parameters and return type annotations by adopting Prettier's behaviour to first break the return type annotation in some situations before breaking the parentheses (see `shouldGroupCallArguments`). 

Implementing this change required to extract the shared method formatting logic into a `FormatMethodMember` helper struct.

## Test Plan

Verified that e.g. object types now break before the parentheses. 

File Based Average Prettier Similarity: 78.03% -> 78.10%
Line Based Average Prettier Similarity: 73.10% -> 73.19%


MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/js/classes/method_class_member.rs
```diff
@@ -1,44 +1,217 @@
...
+                space(),
+                FormatMethodMember::from(node.clone())
+            ]
+        ]
+    }
+}
+
+declare_node_union! {
+    /// Formats the type parameters, parameters, and return type annotation of a method
+    pub(crate) FormatMethodMember =
```
ematipico, line 31: nit: `JsAnyFormatMethodMember` ? It's mostly to use one naming convention.

crates/rome_js_formatter/src/js/declarations/function_declaration.rs
```diff
@@ -120,22 +135,92 @@ impl Format<JsFormatContext> for FormatFunction {
...
-        write!(f, [self.body().format()])
+        if let Some(body) = self.body()? {
+            write!(f, [space(), body.format()])?;
+        }
+
+        Ok(())
     }
 }
+
+// Grouping the parameters has the effect that the return type will break first.
```
ematipico, line 176:
```suggestion
/// Grouping the parameters has the effect that the return type will break first.
```

crates/rome_js_formatter/src/js/declarations/function_declaration.rs
```diff
@@ -120,22 +135,92 @@ impl Format<JsFormatContext> for FormatFunction {
...
+pub(crate) fn should_group_function_parameters(
+    type_parameters: Option<&TsTypeParameters>,
+    parameter_count: usize,
+    return_type: Option<SyntaxResult<TsAnyReturnType>>,
+    formatted_return_type: &mut Memoized<impl Format<JsFormatContext>, JsFormatContext>,
```
ematipico, line 181: First time I see this pattern, and it seems weird (anti-pattern?). Do you foresee future cases where we might need to use this pattern?

crates/rome_js_formatter/src/js/declarations/function_declaration.rs
```diff
@@ -120,22 +135,92 @@ impl Format<JsFormatContext> for FormatFunction {
...
+            _ => return Ok(false),
+        }
+    }
+
+    let result = if parameter_count != 1 {
+        false
+    } else {
+        // THIS is a hack that is necessary to avoid that the formatter doesn't insert a space
+        // between `)` and the `:` of the return type annotation IF there's a an inline comment
+        // after the last comment that has been written. This can be deleted once the comments refactor lands.
```
ematipico, line 211: Is it worth adding a `TODO` with a link to the issue of the refactor?

crates/rome_js_formatter/src/js/declarations/function_declaration.rs
```diff
@@ -120,22 +135,92 @@ impl Format<JsFormatContext> for FormatFunction {
...
+        let is_last_content_inline_comment = f.state().is_last_content_inline_comment();
+        f.state_mut().set_last_content_inline_comment(false);
+
+        let group = matches!(
+            return_type,
+            TsAnyReturnType::TsType(TsType::TsObjectType(_) | TsType::TsMappedType(_))
+        ) || formatted_return_type.inspect(f)?.will_break();
```
ematipico, line 218: Could you update this [line](https://github.com/rome/tools/blob/main/crates/rome_formatter/src/buffer.rs#L489), this [line](https://github.com/rome/tools/blob/main/crates/rome_formatter/src/buffer.rs#L542-L545) - need to fix the markdown title too, it's commented, and add the try operator, please?

ematipico (approved):

crates/rome_js_formatter/src/js/declarations/function_declaration.rs
MichaReiser: Do you mean the need for passing in "Memoized" or the use of `impl` vs a generic type parameter? 

* Memoized: I need to inspect the content. There's no other way of doing this than passing in the memorized document
* `impl` vs generic type parameter. This are the same

PR #2992, opened 2022-08-02
feat: add no extra boolean cast

ktfth:
Description
From umbrella ticket https://github.com/rome/tools/issues/2642

[noExtraBooleanCast](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noExtraBooleanCast.md)
Disallow unnecessary boolean casts
Example fix:
if (Boolean(foo)) {} -> if (foo) {}


IWANABETHATGUY:
https://github.com/rome/tools/pull/2868,  seems duplicate.

ematipico:
Thank you @ktfth for your contribution, but unfortunately there's already a PR to implement this rule.

I suggest to check first if a rule hasn't been claimed/assigned by anyone, so we can avoid duplicates. In the umbrella issue we also wrote instructions of how to claim a rule.

PR #2993, opened 2022-08-03
feat(rome_js_formatter): Hug function parameters

MichaReiser:
This PR implements the heuristic for when we want that the `(` and `)` *hug* the first parameter. This is desired when:

* It is a test function
* The function has a single parameter AND we want that the first token of that parameter is printed on the same line as `(`. This is the case for arrays and object literals OR identifiers where the type annotation is an object type

This PR further unifies the formatting logic of normal parameters and constructor parameters. 

This PR **does not** implement the special formatting of last argument yet.

## Test Plan

File Based Average Prettier Similarity: 78.10% -> 78.53%
Line Based Average Prettier Similarity: 73.19% -> 74.03%


MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/js/bindings/parameters.rs
```diff
@@ -1,22 +1,229 @@
...
-            l_paren_token,
-            items,
-            r_paren_token,
-        } = node.as_fields();
+        FormatParameters::from(node.clone()).fmt(f)
+    }
+}
+
+declare_node_union! {
+    pub(crate) FormatParameters = JsParameters | JsConstructorParameters
```
ematipico, line 23: nit: I think we should follow one "naming style" across the board. There are other unions in the formatter that were called with the prefix `JsAny*`.

crates/rome_js_formatter/src/js/bindings/parameters.rs
```diff
@@ -1,22 +1,229 @@
...
     }
 }
+
+#[derive(Copy, Debug, Clone, Eq, PartialEq)]
+pub enum ParameterLayout {
+    /// Enforce that the opening and closing parentheses aren't separated from the first token of the parameter.
+    /// For example, to enforce that the `{`  and `}` of an object expression are formatted on the same line
+    /// as the `(` and `)` tokens even IF the object expression itself breaks across multiple lines.
+    Hug,
+    Default,
```
ematipico, line 138: How's the default layout? Should we explain it?

crates/rome_js_formatter/src/js/bindings/parameters.rs
```diff
@@ -1,22 +1,229 @@
...
+    }
+
+    fn r_paren_token(&self) -> SyntaxResult<JsSyntaxToken> {
+        match self {
+            FormatParameters::JsParameters(parameters) => parameters.r_paren_token(),
+            FormatParameters::JsConstructorParameters(parameters) => parameters.r_paren_token(),
+        }
+    }
+
+    fn is_in_test_call(&self) -> SyntaxResult<bool> {
```
ematipico, line 100: Could we pair the function with a piece of documentation that shows a snippet of a test call?

It's mostly for helping to understand the actual "shape" of the AST, because by looking at the code I can't understand it.

crates/rome_js_formatter/src/js/bindings/parameters.rs
```diff
@@ -1,22 +1,229 @@
...
+#[derive(Copy, Debug, Clone, Eq, PartialEq)]
+pub enum ParameterLayout {
+    /// Enforce that the opening and closing parentheses aren't separated from the first token of the parameter.
+    /// For example, to enforce that the `{`  and `}` of an object expression are formatted on the same line
+    /// as the `(` and `)` tokens even IF the object expression itself breaks across multiple lines.
+    Hug,
+    Default,
+}
+
+fn should_hug_function_parameters(parameters: &FormatParameters) -> FormatResult<bool> {
```
ematipico, line 141: I think it will useful to list the conditions of when a function parameter should be/not be hugged

crates/rome_js_formatter/src/js/bindings/parameters.rs
```diff
@@ -1,22 +1,229 @@
...
+    if list.len() != 1 {
+        return Ok(false);
+    }
+
+    if parameters.r_paren_token()?.has_leading_comments() {
+        return Ok(false);
+    }
+
+    // SAFETY: Safe because of the length check above
+    let only_parameter = list.first().unwrap()?;
```
ematipico, line 158: Where do we check if `list.len()` is greater than 0? Asking because `!= 1` is not a safe check

crates/rome_js_formatter/src/js/lists/parameter_list.rs
```diff
@@ -8,23 +14,180 @@ impl FormatRule<JsParameterList> for FormatJsParameterList {
...
+
+impl AnyParameterList {
+    pub fn len(&self) -> usize {
+        match self {
+            AnyParameterList::JsParameterList(parameters) => parameters.len(),
+            AnyParameterList::JsConstructorParameterList(parameters) => parameters.len(),
+        }
+    }
+
+    pub fn first(&self) -> Option<SyntaxResult<AnyParameter>> {
```
ematipico, line 135: It's a bit weird seeing `Option<Result<>>`. I would have expected a `Result<Option<>>`, so we can handle the error earlier, then unwrap `Option` knowing that the error has been handled already. Can't we flip it with [`transpose()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose)?

ematipico (approved):
No particular concerns, mostly questions and documentation suggestions

crates/rome_js_formatter/src/js/bindings/parameters.rs
MichaReiser: The function early returns if the length if not equal to 1, or in other words, if the list contains zero or two or more elements. 

```
    if list.len() != 1 {
        return Ok(false);
    }
```

This guarantees that the list always contains exactly one element.

crates/rome_js_formatter/src/js/lists/parameter_list.rs
MichaReiser: This is in line with all other `AstSeparatedList` methods.

crates/rome_js_formatter/src/js/bindings/parameters.rs
ematipico: Completely missed that! Thank you

crates/rome_js_formatter/src/js/bindings/parameters.rs
MichaReiser: I renamed them all to `FormatJsAny*` to align with the generated formatter rules

crates/rome_js_formatter/src/js/bindings/parameters.rs
MichaReiser: I prefer not to explicitly list the same conditions as in the implementation but I referred to the `Hug` layout documentation and extended it with an example and when it's desired to "hug" a parameter.

PR #2995, opened 2022-08-03
feat(rome_formatter): `Align` IR element

MichaReiser:
This PR adds an alignment `FormatElement` that allows explicitly aligning content by a specific amount of spaces rather than using `indent` that uses the configured indent string.

Aligning by an explicit number of spaces is e.g. needed when formatting conditional expressions where the content should be aligned to `"? "` and `: `.

## Prettier compatibility 

Prettier's  `align` IR supports not only positive numbers but also 
* negative numbers (for dedenting)
* NEGATIVE_INFINITE: TO dedent to root
* strings: To use a specific indent string sequence other than spaces

I'll add the functionality once needed (NEGATIVE_INFINITE will be needed for template literals)

## Tests

I added a new doc test that demonstrates the new IR element.

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -8,6 +8,7 @@ use rome_rowan::{Language, SyntaxNode, SyntaxToken, SyntaxTokenText, TextLen};
 use std::borrow::Cow;
 use std::cell::Cell;
 use std::marker::PhantomData;
+use std::num::NonZeroU8;
 use std::ops::Deref;
 
 /// A line break that only gets printed if the enclosing `Group` doesn't fit on a single line.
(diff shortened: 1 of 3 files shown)


MichaReiser:
!bench_formatter

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
             mode: PrintMode::Expanded,
         }
     }
 }
 
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+enum Indent {
+    /// Indent by the specified count of indention levels
+    Simple(u16),
+    /// Indent `indent` indention levels and add `align` spaces.
```
ematipico, line 583: > Indent `indent` indention

LOL! Any chance we can rephrase it a bit? Like:
> Indents the given `indent`, by the specified indentation levels. It then adds the spaces specified by `align`

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
+    IndentAlign { indent: u16, align: NonZeroU8 },
+}
+
+impl Indent {
+    const fn is_empty(&self) -> bool {
+        matches!(self, Indent::Simple(0))
+    }
+
+    const fn empty() -> Indent {
```
ematipico, line 592: nit: `reset` or `start`? It feels weird talking of an "empty indentation"

crates/rome_formatter/src/builders.rs
```diff
@@ -664,6 +665,93 @@ impl<Context> std::fmt::Debug for Indent<'_, Context> {
...
+{
+    Align {
+        count: NonZeroU8::new(count).expect("Alignment count must be a non-zero number."),
+        content: Argument::new(content),
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct Align<'a, Context> {
+    count: NonZeroU8,
```
ematipico, line 724: Is there a particular reason why we limit align to be a `u8`?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
+            Indent::IndentAlign { align, .. } => (*align).into(),
+        }
+    }
+
+    /// Indents by one additional level
+    fn indented(self) -> Self {
+        match self {
+            Indent::Simple(count) => Indent::Simple(count + 1),
+            // Increase the indent AND convert the align to an indent
+            Indent::IndentAlign { indent, .. } => Indent::Simple(indent + 2),
```
ematipico, line 615: Even with comment, I can't understand why it's `+ 2`.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
+
+    fn align(&self) -> u8 {
+        match self {
+            Indent::Simple(_) => 0,
+            Indent::IndentAlign { align, .. } => (*align).into(),
+        }
+    }
+
+    /// Indents by one additional level
+    fn indented(self) -> Self {
```
ematipico, line 611: nit: `into_indent`?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
+        match self {
+            Indent::Simple(count) => Indent::Simple(count + 1),
+            // Increase the indent AND convert the align to an indent
+            Indent::IndentAlign { indent, .. } => Indent::Simple(indent + 2),
+        }
+    }
+
+    fn aligned(self, count: NonZeroU8) -> Self {
```
ematipico, line 619: nit: `into_align`?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
+        match self {
+            Indent::Simple(indent_count) => Indent::IndentAlign {
+                indent: indent_count,
+                align: count,
+            },
+            // Convert the existing align to an indent
+            Indent::IndentAlign { indent, .. } => Indent::IndentAlign {
+                indent: indent + 1,
```
ematipico, line 627: I can't understand why there's a `+ 1`

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -556,12 +570,111 @@ impl PrintElementArgs {
...
+                }
+            }
+            _ => return other.partial_cmp(self).map(|ordering| ordering.reverse()),
+        })
+    }
+}
+
+impl Default for Indent {
+    fn default() -> Self {
+        Indent::Simple(0)
```
ematipico, line 674: It doesn't make much sense to have a default and also a const function called `empty` that returns `Indent::Simple(0)`. Should we stick to just one?

crates/rome_formatter/src/printer/mod.rs
ematipico:
```suggestion
    const fn empty() -> Self {
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -766,8 +881,10 @@ fn fits_element_on_line<'a, 'rest>(
         ),
 
         FormatElement::Text(token) => {
-            state.line_width += state.pending_indent as usize * options.indent_string.len();
-            state.pending_indent = 0;
+            state.line_width += state.pending_indent.indent() as usize
+                * options.indent_string.len()
+                + state.pending_indent.align() as usize;
+            state.pending_indent = Indent::default();
```
ematipico, line 887: Here we use `default`, and we have another case where use `Indent::empty`. I think we should stick to just one.

crates/rome_formatter/src/builders.rs
MichaReiser: I don't see a use case where you want to indent some content by more than 128 spaces. Most use cases will likely use the values 2 and 3 (if you use for, than you probably want to use `indent`).

This may change if we use `align` for markdown because it's then necessary to align by as many spaces as the user has configured but that's a problem for the future.

crates/rome_formatter/src/printer/mod.rs
MichaReiser: It converts the `align` to an indent (+1) and adds the indent as requested by the caller.

crates/rome_formatter/src/printer/mod.rs
MichaReiser: If you have `align(indent(a))`, then the printer handles this as `indent(indent(a))`. My guess is that this is mainly a performance optimisation because it would otherwise be required to track the exact nesting of indent/align which would make things unnecessarily complicated.

crates/rome_formatter/src/printer/mod.rs
MichaReiser: This isn't converting self to an `indent` but adds an indent level. Let me rename it to `increment_indent` and `increment_align`

ematipico: Oooh OK! I was also confused by the signature of the function. Thank you!

crates/rome_formatter/src/printer/mod.rs
ematipico: FYI: https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2889-2894

`default` uses `new`

leops (approved):

PR #3002, opened 2022-08-04
feat(rome_js_formatter): Binary parentheses

MichaReiser:
This PR refines the rules for setting parentheses around binary expression to more closely match Prettier's behaviour.

It does so by re-using `OperatorPrecedence` enum from the parser crate (which is why it is moved to `rome_js_syntax`).

## Tests

Reviewed the updates to the prettier snapshots.

**File Based Average Prettier Similarity**: 78.58%  -> 78.85%
**Line Based Average Prettier Similarity**: 74.00%   -> 74.25%


crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -177,41 +200,41 @@ pub(crate) enum BinaryLikeOperator {
...
+            JsAnyBinaryLikeLeftExpression::JsAnyExpression(JsAnyExpression::JsLogicalExpression(_))
+        )
+    {
+        return Ok(parent_operator != current_operator);
+    }
+
+    let current_precedence = current_operator.precedence();
+    let parent_precedence = parent_operator.precedence();
+
+    #[allow(clippy::if_same_then_else)]
```
ematipico, line 224: I suppose you added this because it helps for readability?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -831,30 +854,46 @@ impl JsAnyBinaryLikeExpression {
...
-                    instance_of.left()?,
-                    JsAnyExpression::JsInstanceofExpression(_)
-                )
-            }
-            JsAnyBinaryLikeExpression::JsInExpression(in_expression) => matches!(
-                in_expression.property()?,
-                JsAnyInProperty::JsAnyExpression(JsAnyExpression::JsInExpression(_))
-            ),
-        })
+        let left = self.left()?;
```
ematipico, line 857: Is the comment still valid? Should it be updated?

crates/rome_js_formatter/src/utils/binary_like_expression.rs
```diff
@@ -831,30 +854,46 @@ impl JsAnyBinaryLikeExpression {
...
+    } else if parent_precedence.is_equality() && child_precedence.is_equality() {
+        false
+    } else if (matches!(
+        child_operator,
+        BinaryLikeOperator::Binary(JsBinaryOperator::Remainder)
+    ) && parent_precedence.is_multiplicative())
+        || (matches!(
+            parent_operator,
+            BinaryLikeOperator::Binary(JsBinaryOperator::Remainder)
+        ) && child_precedence.is_multiplicative())
```
ematipico, line 887: The formatting here is really hard to follow for me 🤣 I wish `rustfmt` was at prettier level here. Just a rant, carry on!

ematipico (approved):
Great progresses! 🎆

crates/rome_js_formatter/src/utils/binary_like_expression.rs
MichaReiser: Yes, I find it easier to reason if every `if` handles a specific case rather than grouping them all together because the main point here isn't about the body (which is a simple bool) but about the condition.

PR #3007, opened 2022-08-04
feat(rome_formatter): `Dedent` IR

MichaReiser:
## Summary

This PR adds a new IR that allows to reduce the indention for some content nested inside of an `align` or `indent`.

The behaviour of the element depends on whether the closest ancestor is an `align` or `indent`

* `align`: Removes the `align` spaces
* `indent`: Reduces the indention level by one *EXCEPT* if the level is 0, in which case this is a no-op

This is necessary for the formatting of nested conditional expressions where the `consequent` gets wrapped in an `align` but any nested consequent should be indented by one level (and not an indent). That's why the formatter uses `dedent(indent(...))` to undo the `align` but enforce an `indent`.

## Test Plan

I added a new builder test showing the behaviour of the `dedent` element.

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -691,10 +691,14 @@ impl<Context> std::fmt::Debug for Indent<'_, Context> {
 ///             text("Indented, not aligned")
 ///         ]))
 ///     ]),
+///     dedent(&format_args![
+///         hard_line_break(),
+///         text("Dedent on root level is a no-op.")
+///     ])
 /// ]).unwrap();
 ///
 /// assert_eq!(
-///     "root\n  aligned\nnot aligned\n\tIndented, not aligned",
+///     "root\n  aligned\nnot aligned\n\tIndented, not aligned\nDedent on root level is a no-op.",
 ///     block.print().as_code()
 /// );
 /// ```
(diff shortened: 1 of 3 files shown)


crates/rome_formatter/src/builders.rs
```diff
@@ -665,6 +665,75 @@ impl<Context> std::fmt::Debug for Indent<'_, Context> {
     }
 }
 
+/// It reduces the indention for the given content depending on the closest [indent] or [align] parent element.
+/// * [align] Undoes the spaces added by [align]
+/// * [indent] Reduces the indention level by one
+///
+/// This is a No-op if the indention level is zero.
```
ematipico, line 672: Maybe we should create a test that shows that using `dedent` when the indentation level is zero won't overflow? For example:

```rs
dedent(dedent(dedent()))
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -551,13 +555,18 @@ impl PrintElementArgs {
         }
     }
 
-    pub fn indented(mut self) -> Self {
+    pub fn increment_indent_level(mut self) -> Self {
```
leops, line 558: Nit: unify the naming between `increment_indent_level` / `decrement_indent` (and `increment_level` / `decrement_indent`)

leops (approved):

PR #3014, opened 2022-08-05
feat(npm): first draft of Node.js API, frontend

ematipico:
## Summary

This PR closes https://github.com/rome/tools/issues/3013

This PR kicks off the code around the Node.js API. The design is still up to debate and change, but here's the list of things I've done.

### Single `Rome` class

Expose a single `Rome` class, from which someone can use different APIs. The reason why I wanted to created a class instead of single functions, is because we could leverage caching and other goodies. An example of what we can do with this design, is to persist the same format options across format runs:

```js
import { Rome } from "rome";

const rome = new Rome();

rome.format_options({
	quoteStyle: "single"
});
const source1 = `function a() {}`;
const source2 = `function b() {  return {} }`;
let result1 = rome.format(source1);
let result2 = rome.format(source2);
```

	  
Although it has the drawback that providing a new configuration via `rome.format(source, options)`, will override the one provided at the beginning. Personally I am not too concerned by this case, as the formatting should be consistent across runs, so I would see the first pattern more used. But this is up to debate and change, obviously.

### `format` function

To quickly format some content with the given options, which are optional.

### Workspace architecture

As discussed with @leops on a private channel, we thought that emulating the `Workspace` APIs might play well when we will wire the calls to WASM/daemon. This can set a good basis of how are APIs will look like, boundaries and less changes across the board;

## Test Plan

There aren't any tests just yet. I just added this source code to the workflow of our CI, and turned off a bunch false positives from our linter.


MichaReiser:
It would help me if we outline the purpose of these APIs because I can see two different use cases:

* single format / or check call, doesn't use Rome's workspace functionalty
* Using Rome's API to trigger different operations on the current workspace. 

I think using classes makes sense for the latter but feels a bit overkill to me for the former. Which problem are we trying to solve?

ematipico:
> It would help me if we outline the purpose of these APIs because I can see two different use cases:
> 
>     * single format / or check call, doesn't use Rome's workspace functionalty
> 
>     * Using Rome's API to trigger different operations on the current workspace.
> 
> 
> I think using classes makes sense for the latter but feels a bit overkill to me for the former. Which problem are we trying to solve?

We can expose both APIs - single functions and workspace, and gather feedback once we deliver a first prototype. 

One use case that I could see is the `check` functionality inside a webapck plugin, for instance: I am not familiar with webpack plugins, but I know that the compilation works on single files. Here using a workspace link approach could work, to persist caching/options.

Another case is code generation, where using function calls might work better. Also for quick prototyping.

MichaReiser:
> One use case that I could see is the check functionality inside a webapck plugin, for instance: I am not familiar with webpack plugins, but I know that the compilation works on single files. Here using a workspace link approach could work, to persist caching/options.

There's probably a lot that we would need to take into consideration when we intend to do that. 

For such use case, I would argue that a simple `check` or `format` function is easier to use. On top of that, tools like `webpack` or `parcel` often bring their own caching where it may be questionable to role out our own caching on top of that. 

That's why I think it is important to understand the use case of the API we plan to ship first because the two may have very different requirements. In my view, publishing a "simple" `format` and `check` API will be useful for many users and may be easier to integrate than a stateful `Rome` object that comes with caching. 

Integrating the stateful implementation may be desired in the long term, particularly once we have cross file analysis to get decent performance.

MichaReiser:
What would help me assess the API is if you could write some examples on how the API is being used. We can "make" assumptions that we have a `transform` function to if that simplifies things to come up with integrations. 

Something else that may be useful is to look at other APIs like Prettier to see how it is structured. Ideally, replacing Rome with Prettier should be a three line change
* one line in package json that replaces rome with prettier
* Import change
* call to Romejs

ematipico:
> What would help me assess the API is if you could write some examples on how the API is being used. We can "make" assumptions that we have a `transform` function to if that simplifies things to come up with integrations.
> 
> Something else that may be useful is to look at other APIs like Prettier to see how it is structured. Ideally, replacing Rome with Prettier should be a three line change
> 
>     * one line in package json that replaces rome with prettier
> 
>     * Import change
> 
>     * call to Romejs

Sure, let's asses the APIs in a different channel. I will add some examples in the main issue. I'd like to merge this PR to kick things off. Is there something that blocks the whole PR?

MichaReiser:
> > What would help me assess the API is if you could write some examples on how the API is being used. We can "make" assumptions that we have a `transform` function to if that simplifies things to come up with integrations.
> > Something else that may be useful is to look at other APIs like Prettier to see how it is structured. Ideally, replacing Rome with Prettier should be a three line change
> > ```
> > * one line in package json that replaces rome with prettier
> > 
> > * Import change
> > 
> > * call to Romejs
> > ```
> 
> Sure, let's asses the APIs in a different channel. I will add some examples in the main issue. I'd like to merge this PR to kick things off. Is there something that blocks the whole PR?

What's in your view the benefit of merging it now? 

My understanding is that this was a proposal in a PR form, it isn't used anywhere, isn't intended to be used, and we don't know if it fits our use cases. 

I haven't looked at the implementation yet because I'm not sure if it makes sense doing so if we haven't agreed on an API yet.

npm/rome/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+			return this.workspace.format_file({
+				path: TEMPORARY_CONTENT,
+			});
+		} catch (err) {
+			if (err instanceof FormatError) {
+				console.error(err.missing_content());
+			} else {
+				console.log("Unknown error");
+				console.log(err);
+			}
```
MichaReiser, line 87: We should propagate the exception instead of silently ignoring the error.

npm/rome/tsconfig.json
```diff
@@ -0,0 +1,103 @@
...
+
+    /* Emit */
+     "declaration": true,
+    // "declarationMap": true,                           /* Create sourcemaps for d.ts files. */
+    // "emitDeclarationOnly": true,                      /* Only output d.ts files and not JavaScript files. */
+     "sourceMap": true,                                /* Create source map files for emitted JavaScript files. */
+    // "outFile": "./",                                  /* Specify a file that bundles all outputs into one JavaScript file. If 'declaration' is true, also designates a file that bundles all .d.ts output. */
+     "outDir": "./dist",
+    // "removeComments": true,                           /* Disable emitting comments. */
+    // "noEmit": true,                                   /* Disable emitting files from a compilation. */
```
MichaReiser, line 54: Can we delete the unnecessary boilerplate? I find it very hard to find any real options.

ematipico:
Closing. Will reopen a new PR with only the workspace part.

PR #3016, opened 2022-08-05
refactor(rome_service): change how rules config is shaped

### LINKED ISSUE #3015: review rule configuration

ematipico:
## Summary

Closes #3015 

The first attempt of the configuration of the rules tried to be too smart (using `IndexMap`) but it wasn't well designed and it couldn't be extended easily and as a consequence we have `rules.js.rules`, which was very redundant.

Fortunately `flatten` directive form `serde` seems to fit the bill here, and it works as expected.

## Future optimization

At the moment the enabled/disabled rules are always computed every time we get a diagnostic from a rule. Same for when we have to create an `AnalysisFilter`. This is not efficient. We should look into ways to cache this information, although this would require some work around lifetimes, because `RuleFilter` only works with references and caching them might only be possible if we use the lifetime coming `App`. 

## Test Plan

Updated the tests, and now it all passes! We will need to change the configuration once this lands.

### DIFF
--- a/crates/rome_cli/tests/configs.rs
+++ b/crates/rome_cli/tests/configs.rs
@@ -36,16 +36,14 @@ pub const CONFIG_ALL_FIELDS: &str = r#"{
     "enabled": true,
     "rules": {
         "js": {
-            "rules": {
-                "noDeadCode": "off",
-                "useSimplifiedLogicExpression": "warn",
-                "noCatchAssign": "error",
-                "noLabelVar": {
-                    "level": "warn"
-                },
-                "useTemplate": {
-                    "level": "error"
-                }
+            "noDeadCode": "off",
+            "useSimplifiedLogicExpression": "warn",
+            "noCatchAssign": "error",
+            "noLabelVar": {
+                "level": "warn"
+            },
+            "useTemplate": {
+                "level": "error"
             }
         }
     }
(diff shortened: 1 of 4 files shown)


leops (commented):
Originally the point of using `IndexMap` was to preserve the ordering of the rules in the original file if the configuration gets re-serialized, this is no longer the case now and the rules will always get printed in the order defined in `rules.rs`. Would it have been possible to use `#[serde(flatten)]` instead to inline the `rule` map into its parent object ?

ematipico:
> Originally the point of using `IndexMap` was to preserve the ordering of the rules in the original file if the configuration gets re-serialized, this is no longer the case now and the rules will always get printed in the order defined in `rules.rs`. Would it have been possible to use `#[serde(flatten)]` instead to inline the `rule` map into its parent object ?

True, but unfortunately I bumped into some hiccups.

I tried different approaches: 
- using an `enum` for `rules`, something like `rules: RuleEnume::Js(IndexMap)` with a combination of `#[serde(untagged)]` but unfortunately it didn't workout as I wanted;
- I evaluted `#[serde(flatten)]`, but unfortunately doesn't work with [deny_unknown_fields](https://serde.rs/container-attrs.html#deny_unknown_fields), so I am not sure if will workout. I could give it a shot.

leops (approved):

PR #3018, opened 2022-08-05
refactor(playground): refactor the playground to use the WASM bindings for the workspace

### LINKED ISSUE #3012: wire the playground and the workspace

leops:
## Summary

Fixes #3012 

This PR refactors the Playground to use the WASM bindings to the Workspace API. These bindings currently live in the playground crate itself along with a few utilities, but I plan on eventually moving this code out into its own crate in a later refactor (as part of the Node API umbrella).

The core of the change is relatively simple, and only requires adding a few methods to the workspace in order to access the internal debug representations rendered by the playground (like the formatter IR or the control flow graph). The WASM binding glue is implemented over the existing serde implementations for all the Workspace-related types, using the `serde-serialize` feature of `wasm-bindgen`. Unfortunately since these types go through serialization they are represented as untyped `any` values in the generated TypeScript definitions for the bindings (#3031 will add auto-generated definitions for these)

## Test Plan

Many of the changes are statically checked or execute at compile time, but unfortunately the playground doesn't really any testing infrastructure in place so that will have to be manually tested in the PR deployment

### DIFF
--- a/crates/rome_service/src/file_handlers/javascript.rs
+++ b/crates/rome_service/src/file_handlers/javascript.rs
@@ -1,22 +1,24 @@
-use rome_analyze::{AnalysisFilter, ControlFlow, Never, RuleCategories, RuleFilter};
+use rome_analyze::{AnalysisFilter, ControlFlow, Never, QueryMatch, RuleCategories, RuleFilter};
 use rome_diagnostics::{Applicability, CodeSuggestion, Diagnostic};
 use rome_formatter::{FormatError, Printed};
 use rome_fs::RomePath;
 use rome_js_analyze::utils::rename::RenameError;
-use rome_js_analyze::{analyze, RuleError};
+use rome_js_analyze::{analyze, analyze_with_inspect_matcher, RuleError};
 use rome_js_formatter::context::QuoteStyle;
 use rome_js_formatter::{context::JsFormatContext, format_node};
 use rome_js_parser::Parse;
 use rome_js_semantic::semantic_model;
-use rome_js_syntax::{JsAnyRoot, JsLanguage, SourceType, TextRange, TextSize, TokenAtOffset};
+use rome_js_syntax::{
+    JsAnyRoot, JsLanguage, JsSyntaxNode, SourceType, TextRange, TextSize, TokenAtOffset,
+};
 use rome_rowan::{AstNode, BatchMutationExt, Direction};
 
-use crate::workspace::{
-    CodeAction, FixAction, FixFileMode, FixFileResult, PullActionsResult, RenameResult,
-};
 use crate::{
     settings::{FormatSettings, Language, LanguageSettings, LanguagesSettings, SettingsHandle},
-    workspace::server::AnyParse,
+    workspace::{
+        server::AnyParse, CodeAction, FixAction, FixFileMode, FixFileResult, GetSyntaxTreeResult,
+        PullActionsResult, RenameResult,
+    },
     RomeError, Rules,
 };
 
@@ -73,7 +75,9 @@ impl ExtensionHandler for JsFileHandler {
     fn capabilities(&self) -> super::Capabilities {
         super::Capabilities {
             parse: Some(parse),
-            debug_print: Some(debug_print),
+            debug_syntax_tree: Some(debug_syntax_tree),
+            debug_control_flow: Some(debug_control_flow),
+            debug_formatter_ir: Some(debug_formatter_ir),
             lint: Some(lint),
             format: Some(format),
             code_actions: Some(code_actions),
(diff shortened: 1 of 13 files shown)


ematipico:
Is there any chance that we could split the schema generation in a separate PR?

leops:
> Is there any chance that we could move the schema generation changes in a separate PR?

Sure, the PR is already split into two different commit so that should be easy enough to do

website/playground/src/lib.rs
```diff
@@ -1,275 +1,360 @@
...
     }
 
-    #[wasm_bindgen(getter)]
-    pub fn errors(&self) -> String {
-        self.errors.clone()
+    pub fn get_syntax_tree(
+        &self,
+        params: IGetSyntaxTreeParams,
+    ) -> Result<IGetSyntaxTreeResult, Error> {
+        let params: GetSyntaxTreeParams = params.into_serde().map_err(into_error)?;
```
ematipico, line 99: Is `into_serde` a blanket trait applied to all data structures that implement `Deserialize`?

website/playground/src/romeWorker.ts
```diff
@@ -1,20 +1,70 @@
...
+import {
+	SourceType,
+	LoadingState,
+	PlaygroundState,
+	IndentStyle,
+	RomeOutput,
+	QuoteStyle,
+} from "./types";
+
+let workspace: Workspace | null = null;
```
ematipico, line 11: `undefined` instead of `null`?

website/playground/src/romeWorker.ts
```diff
@@ -25,31 +75,96 @@ self.addEventListener("message", async (e) => {
...
+						format_with_errors: true,
+						line_width: lineWidth,
+					},
+					linter: {
+						enabled: true,
+					},
+					languages: {
+						javascript: {
+							format: {
+								quote_style: quoteStyle === QuoteStyle.Double ? "Double" : "Single",
```
ematipico, line 93: Should we use the enum instead of the strings? Less prone to typos 

```suggestion
								quote_style: quoteStyle === QuoteStyle.Double ? QuoteStyle.Double : QuoteStyle.Single,
```

website/playground/src/romeWorker.ts
```diff
@@ -25,31 +75,96 @@ self.addEventListener("message", async (e) => {
...
+
+			const printer = new DiagnosticPrinter(path.path, code);
+			for (const diag of diagnostics.diagnostics) {
+				printer.print(diag);
+			}
+
+			const printed = workspace.format_file({
+				path,
+				indent_style:
+					indentStyle === IndentStyle.Tab ? "Tab" : { Space: indentWidth },
```
ematipico, line 152: I believe this code is out of date, I remember we removed `indent_style` from the request

website/playground/src/romeWorker.ts
```diff
@@ -25,31 +75,96 @@ self.addEventListener("message", async (e) => {
...
+			const syntax_tree = workspace.get_syntax_tree({
+				path,
+			});
+			const control_flow_graph = workspace.get_control_flow_graph({
+				path,
+				cursor: cursorPosition,
+			});
+			const formatter_ir = workspace.get_formatter_ir({
+				path,
+				indent_style:
```
ematipico, line 135: I believe this code is out of date, I remember we removed `indent_style` from the request

crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -73,7 +75,9 @@ impl ExtensionHandler for JsFileHandler {
     fn capabilities(&self) -> super::Capabilities {
         super::Capabilities {
             parse: Some(parse),
-            debug_print: Some(debug_print),
+            debug_syntax_tree: Some(debug_syntax_tree),
```
MichaReiser, line 78: Nit: How about grouping these insides of a `DebugCapabilities` struct instead?

crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -124,9 +128,67 @@ where
     }
 }
 
-fn debug_print(_rome_path: &RomePath, parse: AnyParse) -> String {
+fn debug_syntax_tree(_rome_path: &RomePath, parse: AnyParse) -> GetSyntaxTreeResult {
```
MichaReiser, line 131: It may be desirable to put the debug features behind a feature flag so that these can be excluded in builds where it is unnecessary to shove off some bytes. 

The other thing I'm thinking is if we should move all `debug` functionality to a `debug()` handler and implement the methods on there.

crates/rome_service/src/workspace/server.rs
```diff
@@ -179,6 +183,37 @@ impl Workspace for WorkspaceServer {
...
+
+        let parse = self.get_parse(params.path.clone())?;
+        let printed = printer(&params.path, parse, params.cursor);
+
+        Ok(printed)
+    }
+
+    fn get_formatter_ir(&self, params: GetFormatterIRParams) -> Result<String, RomeError> {
+        let capabilities = self.features.get_capabilities(&params.path);
+        let printer = capabilities
```
MichaReiser, line 203: Nit: rename to `format`?

crates/rome_service/src/workspace/server.rs
```diff
@@ -179,6 +183,37 @@ impl Workspace for WorkspaceServer {
...
+            .debug_formatter_ir
+            .ok_or_else(|| RomeError::SourceFileNotSupported(params.path.clone()))?;
+
+        let parse = self.get_parse(params.path.clone())?;
```
MichaReiser, line 207: This used to be more optimized in the current playground as it was able to create the formatted IR once, print the IR, and then print it to a string without having to `parse` and `format` the same node twice.

website/playground/src/romeWorker.ts
MichaReiser: That's very much personal taste. I prefer `null` so that you at least know that the variable has been initialized.

website/playground/src/romeWorker.ts
```diff
@@ -25,31 +75,96 @@ self.addEventListener("message", async (e) => {
...
+					indentStyle === IndentStyle.Tab ? "Tab" : { Space: indentWidth },
+			});
+
+			const diagnostics = workspace.pull_diagnostics({
```
MichaReiser, line 139: I do like the workspace but having this stateful workspace API clearly shows that it may be overkill in a use case like the playground where the caching has no added value but the statefulness adds a lot of complexity and introduce risks for a wide range of bugs. CC: @ematipico for one potential use case.

website/playground/src/lib.rs
leops: Almost yes, it's actually just a method on `JsValue` defined as `fn into_serde<T: Deserialize>(&self) -> Result<T>`, and it's possible to call it on the `extern` type aliases because these all get resolved to `JsValue` by `wasm-bindgen`

website/playground/src/romeWorker.ts
leops: Unfortunately because TypeScript doesn't resolve the value of enums during compilation the `QuoteStyle.Double` expression and `import { QuoteStyle } from 'rome_playground'` statement would remain in the emitted code. This means the bundler will expect the `rome_playground` module have an export named `QuoteStyle`, but the JS code generated by `wasm-bindgen` doesn't include this

crates/rome_service/src/workspace/server.rs
leops: With the current implementation of the workspace, the result of `get_parse` is already cached between calls so the file is only getting parsed once, and it would be trivial to extend this caching mechanism to the Formatter IR to allow it to be reused between calls to `format` and `debug_formatter_ir`

website/playground/src/romeWorker.ts
leops: It's true that at the moment this change makes no practical difference for the playground (it might even regress performance a bit since there are more JS to WASM calls), but eventually I think the playground is an editor experience like any other and it would benefit from performance improvements permitted by the stateful language server implementation like reusing the green node cache between parsing session

crates/rome_service/src/file_handlers/javascript.rs
leops: Unfortunately we don't really have any binary distribution where this feature would be disabled, the only binary that doesn't use these capabilities is the CLI and it may end up getting merged with the LSP in a future PR

website/playground/src/romeWorker.ts
leops: Yes that's an habit from programming in Rust, I like my variables to be explicitly initialized

ematipico (approved):

PR #3019, opened 2022-08-05
feat(rome_js_semantic): semantic model "is_exported" query

xunilrj:
## Summary

This PR implements the "is_exported" query in the SemanticModel.
It will be first used by the ```noUnusedVariables``` lint rule (https://github.com/rome/tools/issues/2979) to avoid false flags like 

```js
error[[js/noUnusedVariables](https://rome.tools/docs/lint/rules/noUnusedVariables/)]: This function is unused.
  ┌─ main.js:1:17
  │
1 │ export function f() {
```

## Test Plan

```
>  cargo test -p rome_js_semantic -- ok_semantic_model_is_exported
```


ematipico (commented):
Please don't hate the messenger, but we should also add the `module.exports` syntax.

xunilrj:
> Please don't hate the messenger, but we should also add the `module.exports` syntax.

That is true. I will do it in another PR.

PR #3021, opened 2022-08-06
feat(rome_formatter): `align > indent` nesting with `IndentStyle::Space`

MichaReiser:
## Summary

This changes the behaviour of the printer if an `indent` is nested inside an `align`.
The existing behaviour was to convert the `align` to an `indent`.

This PR changes the behaviour to keep the spaces as is, so that an `align(2) > indent` results in a `______` sequence with `IndentStyle::Space(4)`

This PR further improves `Memoized` so that it avoids creating an `Interned` element on the heap if it has no content.
## Test Plan

I expanded the doc test on `align` to show how the IR behaves differently depending on the `IndentStyle`

There are two snapshots that changed because of a fix in `fits` in combination with the indent width ([see this comment](https://github.com/rome/tools/pull/3021#discussion_r939568614))

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -747,6 +747,9 @@ impl<Context> std::fmt::Debug for Dedent<'_, Context> {
 /// Using [indent] is preferred in all other situations as it respects the users preferred indent character.
 ///
 /// # Examples
+///
+/// ## Tab indention
+///
 /// ```
 /// use std::num::NonZeroU8;
 /// use rome_formatter::{format, format_args};
@@ -779,9 +782,52 @@ impl<Context> std::fmt::Debug for Dedent<'_, Context> {
 /// );
 /// ```
 ///
-/// You can see that the printer indents the function's `}` by two spaces as specified by the IR and not with a tab.
-/// You can further see that alignments behave the same as an `indent` if you nest `indent` inside an align.
+/// You can see that:
 ///
+/// * the printer indents the function's `}` by two spaces as specified by the IR and not with a tab.
+/// * that `align` acts the same as an `indent` if you nest `indent` inside an align.
+///
+/// ## Spaces indention
+///
+/// ```
+/// use std::num::NonZeroU8;
+/// use rome_formatter::{format, format_args, IndentStyle};
+/// use rome_formatter::prelude::*;
+///
+/// let context = SimpleFormatContext {
+///     indent_style: IndentStyle::Space(4),
+///     ..SimpleFormatContext::default()
+/// };
+///
+/// let block = format!(context, [
+///     text("a"),
+///     hard_line_break(),
+///     text("?"),
+///     space(),
+///     align(2, &format_args![
+///         text("function () {"),
+///         hard_line_break(),
+///         text("}"),
+///     ]),
+///     hard_line_break(),
+///     text(":"),
+///     space(),
+///     align(2, &format_args![
+///         text("function () {"),
+///         block_indent(&text("console.log('test');")),
+///         text("}"),
+///     ]),
+///     text(";")
+/// ]).unwrap();
+///
+/// assert_eq!(
+///     "a\n? function () {\n  }\n: function () {\n      console.log('test');\n  };",
+///     block.print().as_code()
+/// );
+/// ```
+///
+/// Nesting an `indent` inside of an `align` has a different semantic when using spaces over tabs as indent sequence.
+/// With spaces, the outer `align` doesn't get converted into an `indent` and is, instead, kept as is.
 pub fn align<Content, Context>(count: u8, content: &Content) -> Align<Context>
 where
     Content: Format<Context>,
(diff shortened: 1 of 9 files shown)


MichaReiser:
!bench_formatter

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -918,7 +942,8 @@ fn fits_element_on_line<'a, 'rest>(
         ),
 
         FormatElement::Text(token) => {
-            state.line_width += state.pending_indent.level() as usize * options.indent_string.len()
```
MichaReiser, line 921: using `indent_string` here resulted in incorrect results if the printer used tabs because the tab then only took the space of one character instead of what's configured in `tab_width`. 

This is why the two snapshots in the `specs` folder changed.

MichaReiser:
!bench_formatter

crates/rome_formatter/src/builders.rs
```diff
@@ -779,9 +782,52 @@ impl<Context> std::fmt::Debug for Dedent<'_, Context> {
 /// );
 /// ```
 ///
-/// You can see that the printer indents the function's `}` by two spaces as specified by the IR and not with a tab.
-/// You can further see that alignments behave the same as an `indent` if you nest `indent` inside an align.
+/// You can see that:
+///
+/// * the printer indents the function's `}` by two spaces as specified by the IR and not with a tab.
+/// * that `align` acts the same as an `indent` if you nest `indent` inside an align.
```
ematipico, line 788: I find this phrase difficult to understand. Should we specify what "acts the same as" means here? Because I don't get it. Also, I don't understand the nesting part. It seems that I was supposed to know something about it but the documentation doesn't mention it 

```suggestion
/// * that `align` acts the same as an `indent` if you nest `indent` inside an `align`.
```

crates/rome_formatter/src/builders.rs
```diff
@@ -779,9 +782,52 @@ impl<Context> std::fmt::Debug for Dedent<'_, Context> {
...
+///     text(";")
+/// ]).unwrap();
+///
+/// assert_eq!(
+///     "a\n? function () {\n  }\n: function () {\n      console.log('test');\n  };",
+///     block.print().as_code()
+/// );
+/// ```
 ///
+/// Nesting an `indent` inside of an `align` has a different semantic when using spaces over tabs as indent sequence.
```
ematipico, line 829: What's this difference in semantic? I mean, we explain what happens to `align` but we don't explain what happens to `indent`.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -80,15 +80,25 @@ impl<'a> Printer<'a> {
...
             FormatElement::Text(token) => {
                 if !self.state.pending_indent.is_empty() {
-                    self.print_str(
-                        self.options
-                            .indent_string
-                            .repeat(self.state.pending_indent.level() as usize)
-                            .as_str(),
-                    );
+                    let (indent_char, repeat_count) = match self.options.indent_style() {
+                        IndentStyle::Tab => ('\t', 1),
```
ematipico, line 84: I am not sure this is correct. Should it be `2`? https://github.com/rome/tools/blob/main/crates/rome_formatter/src/printer/printer_options/mod.rs#L33

And if `1` is the correct amount, why?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -622,12 +639,22 @@ impl Indention {
 
     /// Increments the level by one.
     ///
-    /// It sets `align` to 0 if the current value is [Indent::Align] and increments the level by one.
-    fn increment_level(self) -> Self {
+    /// The behaviour if this is an [Indent::Align] depends on the indent character used by the printer:
```
ematipico, line 642:
```suggestion
    /// The behaviour if this is an [Indent::Align] depends on the [indentation style](IndentStyle) used by the printer:
```

ematipico (commented):
Could you add in the PR how the new changes affected the new snapshots?

crates/rome_formatter/src/printer/mod.rs
MichaReiser: 1 is correct because we only want to print 1 tab character. The 2 in the options defines the 'width' each single tab should account for.

MichaReiser:
> Could you add in the PR how the new changes affected the new snapshots?

Hmm, I added a comment [here](https://github.com/rome/tools/pull/3021#discussion_r939568614) but it no longer showed up because I updated the code since then. I'll add a link in the summary.

crates/rome_formatter/src/builders.rs
MichaReiser: I tried to rephrase the explanation. I hope that helps

crates/rome_formatter/src/builders.rs
MichaReiser: The `indent` keeps the same semantic. I rephrased the sentence to, hopefully, make that clearer.

ematipico (approved):

PR #3022, opened 2022-08-06
fix(rome_js_parser):  Export anonymous function in .d.ts 

IWANABETHATGUY:
## Summary
1. Part of https://github.com/rome/tools/issues/2310 
2. Fixing `Export anonymous function in .d.ts` syntax error.
3. Introduce a new file type `.d.ts`, and add codegen logic.

## Test Plan
1. See the new added snapshot.


IWANABETHATGUY:
!bench_parser

IWANABETHATGUY:
!bench

IWANABETHATGUY:
!bench_parser

crates/rome_js_parser/src/syntax/function.rs
```diff
@@ -355,7 +372,16 @@ pub(crate) fn parse_ambient_function(p: &mut Parser, m: Marker) -> CompletedMark
     }
 
     p.expect(T![function]);
-    parse_binding(p).or_add_diagnostic(p, expected_binding);
+    match parse_binding(p) {
+        binding @ Absent => {
+            if kind == AmbientFunctionKind::Export {
+                binding.ok();
+            } else {
+                binding.or_add_diagnostic(p, expected_binding);
```
MichaReiser, line 380: Does that mean that the binding is required in none `export` context and is otherwise optional. 

Our AST strictly encodes optionality. E.g. see `JsFunctionExportDefaultDeclaration` for a function where the `id` is optional opposed to `JsFunctionDeclaration` where it is not. 

We may need to apply the same approach here but I must first better understand what this PR is solving/ changing.

MichaReiser (changes requested):
I find it difficult to understand what this PR is changing. 

Please add a more descriptive summary explaining which syntax was invalid before and is valid now.

crates/rome_js_parser/src/syntax/function.rs
IWANABETHATGUY: > Does that mean that the binding is required in none `export` context and is otherwise optional.
> 
> Our AST strictly encodes optionality. E.g. see `JsFunctionExportDefaultDeclaration` for a function where the `id` is optional opposed to `JsFunctionDeclaration` where it is not.
> 
> We may need to apply the same approach here but I must first better understand what this PR is solving/ changing.

Sorry for the late replay, 
1. For the first question yes.
2. I like your idea.

PR #3023, opened 2022-08-07
fix(rome_js_parser): Syntax error when using optional binding on .d.ts

IWANABETHATGUY:
## Summary
1. Part of #2310
2. Fix syntax error when using Optional binding on `.d.ts` or `.ts`

## Test Plan
1. Checkout the new snapshot

### DIFF
--- a/crates/rome_js_parser/src/syntax/function.rs
+++ b/crates/rome_js_parser/src/syntax/function.rs
@@ -918,11 +918,6 @@ impl ParameterContext {
         self == &ParameterContext::Setter
     }
 
-    // Comment for now, maybe it would be used in the future
-    // pub fn is_implementation(&self) -> bool {
-    //     self == &ParameterContext::Implementation
-    // }
-
     pub fn is_parameter_property(&self) -> bool {
         self == &ParameterContext::ParameterProperty
     }
--- a/crates/rome_js_parser/src/syntax/function.rs
+++ b/crates/rome_js_parser/src/syntax/function.rs
@@ -918,9 +918,10 @@ impl ParameterContext {
         self == &ParameterContext::Setter
     }
 
-    pub fn is_implementation(&self) -> bool {
-        self == &ParameterContext::Implementation
-    }
+    // Comment for now, maybe it would be used in the future
+    // pub fn is_implementation(&self) -> bool {
+    //     self == &ParameterContext::Implementation
+    // }
 
     pub fn is_parameter_property(&self) -> bool {
         self == &ParameterContext::ParameterProperty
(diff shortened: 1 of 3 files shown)


IWANABETHATGUY:
An optional binding pattern parameter  in an implementation signature should be valid in both `d.ts` and `ts`
## TS
https://www.typescriptlang.org/play?#code/C4TwDgpgBAogTnA9nAzgFQgD2AeTMAS0QDsUoBeKAbwCgooUIwBDOZ4ZALimeJAG46UACbtmANVbdeAmgF8awiAGMANq2jFmAWwgoWy6AEFqQrGGTAoa5ijJGAVgDdT9ehATJ0WYAAoPSKgA-NwowHAExADmANoAulAAPlDEAK6qqklQqcRKAGaREMIANNQMTKzsyKWiwBKsUHIhsJ6oGNh4hCQoAJSh4ZFRgvQKckA

## D.TS
https://www.typescriptlang.org/play?filetype=d.ts#code/C4TwDgpgBAogTnA9nAzgFQgD2AeTMAS0QDsUoBeKAbwCgooUIwBDOZ4ZALimeJAG46UACbtmANVbdeAmgF8awiAGMANq2jFmAWwgoWy6AEFqQrGGTAoa5ijJGAVgDdT9ehATJ0WYAAoPSKgA-NwowHAExADmANoAulAAPlDEAK6qqklQqcRKAGaREMIANNQMTKzsyKWiwBKsUHIhsJ6oGNh4hCQoAJSh4ZFRgvQKckA

IWANABETHATGUY:
!bech_parser

crates/rome_js_parser/src/syntax/function.rs
```diff
@@ -978,29 +979,24 @@ pub(crate) fn parse_formal_parameter(
...
-                p.error(
-                    p.err_builder(
-                        "A parameter property may not be declared using a binding pattern.",
-                    )
+            valid = false;
+            p.error(
+                p.err_builder("A parameter property may not be declared using a binding pattern.")
                     .primary(binding_range, ""),
-                );
-            } else if parameter_context.is_implementation() && is_optional {
```
MichaReiser, line 995: This error is to catch

```
class Test {
  errorsText({ separator: dataVar }?: string[]): string {
    return "a";
  }
}
```

which isn't valid.

IWANABETHATGUY:
!bench_parser

crates/rome_js_parser/src/syntax/function.rs
IWANABETHATGUY: https://github.com/microsoft/TypeScript/blob/5f64ae878e892ab293e54e779f470fd0ccc0db2c/src/compiler/checker.ts#L35299-L35301, This is the related code in Typescript, 
```rs
 if ((node.questionToken || isJSDocOptionalParameter(node)) && isBindingPattern(node.name) && (func as FunctionLikeDeclaration).body) {
                error(node, Diagnostics.A_binding_pattern_parameter_cannot_be_optional_in_an_implementation_signature);
            }
```
We can't check if `function.body` exists or not in the parser phase, so I remove the checking logic. Maybe we could put it in the analyzer?

crates/rome_js_parser/src/syntax/function.rs
```diff
@@ -978,29 +979,24 @@ pub(crate) fn parse_formal_parameter(
             false
         };
 
+        // test ts ts_parameter_option_binding_pattern
```
MichaReiser, line 982: Would you mind deleting all commented out code. We can re-add it if necessary

MichaReiser (approved):

crates/rome_js_parser/src/syntax/function.rs
IWANABETHATGUY: Do you mean inline testing or https://github.com/rome/tools/pull/3023/files#diff-875082649067c72f4a3af6874edd9b860b6db8648fdcf80bad930dea0bf96f85R921-R924

MichaReiser: whoops, I commented on the wrong line. I meant removing the `// Comment for now, maybe it would be used in the future`

IWANABETHATGUY: Done

PR #3024, opened 2022-08-07
feat(rome_js_formatter): Conditional expression formatting

MichaReiser:
## Summary

This PR implements conditional expression formatting similar to Prettier with the exception of the special handling of conditional expressions containing JSX (PR is already huge). 

## Test Plan

I manually reviewed the snapshot changes. Some changes may appear as regressions but are mainly due to other syntax formatting (binary expression, call arguments). I ignored comments in this PR (except to make sure comments are stable) as we plan to refactor comments. 

File Based Average Prettier Similarity: 78.85% -> 79.72%
Line Based Average Prettier Similarity: 74.25% -> 76.49%


crates/rome_formatter/src/arguments.rs
```diff
@@ -49,7 +50,7 @@ impl<'fmt, Context> Argument<'fmt, Context> {
     }
 
     /// Formats the value stored by this argument using the given formatter.
-    #[inline]
+    #[inline(always)]
```
MichaReiser, line 53: I ran into issues with stack overflows on the playground in debug builds. Adding `inline(always)` ensures that this "trivial" calls are also inlined in debug mode which results in nicer stack traces

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -333,7 +333,7 @@ impl<'a> Printer<'a> {
         // If the indentation level has changed since these line suffixes were queued,
         // insert a line break before to push the comments into the new indent block
         // SAFETY: Indexing into line_suffixes is guarded by the above call to is_empty
-        let has_line_break = self.state.line_suffixes[0].args.indent < args.indent;
+        let has_line_break = self.state.line_suffixes[0].args.indent.level() < args.indent.level();
```
MichaReiser, line 335: Comparing the `align` isn't necessary as it turns out, this hack is only for moving comments after a `(`, `[`, or `{` into the indent.

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
```diff
@@ -174,9 +183,8 @@ fn parenthesis_can_be_omitted(node: &JsParenthesizedExpression) -> SyntaxResult<
     }
 
     let parent_precedence = FormatPrecedence::with_precedence_for_parenthesis(parent.as_ref());
-    let node_precedence = FormatPrecedence::with_precedence_for_parenthesis(Some(node.syntax()));
```
MichaReiser, line 177: `node.syntax().kind()` is always a `JS_PARENTHESIZED_EXPRESSION` that resolves to `FormatPrecedence::Low`. There's no need to call this function.

crates/rome_js_formatter/tests/specs/prettier/js/sequence-break/break.js.snap
```diff
@@ -77,7 +60,35 @@ for (aLongIdentifierName = 0, aLongIdentifierName = 0, aLongIdentifierName = 0,
 -    : function () {
 -        return 0;
 -      });
-+    }
++      }
++),
++  (
```
MichaReiser, line 65: This is an issue with parenthesized expressions

crates/rome_js_formatter/tests/specs/prettier/js/ternaries/nested.js.snap
```diff
@@ -148,228 +130,165 @@ a
 -      </Content>
 -    </Message>
 -  );
-+const StorybookLoader = ({ match }) => (
 +  match.params.storyId === "button"
```
MichaReiser, line 133: JSX special case is intentionally excluded from this PR

crates/rome_js_formatter/tests/specs/prettier/typescript/as/ternary.ts.snap
```diff
@@ -53,51 +53,18 @@ bifornCringerMoshedPerplexSawder =
...
-   ) as Fooooooooooo;
+@@ -29,22 +29,23 @@
  }
  
  function foo() {
 -  void ((
 -    coooooooooooooooooooooooooooooooooooooooooooooooooooond
--      ? baaaaaaaaaaaaaaaaaaaaar
++  void (
++    (coooooooooooooooooooooooooooooooooooooooooooooooooooond
```
MichaReiser, line 63: I only implemented the parentheses handling for expressions but not types. It will be easier to address this with an overall solution that avoids "resolving" expressions everywhere.

MichaReiser:
!bench_formatter

MichaReiser:
> 
> ```
> formatter/checker.ts                     1.00    375.1±3.34ms     6.9 MB/sec    1.00    375.3±3.07ms     6.9 MB/sec
> formatter/compiler.js                    1.00    229.0±1.20ms     4.6 MB/sec    1.01    230.7±1.00ms     4.5 MB/sec
> formatter/d3.min.js                      1.00    181.0±1.67ms  1482.9 KB/sec    1.02    185.4±1.54ms  1448.0 KB/sec
> formatter/dojo.js                        1.00     12.9±0.06ms     5.3 MB/sec    1.01     12.9±0.05ms     5.3 MB/sec
> formatter/ios.d.ts                       1.00    259.6±1.42ms     7.2 MB/sec    1.01    261.3±0.95ms     7.1 MB/sec
> formatter/jquery.min.js                  1.00     49.3±0.53ms  1715.7 KB/sec    1.03     51.0±0.55ms  1660.0 KB/sec
> formatter/math.js                        1.00    378.9±2.10ms  1749.9 KB/sec    1.04    393.1±3.05ms  1686.8 KB/sec
> formatter/parser.ts                      1.00      8.7±0.03ms     5.6 MB/sec    1.01      8.7±0.01ms     5.5 MB/sec
> formatter/pixi.min.js                    1.00    204.8±2.03ms     2.1 MB/sec    1.02    209.1±2.51ms     2.1 MB/sec
> formatter/react-dom.production.min.js    1.00     61.2±0.87ms  1925.2 KB/sec    1.01     62.0±0.48ms  1902.4 KB/sec
> formatter/react.production.min.js        1.00      3.2±0.02ms  1991.0 KB/sec    1.01      3.2±0.01ms  1979.9 KB/sec
> formatter/router.ts                      1.00      6.7±0.05ms     9.1 MB/sec    1.00      6.8±0.01ms     9.1 MB/sec
> formatter/tex-chtml-full.js              1.00    486.5±2.11ms  1918.2 KB/sec    1.03    502.6±3.70ms  1856.5 KB/sec
> formatter/three.min.js                   1.00    237.9±2.03ms     2.5 MB/sec    1.01    240.6±1.26ms     2.4 MB/sec
> formatter/typescript.js                  1.00  1537.7±16.05ms     6.2 MB/sec    1.00  1544.5±10.96ms     6.2 MB/sec
> formatter/vue.global.prod.js             1.00     80.2±1.17ms  1538.1 KB/sec    1.04     83.1±1.36ms  1484.5 KB/sec
> ```

I'm surprised this isn't worse. I guess, there aren't so many conditional expressions in our benchmark files

crates/rome_js_formatter/src/jsx/auxiliary/space.rs
```diff
@@ -1,39 +0,0 @@
-use crate::context::QuoteStyle;
```
MichaReiser, line 1: I had to move this is a manually created file and re-generating the formatter deletes the file. It's now part of `utils/jsx`

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -0,0 +1,669 @@
...
+            return false;
+        }
+
+        match parent {
+            None => false,
+            Some(parent) => {
+                let argument = match parent.kind() {
+                    JsSyntaxKind::JS_INITIALIZER_CLAUSE => {
+                        let initializer = JsInitializerClause::unwrap_cast(parent);
+                        initializer.expression().ok().map(JsAnyExpression::from)
```
ematipico, line 530: I am not sure sure we should "mute" a `SyntaxResult`. Is there a specific reason for doing so?

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -0,0 +1,669 @@
...
+                    }
+                    _ => None,
+                };
+
+                argument.map_or(false, |argument| resolve_expression(argument) == expression)
+            }
+        }
+    }
+
+    /// Returns `true` if this is the root conditional expression and the parent is a static member expression.
```
ematipico, line 563:
```suggestion
    /// Returns `true` if this is the root conditional expression and the parent is a [static member expression](rome_js_syntax::JsStaticMemberExpression).
```

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -0,0 +1,669 @@
...
+                .map(resolve_expression_syntax)
+                .as_ref()
+                == Ok(conditional.syntax())
+        }
+        JsSyntaxKind::JS_EXTENDS_CLAUSE => true,
+        JsSyntaxKind::JS_ARROW_FUNCTION_EXPRESSION => {
+            let arrow = JsArrowFunctionExpression::unwrap_cast(parent.clone());
+
+            match arrow.body() {
+                Err(_) => false,
```
ematipico, line 633: I am not sure we should mute an error. I think we should bounce it to the parent

crates/rome_js_formatter/src/utils/parens.rs
```diff
@@ -0,0 +1,65 @@
+use rome_js_syntax::{JsAnyExpression, JsAnyInProperty};
+use rome_rowan::SyntaxResult;
+
+/// Returns `true` if the `node`s first token isn't a lookahead token (a `{`).
```
ematipico, line 4: I can't see `node` in the signature of the function. Only `expression`. Should be specific where `node` is coming from?

And.. what's a lookahead token???

crates/rome_js_formatter/src/utils/parens.rs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+    };
+
+    Ok(result)
+}
+
+/// Resolves the left hand side of a binary like expression.
+///
+/// Returns the `left` most node or the passed in expression
```
ematipico, line 41: The signature of the function wants a `JsAnyExpression` but here we mention a binary like expression. Should we use `JsAnyBinaryLikeExpression` instead? That node already contains all the variants you need.

Or maybe we should create a function inside the `JsAnyBinaryLikeExpression` struct?

crates/rome_js_formatter/src/js/expressions/arrow_function_expression.rs
```diff
@@ -80,22 +80,50 @@ impl FormatNodeRule<JsArrowFunctionExpression> for FormatJsArrowFunctionExpressi
         // going to get broken anyways.
         // TODO: Explore the concept of hierarchical line breaking
         //   or vertical stacking for arrow selfs
```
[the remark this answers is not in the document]
MichaReiser: Why so? I didn't address this.

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -0,0 +1,669 @@
...
+
+        // This tries to find the start of a member chain by iterating over all ancestors of the conditional
+        // expression, while skipping parenthesized expression.
+        // The iteration "breaks" as soon as a non-member-chain node is found.
+        for ancestor in ancestors {
+            let ancestor = match ancestor.kind() {
+                JsSyntaxKind::JS_CALL_EXPRESSION => {
+                    let call_expression = JsCallExpression::unwrap_cast(ancestor);
+
+                    if call_expression.callee().map(resolve_expression).as_ref() == Ok(&expression)
```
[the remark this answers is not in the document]
MichaReiser: We're testing if the `callee` is the same node as `expression`. This will always be false if it is `Err` (can never be the same as the node). But this isn't muting the `Err`, the `Err` has probably already been handled when formatting the parent node.

crates/rome_js_formatter/src/utils/conditional.rs
MichaReiser: It's safe to do so here because the function will return the "safe" default of `false` if the returned value from here is `Err`. 

The errors get handled as part of the parent node.

crates/rome_js_formatter/src/utils/conditional.rs
MichaReiser: We're testing here if the body is of a specific node. If that node is missing, then there's nothing to compare against and we are good ;)

crates/rome_js_formatter/src/utils/parens.rs
MichaReiser: I thought about that but the idea here is that you can call it with any expression and it either returns the left most from a binary like or the expression itself. 

It could be an option to use `JsAnyBinaryLike` internally but it involves a few more casts.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -689,46 +688,6 @@ impl Indention {
     }
 }
 
-/// Compares two [Indent]s by first comparing the indent level and, if equal, comparing the `align`.
-/// This implementation assumes that `width(align) < width(1-level)`.
-impl PartialOrd for Indention {
```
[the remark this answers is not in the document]
MichaReiser: It's no longer needed, comparing the indent level is sufficient.

crates/rome_js_formatter/src/js/expressions/arrow_function_expression.rs
```diff
@@ -80,22 +80,50 @@ impl FormatNodeRule<JsArrowFunctionExpression> for FormatJsArrowFunctionExpressi
...
 
         if body_has_soft_line_break {
             write![f, [body.format()]]
         } else {
-            write!(f, [group(&soft_line_indent_or_space(&body.format()))])
+            write!(
+                f,
+                [group(&soft_line_indent_or_space(&format_with(|f| {
+                    if should_add_parens {
+                        write!(f, [if_group_fits_on_line(&text("("))])?;
```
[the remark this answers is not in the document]
MichaReiser: I don't think so. `fits in`... you can stuff it into something. `fits on` You can put it onto something?

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -0,0 +1,670 @@
...
+    fn fmt(&self, f: &mut Formatter<JsFormatContext>) -> FormatResult<()> {
+        match self {
+            ExpressionOrType::JsAnyExpression(expression) => expression.format().fmt(f),
+            ExpressionOrType::TsType(ty) => ty.format().fmt(f),
+        }
+    }
+}
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+enum ConditionalLayout {
```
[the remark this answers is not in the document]
MichaReiser: I can add some limited examples but the issue here is that there are multiple other factors flowing into how the node gets ultimately formatted.

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -0,0 +1,670 @@
...
+    /// Conditional expression that is the `alternate` of another conditional expression
+    NestedAlternate { parent: JsAnyConditional },
+
+    /// Conditional expression that is the `test` of another conditional expression.
+    NestedTest { parent: JsAnyConditional },
+
+    /// Conditional expression that is the consequent of another conditional expression.
+    NestedConsequent { parent: JsAnyConditional },
```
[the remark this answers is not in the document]
MichaReiser: I can add the examples but the documentation says exactly when this layout is chosen. It may miss the part of what the consequence is of choosing this layout.

crates/rome_js_formatter/src/js/expressions/parenthesized_expression.rs
MichaReiser: This will all go away in my next PR

crates/rome_js_formatter/src/utils/conditional.rs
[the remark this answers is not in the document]
ematipico: My main concern was just if we are too technical, this could cause more time to understand the documentation. Just that.

crates/rome_js_formatter/src/utils/conditional.rs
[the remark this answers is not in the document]
ematipico: I am really concerned that this will shoot us in the back on the long run:
- it can set a precedent, and future implementations might copy this pattern;
- if the implementation of the algorithm of the a parent changes, we could lose an error;
- in a format range, depending on what's the range selected, we might not have a parent;

If you feel strongly about this pattern, then could you please add `// SAFETY` comments explaining why it's safe to mute the error?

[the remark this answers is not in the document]
MichaReiser: I rather don't want to clutter all the code here with `SAFETY` comments as this isn't doing anything unsafe. This pattern will become more widely used when working on parentheses and I really don't want to add these comments for all checks there (that would result in like 500 SAFETY comments). 

It's fine to copy this pattern. This code isn't part of the formatting. It's only testing if a certain condition holds without itself requiring a `SyntaxResult` return type because it knows what the *safe* default value is. 

The implementation works correctly in all the cases you mentioned because `call_expression.callee() == Err` then means that the node is missing. What this line is testing is if `call_expression.callee() == expression`, which can only be true if the node is  not missing.

crates/rome_js_formatter/tests/specs/prettier/js/unary-expression/urnary_expression.js.snap
```diff
@@ -1,5 +1,8 @@
 ---
 source: crates/rome_js_formatter/tests/prettier_tests.rs
+assertion_line: 271
+info:
+  test_file: js/unary-expression/urnary_expression.js
```
ematipico, line 5: I believe these changes can be reverted

crates/rome_rowan/src/ast/mod.rs
```diff
@@ -124,6 +124,32 @@ pub trait AstNode: Clone {
...
+    /// * [Ok] if the passed node can be cast into this [AstNode]
+    /// * [Err] if the node is of another kind
+    fn try_cast_node<T: AstNode<Language = Self::Language>>(node: T) -> Result<Self, T> {
+        if Self::can_cast(node.syntax().kind()) {
+            Ok(Self::unwrap_cast(node.into_syntax()))
+        } else {
+            Err(node)
+        }
+    }
+
```
ematipico, line 152: nit: would it be possible to add an example for each new API?

ematipico (approved):

PR #3026, opened 2022-08-08
refactor(rome_rowan): Remove `AstNode` constraint from `BatchMutation`

MichaReiser:
## Summary

This PR removes the `N: AstNode<Language=L>` constraint from the `BatchMutation` and changes the `root` to a `SyntaxNode<L>`.

The main motivation for this change is that I want to use the `BatchMutation` for removing all `ParenthesizedExpression`s in the formatter but this isn't possible because:

* The formatter accepts any Js node as the root and there's no `JsAnyNode`
* It isn't guaranteed that the `root` type doesn't change in case the `root` node is a `JsParenthesizedExpression` that gets removed.

## Tests

`cargo test && rustcs`

### DIFF
--- a/crates/rome_js_analyze/src/lib.rs
+++ b/crates/rome_js_analyze/src/lib.rs
@@ -138,7 +138,7 @@ mod tests {
 /// Series of errors encountered when running rules on a file
 #[derive(Debug, PartialEq, Copy, Clone)]
 pub enum RuleError {
-    /// The rule with the specified name replaced the root of the file with a node that is not a [JsAnyRoot].
+    /// The rule with the specified name replaced the root of the file with a node that is not a valid root for that language.
     ReplacedRootWithNonRootError { rule_name: &'static str },
 }
 
--- a/crates/rome_js_analyze/src/lib.rs
+++ b/crates/rome_js_analyze/src/lib.rs
@@ -8,6 +8,7 @@ use rome_js_syntax::{
     suppression::{parse_suppression_comment, SuppressionCategory},
     JsLanguage,
 };
+use std::error::Error;
 
 mod analyzers;
 mod assists;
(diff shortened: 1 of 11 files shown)


crates/rome_service/src/file_handlers/javascript.rs
```diff
@@ -262,7 +262,7 @@ fn fix_all(params: FixAllParams) -> Result<FixFileResult, RomeError> {
         match action {
             Some(action) => {
                 if let Some((range, _)) = action.mutation.as_text_edits() {
-                    tree = action.mutation.commit();
+                    tree = JsAnyRoot::unwrap_cast(action.mutation.commit());
```
ematipico, line 265: `// SAFETY` comment? It's still an unwrap.

ematipico (approved):

crates/rome_service/src/file_handlers/javascript.rs
MichaReiser: Not sure if we can add a safety comment here as it isn't really safe and was simply hidden away in the old API. It's just assuming that no one is changing the root but there's no guarantee that this is indeed the case. @leops what would you propose?

xunilrj: One of the reasons for the ```N``` generic parameter was actually to know which type to return here.

MichaReiser: It always used to be a `JsAnyRoot` and we can always add a type param to `fix_all`.

leops: I think just to be safe we should use `JsAnyRoot::cast` and exit with an error if the mutation replaced the root node of the tree with an invalid node. It's still an error, lint rules aren't supposed to emit code actions that do this kind of things and the `replace_node` API tries to guard against it by requiring the two AST nodes to have the same type (although `replace_element` exists as an "escape hatch"), but at least it won't crash the language server process if it ever happens

MichaReiser: I added a `RuleError` that has a specific error and points to the misbehaving rule.

crates/rome_js_analyze/src/lib.rs
```diff
@@ -133,3 +134,25 @@ mod tests {
         );
     }
 }
+
+/// Series of errors encountered when running rules on a file
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum RuleError {
+    /// The rule with the specified name replaced the root of the file with a node that is not a [JsAnyRoot].
```
ematipico, line 141: nit: in the future we won't have only `JsAnyRoot`

PR #3030, opened 2022-08-08
refactor(rome_rowan): Remove `Clone` requirement from trivia mutation methods

MichaReiser:
Remove the requirement for `Clone` on the trivia mutation methods by changing the
implementation so that it only iterates once over the new trivia rather than twice.

### DIFF
--- a/crates/rome_rowan/src/syntax/token.rs
+++ b/crates/rome_rowan/src/syntax/token.rs
@@ -191,26 +191,27 @@ impl<L: Language> SyntaxToken<L> {
     #[must_use = "syntax elements are immutable, the result of update methods must be propagated to have any effect"]
     pub fn with_leading_trivia<'a, I>(self, trivia: I) -> Self
     where
-        I: Iterator<Item = (TriviaPieceKind, &'a str)> + Clone + ExactSizeIterator,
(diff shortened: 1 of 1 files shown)


MichaReiser:
!bench_analyzer

leops (approved):

crates/rome_rowan/src/syntax/token.rs
```diff
@@ -191,26 +191,27 @@ impl<L: Language> SyntaxToken<L> {
...
+        let mut token_text = String::new();
+        let trivia = trivia.map(|(kind, text)| {
+            token_text.push_str(text);
+            TriviaPiece::new(kind, TextSize::of(text))
+        });
+
+        let leading = GreenTrivia::new(trivia);
 
-        text.push_str(self.text_trimmed());
+        token_text.push_str(self.text_trimmed());
```
xunilrj, line 204: You may be able to squeeze a lit bit more juice calling ```self.raw.green().leading_trailing_total_len()```directly.

xunilrj (approved):

crates/rome_rowan/src/syntax/token.rs
MichaReiser: I implemented this change by slicing the `self.text()` directly.

MichaReiser:
!bench_analyzer

Issue #3029, opened 2022-08-08
Implement the low-level transport for the JavaScript API


leops:
Intended changes for this task are:
- Merge the CLI and Language Server binary (the server is run by calling `rome __run_server`)
- Add a new `rome __print_socket` command that ensures the server is running in a daemon process and prints the domain socket / named pipe it's listening on to stdout

ematipico:
@leops you might want to take a look at the old CLI commands as inspiration: 
- [lsp](https://github.com/rome/tools/blob/archived-js/website/src/_includes/docs/cli-commands.md#rome-lsp)
- [deamon commands](https://github.com/rome/tools/blob/archived-js/website/src/_includes/docs/cli-commands.md#rome-restart)

leops:
> @leops you might want to take a look at the old CLI commands as inspiration:
> 
>     * [lsp](https://github.com/rome/tools/blob/archived-js/website/src/_includes/docs/cli-commands.md#rome-lsp)
> 
>     * [deamon commands](https://github.com/rome/tools/blob/archived-js/website/src/_includes/docs/cli-commands.md#rome-restart)

One thing I'm considering is to prefix these CLI commands with a double underscore instead (like `__run_server` and `__print_socket`) to signify these are internal commands, not intended to be used outside of the stable JS wrapper library at the moment

ematipico:
It makes sense. `esbuild` just doesn't document them. It has a command called `node` and a bunch of arguments, for example `--ping` to check if the service is running.

PR #3031, opened 2022-08-08
refactor(rome_service): add schemars derive attributes for most serializable types

leops:
## Summary

This changes adds a build script to the Rust crate for the playground that automatically generates type definitions for the Workspace protocol. This is implemented by adding additional derives for the JsonSchema trait from the schemars crate to all serializable types. These derives macros generate JSON schema metadata for the API types at compilation time, which are used for two purposes: the most straightforward is generating a schema file for the rome.json configuration file (with the new cargo codegen-schema command) that's embedded in the VSCode extension and provides autocomplete and validation for the configuration file format. The second usage of these metadata is in the build.rs script for the playground, to generate TypeScript definitions from the schema

## Test Plan

This change is only additional compile-time metadata and using it to generate TypeScript definitions, which are tested when the plyaground is built with the TypeScript Compiler


crates/rome_service/Cargo.toml
```diff
@@ -24,14 +24,19 @@ rome_js_semantic = { path = "../rome_js_semantic" }
 rome_rowan = { path = "../rome_rowan" }
 rome_text_edit = { path = "../rome_text_edit" }
 indexmap = { version = "1.9.1", features = ["serde"] }
+schemars = { version = "0.8.10", features = ["indexmap1"], optional = true }
 
 [features]
+schemars = ["dep:schemars", "rome_formatter/serde"]
 serde_workspace = [
```
ematipico, line 31: I wonder if we should rename the feature to simply `workspace`? Mostly because now this PR adds a new constraint, which is the JSON schema and it's not directly related to serde.

editors/vscode/configuration_schema.json
```diff
@@ -0,0 +1,336 @@
+{
```
ematipico, line 1: This is a killer addition! It will force us to write better documentation for each property! This further demonstrates that we can use rust to extract the rustdoc for the documentation of our website.

crates/rome_service/Cargo.toml
leops: I eventually intend for the `serde_workspace` feature to go away (once it starts being used in the CLI / LSP binary as part of #3029) so the workspace types will always implement `Serialize` and `Deserialize`, but `schemars` will remain behind a feature flag since it's only needed in codegen

ematipico (approved):
Have you actually checked if VSCode provides autocomplete when typing in a `rome.json` file? It looks good to me. Just fixing CI and 🚢  it!

leops:
> Have you actually checked if VSCode provides autocomplete when typing in a `rome.json` file?

I had to fix a packaging issue since the schema wasn't being included in the extension bundle, but otherwise it seems to work correctly:

config

MichaReiser (approved):

PR #3033, opened 2022-08-09
feat(rome_js_syntax): Node factories for unknown nodes

MichaReiser:
This PR adds node factory methods for unknown nodes. This is certainly more of an edge case, but I have a use case in the formatter when encountering an invalid parenthesized expression in which case the node should be replaced
as an unknown node so, for as far as the formatter is concerned, parenthesized expressions do not exist.

The new API can be used like this:

```rust
let unknown = rome_js_factory::make::js_unknown_expression(
    root.slots().map(|slot| slot.into_syntax_element()),
);
 ```

## Alternatives

An alternative would be to directly use the `SyntaxNode::new_detached` APIs. This works but is less accessible and
it might be convenient to search for all places where we construct unknown nodes.

### DIFF
--- a/crates/rome_css_factory/src/generated/node_factory.rs
+++ b/crates/rome_css_factory/src/generated/node_factory.rs
@@ -812,3 +812,10 @@ where
         }),
     ))
 }
+pub fn css_unknown<I>(slots: I) -> CssUnknown
+where
+    I: IntoIterator<Item = Option<SyntaxElement>>,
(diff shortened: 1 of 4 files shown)


ematipico:
>  This is certainly more of an edge case, but I have a use case in the formatter when encountering an invalid parenthesized expression in which case the node should be replaced
as an unknown node

Shouldn't we change how the parser emits the CST/AST in this case?

leops (approved):

PR #3035, opened 2022-08-09
test(rome_cli): add snapshots for file system and console session

ematipico:
## Summary

This PR improves the CLI tests by creating a snapshot. The snapshot will contain:
- the content of the configuration file, if it exists
- the files contained inside the memory file system during the test run of a single test (excluded the configuration file)
- the messages printed by the console, if any (excluded the timing messages)

The reason why I created snapshots it's because I found myself at struggle sometimes, when I needed to test specific cases. Also, there have been few issue before where some messages where not supposed to be there, and I couldn't catch them early. The snapshots, somehow, might help the testing experience a bit.

The snapshot have been added to some existing cases, but not all of them. I had to exclude some runs because of some issue that I couldn't debug (some test stalls for some reason).

## Test Plan

The CI should not fail

### DIFF
--- a/crates/rome_cli/tests/snap_test.rs
+++ b/crates/rome_cli/tests/snap_test.rs
@@ -54,7 +54,11 @@ impl CliSnapshot {
                 //
                 // This is a workaround, and it might not work for all cases.
                 if !message_content.contains("files") {
+                    content.push_str("```block");
+                    content.push('\n');
                     content.push_str(message_content);
+                    content.push('\n');
+                    content.push_str("```");
                     content.push_str("\n\n")
                 }
             }
--- a/crates/rome_cli/tests/snap_test.rs
+++ b/crates/rome_cli/tests/snap_test.rs
@@ -3,6 +3,7 @@ use rome_console::{markup, BufferConsole, Markup};
 use rome_diagnostics::termcolor::NoColor;
 use rome_fs::{FileSystemExt, MemoryFileSystem};
 use std::collections::HashMap;
+use std::fmt::Write as _;
 use std::path::PathBuf;
 
 #[derive(Default)]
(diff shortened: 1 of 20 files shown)


leops:
I wonder if we could transition the CLI test suite to be fully fixture and snapshot based: each test is a directory containing an `fs` directory with the initial state of the memory filesystem, an `stdin` file containing the command line to be run and eventually the console inputs for interactive commands, and the snapshot file with the modified files and console output

ematipico:
> I wonder if we could transition the CLI test suite to be fully fixture and snapshot based: each test is a directory containing an `fs` directory with the initial state of the memory filesystem, an `stdin` file containing the command line to be run and eventually the console inputs for interactive commands, and the snapshot file with the modified files and console output

I really like the idea, I think this could work. But I would like to apply this proposal in another PR, because your will make a lot changes. If you're OK with the structure of the snapshots, I think we can merge it?

crates/rome_cli/tests/snap_test.rs
```diff
@@ -0,0 +1,111 @@
...
+                content.push('\n');
+                content.push_str("```");
+                content.push_str("\n\n")
+            }
+        }
+
+        if !self.messages.is_empty() {
+            content.push_str("## Messages\n\n");
+
+            for message in &self.messages {
```
leops, line 50: Nit: wrap the messages in a code block to have them rendered in a `<pre>` block in markdown previews and respect the line breaks

leops (approved):
I'm still a bit concerned about why the `upgrade_severity` test would suddenly start hanging, but I couldn't reproduce the issue locally (I tried to run the test with and without a debugger attached but the issue never manifested). Otherwise the format of the snapshots themselves seems alright to me

Issue #3028, opened 2022-08-08
🐛 `noCamelCase` doesn't work on method names


ematipico:
### Environment information

```bash
Rome: v0.8.0
```

### What happened?

The rule `js/noCamelCase` doesn't catch the name of a method inside a class.

Link to [playground](https://play.rome.tools/?code=class+B++%7B%0A++format_something%28%29+%7B%7D%0A%7D%0Alet+a_something%3B&lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&typescript=false&jsx=false#YwBsAGEAcwBzACAAQgAgACAAewAKACAAIABmAG8AcgBtAGEAdABfAHMAbwBtAGUAdABoAGkAbgBnACgAKQAgAHsAfQAKAH0ACgBsAGUAdAAgAGEAXwBzAG8AbQBlAHQAaABpAG4AZwA7AA==) here

### Expected result

It should emit an error

### Code of Conduct

- [X] I agree to follow Rome's Code of Conduct

PR #3037, opened 2022-08-09
fix(rome_js_analyze): use camel case for class members

xunilrj:
## Summary

Closes https://github.com/rome/tools/issues/3028.

## Test Plan

```
> cargo test -p rome_js_analyze -- camel
```


crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
```diff
@@ -91,15 +126,27 @@ impl Rule for UseCamelCase {
...
+
+        match ctx.query() {
+            JsAnyCamelCaseName::JsIdentifierBinding(binding) => {
+                let renamed =
+                    batch.rename_node_declaration_with_retry(model, binding.clone(), candidates);
+                if renamed {
+                    Some(JsRuleAction {
+                        category: ActionCategory::Refactor,
+                        applicability: Applicability::Always,
+                        message: markup! { "Rename this symbol to camel case" }.to_owned(),
```
leops, line 138: Nit: this message could be improved slightly by inspecting the surroundings of the `binding` node to guess a more precise term than "symbol" like "variable", "method", "function", "parameter" ...

leops (approved):

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
xunilrj: Done

PR #3043, opened 2022-08-10
feat(rome_cli): format code via `stdin`

### LINKED ISSUE #3042: being able to read piped `stdin`

ematipico:
## Summary

Closes #3042 

The implementation of `read` is very simple at the moment and it doesn't take in consideration the `prompt` for now. But it's there because it will be needed in the future.

I also enhanced the CLI snapshot to take the "in" messages in consideration.

## Test Plan

New test cases added

### DIFF
--- a/crates/rome_cli/src/execute.rs
+++ b/crates/rome_cli/src/execute.rs
@@ -75,7 +75,7 @@ impl ExecutionMode {
 }
 
 /// Based on the [mode](ExecutionMode), the function might launch a traversal of the file system
-/// or  
+/// or handles the stdin file.
 pub(crate) fn execute_mode(
     mode: ExecutionMode,
     mut session: CliSession,
@@ -102,6 +102,13 @@ pub(crate) fn execute_mode(
                 console.log(markup! {
                     {printed.as_code()}
                 });
+            } else {
+                console.log(markup! {
+                    {content}
+                });
+                console.error(markup!{
+                    <Warn>"The content was not formatted because the formatter is currently disabled."</Warn>
+                })
             }
         }
 
--- a/crates/rome_cli/src/execute.rs
+++ b/crates/rome_cli/src/execute.rs
@@ -1,0 +1,112 @@
+use crate::traversal::traverse;
+use crate::{CliSession, Termination};
+use rome_console::{markup, ConsoleExt};
+use rome_fs::RomePath;
+use rome_service::workspace::{
+    FeatureName, FixFileMode, FormatFileParams, OpenFileParams, SupportsFeatureParams,
+};
+use std::path::PathBuf;
+
+#[derive(Clone)]
+pub(crate) enum ExecutionMode {
+    /// This mode is enabled when running the command `rome check`
+    Check {
+        /// The maximum number of diagnostics that can be printed in console
+        max_diagnostics: u16,
+
+        /// The type of fixes that should be applied when analyzing a file.
+        ///
+        /// It's [None] if the `check` command is called without `--apply` or `--apply-suggested`
+        /// arguments.
+        fix_file_mode: Option<FixFileMode>,
+    },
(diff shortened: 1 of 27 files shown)


crates/rome_cli/tests/snapshots/format_stdin_with_errors.snap
```diff
@@ -0,0 +1,6 @@
+---
```
ematipico, line 1: I plan to enhance the snapshot testing, and writing also `Termination` errors, so it's useful to have it here with nothing, for now

crates/rome_cli/src/traversal.rs
```diff
@@ -298,7 +315,7 @@ fn print_messages_to_console(
     has_errors
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
```
ematipico, line 318: I had to remove the `Copy` because of `String` and `PathBuf`

crates/rome_cli/tests/main.rs
```diff
@@ -461,6 +462,86 @@ mod check {
...
+                })
+                .count(),
+            1
+        );
+
+        assert_cli_snapshot("downgrade_severity", fs, console);
+    }
+
+    #[test]
+    fn upgrade_severity() {
```
ematipico, line 507: I totally forgot to enable this test again. I moved it from the together with the other test in the `check` section. They were in the `format` section.

crates/rome_cli/src/traversal.rs
```diff
@@ -65,8 +65,25 @@ pub(crate) fn traverse(mode: TraversalMode, mut session: CliSession) -> Result<(
     let workspace = &*session.app.workspace;
     let console = &mut *session.app.console;
 
+    // don't do any traversal if there's some content coming from stdin
+    if let Some((path, content)) = mode.as_stdin_file() {
```
leops, line 69: This code doesn't seem to be relying on anything traversal-related, could it be moved to `commands/format.rs` ?

ematipico: I would like to make it generic, because I am sure that not only the formatter will take advantage of `stdin`.

crates/rome_cli/src/execute.rs
```diff
@@ -0,0 +1,112 @@
...
+        let workspace = &*session.app.workspace;
+        let console = &mut *session.app.console;
+        let rome_path = RomePath::new(path, 0);
+
+        if mode.is_format() {
+            let can_format = workspace.supports_feature(SupportsFeatureParams {
+                path: rome_path.clone(),
+                feature: FeatureName::Format,
+            });
+            if can_format {
```
ematipico, line 94: @leops what's your opinion here? At the moment if the formatter is disabled, the CLI doesn't write anything. Instead, should we write the original code?

leops: I guess we could write the original code to `stdout`, along with a warning in `stderr` to let the user know the formatter is disabled ?

ematipico: Makes sense, I will apply the change!

crates/rome_console/src/lib.rs
```diff
@@ -29,6 +30,9 @@ pub enum LogLevel {
 pub trait Console: Send + Sync + RefUnwindSafe {
     /// Prints a message (formatted using [markup!]) to the console
     fn print(&mut self, level: LogLevel, args: Markup);
+
+    /// It reads from a source, and if this source contains something, it's converted into a [String]
+    fn read(&mut self, prompt: Option<Markup>) -> Option<String>;
```
leops, line 35: I think we'll probably want to differentiate between a `read` method that takes no prompt and reads the entire input stream, and a `prompt` method that takes a prompt and only reads the input stream up to the first newline

leops (approved):

PR #3044, opened 2022-08-10
refactor(rome_bin): merge the CLI and Language Server

leops:
## Summary

This change merges the existing command line client and language server binaries into a single `rome_bin` create that exposes all the existing CLI features but can also be spawned as a daemon and provide a backend for the features of the CLI and the editor extension.

Communication with the daemon server is done using the flavor of the JSON-RPC protocol defined in the LSP specification. Since this protocol defines a `Content-Type` header we could eventually use a non-JSON serialization format for communications between the CLI and the server, for instance `bincode` with an additional compression like Brotli if necessary.

The transport layer for the protocol is OS dependent, it uses a domain socket created in the global temporary directory on Unix systems, and a named pipe on Windows. This is abstracted away for client libraries by providing a `rome __print_socket` command that ensures the daemon is running, prints the OS-specific "socket name" to the standard output and exit. For node.js clients like the VSCode extension, the resulting socket name can simply be passed to `net.connect` to open a connection to the server regardless of the underlying operating system.

Importantly, despite allowing for multiple clients to connect to the same server daemon at the same time, each client will be connected to an isolated Language Server / Workspace instance. The implications of allowing concurrent access to a single workspace instance are complex, and it will require more work to get it right while the immediate goal of this change is only to provide the necessary infrastructure for JavaScript client libraries to connect to.

## Test Plan

I've modified the CLI test suite to use "simulated" client-server communication, by creating an in-process instance of the server, wrapping it with an in-memory transport layer, and injecting a workspace client instance using this transport layer into the CLI. This has the advantage testing the protocol encoding and serialization while being lightweight since it doesn't require spawning a new process for every test, but doesn't actually test the OS-specific transport layer (which might be hard to test since it's technically a shared, global state on the machine running the tests)


crates/rome_bin/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+    let mut args = Arguments::from_env();
+
+    if args.contains("__print_socket") {
+        print_server_socket()
+    } else if args.contains("__run_server") {
+        run_server_session()
+    } else {
+        run_cli_session(args)
+    }
+}
```
MichaReiser, line 29: Can you add some documentation around how the `rome_bin` is intended to be used? 

Is it intentional that we have such "cryptic" arguments? Why not `rome start` for starting the server and `rome stop` to shut it down? 

Or a `rome debug` or `rome environment` command to collect information about the system?

crates/rome_bin/src/server.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+fn setup_tracing_subscriber() {
+    /// This filter enables:
+    /// - All spans and events at level info or higher
+    /// - All spans and events in the `rome_lsp` and `rome_js_parser` crates
+    const LOGGING_FILTER: &str = "info,rome_lsp=trace,rome_js_parser=trace";
+
+    let logs_dir = env::temp_dir().join("rome-logs");
+    let file_appender = tracing_appender::rolling::hourly(logs_dir, "server.log");
```
MichaReiser, line 36: Can we mention the logs in the `README` files?

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+/// Implementation of [WorkspaceTransport] for types implementing [AsyncRead]
+/// and [AsyncWrite]
+pub struct SocketTransport {
+    runtime: Runtime,
+    read_recv: UnboundedReceiver<Vec<u8>>,
+    write_send: UnboundedSender<Vec<u8>>,
+}
+
+impl SocketTransport {
+    pub fn new<T>(runtime: Runtime, socket: T) -> Self
```
MichaReiser, line 43: I find it unexpected that a `new` command an async task. I would recommend renaming the method to `start` or `open` to indicate that this is not a normal constructor function.

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+        let (read_send, read_recv) = mpsc::unbounded_channel();
+        let (write_send, mut write_recv) = mpsc::unbounded_channel::<Vec<_>>();
+
+        let read_task = async move {
+            loop {
+                let mut length = None;
+                let mut line = String::new();
+
+                loop {
+                    match socket_read.read_line(&mut line).await? {
```
MichaReiser, line 59: It would be useful to document this protocol somewhere or, at least, mention the protocol

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+                        // A read of 0 bytes means the connection was closed
+                        0 => {
+                            return Err(RomeError::IoError(io::Error::new(
+                                io::ErrorKind::UnexpectedEof,
+                                "unexpected EOF",
+                            )));
+                        }
+                        // A read of two bytes corresponds to the "\r\n" sequence
+                        // that indicates the end of the header section
+                        2 => break,
```
MichaReiser, line 69: Shouldn't we at least check that the bytes are `\r\n`

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+                                "Content-Type" => {
+                                    if !value.starts_with("application/json") {
+                                        return Err(RomeError::IoError(io::Error::new(
+                                            io::ErrorKind::Other,
+                                            "unsupported Content-Type encoding",
+                                        )));
+                                    }
+                                }
+                                _ => {
+                                    // Ignore unknown headers
```
MichaReiser, line 99: It may be helpful to log unknown headers or someone might have a very frustrating experience when they e.g. have a typo in 'Content-Type'

MichaReiser (commented):
What's the main motivation behind merging the CLI and LSP binaries? Or what workflows does it simplify?

crates/rome_bin/src/cli.rs
```diff
@@ -0,0 +1,20 @@
...
+use tokio::runtime::Runtime;
+
+use crate::service::open_transport;
+
+pub fn run_cli_session(args: Arguments) -> Result<(), Termination> {
+    setup_panic_handler();
+
+    // Try to open a connection to an existing Rome server socket, or create an
+    // in-process Workspace server instance if no daemon process is found
+    let runtime = Runtime::new().expect("could not instantiate the Tokio runtime");
```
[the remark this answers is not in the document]
leops: The initialization of the tokio runtime can fail if it cannot create its internal OS object (epoll on Linux, kqueue on macOS or IOCP on Windows), although it's really not supposed to fail if it ever does that's out of our control so it makes sense that this error should be a `Termination` instead of a panic

leops:
> What's the main motivation behind merging the CLI and LSP binaries? Or what workflows does it simplify?

The main reason behind this merge is to allow the `rome` npm package we currently use to distribute the CLI to also expose a programmatic interface with `import { ... } from 'rome'`. This JavaScript API will internally need to defer it implementation to a prebuilt binary of our Rust code, and it didn't really make sense to distribute two different binaries in the same npm package (since pretty much all the code below the Workspace level would be duplicated between the two). But this is intended as an implementation detail, at the moment the daemon server part is not a part of the public command line interface, hence why this new merged binary consists of the public CLI (from the `rome_cli` crate) + the backend server logic layered on top through cryptic internal commands. But I agree this separation is artificial and we'll probably end up exposing the commands for controlling the daemon on the public CLI (making the `rome_bin` crate redundant)

crates/rome_bin/src/service/mod.rs
leops: I added a link to the specification in the PR description

crates/rome_bin/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "rome_bin"
```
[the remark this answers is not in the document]
leops: I though it could be unclear to have a crate named just `rome` since it doesn't really explain what it does, but I could at least rename the resulting binary

crates/rome_bin/src/service/mod.rs
leops: I added an `eprintln` call there, but in practice the `SocketTransport` should only be used to connect to a `rome_lsp` instance
and this branch should only ever be run if there's a version mismatch between the running daemon and the CLI (this is also the case for the `Content-Type` header, the current version of the server doesn't send it since its assumed to be JSON by default, but while I envision unsupported headers can simply be ignored by the CLI the socket should return an error if the server sends in a response with an encoding it doesn't support)

crates/rome_lsp/src/server.rs
```diff
@@ -162,13 +161,94 @@ impl LanguageServer for LSPServer {
...
+    /// Destructure a connection into its inner service instance and socket
+    pub fn into_inner(self) -> (LspService<LSPServer>, ClientSocket) {
+        (self.service, self.socket)
+    }
+
+    /// Accept an incoming connection and run the server async I/O loop to
+    /// completion
+    pub async fn accept<I, O>(self, stdin: I, stdout: O)
+    where
+        I: AsyncRead + Unpin,
```
ematipico, line 247: I read a lot about `Unpin` and now here we use it. Would you mind explain a bit what's the use of `Unpin` here?

crates/rome_lsp/src/server.rs
```diff
@@ -162,13 +161,94 @@ impl LanguageServer for LSPServer {
...
+            .clone()
+            .unwrap_or_else(workspace::server_sync);
+
+        let mut builder = LspService::build(move |client| LSPServer::new(client, workspace));
+
+        builder = builder.custom_method(SYNTAX_TREE_REQUEST, LSPServer::syntax_tree_request);
+
+        // supports_feature is special because it returns a bool instead of a Result
+        builder = builder.custom_method("rome/supports_feature", |server: &LSPServer, params| {
+            ready(Ok(server.session.workspace.supports_feature(params)))
```
ematipico, line 208: What does `ready` do?

crates/rome_lsp/src/server.rs
```diff
@@ -162,13 +161,94 @@ impl LanguageServer for LSPServer {
...
+    /// same workspace from multiple client
+    workspace: Option<Arc<dyn Workspace>>,
+}
+
+/// Helper method for wrapping a [Workspace] method in a `custom_method` for
+/// the [LSPServer]
+macro_rules! workspace_method {
+    ( $builder:ident, $method:ident ) => {
+        $builder = $builder.custom_method(
+            concat!("rome/", stringify!($method)),
```
ematipico, line 180: nit: I might make sense to create a constant called `REQUEST_PREFIX` or something similar, at this point. Less typos.

crates/rome_lsp/src/requests/syntax_tree.rs
```diff
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 use tower_lsp::lsp_types::{TextDocumentIdentifier, Url};
 use tracing::info;
 
-pub const SYNTAX_TREE_REQUEST: &str = "rome/syntaxTree";
+pub const SYNTAX_TREE_REQUEST: &str = "rome_lsp/syntaxTree";
```
ematipico, line 8: Yeah I think it would to have two constants, one for `rome/` and the other for `rome_lsp/`.  So we can document them and teach when to use one or the other.

crates/rome_lsp/src/server.rs
```diff
@@ -162,13 +161,94 @@ impl LanguageServer for LSPServer {
...
+        let mut builder = LspService::build(move |client| LSPServer::new(client, workspace));
+
+        builder = builder.custom_method(SYNTAX_TREE_REQUEST, LSPServer::syntax_tree_request);
+
+        // supports_feature is special because it returns a bool instead of a Result
+        builder = builder.custom_method("rome/supports_feature", |server: &LSPServer, params| {
+            ready(Ok(server.session.workspace.supports_feature(params)))
+        });
+
+        workspace_method!(builder, update_settings);
```
ematipico, line 211: I understand why we create custom methods, but I fail to see where and when these custom methods are called. Could you help me please?

crates/rome_bin/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version = "0.1.0"
```
ematipico, line 3: This should be `0.0.0`

crates/rome_bin/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+edition = "2021"
```
ematipico, line 4:
```suggestion
edition = "2021"
authors = ["Rome Tools"]
license = "MIT"
repository = "https://github.com/rome/tools"
```

We should also add the `description` field. I leave it to you because you have a better context.

crates/rome_bin/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+tokio = { version = "1.15.0", features = ["io-std", "io-util", "net", "time", "rt", "rt-multi-thread"] }
+tower-lsp = { version = "0.17.0"}
+
+tracing = { version = "0.1.31", default-features = false, features = ["std"] }
+tracing-tree = "0.2.0"
+tracing-subscriber = { version = "0.3.5", features = ["env-filter"] }
+tracing-appender = "0.2"
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2.127"
```
ematipico, line 25: What's this used for?

crates/rome_bin/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Rome's main binary distribution, exposes the command line interface defined in
+`rome_cli`, and the language server interface defined in `rome_lsp` and used by
+the `rome` VSCode extension
+
+# Logs
+
+When the server is run in daemon mode, it will output logs to a file created in
+a `rome-logs` directory inside the system temporary directory. The log file
+will be rotated on a hourly basis.
```
ematipico, line 11: What does `rotated` mean in this context?

crates/rome_bin/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+mod cli;
```
ematipico, line 1: If we have both `main.rs` and `lib.rs`, which documentation is shown by default by rustdoc? If that's `lib.rs`, I think we should copy here the content of `README.md` here, to help with the documentation.

crates/rome_bin/src/main.rs
ematipico: I believe this is intentional and not set on stone. Eventually we might have those commands: https://github.com/rome/tools/issues/3029#issuecomment-1210612479

crates/rome_bin/src/server.rs
```diff
@@ -0,0 +1,50 @@
...
+use tracing_tree::HierarchicalLayer;
+
+use crate::service::{print_socket, run_daemon};
+
+pub fn run_server_session() -> Result<(), Termination> {
+    setup_tracing_subscriber();
+
+    let rt = Runtime::new()?;
+    let factory = ServerFactory::default();
+    let span = debug_span!("Running LSP Server", pid = std::process::id());
```
ematipico, line 17: The name of the function and the comment seems to be not in sync. Should the comment just say `Running Server`?

crates/rome_bin/src/server.rs
```diff
@@ -0,0 +1,50 @@
...
+    rt.block_on(run_daemon(factory).instrument(span))?;
```
ematipico, line 18: What does `block_on` do?

crates/rome_bin/src/server.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+fn setup_tracing_subscriber() {
+    /// This filter enables:
+    /// - All spans and events at level info or higher
+    /// - All spans and events in the `rome_lsp` and `rome_js_parser` crates
+    const LOGGING_FILTER: &str = "info,rome_lsp=trace,rome_js_parser=trace";
+
+    let logs_dir = env::temp_dir().join("rome-logs");
+    let file_appender = tracing_appender::rolling::hourly(logs_dir, "server.log");
```
ematipico, line 36: Is there any chance that this log file can contain the date in its name? Also, would you mind adding a comment saying that the specs of the logging are not set on stone yet? Or maybe create an issue... something to not forget this part at least.

crates/rome_bin/src/server.rs
```diff
@@ -0,0 +1,50 @@
...
+    Ok(())
+}
+
+pub fn print_server_socket() -> Result<(), Termination> {
+    let rt = Runtime::new()?;
+    rt.block_on(print_socket())?;
+    Ok(())
+}
+
+fn setup_tracing_subscriber() {
```
ematipico, line 29: Can we document what this subscribe do? I can see that it does a bunch of stuff: filtering, dumping logs into a file, and I see some other stuff at the end

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,205 @@
+use std::{io, panic::RefUnwindSafe};
```
ematipico, line 1: Would you mind adding a module documentation here and create a brief explanation of this module and why OS specific code is needed? You can also copy what you wrote in the description of the PR, it seemed quite clear to me.

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+#[cfg(unix)]
+mod unix;
+#[cfg(unix)]
+use self::unix::open_socket;
+#[cfg(unix)]
+pub(crate) use self::unix::{print_socket, run_daemon};
+
+/// Tries to open a connection to a running daemon instance, returning a
+/// [WorkspaceTransport] instance if the socket is currently active
+pub(crate) fn open_transport(runtime: Runtime) -> io::Result<Option<impl WorkspaceTransport>> {
```
ematipico, line 26: Does that mean that `Ok(None)` means that there's no active connection?

crates/rome_bin/src/service/mod.rs
ematipico: The link will be lost once the PR is closed. I think it's best have it have in our rustdoc, maybe in as module documentation.

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+                                "unexpected EOF",
+                            )));
+                        }
+                        // A read of two bytes corresponds to the "\r\n" sequence
+                        // that indicates the end of the header section
+                        2 => {
+                            if line != "\r\n" {
+                                return Err(RomeError::IoError(io::Error::new(
+                                    io::ErrorKind::Other,
+                                    "unexpected byte sequence",
```
ematipico, line 73: Is it possible to have a better messages? From the website:

> Unique and specific error messages. No generic error messages. This not only helps users understand what went wrong, but should provide maintainers with a unique call site and the necessary information to debug.

For example, here we are using a `IoError`, which is **really** generic and it's used in many callsites. While it makes sense to have this category, here we don't explain that the error occurred inside our internal transport protocol, 

For **this particular case** we don't suggest why the sequence in unexpected, we don't print what we expected and how the maintainer/user should fix the error. Would it be possible to put more info/thought around the error messages of the protocol?

crates/rome_bin/src/service/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+                                "Content-Type" => {
+                                    if !value.starts_with("application/json") {
+                                        return Err(RomeError::IoError(io::Error::new(
+                                            io::ErrorKind::Other,
+                                            "unsupported Content-Type encoding",
+                                        )));
+                                    }
+                                }
+                                _ => {
+                                    eprintln!("ignoring unknown header {name:?}");
```
ematipico, line 108: Would it make sense to debug trace these headers?

crates/rome_bin/src/service/unix.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// Spawn the daemon server process in the background
+fn spawn_daemon() -> io::Result<Child> {
+    let binary = env::current_exe()?;
+
+    let mut cmd = Command::new(binary);
+    cmd.arg("__run_server");
+
+    // Create a new session for the process and make it the leader, this will
+    // ensures the child process is fully detached from its parent and will
```
ematipico, line 40:
```suggestion
    // ensures that the child process is fully detached from its parent and will
```

crates/rome_bin/src/service/unix.rs
```diff
@@ -0,0 +1,127 @@
...
+                                "the server process exited before the connection could be etablished",
+                            ));
+                        }
+                        _ = time::sleep(Duration::from_millis(50)) => {}
+                    }
+                } else {
+                    // Spawn the daemon process and wait a few milliseconds for
+                    // it to become ready then retry the connection
+                    current_child = Some(spawn_daemon()?);
+                    time::sleep(Duration::from_millis(50)).await;
```
ematipico, line 100: Is it safe to rely on waiting for a server for an arbitrary amount of time? What if the daemon takes more than 50ms for reasons that are out of our control?

ematipico:
Now that we merged LSP and CLI, and `tokio` will be part of the CLI, I am curios to know how big is now our final binary. Any stats?

crates/rome_lsp/src/server.rs
```diff
@@ -16,24 +18,21 @@ pub struct LSPServer {
...
-        let session = Session::new(client);
-        Self { session }
+    fn new(client: Client, workspace: Arc<dyn Workspace>) -> Self {
+        Self {
+            session: Session::new(client, workspace),
+        }
     }
 
-    async fn syntax_tree_request(&self, params: SyntaxTreePayload) -> LspResult<Option<Value>> {
+    async fn syntax_tree_request(&self, params: SyntaxTreePayload) -> LspResult<String> {
```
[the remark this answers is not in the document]
leops: I think this might not actually be needed anymore, the code registering the custom methods on the LSP server went through a few iterations before I settled on this design, and in one of them the Rust type system had troubles inferring the correct generic types for this return value

crates/rome_lsp/src/server.rs
leops: The `Unpin` requirement on the `stdin` type here is inherited from `tower_lsp`, I guess because it needs to move the stream inside an async block (the state of a future is pinned while it's executing, so moving any value inside of it would require its type to be unpin-able)

crates/rome_lsp/src/server.rs
leops: The `custom_method` builder expects the provided callback to return a `Future`, so `ready` is used to construct a future that's immediately ready and returns the provided value (this would be equivalent to `Promive.resolve(value)` in JavaScript to create a promise that immediately resolves to the provided value)

crates/rome_lsp/src/server.rs
leops: The advantage of doing this operation with macros is that this all happens at compile time, the actual generated code only contains a simple static string literal (eg. `"rome/$method"`) which doesn't need to allocate anything since this value is static and contained in the generated binary. Unfortunately the `concat!` macro cannot resolve the value of a constant, so it would force the code to be written as `&format!(concat!("{REQUEST_PREFIX}/", stringify!($method)))` and allocate a temporary String

crates/rome_lsp/src/server.rs
leops: The code that calls these custom methods over the IPC transport is in [crates/rome_service/src/workspace/client.rs](https://github.com/rome/tools/blob/f87ecc163d0d89e240ad802091edb879c9d9890d/crates/rome_service/src/workspace/client.rs)

crates/rome_bin/Cargo.toml
leops: The `libc` crate contains bindings to the C standard library, in this case it's used to call the `setsid` function from the POSIX API to detach the daemon process from the CLI instance that's launching it

crates/rome_bin/README.md
leops: The logs will be appended to a file called `/tmp/rome-logs/server.log.yyyy-MM-dd-HH`, so on every hour change the previous file will be closed and a new one will be created. This helps ensuring the size of the files stay manageable, and allows old log files to be reclaimed by the OS temporary file remover (while the current log file is locked by the running daemon)

crates/rome_bin/src/lib.rs
leops: I think the documentation for `lib.rs` is what's emitted by Rustdoc, I'll try to include the content of the README in there

crates/rome_bin/src/server.rs
leops: The `block_on` method takes a `Future` and execute it in the runtime, blocking the current thread until the future has resolved (the value returned by the future is then returned by the `block_on` call). Essentially this is a way of running async code in the middle of sync code

crates/rome_bin/src/service/mod.rs
leops: Yes, this allows differentiating the case where no server is running from other possible I/O errors that may happen while initializing the connection

crates/rome_bin/src/service/unix.rs
leops: This code is running in a loop so if the server isn't ready once the 50 milliseconds timer expires the client will retry the connection (unless the daemon process has exited in the meantime)

crates/rome_bin/README.md
ematipico: Oh great! That was not immediate. I didn't check the details of the API provided by the tracer crates.

leops:
> Now that we merged LSP and CLI, and `tokio` will be part of the CLI, I am curios to know how big is now our final binary. Any stats?

This is only for Windows right now, but here are the binary size I'm getting:

| Binary | Approx. size | Exact size |
|----------|---------|------------------|
| rome_cli | 4,54 MiB | 4 761 088 bytes |
| rome_lsp | 8,17 MiB | 8 573 952 bytes |
| rome_cli + rome_lsp | 12,71 MiB | 13 335 040 bytes |
| rome_bin | 9,62 MiB | 10 094 080 bytes |

crates/rome_bin/src/service/mod.rs
leops: I overhauled the way errors are emitted in the socket transport, using `anyhow` to print structured error messages:
```
remote connection read task exited with an error

Caused by:
    0: failed to parse header from the remote workspace
    1: invalid value for Content-Type expected "application/vscode-jsonrpc", got "application/bincode"
```

ematipico (approved):
Thank you for reviewing the error part! It looks way better now

PR #3045, opened 2022-08-11
feature(rome_js_semantic): Semantic model supporting "module.exports"

xunilrj:
## Summary

This PR improves the semantic model ```ìs_exported``` query to take ```module.exports``` into consideration.

## Test Plan

```
 cargo test -p rome_js_semantic -- ok_semantic_model_is_exported
```

### DIFF
--- a/crates/rome_js_semantic/src/events.rs
+++ b/crates/rome_js_semantic/src/events.rs
@@ -720,18 +720,19 @@ impl SemanticEventExtractor {
         let reference_parent = reference.parent();
         let reference_greatparent = reference_parent.as_ref().and_then(|p| p.parent());
 
-        // export
+        // check export keyword
         matches!(
             reference_parent.as_ref().map(|p| p.kind()),
             Some(JS_EXPORT_NAMED_SHORTHAND_SPECIFIER | JS_EXPORT_NAMED_SPECIFIER)
         ) | {
-            let is_export_default = matches!(
+            // check "export default" keyword
+            matches!(
                 reference_greatparent.as_ref().map(|p| p.kind()),
                 Some(JS_EXPORT_DEFAULT_EXPRESSION_CLAUSE)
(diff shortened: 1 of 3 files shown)


crates/rome_js_semantic/src/events.rs
```diff
@@ -667,13 +720,44 @@ impl SemanticEventExtractor {
...
+    }
+
+    fn is_assignment_left_side_module_exports(&self, node: &JsSyntaxNode) -> bool {
+        match node.kind() {
+            JsSyntaxKind::JS_ASSIGNMENT_EXPRESSION => {
+                let expr = node.clone().cast::<JsAssignmentExpression>();
+                match expr.and_then(|x| x.left().ok()) {
+                    Some(JsAnyAssignmentPattern::JsAnyAssignment(
+                        JsAnyAssignment::JsStaticMemberAssignment(a),
+                    )) => a.syntax().text_trimmed() == "module.exports",
```
leops, line 755: We should probably check each identifier individually, in case someone decides to write `module. /*comment*/ exports = ...`

ematipico: I chime in and add another case that we have, unfortunately, to handle: 

```js
module.exports.namedCjsExport = function namedCjsExport() {};
```

ematipico (commented):
Are cases like:

```js
exports.SimpleMessage = 'Hello world';
```

part of this PR of you plan to add them with another PR?

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

xunilrj:
> part of this PR of you plan to add them with another PR?
> I chime in and add another case that we have, unfortunately, to handle:

Done

ematipico (approved):

PR #3053, opened 2022-08-12
chore: update rust-toolchain to 1.63.0

IWANABETHATGUY:
## Summary
1. The latest stable rust-toolchain has been released, see https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html
2. Fixing some lint errors introduced by this upgrade.

## Test Plan
1. CI should pass.

### DIFF
--- a/crates/rome_analyze/src/visitor.rs
+++ b/crates/rome_analyze/src/visitor.rs
@@ -1,3 +1,4 @@
+#![allow(rustdoc::broken_intra_doc_links)]
 use std::collections::BinaryHeap;
 
 use rome_diagnostics::file::FileId;
--- a/crates/rome_js_analyze/tests/spec_tests.rs
+++ b/crates/rome_js_analyze/tests/spec_tests.rs
@@ -128,7 +128,7 @@ fn markup_to_string(markup: Markup) -> String {
 
     String::from_utf8(buffer).unwrap()
 }
-
+#[allow(clippy::let_and_return)]
 fn diagnostic_to_string(name: &str, source: &str, diag: Diagnostic) -> String {
     let file = SimpleFile::new(name.into(), source.into());
     let text = markup_to_string(markup! {
--- a/crates/rome_rowan/src/green/token.rs
+++ b/crates/rome_rowan/src/green/token.rs
@@ -62,7 +62,7 @@ impl ToOwned for GreenTokenData {
 impl Borrow<GreenTokenData> for GreenToken {
     #[inline]
     fn borrow(&self) -> &GreenTokenData {
-        &*self
+        &**self
     }
 }
 
@@ -79,14 +79,14 @@ impl fmt::Debug for GreenTokenData {
 
 impl fmt::Debug for GreenToken {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let data: &GreenTokenData = &*self;
+        let data: &GreenTokenData = &**self;
         fmt::Debug::fmt(data, f)
     }
 }
 
 impl fmt::Display for GreenToken {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let data: &GreenTokenData = &*self;
+        let data: &GreenTokenData = &**self;
         fmt::Display::fmt(data, f)
     }
 }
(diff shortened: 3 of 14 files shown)


crates/rome_js_analyze/tests/spec_tests.rs
```diff
@@ -128,7 +128,7 @@ fn markup_to_string(markup: Markup) -> String {
 
     String::from_utf8(buffer).unwrap()
 }
-
+#[allow(clippy::let_and_return)]
```
IWANABETHATGUY, line 131: false positive case of clippy, 

```
error: returning the result of a `let` binding from a block
   --> crates/rome_js_analyze/tests/spec_tests.rs:138:5
    |
134 | /     let text = markup_to_string(markup! {
135 | |         {diag.display(&file)}
136 | |     });
    | |_______- unnecessary `let` binding
137 |
138 |       text
    |       ^^^^
    |
    = note: `-D clippy::let-and-return` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_and_return
help: return the expression directly
    |
134 ~     
135 | 
136 ~     markup_to_string(markup! {
137 +         {diag.display(&file)}
138 +     })
```
If I apply this suggesion,

rustc will emit: 
```
error[E0597]: `file` does not live long enough
   --> crates/rome_js_analyze/tests/spec_tests.rs:135:23
    |
134 |       markup_to_string(markup! {
    |  ______________________-
135 | |         {diag.display(&file)}
    | |                       ^^^^^ borrowed value does not live long enough
136 | |     })
    | |_____- a temporary with access to the borrow is created here ...
137 |
138 |   }
    |   -
    |   |
    |   `file` dropped here while still borrowed
    |   ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `impl rome_console::fmt::Display`
```

crates/rome_rowan/src/green/token.rs
```diff
@@ -79,14 +79,14 @@ impl fmt::Debug for GreenTokenData {
 
 impl fmt::Debug for GreenToken {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let data: &GreenTokenData = &*self;
+        let data: &GreenTokenData = &**self;
```
IWANABETHATGUY, line 82:
```
error: deref on an immutable reference
  --> crates/rome_rowan/src/green/token.rs:82:37
   |
82 |         let data: &GreenTokenData = &*self;
```
same for other deference fixings.

IWANABETHATGUY:
The two workarounds should open related tracking issues.

crates/rome_analyze/src/visitor.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(rustdoc::broken_intra_doc_links)]
```
IWANABETHATGUY, line 1: false positive Clippy error introduced by this upgrade.

IWANABETHATGUY:
Tracked here: https://github.com/rome/tools/issues/3054

leops:
Besides fixing the immediate warning and errors added in the upgrade, we should also track the possible improvements we can have thanks to the newly stabilized APIs:
- With the standard library `Mutex` and `RwLock` being const-constructible it's now possible to use these in the `DiffReport` of the Prettier test suite
- From Rust 1.62 it's now possible to derive the `Default` trait on enums, we could track down and remove the manual implementations for these
- From Rust 1.61 the `Termination` trait and `ExitCode` type are now stable, meaning we could implement stable exit codes for the CLI (this could be done in the incoming refactor of diagnostics)

IWANABETHATGUY:
> 

You are right, would you mind opening a new umbrella to tracking these issues? Because It seems that I don't have permission to create an umbrella issue with full functionality (I could not convert the todo items into separate task.)

crates/rome_js_analyze/tests/spec_tests.rs
IWANABETHATGUY: Yes, I think. I will do it later.

MichaReiser (approved):

MichaReiser:
!bench_formatter

MichaReiser:
!bench_parser

IWANABETHATGUY:
The regression seems not ideal, do we need to revert this?

MichaReiser:
I realized it too late that running the benchmarks after merging is useless because it then compares main (which includes 1.63) vs this branch (which is 1.63). 

But it might be worth running the benchmarks locally before/after this commit.

Issue #3046, opened 2022-08-11
☂️ Formatter 0.9.0 


MichaReiser:
### Description

This is an umbrella issue for the work on the Formatter for the 0.9.0 release. 

## Goals

* Rome removes unnecessary parentheses and adds parentheses to improve readability. 
* Improve Prettier compatibility of some core syntax nodes. 

## Tasks 

* [x] Parentheses (2 weeks) #2427 
  * [x] Implement parentheses handling for expressions  #3057, #3079 #2449 
  * [x] Implement parentheses handling for assignments #3083 
  * [x] Implement parentheses handling for types #3108
  * [x] Remove parenthesized nodes from formatting tree #3092
* [x] Prettier compatibility (2 weeks)
  * [x] `JsConditionalExpression`
  * [x] Hug function parameters
  * [x] Parameter and return type formatting
  * [x] return and throw argument
  * [x] `JsTemplate`  #2446 #3063
  * [x] `JsBinaryExpression` #3079
  * [x] `JsArrowFunctionExpression` #3122 
  * [x] `JsArrayExpression` #3126
  * [x] `JsStaticMemberAssignment` and `JsComputedMemberAssignment` #3061
  * [x] JSX formatting #3144
  * [x] JSX in Conditional Expressions #3150

ematipico:
This is a follow up question after https://github.com/rome/tools/pull/3057.

Is there a specific reason why the "needs parentheses" information is tracked inside the formatter phase?

Have considered some other solutions, such as a visitor approach, or even computing this information during the parsing phase and storing it inside our green tree?

I would like to know what are the implications of your decisions in the future developments

MichaReiser:
@ematipico I [replied](https://github.com/rome/tools/pull/3057#issuecomment-1214664439) on the PR as these questions seem relevant to the implementation.

MichaReiser:
I start working on the Arrow function formatting

MichaReiser:
I'll start looking into JSX formatting tomorrow

IWANABETHATGUY:
We are getting close to 90% prettier compatibility, that's awesome.

PR #3057, opened 2022-08-12
feat(rome_js_formatter): Parenthesizing expressions

MichaReiser:
## Summary

Part of #3046 

This PR adds the capability of adding parentheses to improve readability or removing parentheses if they aren't strictly necessary. 

The core architecture is around the `NeedsParentheses` trait that every `JsAnyExpression` now implements. Its main method is `needs_parentheses` which must return `true` if parentheses are necessary if removing the parentheses would change the semantics OR result in a syntax error. `needs_parentheses` may also return `true` in cases where parentheses aren't strictly necessary but improve readability. 

The new `NeedsParentheses` trait is used in the `FormatNodeRule` trait that gets called when formatting every node. It now adds parentheses if `needs_parentheses` returns `true`. 

`FormatJsParenthesizedExpression` has been changed to **always** remove parentheses and rely on `FormatNodeRule` to re-insert parentheses if necessary. This has the downside that any trailing comments of `(` and leading comments of `)` are pushed outside the parentheses. I think that's OK and is something we can address when working on #2768. 

**Note**: This PR implements `needs_parentheses` for all binary like expression but they aren't used yet. I struggled to integrate the logic into the existing binary like formatting and intend to tackle it with anyway needed rewrite to closer match prettier's formatting. I plan to do this next.

## Performance

This PR significantly reduces Rome performance because it is now necessary to call `expression.resolve()` or `expression.resolve_parent` to skip over parenthesized expressions when testing if a child or parent is of a certain kind. 

The only alternative that I see to this approach would be to remove all `ParenthesizedExpressions` before formatting but this will have its own performance cost, requires source maps to map back to the original code in case a node must be printed in verbatim (to recover the parentheses), and is something we still have the chance to do in the future. 

My main concern is more about the added complexity and it's easy to forget a `resolve_expression` or `resolve_expression_parent` call. I don't have good answer to this and there are likely still a few places where I missed to add these calls myself. My recommendation is that we can fix these over time. 

```
group                                    main                                    parentheses
-----                                    ----                                    -----------
formatter/checker.ts                     1.00   219.6±15.94ms    11.8 MB/sec     1.04   227.3±18.85ms    11.4 MB/sec
formatter/compiler.js                    1.00    124.8±1.59ms     8.4 MB/sec     1.12    140.0±9.95ms     7.5 MB/sec
formatter/d3.min.js                      1.00    103.2±7.78ms     2.5 MB/sec     1.16    119.6±8.73ms     2.2 MB/sec
formatter/dojo.js                        1.00      8.1±0.02ms     8.5 MB/sec     1.00      8.1±0.62ms     8.5 MB/sec
formatter/ios.d.ts                       1.00    144.8±1.88ms    12.9 MB/sec     1.09   158.0±15.89ms    11.8 MB/sec
formatter/jquery.min.js                  1.00     29.7±1.87ms     2.8 MB/sec     1.03     30.7±2.13ms     2.7 MB/sec
formatter/math.js                        1.01   235.5±14.90ms     2.7 MB/sec     1.00   232.7±18.55ms     2.8 MB/sec
formatter/parser.ts                      1.02      5.2±0.40ms     9.3 MB/sec     1.00      5.1±0.32ms     9.5 MB/sec
formatter/pixi.min.js                    1.00    127.5±6.82ms     3.4 MB/sec     1.05   134.3±10.27ms     3.3 MB/sec
formatter/react-dom.production.min.js    1.04     37.3±2.29ms     3.1 MB/sec     1.00     35.7±1.18ms     3.2 MB/sec
formatter/react.production.min.js        1.00  1861.6±130.45µs     3.3 MB/sec    1.12      2.1±0.13ms     3.0 MB/sec
formatter/router.ts                      1.00      4.1±0.31ms    15.1 MB/sec     1.13      4.6±0.19ms    13.3 MB/sec
formatter/tex-chtml-full.js              1.00    274.8±6.53ms     3.3 MB/sec     1.14   314.6±22.55ms     2.9 MB/sec
formatter/three.min.js                   1.00   135.7±10.68ms     4.3 MB/sec     1.07    145.5±9.52ms     4.0 MB/sec
formatter/typescript.js                  1.00   911.6±60.87ms    10.4 MB/sec     1.03   938.6±58.19ms    10.1 MB/sec
formatter/vue.global.prod.js             1.00     42.9±0.15ms     2.8 MB/sec     1.25     53.4±3.34ms     2.3 MB/sec
```

## Prettier Difference

I decided to diverge from Prettier's formatting in two cases:

## `ObjectExpression`, `FunctionExpression`, or `ClassExpression` at the beginning of a statement. 

```javascript
// input
{}.b

// Prettier
({}.b)

// Rome
({}).b 
```

Prettier parenthesizes the whole expression whereas Rome parenthesizes the object/function/class expression only. 

The main reason for diverging is that parenthesizing the whole expression requires that the logic is implemented in any expression that starts with another child expression (tagged template, binary expressions, computed member/assignment, sequence, conditional, ... ) and it requires traversing the first expression until it reaches an object expression or any expression that doesn't start with another expression. This is rather expensive and Rome's approach avoids the expensive traversal except for object/function and class expressions, of which there should be fewer. 

## Parenthesized optional chain

```javascript
// Input
(a?.b).c
a?.b.c

// Prettier
(a?.b).c
a?.b.c

// Rome
a?.b.c
a?.b.c
```

Prettier keeps the parentheses if they were present in the source document but never adds them if they were missing. This is the only case where I found that the fact that parens are present in the source document changed how Prettier parenthesizes formatted expression. 

My take is that Rome should apply a consistent formatting regardless if parentheses were or were not present in the source document. 

## Test Plan

**Average compatibility**: 81.82 -> 81.54%
**Compatible lines**: 78.64 -> 78.34%

This PR is **regressing** the compatibility. Mainly due to the following reasons:

* Prettier has custom handling to avoid removing parentheses around closure type cast comments which Rome has not. This is something we can add support for in the future. The way I envision this to work is that we add a `tag` to every `JsParenthesizedExpression` that has a type cast comment (CC: @leops for a use case for node annotations)
* Our handling of comments differs from Prettier's which results in a large number of line changes. We'll tackle this as part of #2768 
* Prettier incompatibility of other syntax like `JsCallArguments`, `JsTemplate`, `JsBinaryLikeExpression`, member like expressions and more. 

I'm in favour of merging this PR regardless of the regressions because these issues should be tackled when working on comments, or when improving the formatting of other syntaxes.


MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/lib.rs
```diff
@@ -412,16 +414,29 @@ where
...
-        Ok(())
+            self.fmt_fields(node, f)
+        }
     }
 
     /// Formats the node's fields.
     fn fmt_fields(&self, item: &N, f: &mut JsFormatter) -> FormatResult<()>;
+
+    /// Returns whether the node requires parens.
+    fn needs_parentheses(&self, item: &N) -> bool {
```
MichaReiser, line 436: Ideally, the default implementation for `needs_parentheses` would be:

```
fn needs_parentheses(&self, item: &N) -> bool where N: NeedsParentheses {
  item.needs_parentheses()
}

fn needs_parentheses(&self, item: &N) -> bool where N: !NeedsParentheses {
  false
}
```

However, negative type constraints do not exist. If anyone has an idea how we can get a similar behaviour please let me know (CC: @leops, @xunilrj)

MichaReiser:
> Is there a specific reason why the "needs parentheses" information is tracked inside the formatter phase?

> Have you considered some other solutions, such as a visitor approach, or even computing this information during the parsing phase and storing it inside our green tree?

The main reasons for computing whether an expression needs to be parenthesized during the formatting phase are:

* The formatter adds more parentheses than necessary to not change precedence. For example, Rome adds parentheses around shift operations to ease readability. 
* The information which nodes need parentheses would become outdated during tree mutations or, it would be necessary to update this information when mutating the tree. Computing this information isn't cheap because it involves many tree navigation (multiple allocations). 
* The knowledge of which expressions need parentheses aren't needed for all use cases of the parser and computing the information during the parse phase would add an unnecessary cost (performance penalty). Besides, an over-approximation
* of which nodes require parentheses is stored in the green tree by having `ParenthesizedExpression`. Agreed, the tree may have more parenthesized expressions than necessary, but all the required once are in place. 
* Our CST doesn't support storing additional data, e.g. by using annotations on the tree. 


> I would like to know what are the implications of your decisions in the future developments

The main implication is that it increases complexity as well as the risk for subtle bugs when we forget to call `resolve_expression` or `resolve_expression_parent`. 


There are a few alternatives to this design, some of which I want to explore after I rewrote binary expression formatting:

* Pre-compute `needs_parentheses`: An option is to pre-compute `needs_parentheses` when extracting the suppression comments and storing the information on the `FormatContext`. This would remove the overhead for resolving the parent for every node. I'm doubtful that this will yield a big performance and it increases complexity a bit, because knowing if something needs parentheses then needs a `FormatContext`. We can try this out as a dedicated PR if we're interested. 
* Remove parenthesized expressions: The idea here is to pre-process the tree to remove all parenthesized expressions (except the ones that have closure type cast comments, or have any skipped token trivia attached to the left or right parens). This will have its own performance overhead but removes the need for calling `resolve_expression` and `resolve_expression_parent` which I see as the main drawback of the current solution. This is something I want to explore after I finished my work on the binary expression formatting. The main challenge is that I'll need to figure out a way to map positions from the transformed tree back to the original tree (source maps?) to be able to print a node in verbatim node if it's formatting (or any child) fails because of a missing child node.

MichaReiser:
!bench_formatter

MichaReiser:
!bench_formatter

ematipico:
> The formatter adds more parentheses than necessary to not change precedence. For example, Rome adds parentheses around shift operations to ease readability.
> 

How does this influence the decision while considering other approaches? 

> The information which nodes need parentheses would become outdated during tree mutations or, it would be necessary to update this information when mutating the tree. Computing this information isn't cheap because it involves many tree navigation (multiple allocations).
> 

Could you give me more information here? In which context an information becomes outdated? Are you referring to a potential visitor approach? Or else?

>   The knowledge of which expressions need parentheses aren't needed for all use cases of the parser and computing the information during the parse phase would add an unnecessary cost (performance penalty). 

What if one day this information will be needed for some other tool, for example a minifier. How hard/complex/easy would be to refactor this piece of logic? 


> Besides, an over-approximation of which nodes require parentheses is stored in the green tree by having `ParenthesizedExpression`. Agreed, the tree may have more parenthesized expressions than necessary, but all the required once are in place.

I didn't understand this bullet point.



> There are a few alternatives to this design, some of which I want to explore after I rewrote binary expression formatting:

> Pre-compute `needs_parentheses`: An option is to pre-compute `needs_parentheses` when extracting the suppression comments and storing the information on the `FormatContext`. This would remove the overhead for resolving the parent for every node. I'm doubtful that this will yield a big performance and it increases complexity a bit, because knowing if something needs parentheses then needs a `FormatContext`. We can try this out as a dedicated PR if we're interested.
> 
>     * Remove parenthesized expressions: The idea here is to pre-process the tree to remove all parenthesized expressions (except the ones that have closure type cast comments, or have any skipped token trivia attached to the left or right parens). This will have its own performance overhead but removes the need for calling `resolve_expression` and `resolve_expression_parent` which I see as the main drawback of the current solution. This is something I want to explore after I finished my work on the binary expression formatting. The main challenge is that I'll need to figure out a way to map positions from the transformed tree back to the original tree (source maps?) to be able to print a node in verbatim node if it's formatting (or any child) fails because of a missing child node.

Could you explain a bit, as an overview, how these bullet points are related to the solution of this PR? Does that mean that the solution of this PR is not the final one? Or are they complementary? If these bullet points are real alternatives, how do you plan to actually understand the pros and cons and eventually propose a final algorithm? 

Just asking because now, I don't know what I have to review in this PR.

MichaReiser:
> > The formatter adds more parentheses than necessary to not change precedence. For example, Rome adds parentheses around shift operations to ease readability.
> 
> How does this influence the decision while considering other approaches?

The main difference this makes is that the problem could otherwise be simplified to only remove unnecessary parentheses because it should be safe for the formatter to assume that the input tree is correctly parenthesized. I didn't pursue this option because it doesn't achieve our goal to be prettier compatible. 

> > The information which nodes need parentheses would become outdated during tree mutations or, it would be necessary to update this information when mutating the tree. Computing this information isn't cheap because it involves many tree navigation (multiple allocations).
> 
> Could you give me more information here? In which context an information becomes outdated? Are you referring to a potential visitor approach? Or else?

Let's say you start with 

```
let a = await call();
a[member];
```

`await call()` doesn't need any parentheses

Now let's do an inline refactor of `a`

```
(await call())[member]
```

It now becomes necessary to parenthesis `await call()` to form a valid syntax tree, meaning, it would be necessary to update the information if `await call()` needs parentheses or not. 


> > The knowledge of which expressions need parentheses aren't needed for all use cases of the parser and computing the information during the parse phase would add an unnecessary cost (performance penalty).
> 
> What if one day this information will be needed for some other tool, for example a minifier. How hard/complex/easy would be to refactor this piece of logic?

You could split the `NeedParentheses` trait to have a method that only returns `true` if parentheses are necessary to avoid syntax errors and another that may also return `true` for "nicety" parentheses. 


> > Besides, an over-approximation of which nodes require parentheses is stored in the green tree by having `ParenthesizedExpression`. Agreed, the tree may have more parenthesized expressions than necessary, but all the required once are in place.
> 
> I didn't understand this bullet point.

What I'm saying is that `ParenthesizedExpression` expressions inside of the tree encode where parentheses may be necessary. It may be possible to remove some, but if we keep the same amount of parenthesized expressions, then it's guaranteed that the tree remains valid IF the input tree was valid.

If this doesn't help, can you explain what you didn't understand

> > There are a few alternatives to this design, some of which I want to explore after I rewrote binary expression formatting:
> 
> > Pre-compute `needs_parentheses`: An option is to pre-compute `needs_parentheses` when extracting the suppression comments and storing the information on the `FormatContext`. This would remove the overhead for resolving the parent for every node. I'm doubtful that this will yield a big performance and it increases complexity a bit, because knowing if something needs parentheses then needs a `FormatContext`. We can try this out as a dedicated PR if we're interested.
> > Remove parenthesized expressions: The idea here is to pre-process the tree to remove all parenthesized expressions (except the ones that have closure type cast comments, or have any skipped token trivia attached to the left or right parens). This will have its own performance overhead but removes the need for calling `resolve_expression` and `resolve_expression_parent` which I see as the main drawback of the current solution. This is something I want to explore after I finished my work on the binary expression formatting. The main challenge is that I'll need to figure out a way to map positions from the transformed tree back to the original tree (source maps?) to be able to print a node in verbatim node if it's formatting (or any child) fails because of a missing child node.
> 
> Could you explain a bit, as an overview, how these bullet points are related to the solution of this PR? Does that mean that the solution of this PR is not the final one? Or are they complementary? If these bullet points are real alternatives, how do you plan to actually understand the pros and cons and eventually propose a final algorithm?

These alternatives would build up on top of this PR. This PR was the most straightforward approach to get something working. 

* Pre-compute: uses the same idea but instead of calling `node.needs_parentheses` inside of `FormatNodeRule` it would call `context.needs_parentheses(node)` which would perform a `Set` lookup. The `Set` of nodes that needs parentheses would be computed ahead of time when extracting the suppressions. The only thing that changes is when `node.needs_parentheses` is called. 
* Remove parenthesized expressions: This will revert the changes to `ExpressionNode.resolve` but the core logic around `NeedsParentheses` with all the implementations will remain the same. Therefore, this is another incremental change.

crates/rome_js_formatter/src/lib.rs
leops: Would it be possible to move the methods of `NeedsParentheses` into `FormatNodeRule` with a default implementation (and `resolve_parent` to an `AstNode` extension trait I guess) ? This could remove the need for an additional level of indirection

MichaReiser: The reason I prefer having an explicit `NeedsParentheses` trait is that it feels more ergonomic to call `node.needs_parentheses` rather than `FormatAnyExpressionRule::needs_parentheses(&node)`. 

I also do like the fact that `resolve_parent` isn't exposed for nodes where it isn't necessary to go through that extra loop. For example, you won't need to call `resolve_parent` on a statement.

crates/rome_js_formatter/src/parentheses.rs
```diff
@@ -0,0 +1,933 @@
...
+    }
+
+    fn resolve_parent(&self) -> Option<JsSyntaxNode> {
+        resolve_parent(self.syntax())
+    }
+
+    /// Returns `true` if this node requires parentheses to form valid syntax or improve readability.
+    ///
+    /// Returns `false` if the parentheses can be omitted safely without changing semantics.
+    fn needs_parentheses_with_parent(&self, parent: &JsSyntaxNode) -> bool;
```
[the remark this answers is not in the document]
MichaReiser: `NeedsParentheses` is only implemented by node kinds that may need parentheses (e.g. not for statement). That's why I think it's desirable that an explicit implementation, even if it is just returning `false` is necessary.

crates/rome_js_formatter/src/lib.rs
```diff
@@ -241,6 +241,8 @@
 //! If tests that are inside the `ok/` folder fail or if tests that are inside the `err/`
 //! folder don't emit, the whole test suite will fail.
 
+extern crate core;
```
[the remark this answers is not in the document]
MichaReiser: Hmm, I wonder why rust analyzer keeps adding this import :D

crates/rome_js_formatter/src/parentheses.rs
```diff
@@ -0,0 +1,933 @@
...
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyFunctionBody, JsAnyLiteralExpression, JsArrowFunctionExpression,
+    JsAssignmentExpression, JsBinaryExpression, JsBinaryOperator, JsComputedMemberAssignment,
+    JsComputedMemberExpression, JsConditionalExpression, JsLanguage, JsSequenceExpression,
+    JsSyntaxKind, JsSyntaxNode,
+};
+use rome_rowan::AstNode;
+
+/// Node that may be parenthesized to ensure it forms valid syntax or to improve readability
+pub trait NeedsParentheses: AstNode<Language = JsLanguage> {
```
ematipico, line 12: Is there a specific reason why this trait is JS specific?

crates/rome_js_formatter/src/parentheses.rs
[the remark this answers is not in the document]
ematipico: Could you add this information in a comment? So we don't lose the reason why we don't want a default implementation.

crates/rome_js_formatter/src/parentheses.rs
```diff
@@ -0,0 +1,933 @@
...
+        self.resolve().into_syntax()
+    }
+
+    /// Consumes `self` and returns the [JsSyntaxNode] of the first expression that isn't a parenthesized expression.
+    fn into_resolved_syntax(self) -> JsSyntaxNode {
+        self.into_resolved().into_syntax()
+    }
+}
+
+/// Resolves to the first parent that isn't a parenthesized expression, assignment, or type.
```
ematipico, line 46: Why assignments and types are excluded by this logic?

crates/rome_js_formatter/src/js/expressions/arrow_function_expression.rs
```diff
@@ -78,55 +84,144 @@ impl FormatNodeRule<JsArrowFunctionExpression> for FormatJsArrowFunctionExpressi
...
+    use crate::{assert_needs_parentheses, assert_not_needs_parentheses};
+    use rome_js_syntax::{JsArrowFunctionExpression, SourceType};
+
+    #[test]
+    fn needs_parentheses() {
+        assert_needs_parentheses!("new (a => test)()`", JsArrowFunctionExpression);
+        assert_needs_parentheses!("(a => test)()", JsArrowFunctionExpression);
+        assert_needs_parentheses!("(a => test).member", JsArrowFunctionExpression);
+        assert_needs_parentheses!("(a => test)[member]", JsArrowFunctionExpression);
+        assert_not_needs_parentheses!("object[a => a]", JsArrowFunctionExpression);
```
ematipico, line 205: Without looking at the code (PR is massive, I can't find it), how does the assertion work for these cases? Because I don't see parenthesis here

ematipico:
If the diverging is intentional and we won't go back to that, could you please add a task/reminder for ourselves to update the website with this decision?  We have to update the section where we track the difference section

crates/rome_js_formatter/src/js/module/import_meta.rs
```diff
@@ -27,4 +28,30 @@ impl FormatNodeRule<ImportMeta> for FormatImportMeta {
...
         ]
     }
+
+    fn needs_parentheses(&self, item: &ImportMeta) -> bool {
+        item.needs_parentheses()
+    }
+}
+
+impl NeedsParentheses for ImportMeta {
+    fn needs_parentheses(&self) -> bool {
```
ematipico, line 38: Sometime we line these functions are inlined, sometimes *always* inlined, and sometimes not. What's the logic behind that?

crates/rome_js_formatter/src/parentheses.rs
MichaReiser: It eases the implementation. Having it in `rome_formatter` prevents us from implementing it on nodes directly because of the orphan rule. There's also not much to the trait, so that other languages can just copy paste it if they need a similar functionality.

crates/rome_js_formatter/src/parentheses.rs
MichaReiser: Because they can be parenthesized as well. It will only be needed in later PRs but I didn't see a reason to not implement it correctly from the beginning.

crates/rome_js_formatter/src/js/expressions/arrow_function_expression.rs
MichaReiser: You can look at the macro documentation in VS Code. 

It searches for the `JsArrowFunctionExpression` in the code and calls `needs_parentheses` on the node and either asserts that the value is `true` or `false`.

crates/rome_js_formatter/src/js/module/import_meta.rs
MichaReiser: I added `inline` for nodes that just return `false`. I added `inline(always)` to some `resolve()` implementations if they just return the node. I didn't add `inlines` if the methods are more complicated.

MichaReiser:
I plan to merge this PR by tomorrow except there are some major conceptual concerns around it.

ematipico (approved):
We could have a module comment in the `parenthesis` file, explaining the new concepts introduced, mainly:
- what “resolving” means, since it has a meaning based on the context and it’s a concept that is now widespread and we will use somewhere else
- what “needs parenthesis” means, mostly to not assume that people already knows

PR #3058, opened 2022-08-13
refactor(rome_js_parser): refactor regex flag

IWANABETHATGUY:
## Summary
1. Refactor `lexer.rs` lexing regular expression, using a `bitflags` to represent regex flag instead of 7 boolean variables.
Reducing memory usage from 7 bytes to 1 byte.

## Test Plan
1. All CI should pass.

### DIFF
--- a/crates/rome_js_parser/src/lexer/mod.rs
+++ b/crates/rome_js_parser/src/lexer/mod.rs
@@ -1534,7 +1534,7 @@ impl<'src> Lexer<'src> {
     #[allow(clippy::many_single_char_names)]
     fn read_regex(&mut self) -> JsSyntaxKind {
         bitflags! {
-            struct RegexFlag: u32 {
+            struct RegexFlag: u8 {
                 const G = 1 << 0;
                 const I = 1 << 1;
                 const M = 1 << 2;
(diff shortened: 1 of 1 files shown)


crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -1555,53 +1565,59 @@ impl<'src> Lexer<'src> {
...
-                                    s = true;
                                 }
                                 b'u' => {
-                                    if u {
+                                    if flag.contains(RegexFlag::U) {
                                         self.diagnostics.push(self.flag_err('u'));
+                                    } else {
+                                        flag |= RegexFlag::U;
                                     }
-                                    u = true;
```
IWANABETHATGUY, line 1592: This statement will be always executed before, using an else branch to avoid that.

IWANABETHATGUY:
!bench_parser

crates/rome_js_parser/src/lexer/mod.rs
```diff
@@ -1530,10 +1530,20 @@ impl<'src> Lexer<'src> {
             "this flag was already used",
         )
     }
-
     #[inline]
     #[allow(clippy::many_single_char_names)]
     fn read_regex(&mut self) -> JsSyntaxKind {
+        bitflags! {
+            struct RegexFlag: u32 {
```
MichaReiser, line 1537: You can reduce this to be a u8 (you only have 7 flags)

crates/rome_js_parser/src/lexer/mod.rs
MichaReiser: Adding the same flag has no effect, which is why I would probably prefer to keep the assignment outside of the else branch.

IWANABETHATGUY:
!bench_parser

MichaReiser (approved):

PR #3061, opened 2022-08-15
feat(rome_js_formatter): Member assignment formatting

MichaReiser:
This PR improves the formatting of static and computed assignment expressions that Prettier formats the same as static/computed member expressions. 

This is done by introducing new `JsAnyStaticMemberLike` and `JsAnyComputedMemberLike` union types that implement the common formatting logic. 

This PR also fixes the formatting of the special case `a[5]` to match prettier's formatting.

## Test Plan

I manually compared Rome's and Prettier's output before updating the snapshots.

**File Based Average Prettier Similarity**: 79.75%   ->79.91% 
**Line Based Average Prettier Similarity**: 76.52%   -> 76.58%


ematipico:
Could you add the updated metrics to the description please?

MichaReiser:
> Could you add the updated metrics to the description please?

Done

ematipico (approved):
Same as the other PR, could we update/add some test case where mark where we expect a certain layout? Other than that, let's 🚢 it!

PR #3062, opened 2022-08-15
feat(rome_formatter): Dedent to Root IR

MichaReiser:
## Summary

This PR extends the `dedent` IR element to support dedenting by one level or to the start of the line.

I intentionally called the mode `root` because it may become necessary in the future that it's possible to mark a root which is different to the start of the line.

This IR is necessary to correctly format `TemplateLiteral`s where inserting any indent characters that weren't present of the template in the source changes the `template`s content at runtime.

## Test Plan

I added a doc test demonstrating how the IR can be used and verifies that the printer correctly resets the indent level.

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -709,12 +709,14 @@ where
 {
     Dedent {
         content: Argument::new(content),
+        mode: DedentMode::OneLevel,
     }
 }
 
 #[derive(Copy, Clone)]
 pub struct Dedent<'a, Context> {
     content: Argument<'a, Context>,
+    mode: DedentMode,
 }
 
 impl<Context> Format<Context> for Dedent<'_, Context> {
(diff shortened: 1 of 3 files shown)


crates/rome_formatter/src/format_element.rs
```diff
@@ -334,6 +338,15 @@ impl PrintMode {
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum DedentMode {
+    /// Reduces the indent by one level
+    OneLevel,
```
ematipico, line 344: The name implies that there are more levels. `Reduced` maybe?

crates/rome_formatter/src/builders.rs
```diff
@@ -738,6 +743,56 @@ impl<Context> std::fmt::Debug for Dedent<'_, Context> {
...
+///     "root\n\tindent level 1\n\t\tindent level 2\n\t\t  two space align\nstarts at the beginning of the line\n\t\tend indent level 2",
+///     block.print().as_code()
+/// );
+/// ```
+///
+/// ## Prettier
+///
+/// This resembles the behaviour of Prettier's `align(Number.NEGATIVE_INFINITY, content)` IR element.
+#[inline]
+pub fn dedent_root<Content, Context>(content: &Content) -> Dedent<Context>
```
ematipico, line 786: nit: `dedent_to_root`

ematipico (approved):

crates/rome_formatter/src/format_element.rs
MichaReiser: I rename dit to `Level. I prefer not to use `Dedent::Reduced` because it may be misunderstood that it reduces the dedent (meaning, it increases the indent).

PR #3065, opened 2022-08-16
feat(rome_js_formatter): implement option quote_properties

### LINKED ISSUE #2482: 📎 Implement option `preserve_quotes`

bhbs:
## Summary

This PR implements option ~`preserve_quotes`~ `quote_properties` following Prettier. ([docs](https://prettier.io/docs/en/options.html#quote-props))
I referred to `--quote-style` for the implementation of the option.

Close #2482 

## Test Plan

- Check new snapshots
  - crates/rome_js_formatter/tests/specs/js/module/string/properties_quotes.js.snap
  - crates/rome_js_formatter/tests/specs/ts/string/parameter_quotes.ts.snap


crates/rome_js_formatter/tests/specs/js/module/assignment/assignment.js.snap
```diff
@@ -1,6 +1,5 @@
 ---
 source: crates/rome_js_formatter/tests/spec_test.rs
-assertion_line: 259
```
bhbs, line 3: `assertion_line` is no longer retained in snapshots after cargo-insta v1.13.0
https://github.com/mitsuhiko/insta/issues/215

bhbs, line 3: `assertion_line` is no longer retained in snapshots after cargo-insta v1.13.0
https://github.com/mitsuhiko/insta/issues/215

bhbs:
After this PR is merged, I’d like to implement this option to VSCode Extension and playground!

ematipico:
> After this PR is merged, I’d like to implement this option to VSCode Extension and playground!

The VSCode extension now always reads the configuration from `rome.json`. We removed all the beta options. Only the playground needs it

crates/rome_js_formatter/src/context.rs
```diff
@@ -21,6 +21,9 @@ pub struct JsFormatContext {
     /// The style for quotes. Defaults to double.
     quote_style: QuoteStyle,
 
+    /// Respect the input use of quotes in object properties.
+    preserve_quotes: bool,
```
MichaReiser, line 25: I would prefer an enum over a bool so that it becomes possible to support all variants that prettier supports without a breaking change:

* preserve
* as-needed (what Rome does today. Default)
* consistent (out of scope)

I would also rename the setting to `quote-properties` to closer match prettier's naming

MichaReiser (commented):
Overall looking good to me. My main suggestion is to change the option to an enum so that it can be extended in the future

crates/rome_js_formatter/src/context.rs
bhbs: Thanks for the review. I will change it!

crates/rome_service/src/configuration/javascript.rs
```diff
@@ -83,6 +83,7 @@ pub struct JavascriptFormatter {
     /// The style for quotes. Defaults to double.
     #[serde(with = "PlainQuoteStyle")]
     pub quote_style: QuoteStyle,
+    pub preserve_quotes: bool,
```
ematipico, line 86: Could you add some documentation, please?

website/src/_includes/docs/project-config.md
```diff
@@ -197,6 +197,12 @@ The type of quote used when representing string literals. It can be `single` or
 
 > Default: `double`
 
+#### `javascript.formatter.preserveQuotes`
+
+Respect the input use of quotes in object properties.
+
+> Default: `false`
+
```
ematipico, line 205: Same here, this change should be reverted.

ematipico (changes requested):

bhbs:
@MichaReiser @ematipico 
Thank you for reviewing!
I addressed your feedback👍

crates/rome_service/src/configuration/javascript.rs
```diff
@@ -98,3 +101,17 @@ impl Default for PlainQuoteStyle {
...
+pub enum PlainQuoteProperties {
+    AsNeeded,
+    Preserve,
+}
+
+impl Default for PlainQuoteProperties {
+    fn default() -> Self {
+        Self::AsNeeded
+    }
+}
```
ematipico, line 117:
```suggestion
#[derive(Deserialize, Default, Serialize, Debug, Eq, PartialEq)]
#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
#[serde(rename_all = "camelCase", remote = "QuoteProperties")]
pub enum PlainQuoteProperties {
    #[default]
    AsNeeded,
    Preserve,
}
```

We can now use macros to assign defaults to enum!

ematipico (commented):
Could you add a new test inside `rome_cli/tests/` where we test an error when passing a wrong value for `--quote-properties`? The test should go inside `mod formatter`. And we can have a snapshot of it, so we can actually see the error message

crates/rome_service/src/configuration/javascript.rs
bhbs: Cool 🦀

MichaReiser: It may be better to call `QuoteStyle::default` here to make sure these two are always in sync. Not sure if there's a `From` implementation that easily allows you to do so.

bhbs:
@ematipico 
Did you intend like this? a6a89ab

ematipico:
> @ematipico 
> Did you intend like this? a6a89ab

Yes, that's it :)

It's okay if it's empty, I plan to change it soon. At least it's ready and there

crates/rome_service/src/configuration/javascript.rs
bhbs: Sorry, I'm not sure, please tell me how to do it

MichaReiser (approved):

ematipico (approved):
This is a great contribution! Thank you very much!

PR #3066, opened 2022-08-16
feat(rome_cli): add `--json` argument to format command

### LINKED ISSUE #3049: implement a `--json` option to the CLI
To allow communication via JSON between runtime and rust workspace

ematipico:
## Summary

Closes #3049 

**The status of the PR is not the final design**. 

There are still things missing that I would like implement in subsequent PRs:
- report errors on `stderr`
- review how messages and reports can live together
- expand `--json` to the `check` command

These were intentionally left out because I had spent enough time to experiment with other crates. 

What I would like to review is: 
- the shape of the information reported after a traversal of the `format` command on a file/folder
- if the usage of three threads makes sense and not flawed

The usage of three threads was necessary to because we need to collect information from two sources:
- the thread of the traversal
- the thread to process the messages 

And we need a third thread to receive these messages.

### Examples

Given the file `example.js`

```js
const a = "hey"

statement(

);

```

If we run the following command `rome format example.js --json`, we get the following JSON:

```json
{
  "formatter": {
    "summary": {
      "duration": { "secs": 0, "nanos": 3975001 },
      "filesCompared": 1,
      "filesWritten": null
    },
    "details": {}
  },
  "errors": {
    "example.js": {
      "diff": {
        "severity": "Error",
        "before": "const a = \"hey\"\n\nstatement(\n\n);\n",
        "after": "const a = \"hey\";\n\nstatement();\n"
      }
    }
  }
}
```

If we run the command `rome format test.js --write --json`, we get the following JSON:

```json
{
  "formatter": {
    "summary": {
      "duration": { "secs": 0, "nanos": 6510659 },
      "filesCompared": null,
      "filesWritten": 1
    },
    "details": {
      "example.js": {
        "newContent": "const a = \"hey\";\n\nstatement();\n"
      }
    }
  },
  "errors": {}
}

```

The paths will be relative to the path where the command was run.

## Test Plan

Added a test case to cover a simple case

### DIFF
--- a/crates/rome_cli/src/execute.rs
+++ b/crates/rome_cli/src/execute.rs
@@ -7,8 +7,18 @@ use rome_service::workspace::{
 };
 use std::path::PathBuf;
 
+/// Useful information during the traversal of files and virtual content
 #[derive(Clone)]
-pub(crate) enum ExecutionMode {
+pub(crate) struct Execution {
+    /// How the information should be collected and reported
+    report_mode: ReportMode,
+
+    /// The modality of execution of the traversal
+    traversal_mode: TraversalMode,
+}
+
+#[derive(Clone)]
+pub(crate) enum TraversalMode {
     /// This mode is enabled when running the command `rome check`
     Check {
         /// The maximum number of diagnostics that can be printed in console
@@ -35,19 +45,53 @@ pub(crate) enum ExecutionMode {
     },
 }
 
-impl ExecutionMode {
+/// Tells to the execution of the traversal how the information should be reported
+#[derive(Clone, Default)]
+pub(crate) enum ReportMode {
+    /// Reports information straight to the console, it's the default mode
+    #[default]
+    Terminal,
+    /// Reports information in JSON format
+    Json,
+}
+
+impl Execution {
+    pub(crate) fn new(mode: TraversalMode) -> Self {
+        Self {
+            report_mode: ReportMode::default(),
+            traversal_mode: mode,
+        }
+    }
+
+    /// Creates an instance of [Execution] by passing [traversal mode](TraversalMode) and [report mode](ReportMode)
+    pub(crate) fn with_report(traversal_mode: TraversalMode, report_mode: ReportMode) -> Self {
+        Self {
+            traversal_mode,
+            report_mode,
+        }
+    }
+
+    /// Tells if the reporting is happening straight to terminal
+    pub(crate) fn should_report_to_terminal(&self) -> bool {
+        matches!(self.report_mode, ReportMode::Terminal)
+    }
+
+    pub(crate) fn traversal_mode(&self) -> &TraversalMode {
+        &self.traversal_mode
+    }
+
     pub(crate) fn get_max_diagnostics(&self) -> Option<u16> {
-        match self {
-            ExecutionMode::Check {
+        match self.traversal_mode {
+            TraversalMode::Check {
                 max_diagnostics, ..
-            } => Some(*max_diagnostics),
+            } => Some(max_diagnostics),
             _ => None,
         }
     }
 
     /// `true` only when running the traversal in [TraversalMode::Check] and `should_fix` is `true`
     pub(crate) fn as_fix_file_mode(&self) -> Option<&FixFileMode> {
-        if let ExecutionMode::Check { fix_file_mode, .. } = self {
+        if let TraversalMode::Check { fix_file_mode, .. } = &self.traversal_mode {
             fix_file_mode.as_ref()
         } else {
             None
@@ -55,20 +99,20 @@ impl ExecutionMode {
     }
 
     pub(crate) fn is_ci(&self) -> bool {
-        matches!(self, ExecutionMode::CI { .. })
+        matches!(self.traversal_mode, TraversalMode::CI { .. })
     }
 
     pub(crate) fn is_check(&self) -> bool {
-        matches!(self, ExecutionMode::Check { .. })
+        matches!(self.traversal_mode, TraversalMode::Check { .. })
     }
 
     pub(crate) fn is_format(&self) -> bool {
-        matches!(self, ExecutionMode::Format { .. })
+        matches!(self.traversal_mode, TraversalMode::Format { .. })
     }
 
     pub(crate) fn as_stdin_file(&self) -> Option<&(PathBuf, String)> {
-        match self {
-            ExecutionMode::Format { stdin, .. } => stdin.as_ref(),
+        match &self.traversal_mode {
+            TraversalMode::Format { stdin, .. } => stdin.as_ref(),
             _ => None,
         }
     }
@@ -76,10 +120,7 @@ impl ExecutionMode {
 
 /// Based on the [mode](ExecutionMode), the function might launch a traversal of the file system
 /// or handles the stdin file.
-pub(crate) fn execute_mode(
-    mode: ExecutionMode,
-    mut session: CliSession,
-) -> Result<(), Termination> {
+pub(crate) fn execute_mode(mode: Execution, mut session: CliSession) -> Result<(), Termination> {
     // don't do any traversal if there's some content coming from stdin
     if let Some((path, content)) = mode.as_stdin_file() {
         let workspace = &*session.app.workspace;
--- a/crates/rome_cli/src/reports/formatter.rs
+++ b/crates/rome_cli/src/reports/formatter.rs
@@ -24,18 +24,18 @@ impl FormatterReport {
 #[derive(Debug, Default, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct FormatterReportSummary {
-    /// how many files were compared
     files_compared: Option<usize>,
-    /// how many files were written
     files_written: Option<usize>,
 }
 
 impl FormatterReportSummary {
+    /// how many files were compared
     pub(crate) fn set_files_compared(&mut self, files_compared: usize) {
         self.files_compared = Some(files_compared)
     }
 
-    pub(crate) fn self_files_written(&mut self, files_written: usize) {
+    /// how many files were written
+    pub(crate) fn set_files_written(&mut self, files_written: usize) {
         self.files_written = Some(files_written)
     }
 }
--- a/crates/rome_cli/src/reports/formatter.rs
+++ b/crates/rome_cli/src/reports/formatter.rs
@@ -3,34 +3,34 @@ use std::collections::HashMap;
 
 #[derive(Debug, Default, Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct FormatterStats {
+pub struct FormatterReport {
     /// Useful information of the execution
-    summary: Option<FormatterStatSummary>,
+    summary: Option<FormatterReportSummary>,
 
     /// The key is the path of the file
-    details: HashMap<String, FormatterStatDetail>,
+    files: HashMap<String, FormatterReportFileDetail>,
 }
 
-impl FormatterStats {
-    pub(crate) fn insert_stat_detail(&mut self, stat: String, detail: FormatterStatDetail) {
-        self.details.insert(stat, detail);
+impl FormatterReport {
+    pub(crate) fn insert_file_content(&mut self, path: String, detail: FormatterReportFileDetail) {
+        self.files.insert(path, detail);
     }
 
-    pub(crate) fn set_summary(&mut self, summary: FormatterStatSummary) {
+    pub(crate) fn set_summary(&mut self, summary: FormatterReportSummary) {
         self.summary = Some(summary);
     }
 }
 
 #[derive(Debug, Default, Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct FormatterStatSummary {
+pub struct FormatterReportSummary {
     /// how many files were compared
     files_compared: Option<usize>,
     /// how many files were written
     files_written: Option<usize>,
 }
 
-impl FormatterStatSummary {
+impl FormatterReportSummary {
     pub(crate) fn set_files_compared(&mut self, files_compared: usize) {
         self.files_compared = Some(files_compared)
     }
@@ -42,7 +42,7 @@ impl FormatterStatSummary {
 
 #[derive(Debug, Default, Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct FormatterStatDetail {
+pub struct FormatterReportFileDetail {
     /// The new content emitted by the formatter
     pub formatted_content: Option<String>,
 }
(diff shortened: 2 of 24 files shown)


MichaReiser:
Would you mind adding a short outline to the PR summary how the architecture changed. You mention three threads. What's the responsibility of each thread, why are multiple threads necessary? Are there specific reasons why you doubt that three threads are a good approach?

ematipico:
> Would you mind adding a short outline to the PR summary how the architecture changed. You mention three threads. What's the responsibility of each thread, why are multiple threads necessary? Are there specific reasons why you doubt that three threads are a good approach?
> 
> Edit: Can you add an example of a formatter and check JSON output to the test section?

Done.

MichaReiser:
Thanks for updating the description. 

Some feedback on the JSON format

* what use cases to you see for the `duration`? In my view, this is something that scripts can easily measure for themselves if they're interested in the duration
* Is the reason that `errors` are reported outside of the `formatter` that these are diagnostics? 
* Do you think it's necessary for us to report the `content` of the written files? A script could read the file if it is interested in the updated content.

crates/rome_cli/src/traversal.rs
```diff
@@ -63,70 +65,112 @@ pub(crate) fn traverse(mode: ExecutionMode, mut session: CliSession) -> Result<(
     let workspace = &*session.app.workspace;
     let console = &mut *session.app.console;
 
-    let (has_errors, duration) = join(
-        || print_messages_to_console(mode.clone(), console, recv_files, recv_msgs),
-        || {
+    let mut has_errors = None;
+    let mut duration = None;
+    let mut stats = None;
+    let send_stats_from_messages = send_stats_from_traversal.clone();
```
MichaReiser, line 71: Can you explain why printing the messages to the console from the current thread is no longer an option and it is instead necessary to spawn a separate thread?

leops:
As an alternative to having a dedicated thread for receiving the stats, the `crossbeam` library we're using for MPSC channels has a `select!` macro that would allow the existing console thread to wait on both the messages and stats channels at the same time, and wake up the thread whenever one of the channels receives a message

crates/rome_cli/src/traversal.rs
ematipico: The printing of messages **was already happening on separate thread**. From this thread we now send the stats to the `stats_thread`, because these messages contain diagnostics, which we want to capture.

ematipico:
> Thanks for updating the description.
> 
> Some feedback on the JSON format
> 
>     * what use cases to you see for the `duration`? In my view, this is something that scripts can easily measure for themselves if they're interested in the duration

This is true. But I thought, since we have the information at hand (and it's really detailed), why not expose it? Also, considering that our APIs can be used by other developers to make other plugins, they can use it for internal measurements.

> 
>     * Is the reason that `errors` are reported outside of the `formatter` that these are diagnostics?

Yes, but I am not sure these errors should be sent together with this JSON on `stdin` or we should `stderr` instead. That's why I left it out. 

Also, I am not sure if `errors` should stay at top level or at "feature" level. If we run the `check` command, we might have diagnostics related to linter, formatter, etc.

> 
>     * Do you think it's necessary for us to report the `content` of the written files? A script could read the file if it is interested in the updated content.

I think it's a valuable information, since we already have it. If the script wants the information, it would require to make another I/O operation. If we scale this situation on folders, there would be lot of I/O operations... It's a trade-off. The JSON would be slim but scripts that require that information would be penalized. What do you think? Should we remove it?

MichaReiser:
> I think it's a valuable information, since we already have it. If the script wants the information, it would require to make another I/O operation. If we scale this situation on folders, there would be lot of I/O operations... It's a trade-off. The JSON would be slim but scripts that require that information would be penalized. What do you think? Should we remove it?

That depends on the use cases that we want to support. Overall, I'm leaning towards removing any information that we don't have an explicit use case for because it will be difficult to remove fields in the future. 

I would currently focus on only adding fields/information for which we have a direct use in the Node JS API, because that's the ultimate goal we're pursuing.  So the question is, what's the result of `formatFiles`? 

> Should `errors` be called `diagnostics`?

crates/rome_cli/src/execute.rs
```diff
@@ -35,51 +45,82 @@ pub(crate) enum ExecutionMode {
...
+    pub(crate) fn with_report(traversal_mode: TraversalMode, report_mode: ReportMode) -> Self {
+        Self {
+            traversal_mode,
+            report_mode,
+        }
+    }
+
+    /// Tells if the reporting is happening straight to terminal
+    pub(crate) fn should_report_to_terminal(&self) -> bool {
+        matches!(self.report_mode, ReportMode::Terminal)
```
MichaReiser, line 76: Nit: `is_terminal()`  to keep in line with `is_ci`, `is_check`

crates/rome_cli/src/stats/formatter.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::collections::HashMap;
+
+#[derive(Debug, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FormatterStats {
+    /// Useful information of the execution
+    summary: Option<FormatterStatSummary>,
+
+    /// The key is the path of the file
+    details: HashMap<String, FormatterStatDetail>,
```
MichaReiser, line 11: Is it important that the details are stored in the same order as they've been formatted, if so, use an `IndexMap`|

Should we omit the contents if the output is empty?

Nit: Rename to `contents` or `files` as it's storying the formatted content by file.

crates/rome_cli/src/stats/formatter.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    pub(crate) fn self_files_written(&mut self, files_written: usize) {
+        self.files_written = Some(files_written)
+    }
+}
+
+#[derive(Debug, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FormatterStatDetail {
```
MichaReiser, line 45: Nit: `FormatterFileDetail`

crates/rome_cli/src/stats/formatter.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct FormatterStats {
+    /// Useful information of the execution
+    summary: Option<FormatterStatSummary>,
+
+    /// The key is the path of the file
+    details: HashMap<String, FormatterStatDetail>,
+}
+
+impl FormatterStats {
+    pub(crate) fn insert_stat_detail(&mut self, stat: String, detail: FormatterStatDetail) {
```
MichaReiser, line 15:
```suggestion
    pub(crate) fn insert_file_content(&mut self, stat: String, detail: FormatterStatDetail) {
```

Can we make `stat` a `Path`? 

Can you document the meaning of `stat` and `detail`. I find it difficult what `stat` is standing for (`path`?)

crates/rome_cli/src/stats/formatter.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    pub(crate) fn self_files_written(&mut self, files_written: usize) {
+        self.files_written = Some(files_written)
+    }
+}
+
+#[derive(Debug, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FormatterStatDetail {
+    /// The new content emitted by the formatter
```
MichaReiser, line 46: Nit: Just `content`

crates/rome_cli/src/stats/formatter.rs
```diff
@@ -0,0 +1,48 @@
+use serde::Serialize;
+use std::collections::HashMap;
+
+#[derive(Debug, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FormatterStats {
```
MichaReiser, line 6: This Pr introduce the new concept`Stats`.  I assume this stands for statistics. My understanding of statistics is that they are number focused but this object also contains many data fields. How about `FormatterOutput`?

Nit: Avoid abbreviations -> `FormatterStatistics`

crates/rome_cli/src/stats/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use serde::Serialize;
+use std::collections::HashMap;
+
+#[derive(Debug, Default, Serialize)]
+pub struct Stats {
+    /// Information relative to the formatter
+    formatter: FormatterStats,
+
+    /// Diagnostics tracked during a generic traversal
+    diagnostics: HashMap<String, StatErrorKind>,
```
MichaReiser, line 17: What's the key?

crates/rome_cli/src/stats/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum StatErrorKind {
+    Diagnostic(StatDiagnostic),
+    Diff(StatDiff),
+}
```
MichaReiser, line 25: Rename to `Diagnostic` as it is used in the `diagnostics` field? What's the meaning of `Stat`?

crates/rome_cli/src/execute.rs
ematipico: I want to keep them different on purpose, because `is_ci` is related to the traversal, while the terminal is related to the reporting

crates/rome_cli/src/stats/formatter.rs
ematipico: > Is it important that the details are stored in the same order as they've been formatted, if so, use an `IndexMap`|
> 

The order is not important, I think.

crates/rome_cli/src/stats/formatter.rs
ematipico: We can't use `Path` because there are situations where we have a `String` and not a `Path`

crates/rome_cli/src/stats/formatter.rs
ematipico: My first implementation with `statistics`, but I learnt that `stats` is actually a synonym of `statistics`, so I suppose we're fine with using `stats`.

> is that they are number focused 

What does this mean?

MichaReiser: My understanding of `statistics` is something along these lines

> [Statistics](https://en.wikipedia.org/wiki/Statistics) is a mathematical science pertaining to the collection, analysis, interpretation, and presentation of data.

That's why my first interpretation was that the `*Stats` structs are used to store the metrics collected from running `rome --show-metrics` but this isn't the case. The reason for this intuition is that I think of numbers if someone speaks about statistics. But this isn't really what we are storing in this struct. 

That's why I recommend renaming the `Stat` types to `Output` or another concept that more closely reflects the data we're storing.

ematipico: I used the concept of `Report`

ematipico:
@leops I would like to defer your suggestion in another PR, mostly because `select` won't work with the current architecture. The current architecture has two receivers for the reports, one inside the traversal and one inside the processor of the messagges. Using `Select` from `crossbeam` won't work **with the current architecture**, if the receiver of the reports is woken up first, we will cause a deadlock because the sender passed to the processor of the messages never goes our of scope (because it's never woken up) by `Select`. I plan to refactor the current architecture, but this work is out of scope for this PR and I would like to tackle it a later PR.

@MichaReiser Are your concerns been covered? Are there any other things pending? I would like to merge this PR and continue to apply new changes from here.

crates/rome_cli/src/reports/formatter.rs
```diff
@@ -0,0 +1,48 @@
+use serde::Serialize;
+use std::collections::HashMap;
+
+#[derive(Debug, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FormatterReport {
```
MichaReiser, line 6: I like the name `Report`. It makes it clear that it relates to `ReportMode`.

crates/rome_cli/src/reports/formatter.rs
```diff
@@ -0,0 +1,48 @@
...
+        self.summary = Some(summary);
+    }
+}
+
+#[derive(Debug, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FormatterReportSummary {
+    /// how many files were compared
+    files_compared: Option<usize>,
+    /// how many files were written
```
MichaReiser, line 29: Nit: I would recommend moving these comments to the `set` methods that are public so that they show up in the documentation. 

The same with The comments on `FormatterReport`

crates/rome_cli/src/reports/formatter.rs
```diff
@@ -0,0 +1,48 @@
...
+    files_written: Option<usize>,
+}
+
+impl FormatterReportSummary {
+    pub(crate) fn set_files_compared(&mut self, files_compared: usize) {
+        self.files_compared = Some(files_compared)
+    }
+
+    pub(crate) fn self_files_written(&mut self, files_written: usize) {
```
MichaReiser, line 38:
```suggestion
    pub(crate) fn set_files_written(&mut self, files_written: usize) {
```

crates/rome_cli/src/reports/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::reports::formatter::{FormatterReportFileDetail, FormatterReportSummary};
+use formatter::FormatterReport;
+use rome_console::codespan::Severity;
+use rome_service::RomeError;
+use serde::Serialize;
+use std::collections::HashMap;
+
+#[derive(Debug, Default, Serialize)]
+pub struct Report {
+    /// Information relative to the formatter
```
MichaReiser, line 12:
```suggestion
    /// Information related to the formatter
```

related?

crates/rome_cli/src/reports/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    Diagnostic(ReportDiagnostic),
+    Diff(ReportDiff),
+}
+
+/// Information computed from a [diagnostic][rome_diagnostics::Diagnostic]
+#[derive(Debug, Serialize)]
+pub struct ReportDiagnostic {
+    pub severity: Severity,
+    pub code: Option<String>,
+    pub title: String,
```
MichaReiser, line 33: Would you mind adding some documentation to the public fields of the structs in this file?

MichaReiser (approved):
Do you plan to create an issue for the follow up on `select` or is it something we don't plan to pursue at the moment?

ematipico:
> Do you plan to create an issue for the follow up on `select` or is it something we don't plan to pursue at the moment?

I plan to revisit this. This is was more of an optimization and I think it's not needed for our case.

Issue #3054, opened 2022-08-12
☂️ Tracking issues or potential improvement we could make after upgrades to the Rust toolchain


IWANABETHATGUY:
# Task
## Issues after upgrade to 1.63.0
- [x] rustdoc::broken_intra_doc_links can't detect the `macro_export` 
- [ ] let_and_return rule will cause a borrow checker issue.

## Potential improvements we could make
- [x] ~From Rust 1.64: Cargo now supports inheriting some package attributes from the workspace (see https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table for the list of supported attributes), we could move some of the attributes currently duplicated between all crates (like `authors`, `repository`, `licence` or `edition`) to the root `Cargo.toml` file~
- [x] From Rust 1.63: With the standard library Mutex and RwLock being const-constructible it's now possible to use these in the DiffReport of the Prettier test suite
- [x] From Rust 1.62 it's now possible to derive the Default trait on enums, we could track down and remove the manual implementations for these
- [x] ~From Rust 1.61 the Termination trait and ExitCode type are now stable, meaning we could implement stable exit codes for the CLI (this could be done in the incoming refactor of diagnostics)~

IWANABETHATGUY:
rustdoc::broken_intra_doc_links : https://github.com/rome/tools/pull/3053#discussion_r944105018
let_and_return : https://github.com/rome/tools/pull/3053#discussion_r944102789

ematipico:
Hi @IWANABETHATGUY , could you please update the description and explain what's this umbrella for?

leops:
> Hi @IWANABETHATGUY , could you please update the description and explain what's this umbrella for?

I think we may "refactor" this umbrella to track both issues that are introduced in upgrades to the Rust toolchain (mostly Clippy false positives), and possible improvements to the existing codebase made possible by newly stabilized APIs (I came up with a short list of these here: https://github.com/rome/tools/pull/3053#issuecomment-1212812230)

IWANABETHATGUY:
> 

Done

ematipico:
Someone already filed an issue for the false positive of `let_and_return`: `https://github.com/rust-lang/rust-clippy/issues/8114

PR #3069, opened 2022-08-16
refactor: Using default derive macro instead of manual implementation for Enum

IWANABETHATGUY:
## Summary
1. Part of #3054 
2. From Rust 1.62 it's now possible to derive the Default trait on enums, we could track down and remove the manual implementations for these
3. See https://rust-lang.github.io/rfcs/3107-derive-default-enum.html

## Test Plan
1. The CI should pass.

### DIFF
--- a/crates/rome_js_syntax/src/source_type.rs
+++ b/crates/rome_js_syntax/src/source_type.rs
@@ -7,7 +7,7 @@ use std::path::Path;
 /// Defaults to the latest stable ECMAScript standard.
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Default)]
 pub enum LanguageVersion {
-	#[default]
+    #[default]
     ES2022,
 
     /// The next, not yet finalized ECMAScript version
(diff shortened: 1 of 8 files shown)


crates/rome_js_syntax/src/source_type.rs
```diff
@@ -5,8 +5,9 @@ use std::path::Path;
 /// The versions are ordered in increasing order; The newest version comes last.
 ///
 /// Defaults to the latest stable ECMAScript standard.
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Default)]
 pub enum LanguageVersion {
+    #[default]
```
xunilrj, line 10: For this one, I can think the old implementation makes more sense.

MichaReiser (approved):

crates/rome_js_syntax/src/source_type.rs
IWANABETHATGUY: Done

PR #3070, opened 2022-08-16
feat(node-api): create a node.js wrapper library for the JSON-RPC workspace protocol

leops:
## Summary

This PR creates a new `@rometools/backend-jsonrpc` library that implements the Workspace API by wrapping the JSON-RPC protocol exposed by the daemon server. The library is split into a few modules that may be imported independently:

- `command` resolves a file system path to the Rome binary
- `socket` ensures the daemon process is spawned and opens a node.js `Socket` connection to the RPC channel
- `transport` wraps an open socket and implements the JSON-RPC protocol as untyped method calls and responses
- `workspace` is automatically generated from the JSON schema of the Workspace API, and wraps a `Transport` instance with an object implementing typed methods calls to the remote workspace

At the top level of the library a single `createWorkspace` method calls each of these in sequence, wrapping each layer with the next one and returning a `Workspace` object ready to use

## Test Plan

I added a small test script that creates creates a workspace instance (using a local build of the binary from the Cargo target directory), executes a few method calls then shuts down


npm/backend-jsonrpc/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Transport } from "./transport";
+import { createWorkspace as wrapTransport, type Workspace } from "./workspace";
+
+/**
+ * Create an instance of the Workspace client connected to a remote daemon
+ * instance through the JSON-RPC protocol
+ * 
+ * @returns A Workspace client, or null if the underlying platform is not supported
+ */
+export function createWorkspace(): Promise<Workspace> | null {
```
ematipico, line 12: Not sure if `Promise<Workspace> | null` is portable, because sometimes this is thenable, sometimes it's not, and a consumer wouldn't know how to handle this. Maybe we should make this function `async` and always return a `Promise`

npm/backend-jsonrpc/src/socket.ts
```diff
@@ -0,0 +1,37 @@
+import { spawn } from "node:child_process";
+import { connect, type Socket } from "node:net";
```
ematipico, line 2: I am not sure if using the `node:` protocol is good for us. We still support node version `14` and the prefix was introduced in some later release? I don't remember which one though. 

If we decided to go this route, we should pin the correct version inside `engine`

npm/backend-jsonrpc/src/transport.ts
```diff
@@ -0,0 +1,258 @@
...
+}
+
+interface PendingRequest {
+	resolve(result: any): void;
+	reject(error: any): void;
+}
+
+export class Transport {
+	private nextRequestId = 0;
+	private pendingRequests: Map<number, PendingRequest> = new Map();
```
ematipico, line 94: Can we document this two private fields?

npm/backend-jsonrpc/src/transport.ts
```diff
@@ -0,0 +1,258 @@
...
+
+	destroy() {
+		this.socket.destroy();
+	}
+
+	private sendMessage(message: JsonRpcMessage) {
+		const body = Buffer.from(JSON.stringify(message));
+		const headers = Buffer.from(
+			`Content-Length: ${body.length}\r\n` +
+				`Content-Type: application/vscode-jsonrpc;charset=utf-8\r\n` +
```
ematipico, line 131: Can we have `application/vscode-jsonrpc` inside a constant? It seems important and it's reused in different places

npm/backend-jsonrpc/src/workspace.ts
```diff
@@ -0,0 +1,295 @@
+import type { Transport } from "./transport";
```
ematipico, line 1: Can we add a comment here saying that this file is autogenerated and should not manually modified?

npm/backend-jsonrpc/src/workspace.ts
```diff
@@ -0,0 +1,295 @@
...
+export interface RenameParams {
+	new_name: string;
+	path: RomePath;
+	symbol_at: number;
+}
+export interface RenameResult {
+	indels: Indel[];
+	range: TextRangeSchema;
+}
+export interface Workspace {
```
ematipico, line 224: Would it be possible to make all names in camel case instead of snake case? Not just here, but in the whole file.

npm/backend-jsonrpc/src/workspace.ts
ematipico, line 1: If this file is autogenerated, should we add it to the `.gitattributes` file?

ematipico (commented):
Is there any chance/way to make the `Transport` testable? I would like to test the error cases, to make sure it works as expected

npm/backend-jsonrpc/src/index.ts
MichaReiser: I would argue that returning `null` should probably be an error instead.

leops: I don't really want to make this an exception since those can't be explicitly typed in the signature of the function, while I intend for this specific "error" to be handled by the higher-level wrapper API by creating a Workspace instance from the `wasm-nodejs` package if the native platform is unsupported

npm/backend-jsonrpc/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@rometools/backend-jsonrpc",
+  "version": "0.8.0-next",
```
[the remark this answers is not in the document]
leops: This version number is the same as the main `rome` package and all the `cli-*` wrappers, but it gets updated automatically in the `generate-packages.mjs` script that runs during releases anyway so I could set it to `0.0.0` as a placeholder

npm/backend-jsonrpc/src/index.ts
MichaReiser: In that case I would probably go with `Promise<Workspace | null>`

npm/backend-jsonrpc/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	await transport.request("initialize", {
+		capabilities: {},
+		client_info: {
+			name: "@rometools/backend-jsonrpc",
+			version: "0.8.0-next",
+		},
+	});
+
+	return wrapTransport(transport);
```
MichaReiser, line 35: What happens if the transport fails to initialize a connection (or create a socket)? Does it throw any errors and if so which one?

npm/backend-jsonrpc/tsconfig.json
```diff
@@ -0,0 +1,103 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
+    // "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./.tsbuildinfo",              /* Specify the path to .tsbuildinfo incremental compilation file. */
```
MichaReiser, line 8: I would prefer if we could delete all the setting that we aren't using. It makes it hard to see what our explicit settings are.

website/playground/build.rs
```diff
@@ -387,16 +68,10 @@ fn write_protocol() -> (String, Vec<String>) {
     // ensure it looks good by running it through the formatter
     let formatted = format_node(JsFormatContext::default(), module.syntax()).unwrap();
```
MichaReiser, line 69: Nice!

MichaReiser (approved):
In my view it would be useful to add documentation to all public types/methods and to add a small readme to the project as well (which also indicates, that this API is considered internal and may change at any point in time).

npm/backend-jsonrpc/src/index.ts
leops: Creating the socket isn't really up to the transport layer (it expects to be provided with an existing `Socket` instance on construction), but the `createSocket` function will throw an exception is the socket cannot be opened (the socket emits an instance of `Error` through the `"error"` event), and the remote call to `initialize` may also throw an RPC error object if the remote server reports an error

MichaReiser: Would it make sense for us to introduce a `RomeBackendError` that is thrown in any situation where the backend runs into an error (and we use `RomeError` in higher level APIs?)

npm/backend-jsonrpc/tsconfig.json
ematipico: I'd prefer to keep them because we haven't figure out a correct build process, so if we need some specific property turned on, we have the benefit of the comments.

npm/backend-jsonrpc/src/index.ts
leops: It could be useful to introduce a custom error class wrapping the RPC error objects since those contain structured informations (they may include a serialized `RomeError` for instance), but I'm not sure we can provide much value by wrapping the I/O errors coming from node itself because of an issue with the connection

ematipico (approved):

PR #3071, opened 2022-08-16
feat(rome_js_semantic): semantic read events for self invoking functions

xunilrj:
## Summary

Expands https://github.com/rome/tools/issues/2979 with correctly not flagging self-invoking functions as unused.

I don't consider this implementation complete yet. But we need to think about how to deal with complex expressions in these cases.

## Test Plan

```
> cargo test -p rome_js_analyze -- no_unused_variables
```

### DIFF
--- a/crates/rome_js_semantic/src/events.rs
+++ b/crates/rome_js_semantic/src/events.rs
@@ -429,22 +429,16 @@ impl SemanticEventExtractor {
         let call = node.clone().cast::<JsCallExpression>()?;
         let callee = call.callee().ok()?;
 
-        match callee.syntax().kind() {
-            JsSyntaxKind::JS_PARENTHESIZED_EXPRESSION => {
-                let expr = callee.as_js_parenthesized_expression()?;
-                let range = expr.syntax().text_range();
-                match result_of(expr) {
-                    Some(JsAnyExpression::JsFunctionExpression(expr)) => {
-                        let id = expr.id()?;
-                        self.stash.push_back(SemanticEvent::Read {
-                            range,
-                            declared_at: id.syntax().text_range(),
-                        });
-                    }
-                    _ => {}
-                }
+        if let JsSyntaxKind::JS_PARENTHESIZED_EXPRESSION = callee.syntax().kind() {
+            let expr = callee.as_js_parenthesized_expression()?;
+            let range = expr.syntax().text_range();
+            if let Some(JsAnyExpression::JsFunctionExpression(expr)) = result_of(expr) {
+                let id = expr.id()?;
+                self.stash.push_back(SemanticEvent::Read {
+                    range,
+                    declared_at: id.syntax().text_range(),
+                });
             }
-            _ => {}
         }
 
         Some(())
--- a/crates/rome_js_semantic/src/events.rs
+++ b/crates/rome_js_semantic/src/events.rs
@@ -3,10 +3,11 @@
 use std::collections::{HashMap, VecDeque};
 
 use rome_js_syntax::{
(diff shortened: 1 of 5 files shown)


crates/rome_js_semantic/src/events.rs
```diff
@@ -197,6 +198,26 @@ struct Scope {
...
+    let mut expr = Some(JsAnyExpression::JsParenthesizedExpression(expr.clone()));
+    loop {
+        match expr {
+            Some(JsAnyExpression::JsParenthesizedExpression(e)) => {
+                expr = e.expression().ok();
+            }
+            Some(JsAnyExpression::JsSequenceExpression(e)) => {
+                expr = e.right().ok();
+            }
+            Some(JsAnyExpression::JsAssignmentExpression(e)) => {
```
MichaReiser, line 212: You want to add `TsAsExpression`, `TsTypeAssertionExpression`, `TsNonNullAssertionExpression`.

MichaReiser:
Have you consider implementing such methods on `JsAnyExpression` and all expression types? I can see this to be useful for many other analysis too 

```
("FOO") == "FOO"
```

xunilrj:
> Have you consider implementing such methods on `JsAnyExpression` and all expression types? I can see this to be useful for many other analysis too

Yes. But I don´t consider this implementation mature yet. I am still going to improve this a little bit inside the ```noUnusedVariables``` until I feel more confident with it.

For example, not sure what to do with the ternary operator yet.

```js
(true?function f(){}:function g(){})();
```

```f``` and ```g``` are used in this scenario. Currently the semantic model don´t have a way to model this.

crates/rome_js_semantic/src/events.rs
xunilrj: See the comment below. It is still missing a lot of cases. Will try to improve it in other PRs.

ematipico (approved):

leops (commented):
So I've been thinking about this, and in the end I'm not really sure this is actually a false positive. Conceptually the `noUnusedVariable` rule (and the idea of variable usage in the semantic model in general) is about the usage of names and symbols, not about their concrete value. So in the case of `(function f(){})()` the rule is actually correctly raising a diagnostic because the name `f` is never used in the entire document, what's incorrect is how the error is explained to the user. I think it would help to have a suggested fix that removes the `id` of the function expression, but that's not a safe change to apply since it has visible side-effects (it changes the `name` property of the resulting function, which might be used for debugging purpose like in `React.memo(function Component() {})` for instance)

MichaReiser:
> So I've been thinking about this, and in the end I'm not really sure this is actually a false positive. Conceptually the `noUnusedVariable` rule (and the idea of variable usage in the semantic model in general) is about the usage of names and symbols, not about their concrete value. So in the case of `(function f(){})()` the rule is actually correctly raising a diagnostic because the name `f` is never used in the entire document, what's incorrect is how the error is explained to the user. I think it would help to have a suggested fix that removes the `id` of the function expression, but that's not a safe change to apply since it has visible side-effects (it changes the `name` property of the resulting function, which might be used for debugging purpose like in `React.memo(function Component() {})` for instance)

This is an excellent point and, in my view, reason to remove `FunctionExpression` and `ClassExpression` from the `noUnusedVariable` rule because these expressions not only bind a name but also have a value. 

Unused expressions should instead be reported by a `noUnreachable` rule (which likely need to handle `FunctionExpression` and `ClassExpression`s with names differently.

ematipico:
We actually have an example in our codebase

https://github.com/rome/tools/blob/main/website/playground/src/MermaidGraph.tsx#L7-L20

PR #3072, opened 2022-08-17
feat(playground): Use full height for right side panels

MichaReiser:
This PR improves our playground to make use of the full height for the right hand side panels rather than 70% (which may result in scroll bars on small screens)

## Test Plan

Screenshot from 2022-08-17 10-07-17
Screenshot from 2022-08-17 10-07-27
Screenshot from 2022-08-17 10-07-35
Screenshot from 2022-08-17 10-07-45

### DIFF
--- a/website/playground/src/DesktopPlayground.tsx
+++ b/website/playground/src/DesktopPlayground.tsx
@@ -83,25 +83,29 @@ export default function DesktopPlayground({
 	}, []);
 
 	return (
-		<div className="divide-y divide-slate-300">
+		<div className="divide-y divide-slate-300 h-screen flex flex-col">
 			<h1 className="p-4 text-xl">Rome Playground</h1>
 			<SettingsMenu
 				settings={settings}
 				setPlaygroundState={setPlaygroundState}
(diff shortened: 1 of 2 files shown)


ematipico:
The issue now is that if we paste a long code, the scroll bar appears in a different way. Is it correct?

MichaReiser:
I noticed that too when debugging binary expression formatting. Should now be fixed.

xunilrj (approved):

ematipico:
I found a strange bug, not related to your PR. If I paste a big chunk of code, the prettier's output doesn't appear.

ematipico (approved):

MichaReiser:
> I found a strange bug, not related to your PR. If I paste a big chunk of code, the prettier's output doesn't appear.

Have you checked the console output for any errors?

ematipico:
> > I found a strange bug, not related to your PR. If I paste a big chunk of code, the prettier's output doesn't appear.
> 
> Have you checked the console output for any errors?

Yes, and no errors

Issue #3073, opened 2022-08-17
Implement a simple facade of Node.js APIs


ematipico:
Follow up of #3036

PR #3074, opened 2022-08-17
feat(npm): API facade and first compile pipeline

ematipico:
## Summary

Part of https://github.com/rome/tools/issues/3073

This PR starts works around the runtime APIs. This PR implements only few methods compared to the one proposed:
- `formatFiles`
- `formatContent`
- `parseContent`

The return type of these methods are not final yet and should not be part of the review. For example having `content` for `formatFiles` doesn't make sense. 

What's should be reviewed should be the input types for each API.

There are some tests that will fail once we start the actual work, there's some workflow with the GitHub actions and a compilation test with just TypeScript. Not sure if we really need a bundler for now.

## Test Plan

Added new tests and updated an existing workflow. The reason why I didn't add `npm/rome` path to the existing format workflow is because we use `setup-rome`, which contains the old version, while I am using the current version on `main` to format and check the JS code.

I will make a new PR to change that across the board, so we are able to dogfood our JS code base while we develop.

### DIFF
--- a/.github/workflows/pull_request_js.yml
+++ b/.github/workflows/pull_request_js.yml
@@ -9,6 +9,7 @@ on:
       - 'website/**'
       - 'editors/**'
       - 'crates/**'
+      - 'npm/rome/**'
 
 env:
   RUST_LOG: info
@@ -54,3 +55,27 @@ jobs:
       - name: Type check
         working-directory: website/playground
         run: pnpm tsc
+  apis-check:
+    name: Checks on APIs project
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR Branch
+        uses: actions/checkout@v3
+        with:
+          submodules: false
+      - name: Cache pnpm modules
+        uses: actions/cache@v3
+        with:
+          path: ~/.pnpm-store
+          key: ${{ runner.os }}-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-
+      - uses: pnpm/action-setup@v2.1.0
+        with:
+          version: 7
+      - name: Install libraries
+        working-directory: npm/rome
+        run: pnpm i
+      - name: CI checks
+        working-directory: npm/rome
+        run: pnpm ci
--- a/.github/workflows/pull_request_js.yml
+++ b/.github/workflows/pull_request_js.yml
@@ -55,27 +55,3 @@ jobs:
       - name: Type check
         working-directory: website/playground
         run: pnpm tsc
-  apis-check:
-    name: Checks on APIs project
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout PR Branch
-        uses: actions/checkout@v3
-        with:
-          submodules: false
-      - name: Cache pnpm modules
-        uses: actions/cache@v3
-        with:
-          path: ~/.pnpm-store
-          key: ${{ runner.os }}-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-
-      - uses: pnpm/action-setup@v2.1.0
-        with:
-          version: 7
-      - name: Install libraries
-        working-directory: npm/rome
-        run: pnpm i
-      - name: CI checks
-        working-directory: npm/rome
-        run: pnpm ci
--- a/npm/rome/src/index.ts
+++ b/npm/rome/src/index.ts
@@ -5,7 +5,7 @@ interface FormatFilesDebugOptions extends FormatFilesOptions {
 	debug: boolean;
 }
 
-interface FormatContentDebugOptions extends FormatFilesOptions {
+interface FormatContentDebugOptions extends FormatContentOptions {
 	/**
 	 * If `true`, you'll be able to inspect the IR of the formatter
 	 */
@@ -16,7 +16,7 @@ interface FormatFilesOptions {
 	/**
 	 * Writes the new content to disk
 	 */
-	write: boolean;
+	write?: boolean;
 }
 
 interface FormatContentOptions {
@@ -28,7 +28,7 @@ interface FormatContentOptions {
 	/**
 	 * The range where to format the content
 	 */
-	range: [number, number];
+	range?: [number, number];
 }
 
 interface FormatResult {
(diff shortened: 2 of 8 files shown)


npm/rome/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	write: boolean;
+}
+
+interface FormatContentOptions extends FormatDebugOptions {
+	/**
+	 * A virtual path of the file. You should add the extension,
+	 * so Rome knows how to parse the content
+	 */
+	filePath: String;
```
leops, line 20: Nit: use the lowercase `string` primitive type

npm/rome/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	ast: String;
+	// Not final
+	errors: String[];
+}
+
+export class Rome {
+	async formatFiles(
+		paths: String[],
+		options: Partial<FormatFilesOptions> | undefined = undefined,
```
leops, line 60: Nit: `options?: Partial<FormatFilesOptions>` seems simpler

npm/rome/src/index.ts
```diff
@@ -0,0 +1,95 @@
+interface FormatDebugOptions {
```
leops, line 1: Instead of having the `Format*Options` inherit from `FormatDebugOptions`, we could have it the other way around with `Format*OptionsWithDebug` variants adding the debug field, and overloaded declarations for the format methods returning a `FormatResultWithDebug` if th'yre called with the debug options type (similarly to how the `child_process` type definitions are declared to return different process types depending on how the `stdio` options are defined)

.github/workflows/pull_request_js.yml
```diff
@@ -56,3 +57,27 @@ jobs:
       - name: Type check
         working-directory: website/playground
         run: pnpm tsc
+  apis-check:
```
leops, line 60: Since the node.js API will eventually need to rely on a binary build of the Rust code (in whatever form that is), these tests should probably be in the main `pull_request` workflow instead

ematipico: Maybe we should have a workflow for this package (and others), because they should run when we modify Rust code and JS code.

leops: Yes I think it makes sense to introduce a Rust + JS workflow to do things like running the CLI in CI mode over the JS codebase for dogfooding purpose (instead of doing it only on changes to the JS code), along with running the tests for the node.js API

npm/rome/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+	 * The CST of the code
+	 */
+	cst: string;
+	/**
+	 * The AST of the code
+	 */
+	ast: string;
+	// Not final
+	errors: string[];
+}
```
MichaReiser, line 69: Can we rename `errors` to diagnostics. The parser may as well emit warnings. 


I don't think it makes sense for us to keep the `cst` and `ast` as string in the future. I recommend moving these into a `ParseDebugResult` and make them both optional (users can request if they want the CST or AST)

npm/rome/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+function isFormatContentDebug(
+	options: any,
+): options is FormatContentDebugOptions {
+	return options.debug !== undefined;
+}
+
+export class Rome {
+	async formatFiles(
+		paths: string[],
+		options?: Partial<FormatFilesOptions>,
```
MichaReiser, line 84: I recommend to manually make all optionos in `FormatFileOptions` optional because users may use `FormatFileOptions` in their code and having to use `Partial<FormatFileOptions>` everywhere is more work for them.

npm/rome/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+	cst: string;
+	/**
+	 * The AST of the code
+	 */
+	ast: string;
+	// Not final
+	errors: string[];
+}
+
+function isFormatFilesDebug(options: any): options is FormatFilesDebugOptions {
```
MichaReiser, line 71: Is there a way to constrain `options to be a `FormatFilesOptions`? Something like

```suggestion
function isFormatFilesDebug(options: FormatFilesDebugOptions | FormatFilesOptions): options is FormatFilesDebugOptions {
```

npm/rome/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+	options: any,
+): options is FormatContentDebugOptions {
+	return options.debug !== undefined;
+}
+
+export class Rome {
+	async formatFiles(
+		paths: string[],
+		options?: Partial<FormatFilesOptions>,
+	): Promise<FormatResult | FormatDebugResult> {
```
MichaReiser, line 85: I think it would be good if we use function overloading here so that users don't have to narrow down the return type

```javascript
async formatFiles(paths: string[], options?: Partial<FormatFilesOptions>): Promise<FormatResult>;
async formatFiles(paths: string[], options?: Partial<FormatDebugFilesOptions>): Promise<FormatResult>;
async formatFiles(path: string[], options?: Partial<FormatFilesOptions | FormatDebugFilesOptions>): Promise<FormatResult | FormatDebugResult> {
  ...
}
```

same for `formatContents`

npm/rome/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+		}
+		return {
+			content: "",
+			errors: [],
+		};
+	}
+
+	async parseContent(
+		content: string,
+		options: ParseOptions,
```
MichaReiser, line 121: Should the options be optional?

npm/rome/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+	// Available when in debug mode
+	ir: string | null;
+}
+
+interface ParseOptions {
+	/**
+	 * A virtual path of the file. You should add the extension,
+	 * so Rome knows how to parse the content
+	 */
+	filePath: string;
```
MichaReiser, line 55: It's important for parsing (even the playground) to be able to define the source type (ECMA version, JSX, Typescript, ...)

npm/rome/tests/index.test.ts
```diff
@@ -0,0 +1,47 @@
+import { describe, it, expect } from "vitest";
+import { Rome } from "../src/index";
+
+describe("Rome formatter", () => {
+	it("should not format files", async () => {
+		const rome = new Rome();
+
+		let result = await rome.formatFiles(["./path/to/file.js"]);
```
MichaReiser, line 8: Would you mind adding tests for the debug APIs to see how ergonomic it is to use the debug specific results

npm/rome/src/index.ts
ematipico: If we remove `ast` and `cst`, what would you expect to return from a parse action? (No debug)

ematipico: I will rename `errors` in another PR. Let's focus on the input types in this PR

npm/rome/src/index.ts
ematipico: Not unfortunately. If we use this type signature, the compiler will compliant as soon as we try to inspect `.debug` because it doesn't exist in `FormatFilesOptions`

npm/rome/src/index.ts
ematipico: No, because we need `filePath` in order to understand what we need to parse

npm/rome/src/index.ts
ematipico: Unfortunately we don't support this feature nowadays (even in the the CLI). We can add it in the future once we add support for it

npm/rome/src/index.ts
leops: I think in this case the TypeScript compiler wants you to write an explicit existence check for the property first:
```typescript
function isFormatFilesDebug(options: FormatFilesOptions): options is FormatFilesDebugOptions {
    return "debug" in options && options.debug === true;
}
```

ematipico: That's very ugly. Let's do it

ematipico:
@MichaReiser @xunilrj any chance to get this approved so I can move onto the next piece of logic?

xunilrj (approved):

PR #3075, opened 2022-08-17
test(rome_cli): more snapshot tests

ematipico:
## Summary

This PR adds more snapshot  testing to our CLI tests.
I found the issue that was preventing us from applying the snapshot testing to the rest of the tests: there some tests where we manually open a file from the file system and inspect their content; when doing so, we lock the file via `.lock_arc()`. This was preventing the snapshot logic to read these files unless they are freed. 

I manually added a `drop(file)` where needed, and now we have more snapshots!

## Test Plan

CI should pass


crates/rome_cli/tests/snapshots/indent_size_parse_errors_negative.snap
```diff
@@ -0,0 +1,6 @@
+---
```
ematipico, line 1: These "empty" snapshots are expected because we don't read the `Termination` error yet. I plan to make another PR where we read the `Result` of a CLI session.

leops (approved):
Since we're starting to have a lot of snapshots, it would be nice to sort them by passing the value of `module_path!()` to `assert_cli_snapshot` as a directory name for the snapshot

PR #3076, opened 2022-08-17
ci: run ci using the version on `main`

ematipico:
## Summary

This PR replace the usage of `setup-rome`, with the usage of the current version we have on `main`.

The main reason for changing this, is that we don't have enough occasions to dogfooding rome, and this change will help use doing so. Now we are starting to have a bunch JS code that we can use as guinea big. 

While doing so, I found a bunch of false positives. There's also some cases that we haven't considered:

```js
app.on(res, req) {
	req.something;
}
```

In this snippet `res` is not used, and the suggested fix asks to remove `res`. Not sure if we should ignore the case or leave it as is.

## Test Plan

The CI should pass


editors/vscode/src/commands/syntaxTree.ts
```diff
@@ -114,7 +114,7 @@ class SyntaxTreeProvider
 
 	provideDocumentLinks(
 		document: TextDocument,
-		token: CancellationToken,
+		// token: CancellationToken,
```
MichaReiser, line 117: Delete?

MichaReiser (approved):
Thank you.

One thing to keep in mind is that this will increase the size of any formatter diff because there are now even more changes (and changes to the playground may cause formatter conflicts as well)

website/playground/workers-site/index.js
```diff
@@ -65,30 +61,9 @@ async function handleEvent(event) {
 					...notFoundResponse,
 					status: 404,
 				});
-			} catch (e) {}
+			} catch {}
```
MichaReiser, line 64: Do we need this second catch block. Isn't there already a catch block on line `51`?

leops (commented):
Currently there's a `test-cli` job in the `pull_request` workflow that does more or less the same thing (build and run the CLI on the editors and website directories) but is marked as `continue-on-error`, I guess it can be removed now.
Also we should probably change the trigger for the `pull_request_js` workflow now to include the Rust codebase as well, to ensure the JS code doesn't go out of sync again when we change the analyzer or formatter (unfortunately it makes the Rust / JS workflow separation somewhat meaningless)

ematipico:
> Unfortunately it makes the Rust / JS workflow separation somewhat meaningless

Are you up to merge `pull_request_js` and `pull_request` and run JS and Rust test in one single workflow action? I think we're arrived at that point now...

leops:
> > Unfortunately it makes the Rust / JS workflow separation somewhat meaningless
> 
> Are you up to merge `pull_request_js` and `pull_request` and run JS and Rust test in one single workflow action? I think we're arrived at that point now...

I don't think I would completely merge the existing workflows, I think we could still have 3 `pull_request_rs`, `pull_request_js` and `pull_request` workflows, and redistribute the various jobs to the correct workflow

ematipico:
> > > Unfortunately it makes the Rust / JS workflow separation somewhat meaningless
> > 
> > 
> > Are you up to merge `pull_request_js` and `pull_request` and run JS and Rust test in one single workflow action? I think we're arrived at that point now...
> 
> I don't think I would completely merge the existing workflows, I think we could still have 3 `pull_request_rs`, `pull_request_js` and `pull_request` workflows, and redistribute the various jobs to the correct workflow

OK, another PR for that :) better to not lose focus. Cheers

PR #3080, opened 2022-08-17
ci: fix paths of snapshots and faster CI

ematipico:
## Summary

Paths that contains `::` are bad for Windows. I fixed it by replacing it with `_`.

Also, changed the workflow to use the release version of our binary when running `ci` of JS files.

## Test Plan

CI should pass


crates/rome_cli/tests/snapshots/main_check/apply_noop.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: crates/rome_cli/tests/snap_test.rs
-assertion_line: 145
+assertion_line: 147
```
MichaReiser, line 3: Someone recently mentioned that the new insta version is no longer adding these lines. Can we upgrade insta to avoid these snapshot changes?

MichaReiser (approved):

crates/rome_cli/tests/snapshots/main_check/apply_noop.snap
ematipico: It's been done already across the board, but it only works on new snapshots, it seems

Issue #3077, opened 2022-08-17
🐛 false positive noShoutyConstants


IWANABETHATGUY:
### Environment information

According https://rome.tools/docs/lint/rules/noShoutyConstants/
> Disallow the use of constants which its value is the upper-case version of its name.

Such code snippet should not emit a warning: 
```js
let lib = "lib"
```

https://play.rome.tools/?code=let+lib+%3D+%22lib%22&lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&typescript=false&jsx=false#bABlAHQAIABsAGkAYgAgAD0AIAAiAGwAaQBiACIA

### What happened?

Rome  emit a diagnostic 

### Expected result

This code snippet will not raise an error.

Also 
for such code snippet: 
```js
export const CON = 'CON'
```
don't raise an error, because it is a common pattern for `redux` and `vuex` when writing action type.

### Code of Conduct

- [X] I agree to follow Rome's Code of Conduct

PR #3081, opened 2022-08-17
fix(rome_js_analyze): fix noShoutyConstants with lowercase and being exported

xunilrj:
## Summary

Closes https://github.com/rome/tools/issues/3077.

## Test Plan

```
> cargo test -p rome_js_analyze -- shouty
```

with two new cases

```js
let foo = "foo";

const A = "A";
export default A;
```

### DIFF
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
@@ -46,11 +46,16 @@ fn is_id_and_string_literal_inner_text_equal(
         .as_js_string_literal_expression()?;
     let literal_text = literal.inner_string_text();
 
-    if id_text == literal_text {
-        Some((id.clone(), literal.clone()))
-    } else {
-        None
+    for (from_id, from_literal) in id_text.chars().zip(literal_text.chars()) {
+        if from_id != from_literal {
+            return None;
+        }
+        if from_id.is_lowercase() || from_literal.is_lowercase() {
+            return None;
+        }
     }
+
+    Some((id.clone(), literal.clone()))
 }
 
 pub struct State {
@@ -74,6 +79,11 @@ impl Rule for NoShoutyConstants {
         if declaration.is_const() {
             if let Some((binding, literal)) = is_id_and_string_literal_inner_text_equal(declarator)
             {
+                let model = ctx.model();
+                if model.is_exported(&binding) {
+                    return None;
+                }
+
                 return Some(State {
                     literal,
                     references: binding.all_references(ctx.model()).collect(),
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
@@ -23,6 +23,25 @@ declare_rule! {
     /// const FOO = "FOO";
     /// console.log(FOO);
     /// ```
+    ///
+    /// ### Valid
+    ///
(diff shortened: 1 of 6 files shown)


xunilrj:
!bench_analyzer

IWANABETHATGUY:
A few suggestions:
1. Would you mind adding some valid cases in our generated markdown? https://github.com/rome/tools/pull/3078/files#diff-fdf4fb740794b3a765f8604dee5b3939df2fa9a5bf566e910c31b7db591373bcR26-R31

crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
```diff
@@ -46,11 +46,16 @@ fn is_id_and_string_literal_inner_text_equal(
         .as_js_string_literal_expression()?;
     let literal_text = literal.inner_string_text();
 
-    if id_text == literal_text {
-        Some((id.clone(), literal.clone()))
-    } else {
-        None
+    for (from_id, from_literal) in id_text.chars().zip(literal_text.chars()) {
```
leops, line 49: Technically both `id_text` and `literal_text` only ever cover a single token (`IDENT` and `JS_STRING_LITERAL`), so I think it would be possible to run this check over two `&str` directly rather than having to implement these helpers for `SyntaxNodeText`

crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
```diff
@@ -46,11 +46,16 @@ fn is_id_and_string_literal_inner_text_equal(
...
+        if from_id != from_literal {
+            return None;
+        }
+        if from_id.is_lowercase() || from_literal.is_lowercase() {
```
leops, line 53: I'm not sure the exit condition for the loop is correct here, from the description of the rule I would expect it to raise a diagnostic for `const foo = "FOO"`

leops:
> A few suggestions:
> 
>     1. Would you mind adding some valid cases in our generated markdown? https://github.com/rome/tools/pull/3078/files#diff-fdf4fb740794b3a765f8604dee5b3939df2fa9a5bf566e910c31b7db591373bcR26-R31

We can also pull the valid cases from the original rule in Rome JS: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noShoutyConstants.md

crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
xunilrj: That is a good question. It is my understanding that the rule is only for uppercase bindings.  ```const foo = "FOO"```  seems fine for me because it correctly gives a camel case name to a constant.

crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
xunilrj: Done.
But I have mixed feelings about this:
+ On one side is probably faster.
- But we are also not using the typed AST here. So I have to take care around quotes.

xunilrj:
> > A few suggestions:

Done.

crates/rome_js_analyze/src/semantic_analyzers/js/no_shouty_constants.rs
ematipico: Yeah the rule should work only for upper cases. That was its initial intention

ematipico (approved):

Issue #3038, opened 2022-08-10
☂️ Analyzer Correctness


xunilrj:
### Description

This is an umbrella issue to improve the Analyzer correctness.

- [x] Increase automated coverage of the linter;
- [x] Run the analyzer against git repositories;
- [ ] Being able to track times for single rules;
- [ ] Run existing rules against the Rome classic code base;
- [ ] Run the semantic model as part of t262 tests;
- [ ] Move parser checks/diagnostics inside lint rules

ematipico:
@xunilrj could you please tick the task that have been implemented? 

As far as I understood the correctness project came to end, so should we close the issue?

xunilrj:
I created a discussion for "Being able to track times for single rules;" https://github.com/rome/tools/discussions/3127 but we have not decided to go with this solution.

PR #3084, opened 2022-08-18
feat(rome_js_semantic): Improvements to scope resolution

xunilrj:
## Summary

This PR start to use the semantic model with our coverage test suites. See https://github.com/rome/tools/issues/3038.

This new "example" actually compares our scope resolution to other tools. Our compliance percentage is still quite low, especially because:

1 - Javascript tools use start/end as characters, we as bytes; when tests cases have Unicode it is very hard to compare values;
2 - Some tools treat trivia differently than us, so we need to sometimes use ```text_range```, and sometimes use the trimmed range;
3 - Some tools create scopes that, as I understand, are not necessary.

## Test Plan

```
> cargo test -p rome_js_semantic
> cargo run -p rome_js_semantic --example coverage_scope --release
```

### DIFF
--- a/crates/rome_js_semantic/src/events.rs
+++ b/crates/rome_js_semantic/src/events.rs
@@ -256,10 +256,20 @@ impl SemanticEventExtractor {
             | JS_FUNCTION_EXPORT_DEFAULT_DECLARATION
             | JS_FUNCTION_EXPRESSION
             | JS_ARROW_FUNCTION_EXPRESSION
+            | JS_CLASS_DECLARATION
+            | JS_CLASS_EXPORT_DEFAULT_DECLARATION
+            | JS_CLASS_EXPRESSION
             | JS_CONSTRUCTOR_CLASS_MEMBER
+            | JS_METHOD_CLASS_MEMBER
             | JS_GETTER_CLASS_MEMBER
             | JS_SETTER_CLASS_MEMBER
-            | JS_FUNCTION_BODY => {
+            | JS_METHOD_OBJECT_MEMBER
+            | JS_GETTER_OBJECT_MEMBER
+            | JS_SETTER_OBJECT_MEMBER
+            | JS_FUNCTION_BODY
+            | TS_INTERFACE_DECLARATION
+            | TS_ENUM_DECLARATION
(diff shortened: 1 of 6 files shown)


ematipico:
> 3 - Some tools create scopes that, as I understand, are not necessary.

Out of curiosity, can you share some example?

xtask/coverage/src/symbols/scope.js
```diff
@@ -0,0 +1,40 @@
+const parser = require("@babel/parser");
```
ematipico, line 1: Should this package be part of the PR? If so, could you please update the description of the PR and explain what's this used for?

ematipico (changes requested):
- there's a `a.txt` file to remove.
- there's a `scopes.txt` file to remove
- we should use `pnpm` as package manager, unless there are exceptions (I suppose here we can safely use `pnpm`)

xunilrj:
> * there's a `a.txt` file to remove.
> * there's a `scopes.txt` file to remove

Argh.. Sorry, I thought I had removed these.

crates/rome_js_semantic/src/events.rs
```diff
@@ -256,18 +256,28 @@ impl SemanticEventExtractor {
...
             | JS_GETTER_CLASS_MEMBER
             | JS_SETTER_CLASS_MEMBER
-            | JS_FUNCTION_BODY => {
+            | JS_METHOD_OBJECT_MEMBER
+            | JS_GETTER_OBJECT_MEMBER
+            | JS_SETTER_OBJECT_MEMBER
+            | JS_FUNCTION_BODY
+            | TS_INTERFACE_DECLARATION
+            | TS_ENUM_DECLARATION
+            | TS_TYPE_ALIAS_DECLARATION => {
```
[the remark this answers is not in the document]
xunilrj: Yes. Any reason you think we should not?

xtask/coverage/src/symbols/scope.js
xunilrj: > Should this package be part of the PR?

No. I will move these to another place.

ematipico (approved):

Issue #2748, opened 2022-06-20
useOptionalChain


ematipico:

denbezrukov:
Would you mind assigning me this task?🙏🏽

ematipico:
@denbezrukov Done!

denbezrukov:
There is a conditional which is used to check `isLeftSideLowerPrecedence`.

- https://github.com/typescript-eslint/typescript-eslint/blob/943523d9c345f01a7e285eb4440be2998614186d/packages/eslint-plugin/src/rules/prefer-optional-chain.ts#L73-L83

We need it for the case then it's necessary to add parentheses around left part of `LogicalExpression` after fixing the rule:

`(await a || {}).b` -> `(await a)?.b`

I found `OperatorPrecedence` enum. 
https://github.com/rome/tools/blob/282bda492a1479167cc884e2b5cd7724c8db3011/crates/rome_js_parser/src/syntax/util.rs#L8-L30

I was wondering if I could make it public and port `getOperatorPrecedence` function from `typescript-eslint`:

- https://github.com/typescript-eslint/typescript-eslint/blob/943523d9c345f01a7e285eb4440be2998614186d/packages/eslint-plugin/src/util/getOperatorPrecedence.ts#L195

What do you think?🙏🏽

denbezrukov:
@ematipico 
Could you please explain how `FormatPrecedence::with_precedence_for_parenthesis` works now? Does It return `FormatPrecedence::High` for expressions which have more precedence over parentheses? And `FormatPrecedence::Low` for expressions which have less precedence over parentheses? 

I guess that `with_precedence_for_optional_operator` should work. I see only one drawback that next time when we need to implement, for example, `with_precedence_for_async_await` we'll again struggle with [Operator Precedence Table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table).

It seems that `getOperatorPrecedence` function from `typescript-eslint` can handle all cases. Because it returns  `OperatorPrecedence` which already derives `Ord` and `PartialOrd`.
https://github.com/rome/tools/blob/282bda492a1479167cc884e2b5cd7724c8db3011/crates/rome_js_parser/src/syntax/util.rs#L7-L8

We can implement this function and use it in `with_precedence_for_optional_operator` and `with_precedence_for_parenthesis`.

E.g. `with_precedence_for_optional_operator`:  
1. pass `JsSyntaxNode` as argument.
[...]

PR #3085, opened 2022-08-19
feat(rome_js_analyze): useOptionalChain #2748

denbezrukov:
## Summary

Resolved https://github.com/rome/tools/issues/2748
The algorithms are described in the documentation for `LogicalAndChain` and `LogicalOrLikeChain`.

Open questions:
1. This PR handles cases only with logical and expressions and logical or/nullish **separately**. 
 `foo && foo.bar && foo.bar.baz` generates one warning and one fix.
 `((foo || {}).bar || {}).baz` also generates one warning and one fix.
`((foo && foo.bar) || {}).baz` generates two warnings and two fixes.
2. `LogicalOrLikeChain` collects nodes which need to transform in the `action` phase.   `LogicalAndChain` collects in the `run` phase. Technically, this can be done for `LogicalAndChain` (we can compare only one sub-chain to generate warning) as well, but I'm afraid it might complicate the code. I will mark this code fragment in the comments.
3. Currently do not handle complex computed properties:
```
foo && foo[bar as string] && foo[bar as string].baz;
foo && foo[1 + 2] && foo[1 + 2].baz;
foo && foo[typeof bar] && foo[typeof bar].baz;
```
4. Better warning message?

## Test Plan

Cargo test

Test cases from [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/prefer-optional-chain.test.ts)

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
+++ b/crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
@@ -1,0 +1,862 @@
+use std::cmp::Ordering;
+use std::collections::VecDeque;
+use std::iter;
+
+use rome_analyze::{
+    context::RuleContext, declare_rule, ActionCategory, Ast, Rule, RuleCategory, RuleDiagnostic,
+};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_factory::make;
+use rome_js_syntax::{
+    JsAnyExpression, JsAnyName, JsComputedMemberExpression, JsLogicalExpression, JsLogicalOperator,
+    JsStaticMemberExpression, OperatorPrecedence, T,
+};
+use rome_rowan::{declare_node_union, AstNode, AstNodeExt, BatchMutationExt, SyntaxResult};
+
+use crate::JsRuleAction;
+
+declare_rule! {
+    /// Enforce using concise optional chain instead of chained logical expressions.
+    ///
+    /// TypeScript 3.7 added support for the optional chain operator.
+    /// This operator allows you to safely access properties and methods on objects when they are potentially `null` or `undefined`.
+    /// The optional chain operator only chains when the property value is `null` or `undefined`.
+    /// It is much safer than relying upon logical operator chaining; which chains on any truthy value.
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// foo && foo.bar && foo.bar.baz && foo.bar.baz.buzz
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// foo.bar && foo.bar.baz.buzz
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// foo !== undefined && foo.bar != undefined && foo.bar.baz !== null && foo.bar.baz.buzz
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// ((foo || {}).bar || {}).baz;
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// (await (foo1 || {}).foo2 || {}).foo3;
+    /// ```
+    ///
+    /// ```ts,expect_diagnostic
+    /// (((typeof x) as string) || {}).bar;
+    /// ```
+    ///
+    /// ### Valid
+    ///
+    /// ```js
+    /// foo && bar;
+    ///```
+    /// ```js
+    /// foo || {};
+    ///```
+    ///
+    /// ```js
+    /// (foo = 2 || {}).bar;
+    ///```
+    ///
+    /// ```js
+    /// foo || foo.bar;
+    ///```
+    ///
+    /// ```js
+    /// foo["some long"] && foo["some long string"].baz
+    ///```
+    ///
+    pub(crate) UseOptionalChain {
+        version: "0.10.0",
+        name: "useOptionalChain",
+        recommended: true,
+    }
+}
+
+pub(crate) enum UseOptionalChainState {
+    LogicalAnd(VecDeque<JsAnyExpression>),
+    LogicalOrLike(LogicalOrLikeChain),
+}
+
+impl Rule for UseOptionalChain {
+    const CATEGORY: RuleCategory = RuleCategory::Lint;
+
+    type Query = Ast<JsLogicalExpression>;
+    type State = UseOptionalChainState;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let logical = ctx.query();
+        let operator = logical.operator().ok()?;
+
+        match operator {
+            JsLogicalOperator::LogicalAnd => {
+                let head = logical.right().ok()?;
+
+                let chain = LogicalAndChain::from_expression(head).ok()?;
+
+                if chain.is_inside_another_chain().ok()? {
+                    return None;
+                }
+
+                let optional_chain_expression_nodes = chain.optional_chain_expression_nodes()?;
+
+                Some(UseOptionalChainState::LogicalAnd(
+                    optional_chain_expression_nodes,
+                ))
+            }
+            JsLogicalOperator::NullishCoalescing | JsLogicalOperator::LogicalOr => {
+                let chain = LogicalOrLikeChain::from_expression(logical)?;
+
+                if chain.is_inside_another_chain() {
+                    return None;
+                }
+
+                Some(UseOptionalChainState::LogicalOrLike(chain))
+            }
+        }
+    }
+
+    fn diagnostic(ctx: &RuleContext<Self>, state: &Self::State) -> Option<RuleDiagnostic> {
+        let range = match state {
+            UseOptionalChainState::LogicalAnd(_) => ctx.query().range(),
+            UseOptionalChainState::LogicalOrLike(state) => state.member.range(),
+        };
+
+        Some(RuleDiagnostic::new(
+            range,
+            markup! {
+                "Change to an optional chain."
+            }
+            .to_owned(),
+        ))
+    }
+
+    fn action(ctx: &RuleContext<Self>, state: &Self::State) -> Option<JsRuleAction> {
+        match state {
+            UseOptionalChainState::LogicalAnd(optional_chain_expression_nodes) => {
+                let mut prev_expression = None;
+
+                for expression in optional_chain_expression_nodes {
+                    let next_expression = prev_expression
+                        .take()
+                        .and_then(|(prev_expression, new_expression)| {
+                            expression
+                                .clone()
+                                .replace_node(prev_expression, new_expression)
+                        })
+                        .unwrap_or_else(|| expression.clone());
+
+                    let next_expression = match next_expression {
+                        JsAnyExpression::JsCallExpression(call_expression) => {
+                            let mut call_expression_builder = make::js_call_expression(
+                                call_expression.callee().ok()?,
+                                call_expression.arguments().ok()?,
+                            )
+                            .with_optional_chain_token(make::token(T![?.]));
+
+                            if let Some(type_arguments) = call_expression.type_arguments() {
+                                call_expression_builder =
+                                    call_expression_builder.with_type_arguments(type_arguments);
+                            }
+
+                            let call_expression = call_expression_builder.build();
+
+                            JsAnyExpression::from(call_expression)
+                        }
+                        JsAnyExpression::JsStaticMemberExpression(member_expression) => {
+                            let operator_token = member_expression.operator_token().ok()?;
+
+                            JsAnyExpression::from(make::js_static_member_expression(
+                                member_expression.object().ok()?,
+                                make::token(T![?.])
+                                    .with_leading_trivia_pieces(
+                                        operator_token.leading_trivia().pieces(),
+                                    )
+                                    .with_trailing_trivia_pieces(
+                                        operator_token.trailing_trivia().pieces(),
+                                    ),
+                                member_expression.member().ok()?,
+                            ))
+                        }
+                        JsAnyExpression::JsComputedMemberExpression(member_expression) => {
+                            let operator_token = match member_expression.optional_chain_token() {
+                                Some(token) => make::token(T![?.])
+                                    .with_leading_trivia_pieces(token.leading_trivia().pieces())
+                                    .with_trailing_trivia_pieces(token.trailing_trivia().pieces()),
+                                None => make::token(T![?.]),
+                            };
+
+                            JsAnyExpression::from(
+                                make::js_computed_member_expression(
+                                    member_expression.object().ok()?,
+                                    member_expression.l_brack_token().ok()?,
+                                    member_expression.member().ok()?,
+                                    member_expression.r_brack_token().ok()?,
+                                )
+                                .with_optional_chain_token(operator_token)
+                                .build(),
+                            )
+                        }
+                        _ => return None,
+                    };
+
+                    prev_expression = Some((expression.clone(), next_expression));
+                }
+
+                let (prev_expression, new_expression) = prev_expression?;
+
+                let logical = ctx.query();
+                let next_right = logical
+                    .right()
+                    .ok()?
+                    .replace_node(prev_expression, new_expression.clone())
+                    .unwrap_or(new_expression);
+
+                let mut mutation = ctx.root().begin();
+
+                mutation.replace_node(JsAnyExpression::from(logical.clone()), next_right);
+
+                Some(JsRuleAction {
+                    category: ActionCategory::QuickFix,
+                    applicability: Applicability::MaybeIncorrect,
+                    message: markup! { "Change to an optional chain." }.to_owned(),
+                    mutation,
+                })
+            }
+            UseOptionalChainState::LogicalOrLike(chain) => {
+                let chain = chain.optional_chain_expression_nodes();
+
+                let mut prev_chain: Option<(JsAnyMemberExpression, JsAnyMemberExpression)> = None;
+
+                for (left, member) in chain {
+                    let left = if let Some((prev_member, next_member)) = prev_chain.take() {
+                        left.replace_node(prev_member, next_member.clone())
+                            .unwrap_or_else(|| next_member.into())
+                    } else {
+                        left
+                    };
+
+                    let left = trim_trailing_space(left)?;
+                    let need_parenthesis =
+                        left.precedence().ok()? < OperatorPrecedence::LeftHandSide;
+
+                    let left = if need_parenthesis {
+                        make::js_parenthesized_expression(
+                            make::token(T!['(']),
+                            left,
+                            make::token(T![')']),
+                        )
+                        .into()
+                    } else {
+                        left
+                    };
+
+                    let next_member = match member.clone() {
+                        JsAnyMemberExpression::JsStaticMemberExpression(expression) => {
+                            let static_member_expression = make::js_static_member_expression(
+                                left,
+                                make::token(T![?.]),
+                                expression.member().ok()?,
+                            );
+                            JsAnyMemberExpression::from(static_member_expression)
+                        }
+                        JsAnyMemberExpression::JsComputedMemberExpression(expression) => {
+                            let computed_member_expression = make::js_computed_member_expression(
+                                left,
+                                expression.l_brack_token().ok()?,
+                                expression.member().ok()?,
+                                expression.r_brack_token().ok()?,
+                            )
+                            .with_optional_chain_token(make::token(T![?.]))
+                            .build();
+
+                            computed_member_expression.into()
+                        }
+                    };
+
+                    prev_chain = Some((member, next_member));
+                }
+
+                let (prev_member, new_member) = prev_chain?;
+
+                let mut mutation = ctx.root().begin();
+
+                mutation.replace_node(prev_member, new_member);
+
+                Some(JsRuleAction {
+                    category: ActionCategory::QuickFix,
+                    applicability: Applicability::MaybeIncorrect,
+                    message: markup! { "Change to an optional chain." }.to_owned(),
+                    mutation,
+                })
+            }
+        }
+    }
+}
+
+/// `LogicalAndChainOrdering` is the result of a comparison between two logical and chain.
+enum LogicalAndChainOrdering {
+    /// An ordering where a chain is a sub-chain of another.
+    /// ```js
+    /// (foo && foo.bar) /* is sub-chain of */ (foo && foo.bar && foo.bar.baz)
+    /// ```
+    SubChain,
+    /// An ordering where a chain is equal to another.
+    /// ```js
+    /// (foo && foo.bar) /* is equal */ (foo && foo.bar)
+    /// ```
+    Equal,
+    /// An ordering where a chain is different to another.
+    /// ```js
+    /// (foo && foo.bar) /* is different */ (bar && bar.bar && bar.bar.baz)
+    /// ```
+    Different,
+}
+
+/// `LogicalAndChain` handles cases with `JsLogicalExpression` which has `JsLogicalOperator::LogicalAnd` operator:
+/// ```js
+/// foo && foo.bar && foo.bar.baz && foo.bar.baz.buzz;
+///
+/// foo.bar && foo.bar.baz && foo.bar.baz.buzz;
+///
+/// foo !== undefined && foo.bar;
+/// ```
+/// The main idea of the `LogicalAndChain`:
+/// 1. Check that the current chain isn't in another `LogicalAndChain`. We need to find the topmost logical expression which will be the head of the first current chain.
+/// 2. Go down thought logical expressions and collect other chains and compare them with the current one.
+/// 3. If a chain is a sub-chain of the current chain, we assign that sub-chain to new current one. Difference between current chain and sub-chain is a tail.
+/// 4. Save the first tail `JsAnyExpression` to the buffer.
+/// 5. Transform every `JsAnyExpression` from the buffer to optional expression.
+///
+/// E.g. `foo && foo.bar.baz && foo.bar.baz.zoo;`.
+/// The logical expression `foo && foo.bar.baz` isn't the topmost. We skip it.
+/// `foo && foo.bar.baz && foo.bar.baz.zoo;` is the topmost and it'll be a start point.
+/// We start collecting a chain. We collect `JsAnyExpression` but for clarity let's use string identifiers.
+/// `foo.bar.baz.zoo;` -> `[foo, bar, baz, zoo]`
+/// Next step we take a next chain and also collect it.
+/// `foo.bar.baz` -> `[foo, bar, baz]`
+/// By comparing them we understand that one is a sub-chain of the other. `[foo, bar, baz]` is new current chain. `[zoo]` is a tail.
+/// We save `zoo` expression to the buffer.
+/// Next step we take a next chain and also collect it.
+/// `foo` -> `[foo]`
+/// By comparing them we understand that one is a sub-chain of the other. `[foo]` is new current chain. `[bar, baz]` is a tail.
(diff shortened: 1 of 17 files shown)


crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+
+        match operator {
+            JsLogicalOperator::LogicalAnd => {
+                let chain = LogicalAndChain::new(logical.right().ok()?);
+
+                if chain.is_inside_another_chain().unwrap_or(false) {
+                    return None;
+                }
+
+                let optional_chain_expression_nodes = chain.optional_chain_expression_nodes()?;
```
denbezrukov, line 107: Here we traversal all logical and chains and collect expression to transform them later. 
We can optimize that with early return then find only one expression which we can transform.
And in action phase traversal all logical chains.

What do you think?

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -446,6 +489,127 @@ impl JsRegexLiteralExpression {
     }
 }
 
+impl JsAnyExpression {
+    /// Try to extract non `JsParenthesizedExpression` from `JsAnyExpression`
+    pub fn omit_parentheses(self) -> JsAnyExpression {
```
MichaReiser, line 494: We've been discussing for a while that it may be useful to have a `resolve` function that resolves the node that this node evaluates to at runtime. 

```
a -> a
call(b) -> call(b)

(a) -> a
(a as number) -> a
(a, b) -> b
```

Would this be useful for your linter and could be a potential replacement for `omit_parentheses` which only solves a very specific case.

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -446,6 +489,127 @@ impl JsRegexLiteralExpression {
...
+            JsAnyExpression::JsConditionalExpression(_) => OperatorPrecedence::Conditional,
+            JsAnyExpression::JsAssignmentExpression(_) => OperatorPrecedence::Assignment,
+            JsAnyExpression::JsInExpression(_)
+            | JsAnyExpression::JsInstanceofExpression(_)
+            | JsAnyExpression::TsAsExpression(_) => OperatorPrecedence::Relational,
+            JsAnyExpression::JsLogicalExpression(expression) => match expression.operator()? {
+                JsLogicalOperator::LogicalAnd => OperatorPrecedence::LogicalAnd,
+                JsLogicalOperator::LogicalOr => OperatorPrecedence::LogicalOr,
+                JsLogicalOperator::NullishCoalescing => OperatorPrecedence::Coalesce,
+            },
```
MichaReiser, line 520:
```suggestion
            JsAnyExpression::JsLogicalExpression(expression) => match expression.operator()?.precedence(),
```

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -446,6 +489,127 @@ impl JsRegexLiteralExpression {
...
+                | JsBinaryOperator::Divide
+                | JsBinaryOperator::Times => OperatorPrecedence::Multiplicative,
+                JsBinaryOperator::Exponent => OperatorPrecedence::Exponential,
+                JsBinaryOperator::UnsignedRightShift
+                | JsBinaryOperator::RightShift
+                | JsBinaryOperator::LeftShift => OperatorPrecedence::Shift,
+                JsBinaryOperator::BitwiseAnd => OperatorPrecedence::BitwiseAnd,
+                JsBinaryOperator::BitwiseOr => OperatorPrecedence::BitwiseOr,
+                JsBinaryOperator::BitwiseXor => OperatorPrecedence::BitwiseXor,
+            },
```
MichaReiser, line 541:
```suggestion
            JsAnyExpression::JsBinaryExpression(expression) => match expression.operator()?.precedence(),
```

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -379,6 +418,10 @@ impl JsObjectExpression {
     pub fn has_trailing_comma(&self) -> bool {
         self.members().trailing_separator().is_some()
     }
+
+    pub fn is_empty(&self) -> bool {
+        self.members().len() == 0
```
MichaReiser, line 423: I think this should work
```suggestion
        self.members().is_empty
```

crates/rome_js_syntax/src/expr_ext.rs
denbezrukov: Thank you!

crates/rome_js_syntax/src/expr_ext.rs
denbezrukov: 👍👍👍 Thx!!!

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -446,6 +489,103 @@ impl JsRegexLiteralExpression {
...
+
+        iter::successors(first, |expression| {
+            let parenthesized = expression.as_js_parenthesized_expression()?;
+            parenthesized.expression().ok()
+        })
+        .last()
+        .unwrap_or(self)
+    }
+
+    pub fn precedence(&self) -> SyntaxResult<OperatorPrecedence> {
```
denbezrukov, line 507: https://github.com/typescript-eslint/typescript-eslint/blob/943523d9c345f01a7e285eb4440be2998614186d/packages/eslint-plugin/src/util/getOperatorPrecedence.ts#L195

crates/rome_js_syntax/src/expr_ext.rs
denbezrukov: Interesting (:
Is it only proposal or I could find an example?

MichaReiser: Only a proposal discussed on [discord](https://discord.com/channels/678763474494423051/978209399120216076/1004723235620917249).

ematipico:
!bench_analyzer

MichaReiser:
Thanks again. I plan to have a closer look at the PR today

crates/rome_js_analyze/src/lib.rs
```diff
@@ -106,6 +106,32 @@ mod tests {
 
     use crate::{analyze, AnalysisFilter, ControlFlow};
 
+    #[ignore]
+    #[test]
+    fn quick_test() {
```
xunilrj, line 111: Should this be here?
Maybe a better name and not ignoring it?

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+
+    fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
+        let logical = ctx.query();
+        let operator = logical.operator().ok()?;
+
+        match operator {
+            JsLogicalOperator::LogicalAnd => {
+                let chain = LogicalAndChain::new(logical.right().ok()?);
+
+                if chain.is_inside_another_chain().unwrap_or(false) {
```
xunilrj, line 103: Can we replace this check, by checking that we are the "root" expression at the beginning of the method?

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+/// `LogicalAndChain` handles cases with `JsLogicalExpression` which has `JsLogicalOperator::LogicalAnd` operator:
+/// ```js
+/// foo && foo.bar && foo.bar.baz && foo.bar.baz.buzz;
+///
+/// foo.bar && foo.bar.baz && foo.bar.baz.buzz;
+///
+/// foo !== undefined && foo.bar;
+/// ```
+/// The main idea of the `LogicalAndChain`:
+/// 1. Check that the current chain isn't in another `LogicalAndChain`. We need to find the topmost logical expression which will be the head of the first current chain.
```
xunilrj, line 332: My fear is that we are, for every expression, trying to find the  "topmost expression"  and bailing if we are not it. How expensive is this?

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+}
+
+impl LogicalAndChain {
+    fn new(head: JsAnyExpression) -> LogicalAndChain {
+        /// Iterate over `JsAnyExpression` and collect every expression which is a part of the chain:
+        /// ```js
+        /// foo.bar[baz];
+        /// ```
+        /// `[JsReferenceIdentifier, JsStaticMemberExpression, JsComputedMemberExpression]`
+        fn collect_chain(expression: JsAnyExpression) -> VecDeque<JsAnyExpression> {
```
xunilrj, line 367: Any reason for this being an "fn", given that this is not recursive? 
+1 for not being recursive!

MichaReiser: I think it simplifies the loop so that it's possible to use `return`

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+///
+#[derive(Debug)]
+pub(crate) struct LogicalAndChain {
+    head: JsAnyExpression,
+    /// The buffer of `JsAnyExpression` which need to make optional chain.
+    buf: VecDeque<JsAnyExpression>,
+}
+
+impl LogicalAndChain {
+    fn new(head: JsAnyExpression) -> LogicalAndChain {
```
MichaReiser, line 361: Nit: I would find it surprising that `new` performs such an expensive operation. Maybe rename to `from_expression`

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+                                Err(_) => return buf,
+                            }
+                        } else {
+                            return buf;
+                        }
+                    }
+                    expression => expression,
+                };
+
+                match &expression {
```
MichaReiser, line 395: Should the chain also handle Typescript type casts `(a as number)?.test` or `(<number> a)?.test` what about parenthesized expressions?

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// foo && foo.bar && foo.bar.baz && foo.bar.baz.buzz
+    /// ```
+    ///
+    /// ```js,expect_diagnostic
+    /// foo.bar && foo.bar.baz.buzz
```
MichaReiser, line 36: Would you mind adding a few tests that have comments to see if they are retained in the refactor

MichaReiser (commented):
This is very impressive. Overall, this is looking good to me.

crates/rome_js_analyze/src/lib.rs
denbezrukov: I wanted to have something to a debug rule.
We already have a simular test for `rome_js_formatter`. https://github.com/rome/tools/blob/65c1e868285341d0aef7767d454842bdf151fa81/crates/rome_js_formatter/src/lib.rs#L713-L738

crates/rome_js_syntax/src/expr_ext.rs
ematipico: I'd suggest to have it outside of this PR. Even though it was proposed on discord, we need to have an issue with some requirements. We don't know where we could have this utility.

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: I thought about that.
I had concerns that these cases were unusual for logical _**and**_ chains. Because it needs to repeat the same expressions for every sub-chains.
E.g.:

`foo && (foo as IObject).bar && (foo as IObject).bar.baz`.

And it's simpler to use other expression in logical _**or**_ chains.
E.g:

`(((foo as IObject) || {}).bar || {}).baz`.

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,855 @@
...
+    /// foo && foo.bar && foo.bar.baz
+    /// ```
+    SubChain,
+    /// An ordering where a chain is equal to another.
+    /// ```js
+    /// foo && foo.bar
+    /// ```
+    /// is equal
+    /// ```js
+    /// foo && foo.bar
```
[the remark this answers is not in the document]
denbezrukov: Yes, you're right.
It's about `(foo && foo.bar) == (foo && foo.bar)`. 
I'll change it to:
    /// ```js
    /// (foo && foo.bar) == (foo && foo.bar)
    /// ```

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: For logical _**and**_ chains:
https://github.com/denbezrukov/tools/blob/5a06890fd2efd821660d9c871c940825111b1c49/crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs#L428-L461

Let's say that we have `foo && foo.bar && foo.bar.baz && foo.bar.baz.buzz`.
We iterate over logical `&&` expressions. 
Collect the chains, the head of which is on the right.
Try to get logical and **parent** and also collect a chain with the right part.

1. `... && foo.bar`:
Collect a right part: `[foo, bar]`.
Get parent: `.. foo.bar && foo.bar.baz`.
Collect a parent right part: `[foo, bar, baz]`.
The right part is a sub-chain of the parent right part.  Skip it.

2. `... && foo.bar && foo.bar.baz`
Collect a right part: `[foo, bar, baz]`.
Get parent: `.. foo.bar.baz && foo.bar.baz.buzz`.
Collect a parent right part: `[foo, bar, baz, buzz]`.
The right part is a sub-chain of the parent right part.  Skip it.

3. `... && foo.bar.baz && foo.bar.baz.buzz`
Collect a right part: `[foo, bar, baz, buzz]`.
Get parent: There is no logical and parent.
This is our head.


I see that there are some repetitions.
I guess that the better way is finding the first chain and go down or up and somehow mark sub-chains (expressions) that they are parts of another chain. And use this mark to skip all conditions for sub-chain in a run phase. 
But I'm not sure if it's possible to implement it.

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -194,6 +196,43 @@ impl JsBinaryExpression {
             Ok(T![>] | T![<] | T![>=] | T![<=] | T![==] | T![===] | T![!=] | T![!==])
         )
     }
+
+    /// Whether this is a comparison operation similar to the optional chain
+    /// ```js
+    /// foo === undefined;
+    /// foo == undefined;
+    /// foo === null;
+    /// foo == null;
```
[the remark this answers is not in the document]
denbezrukov: I agree, 
We can make a relative issue for `resolve` and after merge add more test cases for the rule. (:

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: What do you think about such test cases? What comments should be retained?  

Initial: `foo && foo.bar && foo/*1*/./*2*/bar/*3*/./*4*/baz /*5*/;`
Result: `foo/*1*/?./*2*/bar/*3*/?./*4*/baz /*5*/;`

Initial: `foo && foo[bar] && foo/*1*/[/*2*/bar/*3*/]/*4*/[/*5*/baz /*6*/];`
Result: `foo/* 1 */?.[/* 2 */bar/* 3 */]/*  4 */?.[/* 5 */baz /* 6 */];`

Initial: `((/*1*/foo/*2*/ || {})/*3*/./*4*/bar()/*5*/ || {})/*6*/.baz/*7*/;` 
Result: ???

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: Renamed. 
Thanks.

crates/rome_js_syntax/src/expr_ext.rs
MichaReiser: I would find it interesting to hear @denbezrukov opinion on `resolve` as they have the most insight if and how it might be useful in this context. Getting additional feedback would be helpful to evaluate the usefulness of `resolve`.

denbezrukov: I agree that it can be useful. 
Do I understand correctly semantic meaning of the `resolve` function? 
It returns a expression which will evaluate at runtime?

crates/rome_js_syntax/src/expr_ext.rs
```diff
@@ -446,6 +489,103 @@ impl JsRegexLiteralExpression {
     }
 }
 
+impl JsAnyExpression {
+    /// Try to extract non `JsParenthesizedExpression` from `JsAnyExpression`
+    pub fn omit_parentheses(self) -> JsAnyExpression {
+        let first = self
+            .as_js_parenthesized_expression()
+            .and_then(|expression| expression.expression().ok());
```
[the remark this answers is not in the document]
denbezrukov: There is `.unwrap_or(self)` at the end of the function. For non parenthesized expressions function returns `self`.

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,858 @@
...
+            markup! {
+                "Change to an optional chain."
+            }
+            .to_owned(),
+        ))
+    }
+
+    fn action(ctx: &RuleContext<Self>, state: &Self::State) -> Option<JsRuleAction> {
+        match state {
+            UseOptionalChainState::LogicalAnd(optional_chain_expression_nodes) => {
```
denbezrukov, line 146: @MichaReiser 
What do you think about using here `SyntaxRewriter`? Can it help here? Or is it the same?

MichaReiser: The `BatchMutation` api is generally more convenient to use and it helps you when dealing with trivia. `SyntaxRewriter` is a more blunt tool but can be useful if you have a recursive transform. Let's say you have

```
(a + (b + c))
```

and you want to remove the parenthesized expressions. 

A naive approach (which I implemented first :D) is to have the following batch mutation calls

```
mutation.replace_node(`(b + c)`, `b + c`);
mutation.replace_node(`(a + (b + c))`, `a + (b + c)`);
```

The problem is that the second mutation overrides the changes of the first mutation. This is where the `SyntaxRewriter` can come in handy because it gives you access to the rewritten inner node. 

The other benefit of `BatchMutation` over `SyntaxRewriter` is that you can control how you want to traverse the tree where the `SyntaxRewriter` is rather specific about it (normal visitor).

denbezrukov: Interesting,
Actually I have the same problem.
I need to replace an operator token to an optional one.
Let's say we have
``
a.b.c
``
And we want to have
``
a?.b?.c
``

And I solved it in this PR by constructing a new expression with an optional token and replace the old expression in the next parent node.
1. `a.b` -> `a?.b`.
2. replace `a.b` with `a?.b` in `a.b.c`.
3. `a?.b.c` -> `a?.b?.c`.

Is there a way to make it better? (:

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: https://github.com/rome/tools/pull/3085/files#diff-5fe12e15748e57b059ea8ab0e273380f8442306af2df9d36bd526954ca479133R4598-R4651

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
MichaReiser: I don't know what other precedence there is in the analyzer. @ematipico may know better. Delaying the work sounds interesting to me. My main concern is around what happens if I apply the fix on a node that has errors. 


```
node && node.test && node.test.
```

Note, the missing member in the last static member expression. 

The behaviour I would expect is that either:

* The linter doesn't flag the expression (or only the first chain) because it contains a syntax error
* The linter does flag the expression and is capable to perform the fix. 

What we should avoid is that the linter shows the suggestion but then fails to apply the fix because of the syntax error.

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
MichaReiser: I think ideally, the `BatchMutation` API would be able to recognize this and merge the changes. For now, I would say if it works, it works ;) CC: @xunilrj

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,858 @@
...
+            if main_value_token.token_text_trimmed() != branch_value_token.token_text_trimmed() {
+                return Ok(LogicalAndChainOrdering::Different);
+            }
+        }
+
+        Ok(chain_ordering)
+    }
+
+    /// This function returns a list of `JsAnyExpression` which we need to transform into an option chain expression.
+    fn optional_chain_expression_nodes(mut self) -> Option<VecDeque<JsAnyExpression>> {
```
MichaReiser, line 540: I recommend you returning `None` in the case the list is empty since you already have the `Option` around the vec. This will simplify the checking inside the rule.

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: Do you mean that we also should take comments from `foo/*comment here*/ && foo/*comment here*/.bar && ...`. And place them in a result chain?
I haven't thought about that. Need to check it.

MichaReiser (approved):
This is an awesome PR. Let me know what feedback you plan to address. I think we're otherwise good to merge this. We can still do further improvement in subsequent PRs

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: Fixed. Thanks.

denbezrukov:
@MichaReiser Thank you!

- I'd like to look closer a `node && node.test && node.test.` case. I think that we can apply fix only for the first chain. Because now it  show a warning but then fails to apply the fix. :(
- Need to think about comments that belong to nodes which are removed after fix. I believe that we can collect them and add to the final chain.  
`foo /*1*/ && /*2*/ foo/*3*/.bar && foo.bar.baz` -> `/*1*/ /*2*/ /*3*/ foo?.bar?.baz`
- It would be great to resolve other conversations. Especially about `BatchMutation` .

MichaReiser:
> @MichaReiser Thank you!
> 
>     * I'd like to look closer a `node && node.test && node.test.` case. Because now it  show a warning but then fails to apply the fix. :( I think that we can apply fix only for the first chain. Is it possible to create test cases with syntax errors? It seems that impossible:
>       https://github.com/rome/tools/blob/5c1ca0920eaf88a4c363676dcaa1c81ac120e296/crates/rome_js_analyze/tests/spec_tests.rs#L167-L169
>       
>       May it's better not to show any suggestion if we have syntax errors.
> 
>     * Need to think about comments that belong to nodes which are removed after fix. I believe that we can collect them and add to the final chain.
>       `foo /*1*/ && /*2*/ foo/*3*/.bar && foo.bar.baz` -> `/*1*/ /*2*/ /*3*/ foo?.bar?.baz`
> 
>     * It would be great to resolve other conversations. Especially about `BatchMutation` .

@ematipico @leops what's our recommended way to deal with syntax errors in analyzers?

leops:
In general the lint rules should not emit any signal (diagnostic or code action) on nodes that contain syntax errors. In many case this is a simple as bailing out on syntax errors using `.ok()?` on `SyntaxResult`, but it may require some additional processing to detect syntax errors in "related nodes" for the more complex cases (for instance the `noDeadCode` rule tries to suppress warnings for the entire function if it finds a syntax error in a control flow statement)

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
leops: In theory changing `a.b.c` to `a?.b?.c` should work using a `BatchMutation` with two `replace_token` operations, if it doesn't it's probably a bug

denbezrukov: > In theory changing `a.b.c` to `a?.b?.c` should work using a `BatchMutation` with two `replace_token` operations, if it doesn't it's probably a bug

Yes, you're right that it works. But after this we need to remove the head of the chain. 
`a && a.b && a.b.c` -> `a?.b?.b`

I tried to write something like:

```
for expression in optional_chain_expression_nodes {
      mutation.replace_token(
          expression .operator_token().ok()?,
          next_operator_token,
      );
}

let logical = ctx.query();
let right = logical.right().ok()?;

mutation.replace_node(JsAnyExpression::from(logical.clone()), right);

return Some(JsRuleAction {
    category: ActionCategory::QuickFix,
    applicability: Applicability::MaybeIncorrect,
    message: markup! { "Change to an optional chain." }.to_owned(),
    mutation,
});
```
But it doesn't work =(

xunilrj: This PR will probably fix these limitations: https://github.com/rome/tools/pull/3148

ematipico:
@denbezrukov What's the status of the PR? Just asking because I need to organize the next release.

denbezrukov:
> @denbezrukov What's the status of the PR? Just asking because I need to organize the next release.

I guess the main part is finished. There are questions about `BatchMutation` and comments after fixing. I hope we can fix them in other PRs. If the reviews are happy with the answers and there are no comments on the code, then it's ready to merge. 

To sum up:
1. `BatchMutation` - https://github.com/rome/tools/pull/3148
2. `foo /*1*/ && /*2*/ foo/*3*/.bar && foo.bar.baz`  what should we do with comments of the nodes which will be removed?
This question is important for every lint rule that removes nodes during fix phase. May be we can handle it in `BatchMutation`?

ematipico:
Thank you! I will leave it open for few days once we stabilize our last release and then merge it. Cheers

denbezrukov:
> Thank you! I will leave it open for few days once we stabilize our last release and then merge it. Cheers

Great!

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
```diff
@@ -0,0 +1,862 @@
...
+    /// ```js
+    /// foo || foo.bar;
+    ///```
+    ///
+    /// ```js
+    /// foo["some long"] && foo["some long string"].baz
+    ///```
+    ///
+    pub(crate) UseOptionalChain {
+        version: "0.9.0",
```
ematipico, line 77: This one now should be `0.10.0` now. Once updated, `cargo lintdoc` should be run again.

leops (approved):

crates/rome_js_analyze/src/analyzers/js/use_optional_chain.rs
denbezrukov: Done (:

ematipico:
Thank you @denbezrukov ! This was another great contribution! Thank you for your help!

PR #3086, opened 2022-08-19
feat(rome_js_semantic): semantic model is_exported support for type alias, enum, interface

xunilrj:
## Summary

Address some of the issues of https://github.com/rome/tools/pull/3076

We should be able to remove: 

## Test Plan

```
> cargo test -p rome_js_analyze -- ok_semantic_model_is_exported
```

### DIFF
--- a/crates/rome_js_semantic/src/events.rs
+++ b/crates/rome_js_semantic/src/events.rs
@@ -6,7 +6,7 @@ use rome_js_syntax::{
     JsAnyAssignment, JsAnyAssignmentPattern, JsAssignmentExpression, JsForVariableDeclaration,
     JsIdentifierAssignment, JsIdentifierBinding, JsLanguage, JsReferenceIdentifier, JsSyntaxKind,
     JsSyntaxNode, JsSyntaxToken, JsVariableDeclaration, JsVariableDeclarator,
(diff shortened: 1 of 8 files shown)


crates/rome_js_semantic/src/events.rs
```diff
@@ -272,17 +272,33 @@ impl SemanticEventExtractor {
         Some(is_var)
     }
 
-    fn enter_js_identifier_binding(&mut self, node: &JsSyntaxNode) -> Option<()> {
-        debug_assert!(matches!(node.kind(), JsSyntaxKind::JS_IDENTIFIER_BINDING));
+    fn enter_identifier_binding(&mut self, node: &JsSyntaxNode) -> Option<()> {
+        use JsSyntaxKind::*;
+        debug_assert!(matches!(
+            node.kind(),
+            JS_IDENTIFIER_BINDING | TS_IDENTIFIER_BINDING
```
ematipico, line 279: Can we insert a message here that explains us how to fix this assertion in case it fails?

ematipico (approved):

crates/rome_js_semantic/src/events.rs
xunilrj: Done.

PR #3091, opened 2022-08-21
feat(rome_rowan): `with_*_trivia_pieces`

MichaReiser:
This PR adds the `with_leading_trivia_pieces` and `with_trailing_trivia_pieces` methods to `SyntaxToken` that accept a `SyntaxTriviaPiece` iterator instead of an iterator over `(kind, &str)` tuple.

The motivation for adding this overload is because it is often necessary to merge trivia from different tokens. Unfortunately, just doing

```
token.with_leading_trivia(existing_pieces.map(|piece| (piece.kind(), piece.text()))
```

It does not work because `text()` returns data owned by `piece` which goes out of scope as soon as the `map closure` returns.

 The workaround thus far has been to collect the pieces and map over the piece references which comes at the cost of an additional vector allocation. Avoiding the extra allocation will be more important in the formatter where this method will soon be called for every parenthesized expression in a program.

## Test

I updated the mutation and batch mutation to use the new API and all analyzer tests continue to run.


MichaReiser:
!bench_analyzer

xunilrj (approved):

PR #3092, opened 2022-08-22
feat(rome_js_formatter): Preprocess AST before formatting

MichaReiser:
## Summary

This PR adds a pre-processing step to the formatter that can be used to bring the tree into a shape that better suits the formatter and adds support for:

* Removing unnecessary parentheses of logical expressions with a nested logical expression in the right hand side 
* Preserving parentheses if a node has a leading type cast comment

### Rewriter

The JavaScript uses the pre-processing to:

* Remove any parenthesized nodes except: 
  * If the expression node has a syntax error (missing required child or has a skipped token trivia)
  * If the inner expression is an unknown node
  * If the expression has a closure type-cast comment 
* Re-balances logical expressions: `a && (b && c)` -> `(a && b) && c` so that it works nicely with the binary like expression flattening

This logic is implemented inside of the new [`JsFormatSyntaxRewriter`](https://github.com/rome/tools/blob/8842144d3922e86191c0ac451ce877f2b490d6bf/crates/rome_js_formatter/src/syntax_rewriter.rs). 

### Source Map

The second significant part of this PR is the need for a [source map](https://github.com/rome/tools/blob/8842144d3922e86191c0ac451ce877f2b490d6bf/crates/rome_formatter/src/source_map.rs) to map printer markers to the original source positions and resolve the source text for a transformed node

For example:

```javascript
// Source			// Transformed
(a + b)				a + b
```

Resolving the source text of `a + b` should return `(a + b)` to ensure that the verbatim output (used when the node can't be formatted or if formatting is suppressed) doesn't drop the parentheses. 

Check out the documentation of [source map](https://github.com/rome/tools/blob/8842144d3922e86191c0ac451ce877f2b490d6bf/crates/rome_formatter/src/source_map.rs) for a more in detail explanation. 

## This looks complicated

Yes, this PR introduces some complexity by adding a source map and transformation step to the formatter. In my view, the simplification that the transformation brings by no longer having to use `resolve` and `resolve_parent` inside specific formatter outweighs the added complexity that is local to the `rome_formatter` only. I see it as a trade of added complexity in the infrastructure in favor of simplifying individual formatters that tend to be complicated anyway. 

## This PR is huge

Yes, it is but most changes are 
* Removing the `ExpressionNode` and `AssignmentNode` implementations
* Removing `resolve` and `resolve_parent` calls. 
* Updated snapshot

I recommend you focusing on reviewing

* The [source map implementation](https://github.com/rome/tools/blob/8842144d3922e86191c0ac451ce877f2b490d6bf/crates/rome_formatter/src/source_map.rs)
* The [syntax rewrite](https://github.com/rome/tools/blob/8842144d3922e86191c0ac451ce877f2b490d6bf/crates/rome_js_formatter/src/syntax_rewriter.rs)

## Performance

```
group                                    main                                    parens
formatter/checker.ts                     1.00   226.6±16.42ms    11.5 MB/sec     1.09    247.1±2.08ms    10.5 MB/sec
formatter/compiler.js                    1.00   144.6±10.28ms     7.2 MB/sec     1.11   160.9±13.78ms     6.5 MB/sec
formatter/d3.min.js                      1.00    106.1±1.00ms     2.5 MB/sec     1.26   134.0±12.17ms  2003.1 KB/sec
formatter/dojo.js                        1.00      7.7±0.58ms     8.9 MB/sec     1.08      8.3±0.62ms     8.2 MB/sec
formatter/ios.d.ts                       1.00   167.6±15.02ms    11.1 MB/sec     1.03   173.1±18.89ms    10.8 MB/sec
formatter/jquery.min.js                  1.00     32.5±2.36ms     2.5 MB/sec     1.02     33.2±0.61ms     2.5 MB/sec
formatter/math.js                        1.00   243.6±15.86ms     2.7 MB/sec     1.04   252.7±12.13ms     2.6 MB/sec
formatter/parser.ts                      1.00      5.3±0.43ms     9.1 MB/sec     1.06      5.6±0.46ms     8.7 MB/sec
formatter/pixi.min.js                    1.00    121.9±1.22ms     3.6 MB/sec     1.13    137.6±3.15ms     3.2 MB/sec
formatter/react-dom.production.min.js    1.00     37.3±2.83ms     3.1 MB/sec     1.18     44.1±3.70ms     2.6 MB/sec
formatter/react.production.min.js        1.00  1795.7±12.33µs     3.4 MB/sec     1.14      2.0±0.00ms     3.0 MB/sec
formatter/router.ts                      1.01      4.2±0.35ms    14.7 MB/sec     1.00      4.1±0.05ms    14.8 MB/sec
formatter/tex-chtml-full.js              1.00   325.4±19.21ms     2.8 MB/sec     1.03   335.3±23.16ms     2.7 MB/sec
formatter/three.min.js                   1.00    137.6±1.49ms     4.3 MB/sec     1.12    154.0±1.51ms     3.8 MB/sec
formatter/typescript.js                  1.00   975.5±66.09ms     9.7 MB/sec     1.06  1038.4±75.09ms     9.1 MB/sec
formatter/vue.global.prod.js             1.00     48.6±3.57ms     2.5 MB/sec     1.10     53.6±3.78ms     2.2 MB/sec

```

Formatting `rome-classic`:

* `main`: Formatted 3959 files in 216ms
* `this pr`: Formatted 3959 files in 236ms

The main bottleneck for formatting remains IO

## Test Plan

**Average compatibility**: 82.68 -> 83.70
**Compatible lines**: 79.75 -> 80.79


MichaReiser:
!bench_formatter

MichaReiser:
@ematipico as discussed offline. I added additional documentation and added more tests to the `SyntaxRewriter` and `SourceMap`

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+    /// Returns the source text of the trimmed range of `node`.
+    pub fn trimmed_source_text<L: Language>(&self, node: &SyntaxNode<L>) -> SyntaxNodeText {
+        let range = self.trimmed_source_range(node);
+        self.source_text.slice(range)
+    }
+
+    #[cfg(test)]
+    fn trimmed_source_text_from_transformed_range(&self, range: TextRange) -> SyntaxNodeText {
+        let range = self.trimmed_source_range_from_transformed_range(range);
+        self.source_text.slice(range)
```
ematipico, line 130: Do you think we're able to suggest how a developer should fix this issue if it pops up?

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+        };
+
+        self.source_offset_with_range(transformed_offset, position, range)
+    }
+
+    fn source_offset_with_range(
+        &self,
+        transformed_offset: TextSize,
+        position: RangePosition,
+        deleted_range: Option<&DeletedRange>,
```
ematipico, line 156: This seems to be an important function, would it be possible to expand some concepts?
- for example, maybe we can explain what's `printed`. Like, we don't need to call the variable, but what's the variable - the formatted content?;
- what are these "transformed positions"? 

Maybe some exmaple/design might help to explain why this mapping is needed.

ematipico: About the design decision, why the mutation on `Printed` occurs here inside `TransformSourceMap` and not the other way around? e.g. Having the printer to accept a reference to `TransformSourceMap` and then do the mapping

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+        for range in ranges {
+            *range = self.source_range(*range)
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+struct TrimmedNodeRangeMapping {
+    /// The original trimmed range of the node.
+    ///
```
ematipico, line 263:
```suggestion
/// A transformation that removes the parentheses from the above expression, removes the ranges `0..1` (`(` token)
```

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+/// (a + b)
+/// ```
+///
+/// A transform that removes the parentheses from the above expression removes the ranges `0..1` (`(` token)
+/// and `6..7` (`)` token) and the source map creates one [DeletedRange] for each:
+///
+/// ```text
+/// DeletedRange {
+///     source_range: 0..1,
+///     transformed_offset: 0,
```
ematipico, line 301: The example is oudate: `total_length_preceding_deleted_ranges`

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+/// },
+/// DeletedRange {
+///     source_range: 6..7,
+///     transformed_offset: 1,
+/// }
+/// ```
+///
+/// The first range indicates that the range `0..1` for the `(` token has been removed. The second range
+/// indicates that the range `6..7` for the `)` has been removed and it stores that, up to this point,
```
ematipico, line 310:
```suggestion
/// indicates that the range `6..7` for the character `)` has been removed and it stores that, up to this point,
```

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+    /// The range in the source document of the bytes that have been omitted from the transformed document.
+    source_range: TextRange,
+
+    /// The accumulated count of all removed bytes up to (but not including) the start of this range.
+    total_length_preceding_deleted_ranges: TextSize,
+}
+
+impl DeletedRange {
+    fn new(source_range: TextRange, total_length_preceding_deleted_ranges: TextSize) -> Self {
+        debug_assert!(source_range.start() >= total_length_preceding_deleted_ranges);
```
ematipico, line 323: Can we add a string saying why there's an error and how a developer can fix it?

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
...
+///
+/// * `a` -> `a`: Should not include the leading `(`
+/// * `b` -> `b`: Should not include the trailing `)`
+/// * `a + b` -> `(a + b)`: Should include the leading `(` and trailing `)`.
+/// * `a + b + c + d -> `(a + b) + (c + d)`: Should include the leading `(` and `)` trailing `)` because the expression statement
+///   fully encloses the `a + b` and `c + d` nodes.
+///
+/// This is why the source map also tracks the mapped trimmed ranges for every node.
+#[derive(Debug, Clone)]
+pub struct TransformSourceMap {
```
ematipico, line 45: This is a great documentation, and maybe we should make it as **module comment**?

If a developer imports this struct, they might be more interested in **how to use the struct**. So maybe some examples are more approriate. Considering that this is a generic API, the need to see some examples of usage is important. Same thing for the functions exposed by struct.

crates/rome_formatter/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub mod prelude;
 #[cfg(debug_assertions)]
 pub mod printed_tokens;
 pub mod printer;
+mod source_map;
```
ematipico, line 38: If we make this module public, we will able to publish the module documentation around the source maps of our formatter.

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,598 @@
+use crate::{Printed, SourceMarker, TextRange};
```
ematipico, line 1: Quick question around this module. If it's here, it means that it's generic, but by looking at the explanation of the documentation and the tests, it seems to orbit around parenthesis, which is something tailored around JavaScript.

While it makes sense to have examples for JavaScript because that's where we have our primary usage, do you think we should revisit the tests and show that it works with every token? And maybe explain, somewhere, that this is a generic module?

crates/rome_js_formatter/src/lib.rs
```diff
@@ -492,6 +505,16 @@ impl FormatLanguage for JsFormatLanguage {
         JsCommentStyle
     }
 
+    fn transform(
+        &self,
+        root: &SyntaxNode<Self::SyntaxLanguage>,
+    ) -> Option<(SyntaxNode<Self::SyntaxLanguage>, TransformSourceMap)> {
+        let mut rewriter = JsFormatSyntaxRewriter::new(root);
+        let transformed = rewriter.transform(root.clone());
```
ematipico, line 513: It's really weird that a rewriter has a reference to `root`, but then we need to pass it to `transform`.

I may understand that this is needed because `transform` comes from a trait, but is there a way to make it less "strange" and error prone?

ematipico (commented):
Left some questions. I still need to review some stuff after having answer from my questions

crates/rome_js_formatter/src/builders.rs
```diff
@@ -207,38 +208,76 @@ impl Format<JsFormatContext> for FormatVerbatimNode<'_> {
...
                     .first_leading_trivia()
                     .into_iter()
                     .flat_map(|trivia| trivia.pieces())
                     .skip_while(skip_whitespace)
                 {
+                    let trivia_source_range = source_range(f, leading_trivia.text_range());
+
+                    if trivia_source_range.start() >= trimmed_source_range.start() {
+                        break;
+                    }
```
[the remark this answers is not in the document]
MichaReiser: Let's say you have

```javascript
a;

(/* comment */ b + );  
```

which gets transformed to 

```
a
/* comment */ b + ;
```

Formatting the binary expression `b + ` will fail and the formatter tries to print it in verbatim. To do this, it will get the inner text of the binary expression `b +` which returns `(/* comment */ b + )`. However, `/* comment */` is also a leading trivia of `b`. 

This check prevents that the formatter writes any trivia that overlap with text that is already part of it's `verbatim` representation in the source document.

crates/rome_js_formatter/src/builders.rs
```diff
@@ -207,38 +208,76 @@ impl Format<JsFormatContext> for FormatVerbatimNode<'_> {
...
+                    let trailing_source_range = source_range(f, trailing.text_range());
+                    let is_in_trimmed_range =
+                        trailing_source_range.start() < trimmed_source_range.end();
+
+                    if is_whitespace || is_in_trimmed_range {
+                        trailing_back.next();
+                    } else {
+                        break;
+                    }
+                }
```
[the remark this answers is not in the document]
MichaReiser: The one thing I added here is the same as for the leading trivia. It's here to make sure that the formatter doesn't write any trailing trivia that overlaps with the verbatim source from the original document. 

The other logic is the same as the one before this PR but avoids an additional vector allocation by using `peek` and `rev`.

crates/rome_js_formatter/src/syntax_rewriter.rs
```diff
@@ -0,0 +1,806 @@
...
+        JsArrayExpression, JsBinaryExpression, JsExpressionStatement, JsIdentifierExpression,
+        JsLogicalExpression, JsSequenceExpression, JsStringLiteralExpression, JsSyntaxNode,
+    };
+    use rome_rowan::{AstNode, SyntaxRewriter, TextSize};
+
+    #[test]
+    fn rebalances_logical_expressions() {
+        let root = parse_module("a && (b && c)", 0).syntax();
+
+        let transformed = JsFormatSyntaxRewriter::new(&root).transform(root.clone());
```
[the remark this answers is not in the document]
MichaReiser: I agree, that this is a bit misleading. The `transform` function is inherited from the `SyntaxRewriter` trait. 

Passing a different root than what you pass to `transform` will yield a broken source map. 

I don't consider this being problematic as `JsFormatSyntaxRewriter` is a crate internal type, and its usage is abstracted by `JsFormatLanguage`.

crates/rome_formatter/src/source_map.rs
MichaReiser: Do you have a recommendation? In my view, it's pointing out the potential problem:

* wrong source map
* source range instead of a transformed range (what I did, and is why I added the check)

crates/rome_formatter/src/source_map.rs
MichaReiser: > About the design decision, why the mutation on Printed occurs here inside TransformSourceMap and not the other way around? e.g. Having the printer to accept a reference to TransformSourceMap and then do the mapping

The reason why it's on the `SourceMap` is so that the implementation has access to all internal fields and can use the most efficient algorithm. I would also consider `SourceMap` a more internal concept than `Printed` that even makes it into the Node JS API.

crates/rome_formatter/src/source_map.rs
MichaReiser: > This is a great documentation, and maybe we should make it as module comment?

Thank you. I would prefer keeping it on the source map because it explains how the data is arranged internally and for what reasons. 

> If a developer imports this struct, they might be more interested in how to use the struct. So maybe some examples are more approriate. Considering that this is a generic API, the need to see some examples of usage is important. Same thing for the functions exposed by the struct.

I considered adding examples when I added the additional tests but as you can see, the tests get extremely length and writing such code in docblocks is no fun to write nor read. I would recommend consulting the tests in that case.

crates/rome_formatter/src/source_map.rs
MichaReiser: I wanted to use concrete test cases and parenthesized expressions are a problem that exist in almost all languages (and languages that don't have them may not need to use the rewriter nor the source map). I didn't want to use arbitrary examples because that may be more complicated to understand than a concrete use case like parentheses.

We can extend the source map functionality when new use cases pop up but, in my view, the version we have today should be sufficient because transforms should only ever delete tokens and that's the use case the source map supports today.

I don't mind adding an additional test if you have a very concrete example with a specific reason why you would find it useful. Writing these tests takes a lot of time (The tests are about as much code as the implementation)

crates/rome_js_formatter/src/lib.rs
MichaReiser: We could not expose the `JsFormatSyntaxRewriter` and hide it behind it a function that takes care of it. In my view, this is what this method is doing and there should be no other place in the code base that has to call this function.

crates/rome_formatter/src/lib.rs
MichaReiser: Screenshot 2022-08-26 at 14 20 17
I just checked and the structs are listed even if the module isn't public

crates/rome_formatter/src/source_map.rs
ematipico: If I understood correctly, this source map is about when we delete tokens. If so, any example where token are deleted should fit the bill? For example removing a trailing comma (if we want to delete it).

Because if we there aren't any examples, it doesn't make much sense to me to make it generic.

crates/rome_js_formatter/src/lib.rs
ematipico: It's really error prone this method. Is it possible to have some debug assertion where we check if the first `&root` is equal to the second `root.clone()`.

Because it seems that nothing might prevent us from doing:

```rs
let mut rewriter = JsFormatSyntaxRewriter::new(&first_node);
let transformed = rewriter.transform(second_node.clone());
```

And we should be able to catch these wrong usages.

crates/rome_formatter/src/source_map.rs
ematipico: Please remember that in the past, in our of our last refactors, I was quite slow with the new APIs, but at least there was examples in the documentation that helped me. If there aren't examples, I would be very slow.

crates/rome_js_formatter/src/syntax_rewriter.rs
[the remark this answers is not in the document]
ematipico: As suggested [here](https://github.com/rome/tools/pull/3092#discussion_r956012573), maybe we could have some debug assertion. If the source maps is broken, we should be able to tell the developer how to fix it.

crates/rome_formatter/src/source_map.rs
ematipico: Unfortunately I can't understand the flow of the code... but if it's there, it means it's triggered from some public API. Maybe we can point them to there and explain them what the  cause might be.

crates/rome_js_formatter/src/syntax_rewriter.rs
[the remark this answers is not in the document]
MichaReiser: I resolve this here to centralize the conversation.

crates/rome_formatter/src/source_map.rs
MichaReiser: I'm sorry but I don't know how to come up with a more specific error. The message is thrown at the entry point from the public API your calling and the problem is with the arguments the caller is passing (or they mixed up the source maps, which is what the message is explaining).

crates/rome_formatter/src/source_map.rs
MichaReiser: I'm not saying that examples aren't good. What I'm saying is that it's hard to write examples here because of how verbose the test setup is.

crates/rome_formatter/src/source_map.rs
MichaReiser: That's correct. But why does deleting parentheses doesn't fit that bill which is a concrete use case. Deleting a trailing comment isn't different from that, except that it is an example for which I would not advice to use the pre-processing for.

crates/rome_js_formatter/src/lib.rs
MichaReiser: That's not possible because I can't override the `transform` function (because I won't be able to call the base implementation). I can wrap it for you in a function but feel this is over focusing on an internal implementation detail.

ematipico: Doesn't seem an internal implementation to me. Actually it seems public API. 

Is the snippet that I pasted something that might happen or the compiler is able to catch it? If not, then I think these API should be revisited somehow to prevent these mistakes

crates/rome_formatter/src/source_map.rs
ematipico: Oh, I didn't say that deleting parentheses doesn't fit the bill. I just wanted to confirm if understood correctly the module

MichaReiser: I added a section to the source map documentation mentioning that the following section are using parentheses as a concrete example but that the module isn't limited to parentheses (I haven't mentioned JavaScript because many languages have parenthesized nodes)

crates/rome_formatter/src/source_map.rs
MichaReiser: To extend a bit on my previous answer. This error isn't an error that a user should ever see. If it happens that this error gets triggered then Rome should show a message similar to `This is a bug in Rome. Please create an issue.` The specific message will help us to identify the exact line that's triggering it which is far better than the out of range exception that would otherwise be thrown when using the returned range to slice the source text. 

The situation is different if you as an engineer encounter this message but I would expect you to be somewhat familiar with the source map then because:

* You're either changing the source map implementation and you messed up some indices (happened to me more than I feel comfortable to admit :D) 
* You're using the source map and you're using it incorrectly. The message should give you a hint what to look out for. I think the message we have now is covering the main reasons:
  * You passed a range from the untransformed source tree instead of a transformed index
  * You passed a range from a tree that isn't the tree for which the source map has been created

I consider this sufficient to guide the developer in the next debugging step and helped me to fix at least one bug in the tests quickly. Overall, having the assertion here is significantly better even if the message may not be perfect because the alternative is that you get a overflow exception the first time the source map implementation calls `transformed_start` and I can understand that a dev unfamiliar (or even familiar with the source map) is at a loss and worried that the source map is broken while it is just the case that they're using it incorrectly. 

I hope this helps you understand why I think it is OK for us to move forward with the message we have today and we can improve on the message in the future if someone runs into it, doesn't understand what the issue was but then figures it out and is then able to clarify the message for others.

crates/rome_js_formatter/src/lib.rs
MichaReiser: The snipped you posted is something that is possible and nothing the compiler will catch. But the snipped by itself isn't necessarily incorrect.

The snipped is only incorrect if it happens that `first_node` isn't an ancestor of `second_node`. 

It would be possible to implement a debug assertion inside of `visit_node` that is called for every node in the program. The assertion would have to check if any of `node`'s ancestor is the node passed to `new`. I'm reluctant from adding this check for three reasons:

* It adds a significant perf penalty in debug builds that will slow down our test execution
* Passing the incorrect root will result in an invalid source map. Using it will very quickly panic because of the other debug assertions I added. In that sense, there are safeguards in place further down. 
* But the main reason is that the `JsFormatSyntaxRewriter` is rarely used. I would see good reasons to improve the API if it is some commonly used API like `Format` that is used all over the code base. This is different for `JsFormatSyntaxRewriter` where there is only a single place where it is useful to be called, and that's here in the `JsFormatLanguage`. The risk to call it incorrectly in this function is close to zero because you only happen to have access to one node. While it's possible I think you would almost have to do so intentionally. 

I mitigated the problem now in sofar that I made `JsFormatSyntaxRewriter` private to the `syntax_rewriter` module that no exposes a `transform(JsSyntaxNode)` function. This isn't solving the underlining issue as the incorrect use is still possible in the `syntax_rewriter` module but someone changing that module should have a basic understanding of the rewriter.

crates/rome_js_formatter/src/syntax_rewriter.rs
```diff
@@ -0,0 +1,864 @@
...
+}
+
+/// Chain iterator that chains two iterators over syntax trivia together.
+///
+/// This is the same as Rust's [Chain] iterator but implements [ExactSizeIterator].
+/// Rust doesn't implement [ExactSizeIterator] because adding the sizes of both pieces may overflow.
+///
+/// Implementing [ExactSizeIterator] in our case is safe because our ranges are limited to [u32], which in turn
+/// guarantees that a tree can never have more than 2^32 tokens or pieces and adding a `u32 + u32` is safely in the range
+/// of a `usize`.
```
leops, line 407: The assumption that `u32 + u32 < usize` doesn't hold for WebAssembly where `usize` is 32-bits, but I don't think having two tokens with more than 2^16 trivia pieces is possible anyway

crates/rome_formatter/src/source_map.rs
```diff
@@ -0,0 +1,613 @@
...
+#[derive(Debug, Clone)]
+pub struct TransformSourceMap {
+    source_text: SyntaxNodeText,
+
+    /// The mappings stored in increasing order
+    deleted_ranges: Vec<DeletedRange>,
+
+    /// Key: Start or end position of node for which the trimmed range should be extended
+    /// Value: The trimmed range.
+    mapped_node_ranges: HashMap<TextSize, TrimmedNodeRangeMapping>,
```
leops, line 60: Depending on how much insertion and lookup this `HashMap` gets, it might improve performances to use an `FxHashMap` or `BTreeMap` here as these containers are very efficient for small numeric keys

MichaReiser: The map contains `count(ParenthesizedNode) * 2` entries. Probably not too many but using `FxHashMap` might be a good idea.

crates/rome_js_formatter/src/syntax_rewriter.rs
MichaReiser: Uff, I wasn't aware of this. I'll update the documentation

MichaReiser: I changed it to be safe because a document can not have more than 2^32 trivia pieces because a document is limited to a length of 2^32 and empty pieces aren't allowed.

MichaReiser:
!bench_formatter

ematipico:
> 
> ```
> formatter/checker.ts                     1.00    362.4±3.53ms     7.2 MB/sec    1.33    482.9±3.00ms     5.4 MB/sec
> formatter/compiler.js                    1.00    249.2±1.47ms     4.2 MB/sec    1.11    276.5±1.70ms     3.8 MB/sec
> formatter/d3.min.js                      1.00    202.5±1.45ms  1325.6 KB/sec    1.17    236.4±2.07ms  1135.4 KB/sec
> formatter/dojo.js                        1.00     13.0±0.02ms     5.3 MB/sec    1.10     14.3±0.02ms     4.8 MB/sec
> formatter/ios.d.ts                       1.00    233.7±0.88ms     8.0 MB/sec    1.24    290.8±1.19ms     6.4 MB/sec
> formatter/jquery.min.js                  1.00     55.1±0.91ms  1534.8 KB/sec    1.09     60.4±1.00ms  1401.8 KB/sec
> formatter/math.js                        1.00    381.1±4.65ms  1740.0 KB/sec    1.25    477.6±4.51ms  1388.4 KB/sec
> formatter/parser.ts                      1.00      8.7±0.01ms     5.6 MB/sec    1.08      9.4±0.02ms     5.2 MB/sec
> formatter/pixi.min.js                    1.00    236.2±1.49ms  1902.5 KB/sec    1.12    263.8±2.12ms  1703.6 KB/sec
> formatter/react-dom.production.min.js    1.00     66.5±1.50ms  1772.1 KB/sec    1.13     74.9±1.01ms  1573.4 KB/sec
> formatter/react.production.min.js        1.00      3.3±0.00ms  1932.7 KB/sec    1.10      3.6±0.00ms  1758.9 KB/sec
> formatter/router.ts                      1.00      6.8±0.01ms     9.1 MB/sec    1.08      7.3±0.03ms     8.4 MB/sec
> formatter/tex-chtml-full.js              1.00    544.4±3.65ms  1714.1 KB/sec    1.12    611.1±4.76ms  1527.0 KB/sec
> formatter/three.min.js                   1.00    265.6±1.94ms     2.2 MB/sec    1.12    297.2±3.00ms  2023.0 KB/sec
> formatter/typescript.js                  1.00   1485.7±4.90ms     6.4 MB/sec    1.24   1842.4±6.40ms     5.2 MB/sec
> formatter/vue.global.prod.js             1.00     89.2±1.18ms  1382.9 KB/sec    1.10     98.4±1.14ms  1254.3 KB/sec
> ```

Is it just me, or do the numbers got slightly higher compared to the previous bench run?

ematipico (approved):

MichaReiser:
I compared the numbers again and it's mainly `checker` and `typescript` that now differs significantly. I manually compared the commits before adding the Fx hash map for those two tests to see if there's a serious regression

```
group                      fx                                     map
-----                      --                                     ---
formatter/checker.ts       1.00   276.1±22.30ms     9.4 MB/sec    1.03   283.1±42.19ms     9.2 MB/sec
formatter/typescript.js    1.02  1079.0±74.32ms     8.8 MB/sec    1.00  1060.6±102.66ms     9.0 MB/sec
```

The numbers are in the normal range which is why I think it's fine to go ahead with this PR to unlock further work. 

As discussed offline with @ematipico There are a few things that we want to discuss offline and I'll create a follow up PR addressing the findings from that meeting.

PR #3094, opened 2022-08-23
docs(website): fix typo on project-config.md

fadiinho:
## Summary

At the current docs the keys are `format`, but when using it in rome.json it causes a crash saying:
```
Error: Rome couldn't load the configuration file, here's why: 
unknown field `format`, expected one of `formatter`, `linter`, `javascript` at line 7 column 10
```
This PR fixes the key to the correct one `formatter`.

## Test Plan

Started the server and it was all ok.


ematipico (approved):
Thank you!

PR #3095, opened 2022-08-23
refactor: use `std::sync::Mutex` instead of `parking_lot::Mutex`

IWANABETHATGUY:
## Summary
1. Part of #3054 
2.  With the standard library Mutex and RwLock being const-constructible it's now possible to use these in the DiffReport of the Prettier test suite

## Test Plan
1. Passing CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1521,6 +1521,7 @@ name = "rome_console"
 version = "0.0.0"
 dependencies = [
  "atty",
+ "lazy_static",
  "rome_markup",
  "schemars",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1500,7 +1500,6 @@ dependencies = [
  "hdrhistogram",
  "insta",
  "lazy_static",
- "parking_lot 0.12.1",
  "pico-args",
  "rayon",
  "rome_bin",
@@ -1522,7 +1521,6 @@ name = "rome_console"
 version = "0.0.0"
 dependencies = [
  "atty",
- "lazy_static",
  "rome_markup",
  "schemars",
  "serde",
@@ -1640,7 +1638,6 @@ dependencies = [
  "ctor",
  "iai",
  "insta",
- "parking_lot 0.12.1",
  "quickcheck",
  "quickcheck_macros",
  "rome_diagnostics",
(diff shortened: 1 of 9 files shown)


ematipico (commented):
Another change that should be part of this PR is removing `lazy_static` around the definition of `METRICS`. 

Now that `Mutex` can be defined as `const`, the usage of `lazy_static` is not needed anymore.

IWANABETHATGUY:
> 

Seems not, `rustc` doesn't allow us to create a const `Mutex` that inner value can't create in compile time.
rustc emit errors below: 
```
cannot call non-const fn `HashMap::<CallsiteKey, Mutex<CallsiteEntry>>::new` in constants
calls in constants are limited to constant functions, tuple structs and tuple variants
```

xunilrj:
!bench_analyzer

xunilrj:
!bench_formatter

ematipico:
> > 
> 
> Seems not, `rustc` doesn't allow us to create a const `Mutex` that inner value can't create in compile time. rustc emit errors below:
> 
> ```
> cannot call non-const fn `HashMap::<CallsiteKey, Mutex<CallsiteEntry>>::new` in constants
> calls in constants are limited to constant functions, tuple structs and tuple variants
> ```

Really difficult to give feedback without seeing the code, but removing only `lazy_static` is not correct, we have to do some refactor.

IWANABETHATGUY:
> pile time. rustc emit errors below:

image

IWANABETHATGUY:
The only way I know to create a hashmap in compile time is `phf`

MichaReiser:
> > > 
> > 
> > 
> > Seems not, `rustc` doesn't allow us to create a const `Mutex` that inner value can't create in compile time. rustc emit errors below:
> > ```
> > cannot call non-const fn `HashMap::<CallsiteKey, Mutex<CallsiteEntry>>::new` in constants
> > calls in constants are limited to constant functions, tuple structs and tuple variants
> > ```
> 
> Really difficult to give feedback without seeing the code, but removing only `lazy_static` is not correct, we have to do some refactor.

The problem is that `RwLock` stores a `HashMap` and `HashMap::new` is not `const`. So you can write `RwLock::new(...)` but there's no way to create the map.

MichaReiser (approved):

PR #3097, opened 2022-08-23
feat(npm): hook `*Conent` APIs to WASM bindings

ematipico:
## Summary

Part of https://github.com/rome/tools/issues/3073

In this PR the Node.js API are hooked to the WASM bindings. 

**This PR only focuses on hooking the `*Content` APIs**. The file system APIs are left out on purpose. For the being also custom configuration left out on purpose, and will be implemented as part of a different PR.

Things that are relevant for the review:
- error handling is left out on purpose until I understand how to handle errors coming from the workspace;
- this PR focuses only on the happy path
- all methods of the `Workspace` of the WASM bindings are synchronous, but there were made asynchronous ( with `await`) because it would allow us to easily swap backend in the future without too many problems.
- I changed the release target of `@rometools/backend-jsonrpc` to `es2020` because we don't need to support old syntax. This has also the benefit that the emitted code is easier to debug
- there's an issue with  emitted types; for now I turned off the error with `@ts-expect-error`. Once fixed, we can remove the suppression

## Test Plan

I updated the tests we had with real expected output

### DIFF
--- a/npm/rome/src/index.ts
+++ b/npm/rome/src/index.ts
@@ -1,3 +1,6 @@
+import { main, Workspace, RomePath } from "@rometools/wasm-nodejs";
+
+
 interface FormatFilesDebugOptions extends FormatFilesOptions {
 	/**
 	 * If `true`, you'll be able to inspect the IR of the formatter
@@ -32,19 +35,25 @@ interface FormatContentOptions {
 }
 
 interface FormatResult {
-	// Not final
+	/**
+	 * The new formatted content
+	 */
 	content: string;
 	// Not final
 	errors: string[];
 }
 
 interface FormatDebugResult {
-	// Not final
+	/**
+	 * The new formatted content
+	 */
 	content: string;
 	// Not final
 	errors: string[];
-	// Available when in debug mode
-	ir: string | null;
+	/**
+	 * The IR emitted by the formatter
+	 */
+	ir: string;
 }
 
 interface ParseOptions {
@@ -80,7 +89,57 @@ function isFormatContentDebug(
 	return options?.debug !== undefined;
 }
 
+interface CurrentFile {
+	version: number, path: RomePath
+}
+
 export class Rome {
+	private workspace: Workspace
+	private currentFile: CurrentFile;
+
+	constructor() {
+		// load the web assembly module
+		main();
+		this.workspace = new Workspace();
+		this.currentFile = {
+			version: 0,
+			path: {
+				path: "",
+				id: 0
+			}
+		};
+	}
+
+	private async getWorkspace(): Promise<Workspace> {
+		return Promise.resolve(this.workspace)
+	}
+
+	/**
+	 * It updates the current file. If `true`, the file was correctly updated.
+	 * If `false`, a new version will be created.
+	 * @param path
+	 * @param workspace
+	 * @private
+	 */
+	private updateCurrentFile(path: RomePath): boolean {
+		if (path.path === this.currentFile.path.path) {
+			// same path, let's just update the version
+			this.currentFile = {
+				version: this.currentFile.version++,
+				...this.currentFile.path
+			}
+			return true;
+		} else {
+			// no same path, let's create a new one
+			this.currentFile = {
+				version: 0,
+				path
+			}
+			return false;
+		}
+	}
+
+
 	async formatFiles(paths: string[]): Promise<FormatResult>;
 	async formatFiles(
 		paths: string[],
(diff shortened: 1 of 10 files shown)


npm/rome/src/index.ts
```diff
@@ -80,7 +88,69 @@ function isFormatContentDebug(
 	return options?.debug !== undefined;
 }
 
+interface CurrentFile {
+	version: number;
+	path: RomePath;
+}
+
 export class Rome {
+	private workspace: Workspace | null;
```
ematipico, line 97: This `null` because creating a `Workspace` using `backend-jsonrpc` is an async operation, which means we can't have a `async` call inside `constructor`.

MichaReiser: I would then recommend using a static factory function instead with a private constructor

```
export class Rome {
  private constructor(private workspace: Workspace)  {
    ..
  }

  async create(): Promise<Rome> {
    return new Rome(await loadWorkspace())
  }
```

npm/rome/src/index.ts
```diff
@@ -80,7 +88,69 @@ function isFormatContentDebug(
...
+	}
+
+	/**
+	 * It updates the current file. If `true`, the file was correctly updated.
+	 * If `false`, a new version will be created.
+	 * @param path
+	 * @param workspace
+	 * @private
+	 */
+	private updateCurrentFile(path: RomePath): boolean {
```
MichaReiser, line 134: Can you explain the need for the `currentFile`? 

I've been bitten many times by state like `currentFile` and I t think we should ideally avoid having it because it may introduce subtle race conditions. 

For example, what if you have and the calls complete out of order?

```
Promise.all([await rome.formatContent(...), await rome.formatContent(...)])
```

npm/rome/src/index.ts
ematipico: You mean a static method? 

```js
import {Rome} from "rome";

const rome = await Rome.create()
```

MichaReiser: Yes sorry... I no longer know how to write JS... I assumed it's static because there's no `self` parameter

npm/rome/src/index.ts
ematipico: > Can you explain the need for the `currentFile`?
> 

It's just an implementation detail for calling `open_file` or `change_file` based on the file path. If it hasn't changed, we call `change_file`. We can remove it but the code for checking the path should stay.

npm/rome/src/index.ts
```diff
@@ -121,17 +188,51 @@ export class Rome {
...
+			id: 1,
+		});
+		if (updated) {
+			await this.workspace.change_file({
+				content,
+				version: this.currentFile.version,
+				path: this.currentFile.path,
+			});
+		} else {
+			await this.workspace.open_file({
```
MichaReiser, line 202: What happens if we call `open_file` multiple times with the same path. It feels unfortunate that the client must remember if the file has already been opened or not if this is information that the workspace already has.

npm/rome/src/index.ts
MichaReiser: Is it necessary that it is a field on `Rome`. Wouldn't it be sufficient to make it a local variable inside of the e.g. `formatContents` method? 

If it's about code re-use. It may make sense to introduce a `ContentCommand` that issues all the workspace commands necessary:
* open file or change file
* run a command
* close the file

npm/rome/src/index.ts
```diff
@@ -80,7 +88,66 @@ function isFormatContentDebug(
...
+			this.currentFile = {
+				version: this.currentFile.version++,
+				path: {
+					...this.currentFile.path,
+				},
+			};
+			return true;
+		} else {
+			// no same path, let's create a new one
+			this.currentFile = {
```
MichaReiser, line 143: The current implementation is never closing a file which prevents the `Workspace` from cleaning up resources. 

For how long do we intend to keep files around? Does the client have to track all open files inside of a map to know if it has to call `open` or `update`? At what point in time should we start closing files. Should it be done automatically after a grace period or do we need to expose an explicit API that allows the client to explicitly close a file if necessary? Or should we add a new option that specifies if the file should be kept open even after the call. 

My recommendation for now would be to always close the which is the safest (and easiest) and we can explore alternatives later on.

npm/rome/src/index.ts
ematipico: Each document is saved inside a map, and their key `RomePath`. If we call `open_file` multiple times, we reinsert a new file each time.

I don't know the specifics, but I guess updating a file is cheaper than inserting a new one that already exists

npm/rome/src/index.ts
ematipico: Done.

npm/rome/src/index.ts
ematipico: Done. Removed the state and file closed at operation terminated

npm/rome/src/index.ts
```diff
@@ -80,7 +88,31 @@ function isFormatContentDebug(
...
+	 */
+	public static async create(): Promise<Rome> {
+		return new Rome(await Rome.loadWorkspace());
+	}
+
+	private static async loadWorkspace(): Promise<Workspace> {
+		// load the web assembly module
+		main();
+		return Promise.resolve(new Workspace());
+	}
```
MichaReiser, line 114: Nit: Should extract into a standalone function as it doesn't use any class state?

MichaReiser (approved):

npm/rome/src/index.ts
ematipico: This piece of code will be refactored in the next pass

PR #3102, opened 2022-08-23
feat(npm): use daemon client and error handling

ematipico:
## Summary

Part of #3073 

This allows implements the necessary code to allow to use chose the daemon as backend. Initially I used the concept of `Client` but it doesn't seems a client to me... should I use backend or client?

It also implements error handling when formatting broken code.

### `Rome.create`

This API has been extended to accept an optional object that will contain the type of `BackendKind`. If no options are passed, the **WebAssembly backend is chosen as default choice**. There's no strong reason behind it, we can chose the daemon if we want. 

### Error handling 

The shape of the returned diagnostics **is temporary** and likely to change.

The error handling is **partially implemented**. The error inside the `try/catch` clause has been ignored on purpose until I figure out what we can do with that, mostly because each backend returns a different type of error. Here's an example

Error coming from WebAssembly backend:

```block
Error: the file could not be formatted since it has syntax errors and `format_with_errors` is disabled
    at /Users/emanuele/www/rometools/rome/npm/wasm-nodejs/rome_wasm.js:634:17
    at logError (/Users/emanuele/www/rometools/rome/npm/wasm-nodejs/rome_wasm.js:131:18)
    at module.exports.__wbg_new_3ee7ebe9952c1fbd (/Users/emanuele/www/rometools/rome/npm/wasm-nodejs/rome_wasm.js:633:65)
    at wasm://wasm/01978b02:wasm-function[17396]:0x5527e6
    at wasm://wasm/01978b02:wasm-function[13959]:0x50b6ad
    at wasm://wasm/01978b02:wasm-function[10866]:0x4b5a08
    at wasm://wasm/01978b02:wasm-function[1462]:0x2582b1
    at wasm://wasm/01978b02:wasm-function[1275]:0x23550e
    at wasm://wasm/01978b02:wasm-function[8461]:0x460d83
    at Workspace.formatFile (/Users/emanuele/www/rometools/rome/npm/wasm-nodejs/rome_wasm.js:495:18)
    at Rome.formatContent (/Users/emanuele/www/rometools/rome/npm/rome/dist/index.js:100:61)
    at async /Users/emanuele/www/rometools/rome/npm/rome/tests/wasm.test.mjs:51:16
    at async runTest (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/chunk-runtime-error.975bd80a.mjs:488:5)
    at async runSuite (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/chunk-runtime-error.975bd80a.mjs:576:13)
    at async runSuite (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/chunk-runtime-error.975bd80a.mjs:576:13)
    at async runFiles (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/chunk-runtime-error.975bd80a.mjs:620:5)
    at async startTestsNode (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/chunk-runtime-error.975bd80a.mjs:638:3)
    at async /Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/entry.mjs:74:9
    at async Module.withEnv (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/chunk-runtime-error.975bd80a.mjs:259:5)
    at async run (/Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/vitest@0.22.0/node_modules/vitest/dist/entry.mjs:69:5)
    at async file:///Users/emanuele/www/rometools/rome/npm/rome/node_modules/.pnpm/tinypool@0.2.4/node_modules/tinypool/dist/esm/worker.js:99:20
```

Error coming from the daemon:

```block
{
  code: -32603,
  message: "the file could not be formatted since it has syntax errors and `format_with_errors` is disabled",
  data: "the file could not be formatted since it has syntax errors and `format_with_errors` is disabled"
}
```

This is using the JSON-RPC protocol, so we also need to map the code errors. 

## Test Plan

The tests have been moved to use the **compiled code**. It might slow down the developments, but it's needed so we can use `.mjs` and take advantage of ESM directly. 

I added some new test to check the diagnostics, with snapshots.


MichaReiser (commented):
There's a lot going on in this PR. Do you think it would make sense to split 

* Support Deamon
* Diagnostic printer
* Error handling

into its own PR each?

ematipico:
I can remove the diagnostic printer, but error handling and backend can stay together

npm/rome/src/index.ts
```diff
@@ -83,34 +91,68 @@ function isFormatFilesDebug(
...
+			}
 
-	private static async loadWorkspace(): Promise<Workspace> {
-		// load the web assembly module
-		main();
-		return Promise.resolve(new Workspace());
+			default: {
+				let client = await NodeWasm.createWorkspace();
+				return new Rome(client);
+			}
```
MichaReiser, line 154: Nit:

```
case BackendKind.NODE:
default: {
     let client = await NodeWasm.createWorkspace();
     return new Rome(client);
}
```

npm/rome/src/index.ts
```diff
@@ -149,50 +191,72 @@ export class Rome {
...
 		});
 
-		let code;
 		if (options.range) {
-			const result = await this.workspace.formatRange({
-				path: path,
-				// @ts-expect-error For some reason, passing the tuple works but. Need to understand what's going on
+			const result = await this.backend.workspace.formatRange({
+				path: file.path,
+				// TODO: investigate following error:
```
MichaReiser, line 223: Is this a TODO for this PR or a future PR?

npm/rome/src/nodeWasm.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class NodeWasm {
+	public workspace: Workspace;
+	private constructor(workspace: Workspace) {
+		this.workspace = workspace;
+	}
+
+	/**
+     * It creates a new instance of the class {Rome}
+     */
```
MichaReiser, line 16: Formatting, and the documentation seems copy pasted from Rome

npm/rome/src/nodeWasm.ts
```diff
@@ -0,0 +1,33 @@
...
+	public static async createWorkspace(): Promise<NodeWasm> {
```
MichaReiser, line 17: Should these `createWorkspace` functions be static functions on `NodeWasm` instead? 

Would it make sense to have a `Backend` interface type?

```
interface Backend {
  static createWorkspace(): Promise<this>;
  pullDiagnostics(params: PullDiagnosticsParams): Promise<Diagnostic[]>;
}
```

MichaReiser (commented):
Would you mind documenting the newly introduced types. What's a `Deamon`? 

How do you imagine that the lifetime of a deamon works? 

I would expect that it is necessary (the same as for the wasm backend) to disconnect. However, in case of the deamon it may also be desired to shutdown the deamon entirely. 

I could see Rome having a 

* `dispose()` method that disconnects from the deamon or releases the wasm module 
* `stop()` Similar to `rome stop`: Stops the deamon, for wasm, releases the wasm module

The one thing I don't see such API supporting is an explicit way to `start` a deamon, analogous to `rome start`. I think in that case it would be necessary to have an explicit `Deamon` object that users can instantiate and pass to `Rome.create`

npm/rome/src/index.ts
ematipico: I will follow up this with Leo when he's back. It's not for this PR

ematipico:
> Would you mind documenting the newly introduced types. What's a `Deamon`?
> 
> How do you imagine that the lifetime of a deamon works?
> 
> I would expect that it is necessary (the same as for the wasm backend) to disconnect. However, in case of the deamon it may also be desired to shutdown the deamon entirely.
> 
> I could see Rome having a
> 
>     * `dispose()` method that disconnects from the deamon or releases the wasm module
> 
>     * `stop()` Similar to `rome stop`: Stops the deamon, for wasm, releases the wasm module
> 
> 
> The one thing I don't see such API supporting is an explicit way to `start` a deamon, analogous to `rome start`. I think in that case it would be necessary to have an explicit `Deamon` object that users can instantiate and pass to `Rome.create`

I suppose this requires a separated discussion and PR, because there might different use cases. The premise says that a `daemon` is essentially a long term server where we can have multiple clients communicating with it. As for now, each client has it's own workspace instance, so the clients don't overlap. 

If we give the possibility to the API to shutdown a server, then we might cause some side effects and disrupt clients that are using the same server. For absurd, a dev server and a plugin, both communicating with the same daemon. 

Maybe should let the users choose what kind of lifetime they want. I don't know.

npm/rome/src/nodeWasm.ts
ematipico: Unfortunately it doesn't work:

```
TS2322: Type 'Promise<NodeWasm>' is not assignable to type 'Promise<this>'.  
Type 'NodeWasm' is not assignable to type 'this'.     
'NodeWasm' is assignable to the constraint of type 'this', but 'this' could be instantiated with a different subtype of constraint 'NodeWasm'.
```

Using an interface doesn't allow us to use  the `await` inside the functions, which a downside. Not sure how much value this gives to us. What do you think?

npm/rome/src/index.ts
```diff
@@ -83,34 +91,64 @@ function isFormatFilesDebug(
...
+	 * Use this if you want to communicate with the WebAssembly client built for Node.JS
+	 */
+	NODE,
+	/**
+	 * Use this if you want to communicate with the Daemon
+	 */
+	DAEMON,
 }
 
+type Backend = NodeWasm | Deamon;
```
MichaReiser, line 113: I would prefer an `interface` that a `Backend` can implement if it will be the case that there's more difference between the backends than just how they create the instance. For example, it will allow us to have different implementations of a `disconnect`  method. 

It also makes it clearer that the requirements on a `Backend` are and gives us a place where we can document the fields (and what a backend is)

```
interface Backend {
  get workspace(): Workspace;

  // .. other methods
}
```

I otherwise recommend deleting the `Backend` type all together if they are only a storage of the `Workspace` and instead have `createWasmWorkspace()` and `createDeamonWorkspace()` methods that return the Workspace. In that case, `Workspace` is our unit of abstraction and there's no need to have another abstraciton.

npm/rome/src/nodeWasm.ts
MichaReiser: Not sure what you mean by not using await inside of the function? The interface only defines the signature. But having the factory function on the interface may not be important. 

You can still use await in the implementation (but the `this` may not work). 


Nit: Rename to `createNodeWasm()` or just `create` as this returns a `NodeWasm` instance and not a Workspace

ematipico: > Not sure what you mean by not using await inside of the function? 

I locally fixed the issue, although I won't implement the interface in this PR because there's no use case yet:
- we are not using custom calls anymore
- the `createWorkspace` functions have different signatures depending on the type of backend

I will change the name of the `createWorkspace` function as you suggested.

npm/rome/src/index.ts
ematipico: Yeah it makes sense, but not in this PR because there's no use case anymore. The need will raise when dealing with errors. So I will keep the code as is and implement the interface in the next PRs.

MichaReiser (approved):

PR #3103, opened 2022-08-24
feat(playground): mapping sourcefile editor cursor to ast panel

IWANABETHATGUY:
## Summary
1. Mapping source code cursor to Ast panel. 
2. When you move your cursor in the source file editor, ast panel will auto scroll to the corresponding `SyntaxToken`, this
would be helpful when you want to inspect the ast of a big source file.

https://user-images.githubusercontent.com/17974631/186348228-ae52447e-36ea-493a-935c-69296d002380.mp4

## Test Plan
1. See my screen record or test locally.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -27,7 +27,6 @@ bench_analyzer = "run -p xtask_bench --release -- --feature analyzer"
 coverage = "run -p xtask_coverage --profile=release-with-debug --"
 rome-cli = "run -p rome_bin --release --"
 rome-cli-dev = "run -p rome_bin --"
-check_js = "rome-cli ci editors website npm/rome" # Same command with check_js CI
 
 [profile.release]
 lto = true
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -27,6 +27,7 @@ bench_analyzer = "run -p xtask_bench --release -- --feature analyzer"
 coverage = "run -p xtask_coverage --profile=release-with-debug --"
 rome-cli = "run -p rome_bin --release --"
 rome-cli-dev = "run -p rome_bin --"
+check_js = "rome-cli ci editors website npm/rome" # Same command with check_js CI
 
 [profile.release]
 lto = true
--- a/website/playground/src/DesktopPlayground.tsx
+++ b/website/playground/src/DesktopPlayground.tsx
@@ -38,7 +38,7 @@ export default function DesktopPlayground({
 	const romeAstCodeMirrorExtension = [romeAst()];
 	const romeFormatterIrCodeMirrorExtension = [RomeFormatterIr()];
 	//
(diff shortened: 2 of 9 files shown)


website/playground/src/DesktopPlayground.tsx
```diff
@@ -39,6 +37,10 @@ export default function DesktopPlayground({
 
 	const romeAstCodeMirrorExtension = [romeAst()];
 	const romeFormatterIrCodeMirrorExtension = [RomeFormatterIr()];
+	//
```
ematipico, line 40: To remove?

website/playground/src/DesktopPlayground.tsx
```diff
@@ -39,6 +37,10 @@ export default function DesktopPlayground({
...
+	const romeAstSyntacticDataRef = useRef(null as RomeAstSyntacticData | null);
```
ematipico, line 41:
```suggestion
	const romeAstSyntacticDataRef = useRef<RomeAstSyntacticData | null>(null);
```

Not sure if this suggestions does the same, because I have never used this kind of syntax before

website/playground/vite.config.ts
```diff
@@ -12,4 +12,10 @@ export default defineConfig({
 	resolve: {
 		dedupe: ["@codemirror/state"],
 	},
+	server: {
+		fs: {
+			// Allow fs access wasm file
+			allow: ["../../"],
+		},
+	},
```
ematipico, line 20: What does this code fix?

ematipico (approved):

website/playground/vite.config.ts
IWANABETHATGUY: Fix the same issue as https://github.com/rome/tools/pull/3106/files, this is no longer needed

.cargo/config.toml
```diff
@@ -27,6 +27,7 @@ bench_analyzer = "run -p xtask_bench --release -- --feature analyzer"
 coverage = "run -p xtask_coverage --profile=release-with-debug --"
 rome-cli = "run -p rome_bin --release --"
 rome-cli-dev = "run -p rome_bin --"
+check_js = "rome-cli ci editors website npm/rome" # Same command with check_js CI
```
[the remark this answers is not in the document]
IWANABETHATGUY: I suggest keeping it since it is pretty useful when you want to check formatting issues and lint issues before you commit.

website/playground/src/DesktopPlayground.tsx
ematipico: The suggestion should be applied to the other ref too.

.cargo/config.toml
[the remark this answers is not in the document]
ematipico: We have added the commands inside the `package.json`, they are sufficient. If you wish to have that alias, you can create one in your machine.

IWANABETHATGUY:
Done

PR #3104, opened 2022-08-24
perf(rome_formatter): Improve Printer source map performance

MichaReiser:
## Summary

This PR improves the formatters performance by only adding unique source markers inside the printer and adding `take_*` methods to take the source markers to avoid the need to copy all markers to a new vec.

## Test Plan

`cargo test`. I added a new unit test for mapped markers in #3092 that verifies that only unique markers are added (and are mapped to the right location)

### DIFF
--- a/crates/rome_formatter/src/lib.rs
+++ b/crates/rome_formatter/src/lib.rs
@@ -378,6 +378,12 @@ impl Printed {
         self.sourcemap
     }
 
+    /// Takes the list of [SourceMarker] mapping byte positions in the output string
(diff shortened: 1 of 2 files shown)


MichaReiser:
!bench_formatter

ematipico:
Would it be possible to have the unit test here?

MichaReiser:
> Would it be possible to have the unit test here?

I prefer not to or if it is a requirement, I will close this PR and include the change in the other PR.

PR #3109, opened 2022-08-25
feat(rome_service, rome_formatter): remove `format_with_errors`

ematipico:
## Summary

This removes the option `formatWithErrors` from our configuration. 

The reason why the team decided to remove it that, despite the cool concept, we reached a wall where if we wanted to improve the formatting experience and algorithm, we needed to assume that the CST is error-free. 

If you have questions and would like more details, @MichaReiser is your guy :)

## Test Plan

Updated existing tests. CI should pass.

### DIFF
--- a/crates/rome_cli/tests/configs.rs
+++ b/crates/rome_cli/tests/configs.rs
@@ -27,7 +27,6 @@ pub const CONFIG_DISABLED_FORMATTER: &str = r#"{
 pub const CONFIG_ALL_FIELDS: &str = r#"{
   "formatter": {
     "enabled": true,
-    "formatWithErrors": true,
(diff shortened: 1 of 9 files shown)


MichaReiser:
I don't think this is necessary anymore. The current implementation of parenthesized expression supports CST with syntax errors. It is even conservative when it comes about removing parentheses if the inner node is an unknown expression.

ematipico:
Oh great! If there's issue anymore, I am happy to close :)

MichaReiser:
I don't think there is. I wouldn't say the way it works today isn't perfect because it may mess up the code depending on how well (or bad) the parser recovered from a syntax error. But it isn't worse than in the initial version.

PR #3112, opened 2022-08-25
refactor(rome_formatter): `FormatLanguage` trait

MichaReiser:
This PR introduces a new `FormatLanguage` trait that formatters for a specific language must implement.

The `format_*` function family in the `rome_formatter` have rather complicated type parameters that Rust isn't always able to infer and some have a large number of parameters. Making this worse, new features like the pre-processing of AST nodes and the comments refactoring will further increase the parametrization of these functions.

This is why this PR introduces the new `FormatLanguage` trait that groups the language specific types and expose methods to customize the formatter's behaviour.

## Test Plan

This PR isn't adding any new functionality. Ran `cargo test`

### DIFF
--- a/crates/rome_formatter/src/lib.rs
+++ b/crates/rome_formatter/src/lib.rs
@@ -230,6 +230,7 @@ pub trait CstFormatContext: FormatContext {
     type Style: CommentStyle<Self::Language>;
 
     /// Customizes how comments are formatted
+    #[deprecated(note = "Prefer FormatLanguage::comment_style")]
     fn comment_style(&self) -> Self::Style;
 
     /// Returns a ref counted [Comments].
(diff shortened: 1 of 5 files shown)


crates/rome_formatter/src/token.rs
```diff
@@ -676,6 +677,7 @@ where
                     TriviaPrintMode::Trim => 0,
                 });
 
+            #[allow(deprecated)]
```
MichaReiser, line 680: We'll be able to remove the `comment_style` after the comments refactoring but I want to avoid that new usages are introduced in the mean-time.

crates/rome_formatter/src/lib.rs
```diff
@@ -819,32 +817,61 @@ where
...
+    /// true for statement and declaration nodes, to ensure the entire statement
+    /// gets formatted instead of the smallest sub-expression that fits the range
+    fn is_range_formatting_node(&self, _node: &SyntaxNode<Self::SyntaxLanguage>) -> bool {
+        true
+    }
+
+    /// Returns the formatting options
+    fn options(&self) -> &<Self::Context as FormatContext>::Options;
+
+    /// Creates the [FormatContext] with the given `source map` and `comments`
```
ematipico, line 847: Where's "source map" here?

ematipico (approved):

crates/rome_formatter/src/lib.rs
MichaReiser: I hope you don't mind me merging it as is. It will be passed in the next PR and it avoids me one more merge conflict.

PR #3113, opened 2022-08-26
feat(rome_rowan): `cast_ref` and `match_ast` macro

MichaReiser:
This PR introduces a new `AstNode::cast_ref` method that takes a node by reference and only clones it if it can be cast into the `AstNode`.

Being able to pass a reference instead of the owned value can be useful if it's necessary to test against multiple nodes

```
if let Some(if_statement) = JsIfStatement::cast(node.clone()) {
  ..
} else if let Some(expression_stmt) = JsExpressionStatement::cast(node.clone()) {
  ..
} else if let Some(switch_statement) = JsSwitchStatement::cast(node) {
  ..
}
```

As you can see here, the `node` must be cloned `n-1` times where `n` is the number of branches (Rust may be able to optimize them away but doesn't hurt to help the compiler). Using `cast_ref` reduces the clones to at most one in the case the node matches one branch.

This PR further introduces a `match_ast` macro that allows matching a syntax node against AST nodes. This removes the need to match on `JsSyntaxKind` directly and use `unwrap_cast` all over the code base.

```rust
    match_ast! {
        match parent {
            // Only allows expression in the `object` child.
            JsStaticMemberExpression(_) => true,
            JsStaticMemberAssignment(_) => true,
            JsComputedMemberExpression(member_expression) => {
                 member_expression
                    .object()
                    .map(ExpressionNode::into_resolved_syntax)
                    .as_ref()
                    == Ok(node)
            },
            JsComputedMemberAssignment(assignment) => {
                assignment
                    .object()
                    .map(ExpressionNode::into_resolved_syntax)
                    .as_ref()
                    == Ok(node)
            },
            _ => false,
        }
    }
```

## Tests

I added a doc test to the new macro.

## Credit

All credit for the macro goes to [rust-analyzer](https://github.com/rust-lang/rust-analyzer/blob/c6c0ac26456093b71837e20b0ff51655e0c230f7/crates/syntax/src/lib.rs#L173-L198)

### DIFF
--- a/crates/rome_js_formatter/src/js/expressions/assignment_expression.rs
+++ b/crates/rome_js_formatter/src/js/expressions/assignment_expression.rs
@@ -8,10 +8,11 @@ use crate::parentheses::{
 use rome_formatter::write;
 
 use rome_js_syntax::{
-    JsAnyAssignmentPattern, JsAnyExpression, JsAnyForInitializer, JsAssignmentExpression,
-    JsForStatement, JsSyntaxKind, JsSyntaxNode,
+    JsAnyAssignmentPattern, JsAnyExpression, JsAnyForInitializer, JsArrowFunctionExpression,
+    JsAssignmentExpression, JsComputedMemberName, JsExpressionStatement, JsForStatement,
+    JsSequenceExpression, JsSyntaxKind, JsSyntaxNode,
 };
-use rome_rowan::AstNode;
+use rome_rowan::{match_ast, AstNode};
 
 #[derive(Debug, Clone, Default)]
 pub struct FormatJsAssignmentExpression;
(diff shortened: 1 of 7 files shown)


ematipico:
I am not sure the macro will be usable. I mean, the macro it's great, but the example shows and it might contain lot of code, and I know by experience that `rustfmt` sometime is not able to format rust code inside the macro. It really annoyed me this, because it forced me to format the code myself.

MichaReiser:
> I am not sure the macro will be usable. I mean, the macro it's great, but the example shows and it might contain lot of code, and I know by experience that `rustfmt` sometime is not able to format rust code inside the macro. It really annoyed me this, because it forced me to format the code myself.

I agree that this is somewhat painful but I consider this a tooling issue and shouldn't hinder us adopting syntax that otherwise simplifies the code.

ematipico:
> > I am not sure the macro will be usable. I mean, the macro it's great, but the example shows and it might contain lot of code, and I know by experience that `rustfmt` sometime is not able to format rust code inside the macro. It really annoyed me this, because it forced me to format the code myself.
> 
> I agree that this is somewhat painful but I consider this a tooling issue and shouldn't hinder us adopting syntax that otherwise simplifies the code. 

Sure, no issues. 

On a personal note, I wouldn't use it because of time spent at formatting. Please consider also this - I might not be the only one.

ematipico (approved):

PR #3116, opened 2022-08-27
feat(rome_formatter): `IndentIfGroupBreaks` IR element

MichaReiser:
This PR adds a new `indent_if_group_breaks` IR element that has the same semantics as using `if_group_breaks` with `indent` and `if_group_fits_on_line`

```rust
write!(f, [
  group(&text("Head")).with_group_id(Some(id)),
  if_group_breaks(&indent(&text("indented"))).with_group_id(Some(id)),
  if_group_fits_on_line(&text("indented")).with_group_id(Some(id))
])
```

The advantage of the new IR is that it simplifies the debug output of the IR and removes the need to memoize the content.

## Performance

I don't expect this to boost performance in a significant way. While this change reduces the number of IR elements and removes the need for memorizing, these operations by itself don't have a high enough cost to significantly affect the formatter perf.

## Tests

I added new doc tests and ran `cargo test` to verify that using `indent_if_break` doesn't introduce any regression on existing syntax.

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -1750,7 +1750,7 @@ impl<Context> std::fmt::Debug for IfGroupBreaks<'_, Context> {
 ///
 ///     write!(f, [
 ///         group(&text("(aLongHeaderThatBreaksForSomeReason) =>")).with_group_id(Some(group_id)),
-///         indent_if_break(&format_args![hard_line_break(), text("a => b")], group_id)
+///         indent_if_group_breaks(&format_args![hard_line_break(), text("a => b")], group_id)
 ///     ])
 /// });
 ///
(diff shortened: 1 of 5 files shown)


MichaReiser:
!bench_formatter

MichaReiser:
!bench_formatter

crates/rome_formatter/src/builders.rs
```diff
@@ -1714,6 +1714,126 @@ impl<Context> std::fmt::Debug for IfGroupBreaks<'_, Context> {
...
+///
+/// let formatted = format!(SimpleFormatContext::default(), [content]).unwrap();
+///
+/// assert_eq!(
+///     "(aLongHeaderThatBreaksForSomeReason) =>\na => b",
+///     formatted.print().as_code()
+/// );
+/// ```
+#[inline]
+pub fn indent_if_break<Content, Context>(
```
ematipico, line 1792: nit: I wonder if we can rename it to `indent_if_group_breaks`? I like `group_if_breaks` because it reminds me this IR has effects when it's inside a group. I wonder if we can do the same with this new IR too

ematipico (approved):

crates/rome_formatter/src/builders.rs
```diff
@@ -1789,26 +1789,26 @@ impl<Context> std::fmt::Debug for IfGroupBreaks<'_, Context> {
 /// );
 /// ```
 #[inline]
-pub fn indent_if_break<Content, Context>(
+pub fn indent_if_group_breaks<Content, Context>(
     content: &Content,
     group_id: GroupId,
-) -> IndentIfBreak<Context>
+) -> IndentIfBreakBreaks<Context>
```
ematipico, line 1795: `IndentIfBreakBreaks` ❗ ❗

PR #3119, opened 2022-08-28
fix(rome_js_analyze): false positive useShorthandArrayTypes

### LINKED ISSUE #3111: 🐛 ts/useShorthandArrayType is not taking into consideration the need for parenthesis
### Environment information

Running the analyzer from the  ```main``` branch.

### What happened?

The lint rule ts/useShorthandArrayType is not taking into consideration the need for parenthesis when creating the suggested fix.

Example:

```js
let v = [] as Array<(...args: E) => void>;
console.log(v);
```

https://play.rome.tools/?code=let+v+%3D+%5B%5D+as+Array%3C%28...args%3A+E%29+%3D%3E+void%3E%3B%0Aconsole.log%28v%29%3B&lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&typescript=true&jsx=false#bABlAHQAIAB2ACAAPQAgAFsAXQAgAGEAcwAgAEEAcgByAGEAeQA8ACgALgAuAC4AYQByAGcAcwA6ACAARQApACAAPQA+ACAAdgBvAGkAZAA+ADsACgBjAG8AbgBzAG8AbABlAC4AbABvAGcAKAB2ACkAOwA=

Suggested fix here is:

```
error[[ts/useShorthandArrayType](https://rome.tools/docs/lint/rules/useShorthandArrayType/)]: Use shorthand T[] syntax instead of Array<T> syntax.
  ┌─ main.ts:1:15
  │
1 │ let v = [] as Array<(...args: E) => void>;
  │               ^^^^^^^^^^^^^^^^^^^^^^^^^^^

Suggested fix: Use shorthand T[] syntax to replace
    | @@ -1,2 +1,2 @@
0   | - let v = [] as Array<(...args: E) => void>;
  0 | + let v = [] as (...args: E) => void[];
1 1 |   console.log(v);
```

### Expected result

I was expecting the suggested fix to be 

```
let v = [] as ((...args: E) => void)[]; // <- see wrapping parenthesis
```

### Code of Conduct

IWANABETHATGUY:
## Summary
1. Fixes #3111 

## Test Plan
1. Checking newly added snapshot

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/ts/use_shorthand_array_type.rs
+++ b/crates/rome_js_analyze/src/analyzers/ts/use_shorthand_array_type.rs
@@ -118,7 +118,9 @@ fn convert_to_array_type(type_arguments: TsTypeArguments) -> Option<TsType> {
                     | TsType::TsFunctionType(_)
                     | TsType::TsConstructorType(_)
                     | TsType::TsConditionalType(_)
-                    | TsType::TsTypeOperatorType(_) => None,
+                    | TsType::TsTypeOperatorType(_)
+                    | TsType::TsInferType(_)
+                    | TsType::TsMappedType(_) => None,
 
                     TsType::TsReferenceType(ty) if is_array_reference(ty).unwrap_or(false) => {
                         if let Some(type_arguments) = ty.type_arguments() {
--- a/crates/rome_js_analyze/src/analyzers/ts/use_shorthand_array_type.rs
+++ b/crates/rome_js_analyze/src/analyzers/ts/use_shorthand_array_type.rs
@@ -112,8 +112,14 @@ fn convert_to_array_type(type_arguments: TsTypeArguments) -> Option<TsType> {
             .filter_map(|param| {
                 let param = param.ok()?;
                 let element_type = match &param {
-                    TsType::TsUnionType(_) => None,
-                    TsType::TsTypeOperatorType(_) => None,
+                    // Intersection or higher types
+                    TsType::TsUnionType(_)
+                    | TsType::TsIntersectionType(_)
+                    | TsType::TsFunctionType(_)
+                    | TsType::TsConstructorType(_)
+                    | TsType::TsConditionalType(_)
+                    | TsType::TsTypeOperatorType(_) => None,
+
                     TsType::TsReferenceType(ty) if is_array_reference(ty).unwrap_or(false) => {
                         if let Some(type_arguments) = ty.type_arguments() {
                             convert_to_array_type(type_arguments)
--- a/crates/rome_js_analyze/tests/specs/ts/useShorthandArrayType.ts.snap
+++ b/crates/rome_js_analyze/tests/specs/ts/useShorthandArrayType.ts.snap
@@ -22,6 +22,14 @@ let valid6: Array<() => string>;
 type valid7<T> = Array<T extends string ? string : number>
 type valid8 = Array<new (string, number) => string>
 // valid end
+
+//parenthesized type
+let valid8: Array<(string & number)>;
+// infer type
+type valid9<T> = T extends Array<infer R> ? R : any;
+// mapped type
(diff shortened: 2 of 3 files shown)


crates/rome_js_analyze/tests/specs/ts/useShorthandArrayType.ts.snap
```diff
@@ -15,7 +16,12 @@ let invalid2: Promise<Array<string>>;
 let invalid3: Array<Foo<Bar>>;
 let invalid4: Array<[number, number]>;
 
-
+// valid
+let valid5: Array<string & number>;
```
MichaReiser, line 20: Would you mind adding some valid and invalid tests for if the type is parenthesized.

crates/rome_js_analyze/src/analyzers/ts/use_shorthand_array_type.rs
```diff
@@ -112,8 +112,14 @@ fn convert_to_array_type(type_arguments: TsTypeArguments) -> Option<TsType> {
...
                 let element_type = match &param {
-                    TsType::TsUnionType(_) => None,
-                    TsType::TsTypeOperatorType(_) => None,
+                    // Intersection or higher types
+                    TsType::TsUnionType(_)
+                    | TsType::TsIntersectionType(_)
+                    | TsType::TsFunctionType(_)
+                    | TsType::TsConstructorType(_)
+                    | TsType::TsConditionalType(_)
+                    | TsType::TsTypeOperatorType(_) => None,
```
MichaReiser, line 121: What about mapped types and `infer`?

crates/rome_js_analyze/tests/specs/ts/useShorthandArrayType.ts.snap
IWANABETHATGUY: I guess all parenthesis typed should be invalid which means should convert to shorthandArrayTypes, 
https://play.rome.tools/?code=type+res+%3D+Array%3C%28%28a%29%29%3E&lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&typescript=true&jsx=false#dAB5AHAAZQAgAHIAZQBzACAAPQAgAEEAcgByAGEAeQA8ACgAKABhACkAKQA+AA==. I will add it later.

MichaReiser: Rome's analyzer should work independently of Rome's formatter. Users can decide to only adopt one of the two.

IWANABETHATGUY: So which behavior do you expect, e.g.
```js
let valid5: Array<(string & number)>;
```
1. keep as is if the inner type should not convert to `shorthandArrayType`
```js
let valid: Array<(foo | bar)>;
```

2. Convert it to shorthandArraytype. (this is current behavior same as rome classic)
```js
let valid: (foo | bar)[];
```

ematipico: The old rule used to apply the following fix:

```ts
let valid5: (string & number)[];
```

crates/rome_js_analyze/src/analyzers/ts/use_shorthand_array_type.rs
IWANABETHATGUY: Done

crates/rome_js_analyze/tests/specs/ts/useShorthandArrayType.ts.snap
IWANABETHATGUY: Done

IWANABETHATGUY:
All the behavior has been aligned to rome classic

ematipico (approved):

leops (approved):

PR #3120, opened 2022-08-28
refactor(rome_formatter): Return reference from `context.comments()`

MichaReiser:
`context.comments` used to return a `Rc<Comments>` so that the lifetime of `comments` isn't bound to the lifetime of the `FormatContext`. This is necessary to support writing to the formatter while holding to a reference of comments:

(Pseudo code):

```rust
let comments = f.context().comments();

for comment in comments.leading_comments(node) {
  write!(f, [comment])?;
}
```

The downside of returning an `Rc` is that every code path using `comments` pays for the overhead of incrementing the `Rc` counter and using an `Rc` in the signature leaks the abstraction.

This PR changes `Comments` to use an `Rc` internally to enable cheap cloning, similar to how `SyntaxNode` works. This allows `context.comments()` to return a reference to `Comments` instead and only code paths that need to write to the formatter have to pay the overhead of cloning the comments.

## Tests

`cargo test` as this PR isn't adding any new functionality

### DIFF
--- a/crates/rome_formatter/src/comments.rs
+++ b/crates/rome_formatter/src/comments.rs
@@ -143,7 +143,7 @@ pub trait CommentStyle<L: Language> {
 /// Cloning `comments` is cheap as it only involves bumping a reference counter.
 #[derive(Debug, Default, Clone)]
 pub struct Comments<L: Language> {
-    /// The use of a [Rc] is necessary to achieve that [Comments] has a lifetime that is independent of the [crate::Formatter].
+    /// The use of a [Rc] is necessary to achieve that [Comments] has a lifetime that is independent from the [crate::Formatter].
     /// Having independent lifetimes is necessary to support the use case where a (formattable object)[crate::Format]
     /// iterates over all comments and writes them into the [crate::Formatter] (mutably borrowing the [crate::Formatter] and in turn its context).
     ///
--- a/crates/rome_formatter/src/comments.rs
+++ b/crates/rome_formatter/src/comments.rs
@@ -144,7 +144,7 @@ pub trait CommentStyle<L: Language> {
 #[derive(Debug, Default, Clone)]
 pub struct Comments<L: Language> {
     /// The use of a [Rc] is necessary to achieve that [Comments] has a lifetime that is independent of the [crate::Formatter].
-    /// Having independent lifetimes is necessary to support the use case where a (formattable object)[Format]
+    /// Having independent lifetimes is necessary to support the use case where a (formattable object)[crate::Format]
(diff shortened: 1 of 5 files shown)


MichaReiser:
!bench_formatter

crates/rome_formatter/src/comments.rs
```diff
@@ -138,21 +139,26 @@ pub trait CommentStyle<L: Language> {
...
 #[derive(Debug, Default, Clone)]
 pub struct Comments<L: Language> {
-    /// Stores the nodes that have at least one leading suppression comment.
-    suppressed_nodes: HashSet<SyntaxNode<L>>,
-
-    /// Stores all nodes for which [Comments::is_suppressed] has been called.
-    /// This index of nodes that have been checked if they have a suppression comments is used to
-    /// detect format implementations that manually format a child node without previously checking if
-    /// the child has a suppression comment.
+    /// The use of a [Rc] is necessary to achieve that [Comments] has a lifetime that is independent of the [crate::Formatter].
```
ematipico, line 146:
```suggestion
    /// The use of a [Rc] is necessary to achieve that [Comments] has a lifetime that is independent from the [crate::Formatter].
```

crates/rome_formatter/src/comments.rs
```diff
@@ -138,21 +139,26 @@ pub trait CommentStyle<L: Language> {
...
+    /// Having independent lifetimes is necessary to support the use case where a (formattable object)[crate::Format]
+    /// iterates over all comments and writes them into the [crate::Formatter] (mutably borrowing the [crate::Formatter] and in turn its context).
```
ematipico, line 148:
```suggestion
    /// iterates over all comments, and writes them into the [crate::Formatter] (mutably borrowing the [crate::Formatter] and in turn its context).
```

crates/rome_formatter/src/comments.rs
```diff
@@ -138,21 +139,26 @@ pub trait CommentStyle<L: Language> {
...
+    /// for leading in f.context().comments().leading_comments(node) {
+    ///     ^
+    ///     |- Borrows comments
+    ///   write!(f, [comment(leading.piece.text())])?;
+    ///          ^
+    ///          |- Mutably borrows the formatter, state, context, and comments (if comments aren't cloned)
+    /// }
+    /// ```
+    ///
+    /// Using an `Rc` here allows cheaply cloning [Comments] for these use cases.
```
ematipico, line 160:
```suggestion
    /// Using an `Rc` here allows to cheaply clone [Comments] for these use cases.
```

ematipico (approved):

Issue #2991, opened 2022-08-02
☂️ Documentation changes 0.9.0


ematipico:
# Tentative date, 6th September 2022

https://github.com/orgs/rome/projects/3/views/4?query=is%3Aopen+sort%3Aupdated-desc

# Focus of this release
- stabilize configuration file, make sure it's exhaustive but minimal at the same time
- correctness, make sure that parser and linter work as expected, we catch errors early, and we monitor possible performance regressions
- formatter stability
- Node.js API

MichaReiser:
# Documentation changes

Document the decision on how Rome parenthesizes`class`, `function`, and `object` expressions at the begin of a statement and optional chain expressions differently: 

## `ObjectExpression`, `FunctionExpression`, or `ClassExpression` at the beginning of a statement. 

```javascript
// input
{}.b

// Prettier
({}.b)

// Rome
({}).b 
```

Prettier parenthesizes the whole expression whereas Rome parenthesizes the object/function/class expression only. 

The main reason for diverging is that parenthesizing the whole expression requires that the logic is implemented in any expression that starts with another child expression (tagged template, binary expressions, computed member/assignment, sequence, conditional, ... ) and it requires traversing the first expression until it reaches an object expression or any expression that doesn't start with another expression. This is rather expensive and Rome's approach avoids the expensive traversal except for object/function and class expressions, of which there should be fewer. 

## Parenthesized optional chain

```javascript
// Input
(a?.b).c
a?.b.c

// Prettier
(a?.b).c
a?.b.c

// Rome
a?.b.c
a?.b.c
```

Prettier keeps the parentheses if they were present in the source document but never adds them if they were missing. This is the only case where I found that the fact that parens are present in the source document changed how Prettier parenthesizes formatted expression. 

My take is that Rome should apply a consistent formatting regardless if parentheses were or were not present in the source document. 


https://github.com/rome/tools/pull/3057

MichaReiser:
Document quote properties option https://github.com/rome/tools/pull/3065

xunilrj:
Two examples of valid ```noUnusedVariables```:
1 - Starting with an underscore;
2 - Named React.
See https://github.com/rome/tools/pull/3124/files#diff-78c747a14dde0733f5baaa828d4ad30bcca527d46dfd380688bbad45eb291448R67

PR #3124, opened 2022-08-29
feat(rome_js_analyze): no unused starting with underscore; small hack for React and removing suggested fix

xunilrj:
## Summary

This improves noUnusedVariables in three ways and removes the suggestion of removing the unused variables:

1 - It does not flag variables starting with "_";
2 - It does not flag React because some old codebases import React and use it only after JSX transpiling;
3 - It ignores function and class expressions (we are going to deal with these in another rule);

The suggestion was removed, because it suggested removing the first parameter of a function, and of course, completely changing the meaning of the other parameters. We need to think a little bit about what we want to suggest.

## Test Plan

```
> cargo test -p rome_js_analuze -- no_unused
```

### DIFF
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
@@ -1,16 +1,13 @@
-use crate::{semantic_services::Semantic, utils::batch::JsBatchMutation, JsRuleAction};
-use rome_analyze::{
-    context::RuleContext, declare_rule, ActionCategory, Rule, RuleCategory, RuleDiagnostic,
-};
+use crate::semantic_services::Semantic;
+use rome_analyze::{context::RuleContext, declare_rule, Rule, RuleCategory, RuleDiagnostic};
 use rome_console::markup;
-use rome_diagnostics::Applicability;
 use rome_js_semantic::{AllReferencesExtensions, SemanticScopeExtensions};
 use rome_js_syntax::{
-    JsCatchDeclaration, JsConstructorParameterList, JsConstructorParameters, JsFormalParameter,
-    JsFunctionDeclaration, JsIdentifierBinding, JsParameterList, JsParameters, JsSyntaxKind,
+    JsClassExpression, JsConstructorParameterList, JsConstructorParameters, JsFunctionDeclaration,
+    JsFunctionExpression, JsIdentifierBinding, JsParameterList, JsParameters, JsSyntaxKind,
     JsVariableDeclarator,
 };
-use rome_rowan::{AstNode, BatchMutationExt};
+use rome_rowan::AstNode;
 
 declare_rule! {
     /// Disallow unused variables.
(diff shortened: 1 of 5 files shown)


ematipico:
@xunilrj once this PR is merged, could you please add a comment to https://github.com/rome/tools/issues/2991, and remind to update the documentation of the rule with these two gotchas? - the exemption of React, and variables that start with underscore.

Boshen:
FYI for parameters, eslint has the `after-used` option, which I think is pretty nice for a default.

> after-used - unused positional arguments that occur before the last used argument will not be checked, but all named arguments and all positional arguments after the last used argument will be checked.

https://eslint.org/docs/latest/rules/no-unused-vars#args

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -198,43 +211,4 @@ impl Rule for NoUnusedVariables {
 
         Some(diag)
     }
-
-    fn action(ctx: &RuleContext<Self>, _: &Self::State) -> Option<JsRuleAction> {
```
leops, line 202: I think we could still keep the code action for variables, but instead of removing the whole declaration it could suggest to prefix the name with an underscore ?

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
```diff
@@ -117,12 +114,28 @@ impl Rule for NoUnusedVariables {
...
     fn run(ctx: &RuleContext<Self>) -> Option<Self::State> {
         let binding = ctx.query();
-        let model = ctx.model();
+        let name = binding.name_token().ok()?;
+        let name = name.token_text_trimmed();
+        let name = name.text();
+
+        // Old code import React but do not used directly
+        // only indirectly after transpiling JSX.
+        if name.starts_with('_') || name == "React" {
```
leops, line 123: The check for `React` should only apply to JSX files, although at the moment I don't think we have access to the `SourceType` for the file in the `RuleContext`

xunilrj: No, we don´t.
It is a little bit hacky now. I expect this to be turned into an option in the future.

xunilrj:
> FYI for parameters, eslint has the `after-used` option, which I think is pretty nice for a default.
> 
> > after-used - unused positional arguments that occur before the last used argument will not be checked, but all named arguments and all positional arguments after the last used argument will be checked.
> 
> https://eslint.org/docs/latest/rules/no-unused-vars#args

hum... Honestly not sure about this. One of the reasons I like noUnusedVariables is to get typos. So, for example, the code below would not flag ```req``` as unused.

Now... today we only have two ways to suppress these:

1 - Suppression lines;
2 - Starting with an underscore.

Are these fine and enough?

```js
const server = http.createServer(function(req, res) {
    console.log(re); //try to use req but typo
    console.log(res); // correctly uses res
});
```

crates/rome_js_analyze/src/semantic_analyzers/js/no_unused_variables.rs
xunilrj: I would prefer to remove it for now. And we can return it after it is a little bit more mature.

ematipico (approved):

PR #3125, opened 2022-08-29
feat(rome_formatter): Custom separator per `Fill` item

MichaReiser:
This PR refactors `Fill` to support different separators for every item rather than one uniform separator that separates all elements in the list.

This is necessary to implement `fill` for array expressions where the separator should be a:

* soft line break or space: if there are 0 - 1 line breaks before the node
* empty line: if there are more than 1 new lines before the node.

Used in #3126

### DIFF
--- a/crates/rome_formatter/src/builders.rs
+++ b/crates/rome_formatter/src/builders.rs
@@ -1654,16 +1654,16 @@ impl<Context> IfGroupBreaks<'_, Context> {
     ///                 text("["),
     ///                 soft_block_indent(&format_with(|f| {
     ///                     f.fill()
-    ///                         .entry(&text("1,"), &soft_line_break_or_space())
-    ///                         .entry(&text("234568789,"), &soft_line_break_or_space())
-    ///                         .entry(&text("3456789,"), &soft_line_break_or_space())
-    ///                         .entry(&format_args!(
+    ///                         .entry(&soft_line_break_or_space(), &text("1,"))
+    ///                         .entry(&soft_line_break_or_space(), &text("234568789,"))
+    ///                         .entry(&soft_line_break_or_space(), &text("3456789,"))
+    ///                         .entry(&soft_line_break_or_space(), &format_args!(
     ///                             text("["),
     ///                             soft_block_indent(&text("4")),
     ///                             text("]"),
     ///                             if_group_breaks(&text(",")).with_group_id(Some(group_id))
-    ///                         ), &soft_line_break_or_space())
-    ///                         .finish()
+    ///                         ))
+    ///                     .finish()
     ///                 })),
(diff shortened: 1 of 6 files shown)


crates/rome_formatter/src/builders.rs
```diff
@@ -2087,45 +2073,70 @@ impl<'a, 'buf, Context> FillBuilder<'a, 'buf, Context> {
...
+        entry: &dyn Format<Context>,
+        separator: &dyn Format<Context>,
+    ) -> &mut Self {
         self.result = self.result.and_then(|_| {
             let mut buffer = VecBuffer::new(self.fmt.state_mut());
             write!(buffer, [entry])?;
 
-            self.items.extend(buffer.into_vec());
+            let entries = buffer.into_vec();
+            self.items.reserve((entries.len() * 2).saturating_sub(1));
```
ematipico, line 2104: I suppose we reserve double the entries because each entry is paired with another content, the `separator`. In case some of the separators we stored are empty - `separator.is_empty()` is `true` -, is the reservation of the slots still valid?

crates/rome_formatter/src/builders.rs
```diff
@@ -2087,45 +2073,70 @@ impl<'a, 'buf, Context> FillBuilder<'a, 'buf, Context> {
...
             let mut buffer = VecBuffer::new(self.fmt.state_mut());
-            write!(buffer, [entry])?;
-
-            let item = buffer.into_element();
 
-            if !item.is_empty() {
-                self.items.push(item);
+            if !self.items.is_empty() {
+                write!(buffer, [separator])?;
+                self.items.push(buffer.take_element());
```
ematipico, line 2134: Why do we push the separator first? I'm just confused because `separator` is the second item.. so I assumed it comes second in order of the printed items.

crates/rome_formatter/src/builders.rs
MichaReiser: It's even true if the `separator` is empty because we must guarantee that `items` has the format 

```
[
  content,
  separator,
  content,
  separator
]
```

or the printer may incorrectly assume that some `content` is a separator or the other way. That's why the `Content` is 

```
[
  content
  List[],
  content,
  List[],
]
```

In case the separators are empty.

crates/rome_formatter/src/builders.rs
MichaReiser: Let's change the order.

ematipico (approved):

crates/rome_formatter/src/builders.rs
ematipico: I suppose the confusion comes from the word "separator", and we can think of it as a comma. But I think it's fine.

PR #3126, opened 2022-08-29
feat(rome_js_formatter): Heuristic for fill arrays

MichaReiser:
feat(rome_js_formatter): Heuristic for flat arrays

Align Rome's heuristic with Prettier's of when use `fill` vs a group.

There's one difference that I intend not to fix because I believe it is a bug in Prettier.

Rome's printer automatically inserts a `group` around `fill` elements so that `fits` returns `false` if it encounters any hard line break.
Prettier doesn't do so and any hard line break will result in `fits` return true.

This difference is meaningful for

```javascript
[
// comment2
-380014,
-253951682
  ]

```

where the first element contains hard line breaks because of the `//comment`. Prettier assumes that the content fits and, thus, prints `-253...` on the same line. Rome does not and istead, prints a line break after `-3800..`.

I can demonstrate that this leads to issues where Prettier incorrectly exceeds the line width: [example](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBtAOlA9FgBJAWwIRgCZMBaAZgA4AGOgRgBYq32POvufe-+ANJVIBWKgE4RjAGw1yUXLgC6mEAJAQADjACW0AM7JQAQwBOpiAHcACmYSGUxgDaXjAT0PqARqeNgA1nAwAMrGxAAyOlBwyABmzvpw3r4BQcGaflEA5sgwpgCuSSBwBF5wACblFeHGUFn5xllwAGIQpgTGMLp1yCDG+TAQaiAAFjAETgDqIzrw+hlgcMH2szoAbrNuvWD6niBRiaYw1r5ZHXEJRQBW+gAewdlOcACK+RDwF06J6hmmh70wNyaOD6MCmHTaYaacGwSY6cowEbIeg-CyJSa+TS9aEguCmNYxdQARze8BOWgcIBxhwJw1McBJOnpJ0a5yQ8S+RUSBB0uQKXMeL1JMXZl3UMGMXjhCKRSFI4t8Oic2QAwhAiMZes4nMN8okACqShwc74gNaFACSUCqsGCYIhMAAgtbgoCnp9EgBfT1AA)

```javascript
[
// comment2
-380014333333333333333333333333333333333333333333333333333333333333333333,
-253951682
  ]
```

## Tests

**Average compatibility**: 83.70 -> 84.00
**Compatible lines**: 80.79 -> 81.75


crates/rome_js_formatter/src/js/expressions/array_expression.rs
```diff
@@ -16,24 +19,78 @@ impl FormatNodeRule<JsArrayExpression> for FormatJsArrayExpression {
...
     }
 
     fn needs_parentheses(&self, item: &JsArrayExpression) -> bool {
         item.needs_parentheses()
     }
 }
 
+/// Returns `true` for arrays containing at least two elements if
+/// * all elements are either object xor array expressions
+/// * each child array has at least two elements or each child object has at least two members.
```
ematipico, line 52:
```suggestion
/// Returns `true` for arrays containing at least two elements if:
/// * all elements are either object or array expressions
/// * each child array expression has at least two elements, or each child object expression has at least two members.
```

crates/rome_js_formatter/src/js/lists/array_element_list.rs
```diff
@@ -26,22 +25,63 @@ impl FormatRule<JsArrayElementList> for FormatJsArrayElementList {
...
+    ///
+    /// ```javascript
+    /// [
+    ///     1, 2, 3,
+    ///     5, 6,
+    /// ]
+    /// ```
+    Fill,
+
+    /// Prints every element on a single line if the whole array expression exceeds the line width
```
ematipico, line 74: Are there other conditions for breaking on multiple lines, other than exceeding the line width?

crates/rome_js_formatter/src/js/lists/array_element_list.rs
```diff
@@ -54,39 +94,58 @@ fn can_print_fill(list: &JsArrayElementList) -> bool {
...
+                rome_js_syntax::JsAnyLiteralExpression::JsNumberLiteralExpression(literal),
+            ))) => literal.into_syntax(),
+
+            Ok(JsAnyExpression(JsUnaryExpression(expr))) => {
+                let signed = matches!(expr.operator(), Ok(Plus | Minus));
+                let argument = expr.argument();
+
+                match argument {
+                    Ok(JsAnyLiteralExpression(
+                        rome_js_syntax::JsAnyLiteralExpression::JsNumberLiteralExpression(literal),
```
ematipico, line 123: nit: we could remove the `rome_js_syntax` prefix

crates/rome_js_formatter/src/js/lists/array_element_list.rs
```diff
@@ -54,39 +94,58 @@ fn can_print_fill(list: &JsArrayElementList) -> bool {
...
+            })
+        });
+
+        if separator_has_trailing {
+            return false;
+        }
+
+        let syntax = match item.into_node() {
+            Ok(JsAnyExpression(JsAnyLiteralExpression(
+                rome_js_syntax::JsAnyLiteralExpression::JsNumberLiteralExpression(literal),
```
ematipico, line 114: nit: we could remove the `rome_js_syntax` prefix

website/api/constants.js
```diff
@@ -1,25 +1,242 @@
...
-		"ME", "MF", "MG", "MK", "ML", "MM", "MN", "MO", "MQ", "MR", "MS", "MT",
-		"MU", "MV", "MW", "MX", "MY", "MZ", "NA", "NC", "NE", "NG", "NI", "NL",
-		"NO", "NP", "NR", "NU", "NZ", "OM", "PA", "PE", "PF", "PG", "PH", "PK",
-		"PL", "PM", "PN", "PR", "PS", "PT", "PY", "QA", "RE", "RO", "RS", "RU",
-		"RW", "SA", "SB", "SC", "SE", "SG", "SH", "SI", "SJ", "SK", "SL", "SM",
-		"SN", "SO", "SR", "SS", "ST", "SV", "SX", "SZ", "TA", "TC", "TD", "TF",
-		"TG", "TH", "TJ", "TK", "TL", "TM", "TN", "TO", "TR", "TT", "TV", "TW",
-		"TZ", "UA", "UG", "US", "UY", "UZ", "VA", "VC", "VE", "VG", "VN", "VU",
-		"WF", "WS", "XK", "YE", "YT", "ZA", "ZM", "ZW", "ZZ",
+		"AC",
```
ematipico, line 4: Personal thought: isn't it weird arrays of numbers fill the array, while the array of string literals does not? I suppose prettier has their reasons for doing so

ematipico (approved):

website/api/constants.js
MichaReiser: I considered keeping that special handling for string arrays as it isn't too hard for us to do it but then decided against it, simply to optimize for compatibility for now. I think this is something that we can consider adding again in the future.

crates/rome_js_formatter/src/js/lists/array_element_list.rs
MichaReiser: I think this is conflicting because the method uses a `use JsAnyExpression::*` so that `JsAnyLiteralExpression` now resolves to the `JsAnyExpression::JsAnyLiteralExpression` variant instead of the enum.

crates/rome_js_formatter/src/js/lists/array_element_list.rs
MichaReiser: I mentioned that this can also happen if any element gets printed in expanded mode

PR #3129, opened 2022-08-29
feat(rome_js_formatter): Indent Doc comments

MichaReiser:
This PR adds support for indenting js doc comments.

```javascript
/*
     * a comment
     */
```

Becomes
```javascript
/*
 * a comment
 *
```

The PR introduces a new `LeadingCommentRule` type on the `CstFormatContext` that is used to format any leading comment. This allows languages to implement formatting of a comment's content, something that wasn't possible before.

## Tests

**Average compatibility**: 83.70 -> 84.04
**Compatible lines**: 80.79 -> 81.22


crates/rome_js_formatter/src/comments.rs
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            line.trim_start().starts_with('*')
+        }
+    })
+}
+
+#[derive(Eq, PartialEq, Copy, Clone, Debug, Default)]
+pub struct JsCommentStyle;
+
+impl CommentStyle<JsLanguage> for JsCommentStyle {
```
MichaReiser, line 93: I moved this from `context.rs`

crates/rome_js_formatter/tests/specs/prettier/js/comments/trailing-jsdocs.js.snap
```diff
@@ -35,47 +35,35 @@ const CONNECTION_STATUS = exports.CONNECTION_STATUS = {
...
-@@ -4,24 +4,19 @@
+@@ -4,12 +4,9 @@
    CONNECTING: Object.freeze({ kind: "CONNECTING" }),
    NOT_CONNECTED: Object.freeze({ kind: "NOT_CONNECTED" }),
  });
 -
 -/* A comment */
 -/**
-- * A type that can be written to a buffer.
++/* A comment */ /**
```
MichaReiser, line 45: Prettier inserts a line break here. We aren't able to do this yet because Prettier only does so if the doc is a trailing comment of a node. We'll only have this distinction (at least in a cheap way) after the comments refactor. 

The only difference is that the block comments start on the same line. A user can fix this by manually inserting a line break, which is what they also have to do for other block comments that start on the same line.

crates/rome_js_formatter/tests/specs/prettier/js/comments/trailing-jsdocs.js.snap
```diff
@@ -35,47 +35,35 @@ const CONNECTION_STATUS = exports.CONNECTION_STATUS = {
...
 - */
 -/**
-- * Describes the connection status of a ReactiveSocket/DuplexConnection.
-- * - NOT_CONNECTED: no connection established or pending.
-- * - CONNECTING: when `connect()` has been called but a connection is not yet
-- *   established.
-- * - CONNECTED: when a connection is established.
-- * - CLOSED: when the connection has been explicitly closed via `close()`.
-- * - ERROR: when the connection has been closed for any other reason.
++ */ /**
```
ematipico, line 49: There's a missing line break between doc comments. Is it expected?

ematipico (commented):
You should run `cargo codegen-bindings`

crates/rome_js_formatter/tests/specs/prettier/js/comments/trailing-jsdocs.js.snap
MichaReiser: Kind of, yes [see here](https://github.com/rome/tools/pull/3129#discussion_r957456367)

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/comments.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+/// Returns `true` if `comment` is a multi line block comment:
+///
+/// # Examples
+///
+/// ## Doc Comments
+///
+/// ```javascript
```
ematipico, line 61: This is also a block comment:

```js
/**
Documentation
*/
```

An example: https://github.com/sindresorhus/pageres/blob/main/source/index.ts#L57-L60

crates/rome_js_formatter/src/comments.rs
```diff
@@ -0,0 +1,142 @@
...
+/// /** has no line break */
+///
+/// /*
+///  *
+///  this line doesn't start with a star
+///  */
+/// ```
+///
+fn is_doc_comment(comment: &SyntaxTriviaPieceComments<JsLanguage>) -> bool {
```
ematipico, line 82: A could of thoughts:
- is it easy to add tests for this function? I just think that that if requirements changes (or we fix them), we should be able to have tests that verity our requirements
- what do you think about having a crate with these utilities? Something like `rome_js_ast` or `rome_js_ast_utils`. I noticed that we start having various utilities around comments, and maybe these utilities are best put in a separate crate. This will be useful also for documentation generation

crates/rome_js_formatter/src/comments.rs
```diff
@@ -0,0 +1,142 @@
...
+            kind,
+            JsSyntaxKind::R_BRACK
+                | JsSyntaxKind::R_CURLY
+                | JsSyntaxKind::R_PAREN
+                | JsSyntaxKind::COMMA
+                | JsSyntaxKind::SEMICOLON
+                | JsSyntaxKind::DOT
+                | JsSyntaxKind::EOF
+        )
+    }
```
ematipico, line 141: These functions are stateless and don't read `&self`. Should they be just normal functions?

crates/rome_js_formatter/src/comments.rs
MichaReiser: It is a block comment but not a documentation comment which Prettier isn't formatting too [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACGATwzjwHkMYBLaAZzwF49gAdKNlmAegCoeP8eIcLwBBDBgA2hPGACutGBAC2eAFIBDAG4aAymGxVKBCAQAWpAO5wARhg0BzOADo8ukmCoAzGbRWlNHX1DYxxzUnsYMxM8DTwvKkkXAR4uAWw4DQATaGk8WgMjGAB+JHyYQygHAG42AF8QABoQCEoaKFpkUA1sbAhLAAUehE6UDUlLDUJO5ptsDTAAazgYXQ1lOAAZKig4ZC9x2jhZ+aWV9wWdh2QKuWPUZRs4LKznzY0quUc4ADEcZQ0MGoVWQIA0ciUTRAZhgykkAHUzFR4LR7GA4LoRsiqFpkYRQWBaDMQDsjrgBvMHAD9od7gArWhoXRXJIARTkEHgNMkR2a9mwZNBRBIBRCMChGEqMHhVCyUWQAA4AAx8vpHeHzDCgyVwMlaPbNACOHPgFNaozBtAAtLtns8oRljVQMhTHNSkAcefcjsoqDdsHdmrQWXB2Zy9h7ac0YBobDK5WZkAAmaPzRJXADCKndqFoAFYoQo4AAVWOjT28kBaO4ASSgr1gwSKonruiISW5RzqdSAA)

crates/rome_js_formatter/src/comments.rs
MichaReiser: I don't want to restrict implementations of the trait to only be stateless. It should be up to them if they need to read any data. Rust will optimize this away. 

I also plan to remove this trait in my comments refactor.

crates/rome_js_formatter/src/comments.rs
MichaReiser: * I can add a couple of tests. I'll have to see if it's possible to do so as doc tests or if they get too lengthy. 
* I can see how some of these might be useful in other contexts. I still prefer to keep them in the formatter for now until we have use cases and we can move the implementation then. Similar how `OperatorPrecedence` used to be inside of the parser but was moved out when it became evident that it is useful in formatting and linting.

crates/rome_js_formatter/src/comments.rs
ematipico: Maybe it's a Prettier bug? Because I am able to see the documentation. Screenshot from VSCode: 

Screenshot 2022-08-30 at 10 09 42

MichaReiser: Prettier's code is very explicit about the requirement that every line must start with a `*`. 

I think this is a very good reason why `is_doc_comment` should be inside of the formatter. For as far as the formatter is concerned, this isn't a doc comment. It doesn't mean that other tools may use other heuristics. 

One reason why a formatter may be more restrictive than other tools is that it must guarantee that it properly formats the comment in the way the user expects. We may not be able to do so if we don't know enough about the comment.

ematipico (approved):

PR #3133, opened 2022-08-30
perf(rome_formatter): Make source maps optional

MichaReiser:
The Printer source maps are only needed for range formatting but the Printer used to generate them all the time. This has an even higher cost since we rewrite the source tree because the markers must be mapped back to the original position in the source tree.

This PR requires consumers to explicitly enable source maps if they need the source mapping information.


MichaReiser:
!bench_formatter

MichaReiser:
!bench_formatter

PR #3135, opened 2022-08-30
feat(rome_js_analyze): no unused variables supporting ts overload

xunilrj:
## Summary

Part of #2979 

This PR allows Typescript function overloading to not be flagged as unused.

```ts
function used_overloaded(): number;
function used_overloaded(s: string): string;
function used_overloaded(s?: string) {
    return s;
}
used_overloaded();
```

## Test Plan

```
> cargo test -p rome_js_analyze -- no_unused
```

### DIFF
--- a/crates/rome_js_analyze/tests/specs/js/noUnusedVariables/noUnusedVariables.ts.snap
+++ b/crates/rome_js_analyze/tests/specs/js/noUnusedVariables/noUnusedVariables.ts.snap
@@ -13,6 +13,12 @@ class D {
 }
 console.log(new D());
 
+function unused_overloaded(): number;
+function unused_overloaded(s: string): string;
+function unused_overloaded(s?: string) {
+  return s;
+}
+
 // Valid
 
 interface A {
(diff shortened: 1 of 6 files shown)


crates/rome_js_analyze/tests/specs/js/noUnusedVariables/noUnusedVariables.ts.snap
```diff
@@ -81,4 +94,16 @@ warning[js/noUnusedVariables]: This parameter is unused.
 
 ```
 
+```
+warning[js/noUnusedVariables]: This function is unused.
+   ┌─ noUnusedVariables.ts:12:10
+   │
+12 │ function unused_overloaded(s?: string) {
```
ematipico, line 101: I believe this diagnostic **should not be here**? If I understood the purpose of the PR correctly...

xunilrj: This one is fine because all three overloaded functions are not used.

leops (approved):

PR #3136, opened 2022-08-30
chore: update contribution guidelines

ematipico:
## Summary

Updated `CONTRIBUTING.md` with some guidelines around how the flow works for the Node.js APIs.

I also updated the LSP section, which was outdated. It has changed since we merged the daemon feature.

## Test Plan

Just check the rendered result

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,15 +26,17 @@ Rome can be used as a language server by following the instructions below.
 
 ## Language Server and VS Code Extension Development
 
-The Rome language server is the binary crate `rome_lsp` which can be built using:
-```
-cargo build --bin rome_lsp
+The Rome language server is the binary crate `rome` which can be built using the command:
+
+```bash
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -58,6 +60,22 @@ If files specific to your local development environment should be ignored, pleas
 
 You can find more information on this process [here](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer).
 
+## Node.js development
+
+The package `npm/rome` offers some runtime API, that rely on different backends:
```
MichaReiser, line 65: Nit: The npm module `npm/rome` contains Rome's Node JS API that supports different backends.

MichaReiser (approved):
Nit: Would it make sense to add a link to the `Contributing` file (right section) into the `Readme` of `npm/rome`? That was the first file I looked for instructions.

PR #3137, opened 2022-08-30
fix(rome_bin): improve error handling for Unix domain sockets

leops:
## Summary

This PR fixes how the domain socket file used to communicated between the daemon server and workspace clients is handled.
Specifically, since the file is left behind in the system temporary directory when the daemon process is killed, the `open_socket` needs to handle the `ConnectionRefused` error in the same way as the `NotFound` error and consider the daemon is not running.
I also extended this behavior to the `print_socket` with an additional safeguard as it will only try to connect a limited number of times before failing with the last error reported by the system.
Finally, since the server cannot bind to an existing file I added a call to `remove_file` on the domain socket file if it exists when the `run_daemon` function is called

## Test Plan

I tested this manually using Ubuntu on WSL: I checked that after spawning the server process with `cargo rome-cli-dev __run_server` then killing it, the main CLI started with `cargo rome-cli-dev format editors/vscode/src/main.ts` was not using the daemon server nor failing to initialize a connection, and the server process could be started again without failing on the pre-existing socket file. I guess it would probably benefit from being tried out on a macOS machine, as well as a "physical" Linux setup instead of the "virtual" Windows subsystem

### DIFF
--- a/crates/rome_bin/src/service/unix.rs
+++ b/crates/rome_bin/src/service/unix.rs
@@ -1,6 +1,6 @@
 use std::{
     convert::Infallible,
-    env,
+    env, fs,
     io::{self, ErrorKind},
     path::PathBuf,
     time::Duration,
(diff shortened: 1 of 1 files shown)


ematipico:
I am going to test it on my machine

ematipico:
It seems to be working as expected:

- I connected my VSCode extension to the built binary, and made sure that formatting worked;
- then I closed the extension and made sure that the daemon process was still running;
- called the format command via CLI, it worked;
- killed the process and called the format command via CLI, it worked;
- opened the VSCode extension, no errors and the connection got created again;
- killed the process while the extension was open and the daemon gets spawned against successfully. Here's the logs

```block
[Info  - 4:37:19 PM] Server initialized with PID: 40405
[Info  - 4:39:07 PM] Connection to server got closed. Server will restart.
[Info  - 4:39:07 PM] Server initialized with PID: 40451
[Info  - 4:39:36 PM] Connection to server got closed. Server will restart.
[Info  - 4:39:36 PM] Server initialized with PID: 40460

```

ematipico (approved):

Issue #3006, opened 2022-08-04
🐛 Extension doesn't work for New Files


ZachHaber:
### Environment information

```bash
Rome Extension: v0.12.0
No Rome CLI installed globally or locally,

VSCode Env Info:
Version: 1.69.2 (user setup)
Commit: 3b889b090b5ad5793f524b5d1d39fda662b96a2a
Date: 2022-07-18T16:12:52.460Z
Electron: 18.3.5
Chromium: 100.0.4896.160
Node.js: 16.13.2
V8: 10.0.139.17-electron.0
OS: Windows_NT x64 10.0.19044
```

### What happened?

I don't have Rome installed separately from the extension.

1. In a project configured with Rome
1. Create a new file
1. Add code to the file and set the language to Typescript or JS
1. Attempt to format the file with Rome

![image](https://user-images.githubusercontent.com/19843782/182888858-fc107098-7839-45a5-b4f9-4513bd48dd19.png)

The Rome Trace doesn't even see the attempt at running the formatter either.

### Expected result

Rome should be able to format new files in VSCode. Even if that's the default because there's no approach to get access to the `rome.json` associated due to the lack of file paths.

### Code of Conduct

- [X] I agree to follow Rome's Code of Conduct

leops:
Some technical information about this: at the moment the extension only activates on documents with the `file://` scheme because we rely on the name of the file (and more specifically its extension) to determine the file type. In order to support files that do not have a path on disk yet, we'd need to provide an explicit file type to the workspace (probably derived from the language mode of the editor if we can access it over the LSP)

PR #3139, opened 2022-08-31
feat(vscode): add a status bar item for the language server

leops:
## Summary

Related to #3006

Currently users of the VSCode extension have little to no feedback on whether the extension is active or not for the current file, or on the status of the language server. Taking inspiration from rust-analyzer and Prettier, this PR adds a status bar item for Rome that displays different state when the server is starting up, is ready and active, is inactive because Rome doesn't support the current document (currently this is done at the extension level by checking if the document matches the language selector for the extension, in the future we may request additional information to the language server and display the status of individual features like linting or formatting), or has crashed repeatedly. Additionally, clicking on the status bar item will open the "Rome Trace" log channel in the VSCode output panel

## Screenshots

Pending:
status_pending

Ready:
status_ok

Inactive:
status_inactive

Error:
status_error

## Test Plan

The extension doesn't have automated tests, so I built it and installed it locally and tried switching between supported and unsupported documents, as well as killing the server process 5 times so the extension gave up on respawning it and went into the error state.

### DIFF
--- a/editors/vscode/src/status_bar.ts
+++ b/editors/vscode/src/status_bar.ts
@@ -1,0 +1,115 @@
+import { StatusBarAlignment, StatusBarItem, ThemeColor, window } from "vscode";
+import { State } from "vscode-languageclient";
+import { Commands } from "./commands";
+
+enum Status {
+	Pending = "refresh",
+	Ready = "check",
+	Inactive = "eye-closed",
+	Warning = "warning",
+	Error = "error",
+}
+
+export class StatusBar {
+	private statusBarItem: StatusBarItem;
+
+	private serverState: State = State.Starting;
+	private isActive: boolean = false;
+
+	constructor() {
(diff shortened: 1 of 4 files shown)


ematipico:
Would you mind attaching some screenshot show us how's the change?

editors/vscode/src/status_bar.ts
```diff
@@ -0,0 +1,115 @@
+import { StatusBarAlignment, StatusBarItem, ThemeColor, window } from "vscode";
```
ematipico, line 1: The file should be renamed to `statusBar.ts`

ematipico (approved):
Really difficult to review without some visual, but I think it's fine to merge

leops:
> Would you mind attaching some screenshot show us how's the change?

I've update the PR descriptions with screenshots

PR #3140, opened 2022-08-31
feat(rome_service): allow workspace clients to provide a language hint

### LINKED ISSUE #3006: 🐛 Extension doesn't work for New Files

leops:
## Summary

Fixes #3006

This PR adds a new `language_hint` field to the parameters of the `open_file` Workspace method. It lets clients specify a language ID as a hint for the file type resolver, for instance for files that do not have a file extension yet in the editor. To support this I've merged the `EditorLanguage` enum in `rome_lsp` into the `Language` enum in `rome_service`, and refactored error handling across the code base to make the lack of a file extension a non-fatal error in file type determination.

## Test Plan

I've added an additional test for the LSP to verify we can correctly format files without an extension

### DIFF
--- a/crates/rome_lsp/tests/server.rs
+++ b/crates/rome_lsp/tests/server.rs
@@ -274,6 +274,76 @@ async fn document_lifecycle() -> Result<()> {
 }
 
 #[tokio::test]
+async fn document_no_extension() -> Result<()> {
+    let factory = ServerFactory::default();
+    let (service, client) = factory.create().into_inner();
+    let (stream, sink) = client.split();
+    let mut server = Server::new(service);
+
+    let reader = tokio::spawn(client_handler(stream, sink));
+
+    server.initialize().await?;
+    server.initialized().await?;
+
+    server
+        .notify(
+            "textDocument/didOpen",
+            DidOpenTextDocumentParams {
+                text_document: TextDocumentItem {
+                    uri: Url::parse("test://workspace/document")?,
+                    language_id: String::from("javascript"),
+                    version: 0,
+                    text: String::from("statement()"),
+                },
+            },
+        )
+        .await?;
+
+    let res: Option<Vec<TextEdit>> = server
+        .request(
+            "textDocument/formatting",
+            "formatting",
+            DocumentFormattingParams {
+                text_document: TextDocumentIdentifier {
+                    uri: Url::parse("test://workspace/document")?,
+                },
+                options: FormattingOptions {
+                    tab_size: 4,
+                    insert_spaces: false,
+                    properties: HashMap::default(),
+                    trim_trailing_whitespace: None,
+                    insert_final_newline: None,
+                    trim_final_newlines: None,
+                },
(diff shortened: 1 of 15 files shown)


crates/rome_lsp/tests/server.rs
```diff
@@ -273,6 +273,76 @@ async fn document_lifecycle() -> Result<()> {
...
+                    insert_final_newline: None,
+                    trim_final_newlines: None,
+                },
+                work_done_progress_params: WorkDoneProgressParams {
+                    work_done_token: None,
+                },
+            },
+        )
+        .await?
+        .context("formatting returned None")?;
```
ematipico, line 324: What's this `.context` call? Is it something like `expect`?

crates/rome_service/src/workspace/server.rs
```diff
@@ -110,6 +111,31 @@ impl WorkspaceServer {
...
+        let language = self
+            .documents
+            .get(path)
+            .map(|doc| doc.language_hint)
+            .unwrap_or_default();
+
+        self.features.get_capabilities(path, language)
+    }
+
+    /// Return an error ffactory function for unsupported features at a given path
```
ematipico, line 125:
```suggestion
    /// Return an error factory function for unsupported features at a given path
```

ematipico (approved):

crates/rome_lsp/tests/server.rs
leops: The `context` (and `with_context`) methods come from `anyhow` and wrap an existing `Result` (or in this case `Option`) with a new error object that contains the provided "context" as an the error message.
It's primarily intended to easily let functions propagate inner error results to the outer context while attaching additional information about the error (that's how the error handling in `crates/rome_bin/src/service/mod.rs` works for instance), and is a concept I'm think we could integrate in the diagnostics refactor if we start using diagnostics as a generic tool for error reporting across the codebase.
But in this case I'm using it as a simple way to turn an `Option<T>` into a `Result<T, Error>` with a custom message, so it's very similar to an `expect` method that returns an error instead of panicking

PR #3134, opened 2022-08-30
release: 0.9.0

ematipico:
## Summary

Closes #2991 

## Test Plan

Rendered [changelog](https://github.com/rome/tools/blob/release/0.9.0/CHANGELOG.md).
Rendered website.

PR #3143, opened 2022-09-01
feat(rome_cli): add commands for controlling the daemon server lifecycle

leops:
## Summary

This PR adds two commands to the CLI:
- `rome start` ensures the daemon server is running
- `rome stop` sends a shutdown request to the running daemon server

The shutdown request is currently implemented as a notification signal broadcast within the daemon process that stops the `tokio` driving the server. It will cause the runtime to drop all active tasks, cleaning up all the active connections but making no attempts to fulfill in-flight requests (the `tower_lsp` crate doesn't expose enough control over the processing of incoming requests to do that, but moving to an owned implementation of the LSP / JSON-RPC protocol could unblock this)

The main downside of this approach is that it relies on the daemon process being active and capable of processing requests, and the `stop` command has no way to ensure the request is processed in a timely manner if the server ever became blocked or under heavy load. Ideally the CLI would forcefully terminate the daemon process after a grace period if the shutdown request timed out, but there are no cross-platform ways of doing this in the Rust standard library so it would have to be implemented by using either an external crate like `sysinfo` or calling the OS-specific APIs directly.

Additionally, I've modified the default behavior of the CLI regarding the daemon server: previously the CLI would attempt to connect to a daemon running on the system on startup, and fallback to and in-process workspace instance if no running daemon was found. Now the CLI will never try to connect to a running server instance unless the `--use-server` flag is added to the command line, in which case the CLI will exit with an error if the server is not running.

## Test Plan

As with the rest of the daemon-related code this relies on spawning processes and OS-specific IPC primitives, which are global states of the host machine that are hard to test reliably


crates/rome_cli/src/lib.rs
```diff
@@ -22,6 +23,7 @@ pub use reports::{
     formatter::{FormatterReport, FormatterReportFileDetail, FormatterReportSummary},
     Report, ReportDiagnostic, ReportDiff, ReportErrorKind, ReportKind,
 };
+pub use service::{open_transport, SocketTransport};
```
ematipico, line 26: Somewhere in this file we should add 

```rs
#![doc = include_str!("../README.md")]
```

crates/rome_cli/src/commands/daemon.rs
```diff
@@ -0,0 +1,118 @@
...
+    Ok(())
+}
+
+fn stop_server() -> Result<(), Termination> {
+    let rt = Runtime::new()?;
+
+    if let Some(transport) = open_transport(rt)? {
+        let client = WorkspaceClient::new(transport)?;
+        // This can be an error if the connection is closed before the empty response is sent
+        client.shutdown().ok();
```
ematipico, line 55: Should we add a `map_err` and add a `error!` log? At least we don't lose it entirely

crates/rome_cli/src/commands/daemon.rs
```diff
@@ -0,0 +1,118 @@
...
+/// The events received by the subscriber are filtered at the `info` level,
+/// then printed using the [HierarchicalLayer] layer, and the resulting text
+/// is written to log files rotated on a hourly basis (in
+/// `rome-logs/server.log.yyyy-MM-dd-HH` files inside the system temporary
+/// directory)
+fn setup_tracing_subscriber() {
+    /// This filter enables:
+    /// - All spans and events at level info or higher
+    /// - All spans and events in the `rome_lsp` and `rome_js_parser` crates
+    const LOGGING_FILTER: &str = "info,rome_lsp=trace,rome_js_parser=trace";
```
ematipico, line 101: Do you think it's possible to have something like `rome_*=trace`?

ematipico (commented):
Re: the naming of the commands, Rome classics used to have these commands:
- `rome start`, start the daemon
- `rome stop`, stop the daemon
- `rome dev` , start  the dev server

I have a different suggestion on this regard. What if we called it `remote`? And have commands like:
- `rome remote start` (daemon)
- `rome remote stop` (daemon)
- `rome server dev` (development server)

---

The concept of `daemon` is not beginner friendly, I think we used have - or plan to have, a section of the website where we explain what it is.

crates/rome_cli/src/commands/daemon.rs
leops: The "connection closed" error is the expected behavior, since the server will try to abort every task as fast as possible when it gets the shutdown request. But I don't think it's entirely guaranteed, depending on how the scheduler works it might be possible for the response to be sent before the connection is dropped, and future versions of the server may try to fulfill ongoing requests before stopping, so we can't assume the result will be an error either.

crates/rome_cli/src/commands/daemon.rs
leops: Unfortunately the `EnvFilter` doesn't attempt to parse the crate names in the filter and only checks for exact string equality, so this would only increase the log level for a crate named `rome_*`. In order to increase the filter level for all crates starting with `rome_`, we'd have to implement the filter manually instead of using `EnvFilter`

MichaReiser:
> Re: the naming of the commands, Rome classic used to have these commands:
> 
>     * `rome start`, start the daemon
> 
>     * `rome stop`, stop the daemon
> 
>     * `rome dev` , start  the dev server
> 
> 
> I have a different suggestion on this regard. What if we called it `remote`? And have commands like:
> 
>     * `rome remote start` (daemon)
> 
>     * `rome remote stop` (daemon)
> 
>     * `rome server dev` (development server)
> 
> 
> If not likable, I would for what Rome classic used to have.
> 
> The concept of `daemon` is not beginner friendly, I think we used have - or plan to have, a section of the website where we explain what it is.

What does `rome dev` exactly do? How is it different from `rome start`?

I prefer `rome start` as I find `deamon` to technical. What if we decide to run Rome on a `remote` server? The remote server may decide to shutdown / start the server whenever it likes for as long as it isn't noticeable to the client and `rome start` only askes the server to put everything in place. 


I prefer `rome start` over `rome remote start` as it may give the misconception that rome is running remotely. That might scare people off with stringent code licensing as they may fear that their code gets transmitted externally.

crates/rome_cli/src/commands/daemon.rs
```diff
@@ -0,0 +1,118 @@
...
+            argument: "daemon <command>",
+            source,
+        })?;
+
+    match subcommand.as_deref() {
+        Some("start") => start_server(),
+        Some("stop") => stop_server(),
+
+        // Internal private commands
+        Some("__run_server") => run_server_session(),
```
MichaReiser, line 31: I prefer keeping these commands internal until we have concrete use cases to make them public.

crates/rome_cli/src/commands/daemon.rs
```diff
@@ -0,0 +1,118 @@
...
+        Some(cmd) => Err(Termination::UnknownCommand {
+            command: format!("daemon {cmd}"),
+        }),
+    }
+}
+
+fn start_server() -> Result<(), Termination> {
+    let rt = Runtime::new()?;
+    rt.block_on(ensure_daemon())?;
+    Ok(())
```
MichaReiser, line 46: Nit
```suggestion
    rt.block_on(ensure_daemon())
```

crates/rome_cli/src/commands/daemon.rs
```diff
@@ -0,0 +1,118 @@
...
+    let rt = Runtime::new()?;
+    let factory = ServerFactory::default();
+    let cancellation = factory.cancellation();
+    let span = debug_span!("Running Server", pid = std::process::id());
+
+    rt.block_on(async move {
+        tokio::select! {
+            res = run_daemon(factory).instrument(span) => {
+                match res {
+                    Ok(_) => unreachable!(),
```
MichaReiser, line 73: Would you mind documenting why this is unreachable?

crates/rome_cli/src/commands/daemon.rs
```diff
@@ -0,0 +1,118 @@
...
+                Ok(())
+            }
+        }
+    })
+}
+
+fn print_server_socket() -> Result<(), Termination> {
+    let rt = Runtime::new()?;
+    rt.block_on(print_socket())?;
+    Ok(())
```
MichaReiser, line 88: Nit
```suggestion
    rt.block_on(print_socket())
```

MichaReiser (approved):
I think it's reasonable to not support "kill" at the moment as it is technically more involved and even imposes the question if `kill` should be its own command

crates/rome_cli/src/commands/daemon.rs
leops: The `ensure_daemon` and `print_socket` functions return an `io::Result`, while the `start_server` and `print_server_socket` wrappers return an `anyhow::Result`, so using the `?` operator is necessary to coerce the error into the correct type. Alternatively this would have to be written as `rt.block_on(ensure_daemon()).map_err(anyhow::Error::from)`

crates/rome_cli/src/commands/daemon.rs
leops: The return type of the `run_daemon` function is a stable equivalent to `io::Result<!>`, where the `Ok` branch is statically known to never be taken (the Rust compiler should not generate any code for the branch). Future versions of Rust may be able to detect this branch is dead code and not needed, but at the moment as the never type isn't stabilized it's required that even unreachable branches are written down in the code for now

ematipico (approved):

ematipico:
If you merge this PR before the release, please update https://github.com/rome/tools/pull/3134 to add these new commands to the help command and the website

ematipico:
Regarding `--use-server`, what do you think to make this argument to accept a string instead? This might allow us in the future to extend it and specify a remote location of the server. The default implement could be localhost? Unless if you planned something else for this feature

leops:
> Regarding `--use-server`, what do you think to make this argument to accept a string instead? This might allow us in the future to extend it and specify a remote location of the server. The default implementation could be localhost? Unless if you planned something else for this feature

At the moment I didn't include this because the default value is OS-specific (it's something like `<system temporary directory>/rome-socket` on Unix and `\\.\pipe\rome-service` on Windows) and it's not possible to connect to a remote target at the time, but adding the option for the `--use-server` argument to have an optional value in the future should be a backward-compatible change

ematipico (approved):
Changes look good to me.

I tried the PR locally, the commands lack messaging, so it's really difficult, as a user, to know if the command successfully terminated or not. I would add some simple messaging, other than that, let's 🚢  it!

crates/rome_cli/Cargo.toml
```diff
@@ -15,19 +18,29 @@ rome_service = { path = "../rome_service" }
...
 thiserror = "1.0.30"
 rayon = "1.5.1"
 serde = { version = "1.0.133", features = ["derive"] }
 serde_json = { version = "1.0.74" }
+tokio = { version = "1.15.0", features = ["io-std", "io-util", "net", "time", "rt", "rt-multi-thread", "macros"] }
+anyhow = "1.0.52"
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2.127"
+tokio = { version = "1.15.0", features = ["io-std", "io-util", "net", "time", "rt", "rt-multi-thread", "macros", "process"] }
```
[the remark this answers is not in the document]
leops: Actually it's not necessary, I'll remove it

Issue #2830, opened 2022-07-06
🐛 Formatting of TSX file may differ from Prettier


tsteckenborn:
### Environment information

Probably not relevant due to the visible differences in the playground.

### What happened?

Formatting a TSX file there's a [deviation ](https://play.rome.tools/?lineWidth=80&indentStyle=tab&quoteStyle=double&indentWidth=2&sourceType=module&treeStyle=0&typescript=true&jsx=true#aQBtAHAAbwByAHQAIAB7ACAARgBvAG8AdABlAHIAIAB9ACAAZgByAG8AbQAgACcAfgAvAGMAbwBtAHAAbwBuAGUAbgB0AHMALwBGAG8AbwB0AGUAcgAnADsACgBpAG0AcABvAHIAdAAgAHsAIABIAGUAYQBkAGUAcgAgAH0AIABmAHIAbwBtACAAJwB+AC8AYwBvAG0AcABvAG4AZQBuAHQAcwAvAEgAZQBhAGQAZQByACcAOwAKAGkAbQBwAG8AcgB0ACAAewAgAEgAZQByAG8AIAB9ACAAZgByAG8AbQAgACcAfgAvAGMAbwBtAHAAbwBuAGUAbgB0AHMALwBIAGUAcgBvACcAOwAKAAoAZQB4AHAAbwByAHQAIABkAGUAZgBhAHUAbAB0ACAAZgB1AG4AYwB0AGkAbwBuACAASQBuAGQAZQB4ACgAKQAgAHsACgAKACAAIAByAGUAdAB1AHIAbgAgACgACgAgACAAIAAgADwAPgAKACAAIAAgACAAIAAgADwASABlAGEAZABlAHIAIAAvAD4ACgAgACAAIAAgACAAIAA8AG0AYQBpAG4APgAKACAAIAAgACAAIAAgADwASABlAHIAbwAgAC8APgAKACAAIAAgACAAIAAgADwALwBtAGEAaQBuAD4ACgAgACAAIAAgACAAIAA8AEYAbwBvAHQAZQByACAALwA+AAoAIAAgACAAIAAgACAAPAAvAD4ACgAgACAAKQA7AAoAfQAKAAoACgAKAAoACgAKAAoA) between Prettier Playground and Rome Playground

Input:
```TSX
import { Footer } from '~/components/Footer';
import { Header } from '~/components/Header';
import { Hero } from '~/components/Hero';

export default function Index() {

  return (
    <>
      <Header />
      <main>
[...]

PR #3144, opened 2022-09-01
feat(rome_js_formatter): Format JSX

MichaReiser:
## Summary

This PR improves Rome's formatting of JSX to bring it closer to Prettier's formatting. It improves the formatting of attributes, element heads, and most importantly of the children. 

The formatting of the children is complicated for two reasons:

* It is necessary to merge multiple children. For example, two `{' '}{' '}` should be merged into a single whitespace. Isolating children in isolation is, therefore, not possible. 
* The formatter has to format children twice, once for a flat representation and one if the element doesn't fit on a single line. 

The way this PR solves this challenge is by transforming a `JsxChildList` into an intermediate representation that strips any children that have no semantic meaning and have no effect on the formatting (empty line has no semantic meaning, but is something we want to preserve). It then uses this transformed representation to perform the formatting, applying slightly different formatting between the "flat" and expanded versions. 

This PR does not implement the special handling of JSX inside of conditional expression. This is left for another pull request. 

## Test Plan

**Average compatibility**: 84.94 -> 85.41
**Compatible lines**: 83.08 ->  83.70

I formatted the ant design project with Prettier and then Rome and manually went through the changes. This helped me identify a few bugs that I fixed. The main remaining differences are now caused by call arguments being formatted differently. 

[Verified](https://play.rome.tools/0014af2/?code=import+%7B+Footer+%7D+from+%27%7E%2Fcomponents%2FFooter%27%3B%0Aimport+%7B+Header+%7D+from+%27%7E%2Fcomponents%2FHeader%27%3B%0Aimport+%7B+Hero+%7D+from+%27%7E%2Fcomponents%2FHero%27%3B%0A%0Aexport+default+function+Index%28%29+%7B%0A%0A++return+%28%0A++++%3C%3E%0A++++++%3CHeader+%2F%3E%0A++++++%3Cmain%3E%0A++++++%3CHero+%2F%3E%0A++++++%3C%2Fmain%3E%0A++++++%3CFooter+%2F%3E%0A++++++%3C%2F%3E%0A++%29%3B%0A%7D&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&typescript=false&jsx=true#aQBtAHAAbwByAHQAIAB7ACAARgBvAG8AdABlAHIAIAB9ACAAZgByAG8AbQAgACcAfgAvAGMAbwBtAHAAbwBuAGUAbgB0AHMALwBGAG8AbwB0AGUAcgAnADsACgBpAG0AcABvAHIAdAAgAHsAIABIAGUAYQBkAGUAcgAgAH0AIABmAHIAbwBtACAAJwB+AC8AYwBvAG0AcABvAG4AZQBuAHQAcwAvAEgAZQBhAGQAZQByACcAOwAKAGkAbQBwAG8AcgB0ACAAewAgAEgAZQByAG8AIAB9ACAAZgByAG8AbQAgACcAfgAvAGMAbwBtAHAAbwBuAGUAbgB0AHMALwBIAGUAcgBvACcAOwAKAAoAZQB4AHAAbwByAHQAIABkAGUAZgBhAHUAbAB0ACAAZgB1AG4AYwB0AGkAbwBuACAASQBuAGQAZQB4ACgAKQAgAHsACgAKACAAIAByAGUAdAB1AHIAbgAgACgACgAgACAAIAAgADwAPgAKACAAIAAgACAAIAAgADwASABlAGEAZABlAHIAIAAvAD4ACgAgACAAIAAgACAAIAA8AG0AYQBpAG4APgAKACAAIAAgACAAIAAgADwASABlAHIAbwAgAC8APgAKACAAIAAgACAAIAAgADwALwBtAGEAaQBuAD4ACgAgACAAIAAgACAAIAA8AEYAbwBvAHQAZQByACAALwA+AAoAIAAgACAAIAAgACAAPAAvAD4ACgAgACAAKQA7AAoAfQA=) that the example from the bug report #2830 now gets formatted the same as prettier.


crates/rome_js_formatter/tests/specs/prettier/js/ternaries/nested-in-condition.js.snap
```diff
@@ -70,8 +70,19 @@ const value = (bifornCringerMoshedPerplexSawder
 -  </Element2>
 -);
 +)
-+  ? <Element><Sub /><Sub /><Sub /><Sub /><Sub /><Sub /></Element>
-+  : <Element2><Sub /><Sub /><Sub /></Element2>;
++  ? <Element>
```
MichaReiser, line 73: Matching Prettier's formatting for conditional's containing JSX requires changes to the conditional formatting. I'll follow up with another PR to not further increase the size of this PR.

crates/rome_formatter/src/format_element.rs
```diff
@@ -984,9 +990,15 @@ impl Format<IrFormatContext> for FormatElement {
...
                             dynamic_text(&std::format!("<interned {index}>"), TextSize::default()),
                             space(),
-                            &interned.0.as_ref()
                         ]
-                    )
+                    )?;
+
+                    match interned.0.as_ref() {
+                        element @ FormatElement::Text(_) | element @ FormatElement::Space => {
+                            write!(f, [text("\""), element, text("\"")])
```
MichaReiser, line 998: This fixes an issue where an interned string token wasn't put in quotes which made  it very confusing when reading Rome's format IR

crates/rome_formatter/src/printed_tokens.rs
```diff
@@ -38,7 +38,8 @@ impl PrintedTokens {
                     descendant_offset if descendant_offset < *offset => {
                         panic!("token has not been seen by the formatter: {descendant:#?}.\
                         \nUse `format_replaced` if you want to replace a token from the formatted output.\
-                        \nUse `format_removed` if you to remove a token from the formatted output")
+                        \nUse `format_removed` if you to remove a token from the formatted output.\n\
```
ematipico, line 41:
```suggestion
                        \nUse `format_removed` if you want to remove a token from the formatted output.\n\
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -474,12 +474,28 @@ impl<'a> Printer<'a> {
...
+                            args.with_print_mode(PrintMode::Flat),
+                            &empty_rest,
+                            self,
+                        ) {
+                        PrintMode::Flat
+                    } else {
+                        PrintMode::Expanded
+                    };
+
+                    self.print_all(queue, &[separator], args.with_print_mode(print_mode));
```
ematipico, line 491: What are we fixing/changing here?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -474,12 +474,28 @@ impl<'a> Printer<'a> {
...
                 }
-                (None, _) => {
+                (None, None) => {
                     break;
                 }
+                (None, Some(_)) => {
+                    unreachable!()
```
ematipico, line 497: Could you add a string saying why it's unreachable please?

crates/rome_js_formatter/src/jsx/attribute/expression_attribute_value.rs
```diff
@@ -19,58 +21,100 @@ impl FormatNodeRule<JsxExpressionAttributeValue> for FormatJsxExpressionAttribut
...
+                matches!(argument.tag(), Ok(JsxAnyTag::JsxElement(_)))
+                    && should_inline_jsx_expression(
+                        &argument.into(),
+                        Some(await_expression.syntax().clone()),
+                    )
+            }
+            _ => false,
+        },
+        _ => {
+            // Don't indent conditional expressions only inside of children.
```
ematipico, line 110: Could you rephrase the comment? The "only inside of children" seems off

crates/rome_js_formatter/src/jsx/tag/element.rs
```diff
@@ -1,32 +1,168 @@
...
+    /// ```javascript
+    /// <div>
+    ///   {`A Long Template String That uses ${
+    ///     5 + 4
+    ///   } that will eventually break across multiple lines ${(40 / 3) * 45}`}
+    /// </div>;
+    /// ```
+    Template(JsxExpressionChild),
+
+    /// Default layout used if there's no special formatting for the children
```
ematipico, line 166: How's a default layout? Should we explain it or show a snippet?

ematipico (approved):

crates/rome_formatter/src/printer/mod.rs
MichaReiser: We didn't use to print a trailing separator in will sequences:

```
Fill[a, \n, b, \n]
```

We dropped the trailing `\n` which is probably fine in most cases but the printer should, in my view, never drop any content. So this change ensures that the `\n` following `b` gets printed.

crates/rome_js_formatter/src/jsx/attribute/expression_attribute_value.rs
MichaReiser: Thanks for pointing this out. I was able to move this out into the `JsxExpressionChild` which is the only case where it applies. This then removes the need for the comment altogether.

PR #3146, opened 2022-09-01
feat(rome_service): ignore files

### LINKED ISSUE #2827: support for ignore patterns

ematipico:
## Summary

Closes #2827 

### Configuration 

This PR adds support for ignoring files via configuration. The new field has been added to the `formatter` and the `linter` sections.

```json
{
	"formatter": {
		"ignore": ["path/to"]
	},
	"linter": {
		"ignore": ["*.ts"]
	}
}
```

- the functions `deserialize_globals` and `serialize_globals` have been renamed to `deserialize_set_of_strings` and `serialize_set_of_strings`, so I could apply the same functions for the fields `globals` and `ignore`;

### Matcher

I forked the create [glob](https://github.com/rust-lang/glob) locally. I just copied the engine (parser) of the unix shell style patterns, as we are not interested in actual `glob` function, which does a traversal of the file system. The only changes that I applied to the crate are merely around clippy - it seems there's low activity on the code - and removed some tests. I won't be able to answer technical questions around this crate.

I added license files and credits to the website.

I created a small wrapper called `Matcher` to add patterns and run the matcher against paths and strings. It also holds a `HashMap` to track the already checked paths/strings. 

**There's a catch with this crate**: the crate doesn't support patterns like `"test.ts"`. I tested it myself and I think it's intentional. I added some basic check in the `Matcher` struct to cover this cases, I figured that users might want to ignore single files.

### Workspace

The check is done in two places inside the workspace:
- inside the function `support_features`
- inside a newly created function called `is_file_ignored`

`support_features` now returns a new struct which olds an optional `reason` that explains why the workspace doesn't support a certain file. Before, we were not distinguish the reason why a file couldn't be supported, and this caused some issue that **are not fixed**. If you check the snapshots, you might notice some deletions of some code that was out of place. 

We now use an enum called `UnsupportedReason` which tells the traversal/LSP why a file is not supported, and based on that they can react accordingly. 

`is_file_ignored` is in the workspace `Workspace`. This new function is now used in most of all functions that act on some files, I am not sure if I should extend the functionality also on `get_syntax_tree` and `get_control_flow_graph`.

The usage of `is_file_ignored` has been placed **before the parsing** starts. There's no need to actually parse a file if it's ignored. 

When a file is ignored, we intentionally return an error. It's up to the clients to decide to ignore the error or deal with it. This is a pattern we already apply in case a file is not supported.

## Test Plan

I added two new tests for the CLI. 

I manually built the VSCode extension and made sure that I don't see diagnostics if a file is ignored. And also if the file doesn't get formatted.

### DIFF
--- a/crates/rome_service/src/settings.rs
+++ b/crates/rome_service/src/settings.rs
@@ -77,6 +77,8 @@ pub struct FormatSettings {
     pub format_with_errors: bool,
     pub indent_style: Option<IndentStyle>,
     pub line_width: Option<LineWidth>,
+    /// List of paths/files to matcher
+    pub ignored_files: IndexSet<String>,
 }
 
 impl Default for FormatSettings {
@@ -86,6 +88,7 @@ impl Default for FormatSettings {
             format_with_errors: false,
             indent_style: Some(IndentStyle::default()),
             line_width: Some(LineWidth::default()),
+            ignored_files: IndexSet::default(),
         }
     }
 }
@@ -98,6 +101,9 @@ pub struct LinterSettings {
 
     /// List of rules
     pub rules: Option<Rules>,
+
+    /// List of paths/files to matcher
+    pub ignored_files: IndexSet<String>,
 }
 
 impl Default for LinterSettings {
@@ -105,6 +111,7 @@ impl Default for LinterSettings {
         Self {
             enabled: true,
             rules: Some(Rules::default()),
+            ignored_files: IndexSet::default(),
         }
     }
 }
--- a/crates/rome_service/src/settings.rs
+++ b/crates/rome_service/src/settings.rs
@@ -1,4 +1,4 @@
-use crate::{Configuration, Rules};
+use crate::{Configuration, MatchOptions, Matcher, RomeError, Rules};
 use indexmap::IndexSet;
 use rome_console::codespan::Severity;
 use rome_formatter::{IndentStyle, LineWidth};
@@ -29,10 +29,13 @@ impl WorkspaceSettings {
     }
 
     /// The (configuration)[Configuration] is merged into the workspace
-    pub fn merge_with_configuration(&mut self, configuration: Configuration) {
+    pub fn merge_with_configuration(
+        &mut self,
+        configuration: Configuration,
+    ) -> Result<(), RomeError> {
         // formatter part
         if let Some(formatter) = configuration.formatter {
-            self.formatter = FormatSettings::from(formatter);
+            self.formatter = FormatSettings::try_from(formatter)?;
         }
         let formatter = configuration
             .javascript
@@ -45,11 +48,13 @@ impl WorkspaceSettings {
 
         // linter part
         if let Some(linter) = configuration.linter {
-            self.linter = LinterSettings::from(linter)
+            self.linter = LinterSettings::try_from(linter)?;
         }
 
         let globals = configuration.javascript.and_then(|j| j.globals);
         self.languages.javascript.globals = globals;
+
+        Ok(())
     }
 
     /// It retrieves the severity based on the `code` of the rule and the current configuration.
@@ -78,7 +83,7 @@ pub struct FormatSettings {
     pub indent_style: Option<IndentStyle>,
     pub line_width: Option<LineWidth>,
     /// List of paths/files to matcher
-    pub ignored_files: IndexSet<String>,
+    pub ignored_files: Matcher,
 }
 
 impl Default for FormatSettings {
@@ -88,7 +93,11 @@ impl Default for FormatSettings {
             format_with_errors: false,
             indent_style: Some(IndentStyle::default()),
             line_width: Some(LineWidth::default()),
-            ignored_files: IndexSet::default(),
+            ignored_files: Matcher::new(MatchOptions {
+                case_sensitive: true,
+                require_literal_leading_dot: false,
+                require_literal_separator: false,
+            }),
         }
     }
 }
@@ -103,7 +112,7 @@ pub struct LinterSettings {
     pub rules: Option<Rules>,
 
     /// List of paths/files to matcher
-    pub ignored_files: IndexSet<String>,
+    pub ignored_files: Matcher,
 }
 
 impl Default for LinterSettings {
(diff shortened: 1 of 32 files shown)


crates/rome_service/src/settings.rs
```diff
@@ -72,6 +72,8 @@ pub struct FormatSettings {
     pub format_with_errors: bool,
     pub indent_style: Option<IndentStyle>,
     pub line_width: Option<LineWidth>,
+    /// List of paths/files to matcher
+    pub ignored_files: IndexSet<String>,
```
leops, line 76: Since the point of having separate `Configuration` and `Settings` data structures is to be able to hold the current settings for the workspace in a more optimized shape for fast access, it would be beneficial to hold the `Matcher` struct here instead of the plain set of ignore strings in order to avoid having to reconstruct the matcher on every request and make a better use of its caching feature (but it would require a refactor of the matcher to get rid of the lifetime argument and make it work on owned data instead)

crates/rome_service/src/workspace/server.rs
```diff
@@ -236,9 +292,15 @@ impl Workspace for WorkspaceServer {
             .debug
             .debug_formatter_ir
             .ok_or_else(self.build_capability_error(&params.path))?;
+        let settings = self.settings();
```
leops, line 295: Since this is going to be a common pattern repeated in every method that access a certain file, maybe it could be pulled into a shared helper function (eg. at the start of the `get_parse` method ?)

crates/rome_service/src/workspace.rs
```diff
@@ -76,6 +76,40 @@ pub struct SupportsFeatureParams {
...
+        }
+    }
+
+    fn disabled() -> Self {
+        Self {
+            reason: Some(UnsupportedReason::FeatureNotEnabled),
+        }
+    }
+
+    fn incapable() -> Self {
```
MichaReiser, line 98: Nit: Can we apply a consistent between the variant names of the enum and this helper. `incapable` -> `file_not_supported()`

Nit: This function can all be made `const`

MichaReiser (commented):
How does the traversal of a directory work. Does the implementation use the `glob` patterns to filter out directories/files before calling into format / pull diagnostics of a single file ?

crates/rome_service/src/settings.rs
ematipico: This is a good suggestion. I will need to wait for https://github.com/rome/tools/pull/3157 to be merged first, otherwise I'm kind forced to make the whole data structure serde compatible.

crates/rome_service/src/workspace/server.rs
ematipico: The suggestion makes sense, but now I am wondering how will be able to deduce which files should be ignored.

For example, the function `pull_diagnostics`. This function is a wildcard and just says "give me the diagnostics of this file based on `RuleCategories`". This info doesn't tell us if we are formatting, linting, doing both, etc.

MichaReiser:
I'll leave it to @leops to review this PR as he has more context. 

Would you mind doing some profiling. E.g. compare wall time of formatting/linting rome classic before/after your change?

ematipico:
> Would you mind doing some profiling. E.g. compare wall time of formatting/linting rome classic before/after your change?

Using `time`, commands ran using the `--release` version

## this PR

### `format` command

```
The number of diagnostics exceeds the number allowed by Rome.
Diagnostics not shown: 1684.
Compared 3961 files in 1655ms
Skipped 1405 files
Error: errors where emitted while running checks
6.00s user 3.51s system 32% cpu 29.004 total
```

Here most of the time is spent printing diagnostics

### `check` command
```
The number of diagnostics exceeds the number allowed by Rome.
Diagnostics not shown: 5097.
Checked 5366 files in 744ms
Error: errors where emitted while running checks
2.31s user 0.42s system 216% cpu 1.263 total
```

## `main


### `format` command

```
The number of diagnostics exceeds the number allowed by Rome.
Diagnostics not shown: 1684.
Compared 3961 files in 1614ms
Skipped 1405 files
Error: errors where emitted while running checks
6.04s user 3.46s system 33% cpu 28.442 total
```

### `check` command

```
The number of diagnostics exceeds the number allowed by Rome.
Diagnostics not shown: 5097.
Checked 5366 files in 879ms
Error: errors where emitted while running checks
2.31s user 0.42s system 201% cpu 1.356 total
```

crates/rome_service/src/workspace/server.rs
```diff
@@ -159,17 +166,59 @@ impl WorkspaceServer {
...
 
 impl Workspace for WorkspaceServer {
-    fn supports_feature(&self, params: SupportsFeatureParams) -> Result<bool, RomeError> {
+    fn supports_feature(
+        &self,
+        params: SupportsFeatureParams,
+    ) -> Result<SupportsFeatureResult, RomeError> {
         let capabilities = self.get_capabilities(&params.path);
         let settings = self.settings.read().unwrap();
+        let is_ignored = matches!(self.is_file_ignored(&params.path, &params.feature), true);
```
leops, line 198: Nit: using `matches!` to match on `true` seems pointless ?

crates/rome_service/src/workspace/server.rs
```diff
@@ -284,8 +333,8 @@ impl Workspace for WorkspaceServer {
             .lint
             .ok_or_else(self.build_capability_error(&params.path))?;
 
-        let parse = self.get_parse(params.path.clone())?;
         let settings = self.settings.read().unwrap();
+        let parse = self.get_parse(params.path.clone(), FeatureName::Lint)?;
```
leops, line 337: This should only test for `FeatureName::Lint` if `params.categories` contains `RuleCategories::LINT`, adding a file to the linter ignore list shoud probably not disable syntax errors for that file

ematipico: Please read https://github.com/rome/tools/pull/3146#discussion_r964662460

`pull_diagnostics` is also used to check for syntax errors during the formatting phase. If a file is ignored inside the formatter, we should not parse it. Do you have any suggestion for fix this?

ematipico: I was thinking something like 

```rs
        let feature = if !params.categories.is_syntax() {
            FeatureName::Format
        } else {
            FeatureName::Lint
        };
```

ematipico: Done. It sill think that `pull_diagnostics` is a special case, but for now we manage to cover all cases.

leops (approved):

PR #3149, opened 2022-09-02
test(rome_js_formatter): Remove `JsFormatOptions` `Default` implementation

MichaReiser:
This PR removes the `Default` implementation from `JsFormatOptions` because passing a `SourceType` that matches with the type of the AST is mandatory to get correct formatting results. Assuming a default isn't safe because the formatting may depend on the correct source type.

I noticed that we aren't always passing in the right source type when adding another test to only apply some expensive logic if the source type is JSX and it never took that branch.

## Tests

See now changed prettier tests

### DIFF
--- a/crates/rome_js_formatter/src/context.rs
+++ b/crates/rome_js_formatter/src/context.rs
@@ -10,7 +10,7 @@ use std::fmt::Debug;
 use std::rc::Rc;
 use std::str::FromStr;
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct JsFormatContext {
     options: JsFormatOptions,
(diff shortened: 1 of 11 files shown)


ematipico (approved):
Good catch!

ematipico:
Too late for that, but I just realized that the title should have  `fix` instead of `test`

ematipico:
> > Too late for that, but I just realized that the title should have `fix` instead of `test`
> 
> Would you mind explaining your reasoning? My reasoning was that the PR is improving the correctness of our tests and isn't fixing any production issue (only preventing issues in the future)

The changes of the snapshots. I just checked and they were buggy and now fixed.

PR #3150, opened 2022-09-02
feat(rome_js_formatter): Conditional JSX Chain layout

MichaReiser:
This PR implements Prettier's conditional JSX chain layout.

```
abcdefgh ? (
    <Element>
        <Sub />
        <Sub />
    </Element>
) : (
    <Element2>
        <Sub />
        <Sub />
    </Element2>
);
```

That parenthesizes the `consequent` and `alternate` except for `null`, `undefined`, and nested conditionals in the alternate.

This PR further removes the `Default` implementation from `JsFormatOptions` because passing a `SourceType` that matches with the type of the AST is mandatory to get correct formatting results.

## Tests

I verified the prettier snapshots and added our own tests to cover the exceptions of `null`, `undefined` and nested alternates that don't get parenthesized.

Average compatibility: 85.41 -> 85.53
Compatible lines: 83.70 -> 83.95

It's expected for the tests to fail until #3149 is merged


crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -14,20 +17,61 @@ declare_node_union! {
...
+    }
+}
+
+#[derive(Debug, Copy, Clone, Default)]
+pub struct FormatJsAnyConditionalRule {
+    /// Whether the parent is a jsx conditional chain.
+    /// Gets passed through from the root to the consequent and alternate of [JsConditionalExpression]s.
+    ///
+    /// Doesn't apply for [TsConditionalType].
+    jsx_chain: ConditionalJsxChain,
```
MichaReiser, line 42: I had to create a new `FormatJsAnyConditionalRule` so that it's possible to pass option values without cluttering the `JsAnyConditional` node. The reason for this is that it's necessary to know if a `JsxTagExpression` appears in any of the parent conditional expressions. 

Checking this for every conditional is expensive, so we only check it at the root and then pass the information down to the `consequent` and `alternate` formatting.

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -144,6 +208,246 @@ impl Format<JsFormatContext> for JsAnyConditional {
...
+    /// ).member
+    ///
+    /// // Preferred
+    /// return (
+    ///     a
+    ///         ? b
+    ///         : c
+    /// ).member
+    /// ```
+    fn should_extra_indent(
```
MichaReiser, line 269: I only moved this method from the `impl  JsAnyConditional` to implement it on the rule instead.

crates/rome_js_formatter/src/utils/conditional.rs
```diff
@@ -78,40 +122,60 @@ impl Format<JsFormatContext> for JsAnyConditional {
...
+                            conditional.colon_token().format(),
+                            space(),
+                            format_jsx_chain_alternate(alternate.as_expression().unwrap())
+                        ]
+                    )
+                }
+                _ => {
+                    // Add an extra level of indent to nested consequences.
+                    if layout.is_nested_consequent() {
+                        // if f.context().indent_style().is_tab() {
```
ematipico, line 144:
```suggestion

```

ematipico (approved):

PR #3151, opened 2022-09-02
perf(rome_service): increase the throughput of the remote daemon transport

leops:
## Summary

This PR improves the performance of the remote daemon transport, primarily by allowing both the client and server to process more requests at the same time. The changes include:
- Refactored the `WorkspaceTransport` trait to make it aware of request IDs. This lets the transport layer send out multiple requests to the remote server at the same time and process the responses out of order (this makes use of the `dashmap` crate to store the handles of pending requests in a thread-safe hashmap)
- Use the `OwnedReadHalf` and `OwnedWriteHalf` types provided by `tokio` on Unix, and manually implemented `<Client|Server>ReadHalf` and `<Client|Server>WriteHalf` types on Windows to support "true" parallel I/O operations on the transport socket, rather than relying on the `split` function provided by `tokio` (that relies on a form of spinlocking internally)
- Run the implementation of all the `rome/*` remote calls in the blocking thread pool of the server. This avoids starving the scheduler if a file takes longer to process

Additionally, I also modified the signature of the `supports_feature` method of the Workspace to return a `Result`, since it may error as any other method if there's an issue with the remote connection. Finally, I added a timeout error to all remote calls as a last-resort measure, currently the timeout is fixed to 15 seconds.

## Test Plan

These are internal-only changes that should not be visible to end users, some of these are covered by the tests for the node.js JSON-RPC bindings but the OS-specific I/O code is still mostly untested.

### DIFF
--- a/crates/rome_cli/Cargo.toml
+++ b/crates/rome_cli/Cargo.toml
@@ -33,6 +33,7 @@ serde = { version = "1.0.133", features = ["derive"] }
 serde_json = { version = "1.0.74" }
 tokio = { version = "1.15.0", features = ["io-std", "io-util", "net", "time", "rt", "rt-multi-thread", "macros"] }
 anyhow = "1.0.52"
+dashmap = "5.2.0"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.127"
--- a/crates/rome_cli/src/service/mod.rs
+++ b/crates/rome_cli/src/service/mod.rs
@@ -33,7 +33,7 @@ use tokio::{
     },
     runtime::Runtime,
     sync::{
-        mpsc::{channel, Sender},
+        mpsc::{channel, Receiver, Sender},
         oneshot,
     },
     time::sleep,
@@ -68,25 +68,43 @@ type JsonRpcResult = Result<Box<RawValue>, TransportError>;
 /// Implementation of [WorkspaceTransport] for types implementing [AsyncRead]
 /// and [AsyncWrite]
 ///
-/// This implementation makes use of two "background tasks":
-/// - the "write task" pulls outgoing messages from the "write channel" and
+/// This structs holds an instance of the `tokio` runtime, as well as the
+/// following fields:
+/// - `write_send` is a sender handle to the "write channel", an MPSC channel
+/// that's used to queue up requests to be sent to the server (for simplicity
+/// the requests are pushed to the channel as serialized byte buffers)
+/// - `pending_requests` is handle to a shared hashmap where the keys are `u64`
+/// corresponding to request IDs, and the values are sender handles to oneshot
+/// channel instances that can be consumed to fullfill the associated request
+///
+/// Creating a new `SocketTransport` instance requires providing a `tokio`
+/// runtime instance as well as the "read half" and "write half" of the socket
+/// object to be used by this transport instance. These two objects implement
+/// [AsyncRead] and [AsyncWrite] respectively, and should generally map to the
+/// same underlying I/O object but are represented as separate so they can be
+/// used concurrently
+///
+/// This concurrent handling of I/O is implemented useing two "background tasks":
+/// - the [write_task] pulls outgoing messages from the "write channel" and
 /// writes them to the "write half" of the socket
-/// - the "read task" reads incoming messages from the "read half" of the
+/// - the [read_task] reads incoming messages from the "read half" of the
 /// socket, then looks up a request with an ID corresponding to the received
 /// message in the "pending requests" map. If a pending request is found, it's
 /// fullfilled with the content of the message that was just received
 ///
 /// In addition to these, a new "foreground task" is created for each request.
-/// These tasks create a "oneshot channel" and store it in the pending requests
-/// map using the request ID as a key, then serialize the content of the
-/// request and send it over the write channel. Finally, the task blocks the
-/// current thread until a response is received over the oneshot channel from
-/// the read task, or the request times out
+/// Each foreground task creates a oneshot channel and stores it in the pending
+/// requests map using the request ID as a key, then serialize the content of
+/// the request and send it over the write channel. Finally, the task blocks
+/// the current thread until a response is received over the oneshot channel
+/// from the read task, or the request times out
 pub struct SocketTransport {
     runtime: Runtime,
     write_send: Sender<Vec<u8>>,
-    pending_requests: Arc<DashMap<u64, oneshot::Sender<JsonRpcResult>>>,
+    pending_requests: PendingRequests,
 }
+
+type PendingRequests = Arc<DashMap<u64, oneshot::Sender<JsonRpcResult>>>;
 
 impl SocketTransport {
     pub fn open<R, W>(runtime: Runtime, socket_read: R, socket_write: W) -> Self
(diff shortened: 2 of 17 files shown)


crates/rome_cli/src/service/mod.rs
```diff
@@ -32,153 +57,280 @@ pub(crate) use self::unix::{ensure_daemon, print_socket, run_daemon};
...
+        Ok(Some((read, write))) => Ok(Some(SocketTransport::open(runtime, read, write))),
         Ok(None) => Ok(None),
         Err(err) => Err(err),
     }
 }
 
+type JsonRpcResult = Result<Box<RawValue>, TransportError>;
+
 /// Implementation of [WorkspaceTransport] for types implementing [AsyncRead]
 /// and [AsyncWrite]
```
MichaReiser, line 69: It could help future mantainers if you explained the details of the socket transport a bit. What's stored in the pending requests?

crates/rome_cli/Cargo.toml
```diff
@@ -33,6 +33,7 @@ serde = { version = "1.0.133", features = ["derive"] }
 serde_json = { version = "1.0.74" }
 tokio = { version = "1.15.0", features = ["io-std", "io-util", "net", "time", "rt", "rt-multi-thread", "macros"] }
 anyhow = "1.0.52"
+dashmap = "5.2.0"
```
MichaReiser, line 36: What's the size increase of having this new dependency? Does it affect the wasm version too?

crates/rome_lsp/src/server.rs
```diff
@@ -177,13 +180,22 @@ macro_rules! workspace_method {
...
+
+                let workspace = server.session.workspace.clone();
+                let result = spawn_blocking(move || {
+                    let _guard = span.entered();
+                    catch_unwind(move || workspace.$method(params))
+                });
+
+                result.map(move |result| {
+                    match result {
+                        Ok(Ok(Ok(result))) => Ok(result),
```
MichaReiser, line 193: What's this :D Result stacking at its best :D

crates/rome_lsp/src/session.rs
```diff
@@ -119,7 +119,7 @@ impl Session {
         let lint_enabled = self.workspace.supports_feature(SupportsFeatureParams {
             feature: FeatureName::Lint,
             path: rome_path.clone(),
-        });
+        })?;
```
MichaReiser, line 122: What's the reason that `supports_feature` now returns a `Result`?

MichaReiser (approved):
> increase the throughput of the remote daemon 

how did  you test that this change is increasing the throughput or is the main goal to allow concurrent use?

crates/rome_cli/Cargo.toml
leops: In theory this change should only marginally affect the binary size since `dashmap` is already a dependency of `rome_service`. It should not affect the size of the WASM bundle since the `rome_cli` crate is not part of that distribution

crates/rome_lsp/src/server.rs
leops: The inner `Workspace` method returns a `Result<R, RomeError>`, that gets wrapped into a `Result<_, Box<dyn Any>>` by `catch_unwind`, and then again into a `Result<_, JoinError>` by `tokio::spawn`. The match is needed to unify all three error types into one, although the middle layer could now be removed since spawned tasks wrap the execution of their future in a `catch_unwind` anyway, and it's possible to retrieve the panic message from the `JoinError` instead

crates/rome_lsp/src/session.rs
leops: If the workspace client is connected to a remote server then it's possible for the request to fail because it timed out or the connection was lost. Previously the error was hidden internally and the file was reported as unsupported instead, but that leads to unclear error messages for the user as supported files can be erroneously (and randomly) reported as unsupported by the CLI

leops:
> > increase the throughput of the remote daemon
> 
> how did you test that this change is increasing the throughput or is the main goal to allow concurrent use?

Yes the main goal of the PR is to allow multiple requests to be in-flight at the same time. This leads to the following improvements to performance when running the formatter CLI in release mode on the `rome/tools` repo:

Wall time on `feature/daemon-cli`:
> Formatted 52 files in 128ms

Wall time on `perf/daemon-transport`:
> Formatted 52 files in 48ms

MichaReiser:
> Yes the main goal of the PR is to allow multiple requests to be in-flight at the same time. This leads to the following improvements to performance when running the formatter CLI in release mode on the rome/tools repo:

Nice!

Sorry for all these questions but I'm not familiar with how workspaces work. How is it that allowing multiple requests improves the CLI performance? I would have expected that the CLI makes a single request: "format files: <path>" or is it that we do the traversal inside of the CLI and then call into the deamon?

leops:
> > Yes the main goal of the PR is to allow multiple requests to be in-flight at the same time. This leads to the following improvements to performance when running the formatter CLI in release mode on the rome/tools repo:
> 
> Nice!
> 
> Sorry for all these questions but I'm not familiar with how workspaces work. How is it that allowing multiple requests improves the CLI performance? I would have expected that the CLI makes a single request: "format files: " or is it that we do the traversal inside of the CLI and then call into the deamon?

Exactly, the Workspace API currently has little to no awareness of the file system, so the traversal is handled by the CLI that issues a sequence of open -> format -> close calls for each file it process. The traversal is run in parallel on the rayon thread pool of the CLI, but since the initial implementation of the `WorkspaceTransport` did not allow for concurrent requests for simplicity this meant the traversal threads were spending most their time blocked waiting for unrelated requests to complete.

crates/rome_cli/src/service/mod.rs
```diff
@@ -32,153 +57,295 @@ pub(crate) use self::unix::{ensure_daemon, print_socket, run_daemon};
...
+/// This implementation makes use of two "background tasks":
+/// - the "write task" pulls outgoing messages from the "write channel" and
+/// writes them to the "write half" of the socket
+/// - the "read task" reads incoming messages from the "read half" of the
+/// socket, then looks up a request with an ID corresponding to the received
+/// message in the "pending requests" map. If a pending request is found, it's
+/// fullfilled with the content of the message that was just received
+///
+/// In addition to these, a new "foreground task" is created for each request.
+/// These tasks create a "oneshot channel" and store it in the pending requests
```
ematipico, line 80:
```suggestion
/// Each foreground task creates a "oneshot channel" and stores it in the pending requests
```

crates/rome_cli/src/service/mod.rs
```diff
@@ -32,153 +57,295 @@ pub(crate) use self::unix::{ensure_daemon, print_socket, run_daemon};
...
 impl SocketTransport {
-    pub fn open<T>(runtime: Runtime, socket: T) -> Self
+    pub fn open<R, W>(runtime: Runtime, socket_read: R, socket_write: W) -> Self
     where
-        T: AsyncRead + AsyncWrite + Send + 'static,
+        R: AsyncRead + Unpin + Send + 'static,
+        W: AsyncWrite + Unpin + Send + 'static,
     {
-        let (socket_read, mut socket_write) = split(socket);
+        let (write_send, mut write_recv) = channel(512);
```
ematipico, line 97: What's 512? Would it be possible to have int a `const`? The name of the variable might help.

crates/rome_cli/src/service/mod.rs
```diff
@@ -32,153 +57,295 @@ pub(crate) use self::unix::{ensure_daemon, print_socket, run_daemon};
...
         Err(err) => Err(err),
     }
 }
 
+type JsonRpcResult = Result<Box<RawValue>, TransportError>;
+
 /// Implementation of [WorkspaceTransport] for types implementing [AsyncRead]
 /// and [AsyncWrite]
+///
+/// This implementation makes use of two "background tasks":
```
ematipico, line 71: Can we give some arbitrary names to these background tasks? Naming them might help us across the codebase, so we can mention them using this internal, arbitrary names.

crates/rome_cli/src/service/mod.rs
```diff
@@ -32,153 +57,295 @@ pub(crate) use self::unix::{ensure_daemon, print_socket, run_daemon};
...
+                        response.map(|response| (message, response))
                     }
-                }
+                    Err(err) => Err(err),
+                };
+
+                let (message, response): (_, JsonRpcResponse) = match message {
+                    Ok(message) => message,
+                    Err(err) => {
+                        eprintln!(
```
ematipico, line 138: Is it OK to have a `eprintln` here?

leops: Since this is a serialization or I/O error we don't have enough information to report this error to a specific pending request (since we failed to deserialize the request ID in the first place), and since this a background task we can't easily access the `Console` object to report the error there (specifically we'd have to wrap the console in a `Mutex` to share it between this and the rest of the CLI code), so it seemed like the only option here is to print the error to stderr directly and exit the task

crates/rome_cli/src/service/mod.rs
```diff
@@ -32,153 +57,295 @@ pub(crate) use self::unix::{ensure_daemon, print_socket, run_daemon};
...
+                            "{:?}",
+                            err.context("remote connection write task exited with an error")
+                        );
+                        break;
+                    }
+                };
 
-                let length = length.context("incoming response from the remote workspace is missing the Content-Length header")?;
+                if let Some((_, channel)) = pending_requests.remove(&response.id) {
```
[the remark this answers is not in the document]
leops: The `remove` method is used to take ownership of an entry in the hashmap, if the map contains an entry with the key `response.id` then the entry is removed from the map and the method returns `Some((key, value))`. Using `remove` rather than `get` is important for two reasons, first the values of the hashmap are "oneshot channels" and as the name indicates these can only be used once (the `send` method takes `self` by value) so we need to take ownership of the channel to consume it and fulfill the request. And more generally this also ensures that we don't keep old requests around in the map by removing the corresponding entry once it gets fulfilled.

ematipico (approved):
I tested the branch locally. I tried to run the `check` command using the `--use-server` argument on the Rome classic code base, `internals` folder.

The `main` branch took 52 seconds to lint 5366 files.
This branch took 14 seconds to lint 5366 files.

Issue #3147, opened 2022-09-01
📎 Tracking tasks for implementing Instantiation expressions


IWANABETHATGUY:
### Description

From https://github.com/rome/tools/pull/3118#discussion_r958305550

- [x] https://github.com/microsoft/TypeScript/pull/47607
- [x] https://github.com/microsoft/TypeScript/pull/48659
- [x] https://github.com/microsoft/TypeScript/pull/49353

These tasks should be implemented in order.

denbezrukov:
I'm working.

PR #3152, opened 2022-09-03
feat(rome_js_parser): instantiation expressions

IWANABETHATGUY:
## Summary
1. Part of https://github.com/rome/tools/issues/3147
2. Finishing the first task, align to the implementation of Typescript https://github.com/microsoft/TypeScript/pull/47607
3. Adding this extra SyntaxKind `TsExpressionWithTypeArguments` to align typescript AST shape. 
4. Modifying the shape of `TsTypeOfType`,adding optional `type_arguments`,
now, the `TsTypeOfType` could be written like:  
    ```
    type T21 = typeof Array<string>;
    ```
5. The newly added `SyntaxKind` is used to support Typescript instantiation expression, also it would be sometimes 
ambiguous when try to parse
- `template_literal` inside `member_expression_rest` https://github.com/rome/tools/pull/3152/files#diff-1ca6d1ff770e933910c550807255d1bf03cdebbf60bf469579b528c0c28c1537R706-R709
- `new_expr` https://github.com/rome/tools/pull/3152/files#diff-1ca6d1ff770e933910c550807255d1bf03cdebbf60bf469579b528c0c28c1537R735-R769
- `call_expr_rest` https://github.com/rome/tools/pull/3152/files#diff-1ca6d1ff770e933910c550807255d1bf03cdebbf60bf469579b528c0c28c1537R1653-R1673

### Newly Supported Syntax
```ts
let f1 = fx<string>;
let f2 = fx<string, number>;
let f3 = fx['test']<string>;
const a2 = f.g<number>;  // () => number
const a3 = f<number>.g;  // <U>() => U
const a4 = f<number>.g<number>;  // () => number
const a5 = f['g']<number>;  // () => number
const a7 = (f<number>)['g'];
const a6 = f<number>['g'];  // type Error
const b2 = f?.<number>();
const b3 = f<number>?.();
const b4 = f<number>?.<number>();  // Type Error, expected no type arguments
const x1 = f<true>
(true);
// Parsed as relational expression
const x2 = f<true>
true;
// Parsed as instantiation expression
const x3 = f<true>;
true;

test ts ts_type_instantiation_expression
type StringBox = Box<string>;

test_err ts ts_instantiation_expressions1
const a8 = f<number><number>;  // Relational operator error
const b1 = f?.<number>;  // Error, `(` expected
```

## Test Plan
1. CI should pass
2. Checking the newly added snapshot

### Main
Average compatibility: 85.67
Compatible lines: 84.09

### PR
Average compatibility: 85.81
Compatible lines: 84.10


IWANABETHATGUY:
!bench_parser

xtask/codegen/js.ungram
```diff
@@ -316,6 +316,7 @@ JsAnyExpression =
 	| TsTypeAssertionExpression
```
IWANABETHATGUY, line 316: Adding this extra `SyntaxKind` to align typescript AST shape. Also, this is necessary for instantiation expression due to instantiation expression parsing sometimes needs an intermediate AST which may reparse in different contexts.

IWANABETHATGUY:
!bench_parser

MichaReiser:
Would you mind extening the PR summary by: 

* Adding some examples of the syntax that is supported now
* New AST nodes added and where they are being used. 
* Mention of breaking AST changes (`ExpressionWithArguments` was used in other contexts before)

crates/rome_js_parser/src/syntax/expr.rs
```diff
@@ -701,16 +703,35 @@ fn parse_member_expression_rest(
             BACKTICK => {
                 // test ts ts_optional_chain_call
                 // (<A, B>() => {})?.<A, B>();
-                let m = lhs.precede(p);
+                let m = match lhs.kind() {
+                    TS_EXPRESSION_WITH_TYPE_ARGUMENTS => lhs.undo_completion(p),
+                    _ => lhs.precede(p),
+                };
```
IWANABETHATGUY, line 709: https://github.dev/microsoft/TypeScript/blob/be4e9bac8ff66174d4e856dae06f69aa4ef7e479/src/compiler/parser.ts#L5893-L5899, related to this part of logic. If we found the lhs is `TsExpressionWithTypeArguments` we need to reparse the slots to `template_literal`, 
Considering example below: 
```js
html<A, B>`abcd`
```
If we don't check the kind `lhs` and always do `lhs.precede` 
the expression would be parsed as
```
            expression: JsTemplate {
                tag: TsExpressionWithTypeArguments {
                    expression: JsIdentifierExpression {
                        name: JsReferenceIdentifier {
                            value_token: IDENT@0..4 "html" [] [],
                        },
                    },
                    arguments: TsTypeArguments {
                        l_angle_token: L_ANGLE@4..5 "<" [] [],
                        ts_type_argument_list: TsTypeArgumentList [
                            TsReferenceType {
                                name: JsReferenceIdentifier {
                                    value_token: IDENT@5..6 "A" [] [],
                                },
                                type_arguments: missing (optional),
                            },
                            COMMA@6..8 "," [] [Whitespace(" ")],
                            TsReferenceType {
                                name: JsReferenceIdentifier {
                                    value_token: IDENT@8..9 "B" [] [],
                                },
                                type_arguments: missing (optional),
                            },
                        ],
                        r_angle_token: R_ANGLE@9..10 ">" [] [],
                    },
                },
                type_arguments: missing (optional),
                l_tick_token: BACKTICK@10..11 "`" [] [],
                elements: JsTemplateElementList [
                    JsTemplateChunkElement {
                        template_chunk_token: TEMPLATE_CHUNK@11..15 "abcd" [] [],
                    },
                ],
                r_tick_token: BACKTICK@15..16 "`" [] [],
            },
```

```
html<A, B>`abcd`
^^^^^^^^^^^  ^^^^ element
this part is tag
```
Actually, it should be parsed as https://www.typescriptlang.org/play?jsx=0&ts=4.7.4#code/C4TwDgpgBAQlC8UCGA7EBuAUKSUCCCyaWAFsALYA2APHgDSwB8ABkgEYDGAJs1kA


```
html<A, B>`abcd`
^^^^  ^^^^^   ^^^^ elements                 
  id     type_arguments
```

That's why we need to drop the original `CompltedMarker` and reparse it .

crates/rome_js_parser/src/syntax/expr.rs
```diff
@@ -741,27 +762,21 @@ fn parse_new_expr(p: &mut Parser, context: ExpressionContext) -> ParsedSyntax {
...
-    let expression = parse_primary_expression(p, context.and_ts_type_assertion_allowed(false))
-        .or_add_diagnostic(p, expected_expression);
-
-    if let Some(lhs) = expression {
-        parse_member_expression_rest(p, lhs, context, false, &mut false);
+    if let Some(lhs) = parse_primary_expression(p, context.and_ts_type_assertion_allowed(false))
+        .or_add_diagnostic(p, expected_expression)
+        .map(|expr| parse_member_expression_rest(p, expr, context, false, &mut false))
+    {
+        if let TS_EXPRESSION_WITH_TYPE_ARGUMENTS = lhs.kind() {
```
IWANABETHATGUY, line 769: The reason are similar with https://github.com/rome/tools/pull/3152/files#r966104100

crates/rome_js_parser/src/syntax/expr.rs
```diff
@@ -1645,15 +1660,17 @@ fn parse_call_expression_rest(
...
-        if p.at(T![?.]) {}
-
         if !matches!(p.cur(), T![?.] | T![<] | T![<<] | T!['(']) {
             break lhs;
         }
 
         // Cloning here is necessary because parsing out the type arguments may rewind in which
         // case we want to return the `lhs`.
-        let m = lhs.clone().precede(p);
+        let m = match lhs.kind() {
```
IWANABETHATGUY, line 1669: The corresponding logic of Typescript you could reference https://github.dev/microsoft/TypeScript/blob/be4e9bac8ff66174d4e856dae06f69aa4ef7e479/src/compiler/parser.ts#L5956-L5958

crates/rome_js_parser/src/syntax/expr.rs
```diff
@@ -1665,26 +1682,24 @@ fn parse_call_expression_rest(
...
-        } else if p.at(T!['(']) {
+            type_arguments
+        } else {
+            None
+        };
+
+        if type_arguments.is_some() || p.at(T!['(']) {
             parse_call_arguments(p)
-                .expect("Expected parsed out arguments because the parser is positioned at '('");
+                .or_add_diagnostic(p, |p, range| expected_node("(", range).to_diagnostic(p));
```
IWANABETHATGUY, line 1702: We can't panic now, because there exist two conditions (Or relation) would let code go inside this block.

IWANABETHATGUY: The corresponding logic in Typescript you could reference here. https://github.dev/microsoft/TypeScript/blob/be4e9bac8ff66174d4e856dae06f69aa4ef7e479/src/compiler/parser.ts#L5949-L5961

IWANABETHATGUY:
> * ExpressionWithArguments

You said that 
> ExpressionWithArguments was used in other contexts before  

But I can't search it through the codebase.

IWANABETHATGUY:
> Would you mind extening the PR summary by:
> 
> * Adding some examples of the syntax that is supported now
> * New AST nodes added and where they are being used.
> * Mention of breaking AST changes (`ExpressionWithArguments` was used in other contexts before)

Done

crates/rome_js_formatter/src/ts/types/typeof_type.rs
```diff
@@ -10,13 +10,19 @@ pub struct FormatTsTypeofType;
...
             expression_name,
         } = node.as_fields();
 
         write![
             f,
-            [typeof_token.format(), space(), expression_name.format()]
+            [
+                typeof_token.format(),
+                space(),
+                expression_name.format(),
```
MichaReiser, line 23: What's the reason that `TsTypeofType` now has a `type_arguments` field?

crates/rome_js_parser/src/syntax/expr.rs
MichaReiser: What do you mean by re-parsing. Is it because it calls `undo_completion`? `undo_completion` doesn't re-parse, it only removes the EndNode marker so that children will continue to append to the current node.

crates/rome_js_parser/src/syntax/expr.rs
MichaReiser: `expected_node` expects the name of the node rather than a token. For example, "call arguments". You can use `expected_token` if you want to require a specific token

crates/rome_js_parser/src/syntax/typescript/types.rs
```diff
@@ -1231,6 +1236,35 @@ fn parse_ts_type_predicate(p: &mut Parser) -> ParsedSyntax {
...
+// test ts ts_instantiation_expressions
+// let f1 = fx<string>;
+// let f2 = fx<string, number>;
+// let f3 = fx['test']<string>;
+// const a2 = f.g<number>;  // () => number
+// const a3 = f<number>.g;  // <U>() => U
+// const a4 = f<number>.g<number>;  // () => number
+// const a5 = f['g']<number>;  // () => number
+// const a7 = (f<number>)['g'];
+// const a6 = f<number>['g'];  // type Error
```
MichaReiser, line 1248: This test is giving me the impression that `FormatTsExpressionWithTypeArguments::needs_parentheses` must return if it appears in the object of a member expression. Other positions that are often necessary to require parens is inside of `class extends f<number>`. Can you please double check in which positions parentheses are required. (See `parentheses.rs` for helpers)

xtask/codegen/js.ungram
```diff
@@ -2109,6 +2111,10 @@ TsTypeAssertionExpression =
 	'>'
 	expression: JsAnyExpression
 
+TsExpressionWithTypeArguments =
```
MichaReiser, line 2114: The `JsNewExpression` is defined as

```
JsNewExpression =
	'new'
	callee: JsAnyExpression
	type_arguments: TsTypeArguments?
	arguments: JsCallArguments?
```

Is it possible to have a situation where the `callee` is a `TsExpressionWithTypeArguments` and the `type_arguments` are set as well? Is it still necessary to have the type arguments field on `JsNewExpression`? 

Can you (if not done already), add a test for `new A<test><test>()`

MichaReiser (commented):
The PR overall is looking good to me. But I do have a few follow up questions:

* `ExpressionWithTypeArguments`: My understanding is that this is a new expression type and represents instantiation expressions. What's your thinking behind calling it `ExpressionWithTypeArguments` instead of `TsInstantiationExpression`? Are there other places where it should be used that aren't instantiation expressions? 
* `ExpressionWithTypeArguments` should end in `Expression` to align with our naming convention that all expressions are named `...Expression`: `JsCallExpression`, `JsIdentifierExpression`, ... 
*

crates/rome_js_parser/src/syntax/expr.rs
IWANABETHATGUY: Yes, you are right. the meaning is exactly the same as you said: 
> it only removes the EndNode marker so that children will continue to append to the current node.

crates/rome_js_parser/src/syntax/expr.rs
IWANABETHATGUY: Done

IWANABETHATGUY:
> * `ExpressionWithTypeArguments`: My understanding is that this is a new expression type and represents instantiation expressions. What's your thinking behind calling it `ExpressionWithTypeArguments` instead of `TsInstantiationExpression`? Are there other places where it should be used that aren't instantiation expressions?

1. I do agree with you, but `ExpressionWithTypeArguments` is not the newly added `SyntaxKind` in `Typescript` parser, we don't have this `syntaxKind` because our parser is not a 100% port of Typescript. I add this new `SyntaxKind` because `ExpressionWithTypeArguments` are widely used in parsing `TsInstantiationExpression` (mostly as a middle SyntaxKind, maybe split in different contexts). I could give two solution
    - Aligning to typescript, keep the `ExpressionWithTypeArguments` as is, (cons: violate our naming convention)
    - Rename it to `TsInstantiationExpression`, this may increase the burden of mind of the maintainer when porting typescript in  the future
    (maybe we could add some comments.)

crates/rome_js_formatter/src/ts/types/typeof_type.rs
IWANABETHATGUY: https://github.com/microsoft/TypeScript/pull/47607/files#diff-12a6724be007eb1a19d80018c7a63bbc73525ca793a9b3e5da49a4e86bbf457cR3079-R3085, `TsTypeQuery` is the same thing as our `TsTypeofType`

xtask/codegen/js.ungram
IWANABETHATGUY: https://github.com/rome/tools/pull/3152/commits/7defc3fbf8937b2ec67d0180ecf4a0e6a8734773, Done. The ast is the same as Typescript. https://www.typescriptlang.org/play?jsx=0#code/FAOwpg7gBAggPAFzAZwQPkS9AKAlAbiA

crates/rome_js_formatter/src/ts/types/typeof_type.rs
MichaReiser: What's the relation of this change to instantiation expressions? Would it be better to move it into its own PR?

MichaReiser:
> > * `ExpressionWithTypeArguments`: My understanding is that this is a new expression type and represents instantiation expressions. What's your thinking behind calling it `ExpressionWithTypeArguments` instead of `TsInstantiationExpression`? Are there other places where it should be used that aren't instantiation expressions?
> 
>     1. I do agree with you, but `ExpressionWithTypeArguments` is not the newly added `SyntaxKind` in `Typescript` parser, we don't have this `syntaxKind` because our parser is not a 100% port of Typescript. I add this new `SyntaxKind` because `ExpressionWithTypeArguments` are widely used in parsing `TsInstantiationExpression` (mostly as a middle SyntaxKind, maybe split in different contexts). I could give two solution
>        
>        * Aligning to typescript, keep the `ExpressionWithTypeArguments` as is, (cons: violate our naming convention)
>        * Rename it to `TsInstantiationExpression`, this may increase the burden of mind of the maintainer when porting typescript in  the future
>          (maybe we could add some comments.)

I'm in favour of renaming the node to `TsInstantiationExpression` for the following reasons:
* It aligns with our `*Expression` naming convention
* I can google for typescript instantiation expression and find resources about that syntax. Searching for expressions with type arguments doesn't yield any useful search results. 
* TypeScript may mainly use the `ExpressionWithTypeArguments` name for historical reasons. It used to be an existing node for them. We don't have this historical context and can, therefore, more freely pick a new. 
* Our AST already differs in many places from TypeScript's because of our decision to use a CST and explicitly model required children. Being compatible with existing AST's hasn't been a goal yet, which is why I consider it OK to diverge from TypeScript.

crates/rome_js_formatter/src/ts/types/typeof_type.rs
IWANABETHATGUY: Move it to another pull request need both rework of rome stuff(re-review the CST changes) and me, and doesn't introduce too much benefit. Since this pull request mostly ports the same logic from typescript, these changes should not be a big deal.

IWANABETHATGUY:
> 

Done https://github.com/rome/tools/pull/3152/commits/d2d9e60e3441b6a9f6c57a07ce66c6b946972fe8

crates/rome_js_formatter/src/ts/types/typeof_type.rs
MichaReiser: Alright. I'll do another review then of this PR looking for changes related to `TsTypeQuery`. Would you mind extending the PR summary to include this change together with the reason why it's necessary (a short sample using the syntax)

crates/rome_js_parser/src/syntax/typescript/types.rs
```diff
@@ -525,6 +527,9 @@ fn parse_ts_typeof_type(p: &mut Parser) -> ParsedSyntax {
     let m = p.start();
     p.expect(T![typeof]);
     parse_ts_name(p).or_add_diagnostic(p, expected_identifier);
+    if !p.has_preceding_line_break() {
```
MichaReiser, line 530: If it doesn't already exist, would you mind adding a test where there's a line break between the type name and arguments. I think ideally an example using JSX

```javascript
type X = typeof Array
<div>a</div>;
```

MichaReiser (commented):
Thank you for the updates.

I think we're good to merge this PR after you included the `TsTypeofType` changes in the PR summary, add a new test for a `TsTypeofType` with a line break before the following `<`, and once the `NeedsParentheses` implementation has been double checked in which cases parentheses are necessary.

crates/rome_js_parser/src/syntax/typescript/types.rs
IWANABETHATGUY: Done, same result as typescript https://www.typescriptlang.org/play?#code/C4TwDgpgBAGlC8VSQPYDMoEEBO2CGIAUADwAmAlgG4B8exA9BTQNxA

crates/rome_js_formatter/src/ts/types/typeof_type.rs
IWANABETHATGUY: https://github.com/microsoft/TypeScript/pull/47607#:~:text=The%20argument%20to%20the%20typeof%20type%20operator%20can%20now%20be%20an%20instantiation%20expression.%20For%20example%3A, I think this is why we need to update of `TsTypeQuery`

IWANABETHATGUY:
Let me explain these two commits:
https://github.com/rome/tools/pull/3152/commits/f45c26fe07e4080443a5087e925b0944ba0fb02a
https://github.com/rome/tools/pull/3152/commits/79d2af3e6a9457618bfa8d4c8de02f235032ea20

xtask/codegen/js.ungram
```diff
@@ -316,10 +316,33 @@ JsAnyExpression =
 	| TsTypeAssertionExpression
 	| TsAsExpression
 	| TsNonNullAssertionExpression
+	| TsInstantiationExpression
 	// JSX
 	| JsxTagExpression
 
 
+// reference https://github.dev/microsoft/TypeScript/blob/77374732df82c9d5c1319677dc595868bbc648b5/src/compiler/utilitiesPublic.ts#L1554-L1590
```
IWANABETHATGUY, line 324: This is how typescript parse `TsinstantiationExpression` https://github.dev/microsoft/TypeScript/blob/77374732df82c9d5c1319677dc595868bbc648b5/src/compiler/parser.ts#L7505-L7513, 
expression would be a `LeftHandSideExpression` https://github.dev/microsoft/TypeScript/blob/77374732df82c9d5c1319677dc595868bbc648b5/src/compiler/parser.ts#L5357, 
That's why I need to add these constraints rather than `JsAnyExpression`

crates/rome_js_formatter/src/ts/expressions/instantiation_expression.rs
```diff
@@ -0,0 +1,28 @@
...
+            arguments,
+        } = node.as_fields();
+
+        write![f, [expression.format(), arguments.format()]]
+    }
+}
+
+impl NeedsParentheses for TsInstantiationExpression {
+    fn needs_parentheses_with_parent(&self, parent: &rome_js_syntax::JsSyntaxNode) -> bool {
+        unary_like_expression_needs_parentheses(self.syntax(), parent)
```
IWANABETHATGUY, line 26: Why I write the `needs_parenthesis_with_parent like this?
1. here is how typescript parse `PostUpdateExpression` https://github.dev/microsoft/TypeScript/blob/77374732df82c9d5c1319677dc595868bbc648b5/src/compiler/parser.ts#L5345-L5354, almost the same as how they parse `ExpressionWithTypeArguments` https://github.dev/microsoft/TypeScript/blob/77374732df82c9d5c1319677dc595868bbc648b5/src/compiler/parser.ts#L7504-L7514. 

2. So I think the `needs_parentheses_with_parent` should be the same as `JsPostUpdateExpression` https://github.com/rome/tools/blob/1d9d7a0a273179867d533b7f9ac9e9124fa6b086/crates/rome_js_formatter/src/js/expressions/post_update_expression.rs#L26-L30

IWANABETHATGUY:
@MichaReiser , All issues has been addressed.

xtask/codegen/js.ungram
MichaReiser: I don't understand this explanation. Where's `JsLeftHandSideExpression` used and what for. Would it be possible to avoid this new node because: 

* Some variants are no expressions `JsPrivateName`, `JsxEelment`, `JsxSelfClosingElement`, `JsxFragment`. 
* Node should be named `JsAnyLeftHandSideExpression`
* This models the EcmaScript AST which we intentionally did not as it can be too fine grained to work with.

That's why I would prefer to undo this new expression except there are very good reasons to keep it.

IWANABETHATGUY: Ok, I will remove it.

MichaReiser (approved):

PR #3155, opened 2022-09-05
ci: stale workflow

ematipico:
## Summary

This PR implements a simple cron job to automatically mark PRs/issues in stale state using the 'S-Stable' label. Then, the action proceeds to close them if there's further inactivity.

PRs with the label 'PR: on hold' are exempt. 

## Test Plan

Merge it and wait. The example was copied from the [repository of the action](https://github.com/actions/stale)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,7 +1,7 @@
 name: 'Close stale issues and PRs'
 on:
   schedule:
-    - cron: '30 1 * * *'
+    - cron: '00 12 * * 1-5'
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v5
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          stale-pr-message: 'This PR is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 7 days.'
+          close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
+          close-pr-message: 'This PR was closed because it has been stalled for 7 days with no activity.'
```
MichaReiser, line 15: My impression was that we discussed to not automatically close issues or PRs but instead review them manually (by looking at all issues/PRs with the stale label). 

What's your reasoning for for adding the automatic closing of issues and PRs

.github/workflows/stale.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
```
MichaReiser, line 4: Can we pick a time that isn't in the middle of the night :) I'm aware that this won't be possible for everyone. Might also be nice to not trigger this on weekends to avoid unnecessary notifications

ematipico: What would you suggest?

.github/workflows/stale.yml
ematipico: Completely forgot. I will revert the change

.github/workflows/stale.yml
MichaReiser: How about running the workflow around the time we have our standup, if possible monday-friday only?

MichaReiser (approved):

Issue #2571, opened 2022-05-10
Formatter: Performance [Stretch]


MichaReiser:
This issue is mainly a place where I collect ideas on how we can improve the Formatter's performance (probably in significant ways). 

## Flat `FormatElement` Structure
I expect this to improve the performance because it can greatly reduce the number of allocations the formatter does. 

What's this about. The idea is to change the nested `FormatElement` structure to a flat structure:

* `Empty` -> `[]`
* `List(Vec(Elements))` -> `[...Elements]`
* `Group(content)` -> `[StartGroup, ...Content, EndGroup]`
* `Fill(elements)` -> `[StartFill, ...elements, EndFill]`
* `Indent(content)` -> `[StartIndent, ...content, EndFill]` 

**Why should this improve performance**:

* It removes the need to perform heap allocations for `Groups`, and `Indent`. 
* The `Printer` get's a single continuous array with elements. Removes the deref `Content` / Vec pointers. May even be possible to write the `Printer` in a way that it no longer needs it's inner queue and instead can just use a pointer to the position in the input document.

**Challenges**

The main challenges will be 
* Rewriting the `Printer` to work with a flat list
* Rewriting `split_trivia` to work efficiently. (Maybe no longer needed after re-writting the comment formatting?)

## Shared `FormatElement` Buffer #2634

The gist of the idea is to change `Format` to no longer return the formatted document but instead make the `format` method write to a mutable `formatter.

```rust 
// Before 
fn format(&self, formatter: &Formatter) -> FormatResult<Document> {
  return formatted![formatter, self.if_token().format(), space_token(), self.l_paren().format()]
}

// After
 fn format(&self, formatter: &mut Formatter) -> FormatResult<Document> {
  // Writes the format elements to `formatter` that maintains a shared buffer 
  write!(formatter, self.if_token().format(), space_token())?;
  write!(formatter, self.l_paren().format())?;
}
```

This may look familiar to you. And there's good reason why, this design is heavily inspired by Rust's `format`, `format_args`, and `write`. 

**How does this improve performance**

* Reduces the heap allocation from one allocation per `format` call to 1 for formatting the whole document (excluding the heap allocations happening inside of a specific `Format` implementation)
* ... I think that's more than reason enough

**Why I'm excited about this**

I haven't ran any benchmarks myself but there exists a benchmark for rust template libraries and the interesting thing is that you can group the libs in two groups:

* Libraries that do heap allocations
* Libraries that don't do heap allocations

[...]

PR #3156, opened 2022-09-05
perf(rome_formatter): Add `Buffer.elements`

MichaReiser:
## Summary

It turns out, that using the `PreambleBuffer` in hot code paths is rather expensive. Or in general, nesting `Buffer`s can be rather expensive. 

I discovered this after I implemented a prototype of a Flat IR for our formatter (#2571) where this problem becomes more prominent because `group`, `indent` etc are no longer "breaking" the buffer chain by writing into a new `VecBuffer`. 

This PR introduces a new `Buffer.elements()` method that gives access to a slice of all written elements (that are part of this buffer). This allows introspection of the written content very cheaply (may require iteration, but most inspections are only interested in the last element) and in a more natural way (writing custom buffers or even using `memoized` often feels overly complicated). Another benefit of `elements` over using `inspect` or a custom buffer is that it guarantees that rewinding the buffer is correctly taken into account, something that most uses of `inspect` didn't.

The new `elements` method allows removing most uses of the `PreambleBuffer` by doing a simple `len` comparison or inspecting the slice content. 

This PR deletes the unused `WillBreakBuffer` and `HaslLabelBuffer` buffers in favour of either `Memoized.inspect` or using `buffer.elements()`. We can re-introduce the buffers if it is indeed necessary for some logic to check if any element breaks. 

The downside of this is that it slightly leaks the `Buffer` extraction as it forces the `Buffer` to now always keep hold of all written elements. For example, this prevents us from having a Printer that prints IR elements as they get formatted. I don't think this is a very practical use case with our current Formatter design as it would mean that the printer has to rewind every time the formatter rewinds... meaning that rewinding will result in more work that must the thrown out the window (and overhead for creating the snapshots). 

I even considered changing `Formatter` to always store a `VecBuffer` but there are still a few use cases where a custom `Buffer` can be useful and we should continue supporting those. Buffers on their own aren't a problem, only if they get insanely stacked. For example, it would probably be better to apply a `GroupsElementsBuffer` once in a world where we have a FlatIR rather than nesting it over and over again.

## Test Plan

`cargo tetst`

The performance win for our nested IR are less impressive. But it improves performance by roughly 20-25%  for a flat IR and more improvements are possible once the `GroupsElementsBuffer` is no longer necessary. 

```
group                                    main                                   no-preamble
-----                                    ----                                   -----------
formatter/checker.ts                     1.05   268.4±20.76ms     9.7 MB/sec    1.00   256.6±20.61ms    10.1 MB/sec
formatter/compiler.js                    1.07    150.9±7.27ms     6.9 MB/sec    1.00    141.2±8.06ms     7.4 MB/sec
formatter/d3.min.js                      1.00    120.0±3.83ms     2.2 MB/sec    1.00    119.7±1.18ms     2.2 MB/sec
formatter/dojo.js                        1.12      8.2±0.57ms     8.3 MB/sec    1.00      7.4±0.03ms     9.3 MB/sec
formatter/ios.d.ts                       1.06   179.7±20.03ms    10.4 MB/sec    1.00   170.0±18.19ms    11.0 MB/sec
formatter/jquery.min.js                  1.02     33.7±0.72ms     2.5 MB/sec    1.00     32.9±1.72ms     2.5 MB/sec
formatter/math.js                        1.13   287.2±15.41ms     2.3 MB/sec    1.00   254.5±23.02ms     2.5 MB/sec
formatter/parser.ts                      1.04      5.4±0.20ms     9.0 MB/sec    1.00      5.2±0.08ms     9.3 MB/sec
formatter/pixi.min.js                    1.00    132.6±2.72ms     3.3 MB/sec    1.04    138.1±6.74ms     3.2 MB/sec
formatter/react-dom.production.min.js    1.06     41.3±2.90ms     2.8 MB/sec    1.00     39.1±0.42ms     2.9 MB/sec
formatter/react.production.min.js        1.03      2.1±0.15ms     3.0 MB/sec    1.00  1990.2±117.48µs     3.1 MB/sec
formatter/router.ts                      1.01      4.2±0.10ms    14.8 MB/sec    1.00      4.1±0.09ms    14.9 MB/sec
formatter/tex-chtml-full.js              1.11   337.5±24.09ms     2.7 MB/sec    1.00   303.4±14.37ms     3.0 MB/sec
formatter/three.min.js                   1.04    152.5±3.59ms     3.9 MB/sec    1.00    146.1±2.81ms     4.0 MB/sec
formatter/typescript.js                  1.02  1008.4±46.15ms     9.4 MB/sec    1.00   987.6±59.44ms     9.6 MB/sec
formatter/vue.global.prod.js             1.01     51.1±1.50ms     2.4 MB/sec    1.00     50.6±0.59ms     2.4 MB/sec

```

### DIFF
--- a/crates/rome_formatter/src/buffer.rs
+++ b/crates/rome_formatter/src/buffer.rs
@@ -4,6 +4,7 @@ use crate::{Format, FormatResult, FormatState};
 use std::any::{Any, TypeId};
 use std::fmt::Debug;
 use std::ops::{Deref, DerefMut};
+use std::ptr::NonNull;
 
 /// A trait for writing or formatting into [FormatElement]-accepting buffers or streams.
 pub trait Buffer {
--- a/crates/rome_formatter/src/buffer.rs
+++ b/crates/rome_formatter/src/buffer.rs
@@ -1,5 +1,5 @@
 use super::{write, Arguments, FormatElement};
-use crate::format_element::{LabelId, List};
+use crate::format_element::List;
 use crate::{Format, FormatResult, FormatState};
 use std::any::{Any, TypeId};
 use std::fmt::Debug;
@@ -29,6 +29,12 @@ pub trait Buffer {
     /// ```
     ///
     fn write_element(&mut self, element: FormatElement) -> FormatResult<()>;
+
+    /// Returns a slice containing all elements written into this buffer.
+    ///
+    /// Useful to inspect if some content breaks, has a specific label or if some format
+    /// did write any content.
+    fn elements(&self) -> &[FormatElement];
 
     /// Glue for usage of the [`write!`] macro with implementors of this trait.
     ///
@@ -132,6 +138,10 @@ impl<W: Buffer<Context = Context> + ?Sized, Context> Buffer for &mut W {
         (**self).write_element(element)
     }
 
+    fn elements(&self) -> &[FormatElement] {
+        (**self).elements()
+    }
+
     fn write_fmt(&mut self, args: Arguments<Context>) -> FormatResult<()> {
         (**self).write_fmt(args)
     }
@@ -224,6 +234,10 @@ impl<Context> Buffer for VecBuffer<'_, Context> {
         }
 
         Ok(())
+    }
+
+    fn elements(&self) -> &[FormatElement] {
+        &self
     }
 
     fn state(&self) -> &FormatState<Self::Context> {
@@ -349,6 +363,10 @@ where
         }
     }
 
+    fn elements(&self) -> &[FormatElement] {
+        self.inner.elements()
+    }
+
     fn state(&self) -> &FormatState<Self::Context> {
         self.inner.state()
     }
@@ -408,6 +426,10 @@ where
         self.inner.state_mut()
     }
 
+    fn elements(&self) -> &[FormatElement] {
+        self.inner.elements()
+    }
+
     fn snapshot(&self) -> BufferSnapshot {
         self.inner.snapshot()
     }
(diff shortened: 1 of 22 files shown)


MichaReiser:
!bench_formatter

crates/rome_js_formatter/tests/specs/prettier/js/module-blocks/comments.js.snap
```diff
@@ -34,8 +34,7 @@ const m2 = module/* B1 */{
    export { foo }; /*A5*/
    /*A6*/
 -}; /*A7*/ /*A8*/
-+} /*A7*/
-+/*A8*/
++} /*A7*/ /*A8*/
```
MichaReiser, line 37: I'll have a look at the comments as part of the comments refactoring (which is why this PR shouldn't be part of the 0.9 release)

crates/rome_js_formatter/src/jsx/tag/element.rs
```diff
@@ -38,9 +38,12 @@ impl Format<JsFormatContext> for JsxAnyTagWithChildren {
...
 
             ElementLayout::Default => {
-                let mut format_opening = format_opening.memoized();
+                let opening_start = f.elements().len();
+                write!(f, [format_opening])?;
 
-                let opening_breaks = format_opening.inspect(f)?.will_break();
+                let opening_breaks = f.elements()[opening_start..]
+                    .iter()
+                    .any(FormatElement::will_break);
```
ematipico, line 46: I am conflicted about this code. The new version _seems_ an hack, maybe due to the lack of documentation, maybe the they it's used, or maybe because the memoization is not there anymore. I'd rather prefer to see an API, which also does some safe check.

For the time being, can we have at least a `// SAFETY` comment on `f.elements()[opening_start..]`?

crates/rome_js_formatter/src/js/lists/module_item_list.rs
```diff
@@ -11,7 +11,14 @@ impl FormatRule<JsModuleItemList> for FormatJsModuleItemList {
...
         for module_item in node {
-            join.entry(module_item.syntax(), &format_or_verbatim(&module_item));
+            match module_item {
+                JsAnyModuleItem::JsAnyStatement(JsAnyStatement::JsEmptyStatement(empty)) => {
+                    join.entry_no_separator(&empty.format());
+                }
+                _ => {
+                    join.entry(module_item.syntax(), &format_or_verbatim(&module_item));
+                }
+            }
```
ematipico, line 21: What did we change/fix here?

crates/rome_formatter/src/builders.rs
```diff
@@ -1417,6 +1452,14 @@ impl<Context> Buffer for GroupBuffer<'_, Context> {
         Ok(())
     }
 
+    fn elements(&self) -> &[FormatElement] {
+        if self.content.is_empty() {
+            self.inner.elements()
+        } else {
+            &self.content
+        }
+    }
```
ematipico, line 1461: This function might return difference slices that have different lengths. Is it safe for a consumer to rely on the `len` of these elements? 
If we can see it from another point of view, do we have cases where a consumer needs the `content` instead of `inner.elements()` or vice-versa ?

ematipico (approved):

crates/rome_js_formatter/src/jsx/tag/element.rs
MichaReiser: Why do you consider `f.elements()[opening_start..]` a hack? It's selecting a sub slice and is safe API.

crates/rome_js_formatter/src/js/lists/module_item_list.rs
MichaReiser: We don't want to insert empty lines if it is an empty statement because that will get removed. This was previously automatically handled inside of `entry` but turns out to be rather expensive. So it's better to explicitly handle it where it is needed.

crates/rome_formatter/src/builders.rs
MichaReiser: `GroupElementsBuffer` is a hack :D and I deleted it in my local `Comments` branch. This PR also works much better once we have a flat IR because it avoids these weird situations... 

We could probably change it for now to always return `self.content` because no use case is interested in retrieving comments. It would then, at least be safe to hold on to the length (I should probably add documentation to `elements` saying that `elements` is only allowed to grow when called on the same buffer).

crates/rome_js_formatter/src/jsx/tag/element.rs
MichaReiser: I introduced a new `start_recording` method on `Buffer` that ensures an empty slice is returned in case someone rewinds a buffer. 

It is not entirely bullet proof in cases where someone rewinds a buffer, and then writes into it again so that `elements` contains at least as many elements before. It then won't return an empty slice. 

I think that's fine as snapshot are only used rarely and there is some unsafeness about snapshots in general (applying them out of order panics, applying them with the wrong buffer panics etc).

I hope this new API addresses your concerns

```rust
        let mut recording = f.start_recording();
        write!(recording, [self.content])?;

        let is_unknown = recording
            .stop()
            .last()
            .map_or(false, |last| match last.last_element() {
                Some(FormatElement::Verbatim(elem)) => elem.is_unknown(),
                _ => false,
            });
```

crates/rome_formatter/src/buffer.rs
```diff
@@ -427,227 +449,110 @@ pub trait BufferExtensions: Buffer + Sized {
...
+    pub fn write_element(&mut self, element: FormatElement) -> FormatResult<()> {
+        self.buffer.write_element(element)
     }
-}
 
-struct HasLabelledSnapshot {
-    inner: BufferSnapshot,
-    has_label: bool,
-}
+    pub fn stop(self) -> Recorded<'buf> {
```
ematipico, line 534: Why do we return a newtype here?

ematipico (approved):

crates/rome_formatter/src/buffer.rs
MichaReiser: No very specific reason. Felt more complete if `stop` returns a `Recorded` than a slice directly. The new type allows us to implement additional helpers in the future

PR #3157, opened 2022-09-05
chore(rome_service): remove constraints from settings

ematipico:
## Summary

This PR is a follow up of https://github.com/rome/tools/pull/3123

This PR removes all the constraints that we have inside the `WorkspaceSettings`. Now the settings are not publicly exposed anymore, so we can remove all the serde/schemars macros. This will allow to have more choice with the data structures we want to choose.

## Test Plan

This is an internal refactor, so the existing CI should pass.

### DIFF
--- a/crates/rome_cli/src/commands/format.rs
+++ b/crates/rome_cli/src/commands/format.rs
@@ -87,10 +87,10 @@ pub(crate) fn apply_format_settings_from_cli(
 
     match indent_style {
         Some(IndentStyle::Tab) => {
-            workspace_settings.format.indent_style = Some(IndentStyle::Tab);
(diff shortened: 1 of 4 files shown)


MichaReiser:
Would you mind explaining what you mean by "constraints"

> removes all the constraints

ematipico:
> Would you mind explaining what you mean by "constraints"
> 
> > removes all the constraints

`WorkspaceSettings` was tagged with `serde::Deserialize` and `serde::Serialize`, which means we could only choose data structures that are already serializable/de-serializable or data structures that require custom function to serialize/de-serialize them. So in a sense it was a constrain/bump.

Now that we don't need serde here, we can choose all the data structures that we see fit without worrying about their serialization/de-serialization.

PR #3159, opened 2022-09-05
fix(rome_service): correctly compute rules

ematipico:
## Summary

This PR  adds a `rome.json` file to the root of our project and fixes a bug I found.

While I was enabling the non-recommended rules in our codebase, I found out that a certain configuration was not correctly computing the rule to run. This PR applies a fix.

The reason why I added a `rome.json` file to out repository is because at the moment there are some rules that are not run by default, and in order to dogfood rome, we should be enable all the rules if possible. This will help us to find early bugs.

## Test Plan

I verified that the updated tests are correct. I also added a new assertion. I manually verified in the VSCode extension that the enabled rules are applied, as well as the recommended rules.

### DIFF
--- a/crates/rome_cli/tests/main.rs
+++ b/crates/rome_cli/tests/main.rs
@@ -522,6 +522,17 @@ mod check {
                 .count(),
             1
         );
+        assert_eq!(
+            messages
+                .iter()
+                .filter(|m| m.level == LogLevel::Error)
+                .filter(|m| {
+                    let content = format!("{:?}", m.content);
+                    content.contains("js/noUnusedVariables")
+                })
+                .count(),
+            1
+        );
 
         assert_cli_snapshot(module_path!(), "upgrade_severity", fs, console);
     }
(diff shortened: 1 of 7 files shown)


MichaReiser:
>  I found out that a certain configuration was not correctly computing the rule to run. This PR applies a fix.

Would you mind explaining what configuration was not applied correctly. Knowing that would simplify reviewing the PR

ematipico:
> > I found out that a certain configuration was not correctly computing the rule to run. This PR applies a fix.
> 
> Would you mind explaining what configuration was not applied correctly. Knowing that would simplify reviewing the PR

The configuration wrote in the `rome.json` file of this PR. And the configuration of the [updated snapshot](https://raw.githubusercontent.com/rome/tools/main/crates/rome_cli/tests/snapshots/main_check/upgrade_severity.snap).

ematipico:
It seems that there's a bunch of [false positives](https://github.com/rome/tools/runs/8188485278?check_suite_focus=true) around `js/useCamelCase` rule. I am going to turn it off for the time being, and enable it again once we fix it.

PR #3163, opened 2022-09-06
chore: Replace `num_cpus` crate with `std` call

MichaReiser:
Use `std::thread::available_parallelism()` over `num_cpus` in `xtask coverage`

## Test Plan

`xtask coverage`

### DIFF
--- a/xtask/coverage/src/lib.rs
+++ b/xtask/coverage/src/lib.rs
@@ -140,7 +140,7 @@ pub fn run(
         reporter: &mut reporters,
         pool: &yastl::Pool::new(
             std::thread::available_parallelism()
-                .map_or(2, |num| usize::from(num))
+                .map_or(2, usize::from)
                 .max(2),
         ),
     };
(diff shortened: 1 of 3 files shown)


xtask/coverage/src/lib.rs
```diff
@@ -138,7 +138,11 @@ pub fn run(
     let mut context = TestRunContext {
         filter: filter.map(|s| s.to_string()),
         reporter: &mut reporters,
-        pool: &yastl::Pool::new(num_cpus::get().max(2)),
+        pool: &yastl::Pool::new(
+            std::thread::available_parallelism()
+                .map_or(2, |num| usize::from(num))
+                .max(2),
```
ematipico, line 144: If we use `max` as limit, is it possible to just use `map`?

ematipico (approved):

xtask/coverage/src/lib.rs
MichaReiser: The max is required so that we still spawn at least two threads if `available_parallelism` returns 1 because the testing framework has one "orchestration" thread and at least one to run the tests.

PR #3171, opened 2022-09-07
fix(rome_cli): do not pull diagnostics if linter is turned off

### LINKED ISSUE #3167: rome ci does not care about rome.json
### Environment information

```bash
node 18.8.0
macos latest version
rome 0.9.0 - installed using pnpm add -D rome@next
```

### What happened?

`npm run format:ci` which is `rome ci .` does not care about rome.json configuration. 

there isn't any clear way of running formatter on CI and throwing errors if it fails.

```json
{
  "formatter": {
    "indentStyle": "space",
    "indentSize": 2
  },
  "linter": {
    "enabled": false
  }
}
```

Screen Shot 2022-09-06 at 12 29 02 PM

### Expected result

It should care linter enabled false configuration and not check for linting.

### Code of Conduct

ematipico:
## Summary

This PR fixes #3167 

Diagnostics were pulled regardless of the capability of the workspace. I added a small condition before the pulling of diagnostics.

## Test Plan

I created two test cases: one where the formatter is enabled and linter disabled, one where the formatter is disabled and the linter enabled

### DIFF
--- a/crates/rome_cli/src/traversal.rs
+++ b/crates/rome_cli/src/traversal.rs
@@ -599,26 +599,20 @@ fn process_file(ctx: &TraversalOptions, path: &Path, file_id: FileId) -> FileRes
             return Ok(FileStatus::Ignored);
         }
 
-        let categories = if ctx.execution.is_format() {
+        let categories = if ctx.execution.is_format() || !can_lint {
             RuleCategories::SYNTAX
         } else {
             RuleCategories::SYNTAX | RuleCategories::LINT
         };
 
-        let (diagnostics, has_errors) = if can_lint {
-            let result = file_guard
-                .pull_diagnostics(categories)
-                .with_file_id_and_code(file_id, "Lint")?;
+        let result = file_guard
+            .pull_diagnostics(categories)
+            .with_file_id_and_code(file_id, "Lint")?;
 
-            let has_errors = result
-                .diagnostics
-                .iter()
-                .any(|diag| diag.severity >= Severity::Error);
-
-            (result.diagnostics, has_errors)
-        } else {
-            (vec![], false)
-        };
+        let has_errors = result
+            .diagnostics
+            .iter()
+            .any(|diag| diag.severity >= Severity::Error);
 
         // In formatting mode, abort immediately if the file has errors
         match ctx.execution.traversal_mode() {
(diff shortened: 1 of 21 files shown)


crates/rome_cli/tests/snapshots/main_ci/ci_does_not_run_formatter.snap
```diff
@@ -0,0 +1,24 @@
+---
```
ematipico, line 1: This test case is valid. `rome ci` **doesn't** attempt to format files if they contain diagnostics, so I created a snippet that doesn't emit lint diagnostics.

MichaReiser: What's the motivation of not formatting files that have lint diagnostics? Shouldn't we show a user all errors to avoid the "Rust" experience where you fix a syntax error, to then get a compile error, to then get a borrow checker error.

crates/rome_cli/src/traversal.rs
```diff
@@ -604,14 +605,20 @@ fn process_file(ctx: &TraversalOptions, path: &Path, file_id: FileId) -> FileRes
...
+                .any(|diag| diag.severity >= Severity::Error);
 
-        let has_errors = result
-            .diagnostics
-            .iter()
-            .any(|diag| diag.severity >= Severity::Error);
+            (result.diagnostics, has_errors)
+        } else {
+            (vec![], false)
+        };
```
leops, line 621: Calling into `pull_diagnostics` regardless of whether the file supports linting or not is intentional since this method doesn't just return the diagnostics emitted by the lint rules, but also any syntax error that was reported by the parser or the syntax-checking pass of the analyzer. Instead the linter diagnostics should be disabled by not including the `RuleCategories::LINT` in the requested categories if linting is not supported, this way the file will still fail processing (or be skipped depending on user settings) if it contains a syntax error

crates/rome_cli/tests/snapshots/main_ci/ci_does_not_run_formatter.snap
MichaReiser: Sorry. I think I misunderstood. 

We do format and check files that have lint errors but we don't format files that have syntax errors. That makes sense.

leops (changes requested):

leops (approved):

anonrig:
@ematipico When can you release a new version with this fix? This is quite an important issue

ematipico:
> @ematipico When can you release a new version with this fix? This is quite an important issue

After we fix https://github.com/rome/tools/issues/3175, for sure. Probably tomorrow

Issue #3164, opened 2022-09-06
📎 Add section in the website to explain suggested/safe fixes


ematipico:
### Description

We should add a section where we explain this difference. While some members of the team take this for granted, users should be instructed and it makes sense to have a section on our website.

PR #3173, opened 2022-09-07
docs(website): add section to explain how fixes work

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3164

## Test Plan

Rendered website

### DIFF
--- a/website/src/_includes/docs/linter.md
+++ b/website/src/_includes/docs/linter.md
@@ -91,5 +91,5 @@ Lint rules may provide automatic code fixes. Rome distinguishes between two type
 Safe fixes are guaranteed to not change the semantics of your code,
 and can be applied without explicit review.
 
-Suggested fixes may change the semantics of your program, and it's,
-therefore,
+Suggested fixes may change the semantics of your program, and it's, 
+therefore, advised to manually review the changes. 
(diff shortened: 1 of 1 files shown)


website/src/_includes/docs/linter.md
```diff
@@ -50,3 +50,19 @@ All rules are enabled by default, and cannot be disabled. [Suppressions](#suppre
...
+It's important to understand the nature of these fixes, and be aware of them.
+
+Rome rules will emit two types of fixes:
+- safe fixes
+- suggested fixes
+
+Safe fixes can be applied safely, without the worry of changing the semantics of your code. 
+
+Suggested fixes are changes suggested to the user, but they should not be accepted blindly. The reason
+why these category of fixes is not safe, is mostly due to the nature of JavaScript and its side effects.
```
MichaReiser, line 68: Suggested fixes are fixes where applying them may change the semantics of your program and it's, therefore, advised to manually review the changes. 

Not quiet sure how we should write that last sentence explaining why some fixes are suggested only. This isn't a problem specific to JS only. It's an overall limitation that proofing the exact behaviour of any turning machine is non decidable. It just happens that this is more often the case for JavaScript than other languages but even Rust suffers from the undecidability (which is why some clips fixes are suggested only). My recommendation would be to leave the sentence away as the fact that the fix may change the semantics of the program summarises the reason very well.

PR #3174, opened 2022-09-07
chore(website): add RSS feed

ematipico:
## Summary

This PR adds a simple RSS feed to the website. The only issue I found it that we are not able to actually put the whole content of a blog post due to some malformed HTML. 

As alternative, I put the description inside the content

## Test Plan

The rendered website should create a new `/feed.xml` page

### DIFF
--- a/website/.eleventy.js
+++ b/website/.eleventy.js
@@ -13,6 +13,7 @@ const terser = require("terser");
 const CleanCSS = require("clean-css");
 const htmlmin = require("html-minifier");
 const { base64Encode } = require("./utils");
+const pluginRss = require("@11ty/eleventy-plugin-rss");
 
 require("dotenv").config();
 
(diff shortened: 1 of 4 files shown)


MichaReiser:
What's the use case for the RSS feed? What's our goal?

ematipico:
> What's the use case for the RSS feed? What's our goal?

On twitter some user asked us if the blog has an RSS/Atom feed, because they would like to subscribe to it. I don't see any reason to say no, and the plugin is very simple to integrate.

MichaReiser (approved):

website/.eleventy.js
```diff
@@ -198,6 +200,10 @@ module.exports = function (eleventyConfig) {
 		return string.toLowerCase().replace(/\s/g, "-");
 	});
 
+	eleventyConfig.addFilter("withAbsoluteUrl", function (string) {
+		return "https://rome.tools" + string;
```
MichaReiser, line 204: Is it safe to manually concatenated the string passed here? Or should we use some URL module?

ematipico: I think so, the string coming from the filter is the permalink specified in the markdown of the blog post

PR #3176, opened 2022-09-07
fix(rome_cli): correctly compute configuration

### LINKED ISSUE #3175: 🐛 `rome format` ignores command line options
### Environment information

```bash
Version: `main`
```

### What happened?

Run `rome` with `quote-style="single"` 

```bash
cargo run --bin rome --release format ~/Desktop/jquery.min.js --quote-style=single
```

Rome formats the string literals in `jquery.min.js` with double quotes.

### Expected result

The command line arguments take precedence over the configuration and Rome formats the string literals in `jquery.min.js` with single quotes 

### Code of Conduct

ematipico:
## Summary

Fixes #3175 

We were incorrectly computing a `WorkspaceSettings`, while now we should change/create a `Configuration` struct instead and then update the settings of the workspace.

This bug went undetected because we don't have tests that check that arguments passed via CLI are correctly applied.

## Test Plan

I created two new tests, one that applies the generic configuration of the formatter and one that applies the configuration for the javascript language.

### DIFF
--- a/crates/rome_cli/src/commands/check.rs
+++ b/crates/rome_cli/src/commands/check.rs
@@ -2,14 +2,11 @@ use crate::commands::format::apply_format_settings_from_cli;
 use crate::{execute_mode, CliSession, Execution, Termination, TraversalMode};
 use rome_diagnostics::MAXIMUM_DISPLAYABLE_DIAGNOSTICS;
 use rome_service::load_config;
-use rome_service::settings::WorkspaceSettings;
 use rome_service::workspace::{FixFileMode, UpdateSettingsParams};
 
 /// Handler for the "check" command of the Rome CLI
 pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
     let configuration = load_config(&session.app.fs, None)?;
-    let mut workspace_settings = WorkspaceSettings::default();
-
     let max_diagnostics: Option<u16> = session
         .args
         .opt_value_from_str("--max-diagnostics")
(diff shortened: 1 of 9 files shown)


crates/rome_cli/src/commands/check.rs
```diff
@@ -31,13 +28,11 @@ pub(crate) fn check(mut session: CliSession) -> Result<(), Termination> {
...
 
-    if let Some(configuration) = configuration {
-        session
-            .app
-            .workspace
-            .update_settings(UpdateSettingsParams { configuration })?;
-    }
-    apply_format_settings_from_cli(&mut session, &mut workspace_settings)?;
+    let configuration = apply_format_settings_from_cli(&mut session, configuration)?;
+    session
```
MichaReiser, line 32: What does this change mean for concurrent usages of the deamon. Do the CLI argument change for all requests to the workspace?

ematipico: Yes

leops (approved):
This looks good overall, can we just add an additional test that has both a configuration file and CLI arguments override to ensure the priority is handled correctly ?

crates/rome_cli/src/commands/check.rs
MichaReiser: Yes, but what if multiple clients are connected to the same daemon 

```bash
# client 1
rome format --indent-style="tab" --use-server

#client 2
rome format --indent-style="space" --use-server
```


And both commands run at the same time. Which configuration would be used?

ematipico: Both clients have two different instances of `Workspace`, so their configuration won't mix up.

PR #3186, opened 2022-09-08
test(rome_cli): snap termination error

ematipico:
## Summary

This PR increases the coverage of the snapshots of the CLI tests, by tracking the `Termination` message emitted in case of errors. 

All the snapshots that result in errors are updated. While doing so, I found a bug in how we construct the error message of the rules and fixed it. 

## Test Plan

Manually checked that the termination messages are printed in the snapshots

### DIFF
--- a/crates/rome_cli/tests/snap_test.rs
+++ b/crates/rome_cli/tests/snap_test.rs
@@ -26,7 +26,7 @@ struct CliSnapshot {
 }
 
 impl CliSnapshot {
-    pub fn form_result(result: Result<(), Termination>) -> Self {
+    pub fn from_result(result: Result<(), Termination>) -> Self {
         Self {
             in_messages: InMessages::default(),
             configuration: None,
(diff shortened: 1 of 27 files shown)


crates/rome_cli/tests/snapshots/main_configuration/incorrect_globals.snap
```diff
@@ -16,4 +15,8 @@ expression: content
 }
 ```
 
+# Termination Message
+
+Rome couldn't load the configuration file, here's why:
```
ematipico, line 20: This line was repeated twice. The introduction of `ConfigurationError:UnknownRule` fixed the issue.

crates/rome_cli/tests/snap_test.rs
```diff
@@ -21,6 +21,20 @@ struct CliSnapshot {
     pub files: HashMap<String, String>,
     /// messages written in console
     pub messages: Vec<String>,
+    /// possible termination error of the CLI
+    pub termination: Option<Termination>,
+}
+
+impl CliSnapshot {
+    pub fn form_result(result: Result<(), Termination>) -> Self {
```
leops, line 29: This looks like a typo:
```suggestion
    pub fn from_result(result: Result<(), Termination>) -> Self {
```

leops (approved):

PR #3190, opened 2022-09-09
fix(rome_js_analyze): accept const as non camel case

xunilrj:
## Summary

This PR fixes a false flag of all uppercase const as an invalid non-camel case.
Fixes a lot of issues here: https://github.com/rome/tools/runs/8188485278

## Test Plan

```
> cargo test -p rome_js_analyze -- camel_case
```

### DIFF
--- a/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
@@ -59,8 +59,8 @@ fn check_is_camel(name: &str) -> Option<State> {
     }
 }
 
-// It is ok to be non camel case when:
-// 1 - is a const variable (eg: const THIS_IS_OK)
+// It is OK to be non camel case when:
+// 1. it's a const variable (eg: const THIS_IS_OK)
 fn is_non_camel_ok(binding: &JsIdentifierBinding) -> Option<bool> {
     use JsSyntaxKind::*;
     let declarator = binding.parent::<JsVariableDeclarator>()?;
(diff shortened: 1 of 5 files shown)


ematipico (commented):
Could you enable the rule inside `rome.json` file? Doing so it will be a better proof that this PR fixes the false positives

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
```diff
@@ -59,6 +59,21 @@ fn check_is_camel(name: &str) -> Option<State> {
     }
 }
 
+// It is ok to be non camel case when:
+// 1 - is a const variable (eg: const THIS_IS_OK)
```
ematipico, line 63:
```suggestion
/// It is OK to be non camel case when:
/// 1 - is a `const` variable (eg: `const THIS_IS_OK`)
```

crates/rome_js_analyze/tests/specs/js/useCamelCase.js.snap
```diff
@@ -34,6 +34,8 @@ console.log({
 let camelCase;
 let longCamelCase;
 
+const THIS_IS_OK = 1;
```
ematipico, line 37: Does this fix cover cases like `const { FOO } = env`?

ematipico (approved):

crates/rome_js_analyze/tests/specs/js/useCamelCase.js.snap
xunilrj: Yes. I update the test with more cases.
There is one case we do not deal well.

```
warning[js/useCamelCase]: Prefer variables names in camel case.
   ┌─ useCamelCase.js:31:5
   │
31 │ let UPPER_CASE = 1;
   │     ----------

Safe fix: Rename this symbol to camel case
      | @@ -28,7 +28,7 @@
27 27 |   let camelCase;
28 28 |   let longCamelCase;
29 29 |   
30    | - let UPPER_CASE = 1;
   30 | + let uPPERCASE = 1;
31 31 |   let { UPPER_CASE } = 1;
32 32 |   let [ UPPER_CASE ] = 1;
33 33 |  
```
The problem is that it is easy to fix this, but break other acronyms like IS_HTML -> isHtml.
Not sure the best way here.

crates/rome_js_analyze/src/semantic_analyzers/js/use_camel_case.rs
xunilrj: Done.

PR #3207, opened 2022-09-13
feat(rome_js_analyze): `noDangerouslySetInnerHtml` rule

### LINKED ISSUE #3204: `noDangerouslySetInnerHtml`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noDanger.md

ematipico:
## Summary

Closes  #3204 

## Test Plan

I created new test cases


crates/rome_js_analyze/src/semantic_analyzers/react/no_danger.rs
```diff
@@ -0,0 +1,139 @@
...
+    ///
+    /// ```jsx,expect_diagnostic
+    /// React.createElement('div', {
+    //     dangerouslySetInnerHTML: { __html: 'child' }
+    // });
+    /// ```
+    pub(crate) NoDanger {
+        version: "0.10.0",
+        name: "noDanger",
+        recommended: true,
```
leops, line 31: I'm not sure if this rule should be part of the recommended set, in a way it's similar to the `unsafe_code` lint rule in Rust (that is also disabled by default): like `unsafe`, the `dangerouslySetInnerHTML` attribute is an "escape hatch" feature that exists to support a number of perfectly valid use case, and like `unsafe` the name of the attribute explicitly spells out that what the code is doing is dangerous and requires extra care.
While it's also perfectly valid to want to disallow any usage of this attribute in a codebase, I think it could be opt-in rather than opt-out, as requiring users to add a suppression comment on top of the already scary looking attribute name seems a bit too much as a default

ematipico: Let's try to see it from another perspective, recommending the rule forces the users to suppress the rule and add a reason for doing so. Which can be compared to the community pattern of having a `// SAFETY` comment. 

We have an example in our code base, where the rule was suppressed with a "safety" reason. 

Anyway, not a strong argument from my end, we can have it as opt-in rule.

MichaReiser:
> This PR also creates a new group called react, because that's where the rule should be placed.

This is an interesting one. The rule prevents security vulnerabilities, should it, thus, be placed under security instead?

ematipico:
@MichaReiser @leops the rule has been moved under the `nursery` group and it's not recommended, as per internal policy.

We can figure out their final group and recommendation once we decide to take it out from `nursery` group.

MichaReiser:
> @MichaReiser @leops the rule has been moved under the `nursery` group and it's not recommended, as per internal policy.
> 
> We can figure out their final group and recommendation once we decide to take it out from `nursery` group.

Can we then use a more specific name: `react-noDanger` or `react/noDanger` as it is otherwise unclear what this rule is about. I'm even wondering if the name should be `dangerouslySetInnerHTML` as I didn't understood what `noDanger` is about.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_danger.rs
```diff
@@ -0,0 +1,139 @@
...
+                            let property_member =
+                                member.as_js_property_object_member()?.name().ok()?;
+                            let name = property_member.as_js_literal_member_name()?;
+
+                            if name.syntax().text_trimmed() == "dangerouslySetInnerHTML" {
+                                return Some(NoDangerState::Property(name.clone()));
+                            }
+                        }
+                        return None;
+                    } else {
```
MichaReiser, line 106: Are these `else` branches necessary. You seem to be using explicit `return` calls in the paths where you are returning something.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_danger.rs
```diff
@@ -0,0 +1,139 @@
...
+                let name = jsx_attribute.name().ok()?;
+                name.syntax().text_trimmed_range()
+            }
+            NoDangerState::Property(property) => property.syntax().text_trimmed_range(),
+        };
+
+        let diagnostic = RuleDiagnostic::new(
+            text_range,
+            markup! {
+                "Avoid passing content using the "<Emphasis>"dangerouslySetInnerHTML"</Emphasis>" prop."
```
MichaReiser, line 130: Is the same message used for a `React.createElement` call?

crates/rome_js_analyze/src/utils.rs
```diff
@@ -237,3 +240,89 @@ fn ok_to_camel_case() {
...
+            let member = node.member().ok()?;
+            let member = member.as_js_name()?;
+            let identifier = object.as_js_identifier_expression()?.name().ok()?;
+            let identifier_binding = model.declaration(&identifier);
+
+            if let Some(binding_identifier) = identifier_binding {
+                let binding_identifier = binding_identifier
+                    .syntax()
+                    .clone()
+                    .cast::<JsIdentifierBinding>()?;
```
MichaReiser, line 284: Nit: `JsIdentifierBinding::cast_ref(binding_identifier.syntax())?`

crates/rome_js_analyze/src/utils.rs
```diff
@@ -237,3 +240,89 @@ fn ok_to_camel_case() {
...
+///
+/// There are two cases:
+///
+/// First case
+/// ```js
+/// React.createElement()
+/// ```
+/// We check if the node is a static member expression with the specific members. Also, if `React`
+/// has been imported in the current scope, we make sure that the binding `React` has been imported
+/// from the `"react"`.
```
MichaReiser, line 258:
```suggestion
/// from the `"react"` module.
```

crates/rome_js_analyze/src/utils.rs
```diff
@@ -237,3 +240,89 @@ fn ok_to_camel_case() {
...
+                let binding_identifier = identifier_binding
+                    .syntax()
+                    .clone()
+                    .cast::<JsIdentifierBinding>()?;
+                for ancestor in binding_identifier.syntax().ancestors() {
+                    if ancestor.kind() == JS_IMPORT {
+                        return Some(
+                            binding_identifier.syntax().text_trimmed()
+                                == identifier.syntax().text_trimmed()
+                                && identifier.syntax().text_trimmed() == "createElement",
```
MichaReiser, line 316: Same here, we can move this out

ematipico:
> > @MichaReiser @leops the rule has been moved under the `nursery` group and it's not recommended, as per internal policy.
> > We can figure out their final group and recommendation once we decide to take it out from `nursery` group.
> 
> Can we then use a more specific name: `react-noDanger` or `react/noDanger` as it is otherwise unclear what this rule is about. I'm even wondering if the name should be `dangerouslySetInnerHTML` as I didn't understood what `noDanger` is about.

What about `noDangerouslySetInnerHtml`? It's really explicit

crates/rome_js_analyze/src/utils.rs
```diff
@@ -237,3 +240,89 @@ fn ok_to_camel_case() {
...
+///
+/// Second case
+///
+/// ```js
+/// createElement()
+/// ```
+///
+/// The logic of this second case is very similar to the previous one, simply the node that we have
+/// to inspect is different.
+pub(crate) fn is_react_create_element(
```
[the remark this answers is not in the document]
ematipico: This function will be used for all the rules that orbit around React, that's why I put it straight inside the `utils`.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_danger.rs
ematipico: yes

crates/rome_js_analyze/src/utils.rs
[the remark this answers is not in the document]
MichaReiser: How about putting it inside of a react module instead? `utils` tend to become this wild buckets where you find everything and nothing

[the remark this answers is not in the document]
ematipico: I can do that in the next PR around React. Is that OK? About the number, I could find at least four usages in Rome classic

leops (approved):

PR #3192, opened 2022-09-09
refactor(rome_js_analyze): groups

ematipico:
## Summary

> **Note**: This PR is a highly disruptive. It also changes the documentation of the website. Maybe we should merge before the next release.

Relevant discussion https://github.com/rome/tools/discussions/3017#discussioncomment-3567594

This PR refactor the group names, please read the relevant discussion for more details.

## Test Plan

Updated documentation, CLI tests, analyzer tests. The CI should pass because no new functionality has been implemented.

PR #3215, opened 2022-09-13
docs: update metadata of rules

ematipico:
## Summary

This PR updates some parts of the website/repo, to update some information about the new groups

## Test Plan

Rendered website


xtask/lintdoc/src/main.rs
```diff
@@ -491,3 +516,12 @@ fn assert_lint(
 
     Ok(())
 }
+
+pub fn markup_to_string(markup: Markup) -> String {
+    let mut buffer = Vec::new();
+    let mut write = Termcolor(NoColor::new(&mut buffer));
```
leops, line 522: This should probably be using the HTML writer instead of the terminal writer. Additionally, the function could take a mutable reference to the `index` buffer and write directly to the output instead of having to allocate an intermediate string

MichaReiser:
What's the timing you plan to merge this PR. My understanding was that we only update the website with the release that ships the changes.

ematipico:
> What's the timing you plan to merge this PR. My understanding was that we only update the website with the release that ships the changes.

True, but unfortunately that ship has sailed when we merged https://github.com/rome/tools/pull/3192. Part of our website is tight to the source code of our rules. :(

MichaReiser (approved):

ematipico:
Going to merge. I downgraded `noNewSymbol` to **not recommended** because it starts from the `nursery` group

PR #3219, opened 2022-09-14
feat(rome_formatter): Source Map iteration over deleted ranges

MichaReiser:
This PR adds a `DeletedRanges` iterator to the `TransformedSourceMap` that allows iterating over all deleted ranges including the deleted text.

The comments refactoring needs a way to identify if a trivia piece is right before the closing `)` in the original tree because we want to keep trailing comments inside parentheses with the preceding nodes

```javascript
!(
  a
  // some comment
);
b
```

The comment should be a trailing comment of `a` and not become a leading comment of `b` because the transformed tree has the following shape

```javascript
!a
  // Some comment
b;
```

Such an iterator requires a fast way to get the text of a deleted range. This isn't possible with `SyntaxNodeText` because resolving the text requires iterating to the token with the given offset which is `O(n)`. That's why this PR changes the `TransfromedSourceMap` to hold a `String` with the original source text to get `O(1)` access to the original text.

The downside of doing so is that the source text must be built up insideof the `SyntaxRewriter` and is stored twice in memory.

The cost of traversal should offset the cost in the comments builder and the memory allocation is short-lived, not adding a lot of pressure if Rome's running as a server.

## Tests
I added a new unit test that shows how the new iterator can be used

### DIFF
--- a/crates/rome_formatter/src/source_map.rs
+++ b/crates/rome_formatter/src/source_map.rs
@@ -484,7 +484,7 @@ pub struct DeletedRangeEntry<'a> {
 ///
 /// Returns the ranges in increased order by their start position.
 pub struct DeletedRanges<'a> {
-    source_text: &'a String,
+    source_text: &'a str,
 
     /// The mappings stored in increasing order
     deleted_ranges: std::slice::Iter<'a, DeletedRange>,
--- a/crates/rome_formatter/src/source_map.rs
+++ b/crates/rome_formatter/src/source_map.rs
@@ -1,7 +1,9 @@
 use crate::{Printed, SourceMarker, TextRange};
-use rome_rowan::{Language, SyntaxNode, SyntaxNodeText, TextSize};
+use rome_rowan::TextLen;
+use rome_rowan::{Language, SyntaxNode, TextSize};
 use rustc_hash::FxHashMap;
 use std::cmp::Ordering;
+use std::iter::FusedIterator;
 
 /// A source map for mapping positions of a pre-processed tree back to the locations in the source tree.
 ///
@@ -50,7 +52,7 @@ use std::cmp::Ordering;
 /// This is why the source map also tracks the mapped trimmed ranges for every node.
 #[derive(Debug, Clone)]
 pub struct TransformSourceMap {
-    source_text: SyntaxNodeText,
+    source_text: String,
 
     /// The mappings stored in increasing order
     deleted_ranges: Vec<DeletedRange>,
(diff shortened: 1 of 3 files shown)


crates/rome_formatter/src/source_map.rs
```diff
@@ -264,12 +273,6 @@ impl TransformSourceMap {
             marker.source = source;
         }
     }
-
-    fn map_verbatim_ranges(&self, ranges: &mut [TextRange]) {
-        for range in ranges {
-            *range = self.source_range(*range)
-        }
-    }
```
MichaReiser, line 272: mapping the verbatim ranges isn't necessary because these ranges represent ranges in the printed document and not in the source tree. 

This has gone unnoticed thus far because no formatter used the verbatim formatting.

crates/rome_js_formatter/src/syntax_rewriter.rs
```diff
@@ -7,28 +7,23 @@ use rome_js_syntax::{
...
+    AstNode, SyntaxKind, SyntaxRewriter, SyntaxToken, SyntaxTriviaPiece, SyntaxTriviaPieceComments,
     VisitNodeSignal,
 };
 use std::iter::FusedIterator;
 
 pub(super) fn transform(root: JsSyntaxNode) -> (JsSyntaxNode, TransformSourceMap) {
-    let mut rewriter = JsFormatSyntaxRewriter::new(&root);
+    let mut rewriter = JsFormatSyntaxRewriter::default();
     let transformed = rewriter.transform(root);
     (transformed, rewriter.finish())
```
MichaReiser, line 18: @ematipico this will make you happy... It's no longer passing `root` twice ;)

MichaReiser:
!bench_formatter

crates/rome_formatter/src/source_map.rs
```diff
@@ -460,8 +468,77 @@ impl TransformSourceMapBuilder {
...
+
+    /// The text of the removed range
+    pub text: &'a str,
+}
+
+/// Iterator over all removed ranges in a document.
+///
+/// Returns the ranges in increased order by their start position.
+pub struct DeletedRanges<'a> {
+    source_text: &'a String,
```
ematipico, line 487: Why `&String` and not `&str`?

ematipico (approved):

crates/rome_formatter/src/source_map.rs
MichaReiser: No reason. I'm surprised that there's no clippy rule for this.

PR #3222, opened 2022-09-14
feat(rome_diagnostics): add a new version of diagnostics

leops:
## Summary

This PR adds a new, backwards compatible implementation of diagnostics under the `v2` module of `rome_diagnostics`. Unlike the previous implementation, this version of diagnostics is trait-based rather than object based, meaning `Diagnostic` is now a trait rather than a struct. The aim of this change is to allow different diagnostic types to be defined for specific needs at various points of the codebase, and the design of the trait tries to strike a balance between allowing diagnostics to deeply customize how they are present to the user while still exposing enough information to support use cases such as serialization, in-editor display, or generally allowing introspection by mid-level code.

## Diagnostic

Currently the `Diagnostic` trait has the following signature:

```rust
pub trait Diagnostic: Debug {
    /// The category of a diagnostic is string code uniquely identifying this
    /// diagnostic type, such as `lint/correctness/noArguments`, `args/invalid`
    /// or `format/disabled`
    fn category(&self) -> Option<&Category>;

    /// The severity defines whether this diagnostic reports an error, a
    /// warning, an information or a hint to the user
    fn severity(&self) -> Severity;

    /// The description is a text-only explanation of the issue this diagnostic
    /// is reporting, intended for display contexts that do not support rich
    /// markup such as in-editor popovers
    ///
    /// The description should generally be as exhaustive as possible, since
    /// the clients that do not support rendering markup will not render the
    /// advices for the diagnostic either
    fn description(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result;

    /// An explanation of the issue this diagnostic is reporting
    ///
    /// In general it's better to keep this message as short as possible, and
    /// instead rely on advices to better convey contextual explanations to the
    /// user
    fn message(&self, fmt: &mut rome_console::fmt::Formatter<'_>) -> io::Result<()>;

    /// Advices are the main building blocks used compose rich errors. They are
    /// implemented using a visitor pattern, where consumers of a diagnostic
    /// can visit the object and collect the advices that make it up for the
    /// purpose of display or introspection
    fn advices(&self, visitor: &mut dyn Visitor) -> io::Result<()>;

    /// Diagnostics can defines additional advices to be printed if the user
    /// requires more detail about the diagnostic
    fn verbose_advices(&self, visitor: &mut dyn Visitor) -> io::Result<()>;

    /// A diagnostic can be tied to a specific "location": this can be a file,
    /// memory buffer, command line argument ... It may also be tied to a
    /// specific text range within the content of that location. Finally, it
    /// may also provide the source string for that location (this is required
    /// in order to display a code frame advice for the diagnostic)
    fn location(&self) -> Option<Location<'_>>;

    /// Tags convey additional boolean metadata about the nature of a diagnostic:
    /// - If the diagnostic can be automatically fixed
    /// - If the diagnostic resulted from and internal error
    /// - If the diagnostic is being emitted as part of a crash / fatal error
    /// - If the diagnostic is warning about a piece of unused or unnecessary code
    /// - If the diagnostic is warning about a piece of deprecated or obsolete code
    fn tags(&self) -> DiagnosticTags;

    /// Similarly to the `source` method of the [std::error::Error] trait, this
    /// returns another diagnostic that's the logical "cause" for this issue.
    /// For instance, a "request failed" diagnostic may have been cause by a
    /// "deserialization error". This allows low-level error to be wrapped in
    /// higher level concepts, while retaining enough information to display
    /// and fix the underlying issue
    fn source(&self) -> Option<&dyn Diagnostic>;
}
```

## Advices

As outlined in the documentation above, diagnostics are primarily built out of "advices", that are exposed through the `Visitor` trait:

```rust
pub trait Visitor {
    /// Prints a single log entry with the provided category and markup
    fn visit_log(&mut self, category: LogCategory, text: &dyn fmt::Display) -> io::Result<()>;

    /// Prints an unordered list of items
    fn visit_list(&mut self, list: &[&dyn fmt::Display]) -> io::Result<()>;

    /// Prints a code frame outlining the provided source location
    fn visit_frame(&mut self, location: Location<'_>) -> io::Result<()>;

    /// Prints the diff between the `prev` and `next` strings
    fn visit_diff(&mut self, prev: &str, next: &str) -> io::Result<()>;

    /// Prints a Rust backtrace
    fn visit_backtrace(
        &mut self,
        title: &dyn fmt::Display,
        backtrace: &Backtrace,
    ) -> io::Result<()>;

    /// Prints a command to the user
    fn visit_command(&mut self, command: &str) -> io::Result<()>;

    /// Prints a group of advices under a common title
    fn visit_group(&mut self, title: &dyn fmt::Display, advice: &dyn IntoAdvices)
        -> io::Result<()>;
}
```

## Derive

In order to minimize the boilerplate code required to implement a new diagnostic, the `rome_diagnostics_macros` crate provides a procedural macro to automatically derive the `Diagnostic` trait on a struct:

```rust
#[derive(Debug, Diagnostic)]
#[diagnostic(
    severity = Error,
    category = "args/fileNotFound",
    message(
        description = "the file {path} could not be found on disk",
        message(
            "the file "<Emphasis>{self.path}</Emphasis>" could not be found on disk"
        )
    )
)]
struct FileNotFoundDiagnostic {
    #[location(path)]
    path: String,
    /// [IgnoredFiles] is a type implementing [IntoAdvices] that will append
    /// additional advices to the diagnostics if the file could not be found
    /// because it was ignored
    #[advice]
    ignored_files: IgnoredFiles,
}
```

## Error

Since working with plain diagnostic can be complicated since they cannot easily be mixed, any type implementing `Diagnostic` can be upcast (using the `From` trait or `?` operator) into an `Error` struct inspired by the type of the same name in the `anyhow` crate, that essentially behaves as a `Box<dyn Diagnostic>`. For convenience the `rome_diagnostic` also exposes an alias of `std::result::Result` whose error type defaults to `Error`, and may eventually expose additional utility macros inspired by `anyhow` such as `bail!` or `ensure!`

## Context

In order to facilitate injecting additional information into diagnostic in mid-level code, the crate exposes a `DiagnosticExt` trait (implemented on diagnostics) and a `Context` trait (implemented on `Result` like the trait of the same name in `anyhow`) that support adding additional information to an existing diagnostic (currently setting the category, path and source code are supported, as well as wrapping an existing diagnostic with a new "caused by" diagnostic similar to `anyhow`)

## Categories

The category of a diagnostic is defined with the `Category` type that contains the name of the category as well as an optional link to the documentation page, and maybe eventually an in-binary block of documentation text for offline access. This type cannot be instantiated outside of the `rome_diagnostics_categories` crate that acts as a static registry for categories: the set of known diagnostic categories is generated at compile time using a build script, and categories in the registry may either be accessed statically using the `category!` macro (eg. `category!("lint/correctness/noArguments")` or dynamically using the `FromStr` trait (eg. `let category: &Category = name.parse()?;`)

## Adapters

The `rome_diagnostics` crate exposes a number of "adapter diagnostics" under the `adapter` module that wraps types such as `std::error::Error` or `std::io::Error` in a newtype implementing `Diagnostic`

## Serde

When the `serde_diagnostics` feature of the crate is enabled, the `serde` module will provide a `Diagnostic` struct that can be instantiated from any diagnostic type, implements the `Diagnostic` trait but can also be serialized / deserialized using `serde` and has a corresponding `schemars` schema


crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+/// The [Backtrace] type can be used to capture a native Rust stack trace, to
+/// be displayed a diagnostic advice for native errors
+#[derive(Clone, Debug)]
+#[cfg_attr(test, derive(Eq, PartialEq))]
+pub struct Backtrace {
+    inner: BacktraceKind,
+}
+
+impl Default for Backtrace {
+    #[inline(never)]
```
ematipico, line 18: Could you explain why we don't want to inline, please?

crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+impl NativeBacktrace {
+    fn new(top_frame: usize) -> Self {
+        Self {
+            backtrace: ::backtrace::Backtrace::new_unresolved(),
+            top_frame,
+            bottom_frame: bottom_frame(),
+        }
+    }
+
+    fn resolve(&mut self) {
```
ematipico, line 140: Would it be possible to add some documentation? Mainly because I don't know what "resolving a backtrace" means

crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+            if let Some(top_frames) = frames.get(..bottom_frame + 1) {
+                frames = top_frames;
+            }
+        }
+
+        frames
+    }
+}
+
+thread_local! {
```
ematipico, line 181: I read the documentation of this macro but it didn't help. What do we achieve with this macro? Why `BOTTOM_FRAME` needs to be "thread local"? 

(If it's not understood, I don't know what "thread local" means here)

crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+enum BacktraceFrame<'a> {
+    Native(&'a ::backtrace::BacktraceFrame),
+    Serialized(&'a SerializedFrame),
+}
+
+impl BacktraceFrame<'_> {
+    fn ip(&self) -> *mut c_void {
+        match self {
+            Self::Native(inner) => inner.ip(),
+            Self::Serialized(inner) => inner.ip as *mut c_void,
```
ematipico, line 355: What does this cast do?

crates/rome_diagnostics/src/v2/serde.rs
```diff
@@ -0,0 +1,503 @@
...
+    de::{self, SeqAccess},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+
+use super::{
+    diagnostic::DiagnosticTag, error::internal::AsDiagnostic, Backtrace, Category, DiagnosticTags,
+    IntoAdvices, LogCategory, Path, Severity, SourceCode, Visitor,
+};
+
+/// Serializable representation for a [Diagnostic]
```
ematipico, line 15: I think you wanted to link `super::Diagnostic`?

crates/rome_diagnostics/src/v2/serde.rs
```diff
@@ -0,0 +1,503 @@
...
+        Self::new_impl(diag.as_diagnostic())
+    }
+
+    fn new_impl<D: super::Diagnostic + ?Sized>(diag: &D) -> Self {
+        let category = diag.category().map(|category| category.name().to_string());
+
+        let severity = diag.severity();
+
+        let mut description = String::new();
+        write!(description, "{}", PrintDescription(diag)).unwrap();
```
ematipico, line 42: `expect` instead of `unwrap`? At least we can leave some message, since it seems they need stay in production

crates/rome_diagnostics/src/v2/serde.rs
```diff
@@ -0,0 +1,503 @@
...
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[cfg_attr(test, derive(Eq, PartialEq))]
+enum Advice {
+    Log(LogCategory, MarkupBuf),
+    List(Vec<MarkupBuf>),
+    Frame(Location),
+    Diff(String, String),
+    Backtrace(MarkupBuf, Backtrace),
+    Command(String),
+    Group(MarkupBuf, Advices),
```
ematipico, line 255: Would it be possible to document the variants? Some of them are self explanatory, but for example `Group` might need some clarification.

Also `List` might need some clarification, because it seems that it only accepts other markups (and not other advises, for example)

crates/rome_diagnostics/src/v2/context.rs
```diff
@@ -0,0 +1,510 @@
...
+
+    /// Diagnostic type returned by [super::DiagnosticExt::context], uses
+    /// `message` as its message and description, and `source` as its source
+    /// diagnostic
+    pub struct ContextDiagnostic<M, E> {
+        pub(super) message: M,
+        pub(super) source: E,
+    }
+
+    impl<M: fmt::Display, E: Debug> Debug for ContextDiagnostic<M, E> {
```
MichaReiser, line 177: What's the reason that `M` must implement `Display` for the `Debug` implementation (that

crates/rome_diagnostics/src/v2/context.rs
```diff
@@ -0,0 +1,510 @@
...
+                    path: match loc.path {
+                        Path::Argv => Path::Argv,
+                        Path::Memory => Path::Memory,
+                        Path::File(file) => {
+                            if let Some(Path::File(path)) = &self.path {
+                                Path::File(file.or(path.as_deref()))
+                            } else {
+                                Path::File(file)
+                            }
+                        }
```
MichaReiser, line 369: Nit: Add a `as_file` helper on `Path` or even a `as_os_path`

crates/rome_diagnostics/src/v2/diagnostic.rs
```diff
@@ -0,0 +1,131 @@
+use std::{fmt::Debug, io};
+
+use bitflags::bitflags;
+use serde::{Deserialize, Serialize};
+
+use rome_console::fmt;
+
+use super::{Category, Location, Visitor};
+
+/// The `Diagnostic` trait
```
MichaReiser, line 10: This documentation isn't adding any more information than its signature. 

How about you point users to the procmacro and other types that are important when working with diagnostics. 

[See miette's Diagnostic documentation](https://docs.rs/miette/0.10.0/miette/trait.Diagnostic.html)

crates/rome_diagnostics/src/v2/diagnostic.rs
```diff
@@ -0,0 +1,131 @@
...
+pub trait Diagnostic: Debug {
+    /// The category of a diagnostic is string code uniquely identifying this
```
MichaReiser, line 12:
```suggestion
    /// The category of a diagnostic uniquely identifying this
```

crates/rome_diagnostics/src/v2/diagnostic.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    /// The severity defines whether this diagnostic reports an error, a
+    /// warning, an information or a hint to the user
+    fn severity(&self) -> Severity {
+        Severity::Error
+    }
+
+    /// The description is a text-only explanation of the issue this diagnostic
+    /// is reporting, intended for display contexts that do not support rich
+    /// markup such as in-editor popovers
```
MichaReiser, line 27: Nit: I noticed that you prefer to not end comments with a `.` This makes the sentences seem incomplete in my view, as if something is supposed to come after popovers but nothing is...

crates/rome_diagnostics/src/v2/diagnostic.rs
```diff
@@ -0,0 +1,131 @@
...
+        let _ = fmt;
+        Ok(())
+    }
+
+    /// An explanation of the issue this diagnostic is reporting
+    ///
+    /// In general it's better to keep this message as short as possible, and
+    /// instead rely on advices to better convey contextual explanations to the
+    /// user
+    fn message(&self, fmt: &mut fmt::Formatter<'_>) -> io::Result<()> {
```
MichaReiser, line 42: Nit: It may be helpful (not such this method but description as well) to have a few examples of good messages/descriptions. But maybe this is better suited inside of a guide on how to write good diagnostics

crates/rome_diagnostics/src/v2/diagnostic.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+bitflags! {
+    pub struct DiagnosticTags: u8 {
+        /// This diagnostic has a fix suggestion
+        const FIXABLE = 1 << DiagnosticTag::Fixable as u8;
+        /// This diagnostic results from an internal error
+        const INTERNAL = 1 << DiagnosticTag::Internal as u8;
+        /// This diagnostic results from a crash
+        const FATAL = 1 << DiagnosticTag::Fatal as u8;
```
MichaReiser, line 125: What's the reasoning behind making `FATAL` a tag and not a `Severity` level?

crates/rome_diagnostics/src/v2/display.rs
```diff
@@ -0,0 +1,924 @@
...
+/// Display struct implementing the formatting of a diagnostic header
+struct PrintHeader<'fmt, D: ?Sized>(&'fmt D);
+
+impl<'fmt, D: Diagnostic + ?Sized> fmt::Display for PrintHeader<'fmt, D> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> io::Result<()> {
+        let Self(diag) = *self;
+
+        // Wrap the formatter with a counter to measure the width of the printed text
+        let mut slot = None;
+        let mut fmt = CountWidth::wrap(f, &mut slot);
```
MichaReiser, line 64: Nit: What do you think of `CountWidth` to have a `finish` method instead that returns the `width` rather than passing in a mutable slot. That would also ensure that it is impossible to write to `CountWidth` after it has been "consumed".

crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+
+impl serde::Serialize for Backtrace {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        let frames = match &self.inner {
+            BacktraceKind::Native(backtrace) => {
+                let mut backtrace = backtrace.clone();
+                backtrace.resolve();
```
MichaReiser, line 57: Is the reason for cloning the backtrace here because `self` is passed by an immutable reference. What are your thoughts on using internal mutability to resolve the backtrace only once?

crates/rome_diagnostics/src/v2/display/backtrace.rs
MichaReiser: It creates a variable that stores a different value for every thread. 

Thread A: Stores A
Thread B: reads variable -> None 
Thread B: stores B
Thread A: reads variable -> A
Thread B: reads variable -> b


But I share the incentive that some documentation explaining the reason for a thread local would be helpful

crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+thread_local! {
+    static BOTTOM_FRAME: Cell<Option<usize>> = Cell::new(None);
+}
+
+/// Registers a function pointer as the "bottom frame" for this thread: all
+/// instances of [Backtrace] created on this thread will omit this function and
+/// all entries below it on the stack
+pub fn set_bottom_frame(ptr: usize) {
```
MichaReiser, line 188: What are the situation where some downstream module must call `set_bottom_frame` (I assume they have or the function wouldn't be `pub`) and in what order?

crates/rome_diagnostics/src/v2/display/backtrace.rs
```diff
@@ -0,0 +1,430 @@
...
+        }
+    }
+
+    fn colno(&self) -> Option<u32> {
+        match self {
+            Self::Native(inner) => inner.colno(),
+            Self::Serialized(inner) => inner.colno,
+        }
+    }
+}
```
MichaReiser, line 430: Is it possible to add one test showing how the backtrace functionality is used or isn't that possible because the stack will change with every code change and contain relative file paths?

crates/rome_diagnostics/src/v2/error.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// Create a new instance of [ErrorImpl] containing the provided diagnostic
+    /// and a reference to the corresponding vtable, allocate it on the heap
+    /// using [Box] and returns the corresponding raw pointer
+    fn new(diag: D) -> NonNull<ErrorImpl> {
+        let vtable = &ErrorVTable {
+            as_diagnostic: as_diagnostic::<D>,
+            drop: drop_error::<D>,
+        };
+
+        let inner = Box::new(Self { vtable, diag });
```
MichaReiser, line 140: How does it work that rust thinks that `vtable` has a `'static` lifetime?

crates/rome_diagnostics/src/v2/location.rs
```diff
@@ -0,0 +1,391 @@
+use std::ops::Deref;
+
+use rome_text_edit::{TextRange, TextSize};
+use serde::{Deserialize, Serialize};
+
+/// Represents the location of a diagnostic in a file
+#[derive(Debug, Clone, Copy)]
+pub struct Location<'a> {
+    /// The path of the file this diagnostic is associated with
+    pub path: Path<&'a str>,
```
MichaReiser, line 10: Nit: `Path` is, in my view, a concept coupled to file systems. `URIs` or `URLs` on the other hand represent resource identifiers/locators for different "backends". 

What's your opinion on renaming `Path` to `Url` or `resource_locator` (or schema to be more in line with URL). That would also make it slightly less awkward to do `location.path.path()` for files.

crates/rome_diagnostics/src/v2/location.rs
```diff
@@ -0,0 +1,391 @@
...
+                file_id,
+            },
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+/// An id that points into a file database.
+pub struct FileId(usize);
```
MichaReiser, line 170: Can we add a re-exports of `FileId` to `rome_js_parser` so that downstream crates don't have to depend on `rome_diagnostics` only to call `parse_module`?

crates/rome_diagnostics/src/v2/serde.rs
```diff
@@ -0,0 +1,503 @@
...
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+
+use super::{
+    diagnostic::DiagnosticTag, error::internal::AsDiagnostic, Backtrace, Category, DiagnosticTags,
+    IntoAdvices, LogCategory, Path, Severity, SourceCode, Visitor,
+};
+
+/// Serializable representation for a [Diagnostic]
+#[derive(Debug, Serialize, Deserialize)]
```
MichaReiser, line 16: Would it make sense to add `Clone` which is advised for most public  types. The same applies for other types in this module/crate)

crates/rome_diagnostics_categories/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "rome_diagnostics_categories"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
MichaReiser, line 6: Nit: Delete?

crates/rome_diagnostics_categories/src/categories.rs
```diff
@@ -0,0 +1,65 @@
...
+    "lint/correctness/useWhile": "https://rome.tools/docs/lint/rules/useWhile",
+    "lint/nursery/noUnreachable": "https://rome.tools/docs/lint/rules/noUnreachable",
+    "lint/nursery/noUnusedVariables": "https://rome.tools/docs/lint/rules/noUnusedVariables",
+    "lint/nursery/useCamelCase": "https://rome.tools/docs/lint/rules/useCamelCase",
+    "lint/nursery/noNewSymbol": "https://rome.tools/docs/lint/rules/noNewSymbol",
+    "lint/style/noNegationElse": "https://rome.tools/docs/lint/rules/noNegationElse",
+    "lint/style/noShoutyConstants": "https://rome.tools/docs/lint/rules/noShoutyConstants",
+    "lint/style/useSelfClosingElements": "https://rome.tools/docs/lint/rules/useSelfClosingElements",
+    "lint/style/useShorthandArrayType": "https://rome.tools/docs/lint/rules/useShorthandArrayType",
+    ;
```
MichaReiser, line 49: Is the `;` intentional? If it is to separate between the lint categories and general categories, then how about using `[]` to enclose each group?

crates/rome_diagnostics_categories/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+use std::str::FromStr;
+
+/// Metadata for a diagnostic category
+///
+/// This type cannot be instantiated outside of the `rome_diagnostics_categories`
+/// crate, which serves as a registry for all known diagnostic categories
+/// (currently this registry is fully static and generated at compile time)
+#[derive(Debug)]
```
MichaReiser, line 8: Nit: Implement `Copy` and `Clone` and `Eq`, `PartialEq`?

Would you mind going through the public structs that you introduce and check for each if it makes sense to implement `Clone` and `Copy`.

crates/rome_diagnostics_macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "rome_diagnostics_macros"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
MichaReiser, line 6:
```suggestion
```

MichaReiser (approved):
Excellent work @leops! This is looking very promising and I'm excited to see how this will improve our linter and parser diagnostics! 

This PR is overall looking good to me. My sole concern is around the `Visitor` naming which seems a bit odd to me as a consumer of the API because it's unclear where the "visitor" pattern is used (to my understanding, it isn't calling back into my visitor). The `visitor` idea seems to be mostly relevant for traversing the diagnostics/advices which is only done in a few places. 

That's why I would optimise the naming for the consumer and name the methods `add_log`, `add_command`. 

The other thing that would be worth doing is to review the newly introduced public types again and check for which it makes sense to implement `Clone`, `Copy`, `Eq`, and `PartialEq.

crates/rome_diagnostics/src/v2/adapter.rs
```diff
@@ -0,0 +1,61 @@
+//! This modules exposes a number of "adapter diagnostics" that wrap error types
```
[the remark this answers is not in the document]
leops: I initially envisioned the types in this module would be used as `adapter::StdError`, but that pattern still works with the name `adapters` and that's more intuitive anyway so I'll rename it

crates/rome_diagnostics/src/v2/advice.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// Print the advices with the warning style
+    Warn,
+    /// Print the advices with the error style
+    Error,
+}
+
+/// Utility type implementing [IntoAdvices] that emits a single log advice with
+/// the provided category and text
+#[derive(Debug)]
+pub struct LogAdvice<T> {
```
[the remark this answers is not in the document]
leops: I generally try to avoid adding trait boundaries on the type themselves, as those must currently be repeated on everything that uses this type (this constraint may be partially relaxed in future versions of Rust but it probably won't be lifted entirely), so instead the trait boundaries are only declared on `impl` blocks that use them

crates/rome_diagnostics/src/v2/diagnostic.rs
```diff
@@ -0,0 +1,131 @@
...
+    Internal,
+    Fatal,
+    Unnecessary,
+    Deprecated,
+}
+
+bitflags! {
+    pub struct DiagnosticTags: u8 {
+        /// This diagnostic has a fix suggestion
+        const FIXABLE = 1 << DiagnosticTag::Fixable as u8;
```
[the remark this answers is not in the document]
leops: No even with `#[repr(u8)]` Rust would still require an explicit cast from `DiagnosticTag` to `u8`

crates/rome_diagnostics/src/v2/location.rs
```diff
@@ -0,0 +1,391 @@
...
+            Path::Argv => Path::Argv,
+            Path::Memory => Path::Memory,
+            Path::File(file) => Path::File(file.as_deref()),
+        }
+    }
+}
+
+impl Path<&'_ str> {
+    /// Converts a `Path<&str>` to `Path<String>`
+    pub fn to_owned(self) -> Path<String> {
```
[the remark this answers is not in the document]
leops: The `ToOwned` trait requires that the `Owned` type implements `Borrow`, whose `borrow` method must return a reference with the type `&Borrowed`. For the `Path` type this doesn't work since its "borrowed form" is `Path<&str>` and not `&Path<str>`

crates/rome_diagnostics/src/v2/location.rs
```diff
@@ -0,0 +1,391 @@
...
+            Path::Memory => Path::Memory,
+            Path::File(file) => Path::File(file.to_owned()),
+        }
+    }
+}
+
+/// Represents the path of a file on the filesystem
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub enum FilePath<P> {
```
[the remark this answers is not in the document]
leops: The explanation here is the same as with `LogAdvice`, the trait bounds are only specified on the `impl` blocks for the type to avoid "contaminating" every place the type is used with redundant trait bounds

crates/rome_diagnostics/src/v2/display/backtrace.rs
leops: That's because `Backtrace::default` is used as the "top frame" for the backtrace, we ask the compiler to never inline it to try to ensure the function will always create a stack frame so that it (and all the other internal functions above it) can be filtered out of the resulting backtrace

crates/rome_diagnostics/src/v2/display/backtrace.rs
leops: The `ip` of a backtrace symbol if the value of the Instruction Pointer for this frame on the stack. Raw pointer types do not implement `Serialize` and `Deserialize`, since whatever memory the pointer "points to" will almost certainly not exist in the remote location where the pointer is deserialized. But for backtraces this doesn't really matter since that pointer is only printed for debugging purpose but is never dereferenced, so we explicitly serialize it to a `u64` but that means we have to cast it back to a `*mut c_void` to print it to the console with the same formatting as any other pointer

crates/rome_diagnostics_categories/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct Category {
+    name: &'static str,
+    link: Option<&'static str>,
+}
+
+impl Category {
+    /// Return the name of this diagnostic
```
ematipico, line 15:
```suggestion
    /// Return the name of this category
```

crates/rome_diagnostics_categories/src/categories.rs
```diff
@@ -0,0 +1,65 @@
...
+//
+// The `define_dategories` macro is preprocessed in the build script for the
+// crate in order to generate the static registry. The body of the macro
+// consists of a list of key-value pairs defining the categories that have an
+// associated hyperlink, then a list of string literals defining the remaining
+// categories without a link.
+
+define_dategories! {
+    // Lint categories
+    "lint/correctness/noArguments": "https://rome.tools/docs/lint/rules/noArguments",
```
ematipico, line 12: Are these categories autogenerated? If so, we should have a comment line saying so. I not, we really should. If that's not possible, we have to update the `CONTRIBUTING.md` file and add instructions about it.

crates/rome_diagnostics_macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
ematipico, line 1: Could you please update the file with the following information? 

```toml
authors = ["Rome Tools Developers and Contributors"]
repository = "https://github.com/rome/tools"
license = "MIT"
```

`version` should `0.0.0` because it's not a crate we intend to publish for the time being.

crates/rome_diagnostics_categories/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "rome_diagnostics_categories"
```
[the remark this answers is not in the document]
ematipico: Would it be possible to add a description for the crate? Just to explain how the create works. The presence of `build.rs` hints that there's something that it's done at build time, so maybe we should have this written somewhere in the `build.rs`? Just to give some context to the developers

crates/rome_fs/Cargo.toml
```diff
@@ -21,4 +21,4 @@ schemars = { version = "0.8.10", optional = true }
 [dev-dependencies]
 
 [features]
-serde = ["dep:serde", "schemars"]
+serde = ["dep:serde", "schemars", "rome_diagnostics/schema"]
```
ematipico, line 24: What was the reason for creating a new feature just for the diagnostic schema?

crates/rome_diagnostics/examples/cli.rs
```diff
@@ -0,0 +1,63 @@
...
+            &markup! {
+                "Did you mean "<Emphasis>{self.suggested_name}</Emphasis>" instead?"
+            },
+        )?;
+
+        visitor.visit_command(&self.suggested_command)?;
+
+        visitor.visit_log(LogCategory::Info, &"To see all available commands run")?;
+        visitor.visit_command("rome --help")
+    }
```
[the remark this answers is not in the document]
leops: This visitor pattern was originally inspired from the way the `tracing` library exposes the fields on spans and events to subscribers: https://docs.rs/tracing/latest/tracing/field/trait.Visit.html
I could bring the visitor implementation closer to this by renaming `Visitor` to `Visit`, the `visit_*` methods to `record_*`, the `visit` method on `IntoAdvices` to `record`, and maybe the `IntoAdvices` trait to just `Advices` ?

crates/rome_diagnostics/examples/cli.rs
```diff
@@ -0,0 +1,63 @@
...
+        path: Path::Argv,
+        span: TextRange::new(TextSize::from(5), TextSize::from(11)),
+        source_code: String::from("rome formqt file.js"),
+        advices: CliAdvices {
+            suggested_name: String::from("format"),
+            suggested_command: String::from("rome format file.js"),
+        },
+    };
+
+    EnvConsole::default().error(markup!({ PrintDiagnostic(&diag) }));
```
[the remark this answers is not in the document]
leops: There are explicit entries in the `Cargo.toml` file of `rome_diagnostics` to run the examples as part of the tests: https://github.com/rome/tools/blob/77ad64b285a33e27508c9f60e0c87e48260f5878/crates/rome_diagnostics/Cargo.toml#L9-L23
However those "example tests" do not include any assertion, they simply check that the code compiles and runs correctly

crates/rome_diagnostics/src/v2/adapter.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Diagnostic for StdError {
+    fn description(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(fmt, "{}", self.error)
+    }
+
+    fn message(&self, fmt: &mut fmt::Formatter<'_>) -> io::Result<()> {
+        let error = self.error.to_string();
+        fmt.write_str(&error)
+    }
```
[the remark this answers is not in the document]
leops: Yes the documentation for the `ToString` trait discourages from implementing the trait manually, instead recommending that users implement `Display`, so it can generally be assumed that calling `.to_string()` on a type will print the same value as `format!` (in fact this is guaranteed for types outside of the standard library until specialization is stabilized since manually implementing `ToString` on a type that already implements `Display` would conflict for the implementation of `ToString for T where T: Display` in the standard library)

As for the general question, I think it could be even better to use an `AsConsoleDisplay` adapter struct that implements `rome_console::fmt::Display` over any type implementing `std::fmt::Display`, and simply call `fmt.write_markup(markup!( {AsConsoleDisplay(&self.error)} ))` in `message`

crates/rome_diagnostics/src/v2/context.rs
```diff
@@ -0,0 +1,510 @@
...
+    where
+        Self: 'static,
+        M: fmt::Display + 'static,
+        Error: From<internal::ContextDiagnostic<M, Self>>;
+
+    /// Returns a new diagnostic using the provided `category` if `self`
+    /// doesn't already have one
+    fn with_category(self, category: &'static Category) -> Error
+    where
+        Error: From<internal::CategoryDiagnostic<Self>>;
```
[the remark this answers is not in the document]
leops: All the methods in `DiagnosticExt` and `Context` return the public `Error` dynamic type instead of the internal diagnostic wrapper struct

crates/rome_diagnostics/examples/cli.rs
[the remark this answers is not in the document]
ematipico: I personally don't mind the "visitor" pattern. While at the beginning I had the same doubts expressed by Micha (and confusion too), I understood that this is pattern used by the community, as far as I understood.

crates/rome_diagnostics/src/v2/context.rs
leops: Requiring that the context message implements both `Debug` and `Display` seemed needlessly restrictive, so instead the `context` method only requires `Display` and the implementation of `Debug` for the `ContextDiagnostic` wrapper prints the debug representation for the markup generated by displaying the message instead

crates/rome_diagnostics/src/v2/context.rs
leops: While I can see why that helper method makes sense, I'm not sure writing `if let Some(other) = self.path.as_ref().and_then(Path::as_file)` here would make things simpler

crates/rome_diagnostics/src/v2/diagnostic.rs
leops: The `Severity` maps more or less directly to the styling of how the error is presented to the user in the console or the editor, but we do not have any particular style to apply to fatal errors so those will use the `Error` severity, and the tags serves to add the additional information that this error has cause Rome to crash

MichaReiser: I think it isn't a requirement that the styling for all `Severity` are different. This may not even be possible for all outputs. For example, the styling for all severities may be the same for terminals that don't support colors. 

That's why I think it would be fine to use a `Fatal` severity but style it the same as `Error` (for now, may change in the future).

crates/rome_diagnostics/src/v2/display.rs
```diff
@@ -0,0 +1,924 @@
...
+
+        // Print the category of the diagnostic, with a hyperlink if
+        // the category has an associated link
+        if let Some(category) = diag.category() {
+            if let Some(link) = category.link() {
+                fmt.write_markup(markup! {
+                    <Hyperlink href={link}>{category.name()}</Hyperlink>" "
+                })?;
+            } else {
+                fmt.write_markup(markup! {
```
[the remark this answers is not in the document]
leops: Yes the buffering happens at the `Console` level, the `EnvConsole` implementation will lock the `stdout` or `stderr` stream for the duration of a call to `console.log` or `console.error` to ensure the entire markup is printed at once

crates/rome_diagnostics/src/v2/display/backtrace.rs
leops: Cloning and resolving the backtrace seemed simpler as that code would generally only get called once anyway, but I could wrap the content of `NativeBacktrace` in a `Mutex` too (a `RefCell` wouldn't work as backtraces have to be `Send + Sync` to be stored in diagnostics)

crates/rome_diagnostics/src/v2/display/backtrace.rs
leops: I'm adding a few examples on how and where to use this function, but there's essentially no hard requirement to ever call `set_bottom_frame`, it's only used to make backtraces generated by the current thread nicer by hiding irrelevant stack frames at the bottom of the trace (for instance the functions from the OS and Rust runtime that call into the `main` function)

crates/rome_diagnostics/src/v2/display/backtrace.rs
leops: Yes there are a few tests that check how backtraces are printed but that relies on serialized backtraces, since the native backtraces display pointers into the generated code as well as platform specific file paths and function implementations

crates/rome_diagnostics/src/v2/error.rs
leops: The Rust compiler will assign the `'static` lifetime to any reference expression that capture anything from the local scope, in this case the `vtable` initializer expression refers only to constant function pointers (that in turn refer to the `D` generic type but that's also constant at compile time) so the entire expression can be "upgraded" to a static

crates/rome_diagnostics/examples/cli.rs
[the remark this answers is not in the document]
MichaReiser: I like that proposoal

crates/rome_diagnostics/src/v2/display/backtrace.rs
MichaReiser: I guess it's fine then.

crates/rome_diagnostics/src/v2/location.rs
leops: `Url` / `Uri` generally refers to strings that conforms to the various IETF RFCs for URIs, whereas the `path` field is intended to adhere to a specific format and is primarily aimed at being displayed to the user to make it easy to understand where an error is happening. Still I agree that the `Path` name is too tied to the concept of a filesystem path, so I'm considering renaming the enum (and field) to `Resource` instead

crates/rome_diagnostics/src/v2/serde.rs
leops: I didn't derive `Clone` for this type since diagnostics generally cannot be cloned (the `Diagnostic` trait cannot require `Clone` as it needs to be "object safe" for dynamic dispatch), but I guess I could add it

crates/rome_diagnostics_categories/src/categories.rs
leops: The `;` token is used as a separator between the categories that have an associated link and the ones that don't, I reused this syntax from the equivalent "error code registry" in rustc since it doesn't create an intrusive separation between the two halves of the list: https://github.com/rust-lang/rust/blob/cba4a389b3961a2fd72e01bd6cb0b0e065edaf3d/compiler/rustc_error_codes/src/error_codes.rs#L496

crates/rome_diagnostics_categories/src/lib.rs
leops: I have implementations of `Eq`, `PartialEq` and `Hash` for this type in my followup integration PR but I could backport that here. I purposefully didn't implement `Clone` and `Copy` though since throughout most of the code diagnostic categories are represented as `&'static Category` instead, so cloning or copying a value of this kind would simply copy the pointer and retain the static lifetime

crates/rome_diagnostics/src/v2/display.rs
```diff
@@ -0,0 +1,924 @@
...
+        self.0.write_str(" ")?;
+
+        let mut slot = None;
+        let mut fmt = IndentWriter::wrap(self.0, &mut slot, false, "  ");
+        fmt.write_markup(Markup(&[MarkupNode {
+            elements: &[kind],
+            content: text,
+        }]))?;
+
+        self.0.write_str("\n\n")
```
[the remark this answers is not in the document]
leops: The reason why we cannot have self-closing elements is that markup elements only act as "style tags" on their underlying content, so without some special handling in the printer the `<Br></Br>` markup wouldn't print anything (it applies the `Br` style to zero child content, so nothing gets printed)

crates/rome_diagnostics_categories/src/categories.rs
leops: This file isn't automatically generated as it needs to list out all the existing error categories for the whole codebase, which we don't really have a way automatically extract from the code. It may also eventually include additional documentation for the error additionally to the link for the error page, and this documentation would have to be manually written by contributors as new error categories get introduced, this is something I plan on documenting as part of a future PR that will add an "How to write a diagnostic" section to the contributing documentation

crates/rome_fs/Cargo.toml
leops: The `rome_diagnostics` crate doesn't have a `serde` feature since the serialization code is used in tests for the crate, and Cargo doesn't allow enabling a set of features for tests. Instead I created a `schema` feature to disable the `schemars` code, and I think eventually this model will be extended to other crates in the workspace since the `serde` feature ends up being enabled in all builds by `rome_service` while the `schemars` code is only needed by `xtask`

crates/rome_diagnostics_categories/src/lib.rs
MichaReiser: Doing it as part of the follow up PR is fine.

PR #3223, opened 2022-09-14
refactor(website): add script to update the code contributors

ematipico:
## Summary

Rome classic had a [script](https://github.com/rome/tools/blob/archived-js/scripts/update-contributor-credits.ts) to update the credits section of the website.

This PR makes that script in Rust and implements the same logic.

## Test Plan

The rendered website should show the new code contributors


website/src/_includes/contributors.md
```diff
@@ -0,0 +1,666 @@
+<!-- Generated file, do not edit by hand, see `xtask/codegen` -->
```
MichaReiser, line 1: It may be helpful to mention the comment that allows regenerating the file because `xtask codegen` now contains many different commands and it's not always evident which one needs to be used.

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
+use pico_args::Arguments;
+use serde::{Deserialize, Serialize};
+use std::io::Write;
+use xtask::glue::fs2;
+use xtask::*;
+
+/// A token is needed to run this script. To create a token, go to https://github.com/settings/tokens
+/// and give it read access to the repository.
```
MichaReiser, line 8: Nit: May be worth mentioning how to run this command and that the token must be provided as an argument. 

What would even be better is if this text would be part of the help text of the command so that I don't need to read the source code to understand how to use the script

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+fn main() -> Result<()> {
+    let root = project_root().join("website/src/_includes");
+    let mut args = Arguments::from_env();
+    let token: String = args.value_from_str("--token").unwrap();
+    let mut contributors = Vec::new();
+    get_contributors(
+        "https://api.github.com/repos/rome/tools/contributors",
+        &token,
+        &mut contributors,
+    );
```
MichaReiser, line 20: Could `get_constributors` return a `Vec` instead of taking an empty vec as argument?

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+/// and give it read access to the repository.
+///
+/// Only users that have read rights can run this script
+fn main() -> Result<()> {
+    let root = project_root().join("website/src/_includes");
+    let mut args = Arguments::from_env();
+    let token: String = args.value_from_str("--token").unwrap();
+    let mut contributors = Vec::new();
+    get_contributors(
+        "https://api.github.com/repos/rome/tools/contributors",
```
MichaReiser, line 17: Nit: It doesn't seem to be supported to use different URLs. How about hard coding the URL inside of `get_constributors`

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+        &token,
+        &mut contributors,
+    );
+
+    let mut content = Vec::new();
```
MichaReiser, line 22: What's the reason for using a `Vec` instead of a `String`?

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+    writeln!(content, "### Code contributors")?;
+    writeln!(content)?;
+    writeln!(content, "<ul class=\"team-list credits\">")?;
+    for contributor in contributors {
+        let escaped_login = html_escape::encode_text(&contributor.login);
+        let escaped_avatar = html_escape::encode_text(&contributor.avatar_url);
+        writeln!(
+            content,
+            "<li><a href=\"https://github.com/rome/tools/commits?author={}\">",
+            contributor.login
```
MichaReiser, line 35: Should we escape this too?

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+        )?;
+        writeln!(
+            content,
+            "<img src=\"{}\" alt=\"{}\" />",
+            escaped_avatar, contributor.login
```
MichaReiser, line 40: Should the login be escaped too?

MichaReiser: does `encode_text` take care of escaping `"` too? Should we use `encode_double_quoted_attribute_to_string` instead?

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,95 @@
...
+                None
+            };
+            let result: Result<Vec<Contributor>, std::io::Error> = response.into_json();
+            if let Ok(new_contributors) = result {
+                contributors.extend(new_contributors);
+            }
+
+            if let Some(next_url) = next_url {
+                get_contributors(&next_url, token, contributors);
+            }
```
MichaReiser, line 89: It may be simpler to recurse directly
```suggestion
                    return get_contributors(&next_url, token, contributors);
                } else {
                    None
                }
            } else {
                None
            };
            let result: Result<Vec<Contributor>, std::io::Error> = response.into_json();
            if let Ok(new_contributors) = result {
                contributors.extend(new_contributors);
            }
```

xtask/contributors/src/main.rs
ematipico: `get_contributors` is a recursive function and the URL changes while we process the requests

xtask/contributors/src/main.rs
ematipico: No particular reason, I copied what we do inside the `lintdoc` command

xtask/contributors/src/main.rs
MichaReiser: Or rewrite this to a loop. May it make it clearer that we're following an URL here (would allow inline the URL)

xtask/contributors/src/main.rs
ematipico: > Should the login be escaped too?

The `alt` text, it should be escaped

xtask/contributors/src/main.rs
ematipico: I would to have refactor the function and make two functions only to allow that

xtask/contributors/src/main.rs
MichaReiser: I think the reason why `lintdoc` uses a `Vec` is because `write_html` requires a `std::io::Write` 

https://github.com/rome/tools/blob/16c98c6612494582064929f26e5ada58c7ffaf28/xtask/lintdoc/src/main.rs#L130

xtask/contributors/src/main.rs
```diff
@@ -0,0 +1,99 @@
...
+            &mut contributor_html,
+        );
+        contributor_html.push_str("\">");
+        contributor_html.push_str("<img src=\"");
+        html_escape::encode_double_quoted_attribute_to_string(
+            format!("{}", escaped_avatar),
+            &mut contributor_html,
+        );
+        writeln!(content, "{}", &contributor_html)?;
+        writeln!(content, "\" alt=\"{}\" />", contributor.login)?;
```
MichaReiser, line 42: Use `encode_double_quoted_attribute_to_string` to escape the login

MichaReiser (approved):
I think it would be helpful to change `content` to a `String` as we know that it is always valid UTF8 and there's no requirement from any used API to use a `Vec<u8>`.

xtask/contributors/src/main.rs
ematipico: The login in this part of the HTML doesn't have to be escaped, otherwise the result text would be wrong

PR #3224, opened 2022-09-14
feature(vscode): resolve `lspBin` using a relative path

ematipico:
## Summary

Relative discussion https://github.com/rome/tools/discussions/3218

This PR adds a feature to resolve the `lpsBin` path from a relative path. This is done by joining the `lpsBin` path to the workspace folders. If none exists after the resolution, Rome will throw an error.

## Test Plan

I manually tested this and it works. I opened a project and set `"rome.lspBin": "../../tools/target/debug/rome"` to point it to my local debug version of the binary.

### DIFF
--- a/editors/vscode/src/main.ts
+++ b/editors/vscode/src/main.ts
@@ -22,7 +22,6 @@ import { Session } from "./session";
 import { syntaxTree } from "./commands/syntaxTree";
 import { Commands } from "./commands";
 import { StatusBar } from "./statusBar";
-import exp = require("constants");
 
 let client: LanguageClient;
 


editors/vscode/src/main.ts
```diff
@@ -15,11 +15,14 @@ import {
...
 	StreamInfo,
 } from "vscode-languageclient/node";
+import { join, isAbsolute } from "path";
+import { existsSync } from "fs";
 import { setContextValue } from "./utils";
 import { Session } from "./session";
 import { syntaxTree } from "./commands/syntaxTree";
 import { Commands } from "./commands";
 import { StatusBar } from "./statusBar";
+import exp = require("constants");
```
leops, line 25: This looks like an autocomplete mistake

leops (approved):

PR #3225, opened 2022-09-14
doc(editors): how to use rome as a language server in helix

IWANABETHATGUY:

### DIFF
--- a/editors/helix/manual.md
+++ b/editors/helix/manual.md
@@ -1,0 +1,45 @@
+# Configuration
+Currently, rome support `js`, `jsx`, `ts`, `tsx` four languages. Rome use a file socket to connect the editor client, 
+it should be a little different from other language server using a binary like `rust-analyzer`,
+you could use `nc -U ${LANGUAGE_SERVER_SOCKET_PATH}` to connect your rome language server, `LANGUAGE_SERVER_SOCKET_PATH` could be got by
+`rome __print_socket`, by default it is `/tmp/rome-socket`. More details why we need nc you could reference https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers
+
+**languages.toml**
+```toml
+[[language]]
+name = "javascript"
+scope = "source.js"
+file-types = ["js"]
+language-server = { command = "nc", args = ["-U", "/tmp/rome-socket"] }
+formatter = { command = "rome", args = ["format", "--stdin-file-path", "test.js"]}
+auto-format = true
+
+[[language]]
+name = "jsx"
+scope = "source.jsx"
+file-types = ["jsx"]
+language-server = { command = "nc", args = ["-U", "/tmp/rome-socket"] }
+formatter = { command = "rome", args = ["format", "--stdin-file-path", "test.jsx"]}
+auto-format = true
+[[language]]
+name = "typescript"
+scope = "source.ts"
+file-types = ["ts"]
+language-server = { command = "nc", args = ["-U", "/tmp/rome-socket"] }
+formatter = { command = "rome", args = ["format", "--stdin-file-path", "test.ts"]}
+auto-format = true
+
+[[language]]
+name = "tsx"
+scope = "source.tsx"
+file-types = ["tsx"]
+language-server = { command = "nc", args = ["-U", "/tmp/rome-socket"] }
+formatter = { command = "rome", args = ["format", "--stdin-file-path", "test.tsx"]}
+auto-format = true
+
+```
+
+# Limitation
+1. The `rome-socket` will not automatically create when you reboot until you using command `rome __print_socket`. As a workaround,
+you could write a init startup bash script.
+


IWANABETHATGUY:
https://user-images.githubusercontent.com/17974631/190205045-aeb86f87-1915-4d8b-8aad-2c046443ba83.mp4


https://user-images.githubusercontent.com/17974631/190205065-ddfde866-5f7c-4f53-8a62-b6cbb577982f.mp4

MichaReiser:
Woahh this is cool.

IWANABETHATGUY:
Because I am not a native speaker, so maybe some expressions are not appropriate, feel free to change them.

editors/helix/manual.md
```diff
@@ -0,0 +1,51 @@
+# Configuration
+Currently, rome support `js`, `jsx`, `ts`, `tsx` four languages. Rome uses a file socket to connect the editor client, 
+which may be different from other language servers using a binary e.g. `rust-analyzer`.
+you could use `nc -U ${LANGUAGE_SERVER_SOCKET_PATH}` to connect your rome language server, `LANGUAGE_SERVER_SOCKET_PATH` could be got by
+`rome __print_socket`, by default it is `/tmp/rome-socket`. More details why we need nc you could reference https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers
```
ematipico, line 5:
```markdown
Currently, rome supports the following file extensions: `js`, `jsx`, `ts`, `tsx` and `d.ts`. 

Rome uses a file socket to connect the editor client, which may be different from other language servers using a binary e.g. `rust-analyzer`.

You can use `nc -U ${LANGUAGE_SERVER_SOCKET_PATH}` to connect to the Rome language server. `LANGUAGE_SERVER_SOCKET_PATH` is path to where the Rome's socket is created. Rome creates that socket inside the temporary folder of the operative system, inside a folder called `rome-socket`. 

To know the temporary folder of your OS, in unix, use:

	  ```shell
	  echo $TMPDIR
	  ```

In Windows, use:

	  ```shell
	  $env:TEMP
	  ```

More details why we need `nc`, please read the [wiki page of helix](https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers)
```

With github it's really difficult to apply suggestions in markdown. So I just reworded the paragraph in markdown so you can copy-path. You would need to remove the indentation from the shell commands (it was a trick to not break the markdown rendering)

editors/helix/manual.md
```diff
@@ -0,0 +1,51 @@
...
+
+**languages.toml**
+```toml
+[[language]]
+name = "javascript"
+scope = "source.js"
```
ematipico, line 11: I couldn't find this `scope` in their wiki... do you have a link that shows how it's used?

editors/helix/manual.md
```diff
@@ -0,0 +1,51 @@
...
+file-types = ["tsx"]
+language-server = { command = "nc", args = ["-U", "/tmp/rome-socket"] }
+formatter = { command = "rome", args = ["format", "--stdin-file-path", "test.tsx"]}
+auto-format = true
+
+```
+
+# Limitation
+1. The `rome-socket` will not automatically create when you reboot until you use the command `rome __print_socket`. As a workaround,
+you could write a init startup bash script.
```
ematipico, line 44: Rome now has [`start`](https://rome.tools/#rome-start) and [`stop`](https://rome.tools/#rome-stop) commands. Does that mean we don't have this limitation anymore?

ematipico: Yeah makes sense, then the description should be reworded to reflect the new commands

leops: Besides ensuring that the server is active and the socket is created, it's important to call `rome __print_socket` rather than assume the socket is in `/tmp/rome-server` because it may change depending on the OS (the path is completely different Windows, but it may also differ on Unix systems where the temporary directory is not `/tmp`), the version of Rome or the user configuration

IWANABETHATGUY: > Besides ensuring that the server is active and the socket is created, it's important to call `rome __print_socket` rather than assume the socket is in `/tmp/rome-server` because it may change depending on the OS (the path is completely different Windows, but it may also differ on Unix systems where the temporary directory is not `/tmp`), the version of Rome or the user configuration

Thanks, but helix only supports static configuration now which means we don't have the ability to get the path of
rome socket, so I put a default path in there,  I will add some extra explanation about that.

editors/helix/manual.md
IWANABETHATGUY: 1. For the language server part you could refer to https://docs.helix-editor.com/languages.html
2. For formatting, you could refer to https://github.com/helix-editor/helix/wiki/External-binary-formatter-configuration

editors/helix/manual.md
IWANABETHATGUY: Done

editors/helix/manual.md
ematipico: Thank you!

editors/helix/manual.md
```diff
@@ -0,0 +1,60 @@
...
+name = "tsx"
+scope = "source.tsx"
+file-types = ["tsx"]
+language-server = { command = "nc", args = ["-U", "/tmp/rome-socket"] }
+formatter = { command = "rome", args = ["format", "--stdin-file-path", "test.tsx"]}
+auto-format = true
+
+```
+`/tmp/rome-socket` is the default socket file path in Linux, I tested on Linux. If anything goes wrong, please double check if the 
+path equals to `rome __print_socket`
```
ematipico, line 49:
```suggestion
`/tmp/rome-socket` is the default socket file path in Linux. Use the command `rome __print_socket` and use the correct value.
```

editors/helix/manual.md
```diff
@@ -0,0 +1,60 @@
...
+
+# Limitation
+1. The `rome-socket` will not automatically create when you reboot until you use the command `rome __print_socket`. As a workaround,
+you could write a init startup bash script.
```
ematipico, line 53:
```suggestion
1. The socket is not automatically created, and you need to call the command [`rome start`](https://rome.tools/#rome-start) in order to create one. Use the command [`rome stop`](https://rome.tools/#rome-stop) to free the socket.
```

ematipico (approved):

PR #3228, opened 2022-09-15
chore(rome_js_syntax): Delete the unused `JsConstructorParameter` kind

MichaReiser:
## Summary

This PR deletes the unused `JsSyntaxKind::JS_CONSTRUCTOR_PARAMETER`. 

## Test Plan

`cargo build`


ematipico (approved):
I wonder if there's a way to generate these kinds from the grammar.. or make sure there aren't dangling keywords like in this example.

MichaReiser:
> I wonder if there's a way to generate these kinds from the grammar.. or make sure there aren't dangling keywords like in this example.

Yes, that would be a nice improvement. @IWANABETHATGUY would this be something you have an interest in?

IWANABETHATGUY:
> 

Yeah, let us open an issue to track this task.

IWANABETHATGUY:
Could we guarantee that all the node in `KIND_SRC.nodes` are defined in `xx.ungram` ?
More specifically, `SET(KIND_SRC.nodes)` == `SET(xx.ungram.definition)` ?

MichaReiser:
> `KIND_SRC`

I'm not entirely sure what you're saying. What I understand is if it is guaranteed that all nodes are defined in the `.ungram` file and there are no other nodes (or tokens). 

I think this is safe to assume because any node that isn't defined in the `ungram` file won't have a corresponding AST node.

IWANABETHATGUY:
> defined

I mean if some node_kind defined in `KIND_SRC.nodes` https://github.com/rome/tools/blob/5654fae83798224905e7f9c5a6fb9c79b9f9b125/xtask/codegen/src/kinds_src.rs#L187  but not in `xx.ungram` file, vice versa, it will not be safe to apply your advice the generating code would be broken.

IWANABETHATGUY:
sorry for my poor expression.

PR #3227, opened 2022-09-15
refactor(rome_formatter): Comments

MichaReiser:
## Summary

This PR is empty for now but I want to use it to merge all comment-related PRs into it as the individual PRs may not compile on their own. 

I split the whole refactory mainly on file boundaries to avoid having to deal with many merge conflicts between the different PRs. However, this means, that individual PRs won't compile nor will tests be passing. 

This is what this PR is for. It ensures that all changes compile together and that there are no regressing tests. The main motivation of splitting this PR into multiple smaller PRs is to ease reviewing. 

* #3230
* #3239 
* #3240
* #3241
* #3243
* #3244
* #3245
* #3246

### JSX Suppressions

It's now possible to suppress formatting of a JSX element or fragment by using

```javascript
<div>
	{/* rome-format ignore: reason */}
	<span>This element
  does not get formatted</span>
</div>
```

### Fix instability issues

Fixes #2768

### Formatting order

The comment formatting relied on the fact that tokens (and thus nodes) are formatted in the order they get written to the final document to be able to decide whether to insert a space before/after a comment. However, there is no enforcement of this constraint and ignoring it leads to very subtle spacing bugs that are hard to notice and track down. There are even situations where it's surprisingly hard, if not even impossible, to format the document in order if it is necessary to test if some other element will break that only gets written later. 

This PR removes the constraint that tokens must be formatted in order (they still must be written in the correct order to the output buffer).

## Performance

The overall performance is improving. The reasons for this are

* No longer necessary to track the formatted comments inside of the `FormatState`. The `FormatState` snapshot has a size of zero now, significantly reducing the cost of formatting statements. 
* The `GroupElementsBuffer` added a level of indirection to every `write`, `state`, `context` etc. calls. It also required to use a Boxed buffer snapshot to store its `empty` flag
* The old implementation iterated over all trivia pieces to extract the suppression comments AND when formatting the token. The new implementation only iterates over the pieces to extract the comments but not when formatting the tokens. 
* `FxHashMap`s are fast :) 

```
group                                    comments                               main
-----                                    --------                               ----
formatter/checker.ts                     1.00   246.5±18.26ms    10.5 MB/sec    1.07   263.8±20.88ms     9.9 MB/sec
formatter/compiler.js                    1.00   134.0±10.79ms     7.8 MB/sec    1.23   164.9±18.11ms     6.4 MB/sec
[...]

PR #3230, opened 2022-09-15
feat(rome_formatter): Comments multi map

MichaReiser:
## Summary

This PR is part of the Comments refactoring (#3227)

This PR implements a multimap for storing the leading, dangling, and trailing comments for nodes.

A naive implementation using three multimaps: one to store the leading, dangling, and trailing parts of each node, requires between `nodes < allocations < nodes * 3` `vec` allocations (`nodes` = number of nodes with at least one comment). 

 This purpose-built multimap uses a shared `Vec` to store all comments where this is possible (99.99% of comments) and only allocates node-specific `Vec`s for nodes appearing "out of order"

The idea behind the implementation is that comments normally remain in the same order as in the source document. That means:
* All comments for a node are inserted at "the same" time (one at a time, but before inserting the comments of the next node)
* It first inserts the leading, then dangling, and finally, a node's trailing comments.

Using this fact allows storing the vast majority of comments in a single `Vec` and only storing the offsets into this shared vec indicating the start/end of the leading/dangling/trailing comments. The implementation falls back to allocating (up to) three vectors for a node in case either of the above rules is violated.

Using a single vector for the majority of comments is also beneficial when it comes to iterating because:
* Retrieving the comments for a node requires a single hash map lookup instead of 3
* All comments for a node are located next to each other in the vector, making better use of cache locality. 

## Metrics

I used `countme` to verify that the hypothesis that comments appearing out of order are rare by formatting all benchmark files:

```bash
cargo run -p xtask_bench --release --features=count  -- --feature=formatter  --criterion=false
```

```
rome_formatter::comments::map::InOrderEntry           39_383       34_333            0
rome_formatter::comments::map::OutOfOrderEntry            33           22            0
                                                       total     max_live         live
```

Out of the close to 40k comments, only 33 nodes required their "own" `leading`, `dangling`, and `trailing` collections. That's 0.0837925%. Or, in other words. This implementation avoids between 40k and 120k vector allocations compared to using three general purpose multimap implementations.

## Test Plan
I added a handful of tests to the map implementation

### DIFF
--- a/crates/rome_formatter/src/comments/map.rs
+++ b/crates/rome_formatter/src/comments/map.rs
@@ -46,7 +46,7 @@ pub(super) struct CommentsMap<K, V> {
     /// Lookup table to retrieve the entry for a key.
     index: FxHashMap<K, Entry>,
 
-    /// Flat array storing all the values that have been inserted in order.
+    /// Flat array storing all the parts that have been inserted in order.
     parts: Vec<V>,
 
     /// Vector containing the leading, dangling, and trailing vectors for out of order entries.
@@ -120,7 +120,7 @@ impl<K: std::hash::Hash + Eq, V> CommentsMap<K, V> {
                 );
             }
 
-            // Has leading and dangling comments and its comments are at the end of values
+            // Has leading and dangling comments and its comments are at the end of parts
             Some(Entry::InOrder(entry))
                 if entry.trailing_end.is_none() && self.parts.len() == entry.range().end =>
             {
@@ -179,7 +179,7 @@ impl<K: std::hash::Hash + Eq, V> CommentsMap<K, V> {
     #[cold]
     fn entry_to_out_of_order<'a>(
         entry: &'a mut Entry,
-        values: &[V],
+        parts: &[V],
         out_of_order: &mut Vec<Vec<V>>,
     ) -> &'a mut OutOfOrderEntry
     where
@@ -189,9 +189,9 @@ impl<K: std::hash::Hash + Eq, V> CommentsMap<K, V> {
             Entry::InOrder(in_order) => {
                 let index = out_of_order.len();
 
-                out_of_order.push(values[in_order.leading_range()].to_vec());
-                out_of_order.push(values[in_order.dangling_range()].to_vec());
-                out_of_order.push(values[in_order.trailing_range()].to_vec());
+                out_of_order.push(parts[in_order.leading_range()].to_vec());
+                out_of_order.push(parts[in_order.dangling_range()].to_vec());
+                out_of_order.push(parts[in_order.trailing_range()].to_vec());
 
                 *entry = Entry::OutOfOrder(OutOfOrderEntry {
                     leading_index: index,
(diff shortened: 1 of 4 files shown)


crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+pub(super) struct CommentsMap<K, V> {
+    /// Lookup table to retrieve the entry for a key.
+    index: FxHashMap<K, Entry>,
+
+    /// Flat array storing all the values that have been inserted in order.
+    parts: Vec<V>,
+
+    /// Vector of vectors. Stores a triple of leading, dangling, and trailing [Vec] for every out of order entry.
+    ///
+    /// The length of this vec is guaranteed to always be a multiple of 3 where the leading parts are
```
ematipico, line 38:
```suggestion
    /// The length of this vec is guaranteed to always be a multiple of 3, where the leading parts are
```

ematipico: When you refer to "vec", which one do you mean? The outer or the inner? If you meant the outer, then why the `new` function doesn't create three vectors?

crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+            // Has only leading comments and no elements have been pushed since
+            Some(Entry::InOrder(entry))
+                if entry.trailing_start.is_none() && self.parts.len() == entry.range().end =>
+            {
+                self.parts.push(part);
+                entry.increment_leading_range();
+            }
+
+            Some(Entry::OutOfOrder(entry)) => {
+                let leading = &mut self.out_of_order[entry.leading_index()];
```
ematipico, line 77: Is the concept of "out of order" an implementation detail? I ask because it is mentioned in the documentation of the struct, but it's not mentioned when calling `push_leading`

crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+                leading.push(part);
+            }
+
+            Some(entry) => {
```
ematipico, line 81: When do we hit this variant? Asking because I read the function `entry_to_out_of_order` which does `match` against `entry`... so I  am confused 😅

crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+/// shared between all keys to significantly reduce the number of allocated [Vec]s.
+///
+/// The implementation does support insertions that don't comply with the above mentioned rules but
+/// these come at a performance penalty:
+/// * It requires allocating three [Vec], one for the leading, dangling, and trailing parts.
+/// * Already inserted parts must be copied over (by cloning) into these newly allocated [Vec]s.
+/// * Resolving the slices for every part requires an extra level of indirection.
+///
+/// ## Panics
+/// Panics when storing the `u32::MAX` part.
```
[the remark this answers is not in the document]
MichaReiser: I prefer having it at the top as it is a limitation of the whole data structure. But I also think it's not that much of a difference. It's merely a note if someone decides to use it for anything other than comments.

crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+
+            Some(entry) => {
+                let out_of_order =
+                    Self::entry_to_out_of_order(entry, &self.parts, &mut self.out_of_order);
+                self.out_of_order[out_of_order.trailing_index()].push(part);
+            }
+        }
+    }
+
+    #[cold]
```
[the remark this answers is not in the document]
MichaReiser: [cold](https://doc.rust-lang.org/reference/attributes/codegen.html#the-cold-attribute)

It hints to the compiler that this function isn't in the hot path... It's unlikely to be called. This allows the compiler to more heavily optimise the paths where this function isn't called.

crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+    fn entry_to_out_of_order<'a>(
```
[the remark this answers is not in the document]
MichaReiser: It converts an in order entry to an out of order entry.

crates/rome_formatter/src/comments/map.rs
```diff
@@ -0,0 +1,802 @@
...
+            index: FxHashMap::default(),
+            parts: Vec::new(),
+            out_of_order: Vec::new(),
+        }
+    }
+
+    /// Pushes a leading part for `key`.
+    pub fn push_leading(&mut self, key: K, part: V)
+    where
+        V: Clone,
```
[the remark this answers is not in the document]
MichaReiser: It allows me to not add the constraint to `Debug` and other implementations related to the map that don't need the `Clone` functionality

crates/rome_formatter/src/comments/map.rs
MichaReiser: It's the outer vec. The length is a multiple of three and an empty vec is valid because `0 * 3 = 0`.

ematipico (approved):

PR #3234, opened 2022-09-15
feat(rome_text_size): implement the JsonSchema trait on TextRange and TextSize

leops:
## Summary

This is another change extracted from #3222, this implements the `JsonSchema` trait on the `TextRange` and `TextSize` structs directly in the `rome_text_size` crate since now own it, instead relying on manual `-Schema` implementations that were actually incorrect.

## Test Plan

This PR only change how these types are represented in the `schemars` meta-type system, the only visible side-effect for this is that the `TextRange` type is now correctly declared as `[number, number]` in the generated TypeScript bindings

### DIFF
--- a/crates/rome_text_size/src/lib.rs
+++ b/crates/rome_text_size/src/lib.rs
@@ -23,6 +23,8 @@ mod range;
 mod size;
 mod traits;
 
+#[cfg(feature = "schemars")]
+mod schemars_impls;
 #[cfg(feature = "serde")]
 mod serde_impls;
 
--- a/npm/backend-jsonrpc/src/workspace.ts
+++ b/npm/backend-jsonrpc/src/workspace.ts
@@ -216,9 +216,10 @@ export interface GetSyntaxTreeResult {
 	cst: string;
 }
 export interface GetControlFlowGraphParams {
-	cursor: number;
+	cursor: TextSize;
 	path: RomePath;
 }
+export type TextSize = number;
 export interface GetFormatterIRParams {
 	path: RomePath;
 }
@@ -273,7 +274,7 @@ export type Severity = "Help" | "Note" | "Warning" | "Error" | "Bug";
  */
 export interface CodeSuggestion {
 	applicability: Applicability;
-	labels: TextRangeSchema[];
+	labels: TextRange[];
 	msg: MarkupBuf;
 	span: FileSpan;
 	style: SuggestionStyle;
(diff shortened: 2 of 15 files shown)


crates/rome_text_size/src/lib.rs
```diff
@@ -23,6 +23,8 @@ mod range;
 mod size;
 mod traits;
 
+#[cfg(feature = "schemars")]
```
MichaReiser, line 26: Is there a way to document features? If so, would you mind adding a sentence or two explaining what the `schemars` feature is about.

npm/backend-jsonrpc/src/workspace.ts
```diff
@@ -299,10 +300,7 @@ export type Applicability =
 	| "MaybeIncorrect"
 	| "HasPlaceholders"
 	| "Unspecified";
-export interface TextRangeSchema {
-	end: number;
-	start: number;
-}
+export type TextRange = [TextSize, TextSize];
```
MichaReiser, line 303: It's probably self explanatory that the first value is the `start` and the second value is the `end` but I found the old API more ergonomic as there are no functions hiding the "internal" layout of range.

npm/backend-jsonrpc/src/workspace.ts
```diff
@@ -216,9 +216,10 @@ export interface GetSyntaxTreeResult {
 	cst: string;
 }
 export interface GetControlFlowGraphParams {
-	cursor: number;
+	cursor: TextSize;
 	path: RomePath;
 }
+export type TextSize = number;
```
MichaReiser, line 222: It's sad that TypeScript doesn't support opaque types because this type alias doesn't give us any type safety, it only serves documentary purpose.

MichaReiser (approved):

npm/backend-jsonrpc/src/workspace.ts
leops: To be clear this PR doesn't change how the `TextRange` struct is serialized, this type has always been represented as a tuple of two numbers, but because it has a custom implementation of the `serde` traits to do this without a corresponding implementation of `JsonSchema` the generated TypeScript type for the struct was incorrect

ematipico (approved):
Could you update this line and remove the TODO, now that we fixed the issue? https://github.com/rome/tools/blob/main/npm/rome/src/index.ts#L249-L252

PR #3235, opened 2022-09-15
docs: fix line break in `noUnusedVariables` lint rule

kidonng:
## Summary

The docs is missing a line break which resulted in [incorrect rendering](https://rome.tools/docs/lint/rules/nounusedvariables/):

image

### DIFF
--- a/website/src/docs/lint/rules/noUnusedVariables.md
+++ b/website/src/docs/lint/rules/noUnusedVariables.md
@@ -11,6 +11,7 @@ There are two exceptions to this rule:
 
 1. variables that starts with underscore, ex: `let _something;`
 2. the `React` variable;
+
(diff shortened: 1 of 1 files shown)


ematipico:
I am not really sure what's the issue, but that text is auto generated from https://github.com/rome/tools/blob/main/crates/rome_js_analyze/src/semantic_analyzers/nursery/no_unused_variables.rs#L13-L25

You would need to change the text there and run `cargo lintdoc` to update it.

leops:
I think the underlying issue is that we need to insert a newline after the end of a list tag in `lintdoc`:
https://github.com/rome/tools/blob/9c73d1984d7970e28800fd6612f0ab79327f5a20/xtask/lintdoc/src/main.rs#L318-L320

ematipico:
Closed in favour of https://github.com/rome/tools/pull/3247

PR #3236, opened 2022-09-16
fix(rome_formatter): Panic for out of order group id

MichaReiser:
## Summary

This PR fixes an issue where the printer panics if group ids aren't written to the document in the same order as they have been created because `resize` shrinks a vector if its size exceeds the argument.

## Test Plan

I added a new printer test verifying that the printer no longer panics if group ids aren't added in the same order as they have been created with `f.group_id`.


MichaReiser:
!bench_formatter

ematipico (approved):

PR #3238, opened 2022-09-16
feat(rome_js_analyzer): `useButtonType`

### LINKED ISSUE #3233: `useButtonType`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/useButtonType.md

ematipico:
## Summary

Close #3233 

## Test Plan

Created new test cases to cover the new rule.


crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+    type State = UseButtonTypeState;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let node = ctx.query();
+
+        match node {
+            UseButtonTypeQuery::JsxOpeningElement(opening_element) => {
+                let name = opening_element.name().ok()?;
+                // we bail early the current tag is not a button; case sensitive is important
```
MichaReiser, line 69: I think it would be important to explain why `case sensitive` is important rather than just saying that it is.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                if name.syntax().text_trimmed() == "button" {
```
MichaReiser, line 70: `text_trimmed` is somewhat expensive on a node because it must:

* find the first and last token
* Allocate a `SyntaxNodeText` and then do the comparison

The good thing is, you can simplify this (and remove the case sensitiveness as well) by explicitly matching on the `JsxAnyElementName`. The only variant that is of importance for this rule is `JsxName` 

```
fn is_button(name: &JsxAnyElementName) -> bool {
  match name {
    JsxAnyElementName::JsxName(name) => {  
      name.value_token().map_or(false, |name| name.text_trimmed() == "button")
    },
    _ => false
  }
}
```

You can then access the `name` token directly which should be significantly faster

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                    let attributes = opening_element.attributes();
+                    if attributes.len() == 0 {
```
MichaReiser, line 72: Nit
```suggestion
                    if attributes.is_empty() {
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                            if name == "type" {
+                                let initializer = attribute.initializer()?.value().ok()?;
+                                let initializer = initializer.as_jsx_string()?;
+                                if !ALLOWED_BUTTON_TYPES
+                                    .contains(&&*initializer.inner_string_text().ok()?)
+                                {
+                                    return Some(UseButtonTypeState::from(initializer.clone()));
+                                }
+                            }
+                        }
```
MichaReiser, line 89: Nit: I would find it clearer to read if this is split into two:

1. Finding the `type` attribute (using `find`)
2. Handling the value of a potentially invalid type attribute


What is the expected outcome for repeated `type` attributes or when used with jsx spread where it is unclear if the type will be overridden. 

```js
<button type="test" type="button" />
<button type="button", type="test" />
<button type="button" {...rest} />
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                if name.syntax().text_trimmed() == "button" {
+                    let attributes = opening_element.attributes();
+                    if attributes.len() == 0 {
+                        return Some(UseButtonTypeState::from(opening_element.clone()));
+                    } else {
+                        for attribute in attributes {
+                            let attribute = attribute.as_jsx_attribute()?;
+                            let jsx_name = attribute.name().ok()?;
+                            let jsx_name = jsx_name.as_jsx_name()?;
+                            let name = jsx_name.syntax().text_trimmed();
```
MichaReiser, line 79: Same here: Can you access the `value_token` instead and get the trimmed text from their

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                    JsAnyExpression::JsIdentifierExpression(identifier_expression) => {
+                        is_react_create_element(
+                            PossibleCreateElement::from(identifier_expression),
+                            model,
+                        )
+                    }
+                    _ => return None,
+                }?;
+
+                if is_create_element {
```
MichaReiser, line 110: The caller of `is_create_element` still needs a lot of knowledge about how `is_react_create_element` works internally. 

Would it make sense to change `is_react_create_element` to instead return a `struct` with the information necessary for the caller and it would be possible to pass any `JsAnyExpression` to `is_react_create_element`

```rust
fn is_react_create_element(expression: &JsAnyExpression) -> Option<ReactCreateElementCall> {
  .. 
}

struct ReactCreateElementCall { 
  typeName: JsAnyExpression,
  props: JsObjectExpression,
  children: Option<JsArrayExpression>
}
```

This would simplify a lot of the code below

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                                for member in members {
+                                    let member = member.ok()?;
+                                    let property = member.as_js_property_object_member()?;
+                                    let property_name = property.name().ok()?;
+                                    let property_value = property.value().ok()?;
+
+                                    let property_name =
+                                        property_name.as_js_literal_member_name()?;
+                                    // we found the correct member, we can bail
+                                    if property_name.name().ok()? == "type" {
```
MichaReiser, line 140: Nit: Same as for the attribute. I would find this easier to read if it is split into two:

1. Finding a member called `type`
2. Testing if it is an allowed button type

What's the expected outcome for

```javascript
react.createElement('button', {
  type: "button",
  ...more,
})
```

Where it is unknown if `...more` overrides `type`. 

or 

```javascript
react.createElement('button', {
  type: "button",
  type: "rubbish",
}
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                                            .contains(&&*value.inner_string_text().ok()?)
+                                        {
+                                            return Some(UseButtonTypeState::from(value.clone()));
+                                        }
+                                    }
+                                }
+
+                                // if we are here, it means that we haven't found the property "type" and
+                                // we have return an error
+                                return Some(UseButtonTypeState::from(second_argument.clone()));
```
MichaReiser, line 155: Why is it that it must return an error in the case the `type` is missing. My understanding is that `button` without a type is valid HTML.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+                                        if !ALLOWED_BUTTON_TYPES
+                                            .contains(&&*value.inner_string_text().ok()?)
+                                        {
+                                            return Some(UseButtonTypeState::from(value.clone()));
+                                        }
+                                    }
+                                }
+
+                                // if we are here, it means that we haven't found the property "type" and
+                                // we have return an error
```
MichaReiser, line 154:
```suggestion
                                // we have to return an error
```

Why?

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    fn diagnostic(_ctx: &RuleContext<Self>, state: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic::new(
+            state.syntax().text_trimmed_range(),
+            markup! {
+                "Provide an explicit "<Emphasis>"type"</Emphasis>" prop on "<Emphasis>"button"</Emphasis>" elements."
+            },
+        )
+            .footer(Severity::Note, markup! {
+                "The default type of a button is "<Emphasis>"submit"</Emphasis>", which causes a page reload and is not a typical behavior in a React application."
```
MichaReiser, line 176: This sounds as if `submit` should never be used but this isn't true. Not using `submit` results in degraded accessibility and it is better to prevent the default handler of the form. 

This message is also somewhat confusing where the help text proposes to use `submit`. Should I now use it or not?

crates/rome_js_syntax/src/jsx.rs
```diff
@@ -0,0 +1,24 @@
+use crate::{JsxString, TextSize};
+use rome_rowan::{SyntaxResult, SyntaxTokenText};
```
MichaReiser, line 2: All other files implementing extensions use the `syntax_ext` pattern

crates/rome_js_syntax/src/jsx.rs
```diff
@@ -0,0 +1,24 @@
...
+
+impl JsxString {
+    /// Get the inner text of a string not including the quotes
```
MichaReiser, line 5: Could you add a doctest example to the new API

MichaReiser (approved):
What's the motivation behind renaming `util` to `shared`. I don't think it solves any of the problems that I see with a `util`/`shared`/`common`/`arbitrary`.... folder

I also think that it would be helpful to split the utils refactor into its own PR if we want to move forward with it as it is unrelated to the other changes in this PR.

MichaReiser (changes requested):
Wrong button... How can I undo an approve and change it to comment xD

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
ematipico: The rule is meant to **enforce** the use the of `type` on a `button` element. While doing `<button>Button</button>` is correct, most of the times is not what the user meant

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
ematipico: The message comes from Rome classic, so I assumed that the message was already vetted. Can you suggest some change please?

MichaReiser: I think there are two fundamental issues with the messaging:

a) It uses the same message for the case where there is no type and there's a potentially invalid type. We should split the two (e.g. the message seems out of place if I wrote `button type="rest"`)
b) What I pointed out before. that it sounds like `submit` is always the incorrect button type. Have you checked out the eslint plugin react?

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
ematipico: The expected outcome for this rule is to work on a single prop only. The rule is not in charge of checking these cases:
1. a repeated property is broader issue, and maybe we could evaluate a rule for that case;
2. we don't have enough information to check  - **as for today** - the safety of `...rest`

ematipico: > (using `find`)

I am not sure this will improve the readability. In order to find the correct attribute using `.find`, we have to manually **handle six `Option`** types without the try operator.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
ematipico: > Have you checked out the eslint plugin react?

It's not as explicit as we are

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
MichaReiser: You can use `find_map` and you can then use the `try` operator

crates/rome_js_analyze/src/react.rs
```diff
@@ -0,0 +1,132 @@
...
+///
+/// There are two cases:
+///
+/// First case
+/// ```js
+/// React.createElement()
+/// ```
+/// We check if the node is a static member expression with the specific members. Also, if `React`
+/// has been imported in the current scope, we make sure that the binding `React` has been imported
+/// from the `"react"` module.
```
leops, line 30: I couldn't find the point in the function where we check that the source of the import is "react", where does that happen ?

crates/rome_js_analyze/src/react.rs
```diff
@@ -0,0 +1,132 @@
...
+                && member.syntax().text_trimmed() == "createElement";
+
+            if maybe_from_react {
+                let identifier_binding = model.declaration(&identifier);
+                if let Some(binding_identifier) = identifier_binding {
+                    let binding_identifier =
+                        JsIdentifierBinding::cast_ref(binding_identifier.syntax())?;
+                    for ancestor in binding_identifier.syntax().ancestors() {
+                        if ancestor.kind() == JS_IMPORT {
+                            maybe_from_react = binding_identifier.syntax().text_trimmed()
```
leops, line 62: Looking only at the text of the identifier instead of its "semantic provenance" seems a bit fragile, while these are probably not common we could handle a few other cases such as:
```js
import OtherName from "react";
OtherName.createElement();
```
```js
import { createElement as otherName } from "react";
otherName();
```
And I think we should also have a test for this pattern:
```js
import * as React from "react";
React.createElement();
```

crates/rome_js_analyze/src/react.rs
ematipico: This fix will be applied in https://github.com/rome/tools/pull/3256

ematipico: I am going to port the fix in this PR

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_button_type.rs
```diff
@@ -0,0 +1,221 @@
...
+            (markup!{
+                "Provide a correct "<Emphasis>"type"</Emphasis>" prop for the "<Emphasis>"button"</Emphasis>" element."
+            }).to_owned()
+        };
+        Some(RuleDiagnostic::new(
+            state.node.syntax().text_trimmed_range(),
+            message
+        )
+            .footer(Severity::Note, markup! {
+                "The default  "<Emphasis>"type"</Emphasis>" of a button is "<Emphasis>"submit"</Emphasis>", which causes the submission of a form when placed inside a `form` element. "
```
MichaReiser, line 198: Thanks for iterating on the message. 

I'm still conflicted by the rule. Using a button with type submit is exactly what you want inside of a form for accessibility reasons. However, this message incentivices users to not use `submit`.... 

To make it worse, this rule will be in direct conflict with an a11 rule that recommends having at least one `submit` button in every form.

MichaReiser (commented):
The implementation looks good to me but I still think the rule is misleading and results in users favouring `button` over `submit` which results in a degraded accessibility.

MichaReiser (approved):

ematipico:
> The implementation looks good to me but I still think the rule is misleading and results in users favouring `button` over `submit` which results in a degraded accessibility.

Should we just remove the footer note?

PR #3242, opened 2022-09-16
feat(playground): enable nursery lint rules

ematipico:
## Summary

This PR adds a new checkbox to the playground to enable nursery rules. This allows us to debug these rules that are still in development. 

I didn't bother much about where to put that checkbox... I am sure we would need an overhaul of the UI, in a separate place.

## Test Plan

Play with the deployed background

### DIFF
--- a/website/playground/package.json
+++ b/website/playground/package.json
@@ -36,7 +36,6 @@
 		"@vitejs/plugin-react": "^1.0.7",
 		"autoprefixer": "^10.4.2",
 		"postcss": "^8.4.6",
-		"rome": "next",
 		"tailwindcss": "^3.0.19",
 		"typescript": "^4.8.2",
 		"vite": "2.9.5",
--- a/website/playground/package.json
+++ b/website/playground/package.json
@@ -10,7 +10,7 @@
 		"build:wasm": "wasm-pack build --out-dir ../../npm/wasm-web --target web --release --scope rometools ../../crates/rome_wasm",
 		"build:wasm-dev": "wasm-pack build --out-dir ../../npm/wasm-web --target web --dev --scope rometools ../../crates/rome_wasm",
 		"format": "cargo rome-cli-dev format --write ./src",
-		"check": "cargo rome-cli-dev check ./src",
+		"check": "cd ../../ && cargo rome-cli-dev check ./website/playground/src",
 		"tsc": "tsc"
 	},
 	"dependencies": {
--- a/website/playground/src/NurseryRules.tsx
+++ b/website/playground/src/NurseryRules.tsx
@@ -8,10 +8,10 @@ export default function NurseryRules({
 	enabledNurseryRules,
 }: Props) {
 	return (
-		<div className="pl-5 pb-5">
+		<div className="pl-5">
 			<fieldset className="flex items-center">
 				<legend className="sr-only">Linter</legend>
-				<div className="relative flex p-5 pl-1 pt-7 pb-0">
+				<div className="relative flex p-5 pt-0 pl-1">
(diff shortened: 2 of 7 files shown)


website/playground/package.json
```diff
@@ -10,7 +10,7 @@
 		"build:wasm": "wasm-pack build --out-dir ../../npm/wasm-web --target web --release --scope rometools ../../crates/rome_wasm",
 		"build:wasm-dev": "wasm-pack build --out-dir ../../npm/wasm-web --target web --dev --scope rometools ../../crates/rome_wasm",
 		"format": "cargo rome-cli-dev format --write ./src",
-		"check": "cargo rome-cli-dev check ./src",
+		"check": "cd ../../ && cargo rome-cli-dev check ./website/playground/src",
```
MichaReiser, line 13: Why has it become necessary to traverse upwards?

website/playground/package.json
```diff
@@ -36,7 +36,6 @@
 		"@vitejs/plugin-react": "^1.0.7",
 		"autoprefixer": "^10.4.2",
 		"postcss": "^8.4.6",
-		"rome": "next",
```
MichaReiser, line 39: What's the reason behind removing the `rome` dependency

website/playground/src/NurseryRules.tsx
```diff
@@ -0,0 +1,38 @@
+interface Props {
+	setEnabledNurseryRules: (b: boolean) => void;
+	enabledNurseryRules: boolean;
+}
+
+export default function NurseryRules({
```
MichaReiser, line 6: image

The check box isn't properly aligned with the other check boxes

website/playground/src/romeWorker.ts
```diff
@@ -81,31 +82,46 @@ self.addEventListener("message", async (e) => {
...
+				configuration.linter = {
+					enabled: true,
+					rules: {
+						nursery: {
+							noNewSymbol: "error",
+							noDangerouslySetInnerHtml: "error",
+							noUnusedVariables: "error",
+							noUnreachable: "error",
+							useCamelCase: "error",
 						},
```
MichaReiser, line 119: Is there no other way to enable all nursery rules? How to we prevent having to change this list with every new nursery rule that we add (or remove)?

MichaReiser (changes requested):

website/playground/package.json
ematipico: This is an existing bug we have in the resolution of the configuration. At the moment we are only able resolve the configuration file from the working directory.

I am already working on fixing the issue, but it requires some refactor on our file system. This is a temporary workaround to let the command to pick up the `rome.json` file

website/playground/package.json
ematipico: Because we don't use it

website/playground/src/romeWorker.ts
ematipico: Unfortunately, it's not possible at the moment. We can investigate how to solve this issue

website/playground/src/NurseryRules.tsx
ematipico: I am not an expert of tailwind. Do you know how to fix it?

MichaReiser: I have no prior experience with tailwind. I do think that our playground, being actively promoted on the website, is worth spending time on to ensure it looks decent, similar to diagnostics which we polish too.

MichaReiser (approved):
Nit: Maybe rename it to *nursery lint rules* to make it clear that it's about linting and not formatting.

PR #3243, opened 2022-09-16
refactor(rome_formatter): Part 5: Comments formatting

MichaReiser:
## Summary

This PR is part of the comments refactoring #3227.

* It removes the old builders for formatting the leading and trailing token trivia and introduces new builders to format the leading, dangling, and trailing node comments. 
* Introduces a new `soft_line_indent_or_spaced` builder that replaces `format_delimited(..).soft_block_or_spaces()`
* Deletes the no longer needed `GroupsElementsBuffer` (yay)
* Removes any (release) state from `FormatState` as it is no longer necessary to track the last token to know if spaces must be inserted between the last token and comments (is now the responsibility of the formatter that calls `format_dangling_comments`).
* 
 
## Test Plan

Added a doc test for the new `soft_line_indent_or_spaced` builder.


crates/rome_formatter/src/builders.rs
```diff
@@ -1131,6 +1127,100 @@ impl<Context> std::fmt::Debug for BlockIndent<'_, Context> {
...
+///         text("}")
+///     ])
+/// ]).unwrap();
+///
+/// assert_eq!(
+///     "{ a: 5 }",
+///     elements.print().as_code()
+/// );
+/// ```
+pub fn soft_line_indent_or_spaced<Context>(
```
ematipico, line 1187: nit: Initially, I thought that `soft_line_indent_or_spaced` worked the same as  `soft_line_break_or_space` (and other similar functions). Maybe adding a `with_soft_line_indents_or_spaces` can help.

Actually I just had the same problem with `soft_line_indent_or_space`...

ematipico (approved):

crates/rome_formatter/src/builders.rs
MichaReiser: `soft_line_indent_or_spaced` now works the same as `soft_block_indent` and `block_indent` in the sense that it adds spaces/line breaks before and after the content. The main exception here is `soft_line_indent_or_space` that only adds a space/line break BEFORE the content but not after. 

I don't understand what you mean by `with_soft_line_indents_or_spaces`. Would you be able to make an example?

ematipico: > `soft_line_indent_or_spaced` now works the same as `soft_block_indent` and `block_indent` in the sense that it adds spaces/line breaks before and after the content. The main exception here is `soft_line_indent_or_space` that only adds a space/line break BEFORE the content but not after.
> 
> I don't understand what you mean by `with_soft_line_indents_or_spaces`. Would you be able to make an example?

The way my brain works, I get straightaway how `soft_block_indent` and `block_indent` work because of the word "block". I get `soft_line_indent_or_space` because its usage is in line with `space` and `soft_line_break` (and the name too).

I associated this new API `soft_line_indent_or_spaced` to `soft_line_indent_or_space`, `space` and `soft_line_break`, but it works like the blocks. That's all

MichaReiser: How about `soft_line_or_space_block_indent`?

ematipico: I like the second suggestion! Thank you for taking the time for coming up with another name :)

PR #3250, opened 2022-09-19
feat(rome_Formatter): Assert formatting of comments

MichaReiser:
## Summary

This PR adds a debug assertion to verify if all comments in a program have been formatted to prevent accidental "dropping" of comments.

## Test Plan

I removed some comments formatting to verify that the panic is triggered.

### DIFF
--- a/crates/rome_formatter/Cargo.toml
+++ b/crates/rome_formatter/Cargo.toml
@@ -13,7 +13,6 @@ rome_rowan = { path = "../rome_rowan" }
 tracing = { version = "0.1.31", default-features = false, features = ["std"] }
 serde = { version = "1.0.136", features = ["derive"], optional = true }
 cfg-if = "1.0.0"
-indexmap = "1.8.2"
 schemars = { version = "0.8.10", optional = true }
 rustc-hash = "1.1.0"
 countme = "3.0.1"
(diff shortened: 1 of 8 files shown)


crates/rome_formatter/Cargo.toml
```diff
@@ -13,7 +13,6 @@ rome_rowan = { path = "../rome_rowan" }
 tracing = { version = "0.1.31", default-features = false, features = ["std"] }
 serde = { version = "1.0.136", features = ["derive"], optional = true }
 cfg-if = "1.0.0"
-indexmap = "1.8.2"
```
MichaReiser, line 16: The last remaining use of `IndexMap` was for the `Debug` implementation and it was easy enough to replace it with a `FxHashMap<Interned, usize>` so that this dependency is no longer needed.

ematipico: Does the new map keep the ordering? Or is the ordering not needed anymore?

MichaReiser: The `Display` implementation used the `IndexMap` to assign a unique number to every `Interned` element in increasing order by using the `index` assigned by the `IndexMap`. 

However, this can be easily replaced by using a `FxMap<Interned, ussize>` that explicitly stores the index of every element (which index map would probably do so too?). Computing the next index is straightforward as it always is `map.len()` because the implementation only appends entry but never removes any.

ematipico (approved):

Issue #3211, opened 2022-09-13
🐛 In JSX, some spaces are removed


pradel:
### Environment information

```bash
Hey all, I was trying Rome in one of my codebases using prettier and found an issue with spaces in JSX expressions.
As you can see in this playground, a space right after the `a` tag is removed, prettier is behaving normally and keeping that space.

[Playground link](https://play.rome.tools/?code=const+a+%3D+%3Cdiv%3E%0A++Text+%3Ca+data-very-long-prop-breakline-rome-playground+data-other%3E%0A++++++++++++++some+link%0A++++++++++++%3C%2Fa%3E%7B%27+%27%7D%0A++++++++++++%7C+some+other+text%2C%7B%27+%27%7D%0A++++++++++++++++%3C%2Fdiv%3E&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&typescript=false&jsx=false#YwBvAG4AcwB0ACAAYQAgAD0AIAA8AGQAaQB2AD4ACgAgACAAVABlAHgAdAAgADwAYQAgAGQAYQB0AGEALQB2AGUAcgB5AC0AbABvAG4AZwAtAHAAcgBvAHAALQBiAHIAZQBhAGsAbABpAG4AZQAtAHIAbwBtAGUALQBwAGwAYQB5AGcAcgBvAHUAbgBkACAAZABhAHQAYQAtAG8AdABoAGUAcgA+AAoAIAAgACAAIAAgACAAIAAgACAAIAAgACAAIAAgAHMAbwBtAGUAIABsAGkAbgBrAAoAIAAgACAAIAAgACAAIAAgACAAIAAgACAAPAAvAGEAPgB7ACcAIAAnAH0ACgAgACAAIAAgACAAIAAgACAAIAAgACAAIAB8ACAAcwBvAG0AZQAgAG8AdABoAGUAcgAgAHQAZQB4AHQALAB7ACcAIAAnAH0ACgAgACAAIAAgACAAIAAgACAAIAAgACAAIAAgACAAIAAgADwALwBkAGkAdgA+AA==)
```

### What happened?

[...]

PR #3251, opened 2022-09-19
fix(rome_js_formatter): in JSX, some spaces are removed #3211

denbezrukov:
## Summary

Fix https://github.com/rome/tools/issues/3211

[Prettier](https://github.com/prettier/prettier/blob/b0d9387b95cdd4e9d50f5999d3be53b0b5d03a97/src/language-js/print/jsx.js#L144-L180) removes `softline`, `hardline` before and after `jsxWhitespace`.

## Open Question

There is another one difference between Rome and Prettier. Prettier adds `jsxWhitespace` and break a line after `Text`. But IR the same. Can it be a problem with `fill` builder?

Rome:
```
                 ...
		Text <a data-very-long-prop-breakline-rome-playground data-other>
			some link
		</a>
		....

```

Prettier:
```
                 ...
		Text{" "}
		<a data-very-long-prop-breakline-rome-playground data-other>
			some link
		</a>
		 ...
```

[Playground](https://play.rome.tools/?code=const+a+%3D+%3Cdiv%3E%0A++Text+%3Ca+data-very-long-prop-breakline-rome-playground+data-other%3E%0A++++++++++++++some+link%0A++++++++++++%3C%2Fa%3E%7B%27+%27%7D%0A++++++++++++%7C+some+other+text%2C%7B%27+%27%7D%0A++++++++++++++++%3C%2Fdiv%3E&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&typescript=false&jsx=false#YwBvAG4AcwB0ACAAYQAgAD0AIAA8AGQAaQB2AD4ACgAgACAAVABlAHgAdAAgADwAYQAgAGQAYQB0AGEALQB2AGUAcgB5AC0AbABvAG4AZwAtAHAAcgBvAHAALQBiAHIAZQBhAGsAbABpAG4AZQAtAHIAbwBtAGUALQBwAGwAYQB5AGcAcgBvAHUAbgBkACAAZABhAHQAYQAtAG8AdABoAGUAcgA+AAoAIAAgACAAIAAgACAAIAAgACAAIAAgACAAIAAgAHMAbwBtAGUAIABsAGkAbgBrAAoAIAAgACAAIAAgACAAIAAgACAAIAAgACAAPAAvAGEAPgB7ACcAIAAnAH0ACgAgACAAIAAgACAAIAAgACAAIAAgACAAIAB8ACAAcwBvAG0AZQAgAG8AdABoAGUAcgAgAHQAZQB4AHQALAB7ACcAIAAnAH0ACgAgACAAIAAgACAAIAAgACAAIAAgACAAIAAgACAAIAAgADwALwBkAGkAdgA+AA==)

## Test Plan

Add new test cases. 
`cargo test -p rome_js_formatter`

This PR:
**Average compatibility**: 88.44
**Compatible lines**: 89.60

Main:
**Average compatibility**: 88.42
**Compatible lines**: 89.30


crates/rome_js_formatter/src/jsx/lists/child_list.rs
```diff
@@ -152,16 +153,21 @@ impl FormatRule<JsxChildList> for FormatJsxChildList {
                             // <pre className="h-screen overflow-y-scroll" />
                             // adefg
                             // ```
-                            if matches!(non_text, JsxAnyChild::JsxSelfClosingElement(_)) {
+                            if matches!(non_text, JsxAnyChild::JsxSelfClosingElement(_))
+                                && word.len() != TextSize::from(1)
```
denbezrukov, line 157: https://github.com/prettier/prettier/blob/b0d9387b95cdd4e9d50f5999d3be53b0b5d03a97/src/language-js/print/jsx.js#L404

crates/rome_js_formatter/src/utils/jsx.rs
```diff
@@ -122,7 +122,7 @@ impl Format<JsFormatContext> for JsxSpace {
         write![
             formatter,
             [
-                if_group_breaks(&format_args![JsxRawSpace, hard_line_break()]),
+                if_group_breaks(&format_args![JsxRawSpace, soft_line_break()]),
```
denbezrukov, line 125: https://github.com/prettier/prettier/blob/b0d9387b95cdd4e9d50f5999d3be53b0b5d03a97/src/language-js/print/jsx.js#L120

crates/rome_js_formatter/src/jsx/lists/child_list.rs
MichaReiser: Can you share a example of what formatting this is fixing. I intentionally ignored it back then when I implemented the JSX formatting because I didn't understand why it is needed... 

https://github.com/prettier/prettier/blob/b0d9387b95cdd4e9d50f5999d3be53b0b5d03a97/src/language-js/print/jsx.js#L420-L425

Has the same check and integrating this would be somewhat hard.

denbezrukov: [Playground](https://play.rome.tools/?code=b4+%3D+%28%0A%09%3C%3E%0A%09%09%3Cpre+%2F%3Ea%0A%09%3C%2F%3E%0A%29%3B%0A%2F%2F+longer+than+one+length+text+should+insert+hard+line+break%0Ab5+%3D+%28%0A%09%3C%3E%0A%09%09%3Cpre+%2F%3E%0A%09%09ab%0A%09%3C%2F%3E%0A%29%3B%0A%0Ab6+%3D+%28%0A%09%3C%3E%0A%09%09%3Cb+%2F%3Ea+bb+cc+dd+ee%7B%22+%22%7D%0A%09%3C%2F%3E%0A%29%3B%0A%2F%2F+longer+than+one+length+text+should+insert+hard+line+break%0Ab7+%3D+%28%0A%09%3C%3E%0A%09%09%3Cb+%2F%3E%0A%09%09longer+bb+cc+dd+ee%7B%22+%22%7D%0A%09%3C%2F%3E%0A%29%3B&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&typescript=false&jsx=false#YgA0ACAAPQAgACgACgAJADwAPgAKAAkACQA8AHAAcgBlACAALwA+AGEACgAJADwALwA+AAoAKQA7AAoALwAvACAAbABvAG4AZwBlAHIAIAB0AGgAYQBuACAAbwBuAGUAIABsAGUAbgBnAHQAaAAgAHQAZQB4AHQAIABzAGgAbwB1AGwAZAAgAGkAbgBzAGUAcgB0ACAAaABhAHIAZAAgAGwAaQBuAGUAIABiAHIAZQBhAGsACgBiADUAIAA9ACAAKAAKAAkAPAA+AAoACQAJADwAcAByAGUAIAAvAD4ACgAJAAkAYQBiAAoACQA8AC8APgAKACkAOwAKAAoAYgA2ACAAPQAgACgACgAJADwAPgAKAAkACQA8AGIAIAAvAD4AYQAgAGIAYgAgAGMAYwAgAGQAZAAgAGUAZQB7ACIAIAAiAH0ACgAJADwALwA+AAoAKQA7AAoALwAvACAAbABvAG4AZwBlAHIAIAB0AGgAYQBuACAAbwBuAGUAIABsAGUAbgBnAHQAaAAgAHQAZQB4AHQAIABzAGgAbwB1AGwAZAAgAGkAbgBzAGUAcgB0ACAAaABhAHIAZAAgAGwAaQBuAGUAIABiAHIAZQBhAGsACgBiADcAIAA9ACAAKAAKAAkAPAA+AAoACQAJADwAYgAgAC8APgAKAAkACQBsAG8AbgBnAGUAcgAgAGIAYgAgAGMAYwAgAGQAZAAgAGUAZQB7ACIAIAAiAH0ACgAJADwALwA+AAoAKQA7AA==)

MichaReiser: That make sense. Do you know what the use cases are in practice? I didn't implement because I wasn't able to come up with a practical use case where the difference should matter. Are you aware of a practical use case, if so, would you mind adding comment in the code mentioning it? If not, then I'm not sure if we have to use the exact same formatting (feels to me that this might be some special FB internal formatting rule)

Will it be necessary for us to also add the special ruling for the link I posted

MichaReiser:
Regarding the open question, there's one difference that may be of importance:


```
// Prettier
"Text",
ifBreak(['{" "}', softline], " "),

// Rome 
fill([
  "Text",
  [],
  [
    if_group_breaks(["{" "}", soft_line_break]),
    if_group_fits_on_line([" "])
  ],
[],
```

Prettier prints the `ifBreak` into the separator slot of fill (fill is a sequence of item, separator, item, separator....) and a separator gets printed in expanded mode if:

* previous item, separator, next item dont fit on the line

For items, the item is only printed if it doesn't fit on the line, regardless of the next item or separator.

crates/rome_js_formatter/src/jsx/lists/child_list.rs
denbezrukov: No, I don't know.
https://github.com/prettier/prettier/pull/4717 I found this issue. 
And it has the [case](https://play.rome.tools/?code=x+%3D%0A++%3Cdiv%3E%0A++++%3Cdiv%3E%0A++++++First%0A++++%3C%2Fdiv%3E%0A++++-%0A++++%3Cdiv%3E%0A++++++Second%0A++++%3C%2Fdiv%3E%0A++%3C%2Fdiv%3E%0A&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&typescript=false&jsx=false#eAAgAD0ACgAgACAAPABkAGkAdgA+AAoAIAAgACAAIAA8AGQAaQB2AD4ACgAgACAAIAAgACAAIABGAGkAcgBzAHQACgAgACAAIAAgADwALwBkAGkAdgA+AAoAIAAgACAAIAAtAAoAIAAgACAAIAA8AGQAaQB2AD4ACgAgACAAIAAgACAAIABTAGUAYwBvAG4AZAAKACAAIAAgACAAPAAvAGQAaQB2AD4ACgAgACAAPAAvAGQAaQB2AD4ACgA=) .

BTW: I can't find these examples in rome prettier spec. Did we miss `jsx` prettier specs?

MichaReiser: Oh wow, that's why. I was surprised why there are so little JSX tests but the whole folder is missing. I can create a PR to add the tests to our test suite. 

The `:` and `-` examples are nice but I'm a bit worried how it interacts even with any other single letter content which is why I'm not entirely sure it's worth adding but I also don't object since you already added it. The only question is if we need to add the `width == 1` check to the separator as well (code I linked to)

denbezrukov: We can add specs only for `:` and `-`. What do you think?

> Oh wow, that's why. I was surprised why there are so little JSX tests but the whole folder is missing. I can create a PR to add the tests to our test suite.

I can try to add it (:

MichaReiser: Oh sorry, I already did so in https://github.com/rome/tools/pull/3255 . I've only seen your comment now. 

How about any punctuation characters instead? That would include `.` and `,` as well.

denbezrukov:
> Regarding the open question, there's one difference that may be of importance:
> 
> ```
> // Prettier
> "Text",
> ifBreak(['{" "}', softline], " "),
> 
> // Rome 
> fill([
>   "Text",
>   [],
>   [
>     if_group_breaks(["{" "}", soft_line_break]),
>     if_group_fits_on_line([" "])
>   ],
> [],
> ```
> 
> Prettier prints the `ifBreak` into the separator slot of fill (fill is a sequence of item, separator, item, separator....) and a separator gets printed in expanded mode if:
> 
> * previous item, separator, next item dont fit on the line
> 
> For items, the item is only printed if it doesn't fit on the line, regardless of the next item or separator.

Could you please explain more? What is the difference between Prettier `ifBreak` and Rome  `if_group_breaks` + `if_group_fits_on_line` inside fill? :)

crates/rome_js_formatter/src/jsx/lists/child_list.rs
denbezrukov: `.` `,` `?` `-` `:` `;` `/` `…` `!` `—` `‘` `“` `(` `)` `[` `]`

MichaReiser: How about using [is_ascii_punctuation](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_punctuation) ?

denbezrukov: Oops, 
I've missed it (: Thank you!

MichaReiser:
As a small note. I noticed an instability with JSX formatting when adding the prettier tests and the fix is now included in the comments formatting. It required to move the best fitting element to the `FormatJsxElement`. I don't think this will effect this PR much. 


https://github.com/rome/tools/blob/refactor/js-syntax-comments/crates/rome_js_formatter/src/jsx/tag/element.rs

MichaReiser:
> Could you please explain more? What is the difference between Prettier `ifBreak` and Rome `if_group_breaks` + `if_group_fits_on_line` inside fill? :)

The difference isn't around `if_group_fits_on_line` and `if_group_breaks` but the around the fill. I should have added the `fill` to prettier's example too

```
// Prettier
fill([
"Text",
ifBreak(['{" "}', softline], " "),
...

// Rome 
fill([
  "Text",
  [],
  [
    if_group_breaks(["{" "}", soft_line_break]),
    if_group_fits_on_line([" "])
  ],
[],
...
```


Fill, in Rome and prettier, is a sequence of `item`, `separator`, `item`, `separator`. As you can see, Prettier puts the `softline` into the `separator` "slot" whereas Rome adds an empty `separator` and writes the `ifBreak` into an `item` slot. 

The printing of `item` and `separator` slots differ

https://github.com/rome/tools/blob/a124d3654f297830e3c4aa8ce1e594ad078b2a49/crates/rome_formatter/src/printer/mod.rs#L360-L482

The logic is as follow:

Print the first item in flat if it fits on the line, otherwise print it in expanded mode. For all remaining separator & item pairs do:

* Test if `separator` and `item` fit on the same line (as the preceding item): If so, print both in flat
* Otherwise: 
  * Print the separator in expanded mode
  * Test again if the `item` now fits on the line. If so, print in flat mode, otherwise print in expanded mode

That's why we should move the softline break into the separator slot.

MichaReiser (approved):
This is looking good to me. I added a few comments that would be good to address before merging. 

I further recommend rebasing on main to get the prettier snapshot tests.

crates/rome_js_formatter/src/jsx/lists/child_list.rs
```diff
@@ -152,16 +153,21 @@ impl FormatRule<JsxChildList> for FormatJsxChildList {
...
                             } else {
                                 Some(LineMode::Soft)
                             }
                         }
 
                         // Add a hard line break if what comes after the element is not a text or is all whitespace
-                        Some(_) => Some(LineMode::Hard),
+                        Some(JsxChild::NonText(_)) => Some(LineMode::Hard),
 
+                        Some(JsxChild::Newline | JsxChild::Whitespace | JsxChild::EmptyLine) => {
```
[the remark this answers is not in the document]
denbezrukov: [Prettier](https://github.com/prettier/prettier/blob/b0d9387b95cdd4e9d50f5999d3be53b0b5d03a97/src/language-js/print/jsx.js#L144-L180) removes sequences:

- `hardline` `hardline`
- `softline` `hardline`
- `hardline` `softline`
- `softline` `jsxWhitespace`
- `hardline` `jsxWhitespace`
- `jsxWhitespace` `softline`
- `jsxWhitespace` `hardline`
- `jsxWhitespace` `jsxWhitespace`


`JsxChild::Newline` adds hardline, `JsxChild::EmptyLine` adds new line, `JsxChild::Whitespace` is `jsxWhitespace`. If next element after `JsxChild::Word` is `JsxChild::Newline | JsxChild::EmptyLine | JsxChild::Whitespace` we don't need to specify `line_mode`.

crates/rome_js_formatter/src/utils/jsx.rs
```diff
@@ -245,23 +245,61 @@ where
...
+                builder.entry(JsxChild::NonText(child));
+            }
+        }
+    }
+
+    Ok(builder.finish())
+}
+
+#[derive(Debug, Default)]
+struct JsxSplitChildrenBuilder {
```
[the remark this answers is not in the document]
denbezrukov: it is necessary to solve the problem of the sequence of elements. Because we don't know which element will be next to decide whether to keep it or remove it. On push we put an element in the pending state and wait a next element. 
It might be better to solve it by mutating the last element of buffer.

crates/rome_js_formatter/src/utils/jsx.rs
denbezrukov: Yes, I'll rollback it.

MichaReiser (approved):
Looking good. Would you mind adding the prettier compatibility metrics to the test plan. 

And there seems to be an issue with the codegen CI job.

denbezrukov:
After https://github.com/rome/tools/commit/10226631e066e2310977e6811aca96170bb641ca the bug started reproducing again, need to figure out (:

MichaReiser:
> After [1022663](https://github.com/rome/tools/commit/10226631e066e2310977e6811aca96170bb641ca) the bug started reproducing again, need to figure out (:

Hmm that's odd. You aren't introspecting any IR, aren't you? Let me know if you need any help.

denbezrukov:
> Hmm that's odd. You aren't introspecting any IR, aren't you? Let me know if you need any help.

Hm, previous version of `print_fill` had a condition:
```
 let current_and_next_fit = current_fits 
     && fits_on_line( 
         [separator, next_item], 
         args.with_print_mode(PrintMode::Flat), 
         &empty_rest, 
         self, 
     ); 
```
which tested a `separator` and a `next_item`. If elements couldn't be printed on one line the separator would be printed in an expanded mode. 

I'm try to find the same condition in current version. But it seems that now `print_fill_entries` tests only a single element.

MichaReiser:
> > Hmm that's odd. You aren't introspecting any IR, aren't you? Let me know if you need any help.
> 
> Hm, previous version of `print_fill` had a condition:
> 
> ```
>  let current_and_next_fit = current_fits 
>      && fits_on_line( 
>          [separator, next_item], 
>          args.with_print_mode(PrintMode::Flat), 
>          &empty_rest, 
>          self, 
>      ); 
> ```
> 
> which tested a `separator` and a `next_item`. If elements couldn't be printed on one line the separator would be printed in an expanded mode.
> 
> I'm try to find the same condition in current version. But it seems that now `print_fill_entries` tests only a single element.

The corresponding code now is covered by

https://github.com/rome/tools/blob/f3079b4eb5c4b70a0f63915ef0efa448bfb39a75/crates/rome_formatter/src/printer/mod.rs#L459-L462

The `SeparatorItemPairPredicate` takes the elements from the queue until it:
* either reaches the end of the fill (separator only, should be rare)
* it processed the separator and the next item

MichaReiser:
I checked out your branch, generated the IR, and then rebased it on main and again, looked at the IR for:

```
before_break1 =
  <span>
    <span barbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbar /> foo
  </span>
```

**Not rebased**

```
fill([
  group([
    "<",
    "span",
    line_suffix_boundary,
    indent([
      soft_line_break_or_space,
      "barbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbar"
    ]),
    line_suffix_boundary,
    soft_line_break_or_space,
    "/",
    ">"
  ]),
  [
    if_group_breaks(["{" "}", soft_line_break]),
    if_group_fits_on_line([" "])
  ],
  [],
  "foo",
  []
])
```

**Rebased**
```
fill([
    <interned 1> [
      [
        group([
          "<span",
          line_suffix_boundary,
          indent([
            soft_line_break_or_space,
            "barbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbar"
          ]),
          line_suffix_boundary,
          soft_line_break_or_space,
          "/>"
        ])
      ],
      [],
      [
        if_group_breaks(["{" "}", soft_line_break]),
        if_group_fits_on_line([" "])
      ],
      [],
      ["foo"],
      []
    ]
  ])
```

Notice how the 

```
[
  if_group_breaks(["{" "}", soft_line_break]),
  if_group_fits_on_line([" "])
],
```

Is the second child of `fill` for the not rebased version but becomes the third entry after rebasing on main.

denbezrukov:
> I'm not entirely sure what's the best fix. The easy fix is to simply add a method to the multiline builder that let's you write a single entry without caring that item/separator are balanced. I don't mind that overly much but it eases that item/separator position can be "accidentally" flipped.

Yes, you're right! Thank you so much. I've just tried your suggestion and it worked.  I'm going to push it.
What is another way to fix it? Try to take next element in `SeparatorItemPairPredicate` if current one is empty? Is it possible/better?

MichaReiser:
> Yes, you're right! Thank you so much. I've just tried your suggestion and it worked. I'm going to push it.
> What is another way to fix it? Try to take next element in `SeparatorItemPairPredicate` if current one is empty? Is it possible/better?

I don't think this is something that should be fixed inside of the printer because the printer can't know what entry is intended to be the item or separator. Correctly writing the *pairs* is the responsibility of the formatting code.

denbezrukov:
This PR:
Average compatibility: 88.44
Compatible lines: 89.60

Main:
Average compatibility: 88.42
Compatible lines: 89.30

crates/rome_js_formatter/src/jsx/lists/child_list.rs
```diff
@@ -476,19 +493,30 @@ impl MultilineBuilder {
     }
 
     /// Formats an element that does not require a separator
+    /// It is safe to omit the separator because at the call side we must guarantee that we have reached the end of the iterator
+    /// or the next element is a space/newline that should be written into the separator "slot".
     fn write_with_empty_separator(
```
MichaReiser, line 498: I think we should update the name now that the behaviour has changed. To this point, the method ensured that it always writes an empty separator which is why it was called `with_empty_separator`. 

However, the method now intentionally omits the separator, that's why I would call it `write_without_separator`.

crates/rome_js_formatter/src/jsx/lists/child_list.rs
```diff
@@ -476,19 +493,30 @@ impl MultilineBuilder {
     }
 
     /// Formats an element that does not require a separator
+    /// It is safe to omit the separator because at the call side we must guarantee that we have reached the end of the iterator
+    /// or the next element is a space/newline that should be written into the separator "slot".
```
MichaReiser, line 497: I would expect this comment at the call side rather than here on the method because we don't know if it is safe to do so at the method level.

denbezrukov: fixed

denbezrukov:
@MichaReiser 
I was thinking about how we can keep only one method with separator.

- https://github.com/denbezrukov/tools/blob/3880b2d1e39c128448b847535c3a9185697bcd65/crates/rome_js_formatter/src/jsx/lists/child_list.rs#L111-L115
- https://github.com/denbezrukov/tools/blob/3880b2d1e39c128448b847535c3a9185697bcd65/crates/rome_js_formatter/src/jsx/lists/child_list.rs#L204-L208

Here we can call next on the iterator for `JsxChild::Newline `, `JsxChild::Whitespace`,  `JsxChild::EmptyLine`.
For `None` we can return empty `format_with`, because `None` means that we reach the end of the iteration.

- https://github.com/denbezrukov/tools/blob/3880b2d1e39c128448b847535c3a9185697bcd65/crates/rome_js_formatter/src/jsx/lists/child_list.rs#L166-L177
I'm not sure that we can remove `write_without_separator` here. We can do it only if `JsxChild::Newline` and `JsxChild::EmptyLine` always go after `JsxChild::Word`, `JsxChild::NonText`.

crates/rome_js_formatter/src/jsx/lists/child_list.rs
```diff
@@ -139,7 +144,7 @@ impl FormatJsxChildList {
                     let is_trailing_or_only_whitespace = children_iter.peek().is_none();
 
                     if is_trailing_or_only_whitespace || is_after_line_break {
-                        multiline.write_with_empty_separator(&JsxRawSpace, f);
+                        multiline.write_without_separator(&JsxRawSpace, f);
```
MichaReiser, line 147: Nit: Here, we're writing the separator and we obviously don't want to write a separator with a separator :D 

That's why I think it would be good to add a `multiline.write_separator(...)` method that has the same behaviour as `write_without_separator` but only exists to document that we differentiate between items that normally should be written with a separator and separators that... are separators and don't need a separator.

MichaReiser (approved):
Thank you so much for working on this. I have a small recommendation to introduce a new method on the builder to make it more clear that it sometimes writes an item without a separator (requires documentation), a separator (fine to not write a second separator), or an item with a separator.

@ematipico are you OK merging this change during my PTO in case @denbezrukov updates the PR after today?

crates/rome_js_formatter/src/jsx/lists/child_list.rs
denbezrukov: Done (:

ematipico:
> @ematipico are you OK merging this change during my PTO in case @denbezrukov updates the PR after today?

Sure, no problem 😃

denbezrukov:
There is another problem =(

Let's say that we have:
```
 (
  <div>
    <br />
    texttexttexttexttexttexttexttexttexxxxxxxxxxttextxxxtexttext{typeeeggggee}{" "}
  </div>
);
```
Rome output in https://github.com/rome/tools/pull/3251: 
```
<div>
	<br />
	texttexttexttexttexttexttexttexttexxxxxxxxxxttextxxxtexttext{typeeeggggee}{" "}
</div>;

```

Prettier output: 

```
<div>
	<br />
	texttexttexttexttexttexttexttexttexxxxxxxxxxttextxxxtexttext{
		typeeeggggee
	}{" "}
</div>;
```


It has the Rome IR in https://github.com/rome/tools/pull/3251:
```
[
  <interned 0> [group(["<div", line_suffix_boundary, ">"])],
  indent([
    hard_line_break,
    fill([
      <interned 1> [
        ["<br", line_suffix_boundary, " />"],
        [hard_line_break],
        ["texttexttexttexttexttexttexttexttexxxxxxxxxxttextxxxtexttext"],
        [soft_line_break],
        [
          group([
            "{",
            indent([soft_line_break, "typeeeggggee"]),
            soft_line_break,
            line_suffix_boundary,
            "}"
          ])
        ],
        ["{" "}"]
      ]
    ])
  ]),
  hard_line_break,
  "</div",
  line_suffix_boundary,
  ">;",
  hard_line_break
]
```
which is close to the Prettier one:
```
[
  group([
    group(["<", "div", indent([]), ">"]),
    indent([
      hardline,
      fill([
        ["<", "br", " />"],
        hardline,
        "texttexttexttexttexttexttexttexttexxxxxxxxxxttextxxxtexttext",
        softline,
        group([
          "{",
          indent([softline, "typeeeggggee"]),
          softline,
          lineSuffixBoundary,
          "}",
        ]),
        '{" "}',
      ]),
    ]),
    hardline,
    "</",
    "div",
    ">",
  ]),
  ";",
  hardline,
]
```

It matched Prettier output before https://github.com/rome/tools/pull/3307 because when the printer checks 
```
group([
  "{",
  indent([softline, "typeeeggggee"]),
  softline,
  lineSuffixBoundary,
  "}",
]),
```

https://github.com/rome/tools/blob/1fd9b7b5b08e79be36f136f0a94ce1cffbb17777/crates/rome_formatter/src/printer/mod.rs#L198-L233 it went to the second branch and used  the`AllPredicate` to check `fits_on_line` which included the `{" "}` element and used expand mode. After https://github.com/rome/tools/pull/3307 it doesn't include `{" "}` to check `fits_on_line` for the group.

MichaReiser:
Yeah... I looked at it today and haven't been able to come up with a good solution. 

The reason why this works for prettier is that they schedule the item and separator with the corresponding modes and then simply continue printing. That means if the item fits but the separator doesn't, then the print queue looks like this:

```
[ item, mode: flat]
[separator: mode expand]
... other filling entries
```

That means that if the item contains a `group` or `best fitting` and then measures if it fits, it ends at the `separator` that now gets printed in expanded mode (making any soft line break a hard line break). The expanding of the separator guarantees that any inner measuring "terminates". 

I don't think this is entirely ideal because it requires that any `group` or `best_fitting` inside a fill item has to re-measure and the same can be mimicked by:

* Measuring if the `item` and separator fit. 
  * Yes -> print the item in flat mode
  * No: Measure if the `item` with the expanded separator fits
  * Yes -> print the item in flat mode
  * No -> print the item and separator in expanded mode. -> continue to the next item
* Measure if the `separator` and the next item fit
  * Yes -> Print the separator in flat mode -> continue at the top
  * No -> Print the separator in expanded mode

How's this different from what we do today. Today, our implementation measures if the separator and next item fit, but never if the item + next separator do. This is important for the issue we're seeing now because we must expand the item  if neither the flat nor expanded separator fit (there's no real alternative).  

The challenge I'm facing is how to measure if the item + separator in expanded mode fit. The approach I have in mind is to create a `FitsOnLine` struct that keeps the internal fits queue, fits stack, and fits state across multiple `fits_on_line` calls. But I haven't had the time to entirely investigate how to implement it in detail

MichaReiser:
The code for this could be similar to this:

```rust
while (not at end) {
	let mut fits_on_line = FitsOnLine::new(queue, stack, self); 
	
	let item_fits = fits_on_line.fits_fill_entry(PrintMode::Flat); 
	
	// Not even the item fits, print item and separator in expanded mode.
	if !item_fits {
	    drop(fits_on_line);
		self.print_entry(queue, stack , PrintMode::Expanded),
	    self.print_entry(queue, stack, PrintMode:Expanded),
	    continue;
	}

	let separator_fits = fits_on_line.fits_fill_entry(PrintMode::Flat);

	// The item fits, but now the separator doesn't...
    if !separator_fits {
       drop(fits_on_line);
       
		// Now test if the separator fits in expanded mode. 
       let mut fits_on_line = FitsOnLine::new(queue, stack, self); 
	   // skip item, we know it fits
       queue.skip_content(TagKind::Entry); 
	   let separator_fits = fits_on_line.fits_fill_entry(PrintMode::Expanded);
	   drop(fits_on_line);      

		// If the separator fits, then print the item in flat mode, but expand the separator
       if separator_fits {
		self.measured_fits_on_line = true;
        self.print_entry(queue, stack, PrintMode::Flat);
		self.measured_fits_on_line = false;
        self.print_entry(queue, stack, PrintMode::Expanded);
       } else {
		// If not, then print both in expanded mode (the best we can do in this situation).
		self.print_entry(queue, stack , PrintMode::Expanded),
	    self.print_entry(queue, stack, PrintMode:Expanded),
       }
 		continue;
    }

	// we know here that the item + separator fit. Now the question is if the next item fits as well.

	// Test if the next item fits too
	let next_fits = fits_on_line.fits_fill_entry(PrintMode::Flat);

	drop(fits_on_line);

	// Print the item in flat mode because we know the flat or expanded separator always fit.
	self.measured_fits_on_line = true;
	self.print_entry(queue, stack, PrintMode::Flat);
	
	// Print the separator in flat or expanded mode, depending if the next item also fits on the line
	self.measured_fits_on_line = next_fits;
	self.print_entry(queue, stack, if next_fits { PrintMode::Flat } else { PrintMode::Expanded })
	self.measured_fits_on_line = false;

	// Now continue with the next item. We may be clever here and avoid recomputing if the item fits but that might be difficult to accomplish
}

self.measured_fits_on_line = false;
```

The alternative would be to change `Fill` to a `FormatElement` instead of a `Tag` that stores all its items in a `Box<[]>`

denbezrukov:
> 
> * Measuring if the `item` and separator fit.
>   
>   * Yes -> print the item in flat mode
>   * No: Measure if the `item` with the expanded separator fits
>   * Yes -> print the item in flat mode
>   * No -> print the item and separator in expanded mode. -> continue to the next item
> * Measure if the `separator` and the next item fit
>   
>   * Yes -> Print the separator in flat mode -> continue at the top
>   * No -> Print the separator in expanded mode

I'm going to try to implement it in another PR.

MichaReiser:
@denbezrukov How do you feel about updating the failing snapshot test? This would allow us to merge this PR and address the printer issue in a follow up PR.

denbezrukov:
> @denbezrukov How do you feel about updating the failing snapshot test? This would allow us to merge this PR and address the printer issue in a follow up PR.

Done.

PR #3252, opened 2022-09-19
feat(rome_js_analyze): `noRenderReturnValue`

### LINKED ISSUE #3248: `noRenderReturnValue`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noRenderReturnValue.md

ematipico:
## Summary

Closes #3248 

## Test Plan

Added some test cases


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_render_return_value.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+        None
+    }
+
+    fn diagnostic(ctx: &RuleContext<Self>, _state: &Self::State) -> Option<RuleDiagnostic> {
+        let node = ctx.query();
+        Some(RuleDiagnostic::new(
+            node.syntax().text_trimmed_range(),
+            markup! {
+                "Do not depend on the return value from "<Emphasis>"ReactDOM.render()"</Emphasis>"."
```
MichaReiser, line 71: This message is unclear to me:

* what's the meaning of *depending*: Am I not allowed to read it? Or what does depend mean
* Why should I not depend on the value of `React.render` and what should I do instead

Is there maybe some external resource (react documentation) where we can point users to?

ematipico: I extended the documentation. That should answer your questions?

ematipico: I reworded the diagnostic and added shorter version of the documentation with a link

leops (approved):

PR #3255, opened 2022-09-19
test(rome_js_formatter): Add prettier's JSX tests

MichaReiser:
## Summary

This PR extracts the JSX tests from prettier and updates existing tests.

There are a few instability issues, most related to comments. I'll look into those as part of the comments refactoring

## Test Plan

Verified that the new snapshots run as part of `cargo test`

**Average compatibility**: 86.15 -> 86.16
**Compatible lines**: 84.93 -> 84.99


MichaReiser:
I'll handle the comments issue as part of the #3227

ematipico:
What's the compatibility coverage now?

MichaReiser:
> What's the compatibility coverage now?

I added them to the test plan. They aren't changing much (and I prefer to fix any regressions as part of dedicated PRs)

crates/rome_js_formatter/tests/specs/prettier/typescript/typeof-this/decorators.ts
```diff
@@ -0,0 +1,6 @@
+// https://github.com/typescript-eslint/typescript-eslint/pull/4382
+function decorator() {}
```
ematipico, line 2: I don't remember what we do with skipped/unsupported syntax. Do we include them in the test suite?

crates/rome_js_formatter/tests/specs/prettier/jsx/template/styled-components.js
```diff
@@ -0,0 +1,18 @@
+<style jsx>{`
```
ematipico, line 1: What should we do with this test? It would tarnish our coverage

ematipico (approved):

crates/rome_js_formatter/tests/specs/prettier/typescript/typeof-this/decorators.ts
MichaReiser: They are included but "excluded" in the report by testing on the test file name. We don't ignore decorator tests for now

crates/rome_js_formatter/tests/specs/prettier/jsx/template/styled-components.js
MichaReiser: We keep them for now. We only excluded tests where the parser fails to parse the syntax (because it's unlikely that the formatter can then produce any reasonable result)

Issue #3254, opened 2022-09-19
`noUselessFragments`


ematipico:
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noUselessFragment.md

PR #3256, opened 2022-09-20
feat(rome_js_analyze): `noUselessFragment`

ematipico:
## Summary

It implements #3254 

The old rule had also a suggested fix, which I would like to implement in another PR. Having this PR with the correct diagnostics, would help me later to avoid regressions.

Also, I fixed some false positives in the `is_create_react_element` function. There was some logic missing.

It seems now that there are some recurrent patterns, I will refactor them in another PR.

## Test Plan

I added new test cases to cover the new rule and new test cases to cover the false positive I fixed in this PR.


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragment.rs
```diff
@@ -0,0 +1,181 @@
...
+        let node = ctx.query();
+        let model = ctx.model();
+        match node {
+            NoUselessFragmentQuery::JsxFragment(fragment) => {
+                let parent = node.syntax().parent().and_then(|parent| {
+                    if let Some(parent) = JsxTagExpression::cast_ref(&parent) {
+                        parent.syntax().parent()
+                    } else {
+                        Some(parent)
+                    }
```
MichaReiser, line 62: Nit
```suggestion
                    match JsxTagExpression::try_cast(parent) {
                        Ok(parent) => parent.syntax().parent(),
                        Err(parent) => parent
                    } 
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragment.rs
```diff
@@ -0,0 +1,181 @@
...
+                    parent_kind,
+                    Some(
+                        JsSyntaxKind::JS_RETURN_STATEMENT
+                            | JsSyntaxKind::JS_INITIALIZER_CLAUSE
+                            | JsSyntaxKind::JS_CONDITIONAL_EXPRESSION
+                            | JsSyntaxKind::JS_ARROW_FUNCTION_EXPRESSION
+                            | JsSyntaxKind::JS_FUNCTION_EXPRESSION
+                            | JsSyntaxKind::JS_FUNCTION_DECLARATION
+                    )
+                );
```
MichaReiser, line 76: We should be able to move this block into the block where you extract the parent because it is never possible that the parent is any of this list if the "direct parent" isn't `JsxTagExpression`. 

```rust
let matches_allowed_parents = if let Some(expression) = node.parent::<JsxTagExpression>() {
  matches!(expression.syntax().parent().map(|p| p.kind()), .... )
} else {
  false 
}
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragment.rs
```diff
@@ -0,0 +1,181 @@
...
+                let child_list = fragment.children();
+                if child_list.len() <= 1 && !matches_allowed_parents {
+                    return if child_list.is_empty() {
+                        Some(NoUselessFragmentState::Empty)
+                    } else {
+                        // SAFETY: it has at least one item
+                        let attribute = child_list.first().unwrap();
+                        Some(NoUselessFragmentState::Attribute(attribute))
+                    };
+                }
```
MichaReiser, line 87: Nit: You an avoid the `unwrap` by doing

```rust
match child_list.first() {
  Some(first) if child_list.len() == 1 => {
  Some(NoUselessFragmentState::Attribute(first))
  },
  None => Some(NoUselessFragmentState::Empty),
  _ => None
}
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragment.rs
```diff
@@ -0,0 +1,181 @@
...
+                    )
+                );
+
+                let child_list = fragment.children();
+                if child_list.len() <= 1 && !matches_allowed_parents {
+                    return if child_list.is_empty() {
+                        Some(NoUselessFragmentState::Empty)
+                    } else {
+                        // SAFETY: it has at least one item
+                        let attribute = child_list.first().unwrap();
```
MichaReiser, line 84: What's the meaning of `attribute` in this context? If I think of attributes in the jsx context than I think of `<div attribute>`

ematipico: Yeah that was some old naming, sorry. It should be "child", not attribute

crates/rome_js_analyze/tests/specs/nursery/noUselessFragment/withChildren.jsx.snap
```diff
@@ -0,0 +1,46 @@
...
+    <>foo</>
+    <React.Fragment>foo</React.Fragment>
+    <Fragment>foo</Fragment>
+</>
+
+```
+
+# Diagnostics
+```
+warning[nursery/noUselessFragment]: Avoid using unnecessary Fragment.
```
MichaReiser, line 17: should *Fragment* be plural? Avoid using unnecessary Fragments?

crates/rome_js_analyze/tests/specs/nursery/noUselessFragment/withChildren.jsx.snap
```diff
@@ -0,0 +1,46 @@
...
+  ┌─ withChildren.jsx:2:5
+  │
+2 │     <>foo</>
```
MichaReiser, line 20: Is `Fragment` jargon? Is it OK for us to assume that people know that `<>` is called a fragment or should we change the messaging

crates/rome_js_analyze/tests/specs/nursery/noDangerouslySetInnerHtml/createElementBindingInvalid.js.snap
```diff
@@ -18,7 +18,7 @@ createElement('div', {
 # Diagnostics
 ```
 warning[nursery/noDangerouslySetInnerHtml]: Avoid passing content using the dangerouslySetInnerHTML prop.
-  ┌─ createElementBinding.js:4:5
+  ┌─ createElementBindingInvalid.js:4:5
```
MichaReiser, line 21: Would you mind adding an example with a comment (and yes, this is valid)

```javascript
</* test */>empty</>;
<>empty</* comment */ />;
<>empty</ /* comment */>
```

I would argue that the rule should probably not trigger for these cases.

MichaReiser (approved):

crates/rome_js_analyze/tests/specs/nursery/noUselessFragment/withChildren.jsx.snap
ematipico: Yeah it makes sense. I also found out that we don't have a guideline of when using singular and plurals inside a rule.

crates/rome_js_analyze/tests/specs/nursery/noDangerouslySetInnerHtml/createElementBindingInvalid.js.snap
ematipico: The rule as it is now triggers for these cases, it seems. Would you mind give us a reason, please? It seems an edge case, so I can better document the code and rule itself

crates/rome_js_analyze/tests/specs/nursery/noUselessFragment/withChildren.jsx.snap
ematipico: We will have a rule called [`useFragmentSyntax`](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/useFragmentSyntax.md). Maybe that's where we can give more explanation?

crates/rome_js_analyze/tests/specs/nursery/noDangerouslySetInnerHtml/createElementBindingInvalid.js.snap
MichaReiser: My reasoning is that these fragments are entirely useless because it is otherwise not possible to add the comments (except using an expression). 

However, we could reason that we should flag these nonetheless but not provide an autofix because it is unclear where to place the comment?

ematipico: Yeah this rule is not over yet, there are still few cases that doesn't take into account. And the autofix will arrive with another PR :)

ematipico:
@leops the last commit adds a fix and test case to cover what you suggested. Nice catch!

leops (approved):

PR #3262, opened 2022-09-21
fix(rome_js_analyzer): fix false positive #3261

### LINKED ISSUE #3261: 🐛 false positive for `useCamelCase`
### Environment information

```bash
Latest on main
```

### What happened?

See [playground](https://play.rome.tools/?code=function+Component%28%29+%7B%0A%0A%7D%0A%0Aconst+f+%3D+%3CComponent%3Eff%3C%2FComponent%3E%3B%0A%0Af%28%29&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&enabledNurseryRules=true&typescript=false&jsx=false#ZgB1AG4AYwB0AGkAbwBuACAAQwBvAG0AcABvAG4AZQBuAHQAKAApACAAewAKAAoAfQAKAAoAYwBvAG4AcwB0ACAAZgAgAD0AIAA8AEMAbwBtAHAAbwBuAGUAbgB0AD4AZgBmADwALwBDAG8AbQBwAG8AbgBlAG4AdAA+ADsACgAKAGYAKAApAA==)

The rule emits a false positive diagnostic for components in the same file

```jsx
function Component() {}
const f = <Component>ff</Component>;
f()
```

### Expected result

It should not emit diagnostics

### Code of Conduct

ematipico:
## Summary

This PR fixes #3261 

The fix is simple. It iterates over all the `all_reads` of the binding and checks if at least one is used as a `JsxRefereceIdentifier`.

## Test Plan

Added a new test case


MichaReiser (approved):
Would you mind adding a more descriptive title. It avoids that I have to open the issue when e.g. going through the commits using git.

MichaReiser:
!bench_analyzer

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_camel_case.rs
```diff
@@ -118,8 +118,14 @@ impl Rule for UseCamelCase {
...
+                        });
 
-                        if is_variable || is_parameter || is_function || is_exported_function {
+                        if is_jsx_component {
+                            None
+                        } else if is_variable || is_parameter || is_function || is_exported_function
+                        {
                             let name = binding.name_token().ok()?;
                             check_is_camel(name.text_trimmed())
                         } else {
```
MichaReiser, line 131: It may be beneficial for performance to first check if the name is camel case because this is relatively "cheap" and only then check if it is a JSX component

ematipico:
!bench_analyzer

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_camel_case.rs
```diff
@@ -112,43 +112,60 @@ impl Rule for UseCamelCase {
...
+
+                            if !is_jsx_component {
+                                let is_variable =
+                                    binding.parent::<JsVariableDeclarator>().is_some();
+                                let is_parameter = binding.parent::<JsFormalParameter>().is_some();
+                                let is_function =
+                                    binding.parent::<JsFunctionDeclaration>().is_some();
+                                let is_exported_function = binding
+                                    .parent::<JsFunctionExportDefaultDeclaration>()
+                                    .is_some();
```
MichaReiser, line 131: Checking the parent is probably cheaper than checking if the string is camel case because all it requires is to clone the parent node (no allocation necessary) and check its kind. The compiler is probably even able to only clone the parent once (or optimizes it completely away).

I don't know the details about `model.all_reads` but it gives me the impression that it may be slightly more involved then just dereferencing the parent and checking its kind.

ematipico:
!bench_analyzer

PR #3263, opened 2022-09-22
refactor(rome_diagnostics): refactor existing diagnostics to use the new display

leops:
## Summary

This PR changes the printing logic for "legacy" diagnostics to use the new "v2" display. This requires a few additional changes to the rest of the codebase as:
- The `code` for a `Diagnostic` needs to be valid `Category` now (besides changing a few tests that were using "fake" error codes, this required adding the `lint/` prefix to diagnostics emitted by the lint rules, and changing any logic that may depend on these codes accordingly)
- The `DiagnosticHeader` display component no longer exists, this required rewriting a few errors in the CLI to diagnostics to have them reuse the diagnostic header printing logic as intended

## Test Plan

This change rewrites all the parser, analyzer and CLI snapshots that contain diagnostics, as well as all the lint rule documentation pages on the website


MichaReiser (approved):

crates/rome_cli/src/traversal.rs
```diff
@@ -203,6 +201,62 @@ struct ProcessMessagesOptions<'ctx> {
...
+    fn record(&self, visitor: &mut dyn Visit) -> io::Result<()> {
+        // Skip printing the diff for files over 1Mb (probably a minified file)
+        let max_len = self.old.len().max(self.new.len());
+        if max_len >= 1_000_000 {
+            visitor.record_log(LogCategory::Info, &"[Diff not printed for file over 1Mb]")
+        } else {
+            visitor.record_diff(self.old, self.new)
+        }
+    }
+}
```
ematipico, line 245: Shouldn't we apply the same logic to `CIDiffDiagnostic`? If not, this creates a change in behaviour that we have to document somewhere.

crates/rome_cli/src/traversal.rs
```diff
@@ -250,13 +304,25 @@ fn process_messages(options: ProcessMessagesOptions) -> bool {
...
+                            Severity::Warning => v2::Severity::Warning,
+                            Severity::Error => v2::Severity::Error,
+                            Severity::Bug => v2::Severity::Fatal,
+                        },
+                        category: err.code.parse().unwrap_or_else(|()| {
+                            panic!(
+                                "Error code {:?} is not a registered Diagnostic category",
+                                err.code
+                            )
+                        }),
```
ematipico, line 321: I suppose this should caught by our panic hook right? What's inside `err.code`? And if we encounter this error, how should we, developers, fix it?

crates/rome_analyze/src/lib.rs
```diff
@@ -398,14 +399,14 @@ where
                             let diag = match group_rule {
                             Some((group, rule)) => Diagnostic::warning(
                                 file_id,
-                                "Linter",
+                                "suppressions/invalidRule",
```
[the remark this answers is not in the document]
leops: It seems this was already an issue for the corresponding diagnostic category in Rome JS: https://github.com/rome/tools/blob/6b47011c5f3d5ed783855b821cdbc9cf1d687266/internal/diagnostics/descriptions/suppressions.ts#L47-L54
I'm not sure if this was due to a naming convention that Rome JS had for diagnostic categories, and if `unknown` / `invalid` were meant to carry a special meaning for these

crates/rome_diagnostics/src/diagnostic.rs
```diff
@@ -68,6 +69,13 @@ impl Diagnostic {
         title: impl Display,
         code: Option<String>,
     ) -> Self {
+        let code = code.filter(|code| !code.is_empty());
+        if let Some(code) = &code {
+            if code.parse::<&'static v2::Category>().is_err() {
+                panic!("code {code:?} is not a registered diagnostic category");
```
ematipico, line 75: Would it be possible to leave an hint of how a developer can fix the issue? (I suppose we need to update the macros inside `categories.rs` but an hint would be great for colleagues that don't have context)

crates/rome_diagnostics/src/emit.rs
```diff
@@ -54,114 +59,138 @@ impl Emitter<'_> {
...
+
+        if let Some(location) = self.lookup_location(label.span) {
+            visitor.record_frame(location)?;
+        }
+
+        Ok(())
+    }
+}
+
+fn map_severity(severity: Severity) -> LogCategory {
```
ematipico, line 121: nit `map_severity_to_log_category`

crates/rome_diagnostics/src/emit.rs
```diff
@@ -54,114 +59,138 @@ impl Emitter<'_> {
...
+        PrintDiagnostic(self).fmt(fmt)?;
+        Ok(())
+    }
+}
+
+impl v2::Diagnostic for DiagnosticPrinter<'_> {
+    fn category(&self) -> Option<&Category> {
+        self.d.code.as_deref().map(|code| {
+            code.parse()
+                .unwrap_or_else(|()| panic!("code {code:?} is not a valid diagnostic category"))
```
ematipico, line 150: Same here, we should give a hint about how to fix issue, if possible.

ematipico (approved):

crates/rome_cli/src/traversal.rs
leops: Both `CIDiffDiagnostic` and `FormatDiffDiagnostic` make use of `FormatDiffAdvice` internally, so this diff printing logic is shared between the two diagnostic types

crates/rome_cli/src/traversal.rs
leops: This panic would be triggered if a diagnostic is created with a code / category that's not registered in `rome_diagnostic_categories`. This is only a temporary measure though, I removed this panic in #3268 once all the diagnostics have been migrated to use statically-checked categories instead

PR #3264, opened 2022-09-22
feat(rome_js_analyze): noUselessFragment

ematipico:
## Summary

Implements #3254 

This PR implements a **partial fix** to the rule `noUselessFragments`. With partial I mean that it handles code actions only in cases where the "wrong" fragments are inside a list. 

I plan to add the other part of the code action in another PR, particularly for case where the "wrong" fragments are not inside list. 

The code action implements an heuristic to keep the comments when deleting a fragment.

## Test Plan

Updated the existing test cases and manually checked that are valid


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -29,6 +37,13 @@ declare_rule! {
     /// ```
     ///
     /// ```jsx,expect_diagnostic
+    /// <>
+    ///     <>foo</>
+    ///     <SomeComponent />
+    /// </>
```
ematipico, line 43: I added this case so a code fix is shown in the website

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -100,25 +223,132 @@ impl Rule for NoUselessFragments {
...
+                            let mut new_text = String::new();
+                            opening_leading_comments
+                                .into_iter()
+                                .for_each(|piece| new_text.push_str(piece.text()));
+                            new_text.push_str(text.value_token().ok()?.text_trimmed());
+                            closing_leading_comments
+                                .into_iter()
+                                .for_each(|piece| new_text.push_str(piece.text()));
+
+                            let new_jsx_text = jsx_text(jsx_ident(&new_text));
```
ematipico, line 323: Is there a smart way to **not** allocate a `String` and create a `&str` from different `&str`. I really don't like allocating a `String` just do concatenate different strings...

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -100,25 +223,132 @@ impl Rule for NoUselessFragments {
...
+                            // the old comments.
+                            // Leading comments are prepended to the existing content, and the trailing comments
+                            // are appended to the to the existing content
+                            //
+                            // Example:
+                            // </*comment/*> content </ /*comment*/>
+                            //
+                            // Will become
+                            //
+                            // /*comment*/ content /*comment*/
```
MichaReiser, line 313: I'm struggling understanding this comment. What I understand is that this moves the comments into the JSXText? 

If so, I don't think that's a good idea because that has an observable difference to the user. The comments now become part of the rendered output. 

If you want to retain comments, than you have to create a `JsxExpressionChild` that contains the comments.

I also recommend you adding test plans for line comments as these are tricky to get right

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -48,6 +63,114 @@ declare_node_union! {
...
+            }
+        }
+    }
+
+    fn opening_comments(&self) -> Option<Vec<SyntaxTriviaPiece<JsLanguage>>> {
+        let mut comments = Vec::new();
+        let l_angle = match self {
+            NoUselessFragmentsQuery::JsxFragment(fragment) => {
+                let opening_fragment = fragment.opening_fragment().ok()?;
+                opening_fragment.l_angle_token().ok()?
```
MichaReiser, line 102: You have have to retain the leading comments of `>` 

For the `</>` it is even more complicated because you need the comments of all three tokens

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -48,6 +63,114 @@ declare_node_union! {
...
+                .leading_trivia()
+                .pieces()
+                .filter(|piece| piece.is_comments())
+                .chain(
+                    l_angle
+                        .trailing_trivia()
+                        .pieces()
+                        .filter(|piece| piece.is_comments()),
+                ),
+        );
```
MichaReiser, line 121: You may want to keep some whitespace between comments or are you re-inserting the whitespace after. E.g. line comments require a line break at the end

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -48,6 +63,114 @@ declare_node_union! {
...
+    fn remove_node_from_list(&self, mutation: &mut BatchMutation<JsLanguage>) {
+        match self {
+            NoUselessFragmentsQuery::JsxFragment(fragment) => {
+                let old_node = JsxAnyChild::JsxFragment(fragment.clone());
+                mutation.remove_jsx_child_element(old_node);
+            }
+            NoUselessFragmentsQuery::JsxElement(element) => {
+                let old_node = JsxAnyChild::JsxElement(element.clone());
+                mutation.remove_jsx_child_element(old_node);
+            }
```
MichaReiser, line 93: My understanding is that this rule only applies to fragments. Thus, it should never happen that you remove an element? Or what are the situations where the code fix removes an element ?

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
ematipico: So you're suggesting to have a result like this instead? 

```jsx
{/*comment*/} content {/*comment*/}
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
ematipico: Situations where we have fragments expressed as jsx tags: `<Fragment></Fragment>`. The snapshots have some cases that should help understand

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
ematipico: I haven't covered these cases yet. Considering that we need cover them, I'd prefer to so in another PR. The logic will become even more complicated

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
MichaReiser: Yes, or remove the comments but adding the comments to the `text` changes the semantics of the program and isn't something we should do.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
MichaReiser: I would recommend to either:
1. not flag fragments with comments, 
2. drop comments
3. or handle comments correctly

I think it's fine to ship a fix first that does 1 or 2 and than follow up with a PR that adds 3 if necessary but I would refrain from shipping a code fix that doesn't handle comments correctly and may produce invalid syntax because of it.

ematipico: I went for dropping the comments for this PR

crates/rome_js_analyze/src/utils/batch.rs
```diff
@@ -168,6 +176,42 @@ impl JsBatchMutation for BatchMutation<JsLanguage> {
             })
             .unwrap_or(false)
     }
+
+    fn replace_jsx_child_element(
```
leops, line 180: How is this different from calling `mutation.replace_node(old_child, new_child)` directly ?

ematipico: Actually, that's not needed. Thank you for the review. This is me being newbie with the mutations :)

crates/rome_js_analyze/src/utils/batch.rs
```diff
@@ -168,6 +172,22 @@ impl JsBatchMutation for BatchMutation<JsLanguage> {
             })
             .unwrap_or(false)
     }
+
+    fn remove_jsx_child_element(&mut self, node_to_remove: JsxAnyChild) -> bool {
```
leops, line 176: Isn't this also doing the same thing as calling `self.remove_node(node_to_remove)` directly .

MichaReiser (approved):
> I plan to add the other part of the code action in another PR, particularly for case where the "wrong" fragments are not inside list.

I found this sentence hard to understand and misleading. Would you mind clarifying in the summary that this PR only adds support for removing fragments if they are the "only" child of a parent node (and doesn't support tag expressions)

Edit: May also be worth to explain this in a comment as part of the rule

PR #3268, opened 2022-09-22
refactor(rome_diagnostics): refactor diagnostics to use static categories

leops:
## Summary

This PR changes the type of diagnostics code / categories from plain strings to the new `Category` type. I tried to make use of statically checked categories (using the `category!` macro) as much as possible, and one particular area of interest for these are the analyzer rules:

Since some rules never emit any diagnostic (mainly assist rules such as `flipBinExp` or `inlineVariable`) this check needs to happen only for rules that actually end up creating a diagnostic. The solution I came up with is to have the `declare_rule!` macro emit a new `rule_category!` macro in the module context that expands to the category for the rule if it is correctly registered (or a compilation error if it's not). This macro is used by lint rules to create diagnostics with the correct category, and most importantly assist rules never use the macro and consequently never cause the compile error to be emitted in the expanded code.

One downside of this approach is that the analyzer infrastructure is no longer responsible for injecting the category corresponding to a rule into the diagnostics it emits, and individual rules could possible emit diagnostics with a different category, but they would still be restricted to existing categories, and this may in fact turn out to be useful to implement syntax rules that would emit diagnostics with the "parse" error category instead of "lint"

## Test Plan

I changed a number of error categories to re-use the name they had in Rome JS, this has caused some parser and CLI snapshots to change again. For the rest as I mentioned earlier I used static verification as much as possible, so this guarantees that the categories used in the code are valid if it builds correctly.


leops:
Closing and reopening as #3270 because I created the PR from the wrong branch

PR #3269, opened 2022-09-23
feat(rome_js_analyzer): rule `noChildrenProp`

### LINKED ISSUE #3267: `noChildrenProp`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noChildrenProp.md

ematipico:
## Summary

Closes #3267 

This PR also enables other nursery rules in the playground and inside our repository. While doing so, I uncovered a false positive inside `noUselessFragments` rule, which is fixed here.

## Test Plan

Added new test cases for the new rule and the fixed case

### DIFF
--- a/crates/rome_js_analyze/src/semantic_analyzers/nursery/no_children_prop.rs
+++ b/crates/rome_js_analyze/src/semantic_analyzers/nursery/no_children_prop.rs
@@ -1,0 +1,80 @@
+use crate::react::is_react_create_element;
+use crate::semantic_services::Semantic;
+use rome_analyze::context::RuleContext;
+use rome_analyze::{declare_rule, Rule, RuleCategory, RuleDiagnostic};
+use rome_console::markup;
+use rome_js_syntax::{JsCallExpression, JsPropertyObjectMember, JsxAttribute, JsxName};
+use rome_rowan::{declare_node_union, AstNode};
+declare_rule! {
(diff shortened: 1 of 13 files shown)


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_children_prop.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        Some(
+            RuleDiagnostic::new(
+                range,
+                markup! {
+                    "Avoid passing "<Emphasis>"children"</Emphasis>" using a prop"
+                },
+            )
+            .footer_note(markup! {
+                "The canonical way to pass children in React is to use JSX elements or additional arguments to React.createElement."
```
leops, line 91: Nit: this could use the state variant to only mention the correct option for the query node ("JSX elements" for a JSX attribute, and "additional arguments to React.createElement" for a `createElement` call expression)

crates/rome_js_analyze/tests/specs/nursery/noChildrenProp/noChildrenPropValid.jsx
```diff
@@ -0,0 +1,6 @@
+// invalid
```
leops, line 1: This should be `// valid`

PR #3270, opened 2022-09-23
refactor(rome_diagnostics): refactor diagnostics to use static categories

leops:
## Summary

This PR changes the type of diagnostics code / categories from plain strings to the new `Category` type. I tried to make use of statically checked categories (using the `category!` macro) as much as possible, and one particular area of interest for these are the analyzer rules:

Since some rules never emit any diagnostic (mainly assist rules such as `flipBinExp` or `inlineVariable`) this check needs to happen only for rules that actually end up creating a diagnostic. The solution I came up with is to have the `declare_rule!` macro emit a new `rule_category!` macro in the module context that expands to the category for the rule if it is correctly registered (or a compilation error if it's not). This macro is used by lint rules to create diagnostics with the correct category, and most importantly assist rules never use the macro and consequently never cause the compile error to be emitted in the expanded code.

One downside of this approach is that the analyzer infrastructure is no longer responsible for injecting the category corresponding to a rule into the diagnostics it emits, and individual rules could possible emit diagnostics with a different category, but they would still be restricted to existing categories, and this may in fact turn out to be useful to implement syntax rules that would emit diagnostics with the "parse" error category instead of "lint"

## Test Plan

I changed a number of error categories to re-use the name they had in Rome JS, this has caused some parser and CLI snapshots to change again. For the rest as I mentioned earlier I used static verification as much as possible, so this guarantees that the categories used in the code are valid if it builds correctly.


crates/rome_diagnostics_categories/src/categories.rs
```diff
@@ -61,11 +61,13 @@ define_dategories! {
     "internalError/panic",
     "io/skippedFile",
     "io/unhandledFile",
+    "parse",
     "suppressions/unknownGroup",
```
ematipico, line 65: Is `io` beginner friendly? (it's also an acronym right?) Would it make sense to call it `fileSystem`? 

Also (sorry I caught this only now), I think `args` is not beginner friendly. A junior developer might not undestand what it means, it's really technical. To be honest, **I also struggle** to understand what it means, because there's `fileNotFound` later... which something related to the file system. I would suggest an alternative but I would a clarification of the category

crates/rome_diagnostics_categories/src/categories.rs
```diff
@@ -61,11 +61,13 @@ define_dategories! {
...
     "suppressions/unknownRule",
     // Legacy codes
-    "SyntaxError",
     "IO",
     "Lint",
     "Format",
+    // Used in tests
+    "SemanticTests",
```
ematipico, line 72: `semanticTests` so it's in line with the wording of the new categories

ematipico (approved):
The use of `rule_category!()` should be documented somewhere, probably [here](https://github.com/rome/tools/blob/main/crates/rome_analyze/src/rule.rs#L67) and add its link in the contribution document

crates/rome_diagnostics_categories/src/categories.rs
```diff
@@ -61,11 +61,13 @@ define_dategories! {
     "internalError/panic",
     "io/skippedFile",
     "io/unhandledFile",
+    "parse",
```
MichaReiser, line 64: Is the `parse` category shared between all parsers?

MichaReiser (approved):

crates/rome_diagnostics_categories/src/categories.rs
leops: The `parse` category is highly generic and is shared between all syntax error diagnostics in the lexer, parser and potentially future rule in the syntax category in the analyzer

crates/rome_diagnostics_categories/src/categories.rs
leops: I've tried to reword some of these codes, but I/O errors are not necessarily caused by the filesystem as they can also be related to the domain socket or named pipe used by the daemon process to communicate for instance. So I think we still need a generic `internalError/io` category when a low-level error happens.

For the `args/fileNotFound` category it's something I reused from Rome JS along with `flags/invalid`, but I haven't put much thought into it since at the moment those two are only used in examples and tests to re-implement the equivalent diagnostic from the JS version.

PR #3273, opened 2022-09-26
feat(rome_js_formatter): Format Member Chain

MichaReiser:
## Summary

This PR implements member chain formatting the same way as prettier does. 

There are still some differences in combination with call arguments because our call arguments formatting doesn't match Prettier's yet. I plan to tackle this as a separate PR. 

## Test Plan

**Average compatibility**: 88.42 -> 89.48%
**Compatible lines**: 89.3 -> 90.47%


MichaReiser:
!bench_formatter

MichaReiser:
!bench_formatter

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -97,19 +121,29 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+            let mut first_last_breaks = false;
 
-                            in_relevant_range && element.will_break()
-                        }
-                        Err(_) => false,
-                    });
+            for (index, argument) in separated.iter_mut().enumerate() {
+                let breaks = argument.inspect(f)?.will_break();
+
+                any_argument_breaks = any_argument_breaks || breaks;
```
ematipico, line 130:
```suggestion
                any_argument_breaks |= breaks;
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -97,19 +121,29 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+
+                if (should_group_first_argument && index > 0)
+                    || (should_group_last_argument && index < args.len() - 1)
+                {
+                    first_last_breaks = first_last_breaks || breaks;
```
ematipico, line 135:
```suggestion
                    first_last_breaks |= breaks;
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -412,6 +424,80 @@ fn could_group_expression_argument(
     Ok(result)
 }
 
+fn is_commonjs_or_amd_call(
```
ematipico, line 427: Would it be possible to add a doc block with an example of what's a "commonjs" call or an "amd call"?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -412,6 +424,80 @@ fn could_group_expression_argument(
...
+    arguments: &JsCallArguments,
+    call: Option<&JsCallExpression>,
+) -> SyntaxResult<bool> {
+    let call = match call {
+        Some(call) => call,
+        None => return Ok(false),
```
ematipico, line 433: Can't this check be done by the caller? I mean, can't the caller check if `call` actually exists and pass `&JsCallExpression` instead? It find weird that the function checks a "call" but the callee might not exists 😆

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -412,6 +424,80 @@ fn could_group_expression_argument(
...
+                }
+            } else {
+                false
+            }
+        }
+        _ => false,
+    })
+}
+
+fn is_multiline_template_only_args(arguments: &JsCallArguments) -> bool {
```
ematipico, line 488: Can we add a small explanation of what a "multiline template only args" is?

crates/rome_js_formatter/src/utils/member_chain/chain_member.rs
```diff
@@ -1,34 +1,51 @@
...
+    /// `of().test`
+    Start,
+
+    /// Somewhere in the middle.
+    /// `a.b().c()`
+    Middle,
+
+    /// At the end of a call chain (root)
+    /// `a.b.c()`
+    End,
```
ematipico, line 25: Given the example, where exactly are we positioned? in `c()`?

crates/rome_js_formatter/src/utils/member_chain/mod.rs
```diff
@@ -106,339 +106,611 @@ mod chain_member;
...
+                .as_ref()
+                .map_or(false, is_factory);
+
+            has_computed_property || is_factory
+        } else {
+            false
+        }
+    }
+
+    /// It tells if the groups should be break on multiple lines
```
ematipico, line 253:
```suggestion
    /// It tells if the groups should break on multiple lines
```

ematipico (approved):

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: This isn't the same to my understanding because `||` is the logical or whereas `|=` does a bitwise or. The logical or allows to short circuit the right hand side if the left side is already true which isn't the case for bitwise or. 

[https://doc.rust-lang.org/book/appendix-02-operators.html]

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: Same as above. `|=` performs a bitwise or where `||` performance a logical or.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: I added some documentation that link to the amd/commonjs specifications that show the signature.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: I added some documentation and linked to `is_multiline_template_starting_on_same_line` that has an example.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
ematipico: We do the same thing in another part of the code, why here is different? https://github.com/rome/tools/blob/main/crates/rome_cli/src/traversal.rs#L356

MichaReiser: It's important to be different here because it allows the compiler to short circuit to not compute `breaks`. 

I don't know if this is something we want (or don't want) in the link you shared but I would be careful about using bitwise operations (it's most often not what you want).

MichaReiser: Using this example:

```

use ::std::*;

fn some_function(x: usize) -> bool {
    x % 2 == 0
}


pub fn main ()
{


    let mut ord = false;
    
    let mut bitwise_or = false;
    
    for x in 0..10 {
        bitwise_or |= some_function(x);
        ord = ord || some_function(x);
    }
    
    dbg!(ord);
    dbg!(bitwise_or);
}
```

Rust generates very different code for `|=` and `||`

for `|=` 

```asm
        call    example::some_function
        mov     byte ptr [rsp + 159], al
        mov     cl, byte ptr [rsp + 159]
        mov     al, byte ptr [rsp + 199]
        or      al, cl
        and     al, 1
        mov     byte ptr [rsp + 199], al
```

You can see how it unconditionally calls `some_function` and then `or`s the result


for `||`

```asm
        test    byte ptr [rsp + 198], 1
        jne     .LBB20_9
        mov     rdi, qword ptr [rsp + 144]
        call    example::some_function
        mov     byte ptr [rsp + 143], al
        jmp     .LBB20_11
.LBB20_9:
        mov     byte ptr [rsp + 254], 1
.LBB20_10:
        mov     al, byte ptr [rsp + 254]
        and     al, 1
        mov     byte ptr [rsp + 198], al
        jmp     .LBB20_2
.LBB20_11:
        mov     al, byte ptr [rsp + 143]
        and     al, 1
        mov     byte ptr [rsp + 254], al
        jmp     .LBB20_10
```

It first tests if `ord` is true. If so, it jumps to the block `LBB20_9` (not calling `some_function`). It only calls into `some_function` if `ord` is false. 

LLVM may still be able to optimise to the same output BUT using `|` has stricter calling conventions than `||` where it's OK to not compute the right value if the left value is `true`.

[compiler explorer](https://godbolt.org/z/oo1zh44b3)

Whether the use of `|=` is correct in the other places is probably best answered by @leops who wrote the code (in some places it certainly is because it uses bitflags. In some cases it isn't clear to me because left and right are booleans and maybe using `||` would have been more appropriate).

PR #3276, opened 2022-09-26
fix(rome_js_analyzer): add newline if statement has comments in `useBlockStatements`

### LINKED ISSUE #3199: 🐛 Auto-suggestion for `js/useBlockStatements` can break single-line commented `else`
### Environment information

```bash
macOS 12.5.1
VSCode v1.71.0
Rome extension v0.14.1
```

### What happened?

Format the following code, using either the VSCode extension or applying fixes from the commandline:
```ts
if (true) {
  console.log("true");
} else
  console.log("false"); // comment
```

This result in broken code:

```ts
if (true) {
  console.log("true");
} else {
  console.log("false"); // comment }
```

### Expected result

The added `}` is on a separate line, instead of being added to (and swallowed by) the comment.

### Code of Conduct

ematipico:
## Summary

This PR fixes #3199

## Test Plan

Added a new test case

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/correctness/use_block_statements.rs
+++ b/crates/rome_js_analyze/src/analyzers/correctness/use_block_statements.rs
@@ -216,8 +216,16 @@ impl Rule for UseBlockStatements {
 
                     r_curly_token.with_leading_trivia(leading_trivia)
                 } else {
-                    r_curly_token
-                        .with_leading_trivia(iter::once((TriviaPieceKind::Whitespace, " ")))
+                    let has_trailing_comments = stmt.syntax().has_trailing_comments();
+                    // if the node we have to enclose has some leading comments, then we add a new line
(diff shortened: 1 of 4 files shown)


crates/rome_js_analyze/src/analyzers/correctness/use_block_statements.rs
```diff
@@ -216,8 +216,16 @@ impl Rule for UseBlockStatements {
 
                     r_curly_token.with_leading_trivia(leading_trivia)
                 } else {
-                    r_curly_token
-                        .with_leading_trivia(iter::once((TriviaPieceKind::Whitespace, " ")))
+                    let has_trailing_comments = stmt.syntax().has_trailing_comments();
```
leops, line 219: Technically we only need to do this is the trailing trivia has any single-line comments, but I think breaking on every comment kind is fine

leops (approved):

PR #3277, opened 2022-09-26
fix(rome_js_analyzer):  precedence for `useOptionalChain`

### LINKED ISSUE #3257: 🐛 invalid code action of `useOptionalChain`
### Environment information

```bash
linux 
AMD
```

### What happened?

```js
(new foo || {}).bar
```
This code snippet will be fixed as 
```
new foo?.bar
```

which is invalid syntax in typescript https://www.typescriptlang.org/play?#code/HYUw7gBAZg9jD8A6ARgQwE4Cgg

### Expected result

Should not be fixed as 
`new foo?.bar`

### Code of Conduct

ematipico:
## Summary

This PR fixes #3257 

## Test Plan

The PR add two new test cases

### DIFF
--- a/crates/rome_js_syntax/src/lib.rs
+++ b/crates/rome_js_syntax/src/lib.rs
@@ -293,10 +293,12 @@ pub enum OperatorPrecedence {
     Exponential = 15,
     Unary = 16,
     Update = 17,
-    LeftHandSide = 18,
-    Member = 19,
-    Primary = 20,
-    Group = 21,
+    // `new` without arguments list
+    New = 18,
+    LeftHandSide = 19,
+    Member = 20,
+    Primary = 21,
+    Group = 22,
 }
 
 impl OperatorPrecedence {
--- a/crates/rome_js_syntax/src/lib.rs
+++ b/crates/rome_js_syntax/src/lib.rs
@@ -294,7 +294,7 @@ pub enum OperatorPrecedence {
     Unary = 16,
     Update = 17,
     // `new` without arguments list
-    New = 18,
+    NewWithoutArguments = 18,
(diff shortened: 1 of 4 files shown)


leops (commented):
I'm not sure those are really false positive, it looks more like the fix is simply incorrectly removing the parenthesis: `(new foo || {}).bar` should be replaced with `(new foo)?.bar`, but the generated code action emits `new foo?.bar` instead which then gets re-parsed as `new (foo?.bar)` and results in a syntax error

ematipico:
Turns out that the issue was the precedence. If you check the [MDN table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table) the `new` keyword has two different precedence if there or not arguments.

The absence of arguments yields a unique precedence just for this case, which was not tracked in our enum. I updated the code and it now works as @leops suggested.

crates/rome_js_syntax/src/lib.rs
```diff
@@ -293,10 +293,12 @@ pub enum OperatorPrecedence {
     Exponential = 15,
     Unary = 16,
     Update = 17,
-    LeftHandSide = 18,
-    Member = 19,
-    Primary = 20,
-    Group = 21,
+    // `new` without arguments list
+    New = 18,
```
MichaReiser, line 297: The name `new` might be misleading. Someone could assume that this is the default precedence for the majority of `new` expressions which is not the case. That's why I would prefer a more explicit name: `NewWithoutArguments`

MichaReiser (approved):
Would you mind testing if the formatter handles the precedence correctly and, if not, fix the parentheses setting in its own PR?

PR #3280, opened 2022-09-26
refactor(rome_diagnostics): port the display of code frames over from Rome JS

leops:
## Summary

This PR refactors how code frame advices are printed in diagnostics, removing the `codespan` module that previously existed in `rome_console` to instead use a Rust implementation of the code frame display from Rome JS in `rome_diagnostics`.

Removing the `codespan` module means I had to move the `Severity` enum to `rome_diagnostics` (and rewrite all the imports), as well as the `SourceFile` struct that's now internal to the printing logic since the new `SourceCode` struct should be used instead. As part of this I also changed how the `line_index` field in `SourceCode` is typed from a `Vec<TextSize>` / `&[TextSize]` to an opaque `LineIndexBuf` / `&LineIndex` pair of types that encapsulate the logic for retrieving line and column numbers from a byte position in a file.

## Test Plan

Besides have to update the hundreds of snapshots that contain diagnostics, I've generally tried to keep the individual tests that existed for the previous `codespan` implementation and simply move them to check the equivalent code in the new implementation


crates/rome_js_analyze/tests/specs/correctness/noUnnecessaryContinue.js.snap
```diff
@@ -133,10 +137,12 @@ noUnnecessaryContinue.js:9:2 lint/correctness/noUnnecessaryContinue  FIXABLE  
 
   ! Unnecessary continue statement
   
-    ┌─ noUnnecessaryContinue.js:9:2
-    │
-  9 │     continue;
-    │     ^^^^^^^^^
+     7 │ }
+     8 │ while (1) {
+   > 9 │ 	continue;
```
ematipico, line 142: This seems to be not aligned

crates/rome_js_analyze/tests/specs/correctness/noUnnecessaryContinue.js.snap
```diff
@@ -157,10 +163,12 @@ noUnnecessaryContinue.js:14:3 lint/correctness/noUnnecessaryContinue  FIXABLE  
 
   ! Unnecessary continue statement
   
-     ┌─ noUnnecessaryContinue.js:14:3
-     │
-  14 │         continue;
-     │         ^^^^^^^^^
+    12 │ 	if (i > 5) {
+    13 │ 		console.log("foo");
+  > 14 │ 		continue;
```
ematipico, line 168: Not aligned. Is it github or the file itself?

crates/rome_js_analyze/tests/specs/correctness/useSingleVarDeclarator.js.snap
```diff
@@ -46,10 +46,12 @@ useSingleVarDeclarator.js:5:2 lint/correctness/useSingleVarDeclarator  FIXABLE
...
   ! Declare variables separately
   
-    ┌─ useSingleVarDeclarator.js:5:2
-    │
-  5 │     let foo, bar;
-    │     ^^^^^^^^^^^^^
+    3 │ function test() {
+    4 │ 	// Comment
+  > 5 │ 	let foo, bar;
+      │  ^^^^^^^^^^^^^
```
ematipico, line 52: Not aligned

crates/rome_diagnostics/src/v2/display/frame.rs
```diff
@@ -1,53 +1,621 @@
...
+use rome_console::{fmt, markup};
 use rome_text_edit::TextRange;
+use rome_text_size::TextSize;
+
+use crate::v2::{
+    location::{BorrowedSourceCode, LineIndex},
+    LineIndexBuf, Location,
+};
+
+// SAFETY: These constant `NonZeroUsize` are being initialized with non-zero values
```
ematipico, line 18:
```suggestion
// SAFETY: These constants `NonZeroUsize` are being initialized with non-zero values
```

ematipico (approved):
I tried to check some of the new diagnostics and found few that are misaligned, but not sure if it's github or the snapshot itself.

Also, now that we removed `codespan` from our code base, we also need to remove it from our website, credits section.

crates/rome_js_analyze/tests/specs/correctness/noUnnecessaryContinue.js.snap
leops: This issue was caused by tab characters, I've implemented special handling for these to preserve the correct alignment

crates/rome_diagnostics/src/v2/display/frame.rs
```diff
@@ -1,53 +1,664 @@
...
+    };
+
+    let source_file = SourceFile::new(source_code);
 
-        let start = span.start();
-        let end = span.end();
+    let start_index = span.start();
+    let start_location = match source_file.location(start_index) {
+        Ok(location) => location,
+        Err(_) => return Ok(()),
```
MichaReiser, line 41: Why is it `OK` here to return `Ok` if the location is missing?

crates/rome_diagnostics/src/v2/display/frame.rs
```diff
@@ -1,53 +1,664 @@
...
+        Some(ellipsis_start..=ellipsis_end)
+    } else {
+        None
+    };
+
+    // Calculate the maximum width of the line number
+    let max_gutter_len = calculate_print_width(context_end);
+    let mut printed_lines = false;
+
+    for line_index in IntoIter::new(context_start..=context_end) {
```
MichaReiser, line 98: Why is it required to use `IntoIter` instead of `context_start..=context_end` directly?

crates/rome_diagnostics/src/v2/display/frame.rs
```diff
@@ -1,53 +1,664 @@
...
+            line_starts: source_code.line_starts.map_or_else(
+                || Cow::Owned(LineIndexBuf::from_source_text(source_code.text)),
+                Cow::Borrowed,
+            ),
+        }
+    }
+
+    /// Return the starting byte index of the line with the specified line index.
+    /// Convenience method that already generates errors if necessary.
+    fn line_start(&self, line_index: usize) -> io::Result<TextSize> {
```
MichaReiser, line 434: I noticed that you use `io::Result` in this  module rather than using your own Result type. The downside I see is that the caller knows less about the potential errors (may have to deal with more errors than can actually happen) and to me, this module has little to do with `io`. 

What where your reasons for going for `io`?

MichaReiser (approved):

crates/rome_diagnostics/src/v2/display/frame.rs
leops: An error can be returned here if the byte index of the start of the code frame location could not be mapped to a line + column position, generally because the index overflows the length of the source code. In this case opted to transparently skip rendering of the code frame, since printing an error inside of a diagnostic might be confusing to the user

MichaReiser: That makes sense. That's something I haven't thought of. May be worth documenting in some place.

crates/rome_diagnostics/src/v2/display/frame.rs
leops: The `IntoIter` struct exists to implement the `Iterator` trait for ranges of `OneIndexed`, since the `std::ops::RangeInclusive<T>` type already implements `IntoIterator` for all types `T` that implement the `Step` trait, however that trait is currently unstable and thus cannot be implemented for the `OneIndexed` struct

crates/rome_diagnostics/src/v2/display/frame.rs
leops: Ultimately all the code in this module is related to display, and has to execute within an implementation of the `fmt::Display` trait. This trait requires the error type of formatting to be `io::Error` (since ultimately all display operations write the resulting string to an I/O primitive), in turn this means that all the code that executes within the display logic must return an error type that's either `io::Error`, or at least an error that that can be converted into it (in this case, these methods were ported over from `rome_console` and were already set to return an `io::Result` so I didn't have to change anything)

PR #3282, opened 2022-09-27
feat(rome_js_analyzer): rule `noUndeclaredVariables`

### LINKED ISSUE #3281: `noUndeclaredVariables`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/js/noUndeclaredVariables.md

ematipico:
## Summary

Closes #3281 

The rule is completed for this iteration. We miss a feature that allows us to collect the `globals` from the configuration, which we should evaluate for the next iteration.

The hardcoded variables were copied from the old rule so maybe it's not up to date. Are there any other globals we should track?

## Test Plan

Added some test case


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_undeclared_variables.rs
```diff
@@ -0,0 +1,208 @@
...
+        name: "noUndeclaredVariables",
+        recommended: false,
+    }
+}
+
+declare_node_union! {
+    pub(crate) NoUndeclaredVariablesQuery = JsReferenceIdentifier| JsxReferenceIdentifier
+}
+
+const NODE_VARIABLES: [&str; 5] = ["require", "__dirname", "__filename", "module", "exports"];
```
leops, line 22: We should probably move these to a shared file, along with the list of JS builtins in `noShadowRestrictedNames`.
Also we should include those builtins to the list of well-known globals, unless it's being handled somewhere it seems like this rule would flag expressions like `new ArrayBuffer()`

MichaReiser: This list should probably include all names from the `global` object

https://nodejs.org/api/globals.html
https://developer.mozilla.org/en-US/docs/Web/API/Window
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope

https://developer.mozilla.org/en-US/docs/Glossary/Global_object

Eslint useses an environment option which I think is a good idea to have and gets the globals from the npm [`globals`](https://www.npmjs.com/package/globals) package 
https://eslint.org/docs/latest/user-guide/configuring/language-options#specifying-environments

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_undeclared_variables.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+fn is_declared(reference_name: &str, scope: Scope) -> bool {
+    let binding = scope.get_binding(reference_name);
+
+    // TODO: add here the check for global variables defined in the configuration, currently not supported
+    binding.is_some()
+        || NODE_VARIABLES.contains(&reference_name)
+        || TS_VARIABLES.contains(&reference_name)
+        || BROWSER_VARIABLES.contains(&reference_name)
```
MichaReiser, line 207: It could be interesting to use [binary search](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) to find the element rather than going through all elements every time.

Boshen:
FYI you can get the most update to date globals from https://github.com/sindresorhus/globals
I found this from eslint's dependencies.

ematipico:
> FYI you can get the most update to date globals from [sindresorhus/globals](https://github.com/sindresorhus/globals) I discovered this from eslint's dependencies.

Fortunately we came prepared :) https://github.com/rome/tools/blob/archived-js/internal/compiler/scope/globals.ts

We already had forked Sindre's data

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_undeclared_variables.rs
ematipico: I ported the globals we had in Rome classic, which were a fork on Sindre's repository.

In the current rule I only used some of the globals added. I think they should cover the majority of the cases.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_undeclared_variables.rs
ematipico: Unfortunately the map we ported doesn't seem to be sorted, so we can't use the binary search for the time being.

In the future we could have some script that downloads these globals and codegen them and sort them. But I would leave it out from this PR.

ematipico:
!bench_analyzer

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_undeclared_variables.rs
ematipico: The last commit uses the binary search!

ematipico:
!bench_analyzer

ematipico:
The performance is mostly the same - `.contains()` vs `.binary_search`. 

It's also clear that this rule adds a bunch of overhead in general (it checks every single identifier). I suppose it's OK to merge it and keep it under nursery for a while. An enhancement that I might see in the future - suggested by @MichaReiser  - is to give to the analyzer awareness of the realm/environment e.g. node vs browser, language version, typescript etc. 

This should give us some a way to narrow down the sets we need to check against.

IWANABETHATGUY:
I am not sure if the string search routine is the bottleneck or the semantic model, would you mind common out these lines and rerun benchmark?https://github.com/rome/tools/pull/3282/files#diff-c5aab19d6e97b3cf22168f2afc8eeb49f85c7475fd976594d5bad6e1abe7b30dR74-R77, if string search is the bottleneck, then maybe you could try Trie, [fst](https://github.com/BurntSushi/fst), and [aho-corasick](https://github.com/BurntSushi/aho-corasick)

IWANABETHATGUY:
And this one https://github.com/daac-tools/daachorse

ematipico:
!bench_analyzer

ematipico:
@IWANABETHATGUY you were right! The big overhead is just caused by the rule - as expected - but not the check itself! Thank you!

IWANABETHATGUY:
> @IWANABETHATGUY you were right! The big overhead is just caused by the rule - as expected - but not the check itself! Thank you!

You are welcome :)

MichaReiser (approved):
@leops @xunilrj do you know if there's a way for us to speed up the semantic model lookups? 

Or does the semantic model provide an API to get all references to  global variables so that this rule could be changed to instead run once per file.

leops (approved):
It might improve performance to use a faster hasher inside for the semantic model for name lookups, but ultimately there's just an inherent cost to inspecting every identifier in the program. However having the ability to query for the `SemanticModel` itself would be really useful since it could be used to directly enumerate all references in the file that do not have a declaration.

PR #3283, opened 2022-09-27
feat(rome_js_formatter): Member chain formatting

MichaReiser:
## Summary

This PR refactors the definition of when a `best_fitting` element fits. I plan to merge #3273 into this branch before merging to main because chaining the fits definition requires changes to the member chain which are already part of #3273 (reason why some tests are currently failing on this branch).

The existing definition of when a best fitting element fits is the same as for groups:

* It does not contain any content that forces a line break: `expand_parent`, any non-soft line break, ...
* Printing all the content does not exceed the configured print width

This PR changes the definition for best fitting to:
* Printing all its content up to the first **hard** line break does not exceed the configured print width

This is important to support use cases like:

```
expect().some.long.member.chain({
  object: withProperties,
  a: 4,
})
```

Where the whole member chain is a best-fitting element and it should take the first variant regardless of that the `{...}` object expression is printed over multiple lines. 

The main challenge of this change comes from that groups inside of a best fitting variant that contains an element that forces them to break **must** be measured in **expanded** mode. This complicates things because the *fits* function would need to "unwind" to the outermost group (best fitting variants form a boundary) if it finds any content that causes it to expand and then continue to measure in expanded mode. The problem with "unwinding" is that it would require restoring the fits queue and fits stack to the point right before measuring the group. 

I haven't found a way to do so with reasonable runtime which is why I opted for the same approach as prettier by pre-processing the IR before printing (or, in our case, after formatting). The pre-processing traverses through the whole document and sets `expanded` to `propagated` on `Group` if it contains any content that forces it to expand but only if there's no best-fitting element in between. The downside of this approach is that it does not respect "if_group_breaks" nor "if_group_fits" (could potentially be added) but it seems to be working reasonably well. It also comes with the advantage that less back-tracking is required in the printer when measuring if a group fits because it only has to measure groups that contain no hard line break.

## Test Plan

See #3273 that fixes many prettier compatibility issues. I also added a new printer test.


MichaReiser:
!bench_formatter

crates/rome_formatter/src/format_element/document.rs
```diff
@@ -17,6 +19,87 @@ pub struct Document {
...
+impl Document {
+    /// Sets [`expand`](tag::Group::expand) to [`GroupMode::Propagated`] if the group contains any of:
+    /// * a group with [`expand`](tag::Group::expand) set to [GroupMode::Propagated] or [GroupMode::Expand].
+    /// * a non-soft [line break](FormatElement::Line) with mode [LineMode::Hard], [LineMode::Empty], or [LineMode::Literal].
+    /// * a [FormatElement::ExpandParent]
+    ///
+    /// [`BestFitting`] elements act as expand boundaries, meaning that the fact that a
+    /// [`BestFitting`]'s content expands is not propagated past the [`BestFitting`] element.
+    ///
+    /// [`BestFitting`](FormatElement::BestFitting)
```
ematipico, line 31:
```suggestion
    /// [`BestFitting`]: FormatElement::BestFitting
```

ematipico (approved):

PR #3284, opened 2022-09-27
feat(vscode): watch configuration file

ematipico:
## Summary

This PR adds a new feature to the VSCode extension. It allows to watch changes to the `rome.json` file and when new changes occur, a listener sends a request. The server will  listen to this request and will update the workspace settings. 

I also added and change some logs, and used the `debug` trace instead. Our filter used the `DEBUG` level.

## Test Plan

I manually tested the feature

### DIFF
--- a/crates/rome_lsp/src/server.rs
+++ b/crates/rome_lsp/src/server.rs
@@ -6,6 +6,7 @@ use crate::{handlers, requests};
 use futures::future::ready;
 use futures::FutureExt;
 use rome_console::markup;
+use rome_fs::CONFIG_NAME;
 use rome_service::workspace::{RageEntry, RageParams, RageResult};
 use rome_service::{workspace, Workspace};
 use std::collections::HashMap;
@@ -132,6 +133,23 @@ impl LSPServer {
             .await;
         }
 
+        let base_path = self.session.base_path();
+
+        if let Some(base_path) = base_path {
+            let registration_options = DidChangeWatchedFilesRegistrationOptions {
+                watchers: vec![FileSystemWatcher {
+                    glob_pattern: format!("{}/rome.json", base_path.display()),
+                    kind: Some(WatchKind::all()),
+                }],
+            };
+            self.register_capability(Registration {
+                id: "workspace/didChangeWatchedFiles".to_string(),
+                method: "workspace/didChangeWatchedFiles".to_string(),
+                register_options: Some(serde_json::to_value(registration_options).unwrap()),
+            })
+            .await;
+        }
+
         if rename {
             self.register_capability(Registration {
                 id: "textDocument/rename".to_string(),
@@ -214,6 +232,37 @@ impl LanguageServer for LSPServer {
         let _ = params;
         self.session.fetch_client_configuration().await;
         self.setup_capabilities().await;
+    }
+
+    #[tracing::instrument(level = "trace", skip(self))]
+    async fn did_change_watched_files(&self, params: DidChangeWatchedFilesParams) {
+        let file_paths = params
+            .changes
+            .iter()
+            .map(|change| change.uri.to_file_path());
+        for file_path in file_paths {
+            match file_path {
+                Ok(file_path) => {
+                    let base_path = self.session.base_path();
+                    if let Some(base_path) = base_path {
+                        let possible_rome_json = file_path.strip_prefix(&base_path);
+                        if let Ok(possible_rome_json) = possible_rome_json {
+                            if possible_rome_json.display().to_string() == CONFIG_NAME {
+                                self.session.update_configuration().await;
+                                self.session.fetch_client_configuration().await;
+                                // for now we are only interested to the configuration file,
+                                // so it's OK to exist the loop
+                                break;
+                            }
+                        }
+                    }
+                }
+                Err(_) => {
+                    error!("The Workspace root URI {file_path:?} could not be parsed as a filesystem path");
+                    continue;
+                }
+            }
+        }
     }
 
     async fn did_open(&self, params: DidOpenTextDocumentParams) {
--- a/crates/rome_lsp/src/server.rs
+++ b/crates/rome_lsp/src/server.rs
@@ -3,6 +3,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
 use crate::capabilities::server_capabilities;
+use crate::requests::configuration_updated::{ConfigurationUpdatedParams, CHANGE_CONFIG_REQUEST};
 use crate::requests::syntax_tree::{SyntaxTreePayload, SYNTAX_TREE_REQUEST};
 use crate::session::{ClientInformation, Session, SessionHandle, SessionKey};
 use crate::utils::{into_lsp_error, panic_to_lsp_error};
@@ -12,6 +13,7 @@ use futures::FutureExt;
 use rome_console::markup;
 use rome_service::workspace::{RageEntry, RageParams, RageResult};
 use rome_service::{workspace, Workspace};
+use std::sync::Arc;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::sync::Notify;
 use tokio::task::spawn_blocking;
@@ -147,6 +149,13 @@ impl LSPServer {
             })
             .await;
         }
+    }
+
+    async fn change_config_request(&self, _params: ConfigurationUpdatedParams) -> LspResult<()> {
+        info!("Calling change config request");
+        self.session.update_configuration().await;
+        self.session.update_workspace_settings();
+        Ok(())
     }
 }
 
@@ -364,6 +373,7 @@ impl ServerFactory {
         });
 
         builder = builder.custom_method(SYNTAX_TREE_REQUEST, LSPServer::syntax_tree_request);
+        builder = builder.custom_method(CHANGE_CONFIG_REQUEST, LSPServer::change_config_request);
 
         // "shutdown" is not part of the Workspace API
         builder = builder.custom_method("rome/shutdown", |server: &LSPServer, (): ()| {
@@ -374,6 +384,7 @@ impl ServerFactory {
 
         builder = builder.custom_method("rome/rage", LSPServer::rage);
 
+        workspace_method!(builder, supports_feature);
         workspace_method!(builder, supports_feature);
         workspace_method!(builder, update_settings);
         workspace_method!(builder, open_file);
--- a/crates/rome_lsp/src/server.rs
+++ b/crates/rome_lsp/src/server.rs
@@ -1,9 +1,4 @@
-use std::collections::HashMap;
-use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::{Arc, Mutex};
-
 use crate::capabilities::server_capabilities;
-use crate::requests::configuration_updated::{ConfigurationUpdatedParams, CHANGE_CONFIG_REQUEST};
 use crate::requests::syntax_tree::{SyntaxTreePayload, SYNTAX_TREE_REQUEST};
 use crate::session::{ClientInformation, Session, SessionHandle, SessionKey};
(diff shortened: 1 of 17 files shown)


crates/rome_lsp/src/session.rs
```diff
@@ -243,6 +226,30 @@ impl Session {
         }
     }
 
+    /// If updates the [Workspace] settings with the new configuration that was
+    /// read from file.
+    #[tracing::instrument(level = "debug", skip(self))]
+    pub(crate) fn update_workspace_settings(&self) {
+        let mut configuration = self.configuration.write().unwrap();
```
MichaReiser, line 233: Is it safe to call `unwrap` here? Where do we receive the new configuration?

crates/rome_lsp/src/session.rs
```diff
@@ -243,6 +226,30 @@ impl Session {
...
+        // from the LSP, it's not needed anymore
+        if let Some(configuration) = configuration.take() {
+            trace!(
+                "The LSP will now use the following configuration: \n {:?}",
+                &configuration
+            );
+
+            let result = self
+                .workspace
+                .update_settings(UpdateSettingsParams { configuration });
```
MichaReiser, line 245: Nit: The field is named `configuration` but the struct is called `UpdateSettings`. Can we align on a single name: Do we call our configuration settings or configuration?

crates/rome_lsp/src/session.rs
```diff
@@ -243,6 +226,30 @@ impl Session {
...
+
+            if let Err(error) = result {
+                error!("{:?}", &error)
```
MichaReiser, line 248: Do we need to let the user know that updating the configuration failed?

editors/vscode/src/lsp_requests.ts
```diff
@@ -12,3 +12,14 @@ export const syntaxTreeRequest = new RequestType<
 	string,
 	void
 >("rome_lsp/syntaxTree");
+
+// Empty parameters
+export interface ConfigChangeParams {}
```
MichaReiser, line 17: Can we write out the term `Configuration`?

crates/rome_lsp/src/server.rs
```diff
@@ -37,6 +37,13 @@ impl LSPServer {
         let url = params.text_document.uri;
         requests::syntax_tree::syntax_tree(&self.session, &url).map_err(into_lsp_error)
     }
+
+    async fn change_config_request(&self, _params: ConfigurationUpdatedParams) -> LspResult<()> {
```
MichaReiser, line 41: Nit: Can we spell out `configuration`

MichaReiser (commented):
My main concern is around naming where there is a mix between "Settings" and configuration

crates/rome_lsp/src/session.rs
ematipico: > Is it safe to call unwrap here?

The `configuration` is protected by a `RwLock`, which means that if a panic is triggered, something else is currently holing writing rights to the same data. If that panics occurs, it's most likely an architecture issue not concerning this particular part of the code. We use `RwLock` for interior mutability, so I suppose it's safe to call `unwrap`.

> Where do we receive the new configuration?

The `configuration` is read from file system and saved inside `Session::configuration`. When from session we `.take()** (consume) the configuration and give it to the workspace. We use `RwLock<Configuration>` to enable interior mutability.

ematipico:
> My main concern is around naming where there is a mix between "Settings" and configuration

I can see the confusion that settings and configuration can cause, and I think it could make sense to normalize the naming convention. Although, there's these two different names allow us to create a distinction:
- the concept of configuration belongs to the users: user-friendly data structure,  data structure we need to serialize/deserialize 
- the concept of settings belongs to us developers - and the `Workspace`, where the configuration is actually **transformed** into another data structure that fits the needs of Rome's infrastructure. An example is the `"globals": [""]` field, which is transformed into a `Matcher` data structure

If we decided to normalize the naming convention, we might risk to lose this distinction. What do you think? How can we improve this division?

crates/rome_lsp/src/session.rs
ematipico: I think so, but maybe not here.. but I am not sure. 

It definitely makes sense to show an alert to the client if we fail to read the configuration file from disk and if its shape is incorrect. 

If we arrive to this error, is mostly an internal error. What should we tell the user here?

crates/rome_lsp/src/session.rs
ematipico: Would `user_configuration` help?

crates/rome_lsp/src/server.rs
```diff
@@ -228,6 +236,7 @@ impl ServerFactory {
             ready(Ok(Some(())))
         });
 
+        workspace_method!(builder, supports_feature);
```
leops, line 239: It seems the method is now registered twice

crates/rome_lsp/src/session.rs
leops: To be more specific, if another thread is currently accessing an `RwLock` with either read-only or read-write access calling `write` will not return an error, it will only block the current thread until the lock is ready again.
The error returned by `read`, `write` or `lock` (on `Mutex`) is related to the "poisoning" behavior of mutexes in the standards library: in case a thread panics while holding a lock, the corresponding mutex will be put in the "poisoned" state, making it impossible to access its internal state and causing all subsequent attempts to get a lock on mutex to fail.
The reasoning behind this is that we don't know what the thread was doing with the state of the mutex when it panicked, it may have been in the middle of a mutation and the panic could have left the inner value in an invalid in-between state. So in order to be safe by default, Rust disallows reading or writing to this state.
In general the most common way to handle a poison error is unwrapping: if it happens is means another thread has already panicked earlier and may have corrupted the data structure you need to read or write, so there's often no other choice but to "propagate" the panic to the current thread too.

crates/rome_service/src/workspace.rs
```diff
@@ -302,6 +302,13 @@ pub trait Workspace: Send + Sync + RefUnwindSafe {
         params: GetSyntaxTreeParams,
     ) -> Result<GetSyntaxTreeResult, RomeError>;
 
+    /// Triggered when the configuration file has been updated.
+    ///
+    /// This function is usually used by clients like the LSP server
+    fn configuration_updated(&self) -> Result<(), RomeError> {
```
leops, line 308: Having both a `configuration_updated` method and an `update_settings` method at the same time seems a bit awkward: the first method relies on the Workspace being able to reload the configuration file by itself from the disk, while the second relies on the client loading the configuration file and providing its content to the Workspace

ematipico: What would you suggest to do?

leops: I though about this for a bit, we need a way for "configuration overrides" to be injected into the Workspace (to support passing configuration options as CLI arguments), maybe we could have a single "configuration updated" method that would explicitly provide a "source" for the configuration: either it comes from the disk (in this case the Workspace is responsible for loading it, and it has a lower priority), or it comes from memory (either from the CLI arguments or explicitly from the node.js API, takes priority over the configuration loaded from the filesystem).
Unfortunately this still means that while the Workspace is responsible for loading the file from the filesystem, the Workspace client (CLI or Language Server) is responsible for registering a filesystem watch on it and proxying those notifications to the Workspace.

MichaReiser:
> > My main concern is around naming where there is a mix between "Settings" and configuration
> 
> I can see the confusion that settings and configuration can cause, and I think it could make sense to normalize the naming convention. Although, these two different names allow us to create a distinction:
> 
>     * the concept of configuration belongs to the users: user-friendly data structure,  data structure we need to serialize/deserialize
> 
>     * the concept of settings belongs to us developers - and the `Workspace`, where the configuration is actually **transformed** into another data structure that fits the needs of Rome's infrastructure. An example is the `"globals": [""]` field, which is transformed into a `Matcher` data structure
> 
> 
> If we decided to normalize the naming convention, we might risk to lose this distinction. What do you think? How can we improve this division?

Oh, I wasn't aware that these are two different things. It may then make sense to keep the distinct names.

editors/vscode/src/main.ts
```diff
@@ -108,6 +109,23 @@ export async function activate(context: ExtensionContext) {
...
+		workspace.workspaceFolders[0].uri.path,
+		"rome.json",
+	);
+	const watcher = workspace.createFileSystemWatcher(configurationPath);
+
+	watcher.onDidChange(() => {
+		const configurationContent = readFileSync(configurationPath, {
+			encoding: "utf-8",
+		});
+		const configurationAsJson = JSON.parse(configurationContent);
```
MichaReiser, line 123: Can we wrap this in a try catch to avoid parsing the configuration if it has syntax errors?

ematipico: Screenshot 2022-11-01 at 16 41 44

This is the last commit :)

editors/vscode/src/main.ts
```diff
@@ -108,6 +109,29 @@ export async function activate(context: ExtensionContext) {
...
 	handleActiveTextEditorChanged(window.activeTextEditor);
 
+	// we tell the client to add a watcher to the configuration file
+	const configurationPath = join(
+		workspace.workspaceFolders[0].uri.path,
+		"rome.json",
+	);
+	const watcher = workspace.createFileSystemWatcher(configurationPath);
+
+	watcher.onDidChange(() => {
```
leops, line 119: This should not be implemented in the extension itself, instead we should rely on the `workspace/didChangeWatchedFiles` capability so that this works on all LSP-compatible editors instead of just VSCode

ematipico: This is weird though. Creating a watcher using `createFileSystemWatcher` doesn't fire any events of type `didChangeWatchedFiles`

MichaReiser: @leops is it the idea that the workspace would listen to `document/change` methods for configuration files? 

The problem I see with the `didChangeWatchedFiles` notification is that it doesn't contain the content of the changed file. 

[didChangeWatchedFiles ](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles)

leops: My idea was that the LSP server would register the `workspace/didChangeWatchedFiles` capability with the `**/rome.json` pattern when it gets initialized, then simply call `update_configuration`, `fetch_client_configuration` and `setup_capabilities` in the `did_change_watched_files` handler to reload the configuration

ematipico:
@leops that's working now, using the server instead of the client

editors/vscode/src/main.ts
```diff
@@ -108,6 +109,16 @@ export async function activate(context: ExtensionContext) {
 
 	handleActiveTextEditorChanged(window.activeTextEditor);
 
+	// we tell the client to add a watcher to the configuration file
```
leops, line 112: This is not needed anymore, it should be fully handled by the language client now

crates/rome_lsp/src/session.rs
```diff
@@ -267,6 +255,30 @@ impl Session {
...
 
+    /// If updates the [Workspace] settings with the new configuration that was
+    /// read from file.
+    #[tracing::instrument(level = "debug", skip(self))]
+    pub(crate) fn update_workspace_settings(&self) {
+        let mut configuration = self.configuration.write().unwrap();
+
+        // This operation is intended, we want to consume the configuration because once it's read
+        // from the LSP, it's not needed anymore
+        if let Some(configuration) = configuration.take() {
```
leops, line 266: I feel like the whole flow of loading the configuration from the filesystem, merging it with the client configuration and updating the workspace settings is more complicated than it really needs to be, but I guess that's always something we can improve in a later refactor

leops (approved):

crates/rome_lsp/src/session.rs
ematipico: I share the same sentiment, and yes I think it's better to have this refactor in it's own PR.

PR #3285, opened 2022-09-27
feat(rome_js_semantic): Semantic Model support for closures

xunilrj:
## Summary

This PR introduces closures in the semantic model.
One interesting decision is what to do with:

```js
function f(c) {
    console.log(a);
    function g() {
        console.log(b, c);
    }
}
```

The implementations is returning that ```f``` captures ```a```, and it has a inner closure that captures ```b``` and ```c```. Which is closer to the code.

## Test Plan

```
cargo test -p rome_js_semantic -- ok_semantic_model_closure
```

### DIFF
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -1279,5 +1279,4 @@ mod test {
         assert_is_exported(true, "A", "enum A {}; exports = A");
         assert_is_exported(true, "A", "enum A {}; exports.A = A");
     }
-
 }
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -70,33 +70,27 @@ impl<T: HasDeclarationAstNode> IsExportedCanBeQueried for T {
     }
 }
 
-#[derive(Clone, Debug)]
+/// Represents a refererence inside a scope.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
 struct ScopeReference {
     range: TextRange,
 }
 
-impl std::hash::Hash for ScopeReference {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.range.hash(state);
-    }
-}
-
-impl PartialEq for ScopeReference {
-    fn eq(&self, other: &Self) -> bool {
-        self.range == other.range
-    }
-}
-
-impl Eq for ScopeReference {}
-
 #[derive(Debug)]
 struct SemanticModelScopeData {
+    // The scope range
     range: TextRange,
+    // The parent scope of this scope
     parent: Option<usize>,
+    // All children scope of this scope
     children: Vec<usize>,
+    // All bindings of this scope
     bindings: Vec<TextRange>,
+    // Map pointing to the [bindings] vec  of each bindings by its name
     bindings_by_name: HashMap<SyntaxTokenText, usize>,
+    // All read references of a scope
     read_references: Vec<ScopeReference>,
+    // All write references of a scope
     write_references: Vec<ScopeReference>,
 }
 
@@ -107,10 +101,12 @@ struct SemanticModelScopeData {
 #[derive(Debug)]
 struct SemanticModelData {
     root: JsAnyRoot,
+    // All scopes of this model
     scopes: Vec<SemanticModelScopeData>,
     scope_by_range: rust_lapper::Lapper<usize, usize>,
     // Maps the start of a node range to a scope id
     scope_hoisted_to_by_range: HashMap<TextSize, usize>,
+    // Map to each by its range
     node_by_range: HashMap<TextRange, JsSyntaxNode>,
     // Maps any range in the code to its declaration
     declared_at_by_range: HashMap<TextRange, TextRange>,
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -1,7 +1,9 @@
 mod closure;
 
+use crate::{SemanticEvent, SemanticEventExtractor};
+pub use closure::*;
 use rome_js_syntax::{
-    JsAnyRoot, JsArrowFunctionExpression, JsFunctionDeclaration, JsIdentifierAssignment,
+    JsAnyRoot, JsIdentifierAssignment,
     JsIdentifierBinding, JsLanguage, JsReferenceIdentifier, JsSyntaxKind, JsSyntaxNode,
     JsxReferenceIdentifier, TextRange, TextSize, TsIdentifierBinding,
 };
@@ -12,8 +14,6 @@ use std::{
     iter::FusedIterator,
     sync::Arc,
 };
-use crate::{SemanticEvent, SemanticEventExtractor};
-pub use closure::*;
 
 /// Marker trait that groups all "AstNode" that are declarations
 pub trait IsDeclarationAstNode: AstNode<Language = JsLanguage> {
@@ -70,7 +70,6 @@ impl<T: HasDeclarationAstNode> IsExportedCanBeQueried for T {
         Some(model.data.is_exported(range))
     }
 }
-
 
 #[derive(Clone, Debug)]
 struct ScopeReference {
@@ -679,7 +678,6 @@ impl SemanticModel {
         Closure::from_node(self.data.clone(), node)
     }
 }
-
 
 // Extensions
 
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -746,7 +746,7 @@ impl<T: IsDeclarationAstNode> AllReferencesExtensions for T {}
 pub trait ClosureExtensions {
     fn closure(&self, model: &SemanticModel) -> Closure
     where
-        Self: HasClosureAstNode,
+        Self: HasClosureAstNode + Sized,
     {
         model.closure(self)
     }
@@ -998,10 +998,7 @@ pub fn semantic_model(root: &JsAnyRoot) -> SemanticModel {
 mod test {
     use super::*;
     use rome_diagnostics::file::FileId;
-    use rome_js_syntax::{
-        JsArrowFunctionExpression, JsFunctionDeclaration, JsReferenceIdentifier, JsSyntaxKind,
-        SourceType, TsIdentifierBinding,
-    };
+    use rome_js_syntax::{JsReferenceIdentifier, JsSyntaxKind, SourceType, TsIdentifierBinding};
     use rome_rowan::SyntaxNodeCast;
 
     #[test]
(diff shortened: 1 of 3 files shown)


ematipico:
I have few questions:
1. How can we use this feature? 
2. What's the difference between the concept of "closure" and the concept of "scope"?

MichaReiser:
> The implementations is returning that f captures a, and it has a inner closure that captures b and c. Which is closer to the code.

What are the alternatives you considered?

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -603,6 +673,116 @@ impl SemanticModel {
...
+                if self.closure_range.intersect(declaration).is_none() {
+                    captures.insert(reference.clone());
+                }
+            }
+        }
+
+        captures
+    }
+
+    // Return all [Reference] this closure captures
```
MichaReiser, line 735: I think it would be helpful to add examples of what `all_captures` means, particularly how it differs from` children`

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -603,6 +673,116 @@ impl SemanticModel {
...
+                    q.extend(scope.children.iter());
+                }
+            }
+        }
+
+        r
+    }
+
+    /// Return all immediate children closures of this closure
+    pub fn children(&self) -> Vec<Closure> {
```
MichaReiser, line 772: I think it would help to have an example here what child closures are.

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -603,6 +673,116 @@ impl SemanticModel {
...
+        for scope_id in self.children_scopes() {
+            let scope = &self.data.scopes[scope_id];
+            closures.push(Closure {
+                data: self.data.clone(),
+                scope_id,
+                closure_range: scope.range,
+            })
+        }
+
+        closures
```
MichaReiser, line 784: Would it be possible to return an `Iterator` rather than a `Vec`?

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -603,6 +673,116 @@ impl SemanticModel {
...
+    data: Arc<SemanticModelData>,
+    scope_id: usize,
+    closure_range: TextRange,
+}
+
+impl Closure {
+    fn capture_scopes(&self) -> Vec<usize> {
+        let scope = &self.data.scopes[self.scope_id];
+        let mut q = VecDeque::from_iter(scope.children.iter().cloned());
+        let mut r = vec![self.scope_id];
```
MichaReiser, line 702: Can we use more telling names than `q` and `r` (It's unclear to me what `r` is used for)

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -603,6 +673,116 @@ impl SemanticModel {
...
+                JsSyntaxKind::JS_FUNCTION_DECLARATION
+                | JsSyntaxKind::JS_ARROW_FUNCTION_EXPRESSION => {}
+                _ => {
+                    r.push(id);
+                    q.extend(scope.children.iter());
+                }
+            }
+        }
+
+        r
```
MichaReiser, line 717: I think this could potentially be implemented by an iterator that uses a `VecDequeue` to "queue" the scopes to process but avoids the allocation of a vector for `r`

xunilrj:
> What are the alternatives you considered?

Closure of ```f``` capturing everything: ```a```, ```b```, ```c```.

The bigger rationale was.

```js
const [a, setA] = useState(0);
const [b, setB] = useState(0);
const [c, setC] = useState(0);
useEffect(() => {
    const print = (x) => console.log(x, a);
    console.log(b, c);
}, [a,b,c]);
```

Look that print is never called.

If we always returns all captures, it is impossible to reason about nuances as the above. We will need to conclude that the dependency list is fine. It is listing all variables that are captured.

If we return just what is actually captured, we can make a finer analysis and come to the conclusion that ```b``` and ```c``` are not needed in the dependency list. Granted that this analysis would be very complex and not something we will do on v1. But at least this enables us to do it.

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -68,13 +70,35 @@ impl<T: HasDeclarationAstNode> IsExportedCanBeQueried for T {
     }
 }
 
+#[derive(Clone, Debug)]
+struct ScopeReference {
```
ematipico, line 74: Can we document this struct and what's it for?

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -68,13 +70,35 @@ impl<T: HasDeclarationAstNode> IsExportedCanBeQueried for T {
...
+
 #[derive(Debug)]
 struct SemanticModelScopeData {
     range: TextRange,
     parent: Option<usize>,
     children: Vec<usize>,
     bindings: Vec<TextRange>,
     bindings_by_name: HashMap<SyntaxTokenText, usize>,
+    read_references: Vec<ScopeReference>,
+    write_references: Vec<ScopeReference>,
```
ematipico, line 101: Can we document these two new fields?

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -245,6 +307,12 @@ impl Scope {
     pub fn syntax(&self) -> &JsSyntaxNode {
         &self.data.node_by_range[self.range()]
     }
+
+    /// Return the [Closure] associated with this scope if
+    /// this scope comes from a function or arrow function.
```
ematipico, line 312: I think it's worth explaining what this function returns in cases where we call it from a node that is not inside a closure.

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -603,6 +671,11 @@ impl SemanticModel {
     {
         node.is_exported(self)
     }
+
+    /// Returns the [Closure] associated with the node.
+    pub fn closure(&self, node: &impl HasClosureAstNode) -> Closure {
+        Closure::from_node(self.data.clone(), node)
+    }
```
ematipico, line 678: Shouldn't this return a `Option`? I mean, if we call `closure` on a node that is NOT inside a closure, shouldn't this be `None`?

crates/rome_js_semantic/src/semantic_model/closure.rs
```diff
@@ -0,0 +1,178 @@
...
+    type Item = Closure;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(scope_id) = self.scopes.pop() {
+            let scope = &self.data.scopes[scope_id];
+            let node = &self.data.node_by_range[&scope.range];
+            match node.kind() {
+                JsSyntaxKind::JS_FUNCTION_DECLARATION
+                | JsSyntaxKind::JS_FUNCTION_EXPRESSION
+                | JsSyntaxKind::JS_ARROW_FUNCTION_EXPRESSION => {
```
ematipico, line 88: I think we have to track `JsMethodClassMember` and `JsMethodObjectMember` too

ematipico (changes requested):

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: That is why you can only pass ```Node```s that have closure using ```HasClosureAstNode```.

crates/rome_js_semantic/src/semantic_model/closure.rs
xunilrj: hum... you are right.

crates/rome_js_semantic/src/semantic_model.rs
ematipico: Ah, I see it now. It's like the other APIs.. it wasn't immediate. I understood the other APIs only by using them and understanding the error and navigated to the `impl IsDeclarationAstNode`

crates/rome_js_semantic/src/semantic_model/closure.rs
```diff
@@ -2,21 +2,126 @@ use std::sync::Arc;
...
+
+pub enum AnyHasClosureNode {
+    JsFunctionDeclaration(JsFunctionDeclaration),
+    JsFunctionExpression(JsFunctionExpression),
+    JsArrowFunctionExpression(JsArrowFunctionExpression),
+    JsConstructorClassMember(JsConstructorClassMember),
+    JsMethodClassMember(JsMethodClassMember),
+    JsGetterClassMember(JsGetterClassMember),
+    JsSetterClassMember(JsSetterClassMember),
+}
```
ematipico, line 73: We have the macro `declare_node_union`. Wouldn't that cover most of use cases you're having?

ematipico (approved):

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -165,6 +192,29 @@ impl PartialEq for SemanticModelData {
...
+            self.q.extend(scope.children.iter());
+            Some(Scope {
+                data: self.data.clone(),
+                id,
+            })
+        } else {
+            None
+        }
+    }
+}
```
MichaReiser, line 216: Nit: Implement `FusedIterator`

crates/rome_js_semantic/src/semantic_model/closure.rs
```diff
@@ -0,0 +1,416 @@
...
+
+impl HasClosureAstNode for JsSetterClassMember {
+    #[inline(always)]
+    fn node_text_range(&self) -> TextRange {
+        self.syntax().text_range()
+    }
+}
+
+pub enum AnyHasClosureNode {
```
MichaReiser, line 65: Would you mind adding documentation to all newly introduced public nodes

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_js_semantic/src/semantic_model/closure.rs
xunilrj: No, because I need to implement a specific trait.  
Nonetheless I created a macro to make this less horrible.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: Done.

crates/rome_js_semantic/src/semantic_model/closure.rs
xunilrj: Done.

PR #3286, opened 2022-09-27
feat(rome_js_analyzer): code action for `noVoidElementsWithChildren`

### LINKED ISSUE #3272: `noVoidElementsWithChildren`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noVoidElementsWithChildren.md

ematipico:
## Summary

Closes #3272 

The PR implements the code action for the rule `noVoidElementsWithChildren`

## Test Plan

Updated the test cases and added new test case for the documentation


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -57,43 +67,71 @@ fn is_void_dom_element(element_name: &str) -> bool {
...
+    children_prop: Option<UnwelcomedProp>,
+    /// An instance of [ReactCreateElementCall]
+    create_react_element: Option<ReactCreateElementCall>,
+}
+
+declare_node_union! {
+    pub(crate) UnwelcomedProp = JsxAttribute | JsPropertyObjectMember
+}
+
+impl UnwelcomedProp {
```
MichaReiser, line 84: Nit: `Unwelcome` sounds a bit odd to me in this context. 

How about `InvalidProp`, `IllegalProp`?

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -57,43 +67,71 @@ fn is_void_dom_element(element_name: &str) -> bool {
...
-    /// If the current element has children props
+    /// If the current element has children props in style `<Component>'children'</Component>
     children_cause: bool,
     /// If the current element has the prop `dangerouslySetInnerHTML`
-    dangerous_prop_case: bool,
+    dangerous_prop_case: Option<UnwelcomedProp>,
+    /// If the current element has the prop `children`
+    children_prop: Option<UnwelcomedProp>,
+    /// An instance of [ReactCreateElementCall]
+    create_react_element: Option<ReactCreateElementCall>,
```
MichaReiser, line 77: Is it possible to have all combinations of `children_cause`, `dangerous_prop_case`, `children_prop`, `and `crate_react_element` or can we limit the possible combinations and represent them as a union? 

I'm asking because limiting the possibilities can help readers to understand what combinations can exist and using an enum to name them makes them into explicit concepts (that can even be documented). It may also help to avoid some `unreachable` branches in the code.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -206,6 +276,95 @@ impl Rule for NoVoidElementsWithChildren {
...
+                        }
+
+                        if let Some(dangerous_prop) = dangerous_prop {
+                            if dangerous_prop == attribute {
+                                continue;
+                            }
+                        }
+                    }
+                    new_attribute_list.push(attribute)
+                }
```
MichaReiser, line 319: Nit: This can be written using `filter_map`

```rust
let new_attribute_list: Vec<_> = opening_element.attributes.filter_map(|attribute| {
  let children_prop = state
        .children_prop
        .as_ref()
        .and_then(|prop| prop.as_jsx_attribute());

    let dangerous_prop = state
        .dangerous_prop_case
        .as_ref()
        .and_then(|prop| prop.as_jsx_attribute());

    if let JsxAnyAttribute::JsxAttribute(attribute) = &attribute {
        if let Some(children_prop) = children_prop {
            if children_prop == attribute {
              return None;
            }
        }

        if let Some(dangerous_prop) = dangerous_prop {
            if dangerous_prop == attribute {
                return None;
            }
        }
        
        Some(attribute)
}).collect();
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -206,6 +276,95 @@ impl Rule for NoVoidElementsWithChildren {
...
+                )
+                .build();
+                mutation.replace_element(
+                    element.clone().into_syntax().into(),
+                    new_node.into_syntax().into(),
+                );
+            }
+            NoVoidElementsWithChildrenQuery::JsCallExpression(_) => {
+                // SAFETY: safe because create_react_element is always set in case of error around a call expression
+                let create_react_element = state.create_react_element.as_ref().unwrap();
```
MichaReiser, line 338: It would be nice to express this as part of the `state` by creating an enum with different variants for the cases. 

It would leverage Rust's capabilities to guide future developers to not make any changes that violate this constraint in the future. For example, you could move the `crate_react_element` to the `NoVoidElementsWithCHildrenQuery::JsCallExpression` state?

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -206,6 +276,95 @@ impl Rule for NoVoidElementsWithChildren {
...
+                    }
+                }
+                if let Some(children_prop) = state.children_prop.as_ref() {
+                    mutation.remove_node(children_prop.clone());
+                }
+                if let Some(dangerous_prop_case) = state.dangerous_prop_case.as_ref() {
+                    mutation.remove_node(dangerous_prop_case.clone());
+                }
+            }
+            // self closing elements don't have inner children so we can safely just remove the props
```
MichaReiser, line 351:
```suggestion
            // self closing elements don't have inner children so we can safely remove the props
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
ematipico: It really depends on how we want the rule to work. My initial implementation was using an enum, but then I noticed that the old rule works by tracking all the invalid cases.

In this rule, we could have up to three errors at once:
```js
React.createElement('img', {
	children: 'child', // <= diagnostic triggered
	dangerouslySetInnerHTML: 'html' // <= diagnostic triggered
}, 'child') // <= diagnostic triggered for 'child'
```

The upside of using an enum state would make the rule simple to understand for us, on the other hand this would generate up to three diagnostics that will show up to three code actions (this is a limitation of our infrastructure) - basically we shift the burden to the user. 

Tracking all the states at once allows us to emit one single code action and one single diagnostics.

Eventually ALL the suggested fixes are applied and the resultant code will be the same in both cases.

I don't mind both approaches.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -54,34 +64,84 @@ fn is_void_dom_element(element_name: &str) -> bool {
...
+        ///     Some child
+        /// </img>
+        /// ```
+        children_cause: bool,
+        /// If the current element has the prop `dangerouslySetInnerHTML`
+        dangerous_prop_cause: Option<JsxAttribute>,
+        /// If the current element has the prop `children`
+        children_prop: Option<JsxAttribute>,
+    },
+    ReactCreateElement {
```
MichaReiser, line 82: Nit: It may be helpful to document the `ReactCreateElement` and `Jsx` variants so that readers know what these variants represent.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
```diff
@@ -54,34 +64,84 @@ fn is_void_dom_element(element_name: &str) -> bool {
...
+        /// ```
+        children_cause: bool,
+        /// If the current element has the prop `dangerouslySetInnerHTML`
+        dangerous_prop_cause: Option<JsPropertyObjectMember>,
+        /// If the current element has the prop `children`
+        children_prop: Option<JsPropertyObjectMember>,
+        /// An instance of [ReactCreateElementCall]
+        react_create_element: ReactCreateElementCall,
+    },
+}
```
MichaReiser, line 96: Nit: An alternative to this because both variants have many common properties would be something along these lines:

```
// Naming is hard
enum Usage {
  Jsx,
  ReactCreateElement(ReactCreateElementCall)
}

struct NoVoidElementsWithChildrenCause {
  usage: Usage,
  children-cause: bool,
  ...
}
```

This would remove the need for many of the accessors but still allow to store additional data for each variant.

MichaReiser (approved):

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_void_elements_with_children.rs
ematipico: I thought about it too, but I preferred this way because it allowed me to better document the two different cases

PR #3288, opened 2022-09-28
feat(rome_js_analyzer): code action `noUselessFragments`

### LINKED ISSUE #3254: `noUselessFragments`

ematipico:
## Summary

Closes #3254 

This implements the last piece of code action for this rule. The logic might seems strange, but it's mostly because in some cases we need to transform the JSX syntax in valid JS syntax.

For example 

```jsx
<>{foo}</>
// in
foo
```

```jsx
<>foo</>
// in
"foo"
```

For example cases like `<>{..foo}</>` are not entitled to a code action because they would result in `{...foo}` which is not a correct syntax.

## Test Plan

Added test cases to cover the new logic


crates/rome_js_analyze/tests/specs/nursery/noUselessFragments/withJsxTextInvalid.jsx.snap
```diff
@@ -0,0 +1,30 @@
...
+    ┌─ withJsxTextInvalid.jsx:1:1
+    │
+  1 │ <>foo</>
+    │ ^^^^^^^^
+  
+  i Suggested fix: Remove the Fragment
+  
+      | @@ -1 +1 @@
+  0   | - <>foo</>
+    0 | + foo
```
leops, line 25: This fix isn't correct either, it's turning a JSX text literal into an identifier when it should be turned into the string literal `"foo"`

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
```diff
@@ -201,6 +212,55 @@ impl Rule for NoUselessFragments {
...
+                            child.l_curly_token().ok()?,
+                            js_statement_list(statement_list),
+                            child.r_curly_token().ok()?,
+                        );
+
+                        block_statement.syntax().clone()
+                    }
+
+                    // can't apply a code action because it will create invalid syntax
+                    JsxAnyChild::JsxSpreadChild(_) => return None,
```
leops, line 258: It might depend on what framework is being the JSX syntax is being used with here, but it should be safe to inline the expression being spread directly into the parent node here eg. turn `return <>{...foo}</>` into just `return foo`

ematipico: In this cases the rule is not triggered

MichaReiser: Changing it into `return foo` wouldn't be entirely safe as it changes semantics (`foo` no longer gets copied). But I think it should be safe to keep `...foo`. 

@ematipico Would you mind documenting an example of where replacing it would create invalid syntax as this isn't obvious to me.

MichaReiser (approved):

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_useless_fragments.rs
ematipico: Here's a list of valid cases from Rome classic: https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noUselessFragment.md#valid

@MichaReiser here's an [example](https://play.rome.tools/?code=%3C%3E%7B...foo%7D%3C%2F%3E%0A%7B...foo%7D&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&enabledNurseryRules=false&typescript=false&jsx=true#PAA+AHsALgAuAC4AZgBvAG8AfQA8AC8APgAKAHsALgAuAC4AZgBvAG8AfQA=) of incorrect syntax.  Even removing brackets wouldn't help.

PR #3290, opened 2022-09-29
feat(rome_js_formatter): Call arguments formatting

MichaReiser:
## Summary

This PR improves our formatting of call arguments:

* Improve our recognition of call expressions calling a `test` method (`beforeEach`, `it`, ...)
* Preserve empty lines between call arguments
* Refine rules when to group first or last argument
* Add support for Prettier's grouping of the first and last argument. 
* Add special handling for function composition (`compose(sortBy(x => x), flatten, map(x => [x, x*2]));`

## Test Plan

**Average compatibility**: 89.48 -> 91.72%
**Compatible lines**:  90.47 ->   93.49%


MichaReiser:
!bench_formatter

crates/rome_formatter/src/printed_tokens.rs
```diff
@@ -8,7 +8,14 @@ use std::collections::BTreeSet;
 #[derive(Debug, Clone, Default)]
 pub struct PrintedTokens {
     /// Key: Start of a token's range
-    offsets: BTreeSet<TextSize>,
+    offsets: IndexSet<TextSize>,
```
MichaReiser, line 11: I changed the printed tokens to use an `IndexSet` because tests became really slow (not as bad anymore but I had 15s run times before introducing the caching of the body) and about 5s were spent just cloning the `BTreeSet` every time a formatter snapshot was taken. 

This can be avoided with an `IndexSet` by simply storing the length of the set and truncating it when restoring.

crates/rome_js_formatter/src/js/expressions/template_element.rs
```diff
@@ -78,32 +76,8 @@ impl Format<JsFormatContext> for FormatTemplateElement {
 
         let format_inner = format_with(|f: &mut JsFormatter| match self.options.layout {
             TemplateElementLayout::SingleLine => {
-                // The goal is to print the expression on a single line, even if it exceeds the configured print width.
```
MichaReiser, line 81: Since we now have the `RemoveSoftLinesBuffer`... let's use it here too.

MichaReiser:
!bench_formatter

crates/rome_formatter/src/buffer.rs
```diff
@@ -428,6 +431,174 @@ where
...
+/// # Ok(())
+/// # }
+/// ```
+pub struct RemoveSoftLinesBuffer<'a, Context> {
+    inner: &'a mut dyn Buffer<Context = Context>,
+
+    /// Cache of "original" interned elements to the "cleaned" interned elements.
+    ///
+    /// It's fine to not snapshot the cache. The worst that can happen is that it holds on to now unused
+    /// interned elements. But there's little harm in that and all gets cleaned up when dropping the buffer.
```
ematipico, line 485:
```suggestion
    /// It's fine to not snapshot the cache. The worst that can happen is that it holds on interned elements
    /// that are now unused. But there's little harm in that and the cache is cleaned when dropping the buffer.
```

crates/rome_formatter/src/buffer.rs
```diff
@@ -428,6 +431,174 @@ where
...
+///     ]
+/// );
+///
+/// # Ok(())
+/// # }
+/// ```
+pub struct RemoveSoftLinesBuffer<'a, Context> {
+    inner: &'a mut dyn Buffer<Context = Context>,
+
+    /// Cache of "original" interned elements to the "cleaned" interned elements.
```
ematipico, line 482: I don't understand this phrase (especially after I've seen the implementation). Using "original" and "cleaned" makes hard to me because they don't have context. Do you mean the we keep  the interned elements stripped out of the soft lines?

crates/rome_formatter/src/lib.rs
```diff
@@ -565,6 +568,13 @@ pub enum FormatError {
 
     /// In case printing the document failed because it has an invalid structure.
     InvalidDocument(InvalidDocumentError),
+
+    /// Formatting failed because formatting some content encountered a situation where a layout
```
ematipico, line 572:
```suggestion
    /// Formatting failed because some content encountered a situation where a layout
```

crates/rome_formatter/src/lib.rs
```diff
@@ -565,6 +568,13 @@ pub enum FormatError {
...
+    /// choice by an enclosing object resulted in a poor layout for the child object.
```
ematipico, line 573:
```suggestion
    /// chosen by an enclosing node, resulted in a poor layout for the child node.
```
When I read "object" I think the actual objects in JS... maybe node is wider term. I thought about "element" but in can be mistaken for HTML elements.

crates/rome_formatter/src/lib.rs
```diff
@@ -565,6 +568,13 @@ pub enum FormatError {
...
+    ///
+    /// It's up to the enclosing object to pick another layout. This error should not be raised
```
ematipico, line 575:
```suggestion
    /// It's up to the enclosing node to pick another layout. This error should not be raised
```

crates/rome_formatter/src/lib.rs
```diff
@@ -565,6 +568,13 @@ pub enum FormatError {
...
+    /// if there's no outer object that handles the poor layout error to avoid that formatting of the whole document fails.
```
ematipico, line 576:
```suggestion
    /// if there's no outer node that is able to handle this particular error and avoid that formatting of the whole document fails.
```

crates/rome_formatter/src/printed_tokens.rs
```diff
@@ -17,43 +24,58 @@ impl PrintedTokens {
...
 
         if !self.offsets.insert(range.start()) {
             panic!("You tried to print the token '{token:?}' twice, and this is not valid.");
         }
     }
 
+    /// Enables or disables the assertion tracking
+    pub(crate) fn set_enabled(&mut self, enabled: bool) {
+        self.disabled = !enabled;
+    }
```
ematipico, line 41: nit: `set_disabled`? So it's inline with `disabled`

crates/rome_formatter/src/printed_tokens.rs
ematipico: Do we need to update the documentation of `PrintedTokens`? I noticed that there have been few changes to its APIs

crates/rome_js_formatter/src/context.rs
```diff
@@ -16,6 +17,28 @@ pub struct JsFormatContext {
...
     comments: Rc<JsComments>,
 
+    /// Stores the formatted content of one function body.
+    ///
+    /// Used during formatting of call arguments where function expressions and arrow function expressions
+    /// are formatted a second time if they are the first or last call argument.
+    ///
+    /// Caching the body in the call arguments formatting is important because it minimises the cases
+    /// where the algorithm is quadratic in case the function or arrow expression contains another
+    /// call expression with a function or call expression as first or last argument.
```
ematipico, line 27:
```suggestion
    /// Caching the body in the call arguments formatting is important. It minimises the cases
    /// where the algorithm is quadratic, in case the function or arrow expression contains another
    /// call expression with a function or call expression as first or last argument.
```

Long phrase without punctuation... I'm without breath 😆

crates/rome_js_formatter/src/context.rs
```diff
@@ -16,6 +17,28 @@ pub struct JsFormatContext {
...
+    ///
+    /// It's sufficient to only store a single cached body to cover the vast majority of cases
+    /// (there's no exception in any of our tests nor benchmark tests). The only not covered case is when
+    /// a parameter has an initializer that contains a call expression:
```
ematipico, line 31:
```suggestion
    /// (there's no exception in any of our tests nor benchmark tests). The only case not covered 
     is when
    /// a parameter has an initializer that contains a call expression:
```

crates/rome_js_formatter/src/context.rs
```diff
@@ -16,6 +17,28 @@ pub struct JsFormatContext {
...
+    ///
+    /// ```javascript
+    ///  test((
+    ///    problematic = test(() => body)
+    ///  ) => {});
+    ///  ```
+    ///
+    /// This should be are enough for us not to care about it.
```
ematipico, line 39:
```suggestion
    /// This should be enough for us not to care about it.
```

crates/rome_js_formatter/src/context.rs
```diff
@@ -16,6 +17,28 @@ pub struct JsFormatContext {
...
+    cached_function_body: Option<(JsAnyFunctionBody, FormatElement)>,
```
ematipico, line 40: I noticed there's a set and get for this function. Why isn't it there a remove/clean ?

crates/rome_js_formatter/src/js/expressions/arrow_function_expression.rs
```diff
@@ -175,13 +199,16 @@ impl FormatNodeRule<JsArrowFunctionExpression> for FormatJsArrowFunctionExpressi
 }
 
 /// writes the arrow function type parameters, parameters, and return type annotation
-fn format_signature(arrow: &JsArrowFunctionExpression) -> impl Format<JsFormatContext> + '_ {
-    format_with(|f| {
+fn format_signature(
+    arrow: &JsArrowFunctionExpression,
+    first_last_call_arg: bool,
```
ematipico, line 204: Can we update the documentation of the function? It seems that it does something more now.

Should we call it `is_first_last_call_argument`?

crates/rome_js_formatter/src/js/expressions/arrow_function_expression.rs
```diff
@@ -201,10 +228,33 @@ fn format_signature(arrow: &JsArrowFunctionExpression) -> impl Format<JsFormatCo
...
+            write!(
+                recording,
+                [group(&format_args![
+                    group(&format_parameters),
+                    group(&arrow.return_type_annotation().format())
+                ])]
+            )?;
+
+            if recording.stop().will_break() {
+                return Err(FormatError::PoorLayout);
```
ematipico, line 247: I think it's worth documenting why we are emitting the error. We could have a paragraph in the documentation of the function. Perhaps a "Safety" section highlighting **where** this error is captured is also important to have, considering that this error should be used only we know there's a node/object that can capture it.

crates/rome_js_formatter/src/ts/expressions/type_arguments.rs
```diff
@@ -28,15 +28,26 @@ impl FormatNodeRule<TsTypeArguments> for FormatTsTypeArguments {
...
-                let first_argument = first_argument?;
 
+        //   const isArrowFunctionVariable = path.match(
+        //     (node) =>
+        //       !(node[paramsKey].length === 1 && isObjectType(node[paramsKey][0])),
+        //     undefined,
+        //     (node, name) => name === "typeAnnotation",
+        //     (node) => node.type === "Identifier",
+        //     isArrowFunctionVariableDeclarator
+        //   );
```
ematipico, line 39: To delete?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -57,178 +71,685 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+impl Format<JsFormatContext> for FormatCallArgument {
+    fn fmt(&self, f: &mut Formatter<JsFormatContext>) -> FormatResult<()> {
+        self.fmt_with_cache_mode(FunctionBodyCacheMode::default(), f)?;
+        Ok(())
+    }
+}
 
-            if first_last_breaks {
-                return write!(f, [all_arguments_expanded]);
+/// Writes the function arguments and groups the first or last argument depending on `group_layout`.
```
ematipico, line 312:
```suggestion
/// Writes the function arguments, and groups the first or last argument depending on `group_layout`.
```

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -57,178 +71,685 @@ impl FormatNodeRule<JsCallArguments> for FormatJsCallArguments {
...
+            _ => self.argument.fmt(f),
+        }
+    }
+}
+
+/// Helper for formatting the last grouped argument (see [should_group_last_argument]).
+struct FormatGroupedLastArgument<'a> {
+    argument: &'a FormatCallArgument,
+    /// Is this the only argument in the arguments list
+    is_only: bool,
```
ematipico, line 586: nit: `alone` instead of `only`?

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
```diff
@@ -725,54 +1304,49 @@ fn contains_a_test_pattern(callee: &JsAnyExpression) -> SyntaxResult<bool> {
...
+                        .ok()
+                        .map(|name| name.token_text_trimmed())
                 }
-            }
-            _ => break,
-        };
+                _ => None,
+            },
+            _ => None,
+        }
```
ematipico, line 1348: nit: Any chance that we can split this file into smaller ones? 1300 LOC is an heavy document lol

ematipico (approved):

crates/rome_formatter/src/printed_tokens.rs
MichaReiser: Good point. I checked the documentation of `PrintedToken` and the documentation still applies.

crates/rome_formatter/src/buffer.rs
MichaReiser: It's a mapping from the `Interned` element that contains soft line breaks to the `Interned` element with the soft line breaks now removed. I'll try to make it more clear

crates/rome_formatter/src/lib.rs
MichaReiser: I prefer not to use `element` to avoid confusing them with `FormatElement`. I'll probably use `type` to be consistent with `Format` and add a link to `Format`.

crates/rome_js_formatter/src/context.rs
MichaReiser: Yeah sorry, the bad habit of a native german speaker.

crates/rome_js_formatter/src/context.rs
MichaReiser: Mainly because it doesn't harm to leave a cached body around:
* The formatted content is part of the formatted document anyway
* It protects against "obsolete" bodies by testing if the queried body matches to the body for which the formatted content is stored.

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: I prefer `only` over `alone` (alone sounds so sad)

crates/rome_js_formatter/src/js/expressions/call_arguments.rs
MichaReiser: I would prefer to keep the formatting logic inside of this file but I can see if I can move some of the util functions.

Issue #3292, opened 2022-09-30
`noArrayIndexKey`


ematipico:
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noArrayIndexKey.md

PR #3293, opened 2022-09-30
feat(rome_js_analyzer): rule `noArrayIndex`

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3292

## Test Plan

Added test cases to catch all the possible combinations

### DIFF
--- a/crates/rome_js_analyze/src/react.rs
+++ b/crates/rome_js_analyze/src/react.rs
@@ -242,9 +242,7 @@ pub(crate) fn is_react_call_api(
     api_name: &str,
 ) -> Option<bool> {
     // we bail straight away if the API doesn't exists in React
-    if !VALID_REACT_API.contains(&api_name) {
-        return None;
-    }
+    debug_assert!(VALID_REACT_API.contains(&api_name));
     Some(match expression {
         JsAnyExpression::JsStaticMemberExpression(node) => {
             let object = node.object().ok()?;
--- a/crates/rome_js_analyze/src/react.rs
+++ b/crates/rome_js_analyze/src/react.rs
@@ -7,6 +7,12 @@ use rome_js_syntax::{
 };
 use rome_rowan::{AstNode, AstSeparatedList};
 
+/// A trait to share common logic among data structures that "mimic" react APIs
+pub(crate) trait ReactApiCall {
+    /// It scans the current props and returns the property that matches the passed name
+    fn find_prop_by_name(&self, prop_name: &str) -> Option<JsPropertyObjectMember>;
+}
+
 /// A convenient data structure that returns the three arguments of the [React.createElement] call
 ///
 ///[React.createElement]: https://reactjs.org/docs/react-api.html#createelement
@@ -19,9 +25,9 @@ pub(crate) struct ReactCreateElementCall {
     pub(crate) children: Option<JsAnyExpression>,
 }
 
-impl ReactCreateElementCall {
+impl ReactApiCall for ReactCreateElementCall {
     /// It scans the current props and returns the property that matches the passed name
-    pub(crate) fn find_prop_by_name(&self, prop_name: &str) -> Option<JsPropertyObjectMember> {
+    fn find_prop_by_name(&self, prop_name: &str) -> Option<JsPropertyObjectMember> {
         self.props.as_ref().and_then(|props| {
             let members = props.members();
             members.into_iter().find_map(|member| {
@@ -60,52 +66,15 @@ impl ReactCreateElementCall {
 ///
 /// The logic of this second case is very similar to the previous one, simply the node that we have
 /// to inspect is different.
-pub(crate) fn is_react_create_element(
+pub(crate) fn as_react_create_element(
     call_expression: &JsCallExpression,
     model: &SemanticModel,
-) -> Option<ReactCreateElementCall> {
+) -> Option<ReactCreateElementCall>
+where
+    ReactCloneElementCall: ReactApiCall,
+{
     let callee = call_expression.callee().ok()?;
-    let is_react_create_element = match callee {
-        JsAnyExpression::JsStaticMemberExpression(node) => {
-            let object = node.object().ok()?;
-            let member = node.member().ok()?;
-            let member = member.as_js_name()?;
-            let identifier = object.as_js_identifier_expression()?.name().ok()?;
-
-            let mut maybe_from_react = identifier.syntax().text_trimmed() == "React"
-                && member.syntax().text_trimmed() == "createElement";
-
-            if let Some(binding_identifier) = model.declaration(&identifier) {
-                let binding_identifier =
-                    JsIdentifierBinding::cast_ref(binding_identifier.syntax())?;
-                if let Some(js_import) = binding_identifier
-                    .syntax()
-                    .ancestors()
-                    .find_map(|ancestor| JsImport::cast_ref(&ancestor))
-                {
-                    maybe_from_react = js_import.source_is("react").ok()?;
-                }
-            }
-            maybe_from_react
-        }
-        JsAnyExpression::JsIdentifierExpression(identifier) => {
-            let name = identifier.name().ok()?;
-            let mut maybe_react = identifier.syntax().text_trimmed() == "createElement";
-            if let Some(identifier_binding) = model.declaration(&name) {
-                let binding_identifier =
-                    JsIdentifierBinding::cast_ref(identifier_binding.syntax())?;
-                if let Some(js_import) = binding_identifier
-                    .syntax()
-                    .ancestors()
-                    .find_map(|ancestor| JsImport::cast_ref(&ancestor))
-                {
-                    maybe_react = js_import.source_is("react").ok()?;
-                }
-            }
-            maybe_react
-        }
-        _ => return None,
-    };
+    let is_react_create_element = is_react_call_api(&callee, model, "createElement")?;
 
     if is_react_create_element {
         let arguments = call_expression.arguments().ok()?.args();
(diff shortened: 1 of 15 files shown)


crates/rome_js_analyze/src/react.rs
```diff
@@ -60,20 +66,190 @@ impl ReactCreateElementCall {
 ///
 /// The logic of this second case is very similar to the previous one, simply the node that we have
 /// to inspect is different.
-pub(crate) fn is_react_create_element(
+pub(crate) fn as_react_create_element(
```
MichaReiser, line 69: I think both the naming `as` and `is` are somewhat misleading because I would expect a function named `as` to return a value by reference which this function does not. 

What do you think of making the method a static function on `ReactCreateElement` similar to the `AstNode` `cast` method:

```
ReactCloneElementCall::from_call_expression(call_expression, model)
```

crates/rome_js_analyze/src/react.rs
```diff
@@ -60,20 +66,190 @@ impl ReactCreateElementCall {
...
     call_expression: &JsCallExpression,
     model: &SemanticModel,
-) -> Option<ReactCreateElementCall> {
+) -> Option<ReactCreateElementCall>
+where
+    ReactCloneElementCall: ReactApiCall,
```
MichaReiser, line 74: Why is the `where` close constraining the `ReactCreateElementCall` necessary?

crates/rome_js_analyze/src/react.rs
```diff
@@ -60,20 +66,190 @@ impl ReactCreateElementCall {
...
     let callee = call_expression.callee().ok()?;
-    let is_react_create_element = match callee {
+    let is_react_create_element = is_react_call_api(&callee, model, "createElement")?;
+
+    if is_react_create_element {
+        let arguments = call_expression.arguments().ok()?.args();
+        // React.createElement() should not be processed
+        if !arguments.is_empty() {
+            let mut iter = arguments.into_iter();
+            // SAFETY: protected by the `is_empty` check
```
MichaReiser, line 84: Nit, you can avoid the `unwrap` by doing:

```
let mut iter = arguments.iter();

if let Some(first) = iter.next() {
    
}

```

or you can even do

```
match (iter.next(), iter.next(), iter.next()) {
    (Some(Ok(first)), Some(Ok(JsAnyCallArgument::JsAnyExpression(JsObjectExpression(object))), Some(Ok(JsAnyCallArgument::JsAnyExpression(children))))) => {

    },
    (Some(Ok(first)), Some(Ok(JsAnyCallArgument::JsAnyExpression(JsObjectExpression(object))), None) => {

    },
    _ => None
}
```

crates/rome_js_analyze/src/react.rs
```diff
@@ -60,20 +66,190 @@ impl ReactCreateElementCall {
...
+    /// Optional children
+    #[allow(dead_code)]
+    pub(crate) children: Option<JsAnyExpression>,
+}
+
+impl ReactApiCall for ReactCloneElementCall {
+    fn find_prop_by_name(&self, prop_name: &str) -> Option<JsPropertyObjectMember> {
+        self.new_props.as_ref().and_then(|props| {
+            let members = props.members();
+            members.into_iter().find_map(|member| {
```
MichaReiser, line 131: Nit: You can use `iter()` or `into_iter` interchangably (the former doesn't consume `self`)

crates/rome_js_analyze/src/react.rs
```diff
@@ -60,20 +66,190 @@ impl ReactCreateElementCall {
...
+///
+/// Second case
+///
+/// ```js
+/// cloneElement()
+/// ```
+///
+/// The logic of this second case is very similar to the previous one, simply the node that we have
+/// to inspect is different.
+pub(crate) fn as_react_clone_element(
```
MichaReiser, line 167: Nit: Same as for `ReactCreateElementCall`: I would prefer to have it as a static method on the `ReactCloneElement`.

crates/rome_js_analyze/src/react.rs
```diff
@@ -60,20 +66,190 @@ impl ReactCreateElementCall {
...
+) -> Option<ReactCloneElementCall>
+where
+    ReactCloneElementCall: ReactApiCall,
+{
+    let callee = call_expression.callee().ok()?;
+    let is_react_clone_element = is_react_call_api(&callee, model, "cloneElement")?;
+
+    if is_react_clone_element {
+        let arguments = call_expression.arguments().ok()?.args();
+        // React.cloneElement() should not be processed
```
MichaReiser, line 179: Nit: Same as for `ReactCreateElement`: You can use `match` here and entirely avoid `unwrap` and many of the `ok` and `clone` calls.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,340 @@
...
+use rome_js_syntax::{
+    JsArrowFunctionExpression, JsCallExpression, JsExpressionStatement, JsFunctionDeclaration,
+    JsFunctionExpression, JsIdentifierBinding, JsIdentifierExpression, JsMethodClassMember,
+    JsMethodObjectMember, JsParameterList, JsPropertyObjectMember, JsReferenceIdentifier,
+    JsxAttribute, JsxOpeningElement, JsxSelfClosingElement,
+};
+use rome_rowan::{declare_node_union, AstNode, AstSeparatedList};
+
+declare_rule! {
+    /// Prevent the usage of Array index in keys
```
MichaReiser, line 16: Nit: `Discourage` the usage of Array indices as React keys.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,340 @@
...
+        recommended: false,
+    }
+}
+
+declare_node_union! {
+    pub(crate) NoArrayIndexKeyQuery = JsxOpeningElement | JsxSelfClosingElement | JsCallExpression
+}
+
+declare_node_union! {
+    pub(crate) InvalidProp = JsxAttribute | JsPropertyObjectMember
```
MichaReiser, line 45: Nit: How about `KeyPropWithArrayIndex` or simply `KeyProp` as this will always be the key property?

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,340 @@
...
+
+        let value = reference.value_token().ok()?;
+        let function_parent = node.find_function_like_ancestor()?;
+        let key_value = value.text_trimmed();
+
+        let is_inside_array_method = is_inside_array_method(&function_parent)?;
+
+        if is_inside_array_method {
+            if node.is_call_expression() {
+                // 1. We scan the parent and look a potential `React.Children` method call
```
MichaReiser, line 133: ...look for *a* potential...

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn diagnostic(ctx: &RuleContext<Self>, state: &Self::State) -> Option<RuleDiagnostic> {
+        let _node = ctx.query();
+        let NoArrayIndexKeyState {
+            binding_origin: incorrect_key,
+            incorrect_prop,
+        } = state;
+        Some(
+            RuleDiagnostic::new(
+                rule_category!(),
+                incorrect_prop.syntax().text_trimmed_range(),
```
MichaReiser, line 167: I don't know if it is possible to use @leops new diagnostics yet but I think it would be great to explain why using an array index is discouraged.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,340 @@
...
+                markup! {"Avoid using array index as key property in an element."},
```
MichaReiser, line 168: Nit: *an*? array index as key property *of*? an element

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,340 @@
...
+            )
+            .secondary(
+                incorrect_key.syntax().text_trimmed_range(),
+                markup! {"This is where the key comes from."},
```
MichaReiser, line 172: Nit: This is the source of the key value.

MichaReiser (approved):

crates/rome_js_analyze/src/react.rs
```diff
@@ -19,9 +25,9 @@ pub(crate) struct ReactCreateElementCall {
...
 }
 
-impl ReactCreateElementCall {
+impl ReactApiCall for ReactCreateElementCall {
     /// It scans the current props and returns the property that matches the passed name
-    pub(crate) fn find_prop_by_name(&self, prop_name: &str) -> Option<JsPropertyObjectMember> {
+    fn find_prop_by_name(&self, prop_name: &str) -> Option<JsPropertyObjectMember> {
         self.props.as_ref().and_then(|props| {
             let members = props.members();
             members.into_iter().find_map(|member| {
```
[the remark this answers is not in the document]
ematipico: Already [done](https://github.com/rome/tools/blob/main/crates/rome_js_syntax/src/jsx_ext.rs#L80) and [done](https://github.com/rome/tools/blob/main/crates/rome_js_syntax/src/jsx_ext.rs#L131) :)

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
ematipico: I reworded it to `"Avoid using the index of an array as key property in an element"`

crates/rome_js_analyze/src/react.rs
ematipico: Not needed. Removed

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
ematipico: I added two footer notes, which are a small explanation of the documentation (which has been updated too)

crates/rome_js_analyze/src/react.rs
```diff
@@ -40,40 +112,148 @@ impl ReactCreateElementCall {
...
+pub(crate) fn is_react_call_api(
+    expression: &JsAnyExpression,
     model: &SemanticModel,
-) -> Option<ReactCreateElementCall> {
-    let callee = call_expression.callee().ok()?;
-    let is_react_create_element = match callee {
+    api_name: &str,
+) -> Option<bool> {
+    // we bail straight away if the API doesn't exists in React
+    if !VALID_REACT_API.contains(&api_name) {
```
leops, line 245: This could be a `debug_assert!`, this way the check isn't performed on release builds, and panics straight away on debug builds if the provided name isn't valid instead of silently failing

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+}
+
+impl NoArrayIndexKeyQuery {
+    const fn is_call_expression(&self) -> bool {
+        matches!(self, NoArrayIndexKeyQuery::JsCallExpression(_))
+    }
+
+    fn find_attribute_with_key_name(&self, model: &SemanticModel) -> Option<KeyPropWithArrayIndex> {
```
leops, line 85: Nit: `find_key_attribute` might be easier to understand at a glance

leops (approved):

PR #3306, opened 2022-09-30
refactor(rome_diagnostics): refactor the display of diffs to how it worked in Rome JS

leops:
## Summary

This PR brings back the rendering of diffs from Rome JS, and should be the last major change required to align the display of diagnostics in the Rust codebase with the previous JS incarnation.

For the rendering itself the code it mostly a direct adaptation of the existing JS logic for Rust. The only major difference is how one-line diffs are rendered (those are diffs that either perform only insertions or only deletions on a single line, this is different from what the code calls "single line diffs" which is when the whole file only spans a single line and we don't need to print line number). With the old logic those diffs would have been printed like this:

```
    2   │ - ····<>foo</>
      2 │ + ····foo
```

Now such a diff will be printed on a single line, highlighting only the parts the text that were added or removed, similarly to how Rustc displays these kinds of suggestions:

```
    2 │ ····<>foo</>
      │     --   ---
```

Internally, this PR changes how diffs are represented, from a sequence of insert-delete operations (replace a range of the original text with a new string), to a sequence string slices annotated with a "change tag" to know whether they belong to the old revision of the text, the new revision or both versions. Similarly to Rome JS, the Rust version also implements "compressed diffs" where large ranges of text that remained unchanged between the two versions are stored as a simple unchanged line count instead of storing the whole string. Additionally, in order to avoid allocating many short strings in each diff operation, all the string slices that make up a diff are stored in a single "dictionary" string that also de-duplicates identical substrings.

Finally, I also removed a number of unused ways of displaying suggestions in the legacy diagnostics, keeping only the diff display that's actually used (all the other rendering modes can still be implemented using the new diagnostics API)

## Test Plan

Besides updating 75 snapshot files, I've added a few test cases for the generation of compressed diffs and their conversions to LSP `TextEdit` operations


crates/rome_diagnostics/src/v2/display/diff.rs
```diff
@@ -0,0 +1,696 @@
...
+    let mut after_line = OneIndexed::MIN;
+
+    for (i, op) in diff.iter().enumerate() {
+        let op = match op {
+            CompressedOp::DiffOp(op) => op,
+            CompressedOp::EqualLines { line_count } => {
+                let is_first_op = i == 0;
+                for i in 0..=line_count.get() {
+                    // Don't increment the first line if we are the first tuple marking the beginning of the file
+                    if !(is_first_op && i == 0) {
```
MichaReiser, line 31:
```suggestion
                    if !is_first_op {
```

because `is_first_op` already asserts that `i == 0`

MichaReiser: ... ah no, this are different `i`. I think it would help readability to either rename the inner or outer `i` even if it's just naming them `i`, and `j`

crates/rome_diagnostics/src/v2/display/diff.rs
```diff
@@ -0,0 +1,696 @@
...
+                &mut inserted_lines,
+                &mut before_line_to_after,
+                before_line,
+                after_line,
+                tag,
+                new_line,
+            );
+        }
+    }
+
```
MichaReiser, line 120: This is a chunky method. Do you see a way how you could split into 1-2 sub-functions (One building up the data structures and another that is concerned with rendering?)

crates/rome_js_analyze/tests/specs/correctness/inlineVariable.js.snap
```diff
@@ -28,72 +28,72 @@ statement(variable);
...
-2 2 |   
-3   | - if (inlinable) {
-4   | -     notInlinable = inlinable;
-  3 | + if ("value1") {
-  4 | +     notInlinable = "value1";
-5 5 |   }
-6 6 |   
-7 7 |   statement(notInlinable);
+```diff
+@@ -1,8 +1,8 @@
```
MichaReiser, line 32: What's the reason that we're hiding the line numbers here. Is it because these are just snapshots of the diff and not printed diagnostics?

crates/rome_js_analyze/tests/specs/correctness/noCompareNegZero.js.snap
```diff
@@ -73,16 +73,8 @@ noCompareNegZero.js:32:1 lint/correctness/noCompareNegZero  FIXABLE  ━━━
...
-  29 29 |   x === +0;
-  30 30 |   // invalid
-  31    | - x === -0;
-     31 | + x === 0;
-  32 32 |   x == -0;
-  33 33 |   -0 == x;
-  34 34 |   x < -0;
-  
+    32 │ x·===·-0;
+       │       -
```
MichaReiser, line 77: I starred at this for a rather long time to notice that the one `-` below the `-` is not some dust on my screen... May it be possible to emphasise the `-` more in case it's for a single character?

What's the reason that we aren't showing any context lines for this frame?

MichaReiser (approved):

ematipico:
I think the reason why single-line diffs were shown on two lines, was mostly because the logic was easily adaptable for "GitHub diffs", meaning the ability to suggest diffs as GitHub comments on PRs.

I am all about clarity and improvements. Let's just take that into consideration and see if this new logic is adaptable for the GitHub comments

crates/rome_cli/tests/snapshots/main_check/downgrade_severity.snap
```diff
@@ -36,9 +36,8 @@ file.js:1:1 lint/correctness/noDebugger  FIXABLE  ━━━━━━━━━━
   
   i Suggested fix: Remove debugger statement
   
-      | @@ -1 +0,0 @@
-  0   | - debugger;
```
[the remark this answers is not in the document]
leops: The line number starting at 0 was definitely a bug, and it might have been one of the reasons the diff gutter could end up being misaligned

crates/rome_cli/tests/snapshots/main_check/maximum_diagnostics.snap
```diff
@@ -129,13 +117,10 @@ check.js:2:25 lint/correctness/useBlockStatements  FIXABLE  ━━━━━━
...
   i Suggested fix: Wrap the statement with a `JsBlockStatement`
   
-      | @@ -1,5 +1,5 @@
-  0 0 |   
-  1   | - for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);
-    1 | + for(;true;);for(;true;);for(;true;) {}for(;true;);for(;true;);for(;true;);
-  2 2 |   for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);
-  3 3 |   for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);
-  4 4 |   for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);
+     2    │ - for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);for(;true;);
```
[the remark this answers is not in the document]
leops: Yes the first line of the file is not rendered in the context lines because it's empty, this is due to the files used in the snapshot being defined as constant string literals in the test code like this:
```rust
const ERRORS: &str = r#"
for(;true;);for(;true;); ...
"#;
```
Making all those strings start and end with the `\n` character

ematipico (approved):

crates/rome_js_analyze/tests/specs/correctness/inlineVariable.js.snap
leops: The snapshots for code actions used to be rendered using the shared diff printing logic in `rome_console`, but since I moved this code to `rome_diagnostics` it's now no longer easily accessible outside of diagnostics. Instead I've opted to render code actions as actual Git diffs (using the printing logic built-in to the `similar` diffing library), and have the editor or GitHub provide the syntax highlighting or rich display for these as part of the markdown rendering.

crates/rome_js_analyze/tests/specs/correctness/noCompareNegZero.js.snap
leops: In contexts that support rich styling such like the console or the HTML output, both the modified characters and the "diff underline" are emphasized and colored appropriately: https://deploy-preview-3306--rometools.netlify.app/docs/lint/rules/nocomparenegzero/#invalid
Ultimately I think single-character diffs are probably hard to read in most diff rendering mode, but my hope is that this way of printing those diffs makes it at least somewhat easier compared to unified diffs by not requiring the user to mentally parse the same line twice the find the difference, and printing an underline specifically to bring attention to what's above it in contexts that do not support styling.
I also intentionally did not print context lines in this mode, again to help bring the focus on the line being modified and not everything around it. In practice I expect that the line of code being modified will almost always be the same one that was highlighted in the diagnostic's location, and as such it doesn't matter too much that the context isn't printed since it should have been rendered already as part of the diagnostic code frame.

leops:
> I think the reason why single-line diffs were shown on two lines, was mostly because the logic was easily adaptable for "GitHub diffs", meaning the ability to suggest diffs as GitHub comments on PRs.

In practice, due to the diff gutter and the invisible characters printing it's not possible to directly copy the output of a diagnostic into a diff that GitHub understands without some significant editing anyway. But as I noted above in my comment on code action snapshots, the `similar` library already contains all the logic required to print Git-compatible diffs, so maybe that's an alternate rendering mode we could expose through a CLI flag for instance.

PR #3307, opened 2022-09-30
fix(rome_formatter): Printer fill fits

MichaReiser:
## Summary

This fixes the issue why the test is currently failing in #3251 

The specific snipped triggering the issue is:

```javascript
  <div>
    ENddddDSIIIN <div>
      texttexttexttexttexttexttexttexttexttexttexttextextffg
    </div>{" "}
    HRS
  </div>
```

That generates the following IR

```
[
  group(["<div>"]),
  indent([
    hard_line_break,
    fill([
        ["ENddddDSIIIN"],
        [
          if_group_breaks(["{" "}", soft_line_break]),
          if_group_fits_on_line([" "])
        ],
        [
          best_fitting([
            [
              [
                group(["<div>"]),
                "texttexttexttexttexttexttexttexttexttexttexttextextffg",
                "</div>"
              ]
            ]
            [
              [
                group(["<div>"]),
                indent([
                  hard_line_break,
                  fill([
                      ["texttexttexttexttexttexttexttexttexttexttexttextextffg"]
                  ])
                ]),
                hard_line_break,
                "</div>"
              ]
            ]
          ])
        ],
        [
          if_group_breaks(["{" "}", soft_line_break]),
          if_group_fits_on_line([" "])
        ],
        ["HRS"]
      ]
    ])
  ]),
  hard_line_break,
  "</div>;",
  hard_line_break
]
```

The issue is the combination of `fill` with `best_fitting`. 

* Fill tests if `ENddddDSIIIN` fits on the line -> yes. Print the entry
* Fill tests if the space and the `<div>text...</div>` fits on the line -> yes. Print the entry
* The best fitting now tests if its content up to the first hard line break fits on the line. This is includes the following content: `<div>text...</div> HRS`. Notice how it includes the `HRS` which is from the next `fill` entry.

The measuring if the best fitting should stop at the next separator because `fill` will make that separator expand if it otherwise risks exceeding the line width. Now, stopping at the next separator isn't trivial. However, there's no need for `best_fitting` to even test if it fits because `fill` already tested that. 

That's why the fix is somewhat simple. All that is necessary is to set that we know it will fit and that the printer can simply use the first variant.

## But what about...

if the fill element doesn't fit. Do we not have the same problem then? Not really, because the element gets printed in expanded mode and this also applies for the separator coming right after. Since the separator **must** contain a soft line break the printer either decides that the content fits (for best fitting) or expands (group). 

## Test Plan

Tested that  #3251 now formats correctly

### DIFF
--- a/crates/rome_js_formatter/src/jsx/tag/closing_element.rs
+++ b/crates/rome_js_formatter/src/jsx/tag/closing_element.rs
@@ -1,6 +1,6 @@
 use crate::prelude::*;
 
-use rome_formatter::{format_args, write};
+use rome_formatter::write;
 use rome_js_syntax::{JsxClosingElement, JsxClosingElementFields};
 
 #[derive(Debug, Clone, Default)]
@@ -21,13 +21,13 @@ impl FormatNodeRule<JsxClosingElement> for FormatJsxClosingElement {
 
         write![
             formatter,
-            [group(&format_args![
+            [
                 l_angle_token.format(),
                 slash_token.format(),
                 name.format(),
                 line_suffix_boundary(),
                 r_angle_token.format(),
-            ])]
+            ]
         ]
     }
 }
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -986,7 +986,7 @@ fn fits_element_on_line<'a, 'rest>(
                 Some(group_id) => state
                     .group_modes
                     .get_print_mode(group_id)
-                    .unwrap_or_else(|| PrintMode::Flat),
+                    .unwrap_or_else(|| args.mode()),
             };
 
             if group_mode != condition.mode {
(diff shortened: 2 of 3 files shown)


crates/rome_js_formatter/src/jsx/tag/closing_element.rs
```diff
@@ -21,13 +21,13 @@ impl FormatNodeRule<JsxClosingElement> for FormatJsxClosingElement {
 
         write![
             formatter,
-            [group(&format_args![
+            [
```
MichaReiser, line 24: These two changes aren't related to the bug but align our IR with Prettiers and made it easier to debug the issue

MichaReiser:
@denbezrukov this should fix the failing test in #3251 . @ematipico please go ahead with merging this PR and then #3251 on Monday (I'll check in on Saturday and merge the changes myself if they are approved but otherwise have to delegate, I'm sorry)

MichaReiser:
!bench_formatter

PR #3311, opened 2022-10-03
docs: better contribution write-up for lint rules

ematipico:
## Summary

This PR adds a better write-up to explain how to implement new rules.

## Test Plan

Unfortunately there's not preview for the technical documentation. Please just do a proof-read


crates/rome_analyze/CONTRIBUTING.md
```diff
@@ -0,0 +1,276 @@
...
+
+# Folder structure
+
+First of all, you need to identify the crate where you want to implement the rule. If the rule
+is going to be implemented for the JavaScript language (and its super languages), then the rule
+will be implemented inside the `rome_js_analyzer` crate.
+
+Rules are divided by capabilities:
+- `analyzers/` folder contains rules that don't require any particular capabilities;
+- `semantic_analyzer/` folder contains rules that require the use of the semantic model
```
leops, line 15: There's also the `assists/` folder containing "assist rules" that contribute refactor code actions with no associated diagnostics

crates/rome_analyze/CONTRIBUTING.md
```diff
@@ -0,0 +1,276 @@
...
+the CI will fail.
+
+### Create your first rule
+
+Let's say we want to create a new rule called `useAwesomeTricks`, which uses the semantic model.
+
+1. create a new file under `semantic_analyzers/nursery` called `use_awesome_tricks`;
+2. run the cargo alias `cargo codegen analyzer`, this command will update the file called `nursery.rs`
+inside the `semantic_analyzers` folder
+3. from there, use the [`declare_rule`](#declare_rule) macro to create a new struct
```
leops, line 37: Nit: Use "type" here instead of "struct", rules are technically enums and not structs, but that's an implementation detail of the `declare_rule!` macro that should not be relied upon anyway

crates/rome_analyze/CONTRIBUTING.md
```diff
@@ -0,0 +1,276 @@
...
+   impl Rule for UseAwesomeTricks {
+        const CATEGORY: RuleCategory = RuleCategory::Lint;
+        type Query = Semantic<JsAnyExpression>;
+        type State = String;
+        type Signals = Option<Self::State>;
+   
+        fn run(ctx: &RuleContext<Self>) -> Self::Signals {}
+   }
+   ```
+5. the const `CATEGORY` must be `RuleCategory::Lint` otherwise it won't work
```
leops, line 70: I figured out a way of having this value set automatically by the `declare_rule!` macro while I was implementing static diagnostic categories, so this may not have to remain in the long term

crates/rome_analyze/CONTRIBUTING.md
```diff
@@ -0,0 +1,276 @@
...
+needs to be emitted
+9. Implement the optional `diagnostic` function, to tell the user where's the error and why:
+   ```rust,ignore
+   impl Rule for UseAwesomeTricks {
+        // .. code
+        fn diagnostic(_ctx: &RuleContext<Self>, _state: &Self::State) -> Option<RuleDiagnostic> {}
+   }
+   ```
+   While implementing the diagnostic, please keep [Rome's technical principals](https://rome.tools/#technical) in mind
+10. Implement the optional `action` function, if we are able to provide automatic code fix to the rule:
```
leops, line 85: We should mention that code actions emitted by lint rules must have the `QuickFix` category (I plan to add an error for this in debug mode soon), and maybe explain the different applicability levels a code action may have (or at least mention that new unstable / nursery rules should probably use the `MaybeIncorrect` applicability)

crates/rome_analyze/src/rule.rs
```diff
@@ -49,126 +49,8 @@ pub trait RuleMeta {
...
-///     fn diagnostic(ctx: &RuleContext<Self>, _state: &Self::State) -> Option<RuleDiagnostic> {
-///         Some(RuleDiagnostic::new(
-///             rule_category!(),
-///             ctx.query().text_trimmed_range(),
-///             "message",
-///         ))
-///     }
-/// }
-/// ```
+/// Check [crate](module documentation) for a better
```
leops, line 52: We could at least leave the first few introduction lines of the documentation in the code before linking to the module documentation

crates/rome_analyze/CONTRIBUTING.md
ematipico: Should we highlight that these rules are mainly used inside editors/IDEs?

crates/rome_analyze/CONTRIBUTING.md
ematipico: We can update this line together with the PR that applies this change. For now I'd like to keep it, because it took me a while that I had to change this info in order to trigger the rule

leops (approved):

Issue #3313, opened 2022-10-03
`noDangerWithChildren`


ematipico:
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/react/noDangerWithChildren.md

PR #3314, opened 2022-10-03
feat(rome_js_analyzer): rule `noDangerouslySetInnerHtmlWithChildren`

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3313

## Test Plan

Added test cases, mostly taken from Rome classic codebase


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_dangerously_set_inner_html_with_children.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// ```
+    pub(crate) NoDangerouslySetInnerHtmlWithChildren {
+        version: "0.10.0",
+        name: "noDangerouslySetInnerHtmlWithChildren",
+        recommended: false,
+    }
+}
+
+declare_node_union! {
+    pub(crate) DangerouseProp = JsxAttribute | JsPropertyObjectMember
```
leops, line 45: Nit: there's a typo in the name, it should be `DangerousProp`

leops (commented):
How does this rule interact with `noDangerouslySetInnerHtml` ? Does enabling both rules cause 2 diagnostics to be emitted for the same property ?

ematipico:
> How does this rule interact with `noDangerouslySetInnerHtml` ? Does enabling both rules cause 2 diagnostics to be emitted for the same property ?

I just tested it locally. The analyzer doesn't trigger two diagnostics, it first emits the diagnostic for `noDangerouslySetInnerHtml`. Then, if suppressed, it correctly triggers `noDangerouslySetInnerHtmlWithChildren`

xunilrj (approved):

PR #3315, opened 2022-10-03
feat(rome_diagnostics): refactor some existing diagnostics to use the new Diagnostic trait

leops:
## Summary

This PR holds a number of miscellaneous changes related to exposing the new diagnostic system to the node.js API.
The central change to enable this is replacing the `rome_diagnostics::Diagnostic` type in the result of the `pull_diagnostics` Workspace method with the new `rome_diagnostics::v2::serde::Diagnostic` type, the serializable form of the new `Diagnostic` trait.
In order to make this work I had to fix a few bugs around the serialization and lifetime of diagnostic categories.
I also updated the logic that turns Rome diagnostics into LSP diagnostics to be generic over the Diagnostic trait, and changed the error reporting in the CLI filesystem traversal logic to use the new `Error` type provided by `rome_diagnostics` instead of declaring its own `TraversalError`.

## Test Plan

The changes in the filesystem traversal should be covered by the CLI snapshot tests, and I've added an additional test to check how diagnostics are encoded to the LSP

### DIFF
--- a/crates/rome_lsp/tests/server.rs
+++ b/crates/rome_lsp/tests/server.rs
@@ -216,6 +216,9 @@ impl Server {
     }
 }
 
+/// Number of notifications buffered by the server-to-client channel before it starts blocking the current task
+const CHANNEL_BUFFER_SIZE: usize = 8;
+
 #[derive(Debug, PartialEq, Eq)]
 enum ServerNotification {
     PublishDiagnostics(PublishDiagnosticsParams),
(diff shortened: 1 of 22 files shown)


crates/rome_lsp/tests/server.rs
```diff
@@ -249,7 +276,8 @@ async fn basic_lifecycle() -> Result<()> {
     let (stream, sink) = client.split();
     let mut server = Server::new(service);
 
-    let reader = tokio::spawn(client_handler(stream, sink));
+    let (sender, _) = channel(8);
```
ematipico, line 279: nit: we could assign `8` to a `const`, so we can document it (having just 8 doesn't help much), and use it everywhere it's needed (I noticed different places)

ematipico (approved):

PR #3320, opened 2022-10-04
perf(rome_js_analyze): improve the performance of noUndeclaredVariables

leops:
## Summary

This PR improves the performance of the `noUndeclaredVariables` rule by deriving the `Queryable` trait on the `SemanticServices` type. This allows individual rules to query the `SemanticModel` as a whole and "manually" traverse it to emit a sequence of signals, instead of having to match on individual AST nodes. This is particularly useful for `noUndeclaredVariables`, since it's a lot cheaper to iterate on the list of all the unresolved references in the semantic model than to match on all the identifier nodes in the syntax tree.

Side note: while it may seem useful to use this new capability to query the semantic model for other lint rules, it is primarily intended as an advanced feature for optimizations and as such should only be used if you know what you're doing. I'm currently considering a way to implement `Queryable` on `rome_js_semantic` types such as `Reference`, `Scope` or `Binding`, and this would allow rules to directly express queries on semantic model concepts without have to implement the traversal and search logic themselves.

## Test Plan

This is an infrastructure only change, it should not impact the results of the `noUndeclaredVariables` rule (or any other rules)

### DIFF
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -5,7 +5,7 @@ use rome_js_syntax::{
 use rome_rowan::{AstNode, SyntaxNode, SyntaxTokenText};
 use rust_lapper::{Interval, Lapper};
 use std::{
-    collections::{hash_set, BTreeSet, HashMap, HashSet},
+    collections::{BTreeSet, HashMap, HashSet},
     iter::FusedIterator,
     sync::Arc,
 };
@@ -98,7 +98,7 @@ struct SemanticModelData {
     // Maps a declaration range to the range of its "writes"
     declaration_all_writes: HashMap<TextRange, Vec<(ReferenceType, TextRange)>>,
     /// All references that could not be resolved
-    unresolved_references: HashSet<TextRange>,
+    unresolved_references: Vec<TextRange>,
     // All bindings that were exported
     exported: HashSet<TextRange>,
 }
@@ -385,7 +385,7 @@ impl<'a> FusedIterator for ReferencesIter<'a> {}
 
 pub struct UnresolvedReferencesIter<'a> {
     data: &'a SemanticModelData,
-    iter: hash_set::Iter<'a, TextRange>,
+    iter: std::slice::Iter<'a, TextRange>,
 }
 
 impl<'a> Iterator for UnresolvedReferencesIter<'a> {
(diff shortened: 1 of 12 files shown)


leops:
@xunilrj As part of this change, I stored all the unresolved references in the semantic model and exposed a way to iterate on those. Does the implementation I came up with match how you would have envisioned this mechanism to work ?

xunilrj:
> @xunilrj As part of this change, I stored all the unresolved references in the semantic model and exposed a way to iterate on those. Does the implementation I came up with match how you would have envisioned this mechanism to work ?

Any reason it is a ```HashSet<...>```? I would expect a ```Vec```.
If I am raising more than one ```UnresolvedReference``` per range, I would consider it a bug.

xunilrj:
Nice! This opens a lot of room for performance improvements!

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -381,6 +383,29 @@ impl<'a> ExactSizeIterator for ReferencesIter<'a> {
 
 impl<'a> FusedIterator for ReferencesIter<'a> {}
 
+pub struct UnresolvedReferencesIter<'a> {
+    data: &'a SemanticModelData,
+    iter: hash_set::Iter<'a, TextRange>,
+}
+
+impl<'a> Iterator for UnresolvedReferencesIter<'a> {
+    type Item = &'a SyntaxNode<JsLanguage>;
```
xunilrj, line 392: I would also try to return a ```Reference``` instead, to be on par with other functions returning references.

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -428,7 +453,7 @@ impl FusedIterator for ScopeBindingsIter {}
 /// - Declarations: [declaration]
 ///
 /// See [SemanticModelData] for more information about the internals.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
xunilrj, line 456: Do we need this ```Debug``` here? Normally it prints too much info.

leops: I added this because I tried to store the `SemanticModel` inside of the `QueryMatch` enum that requires `Debug` at one point during the refactor, but it's not really needed anymore

xunilrj:
> I wasn't sure if it was "valid" to create a `Reference` object to represent an unresolved reference, for instance if the reference should use the existing read / write types or use a special "unresolved" reference type

We may need to create a new field ```read: bool``` for this event. The event extractor certainly knows if it is a read or a write, it is just not propagating this info.

ematipico (approved):

Issue #3169, opened 2022-09-06
🐛 Typescript Declaration File noUnusedVariable


ZachHaber:
### Environment information

```bash
Rome Extension: v0.14.0
No Rome CLI installed globally or locally,

VSCode Env Info:
Version: 1.71.0 (system setup)
Commit: 784b0177c56c607789f9638da7b6bf3230d47a8c
Date: 2022-09-01T07:36:10.600Z
Electron: 19.0.12
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Windows_NT x64 10.0.19044
Sandboxed: No
```

### What happened?

When viewing a manually created `.d.ts` file, `noUnusedVariables` erroneously triggers.

For reference, this is setting up an environment that's outside the normal browser environment (Adobe Acrobat), so the default typescript lib files don't apply in a lot of cases.

```
declare const ANSB_None: number;

interface MapConstructor {
	new <K, V>(
		iterable?:
			| Iterable<readonly [K, V] | LegacyIterator<readonly [K, V]>>
			| null,
	): Map<K, V>;
}

declare class StopIteration {
	private constructor();
}

```
[...]

PR #3322, opened 2022-10-04
fix(rome_js_analyze): correctly do not flag unused on ts ambient files

xunilrj:
## Summary

Closes https://github.com/rome/tools/issues/3169

## Test Plan

```
> cargo test -p rome_js_analyze -- no_unused
```


crates/rome_js_analyze/src/semantic_analyzers/nursery/no_unused_variables.rs
```diff
@@ -160,7 +161,17 @@ fn is_typescript_unused_ok(binding: &JsIdentifierBinding) -> Option<()> {
         JsSyntaxKind::TS_INDEX_SIGNATURE_PARAMETER => Some(()),
         // example: declare function notUsedParameters(a);
         JsSyntaxKind::TS_DECLARE_FUNCTION_DECLARATION => Some(()),
-        _ => None,
+        _ => {
+            // Anything below a declare
+            if parent
+                .ancestors()
+                .any(|x| matches!(x.kind(), JsSyntaxKind::TS_DECLARE_STATEMENT))
```
ematipico, line 168: nit: this can be written

```rust
	if parent
		.ancestors()
		.any(|x| TsDeclareStatement::can_cast(x.kind())) {}
```

crates/rome_js_analyze/tests/specs/nursery/noUnusedVariables/noUnusedVariables.d.ts
```diff
@@ -0,0 +1,20 @@
+declare const SOMECONSTANT: number;
```
ematipico, line 1: Can change the name of the file to add "valid" in its name? e.g. `noUnusedVariablesValid.d.ts` So we know, somehow, that these cases should not trigger the rule

ematipico (approved):

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_unused_variables.rs
xunilrj: Done.

crates/rome_js_analyze/tests/specs/nursery/noUnusedVariables/noUnusedVariables.d.ts
xunilrj: Done.

Issue #3318, opened 2022-10-04
☂️ Analyzer options


ematipico:
### Description

At the moment the analyzer doesn't have much context about configuration, file type, etc.

It's important that the analyzer is able query these information in the future, this would allow us to create more powerful lint rules:
- lint rules with configuration
- lint rules that can inspect the file 
- read information inside the `rome.json`

Here's a list of tasks that have to be done in order to achieve this level of architecture:

- [x] #3324
- [x] make `AnalyzerOptions` accessible from the `trait` `Rule`
- [x] #3343
- [x] #3344
- [x] #3345
- [x] #3346
- [x] #3392
- [x] #3395

PR #3325, opened 2022-10-04
feat(rome_analyze): accept options inside analyzer

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3324

This PR starts the works needed to implement "options" or "configuration" for the analyzer. The reason why I used the term "option" is because is in line with the formatter terminology.

As highlighted in the [umbrella issue](https://github.com/rome/tools/issues/3318), I decided to split the work in different sages to make the implementation of the feature progressive. There are many changes that we have to do, and I felt that needed to be split in multiple stages.

This PR adds a new `AnalyzerOptions` struct which, for now, will contain two information:
- `globals`, which are derived - for now - by `Configuration::javascript::linter::globals`. In the future this will be computed based on the language
- `rules`, which is derived from the configuration of the rules - the `rome.json` file essentially - that match `RuleSettings`. At the moment the logic that computes the rules is hardcoded. Ideally this will have to be auto generated for each group. I will do such in another PR. `rules` is an `HashMap` mapped like this: "name of the rule" => `Box<RawValue>`. In order to use `RawValue`, I had to add `serde_json` as dependency. I tried to use `Box<impl Any>`, but unfortunately `RawValue` doesn't play well with `Any` because `RawValue` doesn't have a size that the compiler can know at compile time.
- the `analyze` function now needs an `&options` as additional parameter

## Test Plan

The logic of the analyzer slightly changed, but it should only a compile time change and the rest of the existing tests should pass.


crates/rome_analyze/src/lib.rs
```diff
@@ -58,15 +58,110 @@ pub struct Analyzer<'analyzer, L: Language, Matcher, Break> {
...
+
+/// A set of information useful to the analyzer infrastructure
+#[derive(Debug, Clone, Default)]
+pub struct AnalyzerOptions {
+    /// A data structured derived from the [`rome.json`] file
+    configuration: AnalyzerConfiguration,
+}
+
+impl AnalyzerOptions {
+    /// It retrieves the options that belong to a rule, it they exists.
```
xunilrj, line 118: if?

xunilrj (approved):

crates/rome_analyze/src/context.rs
```diff
@@ -13,6 +14,7 @@ where
     query_result: &'a RuleQueryResult<R>,
     root: &'a RuleRoot<R>,
     services: RuleServiceBag<R>,
+    options: &'a AnalyzerOptions,
```
leops, line 17: Maybe we could store the options in the service bag, it could make the change simpler than having to add it as a parameter or field on all the intermediate functions and data structures

crates/rome_analyze/src/lib.rs
```diff
@@ -58,15 +58,110 @@ pub struct Analyzer<'analyzer, L: Language, Matcher, Break> {
     query_matcher: Matcher,
     /// Language-specific suppression comment parsing function
     parse_suppression_comment: SuppressionParser,
-    /// Handles analyzer signals emitted by invidual rules
+    /// Handles analyzer signals emitted by individual rules
     emit_signal: SignalHandler<'analyzer, L, Break>,
 }
 
-pub struct AnalyzerContext<L: Language> {
+/// A convenient new type data structure to store the options that belong to a rule
```
leops, line 65: Let's move all the new options-related types to an `options.rs` module, the `lib.rs` files is already quite large as it is

crates/rome_analyze/src/lib.rs
```diff
@@ -58,15 +58,110 @@ pub struct Analyzer<'analyzer, L: Language, Matcher, Break> {
...
+#[derive(Debug, Clone)]
+pub struct RuleOptions(Box<RawValue>);
+
+impl RuleOptions {
+    /// It returns the [RawValue] for the relative rule
+    #[allow(dead_code)]
+    fn value(&self) -> &RawValue {
```
leops, line 72: This function should be public, otherwise code in `rome_js_analyze` for instance will not be able to access the inner raw value for deserialization

crates/rome_analyze/src/lib.rs
```diff
@@ -58,15 +58,110 @@ pub struct Analyzer<'analyzer, L: Language, Matcher, Break> {
...
+
+    /// Creates a new [RuleOptions]
+    fn new(options: Box<RawValue>) -> Self {
+        Self(options)
+    }
+}
+
+/// A convenient new type data structure to insert and get rules
+#[derive(Debug, Clone, Default)]
+pub struct AnalyzerRules(HashMap<String, RuleOptions>);
```
leops, line 84: Nit: This map might be able to use the `RuleKey` type as a key

crates/rome_analyze/src/lib.rs
```diff
@@ -58,15 +58,110 @@ pub struct Analyzer<'analyzer, L: Language, Matcher, Break> {
...
+    ///     type Query = ();
+    ///     type State = ();
+    ///     type Signals = ();
+    ///
+    ///     fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+    ///         let options = ctx.options();
+    ///     }
+    /// }
+    /// ```
+    pub fn rule_options<F: FnOnce(&RuleOptions) -> ToType, ToType>(
```
leops, line 150: In practice since the `RuleOptions` is a thin wrapper around a `RawValue` the only way to get a value out of it will be through deserialization, so this function could simply require `ToType: Deserialize<'de>` (with the function borrowing self as `&'de self`) and remove the `mapper` argument.

Also for ease of use it would be nice to alias this function on the `RuleContext` with the `rule_name` automatically injected, so that most rules could simply write `ctx.rule_options::<OptionsType>()` and get the appropriate configuration entry back automatically.

ematipico: That's the intended final design, but we are not there yet, and that's why I decided to break down the work in multiple tasks:
- using `Deserialize` requires having `serde` as a *non-optional* dependency, which is not a change we can do lightly (it creates errors)
- we can't get the rule name yet, there are changes that needs to be implemented

crates/rome_analyze/src/context.rs
ematipico: Adding it to the `ServiceBag` involves another series of changes around various traits. Inside `RuleContext`, services has a trait bound to `RuleServiceBag`.

I think we can do this change, but I would prefer to defer it to later for two reasons:
- I am not comfortable with this kind of change 
- options will become a trait itself too - while rules are language agnostic, `globals` are not; so it seems more appropriate to make this change together with this one

leops (approved):

PR #3331, opened 2022-10-04
feat(rome_js_formatter): new print fill entries

denbezrukov:
## Summary

We have a jsx formatting problem with the current `print_fill_entries`.  
The problem is well described in https://github.com/rome/tools/pull/3251#issuecomment-1264447408
This PR tries to implement @MichaReiser [proposal](https://github.com/rome/tools/pull/3251#issuecomment-1264460093)

## Test Plan

Tested that https://github.com/rome/tools/pull/3251 now formats correctly
Screenshot 2022-10-10 at 7 48 19 AM

### DIFF
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -785,9 +785,10 @@ impl<'a, 'print> FitsOnLine<'a, 'print> {
             &self.printer.options,
         );
 
-        if !matches!(self.queue.top(), Some(FormatElement::Tag(Tag::StartEntry)))
-            && !matches!(self.queue.top(), Some(FormatElement::Tag(Tag::EndFill)))
-        {
+        if !matches!(
+            self.queue.top(),
+            Some(FormatElement::Tag(Tag::StartEntry | Tag::EndFill))
+        ) {
             self.queue.skip_content(TagKind::Entry);
         }
 
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -414,14 +414,43 @@ impl<'a> Printer<'a> {
         }
     }
 
+    /// Semantic [print_entry] for the fill item
+    fn print_fill_item(
+        &mut self,
+        queue: &mut PrintQueue<'a>,
+        stack: &mut PrintCallStack,
+        args: PrintElementArgs,
+    ) -> PrintResult<()> {
+        self.print_entry(queue, stack, args)
+    }
+
+    /// Semantic [print_entry] for the separator item
+    fn print_fill_separator(
+        &mut self,
+        queue: &mut PrintQueue<'a>,
+        stack: &mut PrintCallStack,
+        args: PrintElementArgs,
+    ) -> PrintResult<()> {
+        self.print_entry(queue, stack, args)
+    }
+
     /// Tries to fit as much content as possible on a single line.
-    /// Each item forms a virtual group that is either printed in flat or expanded mode.
+    /// Each element forms a virtual group that is either printed in flat or expanded mode.
+    /// Fill content is a sequence of items and separators: [item, separator, item, separator, ...]
     /// It handles different cases:
     ///
-    /// * Try to print a item in flat mode. If it doesn't fit print content and separator in expanded mode.
-    /// * If the item fits in flat mode try to print the separator in flat mode or expanded mode.
-    /// * If the separator fits in flat mode try to print the next item in flat mode,
-    /// * If the next item doesn't fit in flat mode print the separator in expanded mode.
+    /// Assume we have the following sequence: [item, separator, next_item, ...].
+    /// With one iteration of the [print_fill_entries] function, we want to print the item and the separator.
+    /// The question is which [PrintMode] we should use for the item and the separator.
+    /// The best effort is printing the item and the separator in flat mode.
+    ///
+    /// * If the item doesn't fit the line in flat mode, it prints the item and the separator in expanded mode.
+    /// * If the item fits the line in flat mode, it tries to fit the separator in flat or expanded mode.
+    /// * If the separator fits only in expanded, it prints the item in flat mode and the separator in expanded mode.
+    /// * If the separator doesn't fit in any mode, it prints the item and the separator in expanded mode.
+    /// * If the separator fits in flat mode, it tries to fit the next item in flat mode.
+    /// * If the next item doesn't fit in flat mode, it prints the item in flat mode and the separator in expanded mode.
+    /// * If the next item fits in flat mode, it prints the item in flat mode and the separator in flat mode.
     fn print_fill_entries(
         &mut self,
         queue: &mut PrintQueue<'a>,
@@ -437,20 +466,20 @@ impl<'a> Printer<'a> {
         while let Some(FormatElement::Tag(Tag::StartEntry)) = queue.top() {
             let mut fits_on_line = FitsOnLine::new(queue, self);
 
-            let item_fits = fits_on_line.fits_fill_entry(PrintMode::Flat, stack)?;
+            let item_fits = fits_on_line.fits_fill_item(PrintMode::Flat, stack)?;
 
             // Not even the item fits, print item and separator in expanded mode
             if !item_fits {
                 fits_on_line.finish();
 
-                self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                self.print_fill_item(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
 
                 // Check if we have a next item
                 if !matches!(queue.top(), Some(FormatElement::Tag(Tag::StartEntry))) {
                     break;
                 }
 
-                self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                self.print_fill_separator(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
                 continue;
             }
 
@@ -461,12 +490,12 @@ impl<'a> Printer<'a> {
             ) {
                 fits_on_line.finish();
 
-                self.print_entry(queue, stack, args.with_print_mode(PrintMode::Flat))?;
+                self.print_fill_item(queue, stack, args.with_print_mode(PrintMode::Flat))?;
                 break;
             }
 
             let separator_fits_in_flat_mode =
-                fits_on_line.fits_fill_entry(PrintMode::Flat, stack)?;
+                fits_on_line.fits_fill_separator(PrintMode::Flat, stack)?;
 
             // Separator doesn't fit in flat mode
             if !separator_fits_in_flat_mode {
@@ -476,24 +505,32 @@ impl<'a> Printer<'a> {
                 let mut fits_on_line = FitsOnLine::new(queue, self);
 
                 // Skip item, we know it fits
-                let _ = fits_on_line.fits_fill_entry(PrintMode::Flat, stack)?;
+                let _ = fits_on_line.fits_fill_item(PrintMode::Flat, stack)?;
 
                 let separator_fits_in_expanded_mode =
-                    fits_on_line.fits_fill_entry(PrintMode::Expanded, stack)?;
+                    fits_on_line.fits_fill_separator(PrintMode::Expanded, stack)?;
 
                 fits_on_line.finish();
 
                 if separator_fits_in_expanded_mode {
                     // If the separator fits, then print the item in flat mode, but expand the separator
                     self.state.measured_group_fits = true;
-                    self.print_entry(queue, stack, args.with_print_mode(PrintMode::Flat))?;
+                    self.print_fill_item(queue, stack, args.with_print_mode(PrintMode::Flat))?;
 
                     self.state.measured_group_fits = false;
-                    self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                    self.print_fill_separator(
+                        queue,
+                        stack,
+                        args.with_print_mode(PrintMode::Expanded),
+                    )?;
                 } else {
                     // If not, then print both in expanded mode (the best we can do in this situation)
-                    self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
-                    self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                    self.print_fill_item(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                    self.print_fill_separator(
+                        queue,
+                        stack,
+                        args.with_print_mode(PrintMode::Expanded),
+                    )?;
                 }
 
                 continue;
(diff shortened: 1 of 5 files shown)


crates/rome_formatter/src/printer/mod.rs
```diff
@@ -667,6 +668,75 @@ impl GroupModes {
...
+        debug_assert!(saved_queue.is_empty());
+
+        let queue = FitsQueue::new(print_queue, saved_queue);
+        let state = FitsState {
+            pending_indent: printer.state.pending_indent,
+            pending_space: printer.state.pending_space,
+            must_be_flat: true,
+            line_width: printer.state.line_width,
+            has_line_suffix: printer.state.line_suffixes.has_pending(),
+            group_modes: &mut printer.state.group_modes,
```
denbezrukov, line 692: I'm not sure that using `unique` reference is a good idea. Because we use `FitsOnLine` only to check a current entry and mutation printer `group_modes` looks a bad idea.
Maybe we can implement `FitsGroupModes`?
 It keeps shared reference to printer `group_modes` and has an another `group_modes` which will be local for `FitsState`.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +433,102 @@ impl<'a> Printer<'a> {
...
-            // If this was a trailing separator, exit
-            if !matches!(queue.top(), Some(FormatElement::Tag(Tag::StartEntry))) {
-                break;
+            // Separator doesn't fit in flat mode
+            if !separator_fits {
+                // Now test if the separator fits in expanded mode
+                let mut fits_in_line = FitsOnLine::new(queue, self);
+
+                // Skip item, we know it fits
+                let _ = fits_in_line.fits_fill_entry(PrintMode::Flat, stack)?;
```
denbezrukov, line 471: We can use `skip_content` here because it mutates only queue and doesn't mutate `FitsState`.
I was wondering and we could copy `FitsState` before checking `separator` in flat mode. But `FitsState` isn't copy because of `group_modes`

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -659,12 +673,122 @@ impl GroupModes {
...
+        fits.map(|fits| match fits {
+            Fits::Maybe | Fits::Yes => true,
+            Fits::No => false,
         })
     }
+
+    // fn finish(self) {
+    //     self.printer.state.fits_queue = self.queue.finish();
+    //     self.printer.state.fits_queue.clear();
+    // }
```
denbezrukov, line 791: Do we need this? I can't understand that kind of optimisation it was.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -642,7 +657,19 @@ struct PrinterState<'a> {
 #[derive(Debug, Default)]
 struct GroupModes(Vec<Option<PrintMode>>);
 
-impl GroupModes {
```
denbezrukov, line 645: Better naming?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +434,115 @@ impl<'a> Printer<'a> {
...
+
+            // Separator doesn't fit in flat mode
+            if !is_separator_fits_in_flat_mode {
+                fits_on_line.finish();
+
+                // Now test if the separator fits in expanded mode
+                let mut fits_on_line = FitsOnLine::new(queue, self);
+
+                // Skip item, we know it fits
+                let _ = fits_on_line.fits_fill_entry(PrintMode::Flat, stack)?;
```
denbezrukov, line 479: We can use skip_content here because it mutates only queue and doesn't mutate FitsState.
I was wondering and we could copy FitsState before checking separator in flat mode. But FitsState isn't copy because of group_modes

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +434,115 @@ impl<'a> Printer<'a> {
...
+                } else {
+                    // If not, then print both in expanded mode (the best we can do in this situation)
+                    self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                    self.print_entry(queue, stack, args.with_print_mode(PrintMode::Expanded))?;
+                }
+
+                continue;
+            }
+
+            // If we don't have next item print the item and the separator in flat mode
```
ematipico, line 502:
```suggestion
            // If we don't have a next item, print the current item and the current separator in flat mode
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +434,115 @@ impl<'a> Printer<'a> {
...
-                    }),
-                )?;
-
-                current_fits = next_fits;
+                self.print_entry(queue, stack, args.with_print_mode(PrintMode::Flat))?;
+                break;
             }
+
+            // We know here that the item + separator fit. Now the question is if the next item fits as well
+            // Test if the next item fits too
```
ematipico, line 515:
```suggestion
            // We know here that the item and separator both fit. Now the question is if the next item fits as well
            // Test if the next item fits too
```

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +434,115 @@ impl<'a> Printer<'a> {
...
+            let next_fits = fits_on_line.fits_fill_entry(PrintMode::Flat, stack)?;
```
ematipico, line 516: `next_item_fits`, it could be a separator... so it's best to use a better name

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +434,115 @@ impl<'a> Printer<'a> {
...
+            fits_on_line.finish();
+
+            // Print the item in flat mode because we know the flat or expanded separator always fit
+            self.state.measured_group_fits = true;
+            self.print_entry(queue, stack, args.with_print_mode(PrintMode::Flat))?;
```
ematipico, line 521: I am trying to understand this part. In the documentation of the function it's written 

> If the next item doesn't fit in flat mode, it prints the separator in expanded mode.

So now I am confused... what are you printing here? The next item or a separator?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +434,115 @@ impl<'a> Printer<'a> {
...
+
+            // Print the separator in flat or expanded mode, depending if the next item also fits on the line
+            self.state.measured_group_fits = next_fits;
+            self.print_entry(
```
ematipico, line 525: I wonder if we could tweak the APIs a little bit. I managed to understand the flow of the logic thanks to the comments, but if these comments/lines of code shift/change, we lose a lot because we don't know what we are printing anymore.

What if we had two APIs called `self.print_item` and `self.print_separator`, and both will do the same, which is

```rs
fn print_item(queue, stack, args) {
	self.print_entry(queue, stack, args.with_print_mode(PrintMode::Flat))?;
}

fn print_separator(queue, stack, args) {
	self.print_entry(queue, stack, args.with_print_mode(PrintMode::Flat))?;
}
```

As far as I understood, we're are in no position to have a typed item and a typed separator, so maybe these "fake" APIs might help to understand a bit what we're printing. 

What do you think? Would that help? This is just a suggestion

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -642,7 +657,19 @@ struct PrinterState<'a> {
 #[derive(Debug, Default)]
 struct GroupModes(Vec<Option<PrintMode>>);
 
-impl GroupModes {
+trait GroupModesMap: Debug {
+    fn insert_print_mode(&mut self, group_id: GroupId, mode: PrintMode);
```
ematipico, line 661: Can we document this API, please?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -642,7 +657,19 @@ struct PrinterState<'a> {
...
+
+    fn get_print_mode(&self, group_id: GroupId) -> Option<PrintMode>;
```
ematipico, line 663: Can we document this API, please?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -642,7 +657,19 @@ struct PrinterState<'a> {
...
+
+    fn unwrap_print_mode(&self, group_id: GroupId, next_element: &FormatElement) -> PrintMode {
```
ematipico, line 665: Can we document this API, please? Also, let's add a `## Panic` section to explain when this API might panic.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -659,12 +686,126 @@ impl GroupModes {
...
-        self.get_print_mode(group_id).unwrap_or_else(|| {
-            panic!("Expected group with id {group_id:?} to exist but it wasn't present in the document. Ensure that a group with such a document appears in the document before the element {next_element:?}.")
+#[derive(Debug)]
+struct FitsOnLine<'a, 'print> {
+    queue: FitsQueue<'a, 'print>,
+    state: FitsState,
+    printer: &'print mut Printer<'a>,
+    group_modes: GroupModes,
+    bomb: DebugDropBomb,
+}
```
ematipico, line 698: Can we document struct and fields please?

ematipico (commented):
I left some questions around some piece of the documentation, mostly to help me understand the new logic. Once these are addressed, I will take a second look.

Sorry, Micha would help better than me

crates/rome_formatter/src/printer/mod.rs
denbezrukov: If we're here we know that the current item and the separator can be printed in flat mode both. 
Now the question is whether the next item fits as well, we check the next item and if it does, we print **_the current separator_** in flat mode, otherwise in expanded mode.

crates/rome_formatter/src/printer/mod.rs
denbezrukov: Here we check the next item. I guess we can use `next_item_fits_in_flat_mode`?

crates/rome_formatter/src/printer/mod.rs
denbezrukov: I like that! I'll try to add the same method for measuring.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -659,12 +686,126 @@ impl GroupModes {
...
+    }
+}
+
+impl<'a, 'print> FitsOnLine<'a, 'print> {
+    fn new(
+        print_queue: &'print PrintQueue<'a>,
+        printer: &'print mut Printer<'a>,
+    ) -> FitsOnLine<'a, 'print> {
+        let saved_queue = std::mem::take(&mut printer.state.fits_queue);
+        debug_assert!(saved_queue.is_empty());
```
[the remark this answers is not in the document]
denbezrukov: I tried to add, but I don't completely understand this optimisation.
https://github.com/rome/tools/blob/879811c97ab88f500e1322995f787da32e8cb98e/crates/rome_formatter/src/printer/mod.rs#L634-L637
https://github.com/rome/tools/blob/879811c97ab88f500e1322995f787da32e8cb98e/crates/rome_formatter/src/printer/mod.rs#L774-L777
https://github.com/rome/tools/blob/879811c97ab88f500e1322995f787da32e8cb98e/crates/rome_formatter/src/printer/mod.rs#L799-L803

crates/rome_formatter/src/printer/queue.rs
```diff
@@ -335,6 +335,7 @@ impl FitsPredicate for SingleEntryPredicate {
 
                     if *depth == 0 {
                         *self = SingleEntryPredicate::Done;
+                        return Ok(false);
```
MichaReiser, line 338: I know why this is necessary but don't think it's safe to change the value to `false` here without changing the `FitsPredicate`. 

The `FitsPredicate` must return `true` for all elements that must be consumed which should include the end entry. A better fix in my view is to change the `FitsPredicate` to have a `is_end` method that returns `true` if this is the last element that should be processed. This avoids that `all_fits` pops the element right after the `EndEntry` from the queue.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -433,101 +463,126 @@ impl<'a> Printer<'a> {
...
+
+            // Separator doesn't fit in flat mode
+            if !separator_fits_in_flat_mode {
+                fits_on_line.finish();
+
+                // Now test if the separator fits in expanded mode
+                let mut fits_on_line = FitsOnLine::new(queue, self);
+
+                // Skip item, we know it fits
+                let _ = fits_on_line.fits_fill_item(PrintMode::Flat, stack)?;
```
MichaReiser, line 508: Nit: It may be good to wrap this in an `assert` to ensure this always yields true. There is no good reason why it should not but just to be safe.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -659,12 +733,152 @@ impl GroupModes {
...
+            must_be_flat: true,
+            line_width: printer.state.line_width,
+            has_line_suffix: printer.state.line_suffixes.has_pending(),
+        };
+
+        FitsOnLine {
+            queue,
+            state,
+            printer,
+            group_modes: Default::default(),
```
MichaReiser, line 777: It's important that we use the `group_modes` from the printer here because content can reference any group that appears "before" the current content. Meaning, it's possible that some conditional content references a group that has already been printed.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -659,12 +733,152 @@ impl GroupModes {
...
+
+    /// Tests if it's possible to print the content of the queue up to the first hard line break
+    /// or the end of the document on a single line without exceeding the line width.
+    pub fn fits_fill_entry(
+        &mut self,
+        mode: PrintMode,
+        stack: &'print PrintCallStack,
+    ) -> PrintResult<bool> {
+        let saved_stack = std::mem::take(&mut self.printer.state.fits_stack);
+        debug_assert!(saved_stack.is_empty());
```
MichaReiser, line 808: I think it's best to store the `stack` and `queue` as part of the `FitsOnLine` struct because it's crucial that the `queue` and `stack` always stay in sync.

crates/rome_formatter/src/printer/mod.rs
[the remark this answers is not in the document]
MichaReiser: The optimisation aims to avoid the printer allocating a new `Vec` every time it calsl `fits` for the `FitsQueue` and `FitsStack.

The `FitsCallStack` and `FitsQueue` are "two layered" stack:

* Uses a slice and a stack pointer to remember which element to return next from the `PrintQueue`/`PrintCallStack`. This to avoid that the original stack gets modified
* A `Vec` to store any items that are added during `fits`. The optimisation re-uses this vec by storing it on the `printer` to avoid allocating a new one for each `fits` call

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -659,12 +733,152 @@ impl GroupModes {
...
+    }
+}
+
+/// Tracks the mode in which groups with ids are printed.
+/// It inserts the new group mode into the fits group not into printer group.
+#[derive(Debug)]
+struct FitsOnLineGroupModes<'print, 'fits> {
+    printer_group_modes: &'print GroupModes,
+    fits_group_modes: &'fits mut GroupModes,
+}
```
MichaReiser, line 858: I think it's OK to insert into the printer's group modes directly because the states get overridden when printing.

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -808,243 +1024,247 @@ where
...
-    fits_state: &mut FitsState,
-    queue: &mut FitsQueue<'a, 'print>,
-    stack: &mut FitsCallStack<'print>,
-    options: &PrinterOptions,
-) -> PrintResult<Fits>
+struct AllFit<'a, 'fit, 'print, 'groups, 'stack, 'options, P, G> {
+    predicate: P,
+    state: &'fit mut FitsState,
+    queue: &'fit mut FitsQueue<'a, 'print>,
+    group_modes: &'groups mut G,
```
MichaReiser, line 1031: Would it be possible to default `G` to the `FitsGroupModes` or are there cases it uses the printer group modes directly?

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -808,243 +1024,247 @@ where
...
-
-        match fits_element_on_line(element, fits_state, queue, stack, options)? {
-            Fits::Yes => {
-                return Ok(Fits::Yes);
-            }
-            Fits::No => {
-                return Ok(Fits::No);
+    /// Tests if it's possible to print the content of the queue up to the first hard line break
+    /// or the end of the document on a single line without exceeding the line width.
+    fn fits(mut self) -> PrintResult<Fits> {
```
MichaReiser, line 1044: It would be preferrable to take the `Predicate` as a function type parameter rather than a type parameter on the struct because:

* It's only necessary to know the `predicate` in the `fits` function. 
* Adding the type parameter to the `AllFit` has the downside that Rust monomorphses (copies the implementation) of all struct methods for every `P`. This is necessary for the `fits` function (except if we use `&dyn Predicate but that comes at a performance cost) but is something that we can avoid for `fits_element_on_line` that doesn't depend on the predicate all together.

MichaReiser (commented):
Oh man.... I'm very sorry. I read your message that you wanted to work on trailing separator and then didn't check GitHub again. I then thought on Saturday, let's fix this printer bug and started working on it and now we have two implementations. https://github.com/rome/tools/pull/3374/files

We can go either with your PR or mine. Let me know what you prefer. Two improvements that I think would be worthwhile to take over from my PR are:

1. Using an enum for the different fill layouts: https://github.com/rome/tools/blob/27fe822c2be7aab66d02d4d3ba87d47147335068/crates/rome_formatter/src/printer/mod.rs#L631-L645
2. Optimise for the case where `item`, `separator`, and `item` fit by avoiding to measure the last item twice: Once when deciding on the layout for the first pair and once when deciding on the layout for the second pair: https://github.com/rome/tools/blob/27fe822c2be7aab66d02d4d3ba87d47147335068/crates/rome_formatter/src/printer/mod.rs#L451-L492

PR #3332, opened 2022-10-05
fix(rome_fs): improve the messaging of the "unhandled file type" diagnostic

### LINKED ISSUE #3329: 🐛 Confusing error when running `rome check` with a directory
When I run `rome check .` I expected it to use the current directory and only lint valid files. However what happens is that it ends up erroring on the first file that we don't support. eg. I get the output:

```
./infra/terraform/.terraform/providers/registry.terraform.io/go-gandi/gandi/2.0.1/darwin_arm64 internalError/fs

  ✖ unhandled file type


Checked 0 files in 13ms
```

This isn't very helpful and it would be nice if when exploding a directory pattern on the CLI we filter it to supported files only.

I labeled this as a bug as this feels like some extraordinarily unexpected behaviour.

leops:
## Summary

Fixes #3329

This PR changes the rendering of the `UnhandledDiagnostic` diagnostic type, emitted by the `OsFileSystem` when an unknown file type is encountered. It adds an explicit error message and description, along with an information advices providing additional details. It also adds a specialized version of the error message when the unhandled file is a symbolic link, as that's the most common case for this error to be raised. Finally, it downgrades the severity of the diagnostic from error to warning, since this is not a "fatal" error that should cause the traversal to abort.

The diagnostic should now be printed as follow when encountering a symbolic link:
```text
..\test-link\link internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ! Symbolic links are not supported

  i Rome does not currently support visiting the content of symbolic links since it could lead to an infinite traversal loop
```

In the less likely case that the CLI encounters a completely unknown file type, the following message is displayed instead:
```text
..\test-link\unknown internalError/fs ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ! Encountered an unknown file type

  i Rome encountered a file system entry that's neither a file, directory or symbolic link
```

## Test Plan

This diagnostic is emitted by low-level code in response to an event from the underlying file system, so it's unfortunately hard to reliably test and reproduce across platforms.

### DIFF
--- a/crates/rome_cli/tests/main.rs
+++ b/crates/rome_cli/tests/main.rs
@@ -10,7 +10,7 @@ use std::{ffi::OsString, path::Path};
 use pico_args::Arguments;
 use rome_cli::{CliSession, Termination};
 use rome_console::{BufferConsole, Console};
-use rome_fs::{FileSystem, MemoryFileSystem};
+use rome_fs::{ErrorEntry, FileSystem, MemoryFileSystem};
 use rome_service::{App, DynRef};
 
 const UNFORMATTED: &str = "  statement(  )  ";
@@ -104,6 +104,8 @@ const CUSTOM_CONFIGURATION_AFTER: &str = "function f() {
 ";
 
 mod check {
+    use std::path::PathBuf;
+
     use super::*;
     use crate::configs::{
         CONFIG_LINTER_DISABLED, CONFIG_LINTER_DOWNGRADE_DIAGNOSTIC, CONFIG_LINTER_IGNORED_FILES,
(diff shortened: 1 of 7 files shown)


ematipico (approved):
Would it be possible to add some test that catches these cases? I am not sure if we can simulate a symbolic link, but we can definitely simulate the unknown file type

leops:
> Would it be possible to add some test that catches these cases? I am not sure if we can simulate a symbolic link, but we can definitely simulate the unknown file type

The issue is that this diagnostic is emitted by the `OsFileSystem` only, while the tests use the `MemoryFileSystem` that has no concept of symbolic links or unknown file in the first place, but I'll try come up with an implementation for that.

ematipico (approved):

PR #3333, opened 2022-10-05
feat(rome_js_analyzer): rule `noAutofocus`

### LINKED ISSUE #3297: `noAutoFocus`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/noAutofocus.md

jk-gan:
## Summary

Closes #3297 

## Test Plan

Added test cases to test the situations according to [this docs](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/noAutofocus.md).

Screenshot 2022-10-06 at 10 05 50 PM
Screenshot 2022-10-06 at 10 05 56 PM
Screenshot 2022-10-06 at 10 06 02 PM

### DIFF
--- a/crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
+++ b/crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
@@ -1,0 +1,119 @@
+use crate::JsRuleAction;
+use rome_analyze::{context::RuleContext, declare_rule, ActionCategory, Ast, Rule, RuleDiagnostic};
+use rome_console::markup;
+use rome_diagnostics::Applicability;
+use rome_js_syntax::{JsxAttribute, JsxSelfClosingElement, JsxSelfClosingElementFields};
+use rome_rowan::{AstNode, BatchMutationExt};
+
+declare_rule! {
+    /// Avoid the `autoFocus` attribute
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```jsx,expect_diagnostic
+    /// <input autoFocus />
+    /// ```
+    ///
+    /// ```jsx,expect_diagnostic
+    /// <input autoFocus="true" />
+    /// ```
+    ///
+    /// ```jsx,expect_diagnostic
+    /// <input autoFocus={"false"} />
+    /// ```
+    ///
+    /// ### Valid
+    ///
+    /// ```jsx
+    /// <input />
+    ///```
+    ///
+    /// ```jsx
+    /// <input autoFocus={undefined} />
+    ///```
+    ///
+    /// ```jsx
+    /// <Input autoFocus={"false"} />
+    ///```
+    pub(crate) NoAutoFocus {
+        version: "10.0.0",
+        name: "noAutofocus",
+        recommended: false,
+    }
+}
+
+impl Rule for NoAutoFocus {
+    type Query = Ast<JsxSelfClosingElement>;
+    type State = JsxAttribute;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let node = ctx.query();
+        let JsxSelfClosingElementFields {
+            l_angle_token: _,
+            name,
+            type_arguments: _,
+            attributes: _,
+            slash_token: _,
+            r_angle_token: _,
+        } = node.as_fields();
+
+        if name.ok()?.text().trim() == "input" {
+            let attribute = node.find_attribute_by_name("autoFocus").ok()??;
+            match attribute.initializer() {
+                Some(initializer) => match initializer.value().ok() {
+                    Some(value) => {
+                        let value = match value.as_jsx_expression_attribute_value() {
+                            Some(value) => value,
+                            None => return Some(attribute),
+                        };
+                        let value = match value.expression().ok() {
+                            Some(value) => value.text(),
+                            None => return Some(attribute),
+                        };
+                        if value.trim() == "undefined" {
+                            return None;
+                        } else {
+                            return Some(attribute);
+                        }
+                    }
+                    None => {
+                        return Some(attribute);
+                    }
+                },
+                None => {
+                    return Some(attribute);
+                }
+            }
+        }
+
+        None
+    }
+
+    fn diagnostic(_ctx: &RuleContext<Self>, attr: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic::new(
+            rule_category!(),
+            attr.range(),
+            markup! {
+                "Avoid the "<Emphasis>"autoFocus"</Emphasis>" attribute."
+            }
+            .to_owned(),
+        ))
+    }
+
+    fn action(ctx: &RuleContext<Self>, attr: &Self::State) -> Option<JsRuleAction> {
+        let mut mutation = ctx.root().begin();
+
+        mutation.remove_node(attr.to_owned());
+
+        Some(JsRuleAction {
+            category: ActionCategory::QuickFix,
+            applicability: Applicability::MaybeIncorrect,
+            message: markup! { "Remove the "<Emphasis>"autoFocus"</Emphasis>" attribute." }
+                .to_owned(),
+            mutation,
+        })
+    }
+}
--- a/crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
+++ b/crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
@@ -71,7 +71,7 @@ impl Rule for NoAutoFocus {
     fn diagnostic(_ctx: &RuleContext<Self>, attr: &Self::State) -> Option<RuleDiagnostic> {
         Some(RuleDiagnostic::new(
             rule_category!(),
-            attr.range(),
+            attr.syntax().text_trimmed_range(),
             markup! {
                 "Avoid the "<Emphasis>"autoFocus"</Emphasis>" attribute."
             },
--- a/crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
+++ b/crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
@@ -2,9 +2,7 @@ use crate::JsRuleAction;
 use rome_analyze::{context::RuleContext, declare_rule, ActionCategory, Ast, Rule, RuleDiagnostic};
 use rome_console::markup;
 use rome_diagnostics::Applicability;
-use rome_js_syntax::{
-    JsAnyExpression, JsxAttribute, JsxSelfClosingElement, JsxSelfClosingElementFields,
-};
+use rome_js_syntax::{JsAnyExpression, JsxAttribute, JsxSelfClosingElement};
 use rome_rowan::{AstNode, BatchMutationExt};
 
 declare_rule! {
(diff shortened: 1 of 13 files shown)


crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,119 @@
...
+                        }
+                    }
+                    None => {
+                        return Some(attribute);
+                    }
+                },
+                None => {
+                    return Some(attribute);
+                }
+            }
```
ematipico, line 89: I think we can simply the whole `return Some(attribute)`.

I think we change the `match` in

```rust
if let Some(initializer) = attribute.initializer() {

}
return Some(attribute)
```

And inside the `if let` only use the  `return` when we know it should trigger the rule, meaning doing `return None`. The rest of the branches of the code will fallback in the `return Some(attribute)` we have after the `if let`

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    fn diagnostic(_ctx: &RuleContext<Self>, attr: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic::new(
+            rule_category!(),
+            attr.range(),
+            markup! {
+                "Avoid the "<Emphasis>"autoFocus"</Emphasis>" attribute."
+            }
+            .to_owned(),
```
ematipico, line 102: I think `to_owned()` is not needed here

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,119 @@
...
+        ))
+    }
+
+    fn action(ctx: &RuleContext<Self>, attr: &Self::State) -> Option<JsRuleAction> {
+        let mut mutation = ctx.root().begin();
+
+        mutation.remove_node(attr.to_owned());
```
ematipico, line 109: We tend to use the `.clone()` instead of `to_owned()`

ematipico (approved):
Thank you! Overall the PR looks good and it does what intended! I just left few suggestions

jk-gan:
alright, let me update this

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
jk-gan: fixed

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
jk-gan: removed

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
jk-gan: I've updated the code.

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -62,31 +62,19 @@ impl Rule for NoAutoFocus {
...
-                            None => return Some(attribute),
-                        };
-                        if value.trim() == "undefined" {
-                            return None;
-                        } else {
-                            return Some(attribute);
+            if let Some(initializer) = attribute.initializer() {
+                if let Some(value) = initializer.value().ok() {
+                    if let Some(value) = value.as_jsx_expression_attribute_value() {
+                        if let Some(value) = value.expression().ok() {
```
ematipico, line 68: Now that we simplified the chain of returns, we can simplify this even further. All the `if let Some` can be removed using the try operator `?`. We already do something similar in different places:

https://github.com/rome/tools/blob/main/crates/rome_js_analyze/src/semantic_analyzers/nursery/no_array_index_key.rs#L60-L75

In this snippet we traverse the AST using the combination `as_*` functions and the try operator with `ok()?`

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+impl Rule for NoAutoFocus {
+    type Query = Ast<JsxSelfClosingElement>;
+    type State = JsxAttribute;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let node = ctx.query();
```
xunilrj, line 55: If we are only using name it is just simpler to do

```rust
let name = ctx.query().name.ok()?;
```

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,113 @@
...
+        let JsxSelfClosingElementFields {
+            l_angle_token: _,
+            name,
+            type_arguments: _,
+            attributes: _,
+            slash_token: _,
+            r_angle_token: _,
+        } = node.as_fields();
+
+        if name.ok()?.text().trim() == "input" {
```
xunilrj, line 65: This will not trim all trivia. You should use ```name.text_trimmed()```.

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,113 @@
...
+            let attribute = node.find_attribute_by_name("autoFocus").ok()??;
```
xunilrj, line 66: We normaly do ```find_attribute_by_name(...).ok()?``` to avoid the double question mark.

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,113 @@
...
+
+            if let Some(value) = get_attribute_value(&attribute) {
+                if value.text().trim() == "undefined" {
```
xunilrj, line 69: Same thing here. we need to use ```value.text_trimmed()```.

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
jk-gan: sry, I do not fully understand this, can you explain more?

ematipico: @xunilrj The code here is fine, `find_attribute_by_name` returns a `SyntaxResult<Option<>>`, so the double try operator makes sense in this case.

jk-gan:
I've updated the code based on the feedbacks.

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,113 @@
...
+            return Some(attribute);
+        }
+
+        None
+    }
+
+    fn diagnostic(_ctx: &RuleContext<Self>, attr: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic::new(
+            rule_category!(),
+            attr.range(),
```
xunilrj, line 74: You probably want the ```trimmed_range()``` here, otherwise we will highlight all the trivia.

jk-gan: ahh, let me change it

jk-gan: updated

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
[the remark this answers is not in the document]
jk-gan: done

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// ```jsx
+    /// <input />
+    ///```
+    ///
+    /// ```jsx
+    /// <input autoFocus={undefined} />
+    ///```
+    ///
+    /// ```jsx
+    /// <Input autoFocus={"false"} />
```
xunilrj, line 38: If the only difference is the uppercase "I" I would put a comment around it.

crates/rome_js_analyze/tests/specs/nursery/noAutofocus/noAutoFocusValid.jsx
```diff
@@ -0,0 +1,5 @@
+<>
+	<input />
+    <input autoFocus={undefined} />
```
xunilrj, line 3: ```<input autoFocus={null} />``` is valid or invalid?

crates/rome_js_analyze/tests/specs/nursery/noAutofocus/noAutoFocusInvalid.jsx
```diff
@@ -0,0 +1,5 @@
+<>
+    <input autoFocus />
+    <input autoFocus="true" />
+    <input autoFocus={"false"} />
```
xunilrj, line 4: Would be interesting to also see

```jsx
<input autoFocus={false} />
<input autoFocus={a} />
```

crates/rome_js_analyze/tests/specs/nursery/noAutofocus/noAutoFocusValid.jsx
ematipico: I **believe** that null `props` don't get rendered, but I might be wrong.

crates/rome_js_analyze/src/analyzers/nursery/no_auto_focus.rs
ematipico, line 38: I think these cases should fail. Check the original rule: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-autofocus.md#fail

jk-gan: Alright, because I was following this [docs](https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/noAutofocus.md). If we want to follow `eslint-plugin-jsx-a11y`, then it should be any element with `autoFocus` attribute should be avoided. [https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f6ba03cd5c206e4ca04068253ab79ad32eb929ce/src/rules/no-autofocus.js](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f6ba03cd5c206e4ca04068253ab79ad32eb929ce/src/rules/no-autofocus.js)

ematipico: Yeah apologies for that. Probably that rule was buggy to begin with. Yeah it's best to follow the `a11y` plugin, it's been around longer and they catch more cases.

jk-gan: No worries! I will change the implementation to follow `a11y` plugin.

jk-gan: Hey, `eslint-plugin-jsx-a11y` provide a config `ignoreNonDOM` to ignore user created component `autoFocus` attribute. I wonder how we should do that. Should we included the user-created components or just ignore them or provides an additional option to let user to config?

jk-gan: this is how to config to ignore user-created components in `eslint`
```json
"rules": {
      "jsx-a11y/no-autofocus": [ 2, {
        "ignoreNonDOM": true
    }]
}
```

ematipico: As for now, we can't configure rules with options. I think we should always ignore user-created components and just inspect normal elements.

jk-gan: alright

jk-gan:
Hi, I've re-implemented the `noAutofocus` rule. It now will show error on any elements, such as `<input />`, `<div />`, `<button />`, with `autoFocus` attribute (same with `eslint-plugin-jsx-a11y` plugin.

PS: I use `element.name().ok()?.as_jsx_name()?;` to check if the element is an user-created elements or not because I saw from the Rome Playground user-created element name is `JsxReferenceIdentifier`. Do let me know if I'm doing this wrong.

ematipico (approved):

PR #3336, opened 2022-10-05
feat(rome_analyzer): add `noPositiveTabindex` rule

### LINKED ISSUE #3304: `noPositiveTabIndex`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/noPositiveTabindex.md

kaioduarte:
## Summary

Closes #3304 

## Test Plan

image

image


crates/rome_js_analyze/tests/specs/nursery/noPositiveTabindex/invalidJsx.jsx
```diff
@@ -0,0 +1,18 @@
+let test;
+
+test = <div tabIndex={1} />;
+
+test = <div tabIndex={"1"} />;
+
+test = <div tabIndex={'5'} />;
+
+// not working - inner_string_text can't capture the number
+test = <div tabIndex="1" />;
```
kaioduarte, line 10: The trailing space at the end is causing the issue:
```rust
Split(
    SplitInternal {
        start: 0,
        end: 4,
        matcher: StrSearcher {
            haystack: "\"1\" ", // note the trailing space here
        },
    },
)
```

kaioduarte (commented):
I'm opening a draft PR so we can discuss the implementation.

While implementing the rule, I noticed that the helper `inner_string_text` couldn't strip the quotes on `JsxSelfClosingElement`. Can you confirm this is working as intended, or is it a bug?

The same also happens when getting the value for `JsNumberLiteralExpression` and `JsStringLiteralExpression` on `React.createElement`.

Any help is appreciated!

crates/rome_js_analyze/tests/specs/nursery/noPositiveTabindex/invalidJsx.jsx
kaioduarte: cc @ematipico

ematipico: OK, see now that you meant. This is not a bug, it works as intended. That's because internally string literal are represented with two quotes, because we **preserve** them. Check the [playground](https://play.rome.tools/?code=%3Cinput+tabIndex%3D%221%22+%2F%3E&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&enabledNurseryRules=true&typescript=false&jsx=false#PABpAG4AcAB1AHQAIAB0AGEAYgBJAG4AZABlAHgAPQAiADEAIgAgAC8APgA=)

The node that holds the value `"1"` keeps the quotes from source, and the function removes the first ones. So yes, it works as intended.

kaioduarte: > From `"\"1\""` to `"1"`.

I'm actually getting `1" ` with the helper because of the Whitespace token, and this condition is never reached:
https://github.com/rome/tools/blob/main/crates/rome_js_syntax/src/jsx_ext.rs#L29

ematipico: I managed to replicate the issue locally, and yes, it's definitely a bug!

kaioduarte: Will do, thanks for the quick fix!

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+    type Query = Semantic<NoPositiveTabindexQuery>;
+    type State = NoPositiveTabindexState;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let node = ctx.query();
+
+        match node {
+            NoPositiveTabindexQuery::JsxOpeningElement(opening_element) => {
+                let attribute = opening_element.find_attribute_by_name("tabIndex").ok()?;
```
ematipico, line 68: Here you can use a second try operator `?`, because the logic of the rule has to continue only if the attribute exits.

```rs
let attribute = opening_element.find_attribute_by_name("tabIndex").ok()??;
```

Doing so, you can remove the next `if let Some()`, change the signature of `is_jsx_attribute_valid` and avoid the `unwrap` later

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+                        return Some(NoPositiveTabindexState::Attribute(
+                            attribute.unwrap().clone(),
+                        ));
+                    }
+                }
+            }
+            NoPositiveTabindexQuery::JsxSelfClosingElement(self_closing_element) => {
+                let attribute = self_closing_element
+                    .find_attribute_by_name("tabIndex")
+                    .ok()?;
```
ematipico, line 79: Same thing here, a second try operator `?` will simplify the code and avoid the unwrap

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+            NoPositiveTabindexQuery::JsCallExpression(expression) => {
+                let model = ctx.model();
+                let react_create_element =
+                    ReactCreateElementCall::from_call_expression(expression, model)?;
+                let tabindex_prop = react_create_element.find_prop_by_name("tabIndex");
+
+                if let Some(prop) = tabindex_prop {
+                    let value = prop.value().ok()?;
+
+                    if let Some(expression_value) = get_expression_value(&value) {
```
ematipico, line 96: You can use the try operator `?` here so we don't need the `if let Some` anymore.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+                    }
+                }
+            }
+            NoPositiveTabindexQuery::JsCallExpression(expression) => {
+                let model = ctx.model();
+                let react_create_element =
+                    ReactCreateElementCall::from_call_expression(expression, model)?;
+                let tabindex_prop = react_create_element.find_prop_by_name("tabIndex");
+
+                if let Some(prop) = tabindex_prop {
```
ematipico, line 93: You can use the try operator `?` here so we don't need the `if let Some` anymore.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+        .footer(
+            Severity::Note,
+            "Elements with a positive tab index override natural page content order. This causes elements without a positive tab index to come last when navigating using a keyboard.",
+        );
+
+        Some(diagnostic)
+    }
+}
+
+fn get_expression_value(expression: &JsAnyExpression) -> Option<SyntaxTokenText> {
```
ematipico, line 137: Could you add some documentation to this function, explaining what it's doing, please? The `JsAnyExpression` as input and `SyntaxTokenText` as output are really generic, so some explanation would help.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+    match expression {
+        JsAnyExpression::JsAnyLiteralExpression(
+            JsAnyLiteralExpression::JsStringLiteralExpression(literal),
+        ) => literal.inner_string_text().ok(),
+        JsAnyExpression::JsAnyLiteralExpression(
+            JsAnyLiteralExpression::JsNumberLiteralExpression(literal),
+        ) => Some(literal.value_token().ok()?.token_text()),
+        JsAnyExpression::JsUnaryExpression(unary_expression) => {
```
ematipico, line 145: Since this function operates on a small of variants, maybe we can make it a bit more explicit using this pattern we use scattered across the repository:

```rs
use rome_js_syntax::{ 
	JsStringLiteralExpression,  
	JsNumberLiteralExpression, 
	JsUnaryExpression
};

/// Some documentation to explain
declare_node_union! {
	pub(crate) AnyNumberLikeExpression = JsStringLiteralExpression | JsNumberLiteralExpression | JsUnaryExpression
}

impl AnyNumberLikeExpression {
	pub(crate) fn value(&self) -> Option<SyntaxtTextToken> {}
}
```

Then, inside this function, you can do:

```rs
if let Some(number_like_expression) = AnyNumberLikeExpression::cast(expression.clone().syntaxt()) {
	number_like_expression.value()
} else {
	None
}
```

Doing so allows us to be more explicit over a subset of nodes, merge `AstNode` operations on the same enum and better document the operations.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+
+                // Build a string literal expression with the operator and the argument
+                // e.g. operator MINUS and argument value 5 becomes "-5"
+                return get_expression_value(&JsAnyExpression::JsAnyLiteralExpression(
+                    JsAnyLiteralExpression::JsStringLiteralExpression(
+                        make::js_string_literal_expression(make::js_string_literal(
+                            &string_literal,
+                        )),
+                    ),
+                ));
```
ematipico, line 161: I am not sure using the mutation APIs in this place is the best fit.. Instead, why don't we to the transformation from string to number in this function?

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            None
+        }
+        _ => None,
+    }
+}
+
+fn is_jsx_attribute_valid(jsx_attribute: &Option<JsxAttribute>) -> Option<bool> {
```
ematipico, line 170: Could you add a bit of documentation here, please? The "Valid" word depends on the context, and here it's not provided. Also, I would change the name of the function too e.g. `is_jsx_attribute_valid_tabindex`, `attribute_has_valid_tabindex`, etc.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+                return Some(is_valid_tabindex(&literal_string));
+            }
+            JsxAnyAttributeValue::JsxExpressionAttributeValue(value) => {
+                let expression = value.expression().ok()?;
+
+                if let Some(expression_value) = get_expression_value(&expression) {
+                    return Some(is_valid_tabindex(&expression_value));
+                }
+            }
+            _ => return Some(true),
```
ematipico, line 186: This function never returns `None` and it should. For example, here a `None` here makes sense because they are props that should never trigger the rule, e.g. `<div {...props} />`. And then, we change the usage of this function to use the try operator `is_jsx_attribute_valid(&attribute)?`, so we bounce back all the cases that should be ignored  by this rule.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,204 @@
...
+        JsAnyExpression::JsAnyLiteralExpression(
+            JsAnyLiteralExpression::JsStringLiteralExpression(literal),
+        ) => literal.inner_string_text().ok(),
+        JsAnyExpression::JsAnyLiteralExpression(
+            JsAnyLiteralExpression::JsNumberLiteralExpression(literal),
+        ) => Some(literal.value_token().ok()?.token_text()),
+        JsAnyExpression::JsUnaryExpression(unary_expression) => {
+            let argument_expression = unary_expression.argument().ok()?;
+            let argument_value = get_expression_value(&argument_expression);
+            let operator = unary_expression.operator_token().ok()?.token_text_trimmed();
```
ematipico, line 148: We should ignore all the unary operators that don't fit the usage. For example `<div tabIndex={typeof e} />` is a correct code and `typeof e` is an unary operator. We should check the operator to be equals to `JsUnaryOperator::Plus` and `JsUnaryOperator::Minus`

ematipico (changes requested):

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
kaioduarte: I addressed this on this commit [1584b43](https://github.com/rome/tools/pull/3336/commits/1584b439d4d40d43e095b0b1ef623372555cda40). Let me know if it looks good, and thanks for the feedbacks!

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,205 @@
...
+use rome_js_syntax::{
+    JsCallExpression, JsNumberLiteralExpression, JsPropertyObjectMember, JsStringLiteralExpression,
+    JsUnaryExpression, JsxAnyAttributeValue, JsxAttribute, JsxOpeningElement,
+    JsxSelfClosingElement,
+};
+use rome_rowan::{declare_node_union, AstNode};
+
+declare_rule! {
+    /// Prevent the usage of positive integers on tabIndex property
+    ///
```
ematipico, line 17:
```suggestion
    /// Prevent the usage of positive integers on `tabIndex` property
    ///
    /// Avoid positive `tabIndex` property values to synchronize the flow of the page with keyboard tab order.
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// ## Examples
```
ematipico, line 18:
```suggestion
    /// ## Accessibility guidelines
    ///
    /// [WCAG 2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order)
	///
    /// ## Examples
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+declare_node_union! {
+    /// Subset of expressions supported by this rule.
+    ///
+    /// ## Examples
+    ///
+    /// ### JsStringLiteralExpression &mdash; `"5"`
+    /// ### JsNumberLiteralExpression &mdash; `5`
+    /// ### JsUnaryExpression &mdash; `+5` | `-5`
```
ematipico, line 65:
```suggestion
    /// - `JsStringLiteralExpression` &mdash; `"5"`
    /// - `JsNumberLiteralExpression` &mdash; `5`
    /// - `JsUnaryExpression` &mdash; `+5` | `-5`
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let diagnostic = RuleDiagnostic::new(
+            rule_category!(),
+            text_range,
+            markup! { "Avoid positive values for the "<Emphasis>"tabIndex"</Emphasis>" prop." }
+                .to_owned(),
+        )
+        .footer(
+            Severity::Note,
+            "Elements with a positive tab index override natural page content order. This causes elements without a positive tab index to come last when navigating using a keyboard.",
```
ematipico, line 171:
```suggestion
            "Elements with a positive "<Emphasis>"tabIndex"</Emphasis> override natural page content order. This causes elements without a positive tab index to come last when navigating using a keyboard.",
```

crates/rome_js_analyze/tests/specs/nursery/noPositiveTabindex/invalidJsx.jsx.snap
```diff
@@ -0,0 +1,193 @@
...
+```
+
+# Diagnostics
+```
+invalidJsx.jsx:2:7 lint/nursery/noPositiveTabindex ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  ! Avoid positive values for the tabIndex prop.
+  
+    1 │ <>
+  > 2 │ 	<div tabIndex={1} />
```
ematipico, line 30: The diagnostic shows that the error is `tabIndex`, while it should point `{1}` (or `1` if we want to be more precise, but that's a nitpick). If you pair this diagnostic with the phrase `Avoid positive values for the tabIndex prop.`, it can be misleading or lead to a bug.

crates/rome_js_analyze/src/semantic_analyzers/nursery/no_positive_tabindex.rs
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+
+        None
+    }
+}
+pub(crate) enum NoPositiveTabindexState {
+    Attribute(JsxAttribute),
+    MemberProp(JsPropertyObjectMember),
+}
```
ematipico, line 117: Unless we decide to have a different message for attribute and member prop, there's actually not need to have the enum.

A better way to play with the state, it would be to accept a `TextRange` which is returned by `text_ranged_trimmed()`.

**If we find an incorrect attribute value**, we can do something like:

```rs
if let Some(invalid_attribute) = attribute_has_invalid_tabindex(&jsx_any_attribute_value) {
      return Some(invalid_attribute.syntaxt().text_range_trimmed());
}
// ...

if let Some(invalid_property) =  has_invallid_tabindex(&expression_value) {
    return Some(invalid_property.syntaxt().text_range_trimmed());
}
```

To not the change in name of the function: **attribute_has_invalid_tabindex** (`invalid`)

ematipico (changes requested):
The implementation is way better now! I left few more suggestions, and we should fix the incorrectness of the diagnostic.

Also, it seems there are some conflicts to solve

leops (approved):

crates/rome_js_analyze/tests/specs/nursery/noPositiveTabindex/invalidJsx.jsx.snap
kaioduarte: I updated this and the screenshots in the PR description as well.

ematipico (approved):

kaioduarte:
sorry, forgot to update doc files 😞

ematipico:
> sorry, forgot to update doc files 😞

No worries. To fix the CI issue, you need to run `cargo codegen-configuration`

PR #3339, opened 2022-10-05
fix(rome_service): disallow large files from being parsed and analyzed

### LINKED ISSUE #3330: 🐛 `rome check .` seems to run in an infinite loop with 99% CPU usage
### Environment information

```bash
OS Information:
- macOS Monterey, version 12.5.1 (21G83)
- Apple M1 Max

Context:
- Wanted to play with the new version in Tailwind CSS to see how it behaves when running `rome format . --write` or `rome check .`

Installation:
- `npm install rome@next --save-dev`

Versions:
- Node: v18.8.0
- NPM: 8.18.0
```

### What happened?

Every now and then when you release a new version, I try to play with it on a real codebase, in my case I'm playing with things in the Tailwind CSS codebase (https://github.com/tailwindlabs/tailwindcss).

I basically installed rome:
```sh
npm install rome@next --save-dev
```

... once that was done, I executed:
```sh
rome check .
```

It shows a bunch of output already, but then just hangs, and it looks like this: 
image

When I looked at activity monitor I saw this:
image

I eventually had to quit rome by using `ctrl+c` and luckily that worked without me force killing the process.

### Expected result

I would have expected that the command (`rome check .`) did its job and once it was done, that it exited the process successfully.

### Code of Conduct

leops:
## Summary

Fixes #3330

This PR adds a maximum size threshold to the Workspace after which files are no longer parsed, analyzed or formatted. At the moment I've hard-coded this limit to 1 MiB, but it should be trivial to expose this limit in the configuration file or the CLI if users really want to process very large files.

As part of the new message for this new error, I've added a new `Bytes` utility to `rome_console` that prints a quantity in bytes using the correct binary unit (B, KiB, MiB, GiB or TiB)

## Test Plan

I've added new test cases for the CLI that ensures files larger than the size threshold are not checked or formatted, and the correct diagnostic is emitted.

### DIFF
--- a/crates/rome_cli/tests/snapshots/main_ci/file_too_large.snap
+++ b/crates/rome_cli/tests/snapshots/main_ci/file_too_large.snap
@@ -1,0 +1,19 @@
+---
+source: crates/rome_cli/tests/snap_test.rs
+expression: content
+---
+# Emitted Messages
+
+```block
+ci.js lint ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  × The file ci.js could not be parsed because it's too large (the file is 1.0 MiB long, but the size limit is 1.0 MiB)
+  
+
+```
+
+```block
+Skipped 1 files
+```
+
+
--- a/crates/rome_service/src/workspace/server.rs
+++ b/crates/rome_service/src/workspace/server.rs
@@ -169,6 +169,18 @@ impl WorkspaceServer {
                     .parse
                     .ok_or_else(self.build_capability_error(rome_path))?;
 
+                /// Limit the size of files to 1.0 MiB
+                const SIZE_LIMIT: usize = 1024 * 1024;
+
+                let size = document.content.as_bytes().len();
+                if size >= SIZE_LIMIT {
+                    return Err(RomeError::FileTooLarge {
+                        path: rome_path.to_path_buf(),
+                        size,
+                        limit: SIZE_LIMIT,
+                    });
+                }
+
                 let parsed = parse(rome_path, document.language_hint, &document.content);
 
                 Ok(entry.insert(parsed).clone())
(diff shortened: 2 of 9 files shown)


crates/rome_service/src/workspace/server.rs
```diff
@@ -169,6 +169,18 @@ impl WorkspaceServer {
                     .parse
                     .ok_or_else(self.build_capability_error(rome_path))?;
 
+                /// Limit the size of files to 1.0 MiB
+                const SIZE_LIMIT: usize = 1024 * 1024;
```
xunilrj, line 173: How about ```SIZE_LIMIT_IN_BYTES```?

xunilrj (approved):

crates/rome_cli/tests/snapshots/main_ci/file_too_large.snap
```diff
@@ -0,0 +1,19 @@
+---
+source: crates/rome_cli/tests/snap_test.rs
+expression: content
+---
+# Emitted Messages
+
+```block
+ci.js lint ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  × The file ci.js could not be parsed because it's too large (the file is 1.0 MiB long, but the size limit is 1.0 MiB)
```
ematipico, line 10: Is it a coincidence that the input file is `1.0` Megabyte  like the limit? If not, probably that's because we truncated some digits. I am not sure what to do... surely an inexperienced  user might think that's a bug.

ematipico (approved):

crates/rome_cli/tests/snapshots/main_ci/file_too_large.snap
leops: It's a bit of both, in this case the test file is a few bytes larger than 1 MiB but that doesn't show since the values higher than 1 KiB are not printed at full precision. However the file size limit is checked using `>=` so even a file that's precisely 1 MiB in size would be rejected and print the error anyway.

Issue #3343, opened 2022-10-05
change the trait `Rule` to accept `self`


ematipico:
this is needed so the Rule is able to read its own metadata, which is important when querying its own configuration

PR #3347, opened 2022-10-05
feat(rome_analyze): use serde to deserialize rule settings

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3343 
Closes https://github.com/rome/tools/issues/3344 
Closes https://github.com/rome/tools/issues/3345

The function `rule_options` has been renamed `rule_settings` and it's been lifted inside the `AnalyzerContext`. The `AnalyzerContext` now accepts a `RuleKey`, which is needed to retrieve and deserialize the JSON string coming from the configuration.

`serde` is not optional anymore (I thought the entity of the change was bigger).

The handling of the error has been left out on purpose for the time being, and it will be handled in a different PR

## Test Plan

The current CI should work

### DIFF
--- a/crates/rome_analyze/src/context.rs
+++ b/crates/rome_analyze/src/context.rs
@@ -1,7 +1,8 @@
 use crate::{
     registry::RuleRoot, AnalyzerOptions, CannotCreateServicesError, FromServices, Queryable, Rule,
-    ServiceBag,
+    RuleKey, ServiceBag,
 };
+use serde::Deserialize;
 use std::ops::Deref;
 
 type RuleQueryResult<R> = <<R as Rule>::Query as Queryable>::Output;
@@ -11,6 +12,7 @@ pub struct RuleContext<'a, R>
 where
     R: ?Sized + Rule,
 {
+    rule_key: RuleKey,
     query_result: &'a RuleQueryResult<R>,
     root: &'a RuleRoot<R>,
     services: RuleServiceBag<R>,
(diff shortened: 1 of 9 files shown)


crates/rome_analyze/src/context.rs
```diff
@@ -11,6 +12,7 @@ pub struct RuleContext<'a, R>
 where
     R: ?Sized + Rule,
 {
+    rule_key: RuleKey,
```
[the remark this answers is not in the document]
ematipico: Ah yeah! I completely overlooked that we have `R` at hand

crates/rome_analyze/src/context.rs
```diff
@@ -48,6 +49,52 @@ where
...
+    ///     type Query = ();
+    ///     type State = ();
+    ///     type Signals = ();
+    ///
+    ///     fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+    ///         let options = ctx.rule_settings::<RuleSettings>();
+    ///     }
+    /// }
+    /// ```
+    pub fn rule_settings<'de, ToType: Deserialize<'de>>(&'de self) -> Option<ToType> {
```
xunilrj, line 89: Any reason the settings type is not an associated type like other rule types?  

We can default it to ```()```;

and this method can use it directly from ```R::SettingsType```.

ematipico: Yes, it will be in https://github.com/rome/tools/pull/3352

I have broken down the work in smaller PRs :)

leops (approved):

PR #3349, opened 2022-10-06
fix(rome_wasm): fix the WASM diagnostic printer

leops:
## Summary

The `DiagnosticPrinter` was broken since #3315 as it was not deserializing diagnostics from JS back into Rust with the correct type (it was using `rome_diagnostics::Diagnostic` instead of `rome_diagnostics::v2::serde::Diagnostic`). This should fix the outstanding issue of the Playground failing to print the diagnostics.

## Test Plan

I've added a new case in the tests of the WASM API in `@rometools/js-api` to check the diagnostic printer works correctly


ematipico (approved):
CI failing because the JS is not formatted

PR #2975, opened 2022-07-31
feat(rome_js_analyze): no shadow restricted lint rule

xunilrj:
## Summary

Closes https://github.com/rome/tools/issues/2897. Will be available at https://github.com/rome/tools/issues/2835

I created this benchmark https://github.com/rome/tools/pull/2975/files#diff-02554aa7d6b22992e92b6b8fe6366b78706e2a98f0f421cfd4641f79fa7fec8c expecting that a fancy algorithm would be faster, but end up testifying that cache rules and a simple slice::contains was the faster for the small number os strings we test.

The binary that calls this benchmark is only because ```iai``` do not support setup yet (https://github.com/bheisler/iai/pull/24) Basically it runs the test twice. First only the setup, and second the whole test and take the difference.

## Codegen

I changed the codegen for analyzers prefixing it with "self" to avoid collisions with some crates. "regex" is an example.
I also implemented a "codegen all"  to make it simpler to update everything. But running changed things in the formatted and the parser.

## Parser errors

Some examples are commented because they generate parser errors.

```js
//function arguments() {}
//function eval() {}
```

I think our parser should accept these, and let the linter complain. I can change this in another PR.

## Test Plan

```
> cargo rome_cli -- check crates/rome_js_analyze/tests/specs/js/noShadowRestrictedNames.js 
```

PR #3354, opened 2022-10-06
feat(rome_js_semantic): simple globals resolution

xunilrj:
## Summary

This PR implements a simple global resolution for the semantic model.

## Test Plan

```
> cargo test -p rome_js_semantic -- ok_semantic_model_globals
```

### DIFF
--- a/crates/rome_js_semantic/src/semantic_model.rs
+++ b/crates/rome_js_semantic/src/semantic_model.rs
@@ -116,10 +116,12 @@ struct SemanticModelData {
     declaration_all_reads: HashMap<TextRange, Vec<(ReferenceType, TextRange)>>,
     // Maps a declaration range to the range of its "writes"
     declaration_all_writes: HashMap<TextRange, Vec<(ReferenceType, TextRange)>>,
-    /// All references that could not be resolved
-    unresolved_references: Vec<(ReferenceType, TextRange)>,
     // All bindings that were exported
     exported: HashSet<TextRange>,
+    /// All references that could not be resolved
+    unresolved_references: Vec<(ReferenceType, TextRange)>,
+    /// All references that are resolved to globals
+    global_references: Vec<(ReferenceType, TextRange)>,
 }
 
 impl SemanticModelData {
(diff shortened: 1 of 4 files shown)


crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -996,19 +1050,33 @@ impl SemanticModelBuilder {
...
+            global_references: self.global_references,
         };
         SemanticModel::new(data)
     }
 }
 
+#[derive(Default)]
+/// Extra options for the [SemanticModel] creation.
+pub struct SemanticModelOptions {
+    /// All the allowed globals names
```
ematipico, line 1064:
```suggestion
    /// All the allowed global names
```

ematipico (approved):

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -996,19 +1050,33 @@ impl SemanticModelBuilder {
...
+    pub globals: HashSet<String>,
```
leops, line 1065: How does `HashSet` compare in performance against using a pre-sorted `Vec` of strings and a binary search ?

crates/rome_js_semantic/src/semantic_model.rs
```diff
@@ -1041,7 +1109,7 @@ mod test {
             FileId::zero(),
             SourceType::js_module(),
         );
-        let model = semantic_model(&r.tree());
+        let model = semantic_model(&r.tree(), SemanticModelOptions::default());
```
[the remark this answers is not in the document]
xunilrj: > Shouldn't we fill the globals with the ones that we have inside globals/mod.rs? 

Yes, but it should be done at crates\rome_js_analyze\src\lib.rs:76 where you have info from the file being analyzed and can choose the globals.

crates/rome_js_semantic/src/semantic_model.rs
xunilrj: I prefer to merge this like it is, and we can bench in a real case later. I did a similar test while ago tough: https://github.com/rome/tools/pull/2975

Issue #3299, opened 2022-09-30
`noBlankTarget`


ematipico:
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/noTargetBlank.md

PR #3356, opened 2022-10-07
feat(rome_js_analyze): new rule `useBlankTarget`

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3299

## Test Plan

Added various test cases


crates/rome_js_analyze/src/analyzers/nursery/use_blank_target.rs
```diff
@@ -0,0 +1,192 @@
...
+        let target_attribute = target_attribute?;
+        let text = target_attribute
+            .initializer()?
+            .value()
+            .ok()?
+            .as_jsx_string()?
+            .inner_string_text()
+            .ok()?;
+
+        if text == "_blank" {
```
leops, line 102: I think this check should be case-insensitive, at least if I remember correctly browsers do not treat this attribute as case-sentitive

leops (approved):

crates/rome_js_analyze/src/analyzers/nursery/use_blank_target.rs
ematipico: Good catch. Yes, it seems to be case-insensitive

PR #3370, opened 2022-10-07
feat(rome_js_analyze): rule `useAnchorContent`

### LINKED ISSUE #3300: `useAnchorContent`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/useAnchorContent.md

lucasweng:
## Summary

- Closes #3300 

## Test Plan

Added various test cases


crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        Some(node.clone())
+    }
+
+    fn diagnostic(_ctx: &RuleContext<Self>, node: &Self::State) -> Option<RuleDiagnostic> {
+        Some(RuleDiagnostic::new(
+			rule_category!(),
+            node.syntax().text_trimmed_range(),
+            markup! {
+				"Provide screen reader accessible content when using "<Emphasis>"anchor"</Emphasis>" elements."
```
ematipico, line 109:
```suggestion
				"Provide screen reader accessible content when using "<Emphasis>"a"</Emphasis>" elements."
```

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,187 @@
...
+			}
+        ).footer_note(
+			markup! {
+				"All links on a page should have content that is accessible to screen readers."
+			}
+		))
+    }
+}
+
+fn is_accessible_to_screen_reader(element: JsxAnyChild) -> Option<bool> {
```
ematipico, line 119: Can we document this function and explain what "accessible to screen readers" mean?

crates/rome_js_analyze/tests/specs/nursery/useAnchorContent.jsx
```diff
@@ -0,0 +1,19 @@
+<>
+    {/* invalid */}
+    <a />
+		<a><TextWrapper aria-hidden /></a>
```
ematipico, line 4: I am not sure this case should be invalid. I think **it should be valid**, meaning it should not trigger the rule because we don't have enough info.

Here we are passing `aria-hidden` to a *component*, not an element, which means we don't know how this component is going to use the `aria-hidden` property. Hence, the rule should not be triggered for this case.

Here's an [example](https://stackblitz.com/edit/react-ts-6zuacx?file=App.tsx)

crates/rome_js_analyze/tests/specs/nursery/useAnchorContent.jsx
```diff
@@ -0,0 +1,19 @@
...
+		<a><TextWrapper aria-hidden={true} /></a>
+    <a><span aria-hidden="true">anchor content</span></a>
+		<a></a>
+		<a>{null}</a>
+		<a>{undefined}</a>
```
ematipico, line 9: I believe

```jsx
<a aria-hidden>content</a>
```
should be an invalid case, and trigger the rule.

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,187 @@
...
+                .find_attribute_by_name("dangerouslySetInnerHTML")
+                .ok()?
+                .is_some(),
+        })
+    }
+}
+
+impl Rule for UseAnchorContent {
+    type Query = Ast<UseAnchorContentNode>;
+    type State = UseAnchorContentNode;
```
ematipico, line 88: In this case, `State` can be `()`. And inside the `diagnostic` function you can do

```rs
let node = ctx.query();
```

ematipico (changes requested):
Thank you for the contribution! 

The PR looks good overall, although I think there are few cases that need to be reviewed.

crates/rome_js_analyze/tests/specs/nursery/useAnchorContent.jsx
lucasweng: Good point! I was following the eslint-plugin-jsx-a11y [documentation](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/bccf0aeab8dd337c5f134f892a6d3588fbc29bdf/docs/rules/anchor-has-content.md#fail) though. Yet we don't know if the `aria-hidden` property will be used to hide all the content or not. I'll adjust it!

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
lucasweng: Adjusted the wording in https://github.com/rome/tools/pull/3370/commits/f0ace947b425544d935c50454b03bc7a0801bdc0

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
lucasweng: Added some explanation to this function in https://github.com/rome/tools/pull/3370/commits/f0ace947b425544d935c50454b03bc7a0801bdc0

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,227 @@
...
+
+            let aria_hidden_attribute = opening_element
+                .find_attribute_by_name("aria-hidden")
+                .ok()??;
+            !is_hidden_from_screen_reader(aria_hidden_attribute)?
+        }
+        JsxAnyChild::JsxSelfClosingElement(element) => {
+            // We don't check if a component (e.g. <Text aria-hidden />) is using the `aria-hidden` property,
+            // since we don't have enough information about how the property is used.
+            element.name().ok()?.as_jsx_name()?;
```
ematipico, line 169: While this is correct, it's not very explicit (without the comment, someone might think it's an error). Maybe it reads better if we explicitly return 

```rs
if element.name().ok()?.as_jsx_name().is_none() {
	return None;
}
```

Or, to make _even more explicit_, we can do this

```rs
let element_name = element.name().ok()?;
if JsxName::can_cast(element_name.syntax().kind()) {
	return None
}
```

ematipico (approved):
I left some comment, we can improve a bit further the code, but overall the logic seems correct to me!

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let node = ctx.query();
+        if !node.is_anchor()? {
+            return None;
+        }
+
+        // If there's no `aria-hidden` attribute on the a element,
+        // proceed to check the accessibility of its child elements
+        if !node.is_hidden_from_screen_reader().unwrap_or(false) && node.has_accessible_child()? {
```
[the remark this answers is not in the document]
lucasweng: I found the usage of `unwrap_or` not intuitive too. I wasn't sure what is the better way to handle the `None` propagation from the `is_hidden_from_screen_reader` helper function. I didn’t use the `?` operator here because it would propagate `None` immediately without proceeding to run `node.has_accessible_child()?`.

So instead of using `unwrap_or` inside `node.is_hidden_from_screen_reader` (which I also found a bit unintuitive 🤔), it ended up having a `Option<bool>` return signature. 

I adjusted the signature per your suggestions in https://github.com/rome/tools/pull/3370/commits/d8aba30f65da8ff9c06ce8002a76cd463f3ae452, please let me know if we can improve the code further. Thanks!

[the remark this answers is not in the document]
ematipico: I think it's fine the current implementation. Unfortunately the `Option<>` can make this confusing when paired with boolean checks. 

But I also think it's fine to use the try operator `?` even it doesn't feel intuitive. The try operator is needed to bounce any potential errors in our syntax (among other things). That's why the rules are filled with `ok()?`. 

For example [here](https://github.com/rome/tools/blob/21592601fc06f28652b8e43a72c22af632559ecc/crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs#L88) we are bypassing any potential errors inside the code (we have to remember that our parser is recoverable, so it's able to spit an AST that contains unknown nodes, which are errors). While this is fine and it doesn't break any logic, it's a case where we could terminate early instead of moving to the next check.

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,239 @@
...
+                    Ok(Some(aria_hidden_attribute)) => {
+                        is_hidden_from_screen_reader(aria_hidden_attribute).unwrap_or(false)
+                    }
+                    _ => false,
+                }
+            }
+        }
+    }
+
+    /// Check if the a element has content accessible to screen readers.
```
MichaReiser, line 102:
```suggestion
    /// Check if the `a` element has content accessible to screen readers.
```

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,239 @@
...
+declare_node_union! {
+    pub(crate) UseAnchorContentNode = JsxElement | JsxSelfClosingElement
+}
+
+impl UseAnchorContentNode {
+    /// Check if the current element is an anchor
+    fn is_anchor(&self) -> Option<bool> {
+        Some(match self {
+            UseAnchorContentNode::JsxElement(element) => {
+                element.opening_element().ok()?.name().ok()?.text() == "a"
```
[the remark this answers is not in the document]
MichaReiser: @ematipico do you remember the rule where you re-wrote the match of the element name to use the token instead? 

I think that would be beneficial to do here too because comparing `SyntaxNodeText` is more expensive and e.g. not even necessary if `name` is a `JsxMemberName` or `JsxNamespaceName`

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// <a></a>
+    /// ```
+    ///
+    /// ```jsx,expect_diagnostic
+    /// <a aria-hidden>content</a>
+    /// ```
+    ///
+    /// ```jsx,expect_diagnostic
+    /// <a><span aria-hidden="true">content</span></a
+    /// ```
```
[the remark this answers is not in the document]
lucasweng: It should be invalid, but the current implementation doesn't catch this case. I'll fix this issue.

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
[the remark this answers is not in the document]
ematipico: I can't recall unfortunately. I can do this refactor, in a separate PR. This is not the only place where we should check.

crates/rome_js_analyze/src/analyzers/nursery/use_anchor_content.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /// Check if the a element has the `aria-hidden` attribute set to true.
+    fn is_hidden_from_screen_reader(&self) -> bool {
+        match self {
+            UseAnchorContentNode::JsxElement(element) => {
+                if let Ok(opening_element) = element.opening_element() {
+                    match opening_element.find_attribute_by_name("aria-hidden") {
+                        Ok(Some(aria_hidden_attribute)) => {
+                            is_hidden_from_screen_reader(aria_hidden_attribute).unwrap_or(false)
```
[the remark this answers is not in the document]
lucasweng: Adjusted in https://github.com/rome/tools/pull/3370/commits/3119407a9a0b7ded86b3255548deefe50601859a

PR #3371, opened 2022-10-08
feat(rome_js_analyzer): add useAltText lint rule for JSX expression

### LINKED ISSUE #3298: `useAltText`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/useAltText.md

b4s36t4:
## Summary

Closes #3298 following the document given for valid and in-valid conditions.

#3298 

## Test Plan

A lint rule to check for alt attribute in JSX for tags like `input type='image', img, area`.


crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub(crate) UseAltTextNode = JsxString | JsxSelfClosingElement | JsStringLiteralExpression | JsObjectExpression
+}
+
+pub(crate) struct UseAltTextState {
+    node: UseAltTextNode,
+    missing_alt_prop: bool,
+}
+
+impl Rule for UseAltText {
+    type Query = Semantic<UseAltTextQuery>;
```
leops, line 59: The rule could query for `Syntax<UseAltTextQuery>` instead, since it only relies on the syntax tree and doesn't seem to make use of any information from the semantic model

leops (approved):

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
b4s36t4: Which one it is?! 😅

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    pub(crate) UseAltText{
+        version:"0.10.0",
+        name:"useAltText",
+        recommended: false,
+    }
+}
+
+declare_node_union! {
+    pub (crate) UseAltTextQuery = JsxSelfClosingElement  | JsCallExpression
```
ematipico, line 46: I think we should limit ourselves only to `JsxSelfClosingElement` nodes. Which means that union is not needed anymore, and the `Query` will become:

```rs
type Query = Ast<JsxSelfClosingElement>;
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        if initalizer.inner_string_text().ok()? == "image" {
+            return Some(true);
+        }
+        return None;
+    }
+    None
+}
+
+fn is_valid_tag(name: &JsxAnyElementName) -> Option<bool> {
```
ematipico, line 171: Can we document this function? "Valid tag" is not clear, unless we look at the implementation of the function.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,181 @@
...
+                markup! {
+                    "Meaningful alternative text on elements helps users relying on screen
+                readers to understand content's purpose within a page."
+                },
+            ),
+        )
+    }
+}
+
+fn is_valid_input(element: &JsxSelfClosingElement) -> Option<bool> {
```
ematipico, line 156: Can we document this function? "Valid input" needs to have a context; we have to tell developers what a valid input means in this case.

Also, this function could be called `input_has_type_image`

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,181 @@
...
+                state.node.syntax().text_trimmed_range(),
+                message,
+            )
+            .footer(
+                Severity::Note,
+                markup! {
+                    "Meaningful alternative text on elements helps users relying on screen
+                readers to understand content's purpose within a page."
+                },
+            ),
```
ematipico, line 151: nit: you can `footer_note` function and pass only the text.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+pub(crate) struct UseAltTextState {
+    node: UseAltTextNode,
+    missing_alt_prop: bool,
+}
+
+impl Rule for UseAltText {
+    type Query = Semantic<UseAltTextQuery>;
+    type State = UseAltTextState;
```
ematipico, line 60: I think `State` can be a `TextRange`

```rust
type State = TextRange
```

And when you find a property that is incorrect, you just do 

```rust
return Some(element.syntaxt().text_range_trimmed())
```

And the `state` can be used inside the diagnostic. As you noticed the state is only used to retrieve the range.

ematipico (changes requested):
Thank you for the contribution! I left some suggestion and correction to the cases covered

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
ematipico: It's the second one :) just look for other rules that use `Ast<`

crates/rome_js_analyze/tests/specs/nursery/useAltText/area.jsx.snap
```diff
@@ -0,0 +1,99 @@
+---
+source: crates/rome_js_analyze/tests/spec_tests.rs
+expression: area.jsx
+---
+# Input
+```js
+// invalid
+
+<>
+	<area {...props} />
```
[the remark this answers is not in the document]
b4s36t4: This applies to other tags as well?

b4s36t4:
@ematipico I have updated the code, please check it. 
> Note: as part of skipping cases including `{...props}` spread items existing cases mentioned in the doc will became valid. > Please check the snapshots for more context.

crates/rome_js_analyze/tests/specs/nursery/useAltText/area.jsx
```diff
@@ -0,0 +1,17 @@
+// invalid
+
+<>
+	<area alt={undefined} />
```
ematipico, line 4: If we add a case for `undefined`, I think we should also add a case for `alt={null}`

crates/rome_js_analyze/tests/specs/nursery/useAltText/img.jsx
```diff
@@ -0,0 +1,22 @@
...
+	<img src="foo" role="none" />
+	<img alt />
+</>
+
+// valid
+
+<>
+	<img {...props} alt />
+	<img {...props} /> {/* Skipping*/}
+	<img {...props} alt={undefined} /> {/* Skipping*/}
```
ematipico, line 15: Can we remove these cases? It would be better to tackle spread props in a different PR (it's a feature missing in different rules, not just this one). And for example, the first and third cases are false positives.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        None
+    }
+
+    fn diagnostic(_ctx: &RuleContext<Self>, state: &Self::State) -> Option<RuleDiagnostic> {
+        if state.is_empty() {
+            return None;
+        }
+        let message = markup!(
+            "Provide "<Emphasis>"alt"</Emphasis>" when using "<Emphasis>"img"</Emphasis>", "<Emphasis>"area"</Emphasis>", "<Emphasis>"input type='image'"</Emphasis>""
```
ematipico, line 108:
```suggestion
            "Provide the attribute"<Emphasis>"alt"</Emphasis>" when using "<Emphasis>"img"</Emphasis>", "<Emphasis>"area"</Emphasis>" or "<Emphasis>"input type='image'"</Emphasis>""
```
You would need to regenerate the snapshots and documentation

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+        ).to_owned();
+
+        Some(
+            RuleDiagnostic::new(rule_category!(), state, message).footer(
```
ematipico, line 112: You can just use `footer_note`, so you don't need to pass `Severity`

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+                readers to understand content's purpose within a page."
+                },
+            ),
+        )
+    }
+}
+
+/*
+This function checks for the attribute `type` for input element where we checking for the input type which is image.
+ */
```
ematipico, line 125:
```suggestion
/// This function checks for the attribute `type` for input element where we checking for the input type which is image.
```

We should use rust documentation comments

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+        return None;
+    }
+    None
+}
+
+/*
+Function to check that the HTML/JSX tag is valid for the lint rule, which check if the tag should be `img`, `input type='image' and `area`.
+
+ */
```
ematipico, line 144:
```suggestion
/// Function to check that the HTML/JSX tag is valid for the lint rule, which check if the tag should be `img`, `input type='image' and `area`.
```

We should use rust documentation comments

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+
+            let is_spread_available = element
+                .attributes()
+                .iter()
+                .any(|attribute| attribute.as_jsx_spread_attribute().is_some());
+
+            // To-Do - Complex cases might invlove, not triggering rules.
+            if is_spread_available {
+                return None;
+            }
```
ematipico, line 71: I think it's best, for now, to remove the check for spread props. It's a feature/functionality that is missing from various rules and the implementation should do more checks. It's OK to remove it, I will work on it later.

ematipico (approved):

website/src/docs/lint/rules/UseAltText.md
```diff
@@ -0,0 +1,57 @@
+---
```
ematipico, line 1: This file seems incorrect. It should be called `useAltText.md`. Can you delete it and try to generate it again?

website/src/docs/lint/rules/index.md
```diff
@@ -367,6 +367,13 @@ Disallow unnecessary fragments
 This rules prevents void elements (AKA self-closing elements) from having children.
 </div>
 <div class="rule">
+<<<<<<< HEAD
+<h3 data-toc-exclude id="useAltText">
+	<a href="/docs/lint/rules/useAltText">useAltText (since v0.10.0)</a>
+	<a class="header-anchor" href="#useAltText"></a>
+</h3>
+It asserts that alternative text to images or areas, help to rely on to screen readers to understand the purpose and the context of the image.
+=======
```
ematipico, line 376: There's some issue with this file too... You should delete it and generate again

ematipico (changes requested):

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+        return None;
+    }
+    None
+}
+
+/*
+Function to check that the HTML/JSX tag is valid for the lint rule, which check if the tag should be `img`, `input type='image' and `area`.
+
+ */
+fn is_valid_tag(name: &JsxAnyElementName) -> Option<bool> {
```
xunilrj, line 145: "text_trimmed"and the navigation are expensive methods. You can simplify this method, and its call to:

```rs
match name.as_jsx_name().map(|x| x.text_trimmed()) {
    Some("input") => {...}
    Some("img" | "area") => {...}
}
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_alt_text.rs
```diff
@@ -0,0 +1,155 @@
...
+                .attributes()
+                .iter()
+                .any(|attribute| attribute.as_jsx_spread_attribute().is_some());
+
+            // To-Do - Complex cases might invlove, not triggering rules.
+            if is_spread_available {
+                return None;
+            }
+
+            let alt_prop = element.find_attribute_by_name("alt").ok()?;
```
xunilrj, line 73: You may also look at: https://github.com/rome/tools/pull/3458. 
It will improve the code here.

ematipico:
Closing in favour of https://github.com/rome/tools/pull/3502

PR #3373, opened 2022-10-09
feat(rome_js_analyze): rule `useKeyWithMouseEvents`

### LINKED ISSUE #3302: `useKeyWithMouseEvents`
https://github.com/rome/tools/blob/archived-js/website/src/docs/lint/rules/a11y/useKeyWithMouseEvents.md

mrkldshv:
## Summary

Closes #3302.

## Test Plan

Added various test cases


crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
```diff
@@ -0,0 +1,191 @@
...
+                    .text_trimmed();
+
+                return Some(name == "undefined");
+            }
+
+            let name = expression
+                .as_js_any_literal_expression()?
+                .as_js_null_literal_expression()?
+                .syntax()
+                .text_trimmed();
```
ematipico, line 186: I might be wrong, but **I think** don't need to check against the text `"null"`. Having a `JsNullLiteralExpression` should be enough

crates/rome_js_analyze/tests/specs/nursery/useKeyWithMouseEvents.jsx
```diff
@@ -0,0 +1,34 @@
+// invalid
+<div onMouseOver={() => {}} />;
+<div onMouseOut={() => {}} />;
+<div onMouseOver={() => {}} onFocus={undefined} />;
+<div onMouseOut={() => {}} onBlur={undefined} />;
+<div onMouseOver={() => {}} onFocus={null} />;
+<div onMouseOut={() => {}} onBlur={null} />;
+<div onMouseOver={() => {}} {...props} />;
```
ematipico, line 8: I think this should be **a valid case** for now. That's because we don't know if `props` contains `onBlur`/`onFocus`, and **it might**. Because of that, we don't have enough information, and we should not trigger the rule.

ematipico (changes requested):
Thank you for the contribution! I left some suggestion. I think there's some case that should not trigger.

crates/rome_js_analyze/tests/specs/nursery/useKeyWithMouseEvents.jsx
mrkldshv: Hmm, I see. I was looking at eslint [tests](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/mouse-events-have-key-events-test.js) when implementing this rule. Do you think we should deviate from eslint here?

mrkldshv: Hmm, I see. I was looking at eslint [tests](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/mouse-events-have-key-events-test.js) when implementing this rule. Do you think we should deviate from eslint here?

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
mrkldshv: Good point. Will adjust it!

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
```diff
@@ -0,0 +1,191 @@
...
+impl Rule for UseKeyWithMouseEvents {
+    type Query = Semantic<JsxAnyElement>;
+    type State = UseKeyWithMouseEventsState;
+    type Signals = Option<Self::State>;
+
+    fn run(ctx: &RuleContext<Self>) -> Self::Signals {
+        let node = ctx.query();
+
+        match node {
+            JsxAnyElement::JsxSelfClosingElement(element) => {
```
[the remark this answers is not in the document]
mrkldshv: Should we do `find_by_attribute` for `onFocus` only in case we find `onMouseOver` attribute?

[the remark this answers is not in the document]
ematipico: We could, but that would repeat the code again. But I suppose that's fine, I leave it to your discretion :)

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
```diff
@@ -0,0 +1,191 @@
...
+        .initializer()
+        .and_then(|x| {
+            let expression = x
+                .value()
+                .ok()?
+                .as_jsx_expression_attribute_value()?
+                .expression()
+                .ok()?;
+
+            if expression.as_js_identifier_expression().is_some() {
```
xunilrj, line 171: You can use (that are other cases similar to this one)

```rust
if let Some(id) = expression.as_js_identifier_expression() {
    let name = id.name().ok()?.syntax().text_trimmed();
```

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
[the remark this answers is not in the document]
mrkldshv: I was thinking about the way how we share common code for this case. Because we have four attributes to find, it gets pretty repetitive with something like this:
```rs
let on_mouse_over_attribute = element.find_attribute_by_name("onMouseOver").ok()?;
let on_mouse_out_attribute = element.find_attribute_by_name("onMouseOut").ok()?;

match (&on_mouse_over_attribute, &on_mouse_out_attribute) {
    (Some(_), Some(_)) => (
        on_mouse_over_attribute,
        element.find_attribute_by_name("onFocus").ok()?,
        on_mouse_out_attribute,
        element.find_attribute_by_name("onBlur").ok()?,
    ),
    (Some(_), None) => (
        on_mouse_over_attribute,
        element.find_attribute_by_name("onFocus").ok()?,
        on_mouse_out_attribute,
        None,
    ),
    (None, Some(_)) => (
        on_mouse_over_attribute,
        None,
        on_mouse_out_attribute,
        element.find_attribute_by_name("onBlur").ok()?,
    ),
    (None, None) => (on_mouse_over_attribute, None, on_mouse_out_attribute, None),
}
```
I guess there should be a better way of doing it. Any suggestions?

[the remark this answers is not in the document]
ematipico: Another suggestion would be to move the majority of the common functions inside the enum `JsxAnyElement`.

Some example:

```rs
impl JsxAnyElement {
	fn find_on_mouse_over_attribute() {}
	fn find_on_mouse_out_attribute() {}
	fn find_on_focus_attribute() {}
	fn find_on_blur_attribute() {}

	/// here we check the pair "onFocus" and "onMouseOver"
	fn has_focus_attributes() {
		let on_mouse_over = self.find_on_mouse_over_attribute();
		let on_focus = self.find_on_focus_attribute();
	}
    /// here we check the pair "onBlur" and "onMouseOut"
	fn has_blur_attributes() {
		let on_mouse_over = self.find_on_mouse_out_attribute();
		let on_focus = self.find_on_blur_attribute();
	}
}
```

Even though we have four attributes, they pair are fixed, so we don't need to do a matrix.

mrkldshv:
@ematipico @xunilrj Thanks for the review and your suggestions! I've addressed them in the last commits.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    fn has_focus_attributes(&self) -> Option<bool> {
+        if self.find_on_mouse_over_attribute().is_some() {
+            let on_focus_attribute = self.find_on_focus_attribute();
+
+            if on_focus_attribute.is_none() || is_value_undefined_or_null(&on_focus_attribute?) {
+                return None;
+            }
+            return Some(true);
```
ematipico, line 156: I _think_ the return here is not needed... it would go straight to the last `Some(true)`

ematipico (approved):
Thank you! The code looks way better now! :) 
I just left few tiny improvements, but I think the rule can be merged as it is

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
mrkldshv: Hmm, `rust-analyzer` tells me it's required in this case. Probably because it's early return.

crates/rome_js_analyze/src/semantic_analyzers/nursery/use_key_with_mouse_events.rs
```diff
@@ -0,0 +1,239 @@
...
+                None
+            }
+        }
+    }
+
+    fn has_spread_attribute(&self) -> bool {
+        match self {
+            JsxAnyElement::JsxSelfClosingElement(element) => element
+                .attributes()
+                .into_iter()
```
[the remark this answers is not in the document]
mrkldshv: It seems that `JsxAttributeList` doesn't have `iter` method on it.

PR #3374, opened 2022-10-10
fix(rome_formatter): Fill fits

MichaReiser:
## Summary

Implements the fix for https://github.com/rome/tools/pull/3251#issuecomment-1264447408

This PR addresses an issue in the Printer where fill printed the item in flat mode even though the separator neither fits in flat nor expanded mode. In that case, it is necessary to expand the `item` to avoid overflowing the configured print width. 

## Test Plan

See the updated snapshots.

### DIFF
--- a/crates/rome_formatter/src/printer/mod.rs
+++ b/crates/rome_formatter/src/printer/mod.rs
@@ -21,8 +21,7 @@ use crate::printer::call_stack::{
 };
 use crate::printer::line_suffixes::{LineSuffixEntry, LineSuffixes};
 use crate::printer::queue::{
-    AllPredicate, FitsPredicate, FitsQueue, PrintQueue, Queue, SeparatorItemPairPredicate,
-    SingleEntryPredicate,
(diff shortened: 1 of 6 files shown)


MichaReiser:
!bench_formatter

crates/rome_formatter/src/printer/mod.rs
```diff
@@ -760,291 +828,345 @@ impl Default for Indention {
...
+            self.stack.pop(TagKind::Fill)?;
         }
+
+        Ok(fits)
     }
 
-    Ok(Fits::Maybe)
-}
+    /// Tests if the passed element fits on the current line or not.
+    fn fits_element(&mut self, element: &'a FormatElement) -> PrintResult<Fits> {
```
MichaReiser, line 947: The only changes to this method are that I had to rewrite `state` to `self.state`, `queue` to `self.queue`, ... plus the change in indention level.

ematipico (approved):

PR #3381, opened 2022-10-10
feat(rome_js_analyze): compute `AnalyzerOptions` from `Workspace`

ematipico:
## Summary

Closes https://github.com/rome/tools/issues/3392

I also changed the signature of the function `run` to result a `Result`. This would allow us to track the diagnostics emitted from errors. The tracking of the errors will be done in a separate PR, because it requires a refactor of the current diagnostics of the analzyer, which still use the old diagnostics.

## Test Plan

Unfortunately I can't test my changes yet. The current CI should work


crates/rome_service/src/configuration/mod.rs
```diff
@@ -279,33 +280,56 @@ where
 /// The function needs access to a filter, in order to have an easy access to the [metadata] of the
 /// rules.
 ///
-/// [metadata]: crate::rome_analyze::RegistryRuleMetadata
+/// The third argument is closure that accepts a reference to [Configuration] and has to map
```
MichaReiser, line 283: is  **a** closure

crates/rome_service/src/configuration/mod.rs
```diff
@@ -279,33 +280,56 @@ where
...
-        .and_then(|j| j.globals.as_ref())
-    {
-        globals
-            .iter()
-            .map(|global| global.to_string())
-            .collect::<Vec<_>>()
-    } else {
-        vec![]
-    };
+    to_globals: ToGlobals,
```
MichaReiser, line 322: My understanding is that globals are a JS specific concept. Shouldn't it be handled outside of the generic `to_analyzer_configuration` or how do we plan to support other languages?

crates/rome_service/src/configuration/mod.rs
MichaReiser: It can be easier to understand if you reference the variable name. This also guaranteees that the documentation won't be out of date when new arguments are added or arguments are removed:

> `to_globals` is a closure that takes the [Configuration] by reference and returns the globals of the language. 

I don't think it's worth documenting that it returns a `Vec<String>` because that's repeating the function signature.

crates/rome_service/src/configuration/mod.rs
```diff
@@ -358,21 +382,28 @@ fn push_rules<M>(
...
-            }),
-            ..Configuration::default()
-        };
+        let mut settings = WorkspaceSettings::default();
+        settings.languages.javascript.globals =
+            Some(["jQuery".to_string(), "React".to_string()].into());
+
         let filter = AnalysisFilter::default();
-        let analyzer_configuration = to_analyzer_configuration(&configuration, &filter);
+        let analyzer_configuration =
```
MichaReiser, line 396: Is this test now a duplicate of the example? If so, can we delete it?

crates/rome_service/src/configuration/mod.rs
ematipico: It was, but now they can be a CSS thing too :) 

Imagine a third party CSS that creates a variable `--external-font-size`, and it's used inside the user's code:

```css
.some_class {
	font-size: var(--external-font-size) * 0.2
}
```

ematipico: Unfortunately this approach won't scale that well, at least not as much as the same approach we have for `FormatOptions` -> `JsFormatOptions`. Initially I wanted to apply the same approach, but I encountered many complex issues that would have required refactors - which I am not yet comfortable with - inside the analyzer itself.

leops (approved):

PR #3382, opened 2022-10-10
fix(rome_diagnostics): fix the printing of empty diff lines

### LINKED ISSUE #3375: 🐛 new line only formatter diff
### Environment information

* version: `main` (e6dbf2a829387bb2946d693d416152501520e269)

Input file

```ts
declare module "test" {
	interface A {

		prop: string;
	}
}
```

### What happened?

1. Run `rome format`: ` cargo run --bin rome --release format ./test/`
2. Observe how Rome prints an empty diff

```
./test/test.d.ts format ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ℹ Formatter would have printed the following content:
  
  

Compared 1 files in 1277µs
```

### Expected result

I expect a diff that highlights that the empty line before the `prop: string` has been removed.

### Code of Conduct

leops:
## Summary

Fixes #3375

This PR fixes a bug that was causing empty inserted or deleted lines to not be displayed by the diff printing logic. Due to the line-based model used to render diffs those lines were considered to have no content, but they should now be properly accounted for.

## Test Plan

I've added new test cases to check the insertion and deletion of one or many empty lines are all correctly displayed by the diff printer. I've also updated a few snapshots that were affected by this issue


ematipico (approved):
Much better!

PR #3383, opened 2022-10-10
feat(rome_formatter): Measure unicode width

MichaReiser:
Use the `unicode-width` to calculate the width of each character instead of assuming that each character only takes a single column.

I noticed that this is an issue when using Rome to format the ant design project and Rome used different line break positions than Prettier.

This PR also fixes an issue in the diagnostic printing where it incorrectly used `column_number` to retrieve the subslice of the line. The problem with that is that this doesn't account for multi-byte characters.

closing #3379


leops (approved):

MichaReiser:
!bench_formatter

PR #3384, opened 2022-10-10
feat(rome_js_analyze): new rule `useFlatMap`

### LINKED ISSUE #3367: `map_flatten` https://rust-lang.github.io/rust-clippy/master/#map_flatten

ematipico:
## Summary

Closes #3367 

The rule checks for usages like `arr.map(function).flat(function)` and suggests the replace it with `arr.flatMap(function)` instead.

Clippy marks the code action as safe, so I suppose we can do that too? Please let me know if you think that the code action should be suggested.

## Test Plan

Added some simple case.
The rule should not be triggered if at the function`flat` are [passed some argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#parameters).


crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Promotes the use of `.flatMap()` when `flat().map()` are used together.
+    ///
+    ///
+    /// ## Examples
+    ///
+    /// ### Invalid
+    ///
+    /// ```js,expect_diagnostic
+    /// let arr = ["1", "4", ["3", "6"]];
+    /// arr.flat().map(Number);
```
MichaReiser, line 20: Replacing this with `flatMap` doesn't yield the same result for me: 

```
arr.flatMap(Number);
Array(3) [ 1, 4, NaN ]
```

That's why using `flat` followed by `flatMap` may also be easier to write because it's otherwise necessarily to flat any nested arrays manually?

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// ```
+    ///
+    /// ### Valid
+    ///
+    /// ```js
+    /// let arr = ["1", "4", ["3", "6"]];
+    /// arr.flat(1).map(Number);
```
MichaReiser, line 27: Why is this valid when `flat` isn't? `depth` defaults to 1.

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,102 @@
...
+            state.object().ok()?,
+            state.operator_token().ok()?,
+            JsAnyName::JsName(member),
+        );
+
+        mutation.replace_node(node.clone(), static_member_expression);
+
+        Some(JsRuleAction {
+            mutation,
+            message: markup! {"Replace it with "<Emphasis>".flatMap()"</Emphasis>""}.to_owned(),
```
leops, line 97: Nit: the empty `""` at the end of the markup can be removed since it doesn't print anything

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
leops, line 27: With the correct order of operations `.map().flat(1)` should also be considered invalid by the rule (but not `.map().flat(2)`)

leops (changes requested):
I think the order of operations is reversed, the rule is searching for `.flat().map()` when it should be checking for `.map().flat()`

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
ematipico: I suppose the rule should not be triggered if the `depth` is greater than `1`

ematipico: Yes, I think so.

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
ematipico: Ironically, my first implementation was around `map().flat()` 😆
Thank you for flatting it out

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    fn diagnostic(ctx: &RuleContext<Self>, _state: &Self::State) -> Option<RuleDiagnostic> {
+        let node = ctx.query();
+
+        Some(RuleDiagnostic::new(
+            rule_category!(),
+            node.syntax().text_trimmed_range(),
+            markup! {
+                "The call chain "<Emphasis>".flat().map()"</Emphasis>" can be replaced with a simple "<Emphasis>".flatMap()"</Emphasis>"."
```
leops, line 102: The message also needs to be updated to `.map().flat()`

crates/rome_js_analyze/tests/specs/nursery/useFlatMap.js.snap
```diff
@@ -0,0 +1,63 @@
...
+```
+
+# Diagnostics
+```
+useFlatMap.js:2:1 lint/nursery/useFlatMap  FIXABLE  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+
+  ! The call chain .flat().map() can be replaced with a simple .flatMap().
+  
+    1 │ // invalid
+  > 2 │ [0, [12]].map(Number).flat()
```
leops, line 22: You should add a semicolon between the statements here, otherwise the second array gets parsed as a computed member expression (so the whole `[0, [12]].map(Number).flat()[0, [12], [[16]]].map(Number).flat(1)` becomes a single expression)

ematipico: Are you suggesting that the mutation should do that or the code inside `useFlatMap.js` should have the semicolon?

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        let old_static_member_expression = state.callee().ok()?;
+        let old_static_member_expression =
+            old_static_member_expression.as_js_static_member_expression()?;
+        let member = js_name(ident("flatMap"));
+        let new_member_expression = js_static_member_expression(
+            old_static_member_expression.object().ok()?,
+            old_static_member_expression.operator_token().ok()?,
+            JsAnyName::JsName(member),
+        );
```
MichaReiser, line 119: Nit: How about using `old_static_member_expression.with_member(JsAnyname::JsName(member))` because we're only changing the call member name but want to keep everything else as is.

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+        Some(RuleDiagnostic::new(
+            rule_category!(),
+            node.syntax().text_trimmed_range(),
+            markup! {
+                "The call chain "<Emphasis>".map().flat()"</Emphasis>" can be replaced with a simple "<Emphasis>".flatMap()"</Emphasis>"."
+            },
+        ))
+    }
+
+    fn action(ctx: &RuleContext<Self>, state: &Self::State) -> Option<JsRuleAction> {
```
MichaReiser, line 107: Nit: It may be more understandable if `state` is called `flat_call` instead because that's what the state represents in the context of this rule (makes it clear what `callee` and `arguments` represents)

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+        let new_member_expression = js_static_member_expression(
+            old_static_member_expression.object().ok()?,
+            old_static_member_expression.operator_token().ok()?,
+            JsAnyName::JsName(member),
+        );
+        let new_node = js_call_expression(
+            JsAnyExpression::JsStaticMemberExpression(new_member_expression),
+            state.arguments().ok()?,
+        )
+        .build();
```
MichaReiser, line 124: Nit: How about using flat_call.with_callee(new_member_expression)` instead?

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    fn action(ctx: &RuleContext<Self>, state: &Self::State) -> Option<JsRuleAction> {
+        let node = ctx.query();
+        let mut mutation = ctx.root().begin();
+
+        let old_static_member_expression = state.callee().ok()?;
+        let old_static_member_expression =
+            old_static_member_expression.as_js_static_member_expression()?;
+        let member = js_name(ident("flatMap"));
+        let new_member_expression = js_static_member_expression(
```
MichaReiser, line 115: Nit: Rename to `flat_map_member_expression`

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+            JsAnyExpression::JsStaticMemberExpression(new_member_expression),
+            state.arguments().ok()?,
+        )
+        .build();
+
+        mutation.replace_node(node.clone(), new_node);
+
+        Some(JsRuleAction {
+            mutation,
+            message: markup! {"Replace it with "<Emphasis>".flatMap()"</Emphasis>"."}.to_owned(),
```
MichaReiser, line 130: Nit
```suggestion
            message: markup! {"Replace the chain with "<Emphasis>".flatMap()"</Emphasis>"."}.to_owned(),
```

MichaReiser (changes requested):
Can you update the PR description to reflect that the rule is testing for `.map(...).flat()`

MichaReiser (approved):

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
ematipico: Great tip! I was completely clueless I has these APIs at hand! Thank you

crates/rome_js_analyze/src/analyzers/nursery/use_flat_map.rs
```diff
@@ -0,0 +1,135 @@
...
+                let first_argument = first_argument
+                    .as_js_any_expression()?
+                    .as_js_any_literal_expression()?
+                    .as_js_number_literal_expression()?;
+
+                if first_argument.value_token().ok()?.text_trimmed() != "1" {
+                    return None;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ematipico: I added a part of the suggestion, I left

```rs
let first_argument = first_argument
    .as_js_any_expression()?
    .as_js_any_literal_expression()?
    .as_js_number_literal_expression()?;
```

I like it more 😆

ematipico:
@leops could you please review again the PR?

leops (approved):

PR #3389, opened 2022-10-11
fix(rome_js_parser): Allow `arguments` in `d.ts` files

MichaReiser:
Ambient context has slightly different rules than parsing in a normal context. One such difference is that `arguments` and future reserved keywords are valid identifiers.

Our parser already correctly handled this case by setting the `strict_mode` to `None` when entering an ambient context. However, we initialized `strict_mode` with `Module` for `d.ts` files.

This PR correctly initializes the parser state with `strict = None` if the file is a type script definition file.

## Tests

I added a new parser test and verified that the `jquery.d.ts` can now be formatted

```bash
cargo run --bin rome format ../vscode/extensions/html-language-features/server/lib/jquery.d.ts --write
```

### DIFF
--- a/crates/rome_js_parser/src/state.rs
+++ b/crates/rome_js_parser/src/state.rs
@@ -1,7 +1,7 @@
 use crate::Parser;
 use bitflags::bitflags;
 use indexmap::IndexMap;
-use rome_js_syntax::{Language, SourceType};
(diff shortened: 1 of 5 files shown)


ematipico (approved):
I wonder if we should add an option to our playground, so we can test these cases 🤔

PR #3402, opened 2022-10-11
ci: publish workflow for `@rometools/js-api` package

ematipico:
## Summary

This PR adds a new publish workflow for the new package `@rometools/js-api`.

The majority of the code was copied from the existing workflows. 

## Test Plan

As it's a new package, the only way to test would be to actually merge and see if:
- the nightly release works
- change the `package.json` version will kick a stable release


leops (commented):
Since the new workflow has the `workflow_dispatch` trigger isn't it possible to dispatch it manually from the branch before merging to test it ?

MichaReiser (approved):

PR #3404, opened 2022-10-12
chore: use workspace metadata

ematipico:
## Summary

This is a chore PR, it uses the new [workspace feature](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) to move all the repeated info inside the main `Cargo.toml`. Info that have been moved:
- `authors`
- `edition`
- `license`
- `repository`

I didn't change those crates that were a fork of another project. In those crates, I just changed edition.

## Test Plan

Current CI should pass

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,11 @@ members = [
 	"xtask/libs_bench",
 	"xtask/contributors",
 ]
+[workspace.package]
+edition = "2021"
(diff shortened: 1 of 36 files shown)


ematipico:
Closing. There are some crates that generate issues while running  tests. It seems there some edge cases around cargo that needs to be fixed. I didn't investigate, but it seems the feature is not ready to cover some edge cases.

Issue #3295, opened 2022-09-30
📎 Formatter: Object destructuring


MichaReiser:
### Description

Align our object destructuring heuristic for when to break members with Prettier's heuristic [example](http://play.rome.tools/?code=++++const+%7B%0A++++++id2%2C%0A++++++method%3A+isMethod2%2C%0A++++++methodId2%2C%0A++++%7D+%3D+privateNamesMap.get%28bifornCringerMoshedPerplexSawder%29%3Bf&lineWidth=80&indentStyle=tab&quoteStyle=double&quoteProperties=as-needed&indentWidth=2&sourceType=module&enabledNurseryRules=true&typescript=true&jsx=false#IAAgACAAIABjAG8AbgBzAHQAIAB7AAoAIAAgACAAIAAgACAAaQBkADIALAAKACAAIAAgACAAIAAgAG0AZQB0AGgAbwBkADoAIABpAHMATQBlAHQAaABvAGQAMgAsAAoAIAAgACAAIAAgACAAbQBlAHQAaABvAGQASQBkADIALAAKACAAIAAgACAAfQAgAD0AIABwAHIAaQB2AGEAdABlAE4AYQBtAGUAcwBNAGEAcAAuAGcAZQB0ACgAYgBpAGYAbwByAG4AQwByAGkAbgBnAGUAcgBNAG8AcwBoAGUAZABQAGUAcgBwAGwAZQB4AFMAYQB3AGQAZQByACkAOwBmAA==)

## Test cases

* js/destructuring/destructuring.js
* js/destructuring/issue-5988.js

## Real World use cases

* ant-design: components/avatar/group.tsx https://github.com/ant-design/ant-design/blob/975def492b97a2fc5c1e1fc9eda7f1984b93bb4b/components/avatar/group.tsx#L41
[...]

PR #3405, opened 2022-10-12
feat(rome_js_formatter): Object Pattern Formatting

MichaReiser:
This PR aligns the object pattern formatting with prettier. Mainly the logic when to expand a pattern over multiple lines.

This PR further adds an extension trait for `Option<SyntaxNode<L>>` that implements a `kind` helper. I don't remember how many times I've written `node.parent().map(|p| p.kind())`... This new extension trait now allows to simply write `node.parent().kind()`.

## Tests

I verified that the tests mentioned in #3295 are fixed.

**Average compatibility**: 91.84 -> 91.98
**Compatible lines**: 93.85 -> 93.95


crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -50,69 +50,64 @@ declare_node_union! {
...
-                            )
-                        )
-                    });
-                Ok(has_at_least_a_complex_binding)
+                    .flatten()
+                    .any(|property| match property {
+                        JsObjectAssignmentPatternProperty(_) => true,
+                        JsObjectAssignmentPatternShorthandProperty(short) => short.init().is_some(),
+                        _ => false,
+                    })
```
ematipico, line 73: Should we update the comment? It seems outdated now

crates/rome_js_formatter/src/utils/assignment_like.rs
```diff
@@ -50,69 +50,64 @@ declare_node_union! {
...
-                            )
-                        )
-                    });
-                Ok(has_at_least_a_complex_binding)
+                    .flatten()
+                    .any(|property| match property {
+                        JsObjectBindingPatternProperty(_) => true,
+                        JsObjectBindingPatternShorthandProperty(member) => member.init().is_some(),
+                        _ => false,
+                    })
```
ematipico, line 93: Should we update the comment? It seems outdated now

ematipico (approved):

